PR #101887, opened 2026-04-06
Fix flaky test 01550_create_map_type

groeneai:
### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
...

### What this PR does

With randomized `map_serialization_version = with_buckets`, Map keys may be stored in bucket order rather than insertion order. The test's `WHERE a = map(...)` comparison failed because the Map `=` operator compares internal binary representations — a stored bucketed map ≠ a literal basic-format map.

Instead of pinning serialization settings, uses `mapSort()` on both sides of the comparison to normalize key order, making the equality check order-independent while still allowing all serialization variants to be tested.

### Version info
- Merged into: `26.4.1.596`

PR #102245, opened 2026-04-09
Fix flaky test 01550_create_map_type

groeneai:

- Not required

### What this PR does

Extends PR #101887's approach to the remaining `SELECT *` queries on Map columns and the CAST expression:

- `select * from table_map` → `select d, mapSort(m) from table_map` for `Map(Int8, UInt256)` and `Map(Int8, Int8)` tables
- `SELECT CAST(... 'Map(UInt8, String)') AS map, map[1]` → wrap in `mapSort()`

**Root cause:** The CI randomizer can set `map_serialization_version='with_buckets'` + `map_serialization_version_for_zero_level_parts='with_buckets'` + low `map_buckets_min_avg_size`. Bucketed serialization distributes Map keys across hash buckets, which changes key display order on read-back. For example, Int8 keys `(2,3)` get reordered to `{3:1,2:0}` instead of `{2:0,3:1}` with 4 buckets and constant strategy. The specific pair `(1,2)` currently hashes in-order, but this is a SipHash coincidence that could break with any hash change.

PR #101887 (merged April 7) applied `mapSort()` to the `WHERE` comparison. This PR extends the same approach to the two `SELECT *` queries and the `CAST` expression that also display Map values, completing the hardening.

**Verified:** 50/50 passes with full randomization locally, plus explicit verification with forced `with_buckets` serialization settings.

### Version info
- Merged into: `26.4.1.822`


alexey-milovidov (approved):
This is good - just what we needed.

Avogar (approved):

Avogar:
@groeneai merge master into the branch to get recent CI fixes

groeneai:
Done — merged latest master (448416aa94e).

PR #102254, opened 2026-04-09
Vector similarity indexes: support dot product as distance function

### LINKED ISSUE #101431: Vector similarity index: support the dot product
### Company or project name

ClickHouse Inc.

### Use case

[Vector similarity indexes](https://clickhouse.com/docs/engines/table-engines/mergetree-family/annindexes#vector-similarity-index) currently support the cosine distance and the L2 distance. The dot product and the cosine distance are equivalent on normalized datasets. For all other datasets, it would be nice to support the dot product. In usearch, that corresponds to the `ip_k` metric ("inner product").

Yet another supported distance function in the vector similarity index

_No response_

### Additional context

_No response_

RenzoMXD:
Add `dotProduct` (inner product) as a supported distance function for vector similarity indexes.

Currently, vector similarity indexes support `L2Distance` and `cosineDistance`. This adds `dotProduct`, which maps to usearch's `ip_k` metric. Dot product and cosine distance are equivalent on normalized datasets; for non-normalized datasets, dot product support is valuable.

Additionally, this PR validates the sort direction in `useVectorSearch.cpp`: `L2Distance` and `cosineDistance` require `ORDER BY ... ASC` (smaller = more similar), while `dotProduct` requires `ORDER BY ... DESC` (larger = more similar). Previously, sort direction was not checked, meaning queries with the wrong direction could silently return incorrect results.

Closes #101431.

- New Feature

Added `dotProduct` as a supported distance function for vector similarity indexes. Also validate sort direction for all distance functions to prevent silently incorrect results.

### Version info
- Merged into: `26.4.1.883`


tests/queries/0_stateless/04062_vector_search_dot_product.sql
```diff
@@ -0,0 +1,82 @@
+-- Tags: no-fasttest, no-ordinary-database
```
rschu1ze: This test is copypasta from somewhere else. Please delete it and extend 02354_vector_search_queries.sql properly.

tests/queries/0_stateless/02354_vector_search_queries.sql
```diff
@@ -279,3 +279,48 @@ ORDER BY L2Distance(vec, reference_vec)
 LIMIT 3;
 
 DROP TABLE tab;
+
+SELECT '-- dotProduct distance function';
```
rschu1ze, line 370: Please also extend l. 85ff.

alexey-milovidov:
The flaky check failure is fixed in https://github.com/ClickHouse/ClickHouse/pull/102148, let's update the branch.

tests/queries/0_stateless/02354_vector_search_queries.sql
RenzoMXD: fixed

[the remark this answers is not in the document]
rschu1ze: @RenzoMXD This^^ - the tests are destroyed at the moment.

rschu1ze (approved):

RenzoMXD:
@rschu1ze Thanks for merging this PR.
Can you review my PRs (#100535 , #99877) if you have time?

PR #102255, opened 2026-04-09
Separate text index analysis from `MergeTreeReaderTextIndex`

CurtizJ:
- Improvement

Improved analysis of multiple text indexes used simultaneously in the same query. Also, the `use_skip_indexes_on_data_read` setting is no longer required for the direct read optimization of text indexes.

### Version info
- Merged into: `26.4.1.910`


src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
```diff
@@ -180,13 +190,6 @@ void MergeTreeReaderTextIndex::updateAllMarkRanges(const MarkRanges & ranges)
     }
 }
 
-void MergeTreeReaderTextIndex::prefetchBeginOfRange(Priority priority)
```
CurtizJ, line 183: Unfortunately, we have to sacrifice prefetches for now. Maybe I'll add prefetches for skip indexes in the general case later.

tests/queries/0_stateless/02346_text_index_prefetch.sql
```diff
@@ -1,43 +0,0 @@
--- Tags: no-fasttest, no-parallel-replicas
```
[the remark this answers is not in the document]
CurtizJ: It was a deliberate choice to remove this feature and test.

src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
ahmadov: we can remove this line now.

ahmadov, line 416: seems in case `max_rows_to_read = 0`, granule might not be initialized and crash.

ahmadov (approved):
LGTM 👍

src/Storages/MergeTree/MergeTreeIndexReadResultPool.cpp
```diff
@@ -64,7 +66,7 @@ SkipIndexReadResultPtr MergeTreeSkipIndexReader::read(const RangesInDataPart & p
 
         ProfileEventTimeIncrement<Microseconds> watch(ProfileEvents::FilteringMarksWithSecondaryKeysMicroseconds);
 
-        ranges = MergeTreeDataSelectExecutor::filterMarksUsingIndex(
+        auto [filtered_ranges, filtered_hints] = MergeTreeDataSelectExecutor::filterMarksUsingIndex(
```
Ergus, line 69: IIUC With `use_skip_indexes_on_data_read = 1`, text indexes are now in `applicable_skip_indexes`, causing `filterMarksUsingIndex` to re-read the same granule that was already stored in `read_hints.index_granules` Is this intended?

Ergus (commented):
I only have a question:

For remote-filesystem table policies `MergeTreePrefetchedReadPool` calls `reader->prefetchBeginOfRange(priority)` for every prewhere reader before the I/O window opens.

https://github.com/ClickHouse/ClickHouse/blob/3c6d51510dae93528d2de499e895098e141a8d39/src/Storages/MergeTree/MergeTreePrefetchedReadPool.cpp#L161-L163

Now every text index read becomes a synchronous, unmasked latency event (because there is not override) Is this intended?
This PR intentionally drops the prefetch on remote filesystems?

src/Storages/MergeTree/MergeTreeIndexReadResultPool.cpp
CurtizJ: No, we don't read granule twice. With `use_skip_indexes_on_data_read = 1` we shouldn't fill `read_hints.index_granules`, because index analysis is not called in `MergeTreeDataSelectExecutor::filterPartsByPrimaryKeyAndSkipIndexes`.

I can add a test for that case.

CurtizJ: I've added a test. It looks ok.

CurtizJ:
> This PR intentionally drops the prefetch on remote filesystems?

Yes, this is intended. Prefetch in the text index only worked for the sparse index file. Now its analysis has been moved to the common code path in `filterMarksUsingIndex`, which does not have prefetch infrastructure yet. This is not a big problem, because reading tokens from the dictionary typically contributes more latency.

I have a prototype of prefetch support for skip indexes in general and will open a PR soon. However, we need to be careful with prefetches of skip indexes, as they can slow down queries. I also want to add prefetches for the text index dictionary later, but that requires a small refactoring.

Ergus (approved):

zdyj3170101136:
please backport it to 26.2

rschu1ze:
@zdyj3170101136 Only bugfixes get backported, this PR is feature work.

PR #102256, opened 2026-04-09
Fix crash in _post_run when network is unreachable during GH auth

maxknv:
When the network becomes unreachable after a job completes, `_GH_Auth` raises an unhandled `RuntimeError` that kills the entire `_post_run` method. This prevents critical post-run steps from executing, including writing `pipeline_status` which GH Actions needs to determine the job result.

Make `_GH_Auth` catch exceptions and return a boolean so post-run callers can gracefully skip GH API calls instead of crashing. The pre-run call site still fails hard since auth is required before the job runs.

...

### Version info
- Merged into: `26.4.1.735`

### DIFF
--- a/ci/praktika/runner.py
+++ b/ci/praktika/runner.py
@@ -32,14 +32,19 @@ _GH_authenticated = False
 def _GH_Auth():
     global _GH_authenticated
     if _GH_authenticated:
-        return
+        return True
     if not Settings.USE_CUSTOM_GH_AUTH:
-        return
+        return True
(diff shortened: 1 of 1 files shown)


leshikus (approved):
better to add the type here if you change this `-> bool`

ci/praktika/runner.py
```diff
@@ -32,14 +32,19 @@
 def _GH_Auth():
```
leshikus, line 32: here ^^

PR #102283, opened 2026-04-09
Fix assertion failure when table UUID collides with database UUID

groeneai:
In `DatabaseCatalog::getTableImpl`, when a table lookup by UUID finds a
database mapping `{database, nullptr}` instead of a table mapping
`{database, table}`, the assertion `!db_and_table.first &&
!db_and_table.second` fires and crashes the server (debug builds only).

This happens when a user (or fuzzer like BuzzHouse) creates a table with
a UUID that matches an existing database UUID. The `ON CLUSTER` path
through `DDLWorker::processTask` → `tryGetTable` triggers the lookup
before executing the DDL, which hits the assertion.

### Root cause

Databases with UUIDs (e.g., `Atomic` engine) register their UUID in
`DatabaseCatalog`'s UUID map as `{database_ptr, nullptr}` (line 618).
This is a valid state — the "table" slot is null because the entry
represents a database, not a table.

The assertion at line 391 assumed that if either pointer is null, both
must be null. This invariant is violated by database UUID mappings.

### Fix

Handle the `{database, nullptr}` case explicitly before the assertion:
when a UUID lookup finds a database (not a table), treat it as "table
not found" and return gracefully with an `UNKNOWN_TABLE` error.

The assertion is preserved for any other unexpected states.

### How to reproduce

```sql
CREATE DATABASE d0 UUID '00000000-0000-0000-0000-000000000001' ENGINE = Atomic;
CREATE TABLE d0.t0 UUID '00000000-0000-0000-0000-000000000001'
    ON CLUSTER 'default' (c0 Int) ENGINE = MergeTree() ORDER BY tuple();
-- Server crashes on debug builds
```

Fix server crash (assertion failure) when creating a table with a UUID that collides with an existing database UUID via `ON CLUSTER`.

Closes #98615
Related: #98861 (stalled fix PR)

---
> I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en


alexey-milovidov:
@groeneai, the PR https://github.com/ClickHouse/ClickHouse/pull/98861 is not stale - I reviewed it and am going to merge it. We prefer PRs from external contributors if possible.

PR #102011, opened 2026-04-08
Fix exception in throwUnexpectedDataAfterParsedValue for empty Tuple() type

alexey-milovidov:
`throwUnexpectedDataAfterParsedValue` calls `popBack(1)` assuming the parsed value is already committed to the column. For tuples, it is called from inside `addElementSafe`'s impl lambda before `addSize(1)` commits the row. For `Tuple()` with no sub-columns, `column.size()` returns 0, so `popBack(1)` throws "Cannot pop 1 rows from Tuple(): there are only 0 rows".

Fix by calling `addSize(1)` to commit the row before `throwUnexpectedDataAfterParsedValue`, which will serialize it for the error message and then pop it back.

Found by the AST fuzzer: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102000&sha=9190a5549402d12fff136f61209c5405f2aa896c&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%2C%20targeted%2C%20old_compatibility%29

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix exception when casting a string with trailing data to empty `Tuple()` type.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.744`

PR #102289, opened 2026-04-09
Fix logical error when parsing incorrect empty tuples

nihalzp:
The following query throws exception:

```sql
SELECT CAST('()()', 'Tuple');
```

Currently, we increment empty tuple column size after `addElementSafe`'s `impl` has been executed, leading to incorrect column size during validation that happens at the end of `impl` of `SerializationTuple::deserializeTextImpl`. Since `impl` populates the tuple subcolumns, it feels natural that `impl` should also update the column size.

Reverts https://github.com/ClickHouse/ClickHouse/pull/102011.

Fix logical error when parsing incorrect empty tuple string.

### Version info
- Merged into: `26.4.1.1067`


alexey-milovidov:
The flaky check failure is fixed in https://github.com/ClickHouse/ClickHouse/pull/102148, let's update the branch.

Avogar (changes requested):
It's actually better to not use `throwUnexpectedDataAfterParsedValue` in `serializationTuple`, because for Tuple we have special logic of rollback in canse of error during parsing. `throwUnexpectedDataAfterParsedValue` should be used for simple columns with no nested structure that should be rolled back carefully

nihalzp:
I can try removing `throwUnexpectedDataAfterParsedValue` from `serializationTuple`. Apart from that, as changed in this PR, should `impl` of `addElementSafe` have the responsibility to update size via `addSize`?

Avogar:
> should impl of addElementSafe have the responsibility to update size via addSize?

I would say it's ok to keep `addSize` in `addElementSafe` and not in `impl`. Because in case of error in `impl` we also need to check if we added size of not, with current approach in case of error in `impl` we know that we didn't add the size

src/DataTypes/Serializations/SerializationTuple.cpp
```diff
@@ -261,12 +262,22 @@ ReturnType SerializationTuple::deserializeTextImpl(IColumn & column, ReadBuffer
...
+                /// If empty tuple, temporarily increase size to make sure we can read the parsed
+                /// value via serializeText.
+                if (elems.empty())
+                    assert_cast<ColumnTuple &>(column).addSize(1);
+                WriteBufferFromOwnString ostr;
+                serializeText(column, column.size() - 1, ostr, settings);
+
+                /// Revert the temporarily added size increment for empty tuple.
+                if (elems.empty())
+                    assert_cast<ColumnTuple &>(column).popBack(1);
```
Avogar, line 274:
```suggestion
                assert_cast<ColumnTuple &>(column).addSize(1);
                WriteBufferFromOwnString ostr;
                serializeText(column, column.size() - 1, ostr, settings);

                /// Revert the temporarily added size increment for empty tuple.
                assert_cast<ColumnTuple &>(column).popBack(1);
```

The condition `elems.empty()` here is not necessary I believe. We can add size and do pop back as we are 100% sure all elements were deserialized (otherwise we would get exception earlier)

Avogar (approved):

src/DataTypes/Serializations/SerializationTuple.cpp
nihalzp: Yes, even without `elems.empty()`, the results will be correct because `addElementSafe` has this guard `if (element_column.size() > old_size)` before rollback.

The reason I kept `if (elems.empty())` because for non-empty tuple, `popBack(1)` is not inverse of `addSize(1)`. `popBack` also pops from the subcolumns; and as a result, we will be doing rollback in `impl` instead of doing the rollback in the correct place which is `addElementSafe`.

If you have strong preference for removal of `elems.empty()`, I can remove it.

Avogar (approved):

PR #102302, opened 2026-04-09
Fix UBSan error in Lexer::nextToken: null pointer arithmetic

alexey-milovidov:
When `Lexer` is constructed with a null buffer (`begin` = nullptr) and non-zero `max_query_size`, the expression `begin + max_query_size` in `Lexer::nextToken` is undefined behavior (adding a non-zero offset to a null pointer). This was triggered by the `FunctionsStress.stress` test in CI.

The fix adds a null check for `begin` before the pointer arithmetic.

- Not For Changelog

Fix undefined behavior in `Lexer` when constructed with a null buffer and non-zero `max_query_size`.


alexey-milovidov (commented):
Looks good.

CheSema (approved):

CheSema:
Some how the code with the fix is vanished from this PR.
The issue has been secured in https://github.com/ClickHouse/ClickHouse/pull/103489
The similar tests has been added here already.

Issue #52121, opened 2023-07-14
Cannot access named tuple element by its name using dot from Kafka table


ekpdt:
**Describe what's wrong**

Selecting a named tuple element using the dot notation from a Kafka table does not work. This functionality is useful for materializations from a Kafka table to a MergeTree.

Querying without dot works:
https://fiddle.clickhouse.com/0e93effa-ffde-4d1c-b393-a71682eb30f4

Querying the tuple element with dot fails:
https://fiddle.clickhouse.com/ba621665-0f19-4b55-a0af-e81dac7b82b7

Querying the tuple element with `tupleElement` works:
https://fiddle.clickhouse.com/09b02535-2558-4d93-9b80-2e1db6a84278

**Does it reproduce on recent release?**

Tested on 23.6.2.18.

**Enable crash reporting**

N/A

**How to reproduce**

```
CREATE TABLE users (foo Tuple(a String, b String)) 
  ENGINE=Kafka 
  SETTINGS 
  kafka_broker_list = 'foo', 
  kafka_topic_list = 'foo', 
  kafka_group_name = 'foo',
  kafka_format = 'JSONEachRow';

CREATE TABLE mt (a String)
  ENGINE = MergeTree
  ORDER BY tuple();

CREATE MATERIALIZED VIEW mv TO mt AS
  SELECT foo.a FROM users;

SELECT 'OK';
```

**Expected behavior**

Tuple navigation with dot should work like `tupleElement`

**Error message and/or stacktrace**

```
Received exception from server (version 23.6.2):
Code: 47. DB::Exception: Received from localhost:9000. DB::Exception: Missing columns: 'foo.a' while processing query: 'SELECT foo.a FROM default.users', required columns: 'foo.a'. (UNKNOWN_IDENTIFIER)
(query: CREATE MATERIALIZED VIEW mv TO mt AS
  SELECT foo.a FROM users;)
```

**Additional context**

Possibly related: https://github.com/ClickHouse/ClickHouse/issues/35368

BorisTyshkevich:
The problem is related to Materialized View; it is not related to Kafka. 

And it works for simple select.

https://fiddle.clickhouse.com/2c6d9e08-c5c5-408c-a8a4-427e020328d2

[...]

PR #102303, opened 2026-04-09
Fix flaky test 04092_kafka_named_tuple_dot_access

groeneai:
### Changelog category

### Changelog entry
Fix flaky test 04092_kafka_named_tuple_dot_access

### What this PR does

Fixes the flaky test `04092_kafka_named_tuple_dot_access` which was failing on master with two distinct failure modes:

1. **`having stderror` (amd_msan, amd_debug, amd_asan_ubsan):** The Kafka background consumer tries to connect to the non-existent broker and produces `<Error>` level log messages. These get forwarded to client stderr via `send_logs_level` (default: `warning`), which the test runner flags as a failure.

2. **`return code: 159` (arm_asan_ubsan):** `DROP TABLE` hangs for 300s because the consumer is blocked on DNS resolution of the unresolvable hostname `'foo'`.

**Changes:**
- `SET send_logs_level = 'fatal'` — suppresses expected Kafka consumer connection errors from reaching client stderr. This is a well-established pattern used in 36+ stateless tests (e.g., `02479_mysql_connect_to_self.sql`, `04038_max_skip_unavailable_shards.sql`).
- `kafka_broker_list = 'localhost:10000'` (was `'foo'`) — uses a proper resolvable hostname, eliminating the DNS timeout that caused DROP TABLE to hang.

**Verification:**
- Without fix: 10/10 runs produce stderr (confirmed FAIL)
- With fix: 50/50 runs pass with full randomized settings

### Version info
- Merged into: `26.5.1.716`


groeneai:
## Pre-PR Validation Gate (Session: cron:clickhouse-ci-task-worker:20260409-231500)

**a) Deterministic repro?**
Yes — the old `.sql` test deterministically produces stderr from Kafka DNS errors on `amd_msan` and times out on `arm_asan_ubsan` (300s DROP TABLE hang). CIDB: 2 master failures on April 9. Not reproducible locally due to different server logging config and faster local networking, but CI failures are consistent.

**b) Root cause explained?**
The `.sql` test creates a Kafka table with `kafka_broker_list = 'foo'` (unresolvable hostname). The ClickHouse server immediately starts a background Kafka consumer that attempts DNS resolution → fails → logs errors to stderr. The test runner fails any test with non-empty stderr. Separately, on DROP TABLE, the consumer shutdown path tries to gracefully close the rdkafka consumer against an unreachable broker, which hangs for up to 300s on slow sanitizer builds.

Yes — using `clickhouse-local` eliminates both root causes by design: no persistent background consumer starts (no stderr), and no consumer needs shutdown (no DROP hang). `0.0.0.0:9092` resolves instantly. `2>/dev/null` handles any residual stderr.

**d) Test intent preserved?**
Yes — the test still creates a Kafka table with a `Tuple(a String, b String)` column, creates a MV with `SELECT foo.a AS a FROM users`, and verifies it succeeds with `SELECT 'OK'`. The regression for #52121 (UNKNOWN_IDENTIFIER on named tuple dot access) is still covered.

- FAILS without fix: 2 master failures in CIDB today (amd_msan: stderr, arm_asan_ubsan: timeout)
- PASSES with fix: 30/30 local test runner passes with randomized settings

groeneai:
cc @antaljanosbenjamin — could you review this test fix? The test from our PR #101975 (Kafka named tuple dot access regression test) was causing master failures due to DNS resolution errors in stderr and DROP TABLE timeouts when using a fake broker. Converted to use `clickhouse-local` following the same pattern as `03523_storage_kafka_shutdown_smoke_local`.

alexey-milovidov (approved):
Yes, this is ok.

antaljanosbenjamin (commented):
If we use a proper host name, why cannot we just use a normal SQL test to check this? In that case:
- no DNS errors
- no hang because of unreachable host

I think this is not the right way to fix this. @groeneai

groeneai:
@antaljanosbenjamin You're right — using a proper hostname is the simpler fix. I've reverted the `clickhouse-local` / `.sh` conversion and kept it as a normal `.sql` test.

The only change from the original is `kafka_broker_list = 'localhost:9092'` instead of `'foo'`. This follows the same pattern as `01821_table_comment.sql` which uses `localhost:10000`.

With a resolvable hostname:
- No DNS errors → no stderr
- Connection failures are fast (refused, not DNS timeout) → no DROP TABLE hang

30/30 passes locally via `clickhouse-test --test-runs 30`.

groeneai:
@antaljanosbenjamin Thanks for the feedback. You were right that we should keep this as a normal SQL test with a proper hostname. However, `localhost:9092` was still producing stderr errors in CI flaky checks ("Error during draining: Broker: Unknown topic or partition") — likely because the CI environment has a Kafka broker on port 9092 for integration tests.

**v3 fix:** Changed to `localhost:10000` instead. This follows the exact pattern from test `01821_table_comment` (line 35) which also creates a Kafka table with `kafka_broker_list = localhost:10000` and has **zero master failures in 30 days**. Port 10000 has nothing listening, so:
- `localhost` resolves instantly (no DNS error)
- Connection is immediately refused — no "Unknown topic or partition" stderr
- Consumer shutdown is fast (no active connection to drain)

Also rebased onto fresh master for a clean single-commit diff. Verified 50/50 passes locally with full randomization.

groeneai:
@antaljanosbenjamin v4: Found the root cause and a much simpler fix.

**Root cause:** The Kafka background consumer produces `<Error>` level log messages when connecting to the non-existent broker (`Connection refused`, `Invalid response size`). These get forwarded to the client's stderr via `send_logs_level` (default: `warning`), which the test runner then flags as `having stderror`.

This happens with ANY fake broker hostname — `foo`, `localhost:9092`, `localhost:10000` — because the Kafka consumer always tries to connect and logs errors regardless.

**Fix:** `SET send_logs_level = 'fatal'` at the top of the test. This is a well-established pattern used in 36+ stateless tests for exactly this situation (e.g., `02479_mysql_connect_to_self.sql`: *"failed connection tries are ok, if it succeeded after retry"*).

Combined with `localhost:10000` (proper hostname, no DNS timeout on DROP), this keeps the test as a normal `.sql` test while preventing expected consumer errors from reaching stderr.

**Verification:**
- Without fix: 10/10 runs produce stderr (confirmed FAIL)
- With fix: 50/50 runs pass with full randomized settings

Session: cron:clickhouse-ci-task-worker:20260410-224500

antaljanosbenjamin (approved):

groeneai:
Status check (session: cron:clickhouse-ci-task-worker:20260419-104500) — approved 2x and MERGEABLE, only blocker is an unrelated chronic-flaky test.

**Failing CI check:** `Stateless tests (amd_asan_ubsan, distributed plan, parallel, 2/2)` — single test failure `02516_projections_and_context` (2.2s). This test is **chronic flaky across many unrelated PRs**:

```
CIDB 30d for 02516_projections_and_context (FAIL|ERROR):
  PR 96130 (CI randomization):   4
  PR 102303 (ours):              1
  PR 99833, 97547, 100427, 97238, 0 (master): 1 each
  Total: 10 failures / 7 distinct PRs, 1 master hit
```

Not related to the Kafka fix — different code path, different component.

**Our actual test `04092_kafka_named_tuple_dot_access` on PR head `5511ca4e`:** 333 OK runs across all variants + 150 flaky-check runs × 50 reps, zero FAIL.

**Master status for 04092:** clean since 2026-04-09 — 46,314 OK / 0 FAIL across all configurations in the last 10 days. Fix still warranted because the v4 change (`send_logs_level=fatal` + `localhost:10000`) makes the test deterministic under the stderr-forwarding and DNS-timeout regimes that caused the April 9 failures.

Could a maintainer please re-run the flaky check on this PR or proceed to merge? Happy to rebase if preferred.

PR #102304, opened 2026-04-09
Native Common Virtuals for Remaining Storage

Michicosun:

Unfortunately, we need to support this for backward compatibility.

### Version info
- Merged into: `26.4.1.833`


KochetovNicolai (approved):
lgtm

Michicosun:
`FunctionsStress.stress` is broken. 

There are 2 problems I discovered:
1. The printf function faces a problem in the fmtlib when a string with a single control character like `%` is passed.
2. `FunctionBaseDynamicAdaptor` does not implement `isInjective`

PR #102316, opened 2026-04-10
Fix exception in system log flush when table has extra columns

alexey-milovidov:
The `SystemLog::flushImpl` method creates an INSERT query without an explicit column list. When the system log table has been altered to include extra columns (e.g. by a fuzzer or a user), the INSERT pipeline expects all table columns, but the block only contains the columns from `getColumnsDescription`. This causes a "Invalid number of columns in chunk pushed to OutputPort. Expected 23, found 22" exception.

Fix by explicitly specifying column names in the INSERT query, so that extra columns added to the table get their default values instead of causing a mismatch.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=a4b72fe91e7c174e0a5347a49d32b5b3e5719d47&name_0=MasterCI&name_1=BuzzHouse%20%28amd_tsan%29

...

### Version info
- Merged into: `26.5.1.216`


tests/queries/0_stateless/04093_system_log_alter_column.sql
```diff
@@ -0,0 +1,18 @@
+-- Verify that system log tables can be flushed after ALTER TABLE adds a column.
+-- This reproduces the bug where SystemLog::flushImpl would fail with
+-- "Invalid number of columns in chunk pushed to OutputPort" because the INSERT
+-- did not specify an explicit column list.
+
+SET log_processors_profiles = 1;
+
+SELECT 1 FORMAT Null;
+SYSTEM FLUSH LOGS;
```
tiandiwonder: Should be SYSTEM FLUSH LOGS processors_profile_log

tests/queries/0_stateless/04093_system_log_alter_column.sql
```diff
@@ -0,0 +1,18 @@
+-- Verify that system log tables can be flushed after ALTER TABLE adds a column.
```
tiandiwonder: Missing -- Tags: no-parallel
  The test does ALTER TABLE system.processors_profile_log ADD COLUMN / DROP COLUMN on a global system
  table. Without the tag, it races with any concurrent test that flushes logs.

tiandiwonder (approved):
better to test query_log as well.

alexey-milovidov (commented):
Now ok.

alexey-milovidov:
Not quite... I'll have to convert to an sh test and configure clickhouse-local accordingly...

PR #84088, opened 2025-07-21
Better test `02932_set_ttl_where`

alexey-milovidov:
### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

Made it ten times lighter and enabled ordinary databases.

PR #102318, opened 2026-04-10
Reduce data volume in test 02932_set_ttl_where to prevent MSAN timeout

groeneai:
The test `02932_set_ttl_where` verifies that TTL with a WHERE clause containing a subquery
works correctly (no crashes, correct execution). It inserts data, waits for TTL expiry, then
runs `OPTIMIZE TABLE FINAL` which triggers TTL evaluation.

Under MemorySanitizer (MSAN), the `OPTIMIZE TABLE FINAL` with 200K rows and a 10K-entry IN
subquery exceeds the 300-second test timeout due to sanitizer instrumentation overhead on the
per-row TTL WHERE evaluation during merge. This was observed on master (April 9, `amd_msan`
check) and across 4 distinct PRs over 30 days.

The test was previously reduced from 1M to 100K rows in PR #84088, but this is still too much
for MSAN builds. Since this is a correctness test (not performance), further reduction is safe.

Changes:
- Reduce insert volume: 100K → 5K rows per batch (10K total, 20x reduction)
- Reduce TTL WHERE subquery set: 10K → 1K entries

The TTL WHERE + subquery evaluation path is still fully exercised. Verified 50/50 passes
locally with full randomization.

...

### Version info
- Merged into: `26.5.1.739`


groeneai:
**Pre-PR Validation Gate** (session `cron:clickhouse-ci-task-worker:20260410-021500`)

**a) Deterministic repro?** The failure is a deterministic timeout on MSAN builds — `OPTIMIZE TABLE FINAL` with 200K rows + 10K-entry TTL WHERE subquery exceeds the 300s test timeout under MSAN instrumentation overhead. Cannot reproduce the timeout on non-MSAN builds (completes in 0.08s on debug), but the root cause is clear: data volume × MSAN overhead.

**b) Root cause explained?** During `OPTIMIZE TABLE FINAL`, the merge engine evaluates the TTL WHERE clause (`a IN (SELECT number FROM system.numbers LIMIT 10000)`) for every row. With 200K rows, MSAN's per-memory-operation instrumentation causes this evaluation to exceed the 300s test timeout. The subquery materializes a 10K-entry Set, then each row's membership is checked against it — all under MSAN tracking.

**c) Fix matches root cause?** Yes — reduces the data volume that MSAN must instrument. Inserts reduced from 100K → 5K per batch (10K total, 20x reduction), subquery set from 10K → 1K entries (10x reduction). Combined ~200x reduction in per-row evaluation work.

**d) Test intent preserved?** Yes. The test verifies that TTL with WHERE clause containing a subquery executes correctly (no crash, no error). It has no assertions about row counts, performance, or optimization behavior. The feature is equally well tested with 10K rows as with 200K.

**e) Demonstrated in both directions?** The original test times out on MSAN (CI evidence: master failure April 9, `amd_msan` check). The reduced test passes 50/50 locally with full randomization (debug build, ~3.2s per run including 3s of sleeps).

groeneai:
cc @CurtizJ — could you review this? Reduces data volume in test 02932_set_ttl_where (100K→5K rows, 10K→1K subquery entries) to prevent OPTIMIZE TABLE FINAL from timing out on MSAN builds. The test verifies TTL WHERE with subquery correctness and doesn't need large data.

alexey-milovidov (approved):
No problems with this change.

Issue #99226, opened 2026-03-10
Avoid thundering herd effect in the query result cache


rschu1ze:
### Company or project name

ClickHouse Inc.

### Describe the unexpected behaviour

The [query result cache](https://clickhouse.com/docs/operations/query-cache) is susceptible to the [thundering herd effect](https://en.wikipedia.org/wiki/Cache_stampede) (dog piling, cache stampede). When the same query runs multiple times, the first run's result is cached. At the end of the TTL (e.g. 20 sec), the entry becomes stale and new (still identical) queries see cache misses. All these queries will now re-calculate the result and the first query adds it to the cache. It would have been sufficient if only the first query ran and inserted the result.

The thundering herds effect is a missed optimization opportunity, i.e. calculating the query result just once instead of N times. It is not about avoiding potential extra work compared to what the system would do anyways without caching at all (we'd also calculate the result N times as well). It will still be nice to avoid thundering herds.

`CacheBase::getOrSet` implements a token mechanism to prevent it. The query result cache currently uses `CacheBase::set` and `CacheBase::getWithKey`.

Two steps are needed to fix this:
- obviously, the query result cache must use `getOrSet`,
- `getOrSet` accepts a lambda which performs the actual work. "Work" here means the query. We need to change _when_ cache inserts happens.
    - Today, the cache insert is done at the end of a SELECT query (see `finalizeQueryPipelineBeforeLogging` in src/Interpreters/executeQuery.cpp).
    - `getOrSet` should instead be called before the query. It can hash the query AST to calculate the cache key and find out if we need to insert the entry at all. If yes, it starts the query. Parallel runs of the same query will block until the calculation is complete and afterwards return the cached result.

### How to reproduce

Run the same query multiple times at once.

rschu1ze:
Sure, done!

PR #102327, opened 2026-04-10
Add QueryBatchingQueue for batching identical concurrent SELECT queries

### LINKED ISSUE #102306: Query Batching (Aggregation)
### Company or project name

ClickHouse

### Use case

In some scenarios, ClickHouse has to process a high number of concurrent, but almost identical queries. This could be accelerated by combining these almost identical queries into batches in a time period, by allowing a certain delay, and executing common parts of these queries only once.

The simplest scenario is when queries contain a fully identical subquery:

```
WITH t AS (SELECT 'this is a common subquery')
SELECT ...
```

We can implement batching at the query cache level and reuse the upcoming "query cache for subqueries" feature: #99804.

Actually, nothing will be needed on top: you just set up the cache usage and the maximum freshness for a subquery, and it will be automatically batched. Our lock to prevent concurrent cache generation, which is developed in https://github.com/ClickHouse/ClickHouse/issues/99226 will ensure that concurrent queries wait for the first one.

A more advanced scenario is when we have many similar queries:

```
SELECT count() FROM table WHERE id = 123;
SELECT count() FROM table WHERE id = 456;
...
```

In this scenario, we can build on top of the upcoming "query rewrite rules" feature: #88234.

But we have to extend it a little.

First, we should allow collecting parameters into arrays by declaring `Array(T)` parameter in a rewritten query when `T` is in the original query. Example:

```
CREATE RULE rule_name AS
(
    SELECT count() FROM table WHERE id = {x:UInt64};
)
REWRITE TO
(
    SELECT c FROM
    (
        SELECT id, count() AS c FROM table WHERE id IN {x:Array(UInt64)} GROUP BY id
    )
    WHERE id = {x:UInt64};
)
SETTINGS batch_max_queries = 100, batch_max_delay_ms = 1000;
```

When applied normally, it will just put a single element into an array. So we will have an array with just one element.

But we can extend rewrite rules to support batching with `batch_max_queries` and `batch_max_delay_ms` settings. When the rewrite rule activates for a query, it will generate rewritten queries based on the aggregation of all incoming queries in the batch. This aggregation populates Array parameters in the rewritten query.

For example, queries
```
SELECT count() FROM table WHERE id = 123;
SELECT count() FROM table WHERE id = 456;
```
received in a short time period, will be rewritten to:
```
SELECT c FROM (SELECT id, count() FROM table WHERE id IN [123, 456] GROUP BY id) WHERE id = 123;
SELECT c FROM (SELECT id, count() FROM table WHERE id IN [123, 456] GROUP BY id) WHERE id = 456;
```
As we see, the inner subquery is identical for them.

Now they can be run together using the query cache mechanism.

This is an ad-hoc solution, which allows fine customization of which queries will be batched and how.

We are considering a generic solution on the query plan level. But it is way harder to implement.

When thinking about how to express the aggregating parameters, it might be tempting to use `...` syntax, like in C++ variadic templates or in JavaScript array generators. But we already have arrays, and it looks perfect. But there is a problem - it's unclear how to collect parameters from various places into an array of tuples... Maybe we should come back to the `...` syntax.

The delay could be adaptive, similar to our adaptive async inserts, so that if queries are rare, they will be processed with no delay instead of waiting the full timeout.

Batching should be limited to a user (so the batches are collected independently for every user).
They don't have to be split by groups of different settings, as it is handled on the query cache level.

### Additional context

_No response_

Krishnachaitanyakc:
Add a `QueryBatchingQueue` that deduplicates identical concurrent SELECT queries using a leader/follower pattern built on the existing `QueryResultCache`.

When `use_query_batching = 1`, concurrent identical queries (same full AST + user ID) are grouped:
- The first query ("leader") executes normally with `use_query_cache = 1`, writing its result to the query result cache
- Subsequent queries ("followers") wait on a `std::shared_future<void>` for the leader to finish, then read the result from the query result cache
- Batches are flushed after `query_batching_max_wait_ms` (default 100ms) or `query_batching_max_batch_size` (default 100 queries)

Closes #102306

- Experimental Feature

Add `use_query_batching` setting to deduplicate identical concurrent SELECT queries. The first query executes and caches its result; subsequent identical queries read from the query result cache instead of re-executing.

New settings:
- `use_query_batching` (Bool, default `false`) — enable query batching for SELECT queries
- `query_batching_max_wait_ms` (Milliseconds, default `100`) — maximum time a follower waits for the leader to finish
- `query_batching_max_batch_size` (UInt64, default `100`) — maximum queries per batch before flushing

Example:
```sql
SELECT count() FROM large_table SETTINGS use_query_batching = 1;
```

When multiple clients issue this same query concurrently, only one execution occurs — the rest read from the query result cache.

### DIFF
--- a/programs/local/LocalServer.cpp
+++ b/programs/local/LocalServer.cpp
@@ -1073,6 +1073,9 @@ void LocalServer::processConfig()
     /// Initialize a dummy query result cache.
     global_context->setQueryResultCache(0, 0, 0, 0);
 
+    /// Initialize query batching queue.
+    global_context->setQueryBatchingQueue();
+
     /// Initialize allowed tiers
     global_context->getAccessControl().setAllowTierSettings(server_settings[ServerSetting::allow_feature_tier]);
 
--- a/programs/server/Server.cpp
+++ b/programs/server/Server.cpp
@@ -2099,6 +2099,9 @@ try
     }
     global_context->setQueryResultCache(query_result_cache_max_size_in_bytes, query_result_cache_max_entries, query_result_cache_max_entry_size_in_bytes, query_result_cache_max_entry_size_in_rows);
 
+    /// Initialize query batching queue for batching identical concurrent SELECT queries.
+    global_context->setQueryBatchingQueue();
+
 #if USE_EMBEDDED_COMPILER
     size_t compiled_expression_cache_max_size_in_bytes = server_settings[ServerSetting::compiled_expression_cache_size];
     size_t compiled_expression_cache_max_elements = server_settings[ServerSetting::compiled_expression_cache_elements_size];
--- a/src/Core/Settings.cpp
+++ b/src/Core/Settings.cpp
@@ -5426,6 +5426,34 @@ Possible values:
 
 - Any string
 )", 0) \
+    DECLARE(Bool, use_query_batching, false, R"(
+If turned on, identical concurrent `SELECT` queries from the same user will be batched together.
+The first query in the batch executes and writes its result into the query result cache.
+Subsequent queries in the batch read from the cache, avoiding redundant computation.
+
+Requires the query result cache to be enabled at the server level.
+
+Possible values:
+
+- 0 - Disabled
+- 1 - Enabled
+)", 0) \
+    DECLARE(Milliseconds, query_batching_max_wait_ms, 100, R"(
+Maximum time in milliseconds to wait for additional queries to join a batch before executing it.
+Only relevant when `use_query_batching` is enabled.
+
+Possible values:
+
+- Positive integer >= 0.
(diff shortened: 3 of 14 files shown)


alexey-milovidov:
The task https://github.com/ClickHouse/ClickHouse/issues/102306 mentions at least two other PRs that we have to merge first.

alexey-milovidov:
> When use_query_batching = 1, concurrent identical queries (same full AST + user ID) are grouped

It should be the default operation of the query cache - lock on the processing query for the same key. We discussed it in some other issue.

alexey-milovidov:
Here: https://github.com/ClickHouse/ClickHouse/issues/99226

PR #102328, opened 2026-04-10
Cap map serialization for zero-level parts at the base version

groeneai:
`map_serialization_version = 'basic'` on a MergeTree table should guarantee that
all parts — including zero-level parts created by INSERT — use basic (insertion-order)
serialization. However, `map_serialization_version_for_zero_level_parts` could independently
override this to `with_buckets`, causing Map keys to be reordered by hash bucket.

This happened in CI when the test runner randomized `map_serialization_version_for_zero_level_parts`
independently of `map_serialization_version`, producing chronic flaky failures in test
`03993_map_subcolumns_small_compact` (16+ PRs affected over 30 days, 2 master failures).

The fix ensures the effective zero-level map serialization is
`min(map_serialization_version, map_serialization_version_for_zero_level_parts)` in both
the regular and projection write paths in `MergeTreeDataWriter`. This enforces the invariant
that zero-level parts never use a more complex serialization than the base setting — matching
the original design intent of the zero-level override (same-or-simpler, never more complex).

The intended use case for `map_serialization_version_for_zero_level_parts` is:
- `map_serialization_version = 'with_buckets'` + `map_serialization_version_for_zero_level_parts = 'basic'`
  (fast inserts with basic serialization, bucketed after merge) — this continues to work unchanged.

The invalid combination blocked by this fix:
- `map_serialization_version = 'basic'` + `map_serialization_version_for_zero_level_parts = 'with_buckets'`
  (basic for merged parts but bucketed for inserts) — this is now capped to basic for both, which
  matches what `map_serialization_version = 'basic'` promises.

Ref: PR #101024 (closed) identified the same issue but used a different approach.

### Changelog entry:
Fix `map_serialization_version = 'basic'` not being respected for zero-level parts when
`map_serialization_version_for_zero_level_parts` was independently set to `'with_buckets'`,
which caused Map keys to be reordered by hash bucket instead of preserving insertion order.


groeneai:
**Pre-PR Validation Gate** (session: cron:clickhouse-ci-task-worker:20260410-041500)

**a) Deterministic repro:** YES — `clickhouse client --allow_merge_tree_settings --map_serialization_version_for_zero_level_parts with_buckets --max_buckets_in_map 4 --map_buckets_strategy constant --map_buckets_coefficient 1.0 --map_buckets_min_avg_size 0 -q "CREATE TABLE t (...) SETTINGS map_serialization_version='basic'..."` deterministically produces reordered keys (`{'d':200,'c':100}` instead of `{'c':100,'d':200}`).

**b) Root cause:** `map_serialization_version_for_zero_level_parts` is passed independently of `map_serialization_version` to `SerializationInfo::Settings` in `MergeTreeDataWriter::writeTempPartImpl` (and projection path). When CI randomizer sets `map_serialization_version_for_zero_level_parts='with_buckets'` on a table with `map_serialization_version='basic'`, zero-level parts use bucketed serialization which reorders keys by hash bucket, breaking the insertion order that `basic` promises.

**c) Fix matches root cause:** YES — the fix applies `std::min(map_serialization_version, map_serialization_version_for_zero_level_parts)` at both write paths, ensuring zero-level parts never use a more complex serialization than the base setting. `BASIC=0 < WITH_BUCKETS=1` makes min() correct.

**d) Test intent preserved / new test added:** Original test 03993 is untouched. New regression test 04095 creates a table with the conflicting settings combination and verifies insertion order is preserved.

**e) Both directions:** Before fix: `{'d':200,'c':100}`. After fix: `{'c':100,'d':200}`. 50/50 passes for both 03993 and 04095 with full randomization.

**f) Fix is general:** Both call sites in MergeTreeDataWriter (regular + projection write paths) are patched. The enum ordering (BASIC=0, WITH_BUCKETS=1) makes `std::min` the correct general solution for any future serialization versions that maintain this ordering.

groeneai:
cc @Avogar @CurtizJ — could you review this? It fixes a bug where `map_serialization_version = 'basic'` wasn't respected for zero-level parts when `map_serialization_version_for_zero_level_parts` was independently set to `'with_buckets'` (by CI randomization), causing Map keys to be reordered by hash bucket. The fix caps zero-level serialization at the base version using `std::min` in both write paths of `MergeTreeDataWriter`. This was the same issue PR #101024 identified but with a different approach — see @alexey-milovidov's comment that `map_serialization_version = 'basic'` should already prevent bucketed maps in all level parts.

Avogar:
`map_serialization_version_for_zero_level_parts` is supposed to override `map_serialization_version` for zero level parts regardless of the `map_serialization_version` value. If any test fails because of this, let's fix the test by fixing the values of settings.

groeneai:
Thanks for the review, @Avogar — understood. The override behavior of `map_serialization_version_for_zero_level_parts` is intentional, and the fix belongs in the test (pinning the setting to `'basic'` in the CREATE TABLE). PR #102350 by @alexey-milovidov already does exactly this, so deferring to that.

PR #102338, opened 2026-04-10
Fix AST formatting roundtrip for `FROM` and `VALUES` used as function names

alexey-milovidov:
The BuzzHouse fuzzer found that when `FROM` or `VALUES` are used as function names, the AST formatter outputs them without backtick-quoting. When parsing the formatted query back, the parser interprets these as SQL clause keywords instead of function names, causing a SYNTAX_ERROR.

Add `FROM` and `VALUES` to the list of keywords that must be backtick-quoted in `writeProbablyQuotedStringImpl`, alongside the existing `DISTINCT`, `ALL`, `TABLE`, and `SELECT` entries.

https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=101054&sha=1a927fce3bc0eb606438ccc418daafbd3060ad2b&name_0=PR&name_1=BuzzHouse%20%28amd_debug%29

### Version info
- Merged into: `26.5.1.78`


alexey-milovidov (commented):
Looks good.

src/IO/WriteHelpers.cpp
```diff
@@ -97,19 +97,27 @@ void writeException(const Exception & e, WriteBuffer & buf, bool with_stack_trac
...
+        && !isCaseInsensitiveEqual(s, "table")
         /// SELECT unquoted as an identifier would be re-parsed as the SELECT keyword and produce a
         /// different AST, e.g. arrayElement(Identifier("SELECT"), x) formats as SELECT[x], which
         /// re-parses as a subquery (SELECT [x]) with a different structure.
-        && !(s.size() == select_str.size() && 0 == strncasecmp(s.data(), "select", s.size())))
+        && !isCaseInsensitiveEqual(s, "select")
+        /// These keywords cause parsing ambiguity when used as function or identifier names
+        /// because the parser consumes them as clause-starting keywords.
+        && !isCaseInsensitiveEqual(s, "from")
+        && !isCaseInsensitiveEqual(s, "values"))
```
george-larionov, line 120: This looks much nicer than the old implementation

george-larionov (approved):
LGTM

SaltTan:
I have a table with a Tuple(values String).
After an upgrade new replica could not download older parts because of different columns hash.
```
Code: 40. DB::Exception: Part all_900_1859_113 from replica-3 has different columns hash (it may rarely happen on race condition with KILL MUTATION). (CHECKSUM_DOESNT_MATCH)
```
I was lucky that the table was small and I could afford OPTIMIZE FINAL :)

PR #101292, opened 2026-03-31
Fix UBSan in MergeTreeDataPartWriterCompact::cancel

alexey-milovidov:
In `MergeTreeDataPartWriterCompact::addStreams`, `streams_by_codec[codec_id]` inserts a null `shared_ptr` via `operator[]` before `std::make_shared` runs. If the allocation throws, the null entry remains in the map. When `cancel` is later called (e.g. from the background executor's error path), it iterates `streams_by_codec` and dereferences the null entry, causing UBSan to report undefined behavior.

The fix rewrites the insertion to be exception-safe: use `find` to check for an existing entry, and only `emplace` the fully-constructed `shared_ptr` returned by `std::make_shared`. If the allocation throws, the map is not modified, so no null entries can be observed by `cancel`.

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix undefined behavior in `MergeTreeDataPartWriterCompact::cancel` when a stream allocation fails.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.164`
- Backported to: `26.4.2.6`, `26.3.13.10`

PR #102351, opened 2026-04-10
Fix join reorder applying type changes at wrong step

alexey-milovidov:
This PR is a follow-up to #102516, which already fixed the "Cannot fold actions for projection" exception in join reorder optimization with type-changing joins.

#102516 added the regression test as `04097_join_reorder_type_changes.sh`. This PR replaces it with the equivalent `.sql` form (`04095_join_reorder_type_changes.sql`), which is preferable per project convention — it does not require bash and runs through the standard query path. Same schema, data, query, settings, and expected output; only the file format changes.

Original stress test (arm_msan) report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=101292&sha=2d129c6df44b0ed28a05397e55abe0cbbff9bb1e&name_0=PR&name_1=Stress%20test%20%28arm_msan%29

Upstream PR: https://github.com/ClickHouse/ClickHouse/pull/101292
Original fix PR: https://github.com/ClickHouse/ClickHouse/pull/102516

- Not for changelog

### Version info
- Merged into: `26.5.1.496`


alexey-milovidov (commented):
Good, test-only change.

PR #102354, opened 2026-04-10
ci: switch all tests to azurite-rs

alexey-milovidov:
Sometimes real azure can give 503 errors [1], plus it is not possible to run tests locally.

[1]: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=98776&sha=e98e65358583dbe10ac317de0aa9aca927d58213&name_0=PR&name_1=Stateless%20tests%20%28arm_asan%2C%20azure%2C%20parallel%29

Also, we had an incident with Azure yesterday.

### Version info
- Merged into: `26.5.1.170`


alesapin (changes requested):
What's the point of replacing tests with real system which runs in production with some random wibe-coded binary which nobody every read? What is the goal of tests and CI than? It's better to remove tests with Azure than waste compute in this way.

alexey-milovidov:
> some random wibe-coded binary

wibe -> vibe

> What is the goal of tests and CI than?

than -> then

Azurite-rs is a mock-implementation of Azure, similar to Azurite. It implements just enough of the API to pass all the tests for Azurite and all the tests for ClickHouse, so it is suitable for isolated, reliable testing. Vibe-coding allows creating these implementations quickly, with high quality and conformance. The only thing you can't expect is the completeness of the implementation, so if you start using a new part of the API, it might not be supported.

As with any mocks, it has advantages and disadvantages. 
Advantages: - It will make our tests stable, and we will not have to worry whether Microsoft Azure has downtime or not.
Disadvantages: - It's not real.

We can still run tests with the real Microsoft Azure, but it should be outside of the test contour with a tight feedback loop. The tests with real Microsoft Azure can run periodically in the background, or only in the master. We shouldn't allow them to run in every pull request, because Microsoft Azure is unreliable.

alexey-milovidov:
These tests **failed** yesterday, thanks to Microsoft Azure :skull:.

Screenshot_20260410_120436

alesapin:
@alexey-milovidov let's have them in master and private repo? BTW, it was the way how these tests were intended to work, I'm surprised that we run them in each PR (it should be quite expensive). Likely it was enabled by accident. So proposal is to remove this tests run from PR workflow.

alexey-milovidov:
Thanks, this will be good!

alexey-milovidov:
> [Address review: stop running Azure tests in PR workflow](https://github.com/ClickHouse/ClickHouse/pull/102354/commits/65fe336b5afcf3ab946e7cd0b3c6dc83b7f1d36a)

Please, disregard what @alesapin said - it was an exaggeration. Do not remove Azure tests from PRs. Restore them.

alexey-milovidov (commented):
Ok.

PR #102355, opened 2026-04-10
Revert "Revert "Do not invalidate dictionary if there is only one of them""

Avogar:
Reverts ClickHouse/ClickHouse#102272

Let's wait for some time to see if reverting helped with failing `02156_async_insert_query_log` test.


bobrik:
I think it's safe to close this one now that the bug is confirmed: https://github.com/ClickHouse/ClickHouse/pull/103662#issuecomment-4333029838.

PR #100773, opened 2026-03-26
Support hot-reloading keeper server settings

al13n321:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
A few Keeper server settings are now hot-reloaded if config file is changed at runtime: max_requests_batch_size, max_requests_batch_bytes_size, max_request_size, quorum_reads.

---

I got tired of restarting the server all the time when benchmarking, so here's a way to tweak things on the fly.

To make a setting hot-reloadable, add `HOT_RELOAD` flag to its declaration in `CoordinationSettings.cpp`, and use `keeper_context->getCoordinationSettings()` to get the latest value.

This turned out much easier than expected. I always thought this would have to be slow or complicated (rcu! hazard pointers! thread-cached refcount!) and require changing all the call sites. But no, it's fine, you can just cache a shared_ptr in a thread-local variable to avoid all cacheline contention.

This PR only makes a couple easiest settings reloadable, just to make sure it works: `max_requests_batch_size`, `max_requests_batch_bytes_size`, `quorum_reads`.

Maybe we should do this to ServerSettings too, maybe replacing some of the laborious manual change propagation in the config reload callback.

### Version info
- Merged into: `26.4.1.826`

PR #102372, opened 2026-04-10
Fix flaky timeouts in union race condition tests

antonio2368:
Tests `00090`, `00091`, `00093`, `00094` use fixed-iteration loops (10 x 100 queries) that cannot self-terminate before the test runner's 180-second `TOO_LONG` threshold. Under flaky check, `ThreadFuzzer` enables pthread mutex wrapping (50% CPU-migration probability per mutex op), inflating per-query latency 4-6x. Combined with the 14 new randomized query-plan settings from #101638, these tests exceed the timeout on ASan/Debug/Binary builds (100% failure rate on ASan).

Sibling test `00092` had the identical problem and was already fixed in #96976 by switching to a `TIMELIMIT=$((SECONDS + 100))` wall-clock loop. This PR applies the same pattern to the remaining four tests.

Race detection is preserved: incorrect query results still cause `grep` to match, print `Fail!`, and break the loop, producing a reference file mismatch.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100773&sha=43fb80eb293d22cd90d58702f50a018bd651934f&name_0=PR&name_1=Stateless%20tests%20%28amd_debug%2C%20flaky%20check%29

Related: #100773, #95354

...

### Version info
- Merged into: `26.4.1.809`


tests/queries/0_stateless/00090_union_race_conditions_1.sh
```diff
@@ -16,6 +16,7 @@ echo "
     INSERT INTO two_blocks VALUES ('2000-01-02');
 " | $CLICKHOUSE_CLIENT -n
 
-for _ in {1..10}; do seq 1 100 | sed 's/.*/SELECT count() FROM (SELECT * FROM two_blocks);/' | $CLICKHOUSE_CLIENT -n | grep -vE '^2$' && echo 'Fail!' && break; echo -n '.'; done; echo
+TIMELIMIT=$((SECONDS + 100))
```
kssenii, line 19: nit: `TIMELIMIT` is a bit difficult to read, may be `TIME_LIMIT` is better

kssenii (approved):

tests/queries/0_stateless/00090_union_race_conditions_1.sh
antonio2368: It's naming used in other tests with the exact same mechanic so I would prefer keeping it for now.

PR #102373, opened 2026-04-10
S3-backed git submodule cache for CI

maxknv:
Cache submodule git objects in S3 to eliminate 129 individual GitHub clones per build job.

**How it works:**
- Config Workflow computes a content-addressed hash of all submodule SHAs
- On cache miss: clones submodules, tars `.git/modules/`, uploads to S3
- Jobs with `needs_submodules=True` download the archive and extract — `git submodule update` is instant (objects are local)
- On cache hit (most PRs): no submodule cloning at all, just download from S3
- Graceful fallback: if cache is unavailable, jobs clone from GitHub as before

**Changes:**
- `Settings.ENABLE_SUBMODULE_CACHE` — project-level opt-in setting
- `Job.Config.needs_submodules` — marks jobs that need submodule cache restore
- `RunConfig.submodule_cache_hash` — passes hash from Config Workflow to downstream jobs
- `Digest.get_submodule_shas` — shared helper, reused by job cache digest
- `build_clickhouse.py` / `fast_test.py` — use cached objects when available, skip GitHub fetch

### Version info
- Merged into: `26.4.1.831`


maxknv:
Looks good! Faster and much more reliable

ci/praktika/runner.py
```diff
@@ -255,8 +256,40 @@ def _pre_run(self, workflow, job, local_run=False):
                     if artifact.compress_zst:
                         Utils.decompress_file(Path(Settings.INPUT_DIR) / artifact_path)
 
+        if not local_run and job.needs_submodules and Settings.ENABLE_SUBMODULE_CACHE:
+            self._restore_submodule_cache()
+
         return 0
 
+    @staticmethod
+    def _restore_submodule_cache():
```
leshikus, line 265: I think, it would be nice to have at some point the object which will implement this API https://github.com/actions/cache - then we can implement other caches using this functionality as well

ci/settings/settings.py
```diff
@@ -19,6 +19,7 @@ class RunnerLabels:
 DOCKER_BUILD_AMD_RUNS_ON = RunnerLabels.STYLE_CHECK_AMD
 
 CACHE_S3_PATH = f"{S3_BUCKET_NAME}/ci_ch_cache"
+ENABLE_SUBMODULE_CACHE = True
```
leshikus, line 22: At some point maybe it worth to migrate to standard json for settings, this is easier to understand than several `settings` files

ci/praktika/native_jobs.py
```diff
@@ -221,6 +222,76 @@ def _clean_buildx_volumes():
...
+
+        # Check if cache already exists in S3
+        if S3.head_object(s3_path):
+            print(f"Submodule cache hit: {s3_path}")
+            info = f"cache hit: {cache_hash}"
+        else:
+            print(f"Submodule cache miss, creating: {s3_path}")
+            Shell.check("git submodule sync", verbose=True, strict=True)
+            Shell.check("git submodule init", verbose=True, strict=True)
+            if Path("contrib/update-submodules.sh").exists():
```
leshikus: I think it's better to have one variant here instead of two - this would be simpler.

leshikus (approved):
Looks useful and clean

ci/settings/settings.py
maxknv: I also want to do something with settings. I just cannot find a better way. The challenge is that some settings are default (defined in praktika), some must be set by the user, and some could be set by the user.

PR #102162, opened 2026-04-09
Add missing `hostname` column to `DeltaMetadataLog` and `IcebergMetadataLog`

Crazybus:
`system.delta_lake_metadata_log` and `system.iceberg_metadata_log` are the only two `SystemLog`-derived tables that don't have a `hostname` column. Every other system log table (25+) includes it.

The `SystemLog` base class does not inject `hostname` automatically -- each `LogElement` subclass must declare it in `getColumnsDescription` and write it in `appendToBlock`. Both tables were introduced in 25.10 (#86152 and #87263) by copying from each other but neither included it.

This is the same class of bug that was previously fixed for `blob_storage_log` in #62456. The original bulk addition of `hostname` to all system log tables was #55894.

Any tooling that queries `SELECT hostname FROM system.<table>` across all system log tables will get `UNKNOWN_IDENTIFIER` errors on these two tables. The tables are always created by the `SystemLog` infrastructure regardless of whether the `delta_lake_log_metadata` / `iceberg_log_metadata` settings are enabled.

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Documentation entry for user-facing changes

- [x] Documentation is not required (no user-facing changes beyond the bug fix)

### Version info
- Merged into: `26.4.1.700`

PR #102374, opened 2026-04-10
Fix flaky test 04092_delta_iceberg_metadata_log_hostname

groeneai:

- none

### What this PR does

The test `04092_delta_iceberg_metadata_log_hostname` (introduced by #102162) intermittently fails with `UNKNOWN_TABLE` for `system.delta_lake_metadata_log` on master (~5% failure rate, 6 master failures on April 10, 30+ across unrelated PRs).

**Root cause:** System log tables like `delta_lake_metadata_log` and `iceberg_metadata_log` are lazily created — they are only prepared (i.e., the MergeTree table is created in the `system` database) when:
1. The first log entry is flushed, OR
2. `SYSTEM FLUSH LOGS` is explicitly called

The `prepare_system_log_tables_on_startup` server setting defaults to `false`, so these tables do not exist on a fresh server until one of those conditions is met. Unlike `query_log` (which gets entries from every query), `delta_lake_metadata_log` only receives entries during Delta Lake operations — which don't occur in most test runs.

**Fix:** Add `SYSTEM FLUSH LOGS;` at the start of the test to force table preparation, matching the pattern used by similar tests (e.g., `02905_system_logs_hostname.sql` which flushes before querying system log tables).

**Verification:**
- Fresh server WITHOUT flush: `SELECT hostname FROM system.delta_lake_metadata_log LIMIT 0` → Code 60 UNKNOWN_TABLE (deterministic)
- Fresh server WITH flush: passes every time (20/20 runs)

### Version info
- Merged into: `26.4.1.844`


groeneai:
### Pre-PR Validation Gate

**Session:** `cron:clickhouse-ci-task-worker:20260410-111500`

**a) Deterministic repro?**
Yes. On a fresh ClickHouse server (no prior `SYSTEM FLUSH LOGS`), `SELECT hostname FROM system.delta_lake_metadata_log LIMIT 0` deterministically fails with `Code: 60. UNKNOWN_TABLE`. The table does not exist because `createSystemLog()` only calls `prepareTable()` on first flush (line 641 of `SystemLog.cpp`) or explicit `SYSTEM FLUSH LOGS` (via `create_table_force` flag).

**b) Root cause explained?**
System log tables are lazy-prepared. `prepare_system_log_tables_on_startup` defaults to `false`, so tables are created only when data is first flushed. Unlike `query_log` (which receives entries from every query), `delta_lake_metadata_log` only gets entries during Delta Lake operations. On ~5% of CI runs, this test executes before any flush has occurred on the server, so the table doesn't exist.

Yes. `SYSTEM FLUSH LOGS` triggers `notifyFlush` with `should_prepare_tables_anyway=true`, which sets `create_table_force` in the queue `pop()` result, causing `savingThreadFunction()` to call `prepareTable()` even with no pending data. This creates the table.

**d) Test intent preserved?**
Yes. The test still verifies the hostname column exists with `LowCardinality(String)` type. `SYSTEM FLUSH LOGS` only ensures the table is created; it doesn't alter the schema.

Yes. Without fix: deterministic UNKNOWN_TABLE on fresh server. With fix: 20/20 passes on fresh server.

groeneai:
cc @alesapin — could you review this? It fixes a ~5% flaky failure in 04092_delta_iceberg_metadata_log_hostname by adding `SYSTEM FLUSH LOGS` to force lazy system log table creation before querying `system.delta_lake_metadata_log`.

alesapin (approved):

alesapin:
@groeneai Let's specify table we want to flush. generic `SYSTEM FLUSH LOGS` is not allowed.

groeneai:
Done — changed to table-specific flush: `SYSTEM FLUSH LOGS delta_lake_metadata_log, iceberg_metadata_log`. Also merged latest master.

alexey-milovidov (approved):

PR #102381, opened 2026-04-10
Fix icebergTruncate: add support for Decimal128 and Decimal256

maxistua:
`icebergTruncate` rejected `Decimal128` and `Decimal256` with `ILLEGAL_TYPE_OF_ARGUMENT`, even though its error message already advertised `Decimal`. This broke plain `SELECT`s over Iceberg tables partitioned by `truncate(N, col)` on columns such as `Decimal(38, 0)`, since partition pruning builds the same transform. The type check now accepts every decimal, and `executeImpl` gained the two missing `checkDecimal` branches.

Fixed `ILLEGAL_TYPE_OF_ARGUMENT` in `icebergTruncate` when the value argument is `Decimal128` or `Decimal256`. Only `Decimal32` and `Decimal64` were accepted before, which broke queries over Iceberg tables partitioned by `truncate(N, col)` on high-precision decimal columns such as `Decimal(38, 0)`.


tests/queries/0_stateless/03376_iceberg_truncate.sql
```diff
@@ -54,3 +54,11 @@ SELECT icebergTruncate(10, 1::Int64);
 SELECT icebergTruncate(10, -1::Int64);
 SELECT icebergTruncate(50, toDecimal64('10.65', 2));
 SELECT icebergTruncate(3, 'iceberg');
+
+SELECT icebergTruncate(100000, toDecimal32(123456, 0));
+SELECT icebergTruncate(100000, toDecimal64(1234567890123, 0));
+SELECT icebergTruncate(100000, toDecimal128(12345678901234567890, 0));
+SELECT icebergTruncate(100000, toDecimal256(123456789012345678901234567890, 0));
```
evillique, line 61: Exceeds UInt64, so this parses as Float64 and loses precision. Let's use a string literal.

```suggestion
SELECT icebergTruncate(100000, toDecimal256('123456789012345678901234567890', 0));
```

tests/queries/0_stateless/03376_iceberg_truncate.sql
```diff
@@ -54,3 +54,11 @@ SELECT icebergTruncate(10, 1::Int64);
...
+
+SELECT icebergTruncate(100000, toDecimal128(100000, 0));
+SELECT icebergTruncate(100000, toDecimal128(99999, 0));
```
evillique, line 64: All new cases use scale 0 and non-negative values, so scale handling and negative rounding stay untested.

```suggestion
SELECT icebergTruncate(100000, toDecimal128(99999, 0));

SELECT icebergTruncate(50, toDecimal128('10.65', 2));
SELECT icebergTruncate(50, toDecimal256('10.65', 2));
SELECT icebergTruncate(10, toDecimal128('-0.05', 2));
SELECT icebergTruncate(10, toDecimal256('-0.05', 2));

```

PR #95771, opened 2026-02-02
Optimize ReplicatedMergeTree queue locks for readonly operations on tables with mutations

korowa:
### Changelog category (leave one):
- Performance Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Reduce lock contention during readonly operations on ReplicatedMergeTree tables with finished mutations.

### Details

`ReplicatedMergeTree::getMutationsSnapshot` function is called on every select query from replicated table, and due to `apply_patch_parts` it iterates through all known partitions mutations on every call.

In cases when mutations data is large (e.g. table has many partitions and couple of mutations, where each mutation contains stale partition ids, produced by the values from `/block_numbers` ZooKeeper node), and the number of concurrent queries is high enough, the lock contention in `getMutationsSnapshot` may become a bottleneck and cause queries to stuck during planning phase.

The suggestion is to allow shared locks on ReplicatedMergeTree queue state during readonly operations.

<details>
<summary>Benchmarks</summary>

The results are obtained on debug build (non-Linux, so `absl::Mutex` was used as a `SharedMutex` implementation), with the table containing 10 parts + 10k "stale" parts in mutations:

```sql
SELECT count()
FROM system.parts
WHERE `table` = 'bloat'

Query id: 0ebbb16b-62a6-47fe-ac98-fd22bb77cdba

   ┌─count()─┐
1. │      10 │
   └─────────┘

1 row in set. Elapsed: 0.002 sec.

SELECT
    count(),
    sum(length(block_numbers.partition_id))
FROM system.mutations
WHERE `table` = 'bloat'

Query id: 9caddd12-ba8f-48ba-ac6b-2a2eea7f658b

   ┌─count()─┬─sum(length(b⋯tition_id))─┐
1. │      12 │                    74750 │
   └─────────┴──────────────────────────┘

1 row in set. Elapsed: 0.030 sec.
```

Running 2-4-8 parallel SELECT queries for a single non-existing key selection:
```sh
# master -- lock_guard on std::mutex
eduard ClickHouse $ ./build/debug/programs/clickhouse-benchmark -q "select * from bloat where key = 7123 and prt_num = 11015 and prt_str = '11015' settings apply_patch_parts = 1, max_threads = 1" -t 30 -d 0 -c 2
Loaded 1 queries.
Stopping launch of queries. Requested time limit 30 seconds is exhausted.

[...]

PR #102382, opened 2026-04-10
Avoid lock contention from background fetch threads in getMutationsSnapshot

c-end:
We observed a significant increase in SELECT duration (>10x) when there are a lot of background fetches running. The flamegraph indicated lock contention in `ReplicatedMergeTreeQueue::getMutationsSnapshot`. This happened for tables that had no pending data or metadata mutations. It looks like the method could just exit early without acquiring `state_mutex` in this case (with `apply_mutations_on_fly` and `apply_patch_parts` both disabled). We should even be able to exit early if there are parts with an older metadata version by tracking if there are any supported metadata mutations with an atomic counter. The shared lock introduced in https://github.com/ClickHouse/ClickHouse/pull/95771 doesn't seem to avoid this because fetches frequently acquire exclusive locks.

- Performance Improvement

Improve SELECT performance by avoiding lock contention when getting mutation snapshots with many concurrent background fetches.

### DIFF
--- a/src/Storages/MergeTree/ReplicatedMergeTreeQueue.cpp
+++ b/src/Storages/MergeTree/ReplicatedMergeTreeQueue.cpp
@@ -48,6 +48,14 @@ namespace ErrorCodes
     extern const int BAD_ARGUMENTS;
 }
 
+static bool hasSupportedMetadataMutation(const MutationCommands & commands)
+{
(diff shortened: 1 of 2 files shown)


c-end:
@alexey-milovidov could you or someone else take a look at this?

PR #102383, opened 2026-04-10
Detect decompressive transcoding during reading from GCS

### LINKED ISSUE #47980: empty result or ZLIB_INFLATE_FAILED error when using S3 table function and GCS files with Content-Encoding: gzip after upgrading from 22.12 to 23.x
### Details
Issue is Google S3 specific  when using objects with specific metadata => those aren't retrieved anymore in new version

### GCS Problematic Area
There are 3 ways how you can store file in GCS
- can compress file and store as archive, e.g. file.json.gz => this OK
- can leave file as is w/o compression, e.g. file.json => this OK
- can compress file only during transit, however leave original content-type/file name => this not working anymore
  S3 details:
    when using this mode in S3 file stored as file.json in compressed way
    if client calling API with accept encoding gzip => file sent as compressed, client auto-decompresses it
    if client calling API w/o accept encoding header => file sent as decompressed
    E.g. if web browser retrieves such file, even though payload is send compressed, it shown in client as decompressed as plain .json file
    This controlled by setting s3 object metadata Content-Encoding: gzip 
    (see https://cloud.google.com/storage/docs/transcoding)
  

### CH Behavior change
Prior it was possible to retrieve such objects (with metadata Content-Encoding = gzip ) from Clichouse when using compression = none

- e.g. version: 22.12.5.34
    ```
    SELECT count()
    FROM s3('https://storage.googleapis.com/<bucket>/<file_with_content_encoding_set>.json', 'JSONEachRow');  --OK

    ```

after upgrade to 23.2 or 23.1 
  this doesn't work anymore with S3 engine, however still works with URL e.g. if using public bucket.
- e.g. version: 23.2.4.12
```        
    SELECT count()
    FROM s3('https://storage.googleapis.com/<bucket>/<file_with_content_encoding_set>.json', 'JSONEachRow', 'auto', 'none');  --NOK: empty result without exception / changed in new version

    SELECT count()
    FROM s3('https://storage.googleapis.com/<bucket>/<file_with_content_encoding_set>.json', 'JSONEachRow', 'auto', 'gzip');  --NOK: inflateReset failed / but this is same as prior

    SELECT count()
    FROM url('https://storage.googleapis.com/<bucket>/<file_with_header>.json', 'JSONEachRow', 'auto', 'none');  --OK: still works
```

### Observations
Based on logs, guess issue is due to size is calculated as 0
```
 <Trace> StorageS3Source: Downloading object of size 0 from S3 in single thread
```
As I understand size is calculated based on S3 head_object call (change in 23.1 to get it via GetObjectAttributes, doesn't apply, as CGS doesn't support this method)
i.e.
````
LOG_TRACE(log, "Downloading object of size {} from S3 in single thread", object_size);  called in createS3ReadBuffer.
createS3ReadBuffer receives size from StorageS3Source::createReader()
createReader get size from S3::getObjectSize < getObjectInfo() < tryGetObjectInfo() < Aws::S3:headObject
````
And issue is due to Google doesn't return ContentLength property.
I.e. Google responds with HTTP  200 to head call, but w/o ContentLength property
only GCS specific headers like x-goog-stored-content-encoding, x-goog-stored-content-length

### Summary
I understand that this is not really ClickHouse issue, but GCS, and that  case is very specific.
But still, if this worked prior, maybe possible to fix, so that such objects with gzip content encoding would be still parsed either with none or gzip compression by S3 table function.

### PR DESCRIPTION
**Pull request by yariks5s**
- Improvement

Now it is possible to detect .gz files (and read them properly by default) and read from GCS, which are automatically decompressed due to decompression transcoding. Closes https://github.com/ClickHouse/ClickHouse/issues/47980

### CODE REVIEW

File: src/IO/S3/getObjectInfo.cpp [hunk 1]
```diff
@@ -88,6 +88,11 @@ namespace
         if (with_metadata)
             object_info.metadata = result.GetMetadata();
 
+        /// Detect confirmed decompressive transcoding (set by PocoHTTPClient)
+        auto it = object_info.metadata.find("ch-decompressive-transcoding");
```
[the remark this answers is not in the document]
yariks5s: Now even if a user has the metadata key, the override won't trigger unless Content-Length is also missing, that combination with that header is essentially impossible (except only if used intended to tell about decompressive transcoding by himself, which is also OK in this case)

src/IO/S3/PocoHTTPClient.cpp
```diff
@@ -709,6 +709,19 @@ void PocoHTTPClient::makeRequestInternalImpl(
...
 
+            /// Detect GCS decompressive transcoding: the object is gzip on disk
+            /// but GCS decoded it before serving. All three must hold:
+            ///   x-goog-stored-content-encoding: gzip  (stored as gzip)
+            ///   no Content-Encoding                   (server removed it after decoding)
+            ///   no Content-Length                      (server doesn't know decompressed size)
+            if (poco_response.has("x-goog-stored-content-encoding")
+                && poco_response.get("x-goog-stored-content-encoding") == "gzip"
+                && !poco_response.has("content-encoding")
+                && !poco_response.has("content-length"))
```
Michicosun: should we check here `x-goog-stored-content-encoding != identity and !poco_response.has("content-encoding")`

src/Storages/ObjectStorage/ReadBufferIterator.cpp
```diff
@@ -143,7 +143,14 @@ std::unique_ptr<ReadBuffer> ReadBufferIterator::recreateLastReadBuffer()
...
 
-    const auto compression_method = chooseCompressionMethod(current_object_info->getFileName(), configuration->compression_method);
+    auto compression_method = chooseCompressionMethod(current_object_info->getFileName(), configuration->compression_method);
+
+    if (auto metadata = current_object_info->getObjectMetadata();
+        metadata && metadata->is_server_side_decompressed && !metadata->is_size_known)
+    {
+        compression_method = CompressionMethod::None;
+    }
+
```
Michicosun: this probably can be reversed

let's introduce `chooseCompressionMethod` with this combination

src/Storages/ObjectStorage/ReadBufferIterator.cpp
```diff
@@ -143,7 +143,14 @@ std::unique_ptr<ReadBuffer> ReadBufferIterator::recreateLastReadBuffer()
     auto impl
         = createReadBuffer(current_object_info->relative_path_with_metadata, object_storage, context, getLogger("ReadBufferIterator"));
 
-    const auto compression_method = chooseCompressionMethod(current_object_info->getFileName(), configuration->compression_method);
+    auto compression_method = chooseCompressionMethod(current_object_info->getFileName(), configuration->compression_method);
+
+    if (auto metadata = current_object_info->getObjectMetadata();
+        metadata && metadata->is_server_side_decompressed && !metadata->is_size_known)
```
Michicosun: why do we need to check `is_size_known`?

PR #99200, opened 2026-03-10
Implement automatic buckets in Map data type MergeTree serialization

Avogar:
## Linked issues
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1745

### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Added bucketed serialization for Map columns in MergeTree (`map_serialization_version = 'with_buckets'`). Keys are split into hash-based buckets so that reading a single key (`m['key']`) only reads one bucket instead of the entire column, providing 2-49x speedup for single-key lookups depending on map size. The number of buckets and the bucketing strategy can be controlled by new MergeTree settings: `map_serialization_version`, `max_buckets_in_map`, `map_buckets_strategy`, `map_buckets_coefficient`, and `map_buckets_min_avg_size`. The setting `map_serialization_version_for_zero_level_parts` allows keeping basic serialization for inserts to avoid write overhead while merged parts use buckets. Skip indexes on `mapKeys/mapValues` are supported with bucketed serialization. The `optimize_functions_to_subcolumns` pass rewrites `m['key']` to a per-key subcolumn read when buckets are available.

## Bucketed Map Serialization in MergeTree

By default, a `Map` column in MergeTree is stored as a single `Array(Tuple(K, V))` stream.
Reading a single key with `m['key']` requires scanning the entire column — every key-value pair for every row — even if only one key is needed.
For maps with many distinct keys this becomes a bottleneck.

Bucketed serialization (`with_buckets`) splits the key-value pairs into multiple independent substreams (buckets) by hashing the key.
When a query accesses `m['key']`, only the bucket that contains that key is read from disk, skipping all other buckets.

### Enabling Bucketed Serialization

```sql
CREATE TABLE tab (id UInt64, m Map(String, UInt64))
ENGINE = MergeTree ORDER BY id
SETTINGS
    map_serialization_version = 'with_buckets',
    max_buckets_in_map = 32,
    map_buckets_strategy = 'sqrt';
```

To avoid slowing down inserts, you can keep `basic` serialization for zero-level parts (created during `INSERT`) and only use `with_buckets` for merged parts:

```sql
CREATE TABLE tab (id UInt64, m Map(String, UInt64))
ENGINE = MergeTree ORDER BY id
SETTINGS
    map_serialization_version = 'with_buckets',
[...]

PR #102387, opened 2026-04-10
Fix SIGSEGV in ColumnMap::calculateStatisticsForRange on empty columns

dmcwhorter:
`ColumnMap::calculateStatisticsForRange(0, 0)` accessed `offsets[ssize_t(0) - 1]` (out-of-bounds read at `offsets[-1]`) before the `start == end` guard on the return line. This caused a segfault when `MergedData::insertChunk` processed an empty `ColumnMap` chunk during `MergingSortedAlgorithm` execution.

The bug was introduced by #99200 ("Implement automatic buckets in Map data type MergeTree serialization"), backported to 26.3 via #100900.

**Root cause:**

```cpp
// src/Columns/ColumnMap.cpp — before fix
ColumnMap::StatisticsPtr ColumnMap::calculateStatisticsForRange(size_t start, size_t end) const
{
    const auto & offsets = getNestedColumn().getOffsets();
    size_t total_maps_size = offsets[ssize_t(end) - 1] - offsets[ssize_t(start) - 1]; // BUG: offsets[-1]
    return std::make_shared<Statistics>(start == end ? 0 : ...);  // guard is too late
}
```

When called with `start=0, end=0` (empty column), `offsets[ssize_t(0) - 1]` wraps to `offsets[SIZE_MAX]` — UB / segfault. The `start == end` ternary on the return line is a dead guard since the OOB access already occurred.

**Call chain:**
1. `MergedData::insertChunk` checks `columns[i]->hasStatistics()`
2. `ColumnMap::hasStatistics` unconditionally returns `true`
3. `takeOrCalculateStatisticsFrom` → `getOrCalculateStatistics` → `calculateStatisticsForRange(0, size())`
4. When the source `ColumnMap` is empty (`size() == 0`) → `calculateStatisticsForRange(0, 0)` → crash

**Trigger:** `ORDER BY ... LIMIT N BY` on tables with Map columns, when multi-threaded execution produces chunks where some threads yield zero rows.

**Fix:** Early return for the empty range case before accessing `offsets`, and fix the non-empty path to correctly handle `start == 0` without the `ssize_t` cast.

**Stack trace from production crash (v26.3.5.12):**

```
Signal: Segmentation fault (11). Address: 0x10. Access: read.

 2. DB::ColumnMap::takeOrCalculateStatisticsFrom(...) @ 0x1d15156c
 3. DB::MergedData::insertChunk(DB::Chunk&&, unsigned long) @ 0x1eac418d
 4. DB::MergingSortedAlgorithm::insertChunk(unsigned long) @ 0x1eac67e8
 5. DB::MergingSortedAlgorithm::mergeImpl(...) @ 0x1eb3b4ed
 6. DB::IMergingTransform<DB::MergingSortedAlgorithm>::work() @ 0x1ca7de4c
 7. DB::ExecutionThreadContext::executeTask() @ 0x1e65edbd
 8. DB::PipelineExecutor::executeStepImpl(...) @ 0x1e650364
```

**Related issues:** #101667, #101848

Fix segfault in `ColumnMap::calculateStatisticsForRange` when processing empty Map columns during merge-sorted query execution (e.g. `ORDER BY ... LIMIT N BY` on tables with Map columns).

### DIFF
--- a/src/Columns/ColumnMap.cpp
+++ b/src/Columns/ColumnMap.cpp
@@ -468,9 +468,14 @@ void ColumnMap::takeExactDynamicStructureFrom(const IColumn & source)
 
 ColumnMap::StatisticsPtr ColumnMap::calculateStatisticsForRange(size_t start, size_t end) const
 {
+    if (start == end)
+        return std::make_shared<Statistics>(0, 0);
+
     const auto & offsets = getNestedColumn().getOffsets();
-    size_t total_maps_size = offsets[ssize_t(end) - 1] - offsets[ssize_t(start) - 1];
-    return std::make_shared<Statistics>(start == end ? 0 : static_cast<Float64>(total_maps_size) / static_cast<Float64>(end - start), end - start);
+    size_t nested_start = start == 0 ? 0 : offsets[start - 1];
+    size_t nested_end = offsets[end - 1];
+    size_t total_maps_size = nested_end - nested_start;
+    return std::make_shared<Statistics>(static_cast<Float64>(total_maps_size) / static_cast<Float64>(end - start), end - start);
 }
 
 ColumnMap::StatisticsPtr ColumnMap::getOrCalculateStatistics() const
--- a/src/Columns/tests/gtest_column_map.cpp
+++ b/src/Columns/tests/gtest_column_map.cpp
@@ -1,0 +1,111 @@
+#include <Columns/ColumnMap.h>
+#include <Columns/ColumnArray.h>
+#include <Columns/ColumnTuple.h>
+#include <Columns/ColumnString.h>
+#include <Columns/ColumnsNumber.h>
+
+#include <gtest/gtest.h>
+
+
+using namespace DB;
+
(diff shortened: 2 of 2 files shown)


Avogar (changes requested):
But that's not correct. `offsets` is a `PaddedPodArray` that accepts `ssize_t` index in `operator[]` and supports accessing `-1` element (which is always the default value, in our case it's 0)

dmcwhorter:
> But that's not correct. `offsets` is a `PaddedPodArray` that accepts `ssize_t` index in `operator[]` and supports accessing `-1` element (which is always the default value, in our case it's 0)

@Avogar ok got it, thanks, I misunderstood the issue.  Let me just file an issue instead with the failure/stack trace and you guys can decide what needs to happen to fix it.

dmcwhorter:
@Avogar  here's the issue https://github.com/ClickHouse/ClickHouse/issues/102390

Avogar:
Thanks, I will investigate

PR #102391, opened 2026-04-10
Enable logging in FunctionsStress unit test

Algunenano:
`FunctionsStress.stress` uses `LOG_ERROR` to report which functions failed and why, but the root Poco logger has no channel configured — all messages are silently discarded. The "see log above" assertion message refers to output that was never produced, making CI failures impossible to diagnose.

Set up a `ConsoleChannel` on stderr before creating the logger, following the same pattern used by other unit tests.

Example failure with no useful output: https://s3.amazonaws.com/clickhouse-test-reports/PRs/100230/fd4001476012b7f2827bdaebc87f146ed62f0d4b/unit_tests_tsan/job.log

...

### Version info
- Merged into: `26.4.1.839`

### DIFF
--- a/src/Functions/tests/gtest_functions_stress.cpp
+++ b/src/Functions/tests/gtest_functions_stress.cpp
@@ -1,5 +1,6 @@
 #include <gtest/gtest.h>
 
+#include <Poco/ConsoleChannel.h>
 #include <absl/log/globals.h>
 #include <boost/program_options.hpp>
 #include <fmt/ranges.h>
@@ -2103,6 +2104,13 @@ TEST(FunctionsStress, stress)
 #endif
 
     chassert(!logger);
+
+    /// Set up console logging so that LOG_ERROR output is visible in the test log.
+    /// Without this, the root logger has a nullptr channel and all messages are silently discarded.
+    Poco::AutoPtr<Poco::ConsoleChannel> channel(new Poco::ConsoleChannel(std::cerr));
+    Poco::Logger::root().setChannel(channel);
+    Poco::Logger::root().setLevel("information");
+
     logger = getLogger("stress");
 
     /// (This makes exception stack traces much faster, and this test spends a lot of time throwing and catching exceptions.)


src/Functions/tests/gtest_functions_stress.cpp
```diff
@@ -2103,6 +2104,13 @@ TEST(FunctionsStress, stress)
 #endif
 
     chassert(!logger);
+
+    /// Set up console logging so that LOG_ERROR output is visible in the test log.
+    /// Without this, the root logger has a nullptr channel and all messages are silently discarded.
+    Poco::AutoPtr<Poco::ConsoleChannel> channel(new Poco::ConsoleChannel(std::cerr));
+    Poco::Logger::root().setChannel(channel);
+    Poco::Logger::root().setLevel("information");
```
Michicosun, line 2112: let's use colored channel, wdyt?

```
String test_log_level = "none";
if (const char * test_log_level_env = std::getenv("TEST_LOG_LEVEL")) // NOLINT(concurrency-mt-unsafe)
    test_log_level = test_log_level_env;

Poco::AutoPtr<OwnPatternFormatter> pf(new OwnPatternFormatter(/*color_=*/true));
Poco::AutoPtr<DB::OwnFormattingChannel> channel(new DB::OwnFormattingChannel(pf, new Poco::ConsoleChannel(std::cerr)));
Poco::Logger::root().setChannel(channel);
Poco::Logger::root().setLevel(test_log_level);
```

Michicosun (approved):

src/Functions/tests/gtest_functions_stress.cpp
Algunenano: I don't care. I don't run this locally and CI output does not have color

Algunenano: Other tests don't use color, so I'm going to leave it as is. We can introduce it in all of them if we want

PR #102403, opened 2026-04-10
Fix inserting default values into AggregateFunction columns with LowCardinality arguments

### LINKED ISSUE #116950: Default value fails for AggregateFunction with LowCardinality argument
### Company or project name

_No response_

ClickHouse accepts an `AggregateFunction` column with a `LowCardinality(...)` argument, but inserting a row while omitting that column fails because the aggregate state type is normalized without `LowCardinality`.

25.8.16

### How to reproduce

```sql
CREATE TABLE test
(
    id UUID,
    state AggregateFunction(
        argMaxIf,
        LowCardinality(Nullable(String)),
        DateTime,
        Bool
    )
)
ENGINE = AggregatingMergeTree
ORDER BY id;
```

Then:

```sql
INSERT INTO test (id)
VALUES (generateUUIDv4());
```

fails with:

```text
Code: 43. DB::Exception:
Cannot insert filed with type
AggregateFunction(argMaxIf, LowCardinality(Nullable(String)), DateTime, Bool)
into column with type
AggregateFunction(argMaxIf, Nullable(String), DateTime, Bool).
(ILLEGAL_TYPE_OF_ARGUMENT)
```

The normalization can also be observed directly:

```sql
SELECT toTypeName(
    argMaxIfState(
        CAST(NULL AS LowCardinality(Nullable(String))),
        now(),
        false
    )
);
```

returns:

```text
AggregateFunction(argMaxIf, Nullable(String), DateTime, Bool)
```

### Expected behavior

A type accepted in the table definition should be able to receive its default value.

Either:

* `AggregateFunction(..., LowCardinality(T), ...)` should be rejected if the aggregate function normalizes it to `T`, or
* the declared and normalized aggregate-state types should be treated as compatible when inserting the default state.

_No response_

### Related issues and pull requests

* https://github.com/ClickHouse/ClickHouse/issues/33322
* https://github.com/ClickHouse/ClickHouse/issues/41057
* https://github.com/ClickHouse/ClickHouse/issues/33028

### Additional context

_No response_

eclbg:
An aggregate function is always instantiated with `LowCardinality` stripped from its argument types (`AggregateFunctionFactory::get`), but `DataTypeAggregateFunction` keeps the argument types exactly as declared. So `AggregateFunction(argMax, LowCardinality(String), DateTime)` and `AggregateFunction(argMax, String, DateTime)` name one and the same state, and yet the two spellings were compared as raw strings and rejected each other.

As a result, any table with such a column rejected every insert that omitted it:

```sql
CREATE TABLE t (k UInt8, s AggregateFunction(argMax, LowCardinality(String), DateTime)) ENGINE = MergeTree ORDER BY k;

INSERT INTO t (k) VALUES (1);
-- Code: 43. Cannot insert field with type AggregateFunction(argMax, LowCardinality(String), DateTime)
--          into column with type AggregateFunction(argMax, String, DateTime)
```

`defaultValueOfTypeName` failed on the same type names for the same reason.

The comparison now falls back to `DataTypeAggregateFunction::nameMatchesState`, which checks that both names resolve to the same normalized state type and version. `ColumnAggregateFunction::insert`, `ColumnAggregateFunction::tryInsert` and `convertFieldToType` share it.

Nothing about the stored data changes: the state bytes are identical under either spelling, so states stay interchangeable between the two, and the default state written by this fix is the same empty state the spelling without `LowCardinality` already produced.

Closes: https://github.com/ClickHouse/ClickHouse/issues/116950

Fix `Cannot insert field with type ...` when inserting a row that omits an `AggregateFunction` column whose argument types contain `LowCardinality`. `defaultValueOfTypeName` is fixed for the same type names.

---
Workflow [[PR](https://s3.amazonaws.com/clickhouse-test-reports/praktika.html?PR=102403&sha=latest&name_0=PR)]
Sync PR [[sync-upstream/pr/102403](https://github.com/search?q=head%3Async-upstream%2Fpr%2F102403+org%3AClickHouse+type%3Apr&type=pullrequests)]


eclbg:
Thanks for taking this @Algunenano.

@jordiori FYI, I remember you had to work around this.

Algunenano (approved):

clickgapai:
Hi — this looks like a backport candidate for `26.3`, `26.6`, `26.7`, `26.8`. I ran this PR's own regression test against each branch's binary and the bug **reproduces** there (the fix isn't present yet).

PR #101757, opened 2026-04-04
KeeperDispatcher overhaul

al13n321:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
 * Various changes to Keeper that make it around 2x faster overall (better batching, pipelining messages to leader, pipelining log appends).

---

The first two commits can be reviewed separately. First one refactors things to move the hot paths from KeeperDispatcher into a new class KeeperRequestDispatcher. Second one adds KeeperRequestDispatcher2.

I didn't benchmark it much yet. Preliminarily, it's ~30% faster (on the one workload I tried).

Explanation from the code:
```C++
/// Considerations:
///  * Forwarding. If nuraft leader is not on this node, requests are sent to the leader node.
///  * Batching. responseThread() picks as many consecutive requests as possible to send together,
///    to save the overhead of passing tiny individual requests around.
///  * Pipelining. Multiple request batches can be in flight simultaneously, to keep the whole
///    nuraft pipeline busy, for full resource utilization for maximum throughput.
///    E.g. maybe one batch is being processed by followers, while the next batch is being
///    transferred from leader to followers, while the next batch is being preprocessed by leader.
///  * Ordering. Requests from one client session must be executed in the order they arrived.
///    To preserve order even when multiple requests are executed concurrently, we rely on nuraft
///    leader not reordering requests, on tcp not reordering messages, and on all the plumbing in
///    between also preserving order. See KeeperAppendStream.
///    If leader changes or we lose connection to it, we fail all in-flight requests (their outcome
///    is unknown) and open a new stream.
///  * Executing reads locally. Read requests don't need to go through raft, and don't even have to
///    be forwarded to the leader node. But they must be executed at exactly the right time: after
///    all previous write requests in the same client session were applied, before the next
///    request in the session is applied.
///  * Reordering requests. Requests in the same client session can't be reordered, but we're free
///    to interleave requests from different sessions however we like. We use this freedom to group
///    read requests together as much as possible. When we have many consecutive read requests, we
///    can execute them in parallel (while blocking writes, so that znode state doesn't change).
///  * Minimizing load on the commit thread. Commit thread is usually the bottleneck for throughput
///    of the whole keeper cluster. We should do as little as possible in the commit callback.
///  * Flow control. See below.
///
/// KeeperRequestDispatcher2 is in charge of most of flow control in keeper server.
[...]

PR #99123, opened 2026-03-09
[WIP] Keeper: treat a batch of requests as one nuraft entry

al13n321:
### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

(Written by claude, I didn't review it yet, just want to run CI.)

PR #102419, opened 2026-04-10
Refactor Keeper: per-session request/response lifecycle

filimonov:
- Performance Improvement

Refactored Keeper request processing to use per-session request lifecycle with strict FIFO ordering, sharded request queue, and direct response delivery. Improves throughput by 18-64% and reduces latency by 12-37% depending on workload and fsync configuration.

## Refactor Keeper: per-session request/response lifecycle

This PR refactors the ClickHouse Keeper dispatcher to use a per-session request lifecycle model instead of global queues with ad-hoc synchronization. The result is cleaner architecture, stronger FIFO ordering guarantees, and significantly better throughput and latency.

### Summary

Replace Keeper's global queues and fragmented request types with a unified, session-centric architecture.

**Per-session request routing**: Extract `KeeperSession` and `KeeperSessionRegistry` from `KeeperDispatcher`. All requests are classified and routed through sessions: writes and quorum reads go through per-session FIFO queues with a round-robin scheduler; local reads are deferred behind preceding writes via per-session barriers and executed when the write commits.

**Eliminate global response queue**: Replace the global `ConcurrentBoundedQueue` and its dedicated `responseThread` with a `ResponseRouter` callback that delivers responses directly to sessions. `KeeperSession` owns the response delivery with two modes: queue+wakeup (TCP handlers) and callback (`KeeperOverDispatcher`).

**Unified request type**: Merge `KeeperRequestForSession` and `RequestEnvelope` into a single `SessionRequest` class carrying session-level fields, Raft-level fields, and the response once the request completes.

**Error delivery on session termination**: `KeeperSession::finalizeWithErrors` atomically collects in-flight requests, creates error responses for pending ones (preserving already-committed responses), and delivers them before closing. `terminateSession` uses this so clients receive explicit error codes (`ZSESSIONEXPIRED`, `ZCONNECTIONLOSS`) instead of silent disconnect.

**Per-session active request limit**: Add `max_session_active_requests` coordination setting (default 0 = unlimited). `KeeperSessionRegistry` takes `KeeperContextPtr` in its constructor and extracts settings as plain members.

**FIFO mismatch safety**: `onRaftResponse` signals XID mismatch via `DeliveryResult::FifoMismatch`. `routeResponse` terminates the session instead of falling back to `deliverDirect` which would leave the FIFO head unresolved.

### Motivation

The original Keeper dispatcher managed sessions through several loosely coupled global structures: `live_sessions` set, `session_to_response_callback` map, `read_request_queue` map — each with its own mutex. Read-after-write ordering was maintained by intercepting reads in `requestThread` and deferring them in a global map keyed by `(session_id, write_xid)`. Response delivery went through a global `ResponsesQueue` polled by a dedicated `responseThread`.

This design had several problems:
- **Lock contention**: three global mutexes serialized all session operations under high concurrency.
- **Fragile ordering**: read-after-write barriers depended on the dispatcher correctly correlating reads with writes across a global map — any missed correlation silently violated consistency.
- **Unnecessary hops**: every response traversed TCP handler → global request queue → dispatcher → Raft → global response queue → response thread → TCP handler, adding latency at each synchronization point.
- **No per-request observability**: request state was implicit, making it hard to diagnose where time was spent.

### Design

```
┌──────────────┐     ┌───────────────────┐     ┌───────────────────────┐
│KeeperTCPHndlr│     │  KeeperDispatcher │     │   KeeperStateMachine  │
│              │     │                   │     │                       │
│  cached_     │     │  session_registry─┼──┐  │  ResponseRouter ──────┤
│  session ────┼─────┤► putRequest()     │  │  │  (callback, no queue) │
│              │     │  requestThread()  │  │  │                       │
│  response_   │     │  routeResponse()  │  │  │  processReadRequest() │
│  queue ◄─────┼─────┤  onRaftCommit()   │  │  │  processReadRequests()│
│  (per-conn)  │     │                   │  │  │                       │
└──────────────┘     └───────────────────┘  │  └───────────────────────┘
                                            │
                     ┌──────────────────────┘
                     ▼
          ┌─────────────────────┐
          │ KeeperSessionRegistry│
          │                     │
          │  sessions[] ────────┼──► KeeperSession (per client)
          │  findSession()      │     │
          │  registerSession()  │     ├─ active_requests (FIFO deque)
          │  detachSession()    │     ├─ callback (shared_ptr)
          │                     │     ├─ subqueue (→ KeeperSubqueue)
          └─────────────────────┘     ├─ addRequest()
                                      ├─ onRaftResponse()
                                      ├─ onRaftCommitted() → advanceQueue()
                                      ├─ finalizeWithErrors()
                                      └─ deliverDirect() / onWatchNotification()

          ┌─────────────────────┐
          │ KeeperRequestsQueue │
          │                     │
          │  subqueue[0..N-1] ──┼──► KeeperSubqueue (sharded FIFO)
          │  waitAndPullRaft    │     │
          │   Batch()           │     ├─ push()
          │  system_subqueue    │     └─ pullIntoRaftBatch()
          └─────────────────────┘

          ┌─────────────────────┐
          │   SessionRequest    │
          │                     │
          │  state (atomic) ────┼──► RequestState enum
          │  setState()         │     Initial → Received → PendingRaft → InRaft
          │  tryTransitionState()     → RaftResponseReady → Completed
          │  onCommitCompleted()│     → SendingResponse → Sent
          │  per-state metrics  │     (+ PendingLocal → ExecutingLocal branch)
          └─────────────────────┘
```

#### Key components

**`KeeperSession`** — server-side state for a single client connection. Owns an `active_requests` FIFO that enforces strict request ordering within the session. Two executor threads coordinate via the FIFO without cross-thread synchronization beyond the session mutex:
- **SessionThread** (TCP handler): dispatches fast-path local reads inline when the FIFO is empty.
- **QuorumThread** (NuRaft commit callback): attaches Raft responses, pops completed writes, dispatches deferred reads.

The FIFO makes ordering violations immediately visible — any out-of-order response triggers a `chassert` at the FIFO head check in `onRaftResponse`, even if the client would silently tolerate misordering.

**`KeeperSessionRegistry`** — central registry of active sessions. Manages session lifecycle (register, find, detach, shutdown). Uses `shared_mutex` with swap-and-pop vector for O(1) lookups with minimal contention.

**`SessionRequest`** — unified request type carrying session-level fields (session_id, mode, executor), Raft-level fields (zxid, digest, log_idx), response, and an atomic `RequestState`. `setState` validates transitions with `chassert` and updates per-state `CurrentMetrics` gauges (10 metrics from `Initial` through `Sent`), giving real-time visibility into where requests are in the pipeline. OTel span lifecycle is integrated into state transitions.

**`KeeperRequestsQueue`** — sharded concurrent queue replacing the global `ConcurrentBoundedQueue`. Each session pushes quorum requests to its assigned subqueue. `waitAndPullRaftBatch` does round-robin pull across subqueues with deadline-based batch accumulation — larger batches amortize fsync cost.

**`ResponseRouter`** — replaces the global `ResponsesQueue` + `responseThread`. The state machine calls a callback directly, eliminating one queue hop and one thread.

### Performance results

Benchmarked with `keeper-bench` on a 3-node Keeper cluster (Docker, same host), 100 concurrent clients, pipeline depth 8, 4M iterations. Mixed workload: Get (83%), Multi/Create/Remove (17%).

#### With fsync enabled

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| **Read RPS** | 15,059 | 24,771 | **+64%** |
| **Write RPS** | 3,008 | 4,949 | **+64%** |
| **Read p50** | 39.0 ms | 25.7 ms | **-34%** |
| **Write p50** | 39.8 ms | 26.1 ms | **-34%** |
| **Read p99** | 60.4 ms | 38.4 ms | **-37%** |
| **Write p99** | 60.7 ms | 38.8 ms | **-36%** |

#### Without fsync 

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| **Read RPS** | 194,301 | 229,200 | **+18%** |
| **Write RPS** | 38,817 | 45,792 | **+18%** |
| **Read p50** | 2.88 ms | 2.42 ms | **-16%** |
| **Write p50** | 2.99 ms | 2.46 ms | **-18%** |
| **Read p99** | 5.04 ms | 4.41 ms | **-12%** |
| **Write p99** | 5.08 ms | 4.39 ms | **-14%** |


al13n321:
https://github.com/ClickHouse/ClickHouse/pull/101757 - are we working on the same task concurrently? Let's coordinate?

My plan is to add order-preserving pipelining in enough places in keeper until we 100% a cpu core with either the commit thread (`processRequest` + `processLocalRequests`) or the pre-append thread/mutex (`preprocessRequest`). My PR pipelines KeeperDispatcher (see PR summary), then I plan to extract all pre-append calls into a separate thread similar to bg commit thread, to allow `preprocessRequest` to use a whole cpu core and reduce mutex contention and make it easier to profile. Then investigate what other unnecessary blocking is preventing these threads from being busy 100% of the time; this would probably need nuraft's streaming mode (`max_log_gap_in_stream` setting), but currently it slows everything down to a crawl if enabled, for some reason, will investigate. Then probably also do something like https://github.com/ClickHouse/ClickHouse/pull/99123 . Then look into redesigning the KeeperStorage's in-memory data structures to use less memory and be faster.

If that conflicts with what you're doing, let's discuss.

filimonov:
Keeper performance

several problems were identified. 
1) [fixed] reads blocking writes globally, needed session level barriers instead of global waits - done (can be done simpler, but then there are risks).

2) [nuraft] when fsyncs enabled nuraft leader-follower interactions with streaming_replication = false have long dalays, and global locks on the follower side. 

   When fsyncs are enabled the main bottleneck is leader-follower interaction which is synchronous (even with `async_replication`).
   NuRaft waits for follower fsync — the follower is blocked in `handle_append_entries` (lines 988-1006 of `handle_append_entries.cxx`) while waiting for fsync, and the peer `busy_flag_` prevents the leader from sending new messages to that follower until the response comes back.

   So for example even if the leader collected the quorum and wants to notify the follower about the fact that the transaction is done, the follower will know about that only after finishing its own fsync. Profiling confirmed this: follower consensus wait is 3-6ms higher than leader consensus wait — this is the "commit index propagation delay" from leader's `commit()` → `request_append_entries_for_all()` → peer busy → deferred via `set_pending_commit()` → sent only after follower responds.

   Measured: follower `in_raft` = 26.7ms vs leader `in_raft` = 16.7ms (leader-only vs follower-only benchmarks). The ~10ms gap is the average wait for the leader to include the auto-forwarded request in the next consensus round (~half a round of ~17ms). This was confirmed by simultaneous per-node metric collection showing leader consensus = 14ms, follower consensus = 19.5ms.

   It's an architectural limitation of `streaming_replication = false`. There is a dual lock used for the same on leader (PR eBay/NuRaft#74 introduced `cli_lock_` separate from `lock_`). A POC was built for the follower: passing `guard` to `handle_append_entries` and releasing `lock_` during the durability wait (same pattern as `handle_install_snapshot_req`). The patch built, ran stable (no election flaps or errors), but produced only ~1ms improvement — because the follower's extra latency is NOT from `lock_` contention but from the auto-forwarding round-trip through the leader's consensus pipeline. The `auto_fwd_resp_handler` does NOT use `lock_` at all — it runs on an asio thread and directly completes the `cmd_result`. So the lock release helps with vote/heartbeat responsiveness under load (relevant to eBay/NuRaft#355) but does not reduce the follower's write latency.

   Checked upstream eBay/NuRaft master: no issues or PRs address the follower `lock_`-during-durability-wait. PR eBay/NuRaft#74 only fixed the leader side. Issue eBay/NuRaft#379 documents `parallel_log_appending_` but doesn't mention the locking problem.



3) [nuraft] when fsyncs enabled streaming_replication = true improves the latency on the followers significantly (like 2 times) but leads to significant throughput degdadation on the leader

   It looks like when streaming replication is disabled the 'busy' followers work like a natural limitation / backpressure, so more operations go in one commit, and once you enable streaming replication that backpressure goes away so commits become tiny and instead of few commits you have hundreds (and it impacts the leader throughput).

   Measured results (streaming with `asio_opts.streaming_mode_ = true`, `max_log_gap_in_stream_ = 100`):
   - Follower-only: Get RPS 20,324 → 25,010 (+23%), p50 34ms → 26ms, `in_raft` 26.7ms → 14.6ms, commits/fsync 37 → 60.
   - Leader-only: Get RPS 29,500 → 17,590 (-40%), commits/fsync 46 → 32.5.
   - Mixed: Get RPS 23,367 → 1,239 (-95%), commits/fsync 46 → 3.6.

   Root cause identified — three NuRaft call sites fire `request_append_entries` and with streaming the peer is never busy so each sends an actual message:
   1. `handle_cli_req_prelock:99` → `request_append_entries_for_all()` — after every client request batch
   2. `commit():74` → `request_append_entries(pp)` — on every commit index advancement, directly (not via bg thread)
   3. `notify_log_append_completion:1537` → `request_append_entries_for_all()` — after each fsync completion
   Without streaming, p1-p3 are no-ops because the peer is busy. With streaming, each fires a send carrying whatever few entries accumulated since the last send.

   Proof — follower fsync inflation (mixed benchmark with streaming): leader 1,761 fsyncs vs follower 2,715 fsyncs (+54%) for the same 6,368 commits. Followers receive many small `append_entries`, each triggering a separate `end_of_append_batch` → fdatasync.

   Tuning experiments:
   - `max_log_gap_in_stream_` = 50000 vs 100: gap=100 somewhat better (c/f 32.5 vs 24.4 for leader-only) but still regresses vs no streaming. The gap limits logs in flight, not batch size.
   - `use_bg_thread_for_urgent_commit_ = false`: no change — `commit():74` calls `request_append_entries` directly, bypassing `request_append_entries_for_all` and the bg thread entirely.
   - Skipping `request_append_entries_for_all` from `handle_cli_req_prelock` when streaming enabled (routing through `bg_append_ea_->invoke()` instead): marginal improvement for leader-only (17.6k → 18.5k RPS) but mixed still broken because `commit()` path still fragments.

   That seems to be the most promising next optimization target. I believe it should be possible to implement some alternative backpressure mechanisms there (e.g. per-peer send throttle, routing all sends through the bg thread including `commit()`, or minimum batch size for streaming sends).

Fixing 2 or 3 (or both) should 'unlock' the keeper performance, which just spen a lof of time doing nothing in p.2 or doing too many tiny tasks in p.3. 


Other (less imporant) issues:

4) [improved] too many hops of the data with global synchronization (tcp handler - requests queue - dispatcher - nuraft - responces queue - tcp handler) 
   fixed by removing global responces queue, and by doing sharded requests queue (it can potentially allow to run several request thread per subqueue) 

5) [some experiments done, but was not  able to confirm the benefits] dispatcher only support one in-flight operation, whali we are waiting for the previus
   opeation to finish (async_repliaction=false) or waiting it for precommit (async_repliaction=true)
   
   i've quckly tested the option of having more things in-fligh it does not seems it changed anything (as appends are quite fast and bottleneck is somewhere else).
   Confirmed by measurement: the inflight gauge showed 0 during load — the pipeline is naturally balanced (batch collection time ≈ consensus time ≈ 9ms). Multi-inflight doesn't help because there's nothing to pipeline in the steady state.

   Note: one problem here was identified during testing - while on the leader the order of events is preserved even if you have several in-fligh requess (those which are waiting for precommit).
   on the follower automating forwarding from follower to leader can cause reordering of transactions (if you start them in parallel they may be pushed to leader by different connection and reordered)  
   so even if that optimization would work - that forwarding issue should be either fixed on the nuraft side, either just preventing sending requests from the same session / subqueue in parallel.

6) [lot of experiments done, but was not  able to get something much better] batching vs latencies vs backpressure issues. There are several hops when some batching can happen - first when doing append, after on nuraft side (includes leadr - follower interactions), after that on the fsync side. 
   small batches = fragmentation, a lot of operation to do, but low latency
   small batches = small number of operations, but increases latency

   6a) dispatcher forms batches to be sent to nuraft. It's rather on the low-latency side currently (with async_replication=true it was usually producing very tiny batches, like 1-4 requests in each, and by itself id doesn't seems to be the problem - and if fsyncs are disabled it does not hurt).
       attempt to tune that show tiny improvements there, but it requires more carefull testing, and probably does not make too much sense before p2 or p3 will be fixed.
       Note: Keeper-side batch settings (`max_requests_batch_size`, `max_requests_batch_bytes_size`, `max_batch_time_microseconds`) have no impact on NuRaft-side fragmentation — the Keeper batch is submitted as one `putRequestBatch` call, and the splitting happens inside NuRaft when `request_append_entries` fires from multiple call sites. Similarly `max_requests_append_size` / `max_requests_append_bytes_size` (NuRaft's `max_append_size_`) cap the entries per message at 50k, far above actual batch sizes (~46).

   6b) fsyncs have a logic to do one fsync per many commits. Extra measurments show that actually the same thread wastes some time between fsyncs on doing serialization / writing to files, i hava a branch which intoduces double buffering to take serialization cost and many write syscalls out to a different thread,
       but it does not show improvements. Again without p2 or p3 fixed first it seems it doesn't make too much sense on doing that.
       Additionally, group commit in the flushThread was tested (coalescing multiple sealed batches into one fdatasync). It regressed: `last_durable_idx` jumps by many entries at once → NuRaft commit thread completes many `async_result` futures simultaneously → requestThread submits many tiny batches (1.76 commits/fsync instead of 46). Same pathology as streaming — removing backpressure destroys batch efficiency.

   6c) user-level batching - there is PR for that, which may help with that. But when fsyncs are disbled the pefromance of keeper is good even with small batches. So again - probably not worth before fixing p2 or p3.

7) huge tail latencies when fsyncs are disabled
   
   when you don't do fsyncs - some random file operations like truncate, rename, opening or closing file descriptors, or creating new file - which are normally instant becomes very slow because they need to pay the price of flushing the accumlated writes from page cache to disk.
   in practice when fsyncs are disabled every log rotation, or snapshot creation adds signigicant delays. 

   i have a POC fix for that - moving operations related to log rotation (raname, truncate, create new file, remove old file, close file descriptors etc) to the thread pool out of the scope of the main lock. 

8) overhead intoduced by all the alway-on-debugging and performance tracking tools is quite signifacant. The worst issue [fixed] that the request objects are huge because they include (optional) otel spans even if otel is not enabled. And object size is hugs and object construction / copying is very slow.
   Also we access the monotonic clock dozes of times per every request. Normally it's cheap but with high RPS it becomes visible.
   Measured: `clock_gettime` (vdso) consumes 11.6-14.4% of on-CPU perf samples. jemalloc collectively 5-7%. Not blocking at current utilization (<2% CPU total) but would matter at higher scale.

9) [minor] it looke like logging from the nuraft (when enabled) does some extra syscalls (prbably can be avoided). 

10) System-level observations (measured during mixed benchmark, 3-node cluster):
   - CPU utilization: <2% across 32 cores. The system is entirely latency-bound (fdatasync + Raft consensus), not compute-bound.
   - Disk utilization: 3.34% (nvme, ~110 fdatasync/s, 5.7 MiB/s write throughput). The fdatasync latency (~9ms) is the device's inherent commit latency, not contention.
   - fdatasync distribution: leader 99.9% in 8-16ms; follower bimodal 59% in 4-8ms, 40.5% in 8-16ms.
   - Batching: ~46 commits/fsync, ~27KB per fsync. This is near-optimal given the device's fsync rate.
   - Leader and follower have identical futex rates (~30k/s) and context-switch rates (~13.5k/s) when measured simultaneously. An earlier non-simultaneous measurement showed 2x difference which was a timing artifact.
   - On-CPU profile is extremely flat — no single function exceeds 2.2%. Top: request deserialization, jemalloc, `KeeperStorage::processRequest`.

filimonov:
I guess will be absorbed / replaced / covered by https://github.com/ClickHouse/ClickHouse/pull/101757

PR #102422, opened 2026-04-10
Fix full text index settings disabled by compatibility setting

fm4v:
## Linked issues
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1514

The `enable_full_text_index`, `allow_experimental_full_text_index`, and `use_skip_indexes_on_data_read` settings were reverted to `false` when any `compatibility` version older than 26.1 was set. This happened because `applyCompatibilitySetting` iterates all versions above the compat value, and earlier `SettingsChangesHistory` entries had `previous_value=false`, overriding the 26.2 entries that intended to keep them enabled.

This caused `SharedDatabaseCatalog` to fail creating tables with full text indexes (e.g. `otel.otel_logs`), with error `SUPPORT_IS_DISABLED: The text index feature is disabled`, crashing the catalog thread after 10 retries.

Fixed by changing `previous_value` to `true` in all affected history entries so that reverting never disables these settings. Also added `allow_experimental_full_text_index` to `enableAllExperimentalSettings` for DDL replay safety.

Closes #101870
Resolves: https://clickhouse-inc.slack.com/archives/C034YL4P6UC/p1775841294413699

Full text index settings (`enable_full_text_index`, `allow_experimental_full_text_index`, `use_skip_indexes_on_data_read`) are no longer disabled when the `compatibility` setting points to a version older than 26.1. Previously this could prevent `SharedDatabaseCatalog` from creating tables with text indexes.

### Version info
- Merged into: `26.4.1.830`
- Backported to: `26.3.7.10`, `26.2.12.3`


grantholly-clickhouse (approved):

rschu1ze:
I read the PR message and I didn't understand it. But I am also not against this PR.

PR #102426, opened 2026-04-10
Add EBS snapshot-based Docker image cache for CI runners

maxknv:
Add EBS snapshot-based Docker image cache to eliminate Docker pull latency on CI runners.

When `ENABLE_EBS_DOCKER_CACHE` is enabled, Docker build jobs create EBS snapshots of Docker data-root after building images. Runners mount these snapshots as volumes instead of pulling from the registry - all images are available instantly. Transparent fallback to normal docker pull if no snapshot exists.

### DIFF
--- a/ci/docker/style-test/Dockerfile
+++ b/ci/docker/style-test/Dockerfile
@@ -1,5 +1,5 @@
 # docker build -t clickhouse/style-test .
-# EBS docker cache test
+# EBS docker cache test v4
 FROM ubuntu:22.04
 
 RUN apt-get update && env DEBIAN_FRONTEND=noninteractive apt-get install --yes \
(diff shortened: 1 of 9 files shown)


maxknv:
Closing: the benefits of EBS snapshot cache over the existing Docker S3 proxy do not justify the added complexity. Key concerns:

- **Cost**: each of ~2000 concurrent runners needs its own 100 GB EBS volume, making runner volumes the dominant cost.
- **Docker-in-Docker**: integration tests run Docker inside Docker with a separate data volume, so the host EBS cache is invisible to the inner daemon — the heaviest image consumers (integration tests) still pull everything.
- **Different CI/local paths**: adds a CI-only code path that does not exist in local development, making debugging harder.
- **First-run penalty**: snapshots are async (~15 min), so the first CI run after any image change still pulls from the registry.

The RFC and prototype code are preserved in this branch for future reference if the tradeoffs change (e.g., if per-image snapshots or DinD integration become feasible).

PR #101997, opened 2026-04-07
Introduce new `hasPhrase` function

ahmadov:
The first stage for the https://github.com/ClickHouse/ClickHouse/issues/101473. The new function is independent of how the positional data would be stored.

### Changelog category (leave one):

- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

Adds a `hasPhrase` (alias `matchPhrase`) function for phrase search (continuous sequences of tokens). Search is brute-force, i.e. not supported by the text index yet.

### Version info
- Merged into: `26.4.1.772`

Issue #101473, opened 2026-04-01
RFC: Positional Phrase Query Support for Text Index


ahmadov:
## Motivation

ClickHouse's text index currently stores only row IDs in posting lists — it knows *which* rows contain a token but not *where* within the row the token appears. This makes exact phrase queries (e.g., `"connection refused"`) impossible to resolve at the index level. Today, a query like:

```sql
SELECT * FROM logs WHERE matchPhrase(message, 'connection refused')
```

must either skip the text index entirely or fall back to scanning every candidate row returned by individual token lookups for `connection` and `refused`, producing a high false-positive rate.

This RFC proposes adding **optional positional information** to the text index using a Roaringish [[1]] encoding scheme that reduces phrase matching to bulk sorted-array intersection with bitmap arithmetic — avoiding per-document position list decoding entirely.

### Current Text Index Architecture

The text index stores three files per granule:

| File | Contents |
|------|----------|
| `.idx` | Sparse index: first token per dictionary block → offset in `.dct` |
| `.dct` | Dictionary blocks: tokens (front-coded) + per-token posting metadata |
| `.pst` | Posting lists: row IDs as Roaring Bitmaps or delta+bitpacked integers |

Posting lists contain **row IDs only**. The bitpacking codec (`PostingListCodecBitpacking`) already stores posting lists as sorted arrays of delta-compressed integers — this is the foundation we build on.

### Comparison

The traditional approach to phrase search (as in Lucene/Tantivy) stores per-document position arrays alongside posting lists. Phrase check requires decompressing positions per candidate document and running a cursor-based intersection — O(candidates × terms × log(positions)).

The **Roaringish encoding** [[1]] takes a fundamentally different approach: pack (doc_id, position_group, position_bitmap) into integers. Phrase check becomes bulk sorted-array intersection with bitmap shifts — O(|posting_list|) with high throughput, no per-document decode.

The Roaringish approach is chosen because:
- It operates directly on the encoded representation (no intermediate decompression)
- It is inherently vectorizable (sorted intersection + bitwise ops)
- It aligns with ClickHouse's existing delta+bitpacking posting list infrastructure
- The text index is used for **filtering** (which granules/rows to read), not highlighting — exact position recovery is unnecessary

### Roaringish Encoding with `UInt128`

Each token occurrence is encoded as a 128-bit value:

```
┌──────────────────────────────────────────────────────────────────┐
│                          UInt128                                 │
[...]

PR #102438, opened 2026-04-10
Text index: Support `hasPhrase` function in the text index

ahmadov:
Currently, `hasPhrase` does not utilize the text index in any way. This PR adds the missing support for it. However it does not store the positional information yet, intead the `HINT` mode would be used.

The query internally would be converted from:

```sql
WHERE hasPhrase(text_column, 'token1 token2')
```

to:

```sql
WHERE hasAllTokens(text_column, 'token1 token2') AND hasPhrase(text_column, 'token1 token2')
```

The added `hasAllTokens` condition would filter out rows do not contain phrase search tokens.

The proper positional information for the phrase search would be added in the follow-up PRs.

A follow-up of https://github.com/ClickHouse/ClickHouse/pull/101997 for https://github.com/ClickHouse/ClickHouse/issues/101473.

- Improvement

Add text index analysis support for the `hasPhrase` function via the `HINT` mode.

### Version info
- Merged into: `26.4.1.991`


src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
```diff
@@ -748,6 +750,17 @@ bool MergeTreeIndexConditionText::traverseFunctionNode(
         out.text_search_queries.emplace_back(std::make_shared<TextSearchQuery>(function_name, TextSearchMode::All, direct_read_mode, std::move(tokens)));
         return true;
     }
+    if (function_name == "hasPhrase")
```
Ergus, line 768: traverseAtomNode has the hard check:
```  
if (function_arguments_size != 2)
      return false;
```

So, with a third argument like `hasPhrase(col, 'my bla', 'ngrams(3)')` we always bypass the index. A suggestion for a latter PR is that we must check the tokenizer argument and compare with the index one, in that case we could use the index as well.

For now I would suggest to add a test that exercises the 3-argument form against a table that has a text index, to confirm the index is indeed bypassed instead of silently misapplied.

tests/queries/0_stateless/02346_text_index_function_hasPhrase.sql
```diff
@@ -0,0 +1,262 @@
...
+ORDER BY (id);
+
+INSERT INTO tab VALUES
+    (1, '()a()bc()d()'),
+    (2, '()d()bc()a()'),
+    (3, '()a()d()'),
+    (4, '\\a\\bc\\d\\'),
+    (5, '()a()bc()');
+
+SELECT groupArray(id) FROM tab WHERE hasPhrase(message, 'a()bc', 'splitByString([\'()\', \'\\\\\'])');
```
Ergus, line 102: Could we add a few `NOT hasPhrase` tests? IIUC the index may incorrectly prune granules where the tokens exist but not in phrase order. Because the index is in Hint mode, a post-filter must be added but this is only correct if the hint pipeline is wired up properly for negated predicates.

tests/queries/0_stateless/02346_text_index_function_hasPhrase.sql
```diff
@@ -0,0 +1,262 @@
...
+DROP TABLE tab;
+
+SELECT 'Text index analysis';
+
+DROP TABLE IF EXISTS tab;
+CREATE TABLE tab
+(
+    id UInt32,
+    message String,
+    INDEX idx(`message`) TYPE text(tokenizer = 'splitByNonAlpha')
```
Ergus: Does it worth adding a test using the `AsciiCJK`

src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
```diff
@@ -748,6 +750,17 @@ bool MergeTreeIndexConditionText::traverseFunctionNode(
...
+    {
+        /// The sparseGrams and array tokenizers are not supported with the `hasPhrase` function.
+        if (tokenizer->getType() == ITokenizer::Type::SparseGrams || tokenizer->getType() == ITokenizer::Type::Array)
+            return false;
+
+        auto tokens = stringToTokens(value_field);
+        out.function = RPNElement::FUNCTION_HAS_ALL_TOKENS;
+        out.text_search_queries.emplace_back(std::make_shared<TextSearchQuery>(function_name, TextSearchMode::All, direct_read_mode, tokens));
```
Ergus: Minor:: The other cases use `std::move(tokens)` some reason to not use it here.

Ergus (commented):
In general LGTM. Just added a few comments

docs/en/engines/table-engines/mergetree-family/textindexes.md
Ergus, line 934: I think that the new function must be mentioned in this section.

Ergus: Probably here.

docs/en/engines/table-engines/mergetree-family/textindexes.md
```diff
@@ -480,6 +480,24 @@ SELECT count() FROM table WHERE hasAnyTokens(comment, ['clickhouse', 'olap']);
 SELECT count() FROM table WHERE hasAllTokens(comment, ['clickhouse', 'olap']);
 ```
 
+#### `hasPhrase` {#functions-example-hasphrase}
+
+Function [hasPhrase](/sql-reference/functions/string-search-functions.md/#hasPhrase) matches against a phrase: all tokens must appear consecutively and in the same order as in the search string.
+
+Unlike `hasAllTokens`, which only requires all tokens to be present somewhere, `hasPhrase` requires them to appear as a consecutive sequence.
+The search phrase is tokenized using the same tokenizer configured for the index column.
```
Ergus, line 488: If there is not index, the function also works. This doesn't specify what happens then (brute force).

docs/en/engines/table-engines/mergetree-family/textindexes.md
```diff
@@ -847,6 +865,51 @@ SELECT * FROM events WHERE has(data.tags::Array(String), 'bug')
...
+Text index supports phrase search via the `hasPhrase` function.
+All tokens in the phrase must appear consecutively and in the same order in the document.
+
+The text index accelerates phrase search by intersecting the posting lists for all tokens in the phrase to identify candidate granules.
+Within those granules, ClickHouse then verifies exact token adjacency.
+
+`hasPhrase` is supported with tokenizers `splitByNonAlpha`, `splitByString`, `ngrams`, and `asciiCJK`.
+It is **not** supported with the `sparseGrams` and `array` tokenizers.
+
+The phrase string is tokenized using the index's configured tokenizer.
```
Ergus, line 878: See comment in l488

tests/queries/0_stateless/02346_function_hasPhrase.sql
```diff
@@ -152,3 +152,53 @@ SELECT id FROM tab WHERE hasPhrase(message, 'abc', 'ngrams(3)') ORDER BY id;
 SELECT id FROM tab WHERE hasPhrase(message, 'cde', 'ngrams(3)') ORDER BY id;
 
 DROP TABLE tab;
+
+SELECT 'matchPhrase alias';
+
+SELECT hasPhrase('the quick brown fox', 'quick brown');
+SELECT matchPhrase('the quick brown fox', 'quick brown');
```
Ergus, line 161: I don't see any documentation for this alias function.

docs/en/engines/table-engines/mergetree-family/textindexes.md
```diff
@@ -847,6 +865,51 @@ SELECT * FROM events WHERE has(data.tags::Array(String), 'bug')
...
+Tokenizer separator characters in the phrase are ignored: `hasPhrase(text, 'quick+brown')` is equivalent to `hasPhrase(text, 'quick brown')` for the `splitByNonAlpha` tokenizer.
+
+#### Example {#text-index-phrase-search-example}
+
+```sql
+CREATE TABLE tab (
+    id UInt32,
+    text String,
+    INDEX idx(text) TYPE text(tokenizer = splitByNonAlpha))
```
Ergus: Very minor (almost pity ;)): In other examples like l.835 the extra closing ) is in its own line

Ergus (approved):
LGTM (But I will always add some comments ;)

docs/en/engines/table-engines/mergetree-family/textindexes.md
rschu1ze: I think the same quirk also applies to `hasAny/AllTokens` ... and we also didn't document this special case above. No text index ... means no tokenizer ... means no tokenization. I think the current way it's described is fine.

What we could do (but I'm agnostic to it) is to add a 2nd array overload similar to `hasAny/AllTokens` that isn't tokenized by definition.

docs/en/engines/table-engines/mergetree-family/textindexes.md
```diff
@@ -480,6 +480,24 @@ SELECT count() FROM table WHERE hasAnyTokens(comment, ['clickhouse', 'olap']);
 SELECT count() FROM table WHERE hasAllTokens(comment, ['clickhouse', 'olap']);
 ```
 
+#### `hasPhrase` {#functions-example-hasphrase}
+
+Function [hasPhrase](/sql-reference/functions/string-search-functions.md/#hasPhrase) matches against a phrase: all tokens must appear consecutively and in the same order as in the search string.
+
+Unlike `hasAllTokens`, which only requires all tokens to be present somewhere, `hasPhrase` requires them to appear as a consecutive sequence.
+The search phrase is tokenized using the same tokenizer configured for the index column.
+Note that `hasPhrase` is not supported with the `sparseGrams` and `array` tokenizers.
```
rschu1ze: Minor: Better say what works and not what doesn't.

```
The function requires either the `splitByNonAlpha`, `splitByString`, `asciiCJK` or `array` tokenizer.
```

docs/en/engines/table-engines/mergetree-family/textindexes.md
```diff
@@ -480,6 +480,24 @@ SELECT count() FROM table WHERE hasAnyTokens(comment, ['clickhouse', 'olap']);
...
+
+Example:
+
+```sql
+-- Matches: 'clickhouse' and 'olap' must appear consecutively in that order
+SELECT count() FROM table WHERE hasPhrase(comment, 'clickhouse olap');
```
rschu1ze, line 495: Minor: let's use a real-world phrase: `'weather in New York'`.

Doesn't match:
- `New weather in York`.
- `weather and traffic in New York`

src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
```diff
@@ -748,6 +750,17 @@ bool MergeTreeIndexConditionText::traverseFunctionNode(
         out.text_search_queries.emplace_back(std::make_shared<TextSearchQuery>(function_name, TextSearchMode::All, direct_read_mode, std::move(tokens)));
         return true;
     }
+    if (function_name == "hasPhrase")
+    {
+        /// The sparseGrams and array tokenizers are not supported with the `hasPhrase` function.
+        if (tokenizer->getType() == ITokenizer::Type::SparseGrams || tokenizer->getType() == ITokenizer::Type::Array)
```
rschu1ze: See my other comment, I think we should reverse the condition and check for tokenizers that are supported. This will be safer for the case when new un-supported tokenizers are added in future.

tests/queries/0_stateless/02346_function_hasPhrase.sql
rschu1ze: Docs not needed, it's just an alias of `hasPhrase`.

tests/queries/0_stateless/02346_text_index_function_hasPhrase.sql
```diff
@@ -0,0 +1,345 @@
...
+ORDER BY (id);
+
+INSERT INTO tab VALUES
+    (1, 'the quick brown fox'),
+    (2, 'quick brown dog'),
+    (3, 'brown fox jumps'),
+    (4, 'lazy dog sleeps'),
+    (5, 'the lazy fox');
+
+-- ngrams turn a substring search into a token phrase search via the text index
```
rschu1ze: Let's convert to

```
SELECT '---- ngrams turn a substring search into a token phrase search via the text index'`
```
please

rschu1ze (approved):

docs/en/engines/table-engines/mergetree-family/textindexes.md
Ergus: In `hasAny/AllTokens` we use the default tokenizer if not index.

tests/queries/0_stateless/02346_function_hasPhrase.sql
Ergus: We don't document the aliases? At least a sentence to redirect the the original function?

rschu1ze: We do, grep "Alias" [here](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#hasPhrase).

docs/en/engines/table-engines/mergetree-family/textindexes.md
ahmadov: it's same for the hasPhrase function: https://clickhouse.com/docs/sql-reference/functions/string-search-functions#hasPhrase:~:text=Optional%2C%20defaults%20to%20splitByNonAlpha

PR #102439, opened 2026-04-10
feat: add metadata and raw results for perf tests

egor-click:

No changelog entry required.

No user-facing changes.

Motivation: To improve understanding of performance changes over time. The richer metadata (arch, instance type) and raw per-run samples enable applying more statistical methods to catch regressions, reduce false positives, and trace changes back to specific dimensions. The raw data is the foundation for all downstream analysis — it is hard to do better stats / experiment / provide common tooling / have historical insights without access to the individual measurements.

### Version info
- Merged into: `26.4.1.915`

### DIFF
--- a/ci/jobs/performance_tests.py
+++ b/ci/jobs/performance_tests.py
@@ -889,7 +889,6 @@ def main():
                 retries=3,
             )
             if insert_ok:
-                print(f"Inserted [{len(lines)}] raw query metric lines")
                 print(f"Inserted [{line_count}] raw query metric lines")
             else:
                 print(f"Inserted [{line_count}] raw query metric lines - failed")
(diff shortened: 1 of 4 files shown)


egor-click:
I can't add new columns into `query_metrics_v2` because this would break current behavior with 

```
Number of columns doesn't match (source: 13 and result: 19)
```

I will  alter table to add metadata only after this PR is merged

fm4v:
> I can't add new columns into query_metrics_v2 because this would break current behavior with

Add them with DEFAULT column constraint

egor-click:
> > I can't add new columns into query_metrics_v2 because this would break current behavior with
> 
> Add them with DEFAULT column constraint

I created with default but the problem is I believe that we didn't use columns in the insert statement , [this should be fixed in this PR](https://github.com/ClickHouse/ClickHouse/blob/04c9c9cf62be05016fd17fc1040372142741a70e/ci/jobs/performance_tests.py#L43-L63), but idk how to make the change backward compatible because of that

PR #102485, opened 2026-04-12
Add S2 projection index and spatial predicate functions

JasonLi-cn:
ClickHouse has basic S2 cell functions but lacks real spatial predicate support
(within, contains, intersects, etc.) and has no way to accelerate spatial queries
using an index. This adds both.

New spatial predicate functions: `geoWithin`, `geoContains`, `geoCoveredBy`,
`geoCovers`, `geoDWithin`, `geoEquals`, `geoIntersects`, `geoIntersectsBox`,
`geoTouches` — each with explicit Cartesian/Spherical variants plus
BigQuery-compatible `ST_*` aliases. The `st_function_use_spherical` setting
controls whether `ST_*` functions use spherical (default, matching BigQuery) or
Cartesian coordinates.

New `s2` projection index type accelerates spatial queries on geometry columns.
During INSERT, geometries are decomposed into S2 cell coverings stored as a
sorted projection. At query time, spatial predicates are rewritten to
`__s2CoveringIntersects` which uses `KeyCondition` to prune granules via
binary search over the S2 cell union's Hilbert-curve intervals.

- New Feature

Add spatial predicate functions (`geoWithin`, `geoContains`, `geoIntersects`, `geoDWithin`, etc.) with Cartesian/Spherical variants and BigQuery-compatible `ST_*`
aliases. Add `s2` projection index type that accelerates spatial queries on geometry columns using S2 cell coverings. New settings: `enable_s2_index_pruning`,
`s2_max_covering_cells`, `st_function_use_spherical`.

---
Want me to create the PR with this description, or would you like to adjust anything?


rschu1ze:
If we decide to merge this, then it needs to be under an `experimental` setting.

JasonLi-cn:
> If we decide to merge this, then it needs to be under an `experimental` setting.

Agreed. We can rename `enable_s2_index_pruning` to `allow_experimental_s2_index` to gate this feature behind an experimental setting.

PR #102492, opened 2026-04-12
Fix stop_zookeeper not waiting for ZooKeeper JVM in test_keeper_zookeeper_converter

alexey-milovidov:
`stop_zookeeper` used `get_process_pid("zookeeper")` to wait for the process to die, but `get_process_pid` uses the regex `'^[^ ]*zookeeper'` which only matches the first word of the command line. The ZooKeeper Java process starts with `java -Dzookeeper.log.dir=...`, so the first word is `java`, not `zookeeper` — the regex never matched and the function returned immediately without actually waiting for the process to exit.

Under resource-constrained environments (MSan), the JVM takes longer to shut down after `zkServer.sh stop`. When `start_zookeeper` was called later, the old process was still holding port 2181, causing the new JVM to fail with exit code 1.

Fix by using `pgrep -f org.apache.zookeeper.server` to properly detect the running ZooKeeper JVM process. Also increase start retries from 3 to 5 with 3-second delays as an additional safety net.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=00ecef1c87cfebeb52a024307b0a76a68518fb53&name_0=MasterCI&name_1=Integration%20tests%20%28amd_msan%2C%206%2F6%29

...

### Version info
- Merged into: `26.4.1.1054`

### DIFF
--- a/tests/integration/test_keeper_zookeeper_converter/test.py
+++ b/tests/integration/test_keeper_zookeeper_converter/test.py
@@ -38,8 +38,12 @@ def stop_zookeeper():
     # matches the first word of the command line. The ZooKeeper Java process
     # starts with "java -Dzookeeper..." so it never matches. Use a direct pgrep
     # for the ZooKeeper main class to properly detect the running JVM.
+    # The bracket expression [o]rg is a standard trick to keep pgrep -f from
+    # matching the shell process that is running pgrep itself (its argv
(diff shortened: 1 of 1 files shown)


alexey-milovidov (commented):
Ok.

PR #102158, opened 2026-04-09
Add MSan support for fibers and fix dragonbox struct padding

alexey-milovidov:
Re-applies #101621 (which was reverted in #102005) with an updated understanding of the root cause.

**`BOOST_USE_MSAN` for fiber context tracking:** The boost fiber implementation has MSan annotations (`__msan_start_switch_fiber`/`__msan_finish_switch_fiber`) around every `swapcontext` call, but they were gated on `BOOST_USE_MSAN` which was never defined. Without them, MSan loses track of shadow memory across fiber context switches, causing false positives like STID 4179-5154.

**Keep `__msan_unpoison` with a correct explanation:** The original PR #101621 removed `__msan_unpoison` from `FiberStack::allocate`, which caused STID 4148-3044 and the revert. Investigation traced the origin to struct padding in return values (e.g. dragonbox's `fp_t`). Value-initializing the struct does not help: **MSan tracks shadow per-field in LLVM IR, not per-byte of the padded representation.** When any struct with padding is returned by value, the padding shadow is lost. This is a known unfixed LLVM limitation:

- https://github.com/llvm/llvm-project/pull/116615 — Clang coerces small structs with padding to wide integers for register return, making padding bits poison ([alive2 proof](https://alive2.llvm.org/ce/z/DwkvmT))
- https://github.com/llvm/llvm-project/issues/54476 — canonical MSan bug: struct padding false positives (open since 2022)
- https://github.com/llvm/llvm-project/issues/58945 — Clang codegen does not emit writes for padding bytes even with value-initialization
- https://github.com/llvm/llvm-project/issues/162216 — same pattern with `struct addrinfo`

On the main thread stack this is harmless (OS zero-inits pages), but on heap-allocated fiber stacks the dirty padding shadow propagates via stack slot reuse, causing false positives. The `__msan_unpoison` in `FiberStack::allocate` is the correct workaround. Updated the comment to explain why.

A test is included that documents this MSan limitation.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102114&sha=5b68859e4e2e0023ace05c14f750a827f75ed7f5&name_0=PR&name_1=Stress%20test%20%28arm_msan%29
See also: https://github.com/ClickHouse/ClickHouse/pull/102114

Fixes: https://github.com/ClickHouse/ClickHouse/issues/101649

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

[...]

PR #102496, opened 2026-04-12
Revert "Add MSan support for fibers and fix dragonbox struct padding"

alexey-milovidov:
Reverts ClickHouse/ClickHouse#102158

After we have a more targeted fix and a test, check if this change was necessary.

### Version info
- Merged into: `26.4.1.1057`


al13n321 (approved):
Maybe add a comment in src/Common/tests/gtest_dragonbox_msan.cpp that the test was not understood by any human and may be incorrect and meaningless.

alexey-milovidov:
> Maybe add a comment in src/Common/tests/gtest_dragonbox_msan.cpp

The file already deleted after the migration from DragonBox to Żmij (snake/dragon).

PR #102501, opened 2026-04-12
Allow opening Play UI in a new tab via the ClickHouse logo

alexey-milovidov:
Make the ClickHouse logo (databases toggle) in the left panel act as a link so that Ctrl+click, Cmd+click, Shift+click, and middle-click open the current page (with the query) in a new tab. Normal clicks continue to toggle the databases panel as before.

The `<div>` is changed to an `<a>` element whose `href` is set to the current URL (including the query hash) on mousedown, so the browser handles new-tab opening natively.

- Improvement

Allow opening the Play UI in a new tab via Ctrl/Cmd/Shift+click or middle-click on the ClickHouse logo in the left panel.

### Version info
- Merged into: `26.5.1.127`


pamarcos (approved):
Clicking on the icon opens a new Play UI as expected. The only issue I find is the same I found [here](https://github.com/ClickHouse/ClickHouse/pull/101290#discussion_r3065332789). It's unexpected and weird UX that it opens just the SQL statement executed.

Image

which opens 

Image

I guess we'll address it separately because this PR itself is ok

PR #102529, opened 2026-04-13
CI: add CI Tests job to PR workflow

leshikus:
Add a `CI Tests` job that runs `ci/tests/` via pytest on every PR as an early gate alongside `Style check` and `Fast test` — builds cannot start until it passes.

## What changed

**New `CI Tests` job** (`ci/jobs/ci_tests_job.py`, `ci/defs/`, `ci/workflows/pull_request.py`): starts Docker-in-Docker, then runs pytest against `ci/tests/` via `Result.from_pytest_run`. Runs on ARM large inside `clickhouse/integration-tests-runner` (privileged, with a dedicated Docker volume) so tests that need a real ClickHouse server can work. Multi-pattern `--test` filtering uses `shlex.quote` so the full `pytest -k` expression is passed as a single shell token.

**New `ClickHouseService`** (`ci/jobs/scripts/clickhouse_service.py`): a plain context manager that downloads the ClickHouse binary from S3 master builds on first use, copies the default server config, recreates the data directory (avoids the `MISMATCHING_USERS_FOR_PROCESS_AND_DATA` error when UID changes between runs), starts the server in a new session, polls for readiness, and shuts it down via `SIGTERM` / `SIGKILL`.

**praktika fixes**:
- Added `--init` to every `docker run` invocation so Docker's init process reaps zombies and forwards signals correctly.
- Fixed `PermissionError` on `result_ci_tests.json`: when a job runs as root inside Docker, the result file is root-owned; the host process then can't overwrite it. Fixed by running `chown` immediately after `process.wait()`, before `Result.from_fs` and `result.dump()`.
- Added `stopwatch` to `ResultTranslator.from_pytest_jsonl` so the reported duration covers the full parse, not just teardown.

**Test fixture** (`ci/tests/0_stateless/01_parallel_sleep.sh`): replaced the old `00058_select_sleep_3.sh`-based approach with a self-contained script that spawns 5 `sleep` subprocesses and writes output via `echo`, eliminating the `SELECT 1` race and the dependency on a running ClickHouse server.

**Test fix** (`ci/tests/test_cleanup_test_groups.py`): extended the stdout-file deadline to 15 s; polls for a stable process count before asserting; updated all comments to match the current fixture and `write_text_atomic`-based PGID tracking.

...

### Version info
- Merged into: `26.4.1.1090`


ci/jobs/ci_tests_job.py
```diff
@@ -0,0 +1,40 @@
...
+        action="extend",
+    )
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    args = parse_args()
+    pytest_command = "ci/tests/"
+    if args.test:
+        pytest_command += " -k " + " or ".join(args.test)
```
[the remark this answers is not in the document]
leshikus: will address in a separate PR

ci/jobs/scripts/clickhouse_service.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @staticmethod
+    def _download_binary() -> None:
+        dest = Path(temp_dir) / "clickhouse"
+        if dest.exists():
+            print(f"ClickHouse binary already present at [{dest}], skipping download")
+            dest.chmod(0o755)
+            return
+        arch = "aarch64" if Utils.is_arm() else "amd64"
+        url = f"https://clickhouse-builds.s3.us-east-1.amazonaws.com/master/{arch}/clickhouse"
```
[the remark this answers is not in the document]
leshikus: that's ok for CI validation

ci/praktika/runner.py
```diff
@@ -482,6 +482,17 @@ def _run(
...
+            # When running Docker containers as root (non-rootless mode), any files
+            # created by the job will be owned by root.  Fix ownership here, before
+            # reading the result file or writing the host-side result, so that the
+            # host user can open them without a PermissionError.
+            if job.run_in_docker and not no_docker and from_root:
+                print(f"--- Fixing file ownership after running docker as root")
+                uid = os.getuid()
+                gid = os.getgid()
+                chown_cmd = f"docker run --rm --user root --volume {host_dir_q}:{current_dir} --workdir={current_dir} {docker} chown -R {uid}:{gid} {Settings.TEMP_DIR}"
+                Shell.run(chown_cmd)
```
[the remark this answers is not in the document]
leshikus: better to proceed with some artifacts

.claude/skills/clickhouse-pr-description/SKILL.md
```diff
@@ -124,4 +124,6 @@ These patterns make PRs harder to read or signal low effort:
 
 It's fine to mention AI assistance openly. `Co-Authored-By:` in commits or acknowledgements in the PR description are acceptable — ClickHouse is open about AI-assisted development.
 
+When creating a PR, open it as a **Draft** until it is ready for review — this prevents accidental merges and lets CI run without triggering reviewer notifications.
```
maxknv: I think it should be enough to have "When creating a PR, open it as a **Draft**".
Please open a separate pr for this for other folks to review

ci/defs/job_configs.py
```diff
@@ -182,6 +182,14 @@ class JobConfigs:
         allow_merge_on_failure=True,
         enable_gh_auth=True,
     )
+    ci_tests = Job.Config(
+        name=JobNames.CI_TESTS,
+        runs_on=RunnerLabels.ARM_LARGE,
```
maxknv, line 187: ARM_SMALL?

ci/jobs/scripts/clickhouse_proc.py
```diff
@@ -251,6 +251,13 @@ def set_memory_ratio(ratio):
             f"Set max_server_memory_usage_to_ram_ratio to {ratio} in {file_path}"
         )
 
+    def install_light(self):
```
maxknv: just rename _install_light if it needs to be public

ci/defs/job_configs.py
```diff
@@ -182,6 +182,14 @@ class JobConfigs:
         allow_merge_on_failure=True,
         enable_gh_auth=True,
     )
+    ci_tests = Job.Config(
```
maxknv, line 185: let's add digest_config. I think it can be just a "./ci" path in it

tests/queries/0_stateless/00058_select_sleep_3.sh
maxknv: we should keep ci test separate from clickhouse tests. can you move it to ./ci/tests?

maxknv (changes requested):

leshikus:
@maxknv I wonder if using `with ClickHouseService() as service:` than calling `_install_light` and several other methods is more elegant and pythonistic way to describe a service

ci/defs/job_configs.py
leshikus: it launches the whole fuzzer container, e2e can be pretty heavy IMHO

ci/jobs/scripts/clickhouse_proc.py
leshikus: agreed to keep the current service

tests/queries/0_stateless/00058_select_sleep_3.sh
leshikus: moved to ci/tests

maxknv (approved):

PR #89757, opened 2025-11-08
Text Search: Consistent behavior for short ngrams and empty tokens

george-larionov:
1. Making ngram tokenizer behavior consistent, it will now not tokenize ngrams shorter than N
2. Text search will return no rows when search tokens are empty. (Consistent with other implementations)

### Changelog category (leave one):
- Backward Incompatible Change

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Ngram tokenizer will no longer return ngrams less than the length N it was set with. Text Search will return no rows when search tokens are empty.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

### Details

Fixing inconsistent behavior of ngram tokenizer, now it will not return ngrams less than the ngram size N it was set with in any scenario.

PR #102544, opened 2026-04-13
 After stringToTokens returns empty, check the original needle value.

### LINKED ISSUE #102497: `hasToken` with separator-only needle silently returns 0 with text index instead of throwing BAD_ARGUMENTS

_Retrospective finding from a historical scan of [PR #87564](https://github.com/ClickHouse/ClickHouse/pull/87564) (merged 2025-09-26). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`hasToken(text, '()')` returns 0 rows when the table has a text index, but throws BAD_ARGUMENTS when the table has no text index. The text index suppresses the expected exception by incorrectly skipping all granules.

**Root cause:** MergeTreeIndexConditionText.cpp:744: `if (tokens.empty())` does not distinguish between "empty because the needle was empty" and "empty because the needle contained only separator characters". Both cases trigger `tokens.push_back("")`, but only the empty-needle case should skip all granules. Separator-only needles should fall through to the function evaluation which throws BAD_ARGUMENTS.

**Why we believe this is a bug:** MergeTreeIndexConditionText.cpp:743-745 → `stringToTokens('()')` returns empty (no alphanumeric tokens in `()`) → `tokens.empty()` is true → `tokens.push_back("")` → text index looks for empty string token `""` in inverted index → not found → returns false → all granules skipped → function never evaluated → no exception thrown

**Affected locations:**
- `src/Storages/MergeTree/MergeTreeIndexConditionText.cpp:744` — tokens.empty() check in hasToken/hasTokenOrNull text index path

**Impact:** Queries using `hasToken(column, needle)` where needle contains only separator characters (e.g., `()`, `!!!`, `***`, spaces) silently return 0 rows on tables with text indexes instead of raising BAD_ARGUMENTS. Users relying on the exception to detect invalid needles get no feedback. `NOT hasToken(text, '()')` returns all rows instead of throwing.

Yes — confirmed on current `master` (commit `72bdd3a2f384`).

### How to reproduce

```sql
-- Test: hasToken with separator-only needle should throw BAD_ARGUMENTS consistently
-- regardless of whether a text index exists on the table.
-- Covers: MergeTreeIndexConditionText.cpp:744 - tokens.push_back("") incorrectly
-- treats separator-only needles the same as empty needles.

SET allow_experimental_full_text_index = 1;

-- Table WITH text index
DROP TABLE IF EXISTS test_sep_with_idx;
CREATE TABLE test_sep_with_idx (
    id Int,
    text String,
    INDEX idx_text(text) TYPE text(tokenizer = 'splitByNonAlpha')
)
ENGINE = MergeTree()
ORDER BY (id);

INSERT INTO test_sep_with_idx VALUES(1, 'hello world'), (2, 'foo bar');

-- Table WITHOUT text index
DROP TABLE IF EXISTS test_sep_no_idx;
CREATE TABLE test_sep_no_idx (
    id Int,
    text String
)
ENGINE = MergeTree()
ORDER BY (id);

INSERT INTO test_sep_no_idx VALUES(1, 'hello world'), (2, 'foo bar');

-- hasToken with separator-only needle should throw in both cases
SELECT count() FROM test_sep_no_idx WHERE hasToken(text, '()'); -- { serverError BAD_ARGUMENTS }
SELECT count() FROM test_sep_with_idx WHERE hasToken(text, '()'); -- { serverError BAD_ARGUMENTS }

-- NOT hasToken with separator-only needle should also throw in both cases
SELECT count() FROM test_sep_no_idx WHERE NOT hasToken(text, '()'); -- { serverError BAD_ARGUMENTS }
SELECT count() FROM test_sep_with_idx WHERE NOT hasToken(text, '()'); -- { serverError BAD_ARGUMENTS }

-- hasTokenOrNull with separator-only needle: without index returns NULL
-- With text index should also return NULL (not skip granules)
SELECT '-- hasTokenOrNull without index';
SELECT hasTokenOrNull('hello', '()');

SELECT '-- hasTokenOrNull with text index';
SELECT id FROM test_sep_with_idx WHERE hasTokenOrNull(text, '()') IS NULL ORDER BY id;

SELECT '-- hasTokenOrNull without text index';
SELECT id FROM test_sep_no_idx WHERE hasTokenOrNull(text, '()') IS NULL ORDER BY id;

DROP TABLE IF EXISTS test_sep_with_idx;
DROP TABLE IF EXISTS test_sep_no_idx;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/62416668-2149-4e28-b720-aeccb4ed8258)

### Expected behavior

```
-- hasTokenOrNull without index
\N
-- hasTokenOrNull with text index
1
2
-- hasTokenOrNull without text index
1
2
```

```
0
The query succeeded but the server error '36' was expected (query: SELECT count() FROM test_sep_with_idx WHERE hasToken(text, '()'); -- { serverError BAD_ARGUMENTS }).
2
The query succeeded but the server error '36' was expected (query: SELECT count() FROM test_sep_with_idx WHERE NOT hasToken(text, '()'); -- { serverError BAD_ARGUMENTS }).
-- hasTokenOrNull without index
\N
-- hasTokenOrNull wit
```

### Additional context

**Open risks:**
- Same pattern might affect `hasTokenOrNull` with separator needles where NULL semantics differ from 0-row result

**Suggested fix:** Check whether the original value was truly empty before pushing the empty sentinel token: `if (tokens.empty() && value_field.safeGet<String>().empty()) tokens.push_back("");` — for non-empty separator-only needles, return false from `traverseFunctionNode` so the text index pushes FUNCTION_UNKNOWN and all granules are read, allowing the function to throw BAD_ARGUMENTS.

Found during automated review of [PR #87564](https://github.com/ClickHouse/ClickHouse/pull/87564).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr87564_001`_

Ergus:
Fixes: #102497

See issue description there.

Fixed `hasToken` / `hasTokenOrNull` with separator-only needles (e.g. `'()'`, `'!!!'`) on columns with a text index: previously the index silently skipped all granules instead of throwing `BAD_ARGUMENTS` (for `hasToken`) or returning `NULL` (for `hasTokenOrNull`).

### Version info
- Merged into: `26.4.1.905`

### DIFF
--- a/src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
+++ b/src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
@@ -745,8 +745,17 @@ bool MergeTreeIndexConditionText::traverseFunctionNode(
         {
             const String & string_needle = value_field.safeGet<String>();
             if (!string_needle.empty())
-                return false;
-
+            {
+                /// hasToken uses splitByNonAlpha as its tokenizer, so:
+                ///  - A needle without any word character (alphanumeric or non-ASCII) is invalid.
+                ///  - Bypass the index in that case so the row-level evaluation throws BAD_ARGUMENTS (or returns NULL for hasTokenOrNull)
+                ///  -- Consistnt with the no-index behaviour.
+                /// If the needle does contain word characters (e.g. "abc" with ngrams(4)):
+                ///  - It is valid but too short for the index's tokenizer:
+                ///  -- Fall through to push "" so all granules are pruned and the query returns 0 rows.
+                if (std::ranges::none_of(string_needle, [](unsigned char c) { return !isASCII(c) || isAlphaNumericASCII(c); }))
+                    return false;
+            }
             tokens.push_back("");
         }
 
--- a/src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
+++ b/src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
@@ -742,7 +742,13 @@ bool MergeTreeIndexConditionText::traverseFunctionNode(
     {
         auto tokens = stringToTokens(value_field);
         if (tokens.empty())
+        {
+            const String & string_needle = value_field.safeGet<String>();
+            if (!string_needle.empty())
+                return false;
+
             tokens.push_back("");
+        }
 
         out.function = RPNElement::FUNCTION_EQUALS;
         out.text_search_queries.emplace_back(std::make_shared<TextSearchQuery>(function_name, TextSearchMode::All, direct_read_mode, std::move(tokens)));
--- a/tests/queries/0_stateless/02346_text_index_bug102497.reference
+++ b/tests/queries/0_stateless/02346_text_index_bug102497.reference
@@ -1,0 +1,17 @@
+Scalar baseline.
+- hasTokenOrNull with separator-only needle returns NULL
+\N
+Without text index.
+- hasToken with separator-only needle throws BAD_ARGUMENTS
+- hasTokenOrNull with separator-only needle returns NULL per row
+2
+- Empty needle matches nothing, no exception
+0
+2
+With text index.
+- hasToken with separator-only needle throws BAD_ARGUMENTS
+- hasTokenOrNull with separator-only needle returns NULL per row
+2
+- Empty needle matches nothing, no exception
+0
+2
--- a/tests/queries/0_stateless/02346_text_index_bug102497.sql
+++ b/tests/queries/0_stateless/02346_text_index_bug102497.sql
@@ -1,0 +1,58 @@
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/102497
+-- hasToken with separator-only needles should throw BAD_ARGUMENTS consistently,
+-- regardless of whether a text index is present.
+-- hasTokenOrNull with separator-only needles should return NULL per row,
+-- not silently skip all granules.
+
+SELECT 'Scalar baseline.';
+
+SELECT '- hasTokenOrNull with separator-only needle returns NULL';
+SELECT hasTokenOrNull('hello', '()');
+
+SELECT 'Without text index.';
+
+DROP TABLE IF EXISTS tab;
+CREATE TABLE tab (id Int32, text String) ENGINE = MergeTree() ORDER BY (id);
+INSERT INTO tab VALUES (1, 'hello world'), (2, 'foo bar');
+
+SELECT '- hasToken with separator-only needle throws BAD_ARGUMENTS';
+SELECT count() FROM tab WHERE hasToken(text, '()');     -- { serverError BAD_ARGUMENTS }
+SELECT count() FROM tab WHERE NOT hasToken(text, '()'); -- { serverError BAD_ARGUMENTS }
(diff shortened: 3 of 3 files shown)


Ergus:
@george-larionov I think that these changes are in conflict with changes in #89757

src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
```diff
@@ -742,7 +742,22 @@ bool MergeTreeIndexConditionText::traverseFunctionNode(
...
+            if (!string_needle.empty())
+            {
+                /// hasToken uses splitByNonAlpha as its tokenizer, so:
+                ///  - A needle without any word character (alphanumeric or non-ASCII) is invalid.
+                ///  - Bypass the index in that case so the row-level evaluation throws BAD_ARGUMENTS (or returns NULL for hasTokenOrNull)
+                ///  -- Consistnt with the no-index behaviour.
+                /// If the needle does contain word characters (e.g. "abc" with ngrams(4)):
+                ///  - It is valid but too short for the index's tokenizer:
+                ///  -- Fall through to push "" so all granules are pruned and the query returns 0 rows.
+                if (std::ranges::none_of(string_needle, [](unsigned char c) { return !isASCII(c) || isAlphaNumericASCII(c); }))
```
ahmadov, line 756: I guess this condition is always true, right?

ahmadov (approved):

src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
ahmadov: i'd suggest to remove that condition completely and just `return false;`.

Ergus: I implemented like that initially and broke compatibility with previous behavior.

That's why I needed to add the compatibility `none_of` check. See [a5e8f87](https://github.com/ClickHouse/ClickHouse/pull/102544/commits/a5e8f87924e7f7add314db036d1fa73d5db1ee5d)

due to this: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102544&sha=f571e0e083bdef27b127c9b986aecb59a28557da&name_0=PR

Ergus: | I guess this condition is always true, right?

Not really, suppose this: "()" it is ascii but not alphanumeric

ahmadov: I'm confused due to the failed test here https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102544&sha=f571e0e083bdef27b127c9b986aecb59a28557da&name_0=PR. 

The data is inserted here: https://github.com/ClickHouse/ClickHouse/blob/772233087cb52b3a47e1babb2799542595d273da/tests/queries/0_stateless/02346_text_index_short_ngrams_and_empty_tokens.sh#L27-L30

and query is:
```sql
SELECT count() FROM tab WHERE hasToken(message, 'abc');
```

clearly, there is `abc` in the dataset but we expect to match nothing?

Ergus: Because it is too short for ngrams. That's how it was and I didn't want to change behavior

PR #83603, opened 2025-07-10
Fix overflow in NumericIndexedVector

rienath:
Resolves #83591

Now, we will return an error if the user tries to add a number that is too large / small to store in BSI.

### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

PR #88840, opened 2025-10-21
Add pointwise and global min/max functions to `NumericIndexedVector`

FriendLey:
### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

Added pointwise min/max operations and global min/max value functions for NumericIndexedVector. Also improved and refactored the internal implementation of `pointwiseLess` for better correctness, performance, and code reuse across signed and unsigned numeric types. Resolves #88274

#### Motivation:
These additions extend the NumericIndexedVector family with commonly used numerical comparison and reduction capabilities, enabling more expressive analytical queries and feature computation over vectorized numeric aggregates.
This work also addresses user requests for richer NumericIndexedVector operators — see [issue #88274](https://github.com/ClickHouse/ClickHouse/issues/88274).

#### New functions:
 - numericIndexedVectorPointwiseMax(v1, v2) — pointwise maximum between two numeric indexed vectors.
 - numericIndexedVectorPointwiseMin(v1, v2) — pointwise minimum between two numeric indexed vectors.
 - `numericIndexedVectorPointwise*`— functions now accept a Bitmap as the second argument.
 - numericIndexedVectorGetMaxValue(v) — returns the maximum value stored in a single numeric indexed vector.
 - numericIndexedVectorGetMinValue(v) — returns the minimum value stored in a single numeric indexed vector.

All functions are compatible with both signed and unsigned numeric types, as well as floating-point vectors.

#### Other changes:
 - Reimplemented and refactored pointwiseLess that correctly handles signed two's-complement representations.
 - Unified common bitmap selection logic to enhance code reuse and reduce redundant Roaring bitmap operations.
 - Introduced shared helpers for efficient bit-slice accumulation and extrema computation.

#### Example usage:

``` SQL
WITH
    numericIndexedVectorBuild(mapFromArrays([1, 2, 3], [10.0, -2.5, 5.5])) AS vec1,
    numericIndexedVectorBuild(mapFromArrays([1, 2, 3], [3.0, -8.0, 7.2]))  AS vec2
SELECT
    numericIndexedVectorToMap(numericIndexedVectorPointwiseMax(vec1, vec2))  AS max_result,
    numericIndexedVectorToMap(numericIndexedVectorPointwiseMin(vec1, vec2))  AS min_result,
    numericIndexedVectorGetMaxValue(vec1) AS global_max,
[...]

PR #102546, opened 2026-04-13
Fix overflow in scalar pointwise ops on `numericIndexedVector`

### LINKED ISSUE #102210: initializeFromVectorAndValue still has unfixed overflow identical to the one fixed in addValue

### Describe what's wrong

Scalar pointwise operations (e.g. numericIndexedVectorPointwiseAdd(vec, 1.54743e+26)) silently produce wrong results due to Int64 overflow in initializeFromVectorAndValue, even though the same overflow was fixed in addValue by this PR

**Root cause:** AggregateFunctionGroupNumericIndexedVectorDataBSI.h:302 — initializeFromVectorAndValue converts value to Int64 via `Int64(value * static_cast<ScaledValueType>(1ULL << fraction_bit_num))` without any overflow check, identical to the old addValue code that this PR fixed

**Why we believe this is a bug:** numericIndexedVectorPointwiseAdd(vec, scalar) → pointwiseAdd(lhs, rhs, res) at line 545 → rhs_vec.initializeFromVectorAndValue(lhs, rhs) → line 302: Int64(value * ...) overflows

**Affected locations:**
- `src/AggregateFunctions/AggregateFunctionGroupNumericIndexedVectorDataBSI.h:302` — unfixed overflow in initializeFromVectorAndValue

**Impact:** Silent wrong results: adding a large positive scalar to a vector produces negative values. Any scalar pointwise operation (add, subtract, multiply, less, greater, equal, etc.) with an overflowing value corrupts data silently instead of throwing INCORRECT_DATA like addValue now does.

Yes — confirmed on current `master` (commit `9cd08aca73ad`).

### How to reproduce

```sql
-- Test: initializeFromVectorAndValue has the same overflow as addValue but was not fixed
-- Covers: AggregateFunctionGroupNumericIndexedVectorDataBSI.h:302 unfixed overflow in initializeFromVectorAndValue

SET max_execution_time = 30;

DROP TABLE IF EXISTS test_bsi_overflow_f64;
CREATE TABLE test_bsi_overflow_f64 (uin UInt8, value Float64) ENGINE = MergeTree() ORDER BY uin;
INSERT INTO test_bsi_overflow_f64 VALUES (1, 7.3), (2, 8.3);

-- Scalar pointwise add with overflowing float goes through initializeFromVectorAndValue
-- Expected: INCORRECT_DATA error (like addValue throws), Actual: silent overflow producing negative values
WITH (SELECT groupNumericIndexedVectorState(uin, value) FROM test_bsi_overflow_f64) AS vec_1
SELECT numericIndexedVectorToMap(numericIndexedVectorPointwiseAdd(vec_1, 1.54743e+26)); -- { serverError INCORRECT_DATA }

WITH (SELECT groupNumericIndexedVectorState(uin, value) FROM test_bsi_overflow_f64) AS vec_1
SELECT numericIndexedVectorToMap(numericIndexedVectorPointwiseAdd(vec_1, -1.54743e+26)); -- { serverError INCORRECT_DATA }

WITH (SELECT groupNumericIndexedVectorState(uin, value) FROM test_bsi_overflow_f64) AS vec_1
SELECT numericIndexedVectorToMap(numericIndexedVectorPointwiseMultiply(vec_1, 1.54743e+26)); -- { serverError INCORRECT_DATA }

DROP TABLE test_bsi_overflow_f64;

DROP TABLE IF EXISTS test_bsi_overflow_u64;
CREATE TABLE test_bsi_overflow_u64 (uin UInt32, value UInt64) ENGINE = MergeTree() ORDER BY uin;
INSERT INTO test_bsi_overflow_u64 VALUES (1, 100), (2, 200);

-- UInt64 max through scalar pointwise - should also error
WITH (SELECT groupNumericIndexedVectorState(uin, value) FROM test_bsi_overflow_u64) AS vec_1
SELECT numericIndexedVectorToMap(numericIndexedVectorPointwiseAdd(vec_1, 18446744073709551615)); -- { serverError INCORRECT_DATA }

DROP TABLE test_bsi_overflow_u64;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/4a81e1c7-1ed7-4e2d-be9f-031e7612d5d3)

### Expected behavior

```
All four queries should throw INCORRECT_DATA (error code 117), producing no output rows
```

```
{1:-549755813880.7,2:-549755813879.7}
{1:-549755813880.7,2:-549755813879.7}
{1:-5.960464477539063e-8,2:-5.960464477539063e-8}
{1:99,2:199}
```

### Additional context

**Open risks:**
- All 6 callers of initializeFromVectorAndValue (lines 545, 613, 1219, 1516, 1591, 1618) are affected — every scalar pointwise operation

**Suggested fix:** Apply the same overflow checks from addValue (lines 1664-1688) to initializeFromVectorAndValue: check UInt64 > Int64::max, and for Float types check fabs(value) > lim / scaling before converting.

Found during automated review of [PR #83603](https://github.com/ClickHouse/ClickHouse/pull/83603).

---
_ClickGapAI · Confidence: HIGH · Severity: P0 · Finding: `h_pr83603_001`_

FriendLey:
Closes https://github.com/ClickHouse/ClickHouse/issues/102210

Fix silent Int64 overflow in `initializeFromVectorAndValue` for BSI-based `NumericIndexedVector`. Scalar pointwise operations (`numericIndexedVectorPointwiseAdd/Subtract/Multiply/Less/LessEqual/Greater/GreaterEqual` with a scalar) route through `initializeFromVectorAndValue`, which previously converted `value` to `Int64` without any range check — identical to the overflow that was already fixed in `addValue`. As a result, passing a `UInt64`  above `Int64::max` silently produced corrupted negative values instead of throwing.

This PR applies the same overflow checks used in `addValue`:
- For `UInt64`: reject values above `Int64::max`.

This now throws `INCORRECT_DATA` consistently across all 6 callers of `initializeFromVectorAndValue`.

Found during automated review of #83603.

Fixed silent overflow in scalar pointwise operations on numericIndexedVector (e.g. `numericIndexedVectorPointwiseAdd`) when the scalar is out of range. Such inputs now raise `INCORRECT_DATA` instead of returning corrupted values.

### Version info
- Merged into: `26.5.1.517`

### DIFF
--- a/src/AggregateFunctions/AggregateFunctionGroupNumericIndexedVectorDataBSI.h
+++ b/src/AggregateFunctions/AggregateFunctionGroupNumericIndexedVectorDataBSI.h
@@ -298,8 +298,6 @@ public:
           * - When value is a Float32/Float64, fraction_bit_num indicates how many bits are used to represent the decimal, Because the
           *   maximum value of total_bit_num(integer_bit_num + fraction_bit_num) is 64, overflow may occur.
           */
-        using ScaledValueType = std::conditional_t<std::is_floating_point_v<ValueType>, ValueType, UInt64>;
-
         Int64 scaled_value;
         if constexpr (std::is_same_v<ValueType, UInt64>)
         {
@@ -325,7 +323,7 @@ public:
         }
         else
         {
-            scaled_value = static_cast<Int64>(value * static_cast<ScaledValueType>(1ULL << fraction_bit_num));
+            scaled_value = static_cast<Int64>(value * static_cast<UInt64>(1ULL << fraction_bit_num));
         }
         for (size_t i = 0; i < total_bit_num; ++i)
         {
@@ -1326,7 +1324,6 @@ public:
 
         /// Convert the scalar to the same fixed-point two's complement representation
         /// used by initializeFromVectorAndValue, then compare bit by bit.
-        using ScaledValueType = std::conditional_t<std::is_floating_point_v<ValueType>, ValueType, UInt64>;
         UInt64 scaling = 1ULL << lhs.fraction_bit_num;
 
         Int64 scaled_value;
@@ -1341,7 +1338,7 @@ public:
         }
         else
         {
-            scaled_value = static_cast<Int64>(rhs * static_cast<ScaledValueType>(scaling));
+            scaled_value = static_cast<Int64>(rhs * scaling);
         }
 
         UInt64 bit_pattern = static_cast<UInt64>(scaled_value);
--- a/src/AggregateFunctions/AggregateFunctionGroupNumericIndexedVectorDataBSI.h
+++ b/src/AggregateFunctions/AggregateFunctionGroupNumericIndexedVectorDataBSI.h
@@ -301,7 +301,14 @@ public:
         using ScaledValueType = std::conditional_t<std::is_floating_point_v<ValueType>, ValueType, UInt64>;
 
         Int64 scaled_value;
-        if constexpr (std::is_floating_point_v<ValueType>)
+        if constexpr (std::is_same_v<ValueType, UInt64>)
+        {
+            if (value > std::numeric_limits<Int64>::max())
+                throw Exception(ErrorCodes::INCORRECT_DATA,
+                    "Value {} does not fit in Int64. It should, even when using UInt64.", value);
+            scaled_value = static_cast<Int64>(value);
+        }
+        else if constexpr (std::is_floating_point_v<ValueType>)
         {
             UInt64 scaling = 1ULL << fraction_bit_num;
             auto scaled = static_cast<Float64>(value * static_cast<ValueType>(scaling));
--- a/tests/queries/0_stateless/03463_numeric_indexed_vector_overflow.sql
+++ b/tests/queries/0_stateless/03463_numeric_indexed_vector_overflow.sql
@@ -66,3 +66,72 @@ INSERT INTO test SELECT groupNumericIndexedVectorState(toUInt32(2), -1.54743e+26
 INSERT INTO test2 SELECT groupNumericIndexedVectorState(toUInt32(1), 18446744073709551615); -- { serverError INCORRECT_DATA }
 DROP TABLE test;
 DROP TABLE test2;
+
+-- Comprehensive coverage of scalar pointwise ops routing through initializeFromVectorAndValue
+-- Float64: overflow on every scalar pointwise op that touches initializeFromVectorAndValue.
+DROP TABLE IF EXISTS uin_value_f64;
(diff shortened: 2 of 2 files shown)


src/AggregateFunctions/AggregateFunctionGroupNumericIndexedVectorDataBSI.h
```diff
@@ -301,7 +301,14 @@ class BSINumericIndexedVector
         using ScaledValueType = std::conditional_t<std::is_floating_point_v<ValueType>, ValueType, UInt64>;
```
rienath: Now `ScaledValueType` is only used in the final else branch. Worth dropping the alias and inlining `static_cast<UInt64>(1ULL << fraction_bit_num)` at the use site?

rienath: nit: not in this PR's diff, but would be great to fix the following too while we are at it.

The `pointwiseEqual(lhs, scalar)` overload around line 1322 has the same `using ScaledValueType = std::conditional_t<...>` pattern, but there it's fully unnecessary. The floating-point branch handles its own cast, so the alias is only consumed by the else branch which catches every non-float type. For all of those `ScaledValueType` resolves to `UInt64`, so the `conditional_t` is dead.

Let's just remove the `using ScaledValueType ...` there and change `scaled_value = static_cast<Int64>(rhs * static_cast<ScaledValueType>(scaling));` to `scaled_value = static_cast<Int64>(rhs * scaling);`

Feel free to ignore if you want, I will then make a commit with the fix :)

rienath (commented):
Left a nit comment, but otherwise looks good!

src/AggregateFunctions/AggregateFunctionGroupNumericIndexedVectorDataBSI.h
FriendLey: @rienath  Done — dropped both ScaledValueType aliases. In initializeFromVectorAndValue the use site now uses static_cast<UInt64> directly, and in the pointwiseEqual(lhs, scalar) overload rhs * scaling is enough since scaling is already UInt64. Thanks for the suggestion!

By the way, if you have some spare cycles, would you mind taking a look at #88840 as well? It's in the same area — adds pointwise/global min/max to NumericIndexedVector. No rush.

rienath (approved):
Looks good, thanks!

clickgapai:
Hi @FriendLey @rienath — while reviewing this PR I found the following:

- **Bug (P1):** `pointwiseEqual` missing UInt64 overflow check at line 1339-1341 — https://github.com/ClickHouse/ClickHouse/issues/104589

PR #102554, opened 2026-04-13
Speed up test 01655_plan_optimizations_optimize_read_in_window_order

cwurm:
The test was spending ~6.5 seconds almost entirely on connection overhead from 24 separate `$CLICKHOUSE_CLIENT` invocations (~270ms each). This caused CI timeouts, e.g. https://github.com/ClickHouse/ClickHouse/pull/97744.

Batch queries into 5 multi-statement client calls instead of 24:
- All DDL (drops, creates, optimizes) into one call
- Each group of `EXPLAIN PLAN` queries into one call, using `SELECT` as inline markers instead of `echo`, then filtering combined output with `grep`
- Complex ORDER BY section (DDL + queries) into one call
- All final drops into one call

Also reduced `numbers_mt(100000)` to `numbers_mt(1000)` since all queries on the large tables are `EXPLAIN PLAN` which only depends on table schema, not data volume.

...

### Version info
- Merged into: `26.4.1.1045`

### DIFF
--- a/tests/queries/0_stateless/01655_plan_optimizations_optimize_read_in_window_order.sh
+++ b/tests/queries/0_stateless/01655_plan_optimizations_optimize_read_in_window_order.sh
@@ -6,54 +6,48 @@ CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 
 name=test_01655_plan_optimizations_optimize_read_in_window_order
 
-$CLICKHOUSE_CLIENT -nm -q "
-drop table if exists ${name};
-drop table if exists ${name}_n;
-drop table if exists ${name}_n_x;
-create table ${name} engine=MergeTree order by tuple() as select toInt64((sin(number)+2)*65535)%10 as n, number as x from numbers_mt(1000);
-create table ${name}_n engine=MergeTree order by n as select * from ${name} order by n;
-create table ${name}_n_x engine=MergeTree order by (n, x) as select * from ${name} order by n, x;
-optimize table ${name}_n final;
-optimize table ${name}_n_x final;
-"
+$CLICKHOUSE_CLIENT -q "drop table if exists ${name}"
+$CLICKHOUSE_CLIENT -q "drop table if exists ${name}_n"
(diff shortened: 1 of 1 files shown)


alexbakharew:
Hi @cwurm,
normally this test takes 10-12s to finish. This is OK.

This failure with timeout is very sporadic - I see only 3 occurrences in CI DB at all. Also, test was introduced 3 years ago and only in November 2025 there is a first failure. So, I think that using `$CLICKHOUSE_CLIENT` should not be the problem. 

The interesting part is - why it can take more than 3 minutes to execute. 

Have you looked into the server logs?

cwurm:
@alexbakharew Yep, I've looked at the logs, but nothing I can see. There is no error or anything. I know CI can sometimes be very, very slow so if it can take 10-12s under normal conditions can it not take much longer under bad conditions?

alexbakharew:
To be honest, I am not aware that CI can be eventually slow and then fast again. If it is really so, it should be investigated.

I am looking into the logs right now. Will get back shortly

alexbakharew:
@cwurm 
I found out that in this particular run `OPTIMIZE TABLE ... _n FINAL` and `OPTIMIZE TABLE ... _n_x FINAL` took around 2 minutes each.  In all other runs they finish in under a second.

Reason: 
In the flaky check each run creates a new TestCase object, which rolls random MergeTree settings independently. This particular run got this setting:
* index_granularity = 4 — creates 25,000 granules for just 100K rows (other runs had 1482–65529, giving 2–67 granules)

As a result the `merge` takes too much time.

We can prove it in our CI by using `SETTINGS index_granularity = 4` during table creation.

If this theory is right, we can just pin the value of  `index_granularity` to some more real value and the problem should be gone.

cwurm:
Ok, thanks. In this PR, I also changed the number of rows in the table from 100k to 1k, so that should fix it as well. I don't mind which solution we use.

cwurm:
@alexbakharew If you're ok with the approach, can you merge it, please?

alexbakharew:
@cwurm In my opinion the proper fix is to pin `index_granularity` value in this test.

PR #102567, opened 2026-04-13
 Fix stress test log parser losing actual error behind "Parse failure error"

rienath:
When a stress test server crashes and cannot restart, `stress_runner.sh` exits before moving 'stderr.log' to the test output directory. The log parser then receives an empty `stderr_log` path and hits a bare `assert`. The actual error is then never extracted.

Closes https://github.com/ClickHouse/ClickHouse/issues/101321

 Fix stress test log parser losing actual error behind "Parse failure error"

### Version info
- Merged into: `26.4.1.887`

### DIFF
--- a/ci/jobs/scripts/log_parser.py
+++ b/ci/jobs/scripts/log_parser.py
@@ -99,7 +99,10 @@ class FuzzerLogParser:
                 )
             else:
                 if flag_name == "is_sanitizer_error":
-                    assert self.stderr_log
+                    if not self.stderr_log:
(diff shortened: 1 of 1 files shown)


rienath:
This would catch most cases (logical errors, assertions etc). Sanitiser errors logged to `stderr.log` would show a signal without sanitiser-specific detail because there is no `stderr.log` in `result_path`.  @maxknv I don't know why `stderr.log` is only moved to `/test_output/`at the end of `stress_runner.sh`, but it means the log parser can't see it on early exit. If you think it makes sense, a small follow-up to move it when server fails to start too would be welcome!

maxknv (approved):

PR #102586, opened 2026-04-13
Fix OOMs on huge multi requests in keeper

mstetsyuk:

Fix OOMs on huge multi requests in the keeper. For OpenTelemetry tracing, we unconditionally allocate >1 KiB for OpenTelemetry spans in `ZooKeeperRequest` objects - meaning, for really huge multi requests, we try to allocate >10 GiB extra memory. To fix this, we now keep shared data in static memory and use `std::unique_ptr` over `std::optional` in `ZooKeeperOpentelemetrySpans`.

### Version info
- Merged into: `26.4.1.951`

### DIFF
--- a/src/Common/ZooKeeper/ZooKeeperCommon.cpp
+++ b/src/Common/ZooKeeper/ZooKeeperCommon.cpp
@@ -1600,7 +1600,7 @@ ZooKeeperRequestPtr ZooKeeperRequestFactory::get(OpNum op_num) const
         throw Exception(Error::ZBADARGUMENTS, "Unknown operation type {}", op_num);
 
     auto request = it->second();
-    request->spans = std::make_unique<DB::ZooKeeperOpentelemetrySpans>();
+    request->spans = std::make_shared<DB::ZooKeeperOpentelemetrySpans>();
     return request;
 }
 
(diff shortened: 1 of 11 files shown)


src/Coordination/KeeperStorage.cpp
```diff
@@ -4067,12 +4067,12 @@ KeeperResponsesForSessions KeeperStorage<Container>::processLocalRequests(
                 const auto maybe_log_opentelemetry_span = [&](OpenTelemetry::SpanStatus status, const std::string & error_message)
                 {
                     ZooKeeperOpentelemetrySpans::maybeInitialize(
-                        concrete_zk_request.spans.read_process,
-                        concrete_zk_request.tracing_context,
+                        concrete_zk_request.spans->read_process,
```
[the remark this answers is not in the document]
mstetsyuk: false positive

src/Common/ZooKeeper/ZooKeeperCommon.cpp
```diff
@@ -1170,6 +1170,7 @@ void ZooKeeperMultiRequest::readImpl(ReadBuffer & in, RequestValidator request_v
         request->readImpl(in);
         if (request_validator)
             request_validator(*request);
+        request->spans.reset();
```
antonio2368: You allocate then free which doesn't sounds optimal. Maybe just create spans manually where needed?

src/Coordination/KeeperStorage.cpp
antonio2368: spans can be nullptr

antonio2368 (approved):

PR #102587, opened 2026-04-13
Add memory Metric for Projections PK keys

npakeer:

- Improvement

Metrics to Track memory used by Projections Primary Keys and Projections Index Granularity for all tables.

### Version info
- Merged into: `26.5.1.19`


tavplubix:
btw do we already count marks and index granularity from projection parts?

alexey-milovidov (changes requested):
Do not merge before fixing "Logical error: '!res.empty()' (STID: 2508-3ea0)"

npakeer:
> btw do we already count marks and index granularity from projection parts?

@tavplubix  From what I understood, MarkCache memory usage metric is current metric, and is automatically incremented when marks file for any part(including projection parts) is loaded, because the increment of the loaded bytes happens with in the Cache Class during the load.

IndexGranularity, and PK metrics are asynchronous metrics and they are calculated separately for each interval i.e "asynchronous_metrics_update_period_s "  and these metrics for projections were missed. This PR calculates PK and index granularity for projections in separate metrics.

tests/integration/test_asynchronous_metrics_pk_bytes_fields/test.py
```diff
@@ -132,3 +132,123 @@ def res_pk_bytes_allocated():
...
+        return int(node.query(query_proj_pk_bytes_allocated).strip())
+
+    def res_proj_ig_bytes():
+        return int(node.query(query_proj_ig_bytes).strip())
+
+    def res_proj_ig_bytes_allocated():
+        return int(node.query(query_proj_ig_bytes_allocated).strip())
+
+    # metrics should increase after first insert
+    proj_pk_bytes_before, proj_pk_bytes_after = query_until_condition(
```
azat, line 181: Instead of this we can use `SYSTEM RELOAD ASYNCHRONOUS METRICS`

azat (approved):

azat:
Also make sense to add `index_granularity_bytes_in_memory` and `index_granularity_bytes_in_memory_allocated` to `system.projection_parts`

azat:
>Do not merge before fixing "Logical error: '!res.empty()' (STID: 2508-3ea0)"

Has been fixed already, likely https://github.com/ClickHouse/ClickHouse/pull/102604

npakeer:
> Also make sense to add `index_granularity_bytes_in_memory` and `index_granularity_bytes_in_memory_allocated` to `system.projection_parts`

@azat  Thanks for the review. I will work on adding these

PR #102590, opened 2026-04-13
S3queue small refactoring for partitioning mode

kssenii:


  Refactor ObjectStorageQueueOrderedFileMetadata to clean up the processed-path model.
                                                                                                                                                                                     
  Previously processed_node_path (base class member) held the bucket-level processed pointer path, while a separate processed_watch_path member (derived class) appended the
  partition key for HIVE/REGEX modes. This split was redundant and caused several places to manually reconstruct the partition child path.                                           
                                                                        
  Changes:                                                                                                                                                                           
                                                                        
  - processed_node_path is now initialised to the full processed pointer path for the file — including the partition key suffix in HIVE/REGEX modes — by splitting the old           
  getProcessedPath helper into getProcessedBucketPath (bucket-level, no partition suffix) and a new getProcessedPath (appends partition key when applicable).
  - A new processed_bucket_path member stores the bucket-level node path. It is used for version-pinning in setProcessingImpl (preserving the original behaviour for upgrade         
  compatibility) and for writes in doPrepareProcessedRequests.                                                                                                                       
  - processed_watch_path member and its getProcessedWatchPath override are removed; getProcessedNodePath is added to the base class as a non-virtual method alongside
  getFailedNodePath.                                                                                                                                                                 
  - Coordination::Stat * out-parameter removed from both overloads of getProcessingStateFromKeeper; the bucket-level node version is now stored as std::optional<int32_t> 
  processed_bucket_version in ProcessingStateFromKeeper.                                                                                                                             
  - getMaxProcessedFilesByHivePartition renamed to getMaxProcessedFilesByPartition (it handles both HIVE and REGEX modes).
  - getPartitionKey public static method removed; the free function getPartitionKey in the anonymous namespace is used directly.                                                     
  - The range [check_max_processed_path_begin, check_max_processed_path_end) is now used in setProcessingImpl to correctly identify which request failed when                        
  addCheckNotExistsRequest expands to two ops (CREATE + REMOVE) on deployments without native CHECK_NOT_EXISTS.                                                                      
  - A chassert is added to verify that processed_bucket_path == processed_node_path when partitioning is disabled.                                                                   
                                                                                                                                                                                     
  No Keeper node paths or data formats are changed; existing deployments can be upgraded without touching ZooKeeper state.     

<!---
Directly edit documentation source files in the "docs" folder with the same pull-request as code changes

or

Add a user-readable short description of the changes that should be added to docs.clickhouse.com below.

At a minimum, the following information should be added (but add more as needed).
- Motivation: Why is this function, table engine, etc. useful to ClickHouse users?

- Parameters: If the feature being added takes arguments, options or is influenced by settings, please list them below with a brief explanation.

- Example use: A query or command.
-->

<!-- ch-version-info:start -->
### Version info
- Merged into: `26.4.1.987`
<!-- ch-version-info:end -->


bharatnc (approved):
Looks good to me. The PR actually fixes two bugs if I am not wrong:

- preparePartitionProcessedMap was using the wrong path for the partition key 
- The addCheckNotExistsRequest range check was off-by-one

Both are fixed by this PR. Thank you!

PR #102591, opened 2026-04-13
Add regression test for parameterized view crash via remote()/cluster()

groeneai:

None.

### What is the problem?

Parameterized views (e.g., `CREATE VIEW v AS SELECT ... WHERE x = {p:UInt32}`) have empty column metadata. Accessing them through `cluster()` or `remote()` table functions used to crash the server with `LOGICAL_ERROR '!res.empty()'` in debug/sanitizer builds.

The source fix was merged in PR #102604 by @alexey-milovidov. This PR adds a regression test to prevent the bug from recurring.

### What is the fix?

This PR adds regression test `04098_remote_parameterized_view_no_crash` covering SELECT and INSERT via both `cluster()` and `remote()` against a parameterized view. The test verifies a proper `NO_REMOTE_SHARD_AVAILABLE` error is returned instead of a server crash.

**Changes from previous version (reviewer feedback):**
- Removed explicit `CREATE DATABASE` — uses the test's own database with `currentDatabase()` (per @azat)
- Removed `no-parallel` tag — test is isolated to its own database
- Dropped source code changes — superseded by PR #102604
- Resolved merge conflicts by rebasing onto current master (per @nikitamikhaylov)

### How was it tested?

- Verified locally that the test passes 50/50 with randomization on a debug build that includes PR #102604's fix
- Test correctly produces `NO_REMOTE_SHARD_AVAILABLE` for all 4 query variants

### Version info
- Merged into: `26.4.1.1058`

### DIFF
--- a/tests/queries/0_stateless/04098_remote_parameterized_view_no_crash.sql
+++ b/tests/queries/0_stateless/04098_remote_parameterized_view_no_crash.sql
@@ -1,0 +1,15 @@
+-- Regression test: INSERT/SELECT via cluster()/remote() into a parameterized view
+-- should return a proper error, not crash with LOGICAL_ERROR '!res.empty()'.
+
+CREATE VIEW pv AS SELECT 1 AS x WHERE x = {p:UInt32};
+
+-- These should produce errors (parameterized view has no columns), not server crashes.
+-- The empty column list from the local shard causes getStructureOfRemoteTable
+-- to fall through to the remote shard path.
+SELECT * FROM cluster('test_shard_localhost', currentDatabase(), 'pv'); -- { serverError NO_REMOTE_SHARD_AVAILABLE }
+SELECT * FROM remote('127.0.0.1:9000', currentDatabase(), 'pv'); -- { serverError NO_REMOTE_SHARD_AVAILABLE }
+
+INSERT INTO FUNCTION cluster('test_shard_localhost', currentDatabase(), 'pv') SELECT 1; -- { serverError NO_REMOTE_SHARD_AVAILABLE }
+INSERT INTO FUNCTION remote('127.0.0.1:9000', currentDatabase(), 'pv') SELECT 1; -- { serverError NO_REMOTE_SHARD_AVAILABLE }
+
+DROP VIEW pv;


groeneai:
### Pre-PR Validation Gate (session: cron:clickhouse-ci-task-worker:20260413-144500)

```sql
CREATE VIEW v AS SELECT 1 AS x WHERE x = {p:UInt32};
SELECT * FROM cluster('test_shard_localhost', currentDatabase(), 'v');
```
Crashes 100% of the time on debug builds with `Logical error: '!res.empty()'` at `getStructureOfRemoteTable.cpp:159`.

**b) Root cause explained?** Yes. Parameterized views (`CREATE VIEW ... WHERE x = {param:Type}`) intentionally skip `setColumns()` during construction in `StorageView.cpp:138-142` because column types depend on parameter values. This leaves `StorageInMemoryMetadata::columns` empty. When such a view is accessed via `cluster()`/`remote()` table functions, the local shard path in `getStructureOfRemoteTable()` calls `getInMemoryMetadataPtr()->getColumns()` on the view, gets empty `ColumnsDescription`, and the `chassert(!res.empty())` fires.

**c) Fix matches root cause?** Yes. The `chassert` assumed local tables always have columns — this is violated by parameterized views. The fix replaces the crash-inducing assertion with graceful handling: skip the empty result, add an informative message, and fall through to produce a proper user-facing error (`NO_REMOTE_SHARD_AVAILABLE`). This is consistent with how the second loop (remote shards) already handles empty results.

**d) Test intent preserved / New tests added?** Yes. Added regression test `04098_remote_parameterized_view_no_crash` that creates a parameterized view and verifies SELECT/INSERT via both `cluster()` and `remote()` produce proper errors (not crashes).

- Without fix: server crashes with SIGABRT (`Logical error: '!res.empty()'`)
- With fix: proper `NO_REMOTE_SHARD_AVAILABLE` error, server stays alive

**f) Fix is general?** Yes. The fix handles ANY case where a local shard returns empty columns (not just parameterized views). The comment documents parameterized views as the known trigger, but the `if (res.empty())` check is general.

groeneai:
cc @azat @vitlibar — could you review this? It fixes a server crash (SIGABRT in debug/sanitizer builds) when a parameterized view is accessed via remote()/cluster() table functions. The chassert in getStructureOfRemoteTable() assumed local tables always have columns, but parameterized views intentionally have empty metadata.

nikitamikhaylov:
@groeneai Resolve conflicts.

tests/queries/0_stateless/04098_remote_parameterized_view_no_crash.sql
```diff
@@ -0,0 +1,19 @@
+-- Tags: no-parallel
+-- Regression test: INSERT/SELECT via cluster()/remote() into a parameterized view
+-- should return a proper error, not crash with LOGICAL_ERROR '!res.empty()'.
+
+DROP DATABASE IF EXISTS test_04098;
+CREATE DATABASE test_04098;
```
azat: Please, do not create databases, each test has it's own database, if the database is required here for some reason (though I don't see it) you can fallback to `CLICKHOUSE_DATABASE_1` parameter

src/Storages/getStructureOfRemoteTable.cpp
```diff
@@ -156,7 +156,18 @@ ColumnsDescription getStructureOfRemoteTable(
         if (shard_info.isLocal())
         {
             const auto & res = getStructureOfRemoteTableInShard(cluster, shard_info, table_id, context, table_func_ptr);
-            chassert(!res.empty());
+
+            /// Parameterized views have empty column metadata because the actual
+            /// column types depend on parameter values provided at query time.
+            /// Don't crash — fall through to the remote shard path which will
+            /// produce a proper user-facing error.
+            if (res.empty())
```
azat, line 162: I guess it is OK for now, since we cannot skip it if we have parametrized MV, since we don't know it, and extending `DESC TABLE` does not worth it, there is always `SELECT * FROM cluster('test_shard_localhost', view(select * from pv(p=1)))`

Once I will like the test, I will approve it

azat:
@groeneai also it will be nice if you will provide either a link to CI report or at least stacktrace in the issue (you can uploaded the stacktraces to pastila.nl)

groeneai:
@azat @nikitamikhaylov — addressed both review comments:

1. **Removed explicit `CREATE DATABASE`** — the test now uses its own default database with `currentDatabase()`, per @azat's feedback. Also removed the `no-parallel` tag since the test is fully isolated.

2. **Resolved merge conflicts** — rebased onto current master, per @nikitamikhaylov. The source code changes are now dropped entirely since PR #102604 already fixed `getStructureOfRemoteTable()` with the `break` approach. This PR is now test-only (regression test).

Verified locally: 50/50 passes with randomization on a debug build that includes PR #102604's fix.

Session: `cron:clickhouse-ci-task-worker:20260415-171500`

azat:
Already fixed in https://github.com/ClickHouse/ClickHouse/pull/102604, but it worth to add a test

azat (approved):

PR #102597, opened 2026-04-13
Fix fuzzer grouping by use nulls error

Fgrtue:

...

Resolves https://github.com/ClickHouse/ClickHouse/issues/102273.

The investigation shows that the error happens when we use setting `group_by_use_nulls = 1`, and correlated subqeuries, then we wrap the types of grouped elements into Nullable, while correlated subquery does not account for such a change. 

That is why the fix is: after we built the plan for **outer** query, and start building correlated subquery, provide the header of columns to the correlated subquery, so that when visiting correlated column we fix its type if it diverges from the one in the outer query.


Fgrtue:
The error stopped causing the failure of the server. and now just silently lives in the query plan. I adjusted the test so that it is seen what should be the result

tests/queries/0_stateless/04092_correlated_subquery_grouping_sets_nullable.sql
```diff
@@ -0,0 +1,24 @@
...
+SET allow_experimental_correlated_subqueries = 1;
+SET group_by_use_nulls = 1;
+
+SELECT replaceRegexpOne(trim(explain), ' : [0-9]+$', '') AS planner_action
+FROM (
+    EXPLAIN actions = 1, compact = 0, pretty = 0
+    SELECT bitNot((SELECT bitNot(number)))
+    FROM numbers(1)
+    GROUP BY number WITH ROLLUP
+)
```
novikd, line 23: It should be fixed in the `QueryAnalyzer` not in the Planner. The analyzed Query Tree is invalid: https://fiddle.clickhouse.com/0782ad74-9b45-4eff-a6e6-8aeb0c7ee2e5

novikd (dismissed):

alexey-milovidov:
I verified the current state of this on `master` (checked at `e366285fd4d`, 2026-07-18):

- The underlying bug was fixed in the `QueryAnalyzer` by #100365 (merged 2026-07-12) — the same approach @novikd asked for in the review thread here. That fix makes the analyzer apply `nullable_group_by_keys` from the outer scopes to correlated column references, so the correct `Nullable` types are present in the query tree before planning.
- The exact fuzzer query from https://github.com/ClickHouse/ClickHouse/issues/102273 runs cleanly on current `master`.
- The test file `04092_correlated_subquery_grouping_sets_nullable` from this PR passes on current `master` **without any of this PR's code changes** — the planner already records `bitNot(__table3.number) -> Nullable(UInt64)`, and the executed query returns correct results (`0`, `NULL` for the `ROLLUP` row).

Regarding the AI review's concern about `NULL` join keys losing rows during decorrelation: on current `master` I could not construct a wrong result. Supported correlated shapes (e.g. `(SELECT bitNot(number))`) return correct values for the `NULL` grouping rows, and shapes that could distinguish a join miss from a real evaluation (e.g. `(SELECT coalesce(number, 42))`, `(SELECT isNull(number))`) are rejected with a graceful `Cannot evaluate correlated scalar subquery` error rather than returning wrong data.

Since the `GROUPING SETS` shape from the original fuzzer query was not covered by the tests of #100365 (test 04053 covers `WITH ROLLUP`/`WITH CUBE`), I added a regression test for it in #110956, which also closes the original issue.

So this PR appears to be fully superseded by #100365 and can be closed. Thank you for the investigation here, @Fgrtue — it correctly identified the type divergence; the analyzer-side fix ultimately landed via the other PR.

PR #102604, opened 2026-04-13
Fix exception in getStructureOfRemoteTable when local shard returns empty columns

alexey-milovidov:
The `chassert(!res.empty())` assertion in `getStructureOfRemoteTable` can fire in debug builds when `getStructureOfRemoteTableInShard` returns empty columns for a local shard. This can happen due to a race with concurrent DDL operations (e.g. `REPLACE TABLE` or lazy storage initialization), because the `IStorage` constructor creates metadata with empty columns by default.

Restore the original defensive behavior (from before commit f79f12408d8) that falls through to remote shards when the local shard returns empty columns.

BuzzHouse (amd_debug) failure: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102498&sha=0ce3f2c0db68a6f6eae31a5cf54636285b1a640c&name_0=PR&name_1=BuzzHouse%20%28amd_debug%29
PR: https://github.com/ClickHouse/ClickHouse/pull/102498

Fix exception in `getStructureOfRemoteTable` when local shard returns empty columns due to concurrent DDL.

### Version info
- Merged into: `26.4.1.918`

### DIFF
--- a/src/Storages/getStructureOfRemoteTable.cpp
+++ b/src/Storages/getStructureOfRemoteTable.cpp
@@ -156,7 +156,12 @@ ColumnsDescription getStructureOfRemoteTable(
         if (shard_info.isLocal())
         {
             const auto & res = getStructureOfRemoteTableInShard(cluster, shard_info, table_id, context, table_func_ptr);
-            chassert(!res.empty());
+
+            /// Columns may be empty due to a race with concurrent DDL (e.g. REPLACE TABLE or lazy storage initialization).
+            /// In that case, fall through to try remote shards.
+            if (res.empty())
+                break;
+
             return res;
         }
     }


antonio2368 (approved):
The fix is correct, but the root cause analysis is not — this is not a race condition with concurrent DDL.

Looking at the server log from the BuzzHouse failure, `d5.t111` is a `StorageAlias`:

```sql
CREATE TABLE d5.t111 ENGINE = Alias(d3, t43)
```

The target table `d3.t43` never existed — every query against it fails with `UNKNOWN_TABLE` from the very beginning of the test run. `StorageAlias` allows this because it resolves the target lazily at query time (the constructor only rejects self-references and alias-to-alias chains, not missing targets).

The path to empty columns:

1. `getStructureOfRemoteTableInShard` finds a local shard, calls `DatabaseCatalog::getTable("d5.t111")` — succeeds, returns the `StorageAlias`
2. Calls `storage_ptr->getInMemoryMetadataPtr()->getColumns()`
3. `StorageAlias::getInMemoryMetadataPtr` calls `tryGetTargetTable()` which returns null (`d3.t43` doesn't exist)
4. Falls back to `IStorage::getInMemoryMetadataPtr()` — returns `StorageAlias`'s own metadata, which is empty because it was constructed with `IStorage(table_id_)` (no metadata argument)
5. `getColumns()` returns empty `ColumnsDescription` → assertion fires

No concurrency needed to reproduce — a single `StorageAlias` pointing to a nonexistent table is sufficient.

The changelog entry and comment in the code should probably reflect the actual cause (`StorageAlias` with missing target) rather than "race with concurrent DDL".

clickgapai:

**Affected code:** `getStructureOfRemoteTable` in getStructureOfRemoteTable.cpp — core code present in all supported branches.

**Why:** The `chassert` was introduced in commit f79f12408d8 (July 2022), which predates all supported branches. In release builds, the race condition causes `getStructureOfRemoteTable` to return empty columns to callers like `StorageDistributed` constructor, potentially creating tables with no columns and causing query failures.

PR #102610, opened 2026-04-13
Fix `LOGICAL_ERROR` in `S3Queue` concurrent `CREATE TABLE IF NOT EXISTS`

nickitat:

Fix `LOGICAL_ERROR` exception when multiple concurrent `CREATE TABLE IF NOT EXISTS` queries target the same `S3Queue` table on a Shared database.

---

The test will be submitted in private [here](https://github.com/ClickHouse/clickhouse-private/pull/55526). It also contains a detailed explanation. 

### Version info
- Merged into: `26.4.1.973`
- Backported to: `26.3.10.11`, `26.2.17.12`, `26.1.12.6`


src/Storages/StreamingStorageRegistry.cpp
```diff
@@ -46,7 +47,12 @@ void StreamingStorageRegistry::registerTable(const StorageID & storage)
 
     const bool inserted = storages.emplace(storage).second;
     if (!inserted)
+    {
+        if (if_not_exists)
+            throw Exception(ErrorCodes::TABLE_ALREADY_EXISTS, "Table with storage id {} already registered", storage.getNameForLogs());
```
kssenii: When I've used `if_not_exists` - it always had behaviour "do nothing if already exists", while this change is more like "throw if exists, but not a logical error", so it is in general confusing, I'd rather change `LOGICAL_ERROR` to just `TABLE_ALREADY_EXISTS` instead of ambiguity with `if_not_exists`

kssenii (approved):

PR #1731, opened 2018-01-05
NonJoinedBlockInputStream generates duplicate records across block.

amosbird:
This patch fixes the iterator incrementation.

PR #102628, opened 2026-04-13
Fix LOGICAL_ERROR crash in Parquet reader for nullable columns with filter

groeneai:
## Linked issues
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1731
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1508
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1509

Fix LOGICAL_ERROR crash "Unexpected number of rows in column subchunk" in native Parquet V3 reader when reading nullable columns with a WHERE filter.

### What this PR fixes

Two bugs in the native Parquet V3 reader (`src/Processors/Formats/Impl/Parquet/Reader.cpp`):

**1. Inverted null_map check (primary fix, line 1422):**

The `memchr` check searched for byte `0` (non-null marker) instead of byte `1` (null marker) in the null_map. The null_map uses standard ClickHouse convention: `1 = NULL, 0 = NOT NULL`, as [documented in the same file](https://github.com/ClickHouse/ClickHouse/blob/master/src/Processors/Formats/Impl/Parquet/Reader.cpp#L1811-L1814):
```
/// Non-null element: null_map->push_back(0).
/// Null element: null_map->push_back(1).
```

When ALL values in a filtered subchunk were NULL and the output column was non-nullable with `input_format_null_as_default=0`:
- The check incorrectly cleared the null_map (no byte `0` found = no non-null entries)
- The column stayed at 0 rows while `rows_pass` was 1
- Triggered: `Logical error: 'Unexpected number of rows in column subchunk 0 1'`

The inverted check also caused the opposite problem: reading **non-null** data from a Parquet nullable column into a non-nullable ClickHouse column with `null_as_default=0` would incorrectly throw `CANNOT_INSERT_NULL_IN_ORDINARY_COLUMN` even when no NULLs were present.

**2. Disabled `use_filter_in_decoder` for nullable columns:**

The filter-in-decoder optimization processes ALL rows (passing + non-passing) through `processDefLevelsForInnermostColumn`, so the null_map ends up with entries for all rows rather than just filtered rows. Additionally, the decoder applies the filter at consecutive encoded-value indices, but with nullable columns, null rows have no encoded values, so the filter-to-row index mapping is wrong. Disabled this optimization for nullable columns — the standard row-range iteration path correctly handles both issues.

### How it was tested

- Regression test `04099_parquet_nullable_null_check_fix.sh`: creates a Parquet file with nullable string column (NULLs every 3rd row), verifies:
  1. `null_as_default=1` with non-nullable output → NULLs become empty strings ✓
  2. Nullable output → NULLs preserved ✓
  3. Non-nullable output, `null_as_default=0`, filter hits NULL rows → `CANNOT_INSERT_NULL_IN_ORDINARY_COLUMN` (not LOGICAL_ERROR crash) ✓
  4. Non-nullable output, `null_as_default=0`, filter hits only non-null rows → succeeds ✓
- 50/50 passes with `clickhouse-local`
- Confirmed the crash (Aborted/core dump) reproduces without the fix and is resolved with it

Closes https://github.com/ClickHouse/ClickHouse/issues/99019

### Version info
- Merged into: `26.4.1.1084` (included in `26.4` and later)
- Backported to: `26.3.17.51`

### DIFF
--- a/src/Processors/Formats/Impl/Parquet/Reader.cpp
+++ b/src/Processors/Formats/Impl/Parquet/Reader.cpp
@@ -1336,11 +1336,20 @@ void Reader::decodePrimitiveColumn(ColumnChunk & column, const PrimitiveColumnIn
     /// use_filter_in_decoder path reads ALL pages sequentially, so it would crash trying to access
     /// those reset handles. Only use this optimization when reading the whole column chunk
     /// sequentially (no offset index, i.e. data_pages is empty).
+    ///
+    /// Also disabled for nullable columns (need_null_map): the filter-in-decoder path processes
+    /// ALL rows (passing + non-passing) through processDefLevelsForInnermostColumn, so the
+    /// null_map ends up with entries for all rows rather than just filtered rows. Additionally,
+    /// the decoder applies the filter at consecutive encoded-value indices, but with nulls the
+    /// encoded values don't correspond 1:1 to rows (null rows have no encoded values), causing
+    /// the filter to be applied at wrong positions. The standard row-range iteration path
+    /// correctly handles both issues by only reading rows in passing filter ranges.
     const bool use_filter_in_decoder = (column_info.levels.back().rep == 0) &&
         !row_subgroup.filter.filter.empty() &&
         column.page.initialized &&
         !column.page.is_dictionary_encoded &&
-        column.data_pages.empty();
+        column.data_pages.empty() &&
+        !column.need_null_map;
     const size_t subgroup_end_row_idx = row_subgroup.start_row_idx + row_subgroup.filter.rows_total;
 
     if (use_filter_in_decoder)
@@ -1419,7 +1428,9 @@ void Reader::decodePrimitiveColumn(ColumnChunk & column, const PrimitiveColumnIn
     if (subchunk.null_map && !column_info.output_nullable && !options.format.null_as_default)
     {
         const auto & null_map = assert_cast<const ColumnUInt8 &>(*subchunk.null_map).getData();
-        if (memchr(null_map.data(), 0, null_map.size()) != nullptr)
+        /// null_map uses standard ClickHouse convention: 1 = NULL, 0 = NOT NULL.
+        /// Check if any values are null — those can't be inserted into a non-Nullable column.
+        if (memchr(null_map.data(), 1, null_map.size()) != nullptr)
             throw Exception(ErrorCodes::CANNOT_INSERT_NULL_IN_ORDINARY_COLUMN, "Cannot convert NULL value to non-Nullable type for column {}", column_info.name);
         subchunk.null_map = nullptr;
     }
--- a/tests/queries/0_stateless/04099_parquet_nullable_null_check_fix.reference
+++ b/tests/queries/0_stateless/04099_parquet_nullable_null_check_fix.reference
@@ -8,7 +8,7 @@
 6	\N
 --- non-nullable null_as_default=0 should error ---
 CANNOT_INSERT_NULL_IN_ORDINARY_COLUMN
---- non-nullable null_as_default=0 non-null rows ---
+--- non-nullable null_as_default=0 no nulls in file ---
 1	1
 2	2
 4	4
--- a/tests/queries/0_stateless/04099_parquet_nullable_null_check_fix.reference
+++ b/tests/queries/0_stateless/04099_parquet_nullable_null_check_fix.reference
@@ -1,0 +1,14 @@
+--- null_as_default=1 non-nullable ---
+0	
+3	
+6	
+--- nullable output ---
+0	\N
+3	\N
+6	\N
+--- non-nullable null_as_default=0 should error ---
+CANNOT_INSERT_NULL_IN_ORDINARY_COLUMN
+--- non-nullable null_as_default=0 non-null rows ---
+1	1
+2	2
+4	4
--- a/tests/queries/0_stateless/04099_parquet_nullable_null_check_fix.sh
+++ b/tests/queries/0_stateless/04099_parquet_nullable_null_check_fix.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # Tags: no-fasttest
 
-# Regression test for a LOGICAL_ERROR crash in the Parquet V3 reader:
+# Regression test for a LOGICAL_ERROR crash in the Parquet V3 native reader:
 #   "Unexpected number of rows in column subchunk 0 1"
 #
 # The bug was an inverted memchr check in Reader.cpp that searched for byte 0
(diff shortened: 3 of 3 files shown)


groeneai:
### Pre-PR Validation Gate (session: cron:clickhouse-ci-task-worker:20260413-221500)

**a) Deterministic repro?** Yes:
```
clickhouse-local -q "SELECT number AS id, if(number%3=0, NULL, toString(number)) AS val FROM numbers(20) INTO OUTFILE '/tmp/test.parquet' FORMAT Parquet SETTINGS output_format_parquet_use_custom_encoder=0"
clickhouse-local -q "SELECT id, val FROM file('/tmp/test.parquet', Parquet, 'id UInt64, val String') WHERE id = 0 SETTINGS input_format_null_as_default = 0"
```
Crashes 100% of the time without the fix (Aborted/core dump). With the fix: correctly throws `CANNOT_INSERT_NULL_IN_ORDINARY_COLUMN`.

**b) Root cause explained?** Yes. The `memchr` check at Reader.cpp:1422 searched for byte `0` (non-null) instead of byte `1` (null) in the null_map. When all filtered values were NULL (null_map all `1`s), `memchr` for `0` found nothing → didn't throw → cleared null_map → no `expand()` → column stayed at 0 rows → `0 != rows_pass` → LOGICAL_ERROR.

**c) Fix matches root cause?** Yes. Changed `memchr(..., 0, ...)` to `memchr(..., 1, ...)` to correctly detect null entries. Also disabled `use_filter_in_decoder` for nullable columns to prevent a related null_map/filter index mismatch.

**d) Test intent preserved / new test added?** Yes. New regression test `04099_parquet_nullable_null_check_fix` covers 4 scenarios: null_as_default on/off × nullable/non-nullable output.

**e) Demonstrated in both directions?** Yes. Without fix: `Aborted (core dumped)` on NULL rows AND incorrect `CANNOT_INSERT_NULL` on non-null rows. With fix: proper error on NULL rows, success on non-null rows.

**f) Fix is general?** Yes. The memchr fix covers all nullable Parquet columns (not just the specific stress test case). The `use_filter_in_decoder` fix prevents a class of issues where the filter optimization doesn't account for null_map/encoded-value index misalignment.

groeneai:
cc @al13n321 @Avogar — could you review this? It fixes an inverted `memchr` null_map check in the Parquet V3 reader that caused a LOGICAL_ERROR crash ("Unexpected number of rows in column subchunk 0 1") when reading nullable columns with a WHERE filter into non-nullable output with `null_as_default=0`. Also disables the filter-in-decoder optimization for nullable columns to prevent a related null_map/filter index mismatch.

groeneai:
### Test fix pushed (commit b5170b5dd4f)

The regression test `04099_parquet_nullable_null_check_fix` had a ~10-20% failure rate in CI (56 failures across 11 check types). Root causes identified and fixed:

**Issue 1: Legacy reader path.** CI randomizes `input_format_parquet_use_native_reader_v3`. When set to `0`, the legacy Arrow-based reader is used, which reads ALL rows before casting — so `null_as_default=0` errors even when the WHERE clause excludes NULL rows. Fixed by pinning `input_format_parquet_use_native_reader_v3 = 1` in all queries, since this test specifically exercises the V3 reader fix.

**Issue 2: Duplicate grep match.** CI adds `--send_logs_level=warning` to the client, causing `CANNOT_INSERT_NULL_IN_ORDINARY_COLUMN` to appear in both a server log message and the exception. The `grep -o` then produced 2 lines instead of 1. Fixed with `| head -1`.

**Issue 3: Non-deterministic section 4.** The original test tried to read only non-null rows (`WHERE id IN (1, 2, 4)`) from a file containing NULLs with `null_as_default=0`. This fails intermittently because the V3 reader processes data at subchunk granularity — NULLs in the same subchunk cause errors regardless of the WHERE filter. Replaced with reading from a NULL-free file, which reliably validates the fix doesn't break the non-null case.

**Verified:** 50/50 passes with full randomization locally.

Session: cron:clickhouse-ci-task-worker:20260415-161500

Avogar (approved):

Avogar:
[Unit tests (asan_ubsan)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102628&sha=b5170b5dd4ff638e94ef65abd9c9e17f932e2ec7&name_0=PR&name_1=Unit+tests+%28asan_ubsan%29) - https://github.com/ClickHouse/ClickHouse/pull/102890

clickgapai:
Hi @groeneai @Avogar — the changelog category for this PR might need a look.

**Current:** `Bug Fix (user-visible misbehavior in an official stable release)`

**Why:** The PR fixes a LOGICAL_ERROR exception (which causes abort/core dump in debug builds). LOGICAL_ERROR is explicitly listed under 'Critical Bug Fix (crash, data loss, RBAC) or LOGICAL_ERROR' category, not the general 'Bug Fix' category. The linked issue #99019 confirms this reproduces on ClickHouse 26.2.4.23 (a supported stable release).

clickgapai:

**Affected code:** `decodePrimitiveColumn` in `Reader.cpp` — `memchr` null check and `use_filter_in_decoder` condition — introduced in 25.8.

**Why:** The `memchr` null detection bug exists since the V3 reader was introduced in 25.8 (setting added as `false, false` in SettingsChangesHistory for 25.8). It was enabled by default in 25.11 (`false, true`). The `use_filter_in_decoder` bug affects 26.2+. Both cause LOGICAL_ERROR exceptions. The linked issue #99019 confirms the bug reproduces on 26.2.4.23. All supported branches (25.8, 26.1, 26.2, 26.3) contain the V3 reader code; the memchr fix applies to all, and the filter-in-decoder fix applies to 26.2+.

jrogov:
@Avogar I might be misunderstanding how backporting process works in Clickhouse, but shouldn't this also be backported to v26.3?

groeneai:
The backport targets come from the \`vXX.Y-must-backport\` labels, which were added automatically based on the linked core-incidents reports (v25.12 from #1509, v26.2 from #1731), not chosen manually. This fix was already backported to 26.2.19.43. Adding v26.3 is a release-maintainer call, so it is up to Avogar / the release team to apply the label if 26.3 should get it. As the author I do not add backport labels or open manual backport PRs myself.

jrogov:
That's the thing: I can reproduce the issue on the most recent v26.3 LTS version. Looking at the code, it's expected – the necessary changes simply aren't there.

groeneai:
Thanks for confirming - you are right, the fix is absent from 26.3. Only the `v25.12-must-backport` and `v26.2-must-backport` labels were applied here (auto-added from core-incidents #1509 and #1731), so the backport bot created 25.12 and 26.2 PRs but never a 26.3 one. 26.3 is a separate LTS line that was not labeled.

Since 26.3 is confirmed affected, the fix should be backported there too. @Avogar could you add the `v26.3-must-backport` label? The automation will then cherry-pick the fix.

For verification against 26.3 HEAD, the fix is in merge commit 728336b (two parts in the native Parquet reader: the inverted null_map memchr check, and disabling use_filter_in_decoder for nullable columns).

Avogar:
Let's backport to 26.3

groeneai:
The \`v26.3-must-backport\` label is now applied, so the backport automation will cherry-pick this into 26.3 the same way it did for 25.12 and 26.2 (see #105311). Nothing further needed manually. Thanks @jrogov for confirming it reproduces there.

PR #102629, opened 2026-04-13
Keeper: fix missing last_durable_idx rollback

al13n321:

Trivial bug that could break consistency if nodes crash at just the right moment after leader change.

(Add a test? Idk, that would be a lot of code, doesn't seem worth it and the time. Shouldn't things like jepsen find bugs like this?)

### Version info
- Merged into: `26.5.1.119`

### DIFF
--- a/src/Coordination/Changelog.cpp
+++ b/src/Coordination/Changelog.cpp
@@ -2379,6 +2379,17 @@ void Changelog::writeAt(uint64_t index, const LogEntryPtr & log_entry)
     flush();
 
     {
+        /// After flush(), last_durable_idx == old max_log_id. But we are about to
+        /// truncate entries from 'index' onward and rewrite them. The new entries
+        /// are not durable until the write thread fsyncs them, so we must decrease
+        /// last_durable_idx to reflect that entries at 'index' and beyond are no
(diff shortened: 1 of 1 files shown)


antonio2368 (approved):

alexey-milovidov (dismissed):
Do not merge before fixing "Logical error: '!res.empty()' (STID: 2508-3ea0)"

antonio2368:
> Shouldn't things like jepsen find bugs like this?

It should but it didn't. We can probably improve jepsen tests, didn't do it in a long time (for starters it runs for a really short time).

Anyways, modify the changelog to bugfix, everything else is good.

clickgapai:

**Affected code:** `Changelog::writeAt` in `src/Coordination/Changelog.cpp` — core code present in all supported branches.

**Why:** Changelog::writeAt and last_durable_idx bookkeeping predate all supported branches (the durable_idx_mutex/last_durable_idx mechanism existed before 25.8). The bug, if real, can cause NuRaft to ack durability prematurely on log truncation/leader-change paths — a correctness issue for Keeper. git log -S 'last_durable_idx' src/Coordination/Changelog.cpp shows the field has been present for multiple release cycles.

PR #101574, opened 2026-04-02
Fix flaky test_overcommit_tracker/test_user_overcommit

groeneai:
The `test_user_overcommit` integration test became massively flaky after PR #100621 added the `assert overcommitted_killed` check. CIDB shows **456 failures across 170+ distinct PRs** since March 30, primarily in `amd_llvm_coverage` but also in `arm_binary` and `amd_tsan` builds.

## Root cause

The test sends 100 concurrent queries (50 with low `memory_overcommit_ratio_denominator=1`, 50 with high `=80000000`) and expects the overcommit tracker to kill the low-ratio queries when user memory exceeds `max_memory_usage_for_user`.

With the old parameters (`numbers(2500000)` → ~20 MB/query, 2 GB user limit), **100+ queries need to overlap concurrently** to reach 2 GB. Fast builds achieve this easily, but slow CI builds (coverage, ARM, TSAN) serialize queries more, typically peaking at 10-20 concurrent — well below the threshold.

**Two failure modes observed:**
- `overcommitted_killed=False, finished=True` — memory pressure never reached (arm_binary, coverage builds)
- `overcommitted_killed=True, finished=False` — TSAN 5× memory overhead causes ALL queries killed

## Changes

| Parameter | Before | After | Rationale |
|-----------|--------|-------|-----------|
| `max_memory_usage_for_user` | 2 GB | 300 MB | Only 8 concurrent queries needed to trigger |
| `numbers()` | 2,500,000 (~20 MB) | 5,000,000 (~40 MB) | More memory per query |
| Total queries | 100 | 40 | Still sufficient, less total memory |
| Retries | 3 | 5 | Higher success rate |
| TSAN skip | ❌ | ✅ | Joins existing MSan/ASan/coverage skip |

With the new parameters, even 8 concurrent queries (8 × 40 MB = 320 MB) exceed the 300 MB limit. After killing A queries, surviving B queries fit under the limit (7 × 40 MB = 280 MB < 300 MB).

## Verification

- 20/20 local integration test runs pass
- Prior fix attempts (PR #100567 closed, PR #101270 merged but insufficient) only adjusted `max_threads` — this PR fixes the fundamental memory pressure imbalance

Fixes: #101318

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
...

[...]

PR #102640, opened 2026-04-14
Fix residual flakiness in test_overcommit_tracker

groeneai:
Follow-up to #101574 which reduced `test_overcommit_tracker/test_user_overcommit`
failures by ~93%. A ~1% residual failure remained: 2 master failures and ~6 PR
failures per week since the fix was merged on Apr 2.

**Root cause:** both query A (denominator=1, should be killed) and query B
(denominator=80000000, should survive) used `numbers(5000000)` (~40 MB each).
With 20 B queries, their collective memory (800 MB) far exceeded the 300 MB
`max_memory_usage_for_user` limit.  After `UserOvercommitTracker` killed all A
queries (highest `committed/soft_limit` ratio), B's total still exceeded the
limit — so B queries were also killed, failing the `finished` assertion.

**Fix:** reduce B query data from `numbers(5000000)` to `numbers(1000000)` (~8 MB
each) and `QUERY_COUNT` from 40 to 20.  After A is killed, B's collective memory
(10 × 8 MB = 80 MB) fits within the 300 MB limit.  Total demand (480 MB) still
exceeds the limit to trigger the overcommit tracker.

The test still verifies the core invariant: queries with low
`memory_overcommit_ratio_denominator` are killed first while queries with high
denominator survive.

Validated with 50 consecutive local passes.

### Version info
- Merged into: `26.4.1.977`

### DIFF
--- a/tests/integration/test_overcommit_tracker/test.py
+++ b/tests/integration/test_overcommit_tracker/test.py
@@ -21,14 +21,23 @@ def start_cluster():
         cluster.shutdown()
 
 
-# Use larger queries (128 MB each) and a lower user memory limit (300 MB) so that
-# memory pressure is reached even when only a handful of queries overlap in time.
-# This prevents flakiness on slow builds (coverage, ARM) where fewer queries run
-# concurrently and the old 2 GB limit was rarely hit.
+# Query A allocates ~40 MB (5M × 8 bytes) with denominator=1 → highest overcommit
+# ratio → killed first under memory pressure.
+# Query B allocates ~8 MB (1M × 8 bytes) with denominator=80000000 → lowest ratio
+# → survives.
+#
+# User memory limit is 300 MB.  With 10 × A + 10 × B running concurrently the total
+# demand (~480 MB) comfortably exceeds the limit, triggering the overcommit tracker.
+# After all A queries are killed, B's total (~80 MB) fits well within the 300 MB
+# limit, so at least one B query can finish — which is exactly what the test asserts.
+#
+# Previous parameters used numbers(5000000) for both A and B, making B's total
+# (20 × 40 MB = 800 MB) exceed the 300 MB limit even after A was killed.  This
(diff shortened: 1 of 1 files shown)


groeneai:
### Pre-PR Validation Gate

**a) Deterministic repro?**
The failure is timing-dependent (~1% rate in CI on `amd_binary` and `arm_binary` configs). Root cause is fully explained by memory math — not a randomized-setting issue.

**b) Root cause explained?**
`UserOvercommitTracker::pickQueryToExcludeImpl()` kills the query with the highest `committed / soft_limit` ratio. Query A (denominator=1) is always killed first. However, after all A queries are killed, B's collective memory (20 × 40 MB = 800 MB) still exceeds the 300 MB `max_memory_usage_for_user` limit. The tracker then starts killing B queries too. In rare timing windows on CI, ALL B queries are killed before any complete — triggering `assert finished`.

Yes. By reducing B to `numbers(1000000)` (~8 MB each), B's collective memory after A is killed is 10 × 8 MB = 80 MB — well within the 300 MB limit. The overcommit tracker has no reason to kill B queries once A cleanup completes.

**d) Test intent preserved?**
Yes. The test still verifies that queries with low `memory_overcommit_ratio_denominator` (A=1) are killed first, while queries with high denominator (B=80000000) survive. The asymmetric memory makes the expected behavior MORE predictable.

CIDB shows 2 master failures in 12 days with old parameters (`overcommitted_killed=True, finished=False`). 50/50 passes with new parameters locally.

Session: `cron:clickhouse-ci-task-worker:20260414-091500`

groeneai:
cc @azat — could you review this? It fixes residual ~1% flakiness in test_overcommit_tracker by reducing B query memory so that B's collective allocation (80 MB) fits within the 300 MB user limit after A queries are killed. Previously B alone used 800 MB, exceeding the limit and causing B to also be killed.

nikitamikhaylov (approved):

PR #102648, opened 2026-04-14
Fix NOT_FOUND_COLUMN_IN_BLOCK with row policies in SourceStepWithFilter

yariks5s:

Bug with row policy that gives 'Not found column in block' exception in some special cases.

### Version info
- Merged into: `26.4.1.953`
- Backported to: `26.3.10.6`, `26.2.16.5`, `26.1.11.8`


tests/queries/0_stateless/04099_source_step_with_filter_rp.sql
```diff
@@ -0,0 +1,52 @@
+DROP TABLE IF EXISTS t_rp;
```
Ergus, line 7: Let's add an issue dewscription here please. It is an incidence; so, we cannot reference it directly, but a brief explication of the symptoms of what error the user sees can be added to remember in the future if this test fails.

Ergus (approved):
LGTM... it is the same fix I have ;) 

Let's wait for CI

KochetovNicolai (approved):

yariks5s:
The approach was changed because overriding row policies there was inconsistent. Instead, the better approach is to apply updatePrewhereInfo to the raw blocks rather than the updated ones.

@Ergus please lmk if you have objections

Ergus:
| @Ergus please lmk if you have objections

Not at all. I am still understanding all the corner cases in this part of the code ;)

On the other hand, I am a bit more concerned about why you explicitly enabled the analyzer?

yariks5s:
The test was failing with the old analyzer, I will try one more setting to make the test work

Ergus:
@nikitamikhaylov This PR fixes definitively the https://github.com/ClickHouse/support-escalation/issues/7407 we had with 25.12.1497 last week. It probably needs to be backported

cwurm:
Fyi, Codex found a repro that doesn't use parallel replicas (also fixed by this PR) - https://fiddle.clickhouse.com/56f62a71-7e1e-42a4-80d5-a58ef505d2db:
```sql
CREATE TABLE t
(
    a Int32,
    b Int32,
    c Int32,
    d String
)
ENGINE = MergeTree
ORDER BY (d, a)
PRIMARY KEY (d, a);

INSERT INTO t VALUES
    (1, 10, 11, 'fault'),
    (2, 20, 0,  'fault');

CREATE ROW POLICY t_rp
ON t
USING d = substring(currentUser(), 3)
TO CURRENT_USER;

CREATE VIEW v AS
SELECT 101 AS x, 1 AS y;

SELECT x, b
FROM v AS r
INNER JOIN
(
    SELECT a AS k, b
    FROM t
    WHERE c = 11
) AS l
    ON r.y = l.k
SETTINGS optimize_move_to_prewhere = 1
FORMAT JSONEachRow;
```

Leads to:
```
Received exception from server (version 26.4.1):
Code: 10. DB::Exception: Received from localhost:9000. DB::Exception: Not found column d: in block a Int32 Int32(size = 0), b Int32 Int32(size = 0), c Int32 Int32(size = 0). (NOT_FOUND_COLUMN_IN_BLOCK)
```

Ergus:
Hi @yariks5s 

> Fyi, Codex found a repro that doesn't use parallel replicas (also fixed by this PR) - https://fiddle.clickhouse.com/56f62a71-7e1e-42a4-80d5-a58ef505d2db:
> 
> ```sql
> CREATE TABLE t
> (
>     a Int32,
>     b Int32,
>     c Int32,
>     d String
> )
> ENGINE = MergeTree
> ORDER BY (d, a)
> PRIMARY KEY (d, a);
> 
> INSERT INTO t VALUES
>     (1, 10, 11, 'fault'),
>     (2, 20, 0,  'fault');
> 
> CREATE ROW POLICY t_rp
> ON t
> USING d = substring(currentUser(), 3)
> TO CURRENT_USER;
> 
> CREATE VIEW v AS
> SELECT 101 AS x, 1 AS y;
> 
> SELECT x, b
> FROM v AS r
> INNER JOIN
> (
>     SELECT a AS k, b
>     FROM t
>     WHERE c = 11
> ) AS l
>     ON r.y = l.k
> SETTINGS optimize_move_to_prewhere = 1
> FORMAT JSONEachRow;
> ```
> 
> Leads to:
> 
> ```
> Received exception from server (version 26.4.1):
> Code: 10. DB::Exception: Received from localhost:9000. DB::Exception: Not found column d: in block a Int32 Int32(size = 0), b Int32 Int32(size = 0), c Int32 Int32(size = 0). (NOT_FOUND_COLUMN_IN_BLOCK)
> ```

Do you want to add this to the new test `04099_source_step_with_filter_rp‎`

yariks5s:
I will, thanks for finding

archiewood:
This was a savage bug for us to track down. I'm going to put some queries here in case it helps other people end up here and our workaround.

We are using Clickhouse Cloud where this fix has not been released yet



```sql
SELECT *
FROM t_b2b_companies
WHERE partner_category = 'Corporate';
```

* the user running this query has an RLS rule on `partner_name` (another column in t_b2b_companies)
* another user that did not have the RLS rule did not get errors

Error
```
Received exception from server (version 25.12.1):
Code: 10. DB::Exception: Received from xxx.us-central1.gcp.clickhouse.cloud:9440. 
DB::Exception: Not found column partner_name: in block partner_category Nullable(String) Nullable(size = 0, String(size = 0), UInt8(size = 0)). (NOT_FOUND_COLUMN_IN_BLOCK)
```

and later when playing with another setting
Error
```
Received exception from server (version 25.12.1):
Code: 9. DB::Exception: Received from xxx.us-central1.gcp.clickhouse.cloud:9440. 
DB::Exception: Size of filter (433) doesn't match size of column (1): read stage: PrewhereData: column: partner_name: (in file/uri root/t_b2b_companies/live/data.parquet): While executing ParquetV3BlockInputFormat: While executing ReadFromObjectStorage. (SIZES_OF_COLUMNS_DOESNT_MATCH)
```

## Workaround

disable optimize_move_to_prewhere

```sql
SELECT * 
FROM t_b2b_companies
WHERE partner_category='Corporate'
SETTINGS optimize_move_to_prewhere = 0;
```

PR #102661, opened 2026-04-14
Fix crash in cluster discovery static configuration

kssenii:
## Linked issues
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1684

Fixed a server exception in ClusterDiscovery when a static cluster (defined in config) temporarily had no live nodes. 

### Version info
- Merged into: `26.4.1.1023`
- Backported to: `26.3.10.34`, `26.2.17.8`, `26.1.12.4`

### DIFF
--- a/src/Interpreters/ClusterDiscovery.cpp
+++ b/src/Interpreters/ClusterDiscovery.cpp
@@ -487,6 +487,9 @@ void ClusterDiscovery::removeCluster(const String & name, bool is_dynamic)
         std::lock_guard lock(mutex);
         cluster_impls.erase(name);
     }
+    /// For static clusters (defined in config), `clusters_to_update` and `get_nodes_callbacks`
+    /// are initialized once at startup and must persist so the cluster can be re-registered after
+    /// a ZooKeeper session loss. Dynamic clusters own their entries and must clean them up.
     if (is_dynamic)
     {
         clusters_to_update->remove(name);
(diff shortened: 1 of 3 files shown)


src/Interpreters/ClusterDiscovery.cpp
```diff
@@ -470,9 +470,20 @@ bool ClusterDiscovery::upsertCluster(ClusterInfo & cluster_info)
 
     if (nodes_info.empty())
     {
-        String name = cluster_info.name;
-        /// cluster_info removed inside removeCluster, can't use reference to name.
-        removeCluster(name);
+        if (cluster_info.zk_root_index != 0)
```
alesapin: It's very non-obvious why does it mean that cluster is "dynamic"

src/Interpreters/ClusterDiscovery.cpp
```diff
@@ -470,9 +470,20 @@ bool ClusterDiscovery::upsertCluster(ClusterInfo & cluster_info)
...
+            String name = cluster_info.name;
+            /// cluster_info removed inside removeCluster, can't use reference to name.
+            removeCluster(name);
+        }
+        else
+        {
+            /// Static cluster (defined in config) with no nodes. Don't erase watch callback
+            /// (via removeCluster) - nodes may reappear, and the callback is needed to detect that.
+            /// Just hide the cluster from system.clusters until nodes come back.
+            std::lock_guard lock(mutex);
```
alesapin: let's introduce `removeStaticCluster`?

src/Interpreters/ClusterDiscovery.cpp
kssenii: https://github.com/ClickHouse/ClickHouse/blob/7d2c8c7635a4725ed74227bdafc3d15f81c54dae/src/Interpreters/ClusterDiscovery.h#L91-L94

alesapin (approved):

PR #102669, opened 2026-04-14
Add `skip_alias_columns` config for `DefaultSystemLogFlushPolicy`

mstetsyuk:
When system log tables use an S3-backed engine (e.g. `ENGINE = S3(...)` in keeper configs), the `build_id` ALIAS column on `trace_log` causes `Code: 36. DB::Exception: Special columns like MATERIALIZED, ALIAS or EPHEMERAL are not supported for s3 storage`. The `DefaultSystemLogFlushPolicy` always returns `false` for `shouldSkipAliasColumns()`, so alias columns are always included in `CREATE TABLE`.

Add a `default_system_log_flush_policy.skip_alias_columns` server-level config option (default `false`) that is passed to `DefaultSystemLogFlushPolicy` and controls whether alias columns are omitted from system log table DDL.

Deployments using raw S3 engines for system logs (e.g. keeper-as-server) can set:
```xml
<default_system_log_flush_policy>
    <skip_alias_columns>true</skip_alias_columns>
</default_system_log_flush_policy>
```

- Improvement

Add `default_system_log_flush_policy.skip_alias_columns` config option to allow omitting ALIAS columns from system log tables, fixing S3-backed system logs that reject ALIAS columns.

### Version info
- Merged into: `26.4.1.962`


mstetsyuk:
@copilot resolve the merge conflicts in this pull request

bharatnc (approved):

PR #102299, opened 2026-04-09
Fix wrong column comparison in FinishAggregatingInOrderAlgorithm

groeneai:
Fixes #102909

### Changelog category:
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry:
Fix incorrect aggregation results (duplicate rows) when using `optimize_aggregation_in_order=1` with GROUP BY columns ordered differently from the table's sorting key.

### Description

`FinishAggregatingInOrderAlgorithm` merges partially aggregated data from parallel in-order aggregation streams. It uses a sort description to compare rows across streams and determine merge group boundaries via `std::upper_bound`.

**The bug:** In `State::State()`, `sorting_columns` was built by iterating through the sort description and using `emplace_back`, creating a compact array indexed `0..desc.size()-1`. However, the `less()` comparison function uses `elem.column_number` (the column's position in the output header) to index into `sorting_columns`. When these two orderings differ, the wrong columns are compared.

This happens whenever the GROUP BY lists a non-sort-key column before a sort-key column. For example, `GROUP BY a, b` on a table with `ORDER BY b`:
- Output header: `a` at position 0, `b` at position 1
- Sort description: `[b, a]` (sort-key column first)
- Old `sorting_columns`: `[b_col, a_col]` (indexed 0, 1)
- `less()` for `b` uses `column_number=1` → accesses `sorting_columns[1]` = `a_col` ← **WRONG**
- `less()` for `a` uses `column_number=0` → accesses `sorting_columns[0]` = `b_col` ← **WRONG**

The swapped comparison corrupts the `upper_bound` binary search (the data is sorted by `(b, a)` but compared by `(a, b)`), producing incorrect merge group boundaries. Rows with the same key end up in different merge groups, causing duplicate rows in the final result.

**The fix:** Change `sorting_columns` to be indexed by header position (`resize` to `all_columns.size()`, assign at `column_number`) instead of by description iteration order (`emplace_back`). This ensures `less()` accesses the correct column for each sort description entry.

**Reproduction:** `GROUP BY a, b` on a table `ORDER BY b` with `optimize_aggregation_in_order=1` and multiple parallel streams:
- Before fix: 28091–37081 rows instead of 20000 (40–85% inflation from duplicates)
- After fix: consistently 20000

This also fixes the flaky test `00069_duplicate_aggregation_keys` which uses `GROUP BY URL, EventDate, EventDate` on the `test.hits` table (sorted by `CounterID, EventDate, ...`).

### Version info
[...]

PR #102684, opened 2026-04-14
Fix CASE with Dynamic expression returning ELSE for all rows

### LINKED ISSUE #102511: `can_use_transform` missing Dynamic/Variant guard causes CASE with constant WHENs to return ELSE for all rows

_Retrospective finding from a historical scan of [PR #87177](https://github.com/ClickHouse/ClickHouse/pull/87177) (merged 2025-09-17). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`CASE dynamic_col WHEN 1 THEN 'one' WHEN 2 THEN 'two' ELSE 'other' END` returns `'other'` for ALL rows, including those where `dynamic_col = 1` or `dynamic_col = 2`

**Root cause:** caseWithExpression.cpp:182: `can_use_transform` condition only checks for Nullable mismatch but does not exclude Dynamic or Variant types. The `transform` function cannot correctly match values across different Dynamic subtypes because its hash-based lookup includes the type discriminator.

**Why we believe this is a bug:** caseWithExpression.cpp:182 → `can_use_transform` has no check for Dynamic type → routes to transform path → transform.cpp:807 builds hash from `Dynamic(UInt8(1))` via `updateHashWithValue` → executeAnything (line 234) hashes input `Dynamic(Int64(1))` which produces a different hash due to different type discriminator → lookup miss → returns ELSE value

**Affected locations:**
- `src/Functions/caseWithExpression.cpp:182` — can_use_transform missing Dynamic/Variant type guard
- `src/Functions/transform.cpp:807` — hash table built from Dynamic values includes type discriminator in hash

**Impact:** Silent wrong results: any `CASE dynamic_col WHEN <constant> THEN <value> ... ELSE <default> END` query always returns the ELSE branch, ignoring all WHEN matches. This is the common form of CASE (constant WHEN values) and affects all Dynamic columns.

Yes — confirmed on current `master` (commit `339085418513`).

### How to reproduce

```sql
-- Test: CASE with Dynamic expression and constant WHEN/THEN values uses transform path
-- which gives wrong results because Dynamic hash differs by stored subtype
DROP TABLE IF EXISTS t_case_dyn;
CREATE TABLE t_case_dyn (c0 Dynamic) ENGINE = Memory;
INSERT INTO t_case_dyn VALUES (1), (2), (3);

-- This goes through the transform path (all WHEN/THEN are constant)
-- Bug: transform hashes Dynamic(UInt8(1)) differently from Dynamic(Int64(1))
-- so lookups never match, returning ELSE for all rows
SELECT c0, CASE c0 WHEN 1 THEN 'one' WHEN 2 THEN 'two' ELSE 'other' END AS result
FROM t_case_dyn ORDER BY toString(c0);

DROP TABLE t_case_dyn;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/115f4beb-afa3-4672-834b-4cb8b1ef70de)

### Expected behavior

```
1	one
2	two
3	other
```

```
1	other
2	other
3	other
```

### Additional context

**Open risks:**
- Variant types with WHEN values whose type is not a variant arm throw CANNOT_CONVERT_TYPE instead of falling back to multiIf (error code 70 is not caught by NOT_IMPLEMENTED fallback)

**Suggested fix:** Add Dynamic and Variant guards to `can_use_transform`: `&& !isDynamic(removeNullable(args.front().type)) && !isVariant(removeNullable(args.front().type))`. This forces the multiIf path which correctly handles Dynamic/Variant via the equals function.

Found during automated review of [PR #87177](https://github.com/ClickHouse/ClickHouse/pull/87177).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr87177_001`_

Avogar:

Fix CASE with Dynamic expression returning ELSE for all rows. Closes https://github.com/ClickHouse/ClickHouse/issues/102511

### Version info
- Merged into: `26.4.1.1017`

### DIFF
--- a/src/Functions/caseWithExpression.cpp
+++ b/src/Functions/caseWithExpression.cpp
@@ -179,8 +179,15 @@ public:
             auto src_supertype = tryGetLeastSupertype(src_array_types);
             auto dst_supertype = tryGetLeastSupertype(dst_array_types);
 
+            /// The `transform` function cannot handle Dynamic/Variant types because its
+            /// hash-based lookup includes the type discriminator, so values with the same
+            /// logical content but different stored subtypes (e.g. Dynamic(UInt8(1)) vs
+            /// Dynamic(Int64(1))) produce different hashes and never match.
+            auto expr_type = removeNullable(args.front().type);
             bool can_use_transform = src_supertype && dst_supertype
-                && !(src_supertype->isNullable() && !args.front().type->isNullable());
+                && !(src_supertype->isNullable() && !args.front().type->isNullable())
+                && !isDynamic(expr_type)
+                && !isVariant(expr_type);
 
             if (can_use_transform)
             {
--- a/tests/queries/0_stateless/04093_case_with_expression_dynamic.reference
+++ b/tests/queries/0_stateless/04093_case_with_expression_dynamic.reference
@@ -1,0 +1,12 @@
+Dynamic CASE:
+1	one
+2	two
+3	other
+Dynamic CASE with strings:
+a	first
+b	second
+c	other
+Dynamic CASE with NULL:
+\N	null_val
+1	one
+3	other
--- a/tests/queries/0_stateless/04093_case_with_expression_dynamic.sql
+++ b/tests/queries/0_stateless/04093_case_with_expression_dynamic.sql
@@ -1,0 +1,38 @@
+-- Tags: no-fasttest
+-- Test for https://github.com/ClickHouse/ClickHouse/issues/102511
+-- CASE with Dynamic/Variant expression should not use the transform path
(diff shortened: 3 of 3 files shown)


novikd (approved):

Avogar:
[Stateless tests (arm_binary, parallel)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102684&sha=67221a3cc84047276ea7ac8cdf73cb9722f573e0&name_0=PR&name_1=Stateless%20tests%20%28arm_binary%2C%20parallel%29) - https://github.com/ClickHouse/ClickHouse/pull/102299

PR #102700, opened 2026-04-14
Fix SSH exit code

thevar1able:

Report actual exit code via SSH client instead of mapping all errors to `1`. Closes https://github.com/ClickHouse/ClickHouse/issues/101741.

### Version info
- Merged into: `26.4.1.945`

### DIFF
--- a/src/Server/SSH/SSHPtyHandler.cpp
+++ b/src/Server/SSH/SSHPtyHandler.cpp
@@ -101,7 +101,7 @@ public:
     }
 
     bool hasClientFinished() { return client_runner.has_value() && client_runner->hasFinished(); }
-    int getClientExitCode() { return client_runner.has_value() && client_runner->getExitCode(); }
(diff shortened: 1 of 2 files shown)


src/Server/SSH/SSHPtyHandler.cpp
```diff
@@ -101,7 +101,7 @@ class ChannelCallback
     }
 
     bool hasClientFinished() { return client_runner.has_value() && client_runner->hasFinished(); }
-    int getClientExitCode() { return client_runner.has_value() && client_runner->getExitCode(); }
+    int getClientExitCode() { return client_runner.has_value() ? client_runner->getExitCode() : 0; }
```
nikitamikhaylov, line 104: Was there an implicit cast to bool previously?

nikitamikhaylov (approved):

src/Server/SSH/SSHPtyHandler.cpp
thevar1able: Yes.

PR #102702, opened 2026-04-14
Fix flaky test 02974_analyzer_array_join_subcolumn

groeneai:
Add `ORDER BY 1` to all `ARRAY JOIN` queries on Map columns to ensure deterministic row ordering regardless of randomized Map serialization settings.

### Root Cause

When `map_serialization_version_for_zero_level_parts = 'with_buckets'` is combined with `max_buckets_in_map = 98`, `map_buckets_strategy = 'constant'`, and `map_buckets_min_avg_size = 1` (all randomized by the CI test runner), Map entries are serialized into hash-bucket order during zero-level part creation. On read-back, `ARRAY JOIN` iterates entries in bucket order rather than insertion order, reversing the output for keys `'a'` and `'b'`.

The test verifies that `ARRAY JOIN` on Map columns correctly accesses subcolumns (`.values.a`) through subqueries, aliases, and nested Tuples. It does not test Map iteration order. Adding `ORDER BY 1` makes the output deterministic while fully preserving the test's original intent.

### Reproduction

```sql
CREATE TABLE t2 (id Int32, pe Map(String, Tuple(a UInt64, b UInt64)))
ENGINE = MergeTree ORDER BY id
SETTINGS map_serialization_version_for_zero_level_parts = 'with_buckets',
         max_buckets_in_map = 98, map_buckets_strategy = 'constant',
         map_buckets_min_avg_size = 1;
INSERT INTO t2 VALUES (1, {'a': (1, 2), 'b': (2, 3)});
-- Without ORDER BY: ('b',...) before ('a',...) — reversed
SELECT pe, pe.values.a FROM (SELECT * FROM t2) ARRAY JOIN pe SETTINGS enable_analyzer = 1;
-- With ORDER BY 1: ('a',...) before ('b',...) — correct
SELECT pe, pe.values.a FROM (SELECT * FROM t2) ARRAY JOIN pe ORDER BY 1 SETTINGS enable_analyzer = 1;
```

### Validation

- 100/100 passes with full randomization (`--test-runs 100`)
- Deterministic failure without fix confirmed with the 4-setting combination above
- Reference file unchanged — output is identical with ORDER BY

### Changelog entry (a]one-liner human-readable description of the changes):
- (not required)

### Documentation check
- [ ] Documentation is not needed.

### Version info
- Merged into: `26.4.1.984`

### DIFF
--- a/tests/queries/0_stateless/02974_analyzer_array_join_subcolumn.sql
+++ b/tests/queries/0_stateless/02974_analyzer_array_join_subcolumn.sql
@@ -7,17 +7,17 @@ INSERT INTO t2 VALUES (1, {'a': (1, 2), 'b': (2, 3)}),;
 CREATE TABLE t3 (id Int32, c Tuple(v String, pe Map(String, Tuple(a UInt64, b UInt64)))) ENGINE = MergeTree ORDER BY id;
 INSERT INTO t3 VALUES (1, ('A', {'a':(1, 2),'b':(2, 3)}));
 
-SELECT pe, pe.values.a FROM (SELECT * FROM t2) ARRAY JOIN pe SETTINGS enable_analyzer = 1;
-SELECT p, p.values.a FROM (SELECT * FROM t2) ARRAY JOIN pe AS p SETTINGS enable_analyzer = 1;
+SELECT pe, pe.values.a FROM (SELECT * FROM t2) ARRAY JOIN pe ORDER BY 1 SETTINGS enable_analyzer = 1;
+SELECT p, p.values.a FROM (SELECT * FROM t2) ARRAY JOIN pe AS p ORDER BY 1 SETTINGS enable_analyzer = 1;
 
-SELECT pe, pe.values.a FROM t2 ARRAY JOIN pe;
-SELECT p, p.values.a FROM t2 ARRAY JOIN pe AS p;
+SELECT pe, pe.values.a FROM t2 ARRAY JOIN pe ORDER BY 1;
+SELECT p, p.values.a FROM t2 ARRAY JOIN pe AS p ORDER BY 1;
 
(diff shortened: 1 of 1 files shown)


groeneai:
### Pre-PR Validation Gate

**Session:** `cron:clickhouse-ci-task-worker:20260414-171500`

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | ✅ `CREATE TABLE ... SETTINGS map_serialization_version_for_zero_level_parts='with_buckets', max_buckets_in_map=98, map_buckets_strategy='constant', map_buckets_min_avg_size=1` + original queries (no ORDER BY) → always reversed output |
| b | **Root cause explained?** | ✅ Bucketed Map serialization distributes entries across hash buckets during zero-level part creation. On read-back, entries are iterated in bucket order (hash order), not insertion order. For keys 'a' and 'b', hash('b') < hash('a') in bucket space, so 'b' appears first in ARRAY JOIN output |
| c | **Fix matches root cause?** | ✅ `ORDER BY 1` sorts ARRAY JOIN output deterministically by Map entry tuple, independent of internal storage order |
| d | **Test intent preserved?** | ✅ Test verifies ARRAY JOIN on Map subcolumns works through subqueries, aliases, and nested Tuples. Adding ORDER BY doesn't affect this — it only stabilizes row order |
| e | **Both directions?** | ✅ Without fix + bucketed settings → reversed output. With fix + same settings → correct output. 100/100 randomized runs pass |

groeneai:
cc @Avogar — could you review this? It adds ORDER BY 1 to the ARRAY JOIN queries in test 02974 to fix a flaky failure caused by randomized Map bucketed serialization (`map_serialization_version_for_zero_level_parts='with_buckets'`) changing Map entry iteration order.

murphy-4o (approved):

PR #102703, opened 2026-04-14
Fix hex encoding of content sample in cached schema filenames

Avogar:

Fix hex encoding of content sample in cached schema filenames. Closes https://github.com/ClickHouse/ClickHouse/issues/101904

### Version info
- Merged into: `26.5.1.601`


Algunenano (approved):

clickgapai:

**Affected code:** `FormatSchemaInfo::generateSchemaFileName` — introduced in 25.8.

**Why this might apply:** The `format_schema_source` feature was introduced in commit 59297ec0554 on 2025-06-18, before 25.8 was released (2025-08-29). All supported branches contain the buggy code. The bug causes incorrect caching behavior where different schemas get the same cached filename prefix, potentially causing cache collisions.

Issue #53401, opened 2023-08-14
Using RabbitMQ Engine, CPU load is very high


hzhswyz:
**Describe the unexpected behaviour**
I created a table 'user_login_record' using RabbitMQ Engine, then created a table 'user_login_record_view' with the same fields, and then created a materialized view 'user_login_record_mv'(CREATE MATERIALIZED VIEW user_login_record_mv TO user_login_record_view AS SELECT * FROM user_login_record). After I created the materialized view the server's CPU load immediately increased from 1% to 40%.

**How to reproduce**
1: CREATE TABLE user_login_record
(
    `city_id` Int32,
    `district_code` Int32,
    `login_time` DateTime64(3)
)
ENGINE = RabbitMQ
SETTINGS rabbitmq_host_port = '*:5672', rabbitmq_vhost = '/data', rabbitmq_exchange_type = 'fanout', rabbitmq_exchange_name = 'user.login.record', rabbitmq_queue_base = 'clickhouse.user.login.record', rabbitmq_username = '*', rabbitmq_password = '*', rabbitmq_format = 'JSONEachRow', rabbitmq_flush_interval_ms = 10000

2: CREATE TABLE user_login_record_view ON CLUSTER cluster_1S_2R
(
    `city_id` Int32,
    `district_code` Int32,
    `login_time` DateTime64(3)
)
ENGINE = ReplicatedMergeTree
PRIMARY KEY (login_time, district_code)
ORDER BY (login_time, district_code)

3: CREATE MATERIALIZED VIEW user_login_record_mv TO user_login_record_view AS SELECT * FROM user_login_record

**Which ClickHouse server version to use**
ClickHouse server version 23.8.1.322 (official build).

**Sample data for all these tables, use [clickhouse-obfuscator]**
{
    "cityId": 2202,
    "districtCode": 220200,
    "loginTime": 1691733993239
}
**Queries to run that lead to unexpected result**
 CREATE MATERIALIZED VIEW user_login_record_mv TO user_login_record_view AS SELECT * FROM user_login_record
 
**Expected behavior**
CPU load should not significantly increase by 40 times

**Error message and/or stacktrace**
Turn on the debug log and you can see that the backend thread is constantly cycling to obtain data(400/s).Even if there is no data in the queue.
![image](https://github.com/ClickHouse/ClickHouse/assets/44084348/19b3d152-f913-430f-89ad-50a85ede4d1b)

**Additional context**
[...]

Issue #50692, opened 2023-06-08
High memory (and CPU) usage when using RabbitMQ table, ClickHouse v23.4.2.11


vegassor:
I have the following setup:
- RabbitMQ table
- RabbitMQ consumer
- ReplacingMergeTree table

I run a single node ClickHouse in kubernetes. I don't have any queries running.
After upgrading from 22.11.1.1360 to 23.4.2.11 I noticed that memory consumption is higher
![image](https://github.com/ClickHouse/ClickHouse/assets/45504682/480f5315-52d1-402b-8a78-cb031c64f064)

CPU is slightly higher too
![image](https://github.com/ClickHouse/ClickHouse/assets/45504682/818a744c-9063-48a6-9cc7-0acf583130db)

**How to reproduce**
```sql
CREATE TABLE default.event_queue
(
    `_trace_id` String,
    `_publish_time` DateTime64(3),
    `error_message` String
)
ENGINE = RabbitMQ
SETTINGS rabbitmq_host_port = 'rabbitmq.stub.svc:5672',
    rabbitmq_exchange_name = 'stub_events',
    rabbitmq_queue_base = 'staging_event_queue',
    rabbitmq_queue_consume = 1,
    rabbitmq_format = 'JSONEachRow';

CREATE MATERIALIZED VIEW default.event_consumer TO default.event_log
(
    `_trace_id` String,
    `_publish_time` DateTime64(3),
    `error_message` String
) AS
SELECT
    _trace_id,
    _publish_time,
    error_message
FROM default.event_queue;

CREATE TABLE default.event_log
(
    `_trace_id` String,
    `_publish_time` DateTime64(3),
    `error_message` String
)
ENGINE = ReplacingMergeTree
PARTITION BY toYYYYMM(_publish_time)
ORDER BY (_publish_time)
SETTINGS index_granularity = 8192;
```

**Additional context**
I tried to collect trace logs and make flame graphs (using this tool https://github.com/Slach/clickhouse-flamegraph/) if this will be in any way helpful

v22.11
![v22 11 global Memory](https://github.com/ClickHouse/ClickHouse/assets/45504682/8b9d80d5-2b23-4e2a-978f-6748ab921d19)
![v22 11 global MemorySample](https://github.com/ClickHouse/ClickHouse/assets/45504682/0211a2d5-01d3-4768-b96f-6d676599c03e)

v23.4
[...]

PR #102711, opened 2026-04-14
Fix rabbitmq engine performance issues

### LINKED ISSUE #62525: Highload CPU
**Describe the situation**
When using the RabbitMQ engine, there is a highload CPU. Even in the simplest scenario, the CPU load increases by more than 20 percent.

**How to reproduce**
```sql
CREATE TABLE test_order_queue
(
 id UInt32
)
ENGINE = RabbitMQ SETTINGS rabbitmq_host_port = '***', rabbitmq_exchange_name = 'monolith', rabbitmq_exchange_type = 'direct', rabbitmq_routing_key_list = 'non-existent', rabbitmq_format = 'JSONEachRow';

CREATE TABLEtest_order (
    id UInt32
) ENGINE = MergeTree()
ORDER BY id;

CREATE MATERIALIZED VIEW test_order_mv TO test_order AS
SELECT
    id AS id
FROM report.test_order_queue;

```  

* Queries to run that lead to slow performance. Data sent to queue. `{"id": 6}`
* ClickHouse server version to use. `23.8.12.13`
* Used Managed Service for ClickHouse by Yandex Cloud

jaapieaapie1:
- Performance Improvement

Fixed excessive RabbitMQ table engine CPU usage

Closes #62525 #50692 and #53401 

Before:
screenshot-2026-04-14_21-32-10

After:
screenshot-2026-04-14_21-30-30

Ofcourse verified that all functionality stays in tact and it still recovers from errors like disconnections from rabbitmq.

### Version info
- Merged into: `26.4.1.1015`
- Backported to: `26.3.10.45`

### DIFF
--- a/src/Storages/RabbitMQ/RabbitMQHandler.cpp
+++ b/src/Storages/RabbitMQ/RabbitMQHandler.cpp
@@ -68,7 +68,8 @@ int RabbitMQHandler::startBlockingLoop()
 
 void RabbitMQHandler::stopLoop()
 {
-    LOG_DEBUG(log, "Implicit loop stop.");
+    LOG_DEBUG(log, "Stopping background loop.");
+    loop_state.store(Loop::STOP);
 }
 
 void RabbitMQHandler::stopBlockingLoop()
--- a/src/Storages/RabbitMQ/RabbitMQHandler.cpp
+++ b/src/Storages/RabbitMQ/RabbitMQHandler.cpp
@@ -69,6 +69,11 @@ int RabbitMQHandler::startBlockingLoop()
 void RabbitMQHandler::stopLoop()
 {
     LOG_DEBUG(log, "Implicit loop stop.");
+}
+
+void RabbitMQHandler::stopBlockingLoop()
+{
+    LOG_DEBUG(log, "Stopping blocking loop.");
     uv_stop(loop);
 }
 
--- a/src/Storages/RabbitMQ/RabbitMQHandler.cpp
+++ b/src/Storages/RabbitMQ/RabbitMQHandler.cpp
@@ -40,7 +40,10 @@ void RabbitMQHandler::startLoop()
     loop_running.store(true);
 
     while (loop_state.load() == Loop::RUN)
-        uv_run(loop, UV_RUN_NOWAIT);
+    {
+        if (!uv_run(loop, UV_RUN_ONCE))
+            break;
+    }
 
     LOG_DEBUG(log, "Background loop ended");
     loop_running.store(false);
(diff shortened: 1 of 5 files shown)


src/Storages/RabbitMQ/RabbitMQHandler.cpp
```diff
@@ -40,7 +40,10 @@ void RabbitMQHandler::startLoop()
     loop_running.store(true);
 
     while (loop_state.load() == Loop::RUN)
-        uv_run(loop, UV_RUN_NOWAIT);
+    {
+        if (!uv_run(loop, UV_RUN_ONCE))
```
nikitamikhaylov: It seems calling `std::this_thread::yield();` when the `uv_run` return 0 is a better option, because the return value tells you whether the loop still has pending work (!= 0) or is fully drained (== 0).

src/Storages/RabbitMQ/RabbitMQHandler.cpp
```diff
@@ -40,7 +41,10 @@ void RabbitMQHandler::startLoop()
...
```
nikitamikhaylov: And we can return `UV_RUN_NOWAIT`

nikitamikhaylov:
Let's fix tests.

jaapieaapie1:
@nikitamikhaylov the remaining test failure seems unrelated or a random failure?

kssenii (approved):

tests/integration/test_storage_rabbitmq/test_shutdown.py
```diff
@@ -225,6 +225,107 @@ def test_attach_detach_rabbitmq_with_materialized_view(started_cluster):
...
+            rabbitmq_format = 'JSONEachRow',
+            rabbitmq_username = 'root',
+            rabbitmq_password = 'clickhouse',
+            rabbitmq_flush_interval_ms = 100,
+            rabbitmq_max_block_size = 100
+    """
+    )
+
+    # Wait for the connection to be established and the loop to start running.
+    time.sleep(10)
```
kssenii: Let's add a loop instead of sleep to make it more robust

jaapieaapie1: I don't think the sleep is actually necessary anymore because of the wait for logs beneath it. removed.

jaapieaapie1:
Failing test seems unrelated/flaky

kssenii (approved):

kssenii:
Thank you for the fix!

clickgapai:

**Affected code:** `RabbitMQHandler::startLoop` busy-spinning with `uv_run(UV_RUN_NOWAIT)` and `RabbitMQHandler::stopLoop` conflating blocking/non-blocking loop stop semantics — core code present in all supported branches.

**Why:** The busy-spinning CPU waste in `RabbitMQHandler::startLoop` has existed since the RabbitMQ engine was introduced in 2020, predating all supported branches. The fix is minimal (adding `yield` + method split) and safe. Users on all supported branches experience 20%+ unnecessary CPU load when using RabbitMQ tables.

jaapieaapie1:
@kssenii could you add backport labels? If you agree with this needing backporting.

PR #98678, opened 2026-03-04
Fix stale UUID in query storage cache causing UNKNOWN_TABLE for refreshable MVs

alexey-milovidov:
## Summary
- The per-query storage cache in `Context::getOrCacheStorage` caches table lookups by name, storing the UUID for fast subsequent lookups via `tryGetByUUID`
- When a table is dropped and recreated with a new UUID (as happens with refreshable materialized view inner tables on every non-APPEND refresh), the cached UUID becomes stale
- The old code returned null and threw "Table does not exist anymore - maybe it was dropped" instead of invalidating the stale entry and retrying the lookup by name
- Fix: erase the stale cache entry and fall through to `storage_getter` which resolves the table by name and caches the new UUID

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=96130&sha=eba1d6bd0c9b167be37d43e44a0b111d4088cc53&name_0=PR&name_1=Integration%20tests%20%28amd_asan%2C%20targeted%29
https://github.com/ClickHouse/ClickHouse/pull/96130

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
...

### Version info
- Merged into: `26.3.1.388`
- Backported to: `26.2.11.4`, `26.1.8.34`

PR #102724, opened 2026-04-15
lookup table by name not uuid that can be stale

seva-potapov:
- Improvement

during refresh of materialized view we lookup table by name instead of uuid that can be stale

### Details:

IdentifierResolver::resolveTableIdentifier calls resolveStorageID which stamps a UUID onto the StorageID, then passes it to RefreshTask::getAndLockTargetTable. Inside that function, DatabaseCatalog::getTable takes the UUID shortcut in getTableImpl (line 386), which throws UNKNOWN_TABLE when the UUID is stale (after EXCHANGE TABLES + DROP of old table). The 10-retry loop, designed exactly for this race, never fires because the exception propagates out.

### Version info
- Merged into: `26.6.1.845`


tiandiwonder (commented):
Changelog category to Bug Fix?

src/Storages/MaterializedView/RefreshTask.cpp
```diff
@@ -1099,6 +1099,14 @@ std::tuple<StoragePtr, TableLockHolder> RefreshTask::getAndLockTargetTable(const
...
     ///     If this fails, retry until we see a different table by the same name.
 
+    /// Look up by name, not UUID. The caller (IdentifierResolver) may pass a UUID-bearing StorageID
+    /// from resolveStorageID, but that UUID can be stale: EXCHANGE TABLES during refresh replaces
+    /// the target table with a new UUID, and then the old table is dropped. If we pass the stale
+    /// UUID to getTable, it throws UNKNOWN_TABLE via the tryGetByUUID shortcut in getTableImpl,
+    /// bypassing our retry loop. Name-based lookup is safe: the database mutex is held across
+    /// applyState renames, so getTable blocks during the rename and resolves correctly afterward.
+    StorageID name_only_id(storage_id.database_name, storage_id.table_name);
+
```
tiandiwonder, line 1109: better to have a short comment?
```
/// Look up by name (ignore UUID). During refresh, EXCHANGE TABLES may replace the target table
/// with a new UUID; using a stale UUID can fail fast and bypass the retry loop below.
StorageID name_only_id(storage_id.database_name, storage_id.table_name);
```

seva-potapov:
> Changelog category to Bug Fix?

no, because it's too hard to replicate in test and it would fail verification on unpatched clickhouse

tavplubix (commented):
See https://github.com/ClickHouse/ClickHouse/pull/98678/changes#r3094050289 

Resolving by name can cause other issues

seva-potapov:
@tavplubix getAndLockTargetTable is reached via RefreshSet::tryGetTaskForInnerTable, whose InnerTableMap is already keyed by DatabaseAndTableNameHash/Equal (RefreshSet.h:77), so we got here by name in the first place. 

The function's own doc comment (line 1094) starts with "Get table by name", and the retry loop is documented as "retry until we see a different table by the same name" and the UUID shortcut in getTableImpl was just preventing that loop from running.

There's also precedent for stripping the UUID when operating by name in this file: exchangeTargetTable does exactly that at StorageMaterializedView.cpp:644

tavplubix:
> getAndLockTargetTable is reached via RefreshSet::tryGetTaskForInnerTable, whose InnerTableMap is already keyed by DatabaseAndTableNameHash/Equal (RefreshSet.h:77), so we got here by name in the first place.

Are you sure that Claude is right about that? What about the `IdentifierResolver::tryResolveTableIdentifier` path? Looks like the issue that you are trying to fix happens with the `IdentifierResolver::tryResolveTableIdentifier` path, not with `RefreshSet::tryGetTaskForInnerTable` (check the stacktraces in https://pastila.clickhouse.com/?002ca9af/f2560244c841791b4cd4833a9e8fc9b7#L50S7KhXR08mdG2inCaqNA==GCM&nowrap)

> The function's own doc comment (line 1094) starts with "Get table by name"

Unfortunately, the comment doesn't say _why_, so it's a bad comment. Maybe this is relevant for `RefreshSet::tryGetTaskForInnerTable`, idk. But I'm not sure if it's correct for query analysis. AFAIU we should not switch to another table in the middle of query analysis 

> There's also precedent for stripping the UUID when operating by name in this file: exchangeTargetTable does exactly that at StorageMaterializedView.cpp:644

exchangeTargetTable has nothing to do with query analysis. Moreover, the only place where we use `fresh_table` after resetting uuid is 

```
    return exchange ? std::make_optional(fresh_table) : std::nullopt;
```
and therefore
```
            table_to_drop = view->exchangeTargetTable(new_table_id, refresh_context);
```
Now please answer the question: why do we need to reset it here? Maybe because we have just exchanged the tables, and the table we want to drop has a different uuid?

seva-potapov:
@tavplubix thank you for your review, you were right to push back, it made me understand the code path better and I believe I was able to address your feedback. I realised that we already have a similar fallback in [IdentifierResolver.cpp:230-237](https://github.com/seva-potapov/ClickHouse/blob/f49a6c0a1159fb9a1f5ad42625b0be772adcb72f/src/Analyzer/Resolve/IdentifierResolver.cpp#L256) (lines 256-263 after changes in this PR): when a UUID-bearing lookup returns null, we perform a name lookup. The existing fallback codifies "if UUID is gone, the current table at that name is the right answer." since commit 3fc2f551d147. This rework extends the same pattern to the refresh-task branch.

The catch handles both `UNKNOWN_TABLE` (after the old UUID is fully removed) and `TABLE_UUID_MISMATCH` (the post-EXCHANGE-but-pre-DROP window where the UUID still resolves to the now-renamed old storage).

src/Analyzer/Resolve/IdentifierResolver.cpp
```diff
@@ -222,7 +224,31 @@ std::shared_ptr<TableNode> IdentifierResolver::tryResolveTableIdentifier(const I
...
+            ///
+            /// Mirror the established UUID -> name fallback at the if-block below: when the
+            /// original UUID is unusable, retry against the current table at the same name.
+            /// Re-entering `getAndLockTargetTable` (instead of plain `tryGetTable`) preserves
+            /// the retry loop and the `SYSTEM SYNC REPLICA` call that this path needs.
+            if ((e.code() != ErrorCodes::UNKNOWN_TABLE && e.code() != ErrorCodes::TABLE_UUID_MISMATCH)
+                || !storage_id.hasUUID())
+                throw;
+            StorageID name_only_id(storage_id.database_name, storage_id.table_name);
+            std::tie(storage, storage_lock) = refresh_task->getAndLockTargetTable(name_only_id, context);
```
tavplubix, line 338: Looks like it can theoretically fail again with the same exception for the same reason, if refresh happens quickly enough, but it's unlikely on realistic usecases, so it's ok

tavplubix (approved):

PR #93830, opened 2026-01-09
Support for ALTER TABLE ADD ENUM VALUES

ilejn:
### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Implement `ADD ENUM VALUES` in `ALTER TABLE` queries to simplify appending new values to an existing Enum type without the need to specify all current Enum values again.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

---

> [!NOTE]
> **Medium Risk**
> Adds a new ALTER TABLE syntax and plumbing to merge enum definitions, touching SQL parsing and schema-alter preparation logic; mistakes could break ALTER parsing or allow invalid enum remaps.
> 
> **Overview**
> Adds `ALTER TABLE ... MODIFY/ALTER COLUMN ... ADD ENUM VALUES (...)` to append new members to existing `Enum`/`Enum8`/`Enum16` columns without restating the full definition, including `Nullable(Enum*)`.
> 
> Implements parsing/formatting support (`ADD_ENUM_VALUES` keyword, `ASTAlterCommand::add_enum_values`), adds enum-type merge/validation logic (`mergeEnumTypes`) with conflict and range checks (including optional relative numbering), and wires it into `AlterCommands::prepare` to produce the updated column type for the mutation.
> 
> Updates user docs for `Enum` and `ALTER COLUMN`, and adds stateless tests covering success cases and expected errors (non-enum columns, overflow, conflicting name/value, and syntax misuse).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9bf39da786d8e707421c40c2161bc6642bf65ccc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### Version info
- Merged into: `26.6.1.769`

PR #102735, opened 2026-04-15
Fix for unintended mutex recursion in Keeper

ilejn:

Addresses https://github.com/ClickHouse/ClickHouse/pull/93830#issuecomment-4243706437

Not sure that it is the best fix. 
May be someone who understands Keeper better can do it in more elegant way.

### DIFF
--- a/src/Coordination/KeeperStorage.cpp
+++ b/src/Coordination/KeeperStorage.cpp
@@ -1573,6 +1573,9 @@ auto callOnConcreteRequestType(Coordination::ZooKeeperRequest & zk_request, F fu
 namespace
 {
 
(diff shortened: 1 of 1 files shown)


antonio2368:
It was caused by this PR https://github.com/ClickHouse/ClickHouse/pull/102599 which was reverted.

ilejn:
> It was caused by this PR #102599 which was reverted.

Yes, thank you for confirmation, @antonio2368 !

Should I close this PR assuming that the issue is clear and would be handled?

antonio2368:
Yes, let's close the PR. The issue should be fixed.

PR #102736, opened 2026-04-15
allow serialization of all sorting types

seva-potapov:
- Improvement

we allow serialization of all sorting types

### Details:
SortingStep::serialize throws NOT_IMPLEMENTED for non-Full sorting types (e.g. FinishSorting from read-in-order optimization). This prevents the coordinator from serializing the query plan when serialize_query_plan=1 is used with Distributed tables or parallel replicas.


tests/queries/0_stateless/04061_sorting_step_serialize_finish_sorting.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE TABLE t_sorting_dist AS t_sorting_local ENGINE = Distributed(test_cluster_one_shard_three_replicas_localhost, currentDatabase(), t_sorting_local);
+
+INSERT INTO t_sorting_local SELECT number, toString(number) FROM numbers(1000);
+
+-- ORDER BY id matches the primary key, so the old interpreter creates a FinishSorting step
+-- directly during plan building. Serialization for the Distributed query must not throw.
+SELECT id, value FROM t_sorting_dist ORDER BY id LIMIT 5;
+
+DROP TABLE t_sorting_dist;
+DROP TABLE t_sorting_local;
```
azat, line 22: Hm, this test works even without this changes, also how it is possible to have query plan serialization with old analyzer? (cc @KochetovNicolai )

PR #102759, opened 2026-04-15
Apply Poisson sampling correction to collapsed jemalloc heap profiles

antonio2368:
Collapsed jemalloc heap profiles generated by `system.jemalloc_profile_text` and `SYSTEM JEMALLOC FLUSH PROFILE` reported raw sampled bytes without the Poisson sampling correction that jeprof applies. This caused the reported totals to underestimate actual allocation sizes — for example, ~5.9 GB raw vs ~11.2 GB corrected for a profile with the default 524288-byte sampling interval.

The fix parses the `heap_v2/N` header to extract the sampling interval, then scales each sample by `1/(1-exp(-mean_size/interval))` matching jeprof's `AdjustSamples` algorithm. Uses `std::expm1` for numerical stability with small ratios and guards against edge cases (zero bytes, non-finite results, overflow).

Also adds a `WriteBuffer` overload of `symbolizeJemallocHeapProfile` and unit tests.

Apply Poisson sampling correction to collapsed jemalloc heap profiles to match jeprof output. Previously the collapsed format underestimated actual allocation sizes by not accounting for the sampling probability.

No documentation changes needed — this is a correctness fix for existing functionality.

### Version info
- Merged into: `26.4.1.985`
- Backported to: `26.3.10.13`, `26.2.16.7`


src/Processors/Sources/JemallocProfileSource.cpp
```diff
@@ -608,7 +666,7 @@ std::string symbolizeJemallocHeapProfileToString(
 {
     WriteBufferFromOwnString out;
     pullProfileLines(input_filename, format, symbolize_with_inline, out);
-    return out.str();
+    return std::move(out.str());
```
azat, line 669: Shouldn't it always return rvalue?

azat (approved):

src/Processors/Sources/JemallocProfileSource.cpp
antonio2368: It probably makes sense to always return rvalue, not sure if it will break somewhere but we can check.

azat: Of course, but `out.str()` should be an rvalue already, to be precise `out.str()` is an `prvalue` and `std::move(out.str())` is an `xvalue`.

And RVO should be applied here before

I don't understand why clang-tidy does not complain about this line...

antonio2368: I'm not sure I follow. `out.str()` returns `std::string &` so it's lvalue. Am I missing something?

azat: Oh, sorry for the noise, I thought that `str()` returns `std::string`.

PR #102802, opened 2026-04-15
Fix crash in `hasPhrase` with a `NULL` argument

nickitat:

Fix the crash in the function `hasPhrase` with a `NULL` argument.

### Version info
- Merged into: `26.4.1.974`


tests/queries/0_stateless/04099_has_phrase_null_argument.sql
```diff
@@ -0,0 +1 @@
+SELECT hasPhrase('', materialize(NULL));
```
ahmadov: I think it should behave as this test: https://github.com/ClickHouse/ClickHouse/blob/118bba8d9f770e4e7d14f94e057aaa95a4f54a4b/tests/queries/0_stateless/02346_text_index_function_hasAnyAllTokens.sql#L33 The input should be a "const String", any other inputs should not be accepted. Also, could you please move this test to `02346_function_hasPhrase.sql`?

ahmadov (approved):
LGTM! Thanks for the fix!

PR #102804, opened 2026-04-15
Make waitDetachedTableNotInUse cancellable and shutdown-aware

antonio2368:
`DatabaseAtomic::waitDetachedTableNotInUse` was an infinite `while(true)` busy-wait with 100ms sleep, no cancellation, no shutdown check, and no diagnostics. When a concurrent query held a `StoragePtr` (e.g. a `system.replicas` snapshot iterating all replicated tables), the refcount stayed above 1 and `DETACH DATABASE` hung indefinitely, causing 41 hung check failures in 30 days across stress test variants.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=58cd21f1f03e688719d58648aa1b2decb96ab996&name_0=MasterCI&name_1=Stress%20test%20%28amd_debug%29

Apply the same fix pattern from PR #97586 (`waitTableFinallyDropped`):
- Add `std::function<void()> throw_if_cancelled` parameter so `KILL QUERY` works
- Check `DatabaseCatalog::isShuttingDown` to exit on server shutdown
- Log `use_count` every ~10s for diagnostics
- Do a final cleanup pass on shutdown before throwing `UNFINISHED`
- Update all 5 call sites to pass cancellation callbacks
- Add regression test `04099_detach_database_sync_cancellation`

Make `DETACH DATABASE ... SYNC` and other operations that call `waitDetachedTableNotInUse` cancellable via `KILL QUERY` and responsive to server shutdown, preventing indefinite hangs when a concurrent query holds a table reference.

### Version info
- Merged into: `26.5.1.507`


clickgapai:

**Affected code:** DatabaseAtomic::waitDetachedTableNotInUse — core code present in all supported branches.

**Why this might apply:** The bug affects DETACH DATABASE SYNC, ATTACH TABLE, RESTART REPLICA, and similar operations that call waitDetachedTableNotInUse. The affected code (DatabaseAtomic::waitDetachedTableNotInUse) predates all supported branches. PR description mentions 41 hung check failures in 30 days across stress tests.

PR #102817, opened 2026-04-15
Fix part serialization info after column mutation

korowa:

Fix inconsistent part metadata after mutations of columns with non-default serializations

### Details
`MergedColumnOnlyOutputStream` which is used when the mutation should rewrite only specific part columns, creates a writer which uses `data_part` serializations, and later registers new part metadata with `new_serialization_infos`, created in base class and built from storage settings.

For the cases when the mutation overwrites columns with the serializations different from the storage-defined ones, this produces a divergence in on-disk files layout and in-memory part metadata which may lead to LOGICAL_ERROR while reading (e.g. when the reader expects String.size to exist, based on wrong serialization stored in in-memory metadata).

This patch adds synchronization between serializations used by writer and serializations in new part metadata.

### Version info
- Merged into: `26.6.1.1004`


scanhex12 (approved):

alexey-milovidov (approved):
Ok.

alexey-milovidov:
@groeneai, why did Upgrade Check fail here?

groeneai:
Infra flake in job setup, not this PR. The Upgrade check downloads the previous stable release (`v26.5.1.882-stable`) `.deb`s before it starts the server. On commit d637c30 the download of `clickhouse-common-static_26.5.1.882_amd64.deb` exhausted all 4 retries (other packages of the same release downloaded fine), so the script aborted with `failure: Failed to download previous release packages` / `exit 1` in the setup phase. The actual upgrade test never ran: `fatal.log` is empty and CIDB recorded only 3 rows for that run (`Download script exit code` / `Check failed` / `Failures: 1/2`), with no `Server successfully started` or `No Error messages after server upgrade` rows.

The serialization/mutation change here only runs after install+upgrade, which never happened. Same PR, earlier commits, passed Upgrade check fully green (0/12): bd171e0 (2026-05-18), 6b91330 (2026-04-17), and 3 earlier runs. A re-run should clear it.

Job log: https://github.com/ClickHouse/ClickHouse/actions/runs/27117485031/job/80044635535

— Groene AI (session cron:clickhouse-worker-slot-1:20260617-041000)

PR #102853, opened 2026-04-15
Fix server crash from recursive Avro schema

### LINKED ISSUE #102809: Vulnerabilities exploitable by an attacker who controls data that ClickHouse reads
Hi all, I got inspired by [this article](https://blog.calif.io/p/mad-bugs-vim-vs-emacs-vs-claude) (also since [Mythos isnt the only good model at finding vulnerabilities](https://valmarelox.substack.com/p/project-glasswing-a-claude-mythos)), I decided to run the same prompt against ClickHouse.

**I'm absolutely no security expert and no C++ expert so take everything with a grain of salt** but:
the good news is no direct RCE the bad news is a bunch of unwanted crashes.

I'll attach the report and the how to reproduce the first one (I havent tested the other ones reported by Claude):
- get a client running attached to whatever server (be carefull the server is going to be shutdown with this exploit)
- run this query with the attached file: `SELECT * FROM file('malicious.avro', 'Avro')`
- the server will shutdown (and restart again if enabled)

So an external source of data (like a S3 bucket) could in theory crash a CH DB with just a query.

I dont know if this is valuable or just slop but I figured it was worth sharing.

Image

[malicious.avro.zip](https://github.com/user-attachments/files/26748804/malicious.avro.zip)
[CH-vulnerabilities.md](https://github.com/user-attachments/files/26748805/CH-vulnerabilities.md)

*I'm a ClickHouse employee so if you need any more clarification slack me, I can also provide links to the original reporting*

groeneai:
A crafted `.avro` file with a mutually recursive schema (e.g. `TypeA → TypeB → TypeA`) causes `avroNodeToDataType()`, `createDeserializeFn()`, and `createAction()` to recurse infinitely through `AVRO_SYMBOLIC` nodes, overflowing the stack and crashing the server with SIGSEGV. Any query reading such a file (from S3, HTTP, or local path) triggers the crash — no authentication required.

The fix adds explicit cycle detection to all three vulnerable code paths using name-tracking sets that detect back-references immediately (on the first re-entry, not after thousands of stack frames). Non-recursive re-use of named types is unaffected — verified with existing test `02592_avro_records_with_same_names`.

`createSkipFn()` was already protected via `symbolic_skip_fn_map` — the same defensive pattern.

Complementary to #102417 which adds generic `checkStackSize()` guards (catches all deep nesting including non-recursive). This PR specifically detects logical cycles with an immediate, targeted error message.

**Reproduction:**
```sql
-- With the crafted malicious.avro from #102809:
SELECT * FROM file('malicious.avro', 'Avro');  -- SIGSEGV before fix, clean error after
```

**After fix:**
```
Code: 36. Recursive Avro schema is not supported: type 'TypeB' references itself. (BAD_ARGUMENTS)
```

Closes #102809

Fix server crash (SIGSEGV) when reading Avro files with recursive schemas containing cyclic symbolic type references. Now such schemas are detected and rejected with a clear error message instead of crashing.

### Version info
- Merged into: `26.4.1.1106`

### DIFF
--- a/src/Processors/Formats/Impl/AvroRowInputFormat.cpp
+++ b/src/Processors/Formats/Impl/AvroRowInputFormat.cpp
@@ -667,7 +667,15 @@ AvroDeserializer::DeserializeFn AvroDeserializer::createDeserializeFn(const avro
             break;
         }
         case avro::AVRO_SYMBOLIC:
-            return createDeserializeFn(avro::resolveSymbol(root_node), target_type);
+        {
+            const auto & sym_name = root_node->name().fullname();
+            if (!symbolic_deserialize_guard.insert(sym_name).second)
+                throw Exception(ErrorCodes::BAD_ARGUMENTS,
+                    "Recursive Avro schema is not supported: type '{}' references itself", sym_name);
+            auto result = createDeserializeFn(avro::resolveSymbol(root_node), target_type);
+            symbolic_deserialize_guard.erase(sym_name);
+            return result;
+        }
         case avro::AVRO_RECORD:
         {
             if (target.isTuple())
@@ -916,15 +924,22 @@ AvroDeserializer::Action AvroDeserializer::createAction(const Block & header, co
 {
     if (node->type() == avro::AVRO_SYMBOLIC)
     {
+        const auto & sym_name = node->name().fullname();
+        if (!symbolic_deserialize_guard.insert(sym_name).second)
+            throw Exception(ErrorCodes::BAD_ARGUMENTS,
+                "Recursive Avro schema is not supported: type '{}' references itself", sym_name);
+
         /// continue traversal only if some column name starts with current_path
         auto keep_going = std::any_of(header.begin(), header.end(), [&current_path](const ColumnWithTypeAndName & col)
         {
             return col.name.starts_with(current_path);
         });
         auto resolved_node = avro::resolveSymbol(node);
-        if (keep_going)
-            return createAction(header, resolved_node, current_path);
-        return AvroDeserializer::Action(createSkipFn(resolved_node));
+        Action result = keep_going
+            ? createAction(header, resolved_node, current_path)
+            : AvroDeserializer::Action(createSkipFn(resolved_node));
+        symbolic_deserialize_guard.erase(sym_name);
+        return result;
     }
 
     if (header.has(current_path))
@@ -1338,6 +1353,12 @@ NamesAndTypesList AvroSchemaReader::readSchema()
 
 DataTypePtr AvroSchemaReader::avroNodeToDataType(avro::NodePtr node)
 {
+    std::unordered_set<std::string> seen_names;
+    return avroNodeToDataTypeImpl(node, seen_names);
+}
+
+DataTypePtr AvroSchemaReader::avroNodeToDataTypeImpl(const avro::NodePtr & node, std::unordered_set<std::string> & seen_names)
+{
     switch (node->type())
     {
         case avro::Type::AVRO_INT:
@@ -1406,7 +1427,7 @@ DataTypePtr AvroSchemaReader::avroNodeToDataType(avro::NodePtr node)
             return std::make_shared<DataTypeFixedString>(node->fixedSize());
         }
         case avro::Type::AVRO_ARRAY:
-            return std::make_shared<DataTypeArray>(avroNodeToDataType(node->leafAt(0)));
+            return std::make_shared<DataTypeArray>(avroNodeToDataTypeImpl(node->leafAt(0), seen_names));
         case avro::Type::AVRO_NULL:
             return std::make_shared<DataTypeNothing>();
         case avro::Type::AVRO_UNION:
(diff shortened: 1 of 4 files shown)


groeneai:
cc @Avogar @nikitamikhaylov — could you review this? It adds explicit cycle detection for recursive Avro schemas (the vulnerability from #102809). Complementary to #102417 — this catches logical cycles immediately with a targeted error message, while #102417's `checkStackSize()` catches all deep nesting generically. Both can coexist.

**Session:** cron:clickhouse-ci-task-worker:20260415-184500

**Pre-PR validation:**
- (a) Deterministic repro: `clickhouse local -q "SELECT * FROM file('malicious.avro', 'Avro')"` → SIGSEGV (exit 139) without fix, clean error with fix
- (b) Root cause: `AVRO_SYMBOLIC` resolution creates pointer cycle; `avroNodeToDataType`/`createDeserializeFn`/`createAction` recurse infinitely; `createSkipFn` already had protection via `symbolic_skip_fn_map`
- (c) Fix matches cause: name-tracking sets detect cycles on first re-entry
- (d) New test: `04101_avro_recursive_schema_crash` crafts the exact attack vector; existing `02592_avro_records_with_same_names` verifies no false positives
- (e) Both directions confirmed: SIGSEGV without → clean error with
- (f) All 3 vulnerable paths fixed; `createSkipFn` already protected

Avogar (approved):

Avogar:
@groeneai test `01060_avro` fails in the CI with:
```
2026-04-15 23:07:27 Reason: return code:  36
2026-04-15 23:07:27 Code: 36. DB::Exception: Recursive Avro schema is not supported: type 'sub1' references itself: (while reading header): (in file/uri (fd = 0)): While executing AvroRowInputFormat: While executing File. (BAD_ARGUMENTS)
```

Does it mean we tested Avro file with recursive schema and parsed it successfully?

groeneai:
@Avogar thanks for catching that — the failing `01060_avro` is NOT a legitimate recursive-schema read, it was my cycle-detection being too aggressive. Fixed in 919939285ec ("Allow safe skip of cyclic symbolic references in createAction").

### What was happening

`tests/queries/0_stateless/data_avro/nested_complex.avsc` has a truly recursive schema:
```json
{
  "type": "record", "name": "main",
  "fields": [
    {"name": "a", "type": "long"},
    {"name": "b", "type": {
      "type": "record", "name": "sub1",
      "fields": [
        ...
        {"name": "b4_null_sub1", "type": ["null", "sub1"]}   // <- self-reference
      ]
    }}
  ]
}
```

The test reads column `b.b4_null_sub1.sub1.b2_null_str` (goes through the recursive `sub1` link one level, reads a leaf field).

Traversal in `createAction`:
1. `main` (record) -> iterate fields
2. `b` (record `sub1`) -> iterate fields
3. `b.b4_null_sub1` (union `[null, SYMBOLIC sub1]`) -> iterate branches
4. `SYMBOLIC sub1` at path `b.b4_null_sub1.sub1`, `keep_going=true` -> recurse into resolved `sub1` record
5. Inside, `b.b4_null_sub1.sub1.b4_null_sub1` is another union with `SYMBOLIC sub1`
6. That second `SYMBOLIC sub1` is at path `b.b4_null_sub1.sub1.b4_null_sub1.sub1`. The requested column does **not** start with this path, so `keep_going=false` and the old code took the safe `createSkipFn` path (which already handles cycles via `symbolic_skip_fn_map`).

My original patch inserted into `symbolic_deserialize_guard` **before** the `keep_going` check, so the second occurrence of `"sub1"` was flagged as a cycle even though we'd never actually recurse into it. That's the `BAD_ARGUMENTS` you saw.

### The fix (919939285ec)

```cpp
if (node->type() == avro::AVRO_SYMBOLIC)
{
    auto keep_going = std::any_of(header.begin(), header.end(), ...);
    auto resolved_node = avro::resolveSymbol(node);

    // Safe skip path handles cycles via symbolic_skip_fn_map — no guard needed.
    if (!keep_going)
        return AvroDeserializer::Action(createSkipFn(resolved_node));

    // Only guard the genuinely-recursive path.
    const auto & sym_name = node->name().fullname();
    if (!symbolic_deserialize_guard.insert(sym_name).second)
        throw Exception(..., "Recursive Avro schema is not supported: type '{}' references itself", sym_name);
    Action result = createAction(header, resolved_node, current_path);
    symbolic_deserialize_guard.erase(sym_name);
    return result;
}
```

So: schemas that are truly recursive remain readable as long as the requested column paths are finite. Only the case where the user *does* want to descend into a cycle (which would require infinite unrolling) still errors out. The `04101_avro_recursive_schema_crash` regression test still catches the original SIGSEGV — that path goes through `avroNodeToDataTypeImpl` (schema inference, `DESCRIBE` / `SELECT *`) and through `createDeserializeFn` when the target type itself is recursive, both of which keep the unconditional guard.

### Verified locally

- `01060_avro`: FAIL -> PASS (confirmed the exact `Code: 36` error without the patch, exactly as you pasted, and PASS with it).
- `04101_avro_recursive_schema_crash`: PASS (regression test still catches the malicious file).
- `02592_avro_records_with_same_names`: PASS (no false positives for records with reused field names).
- 10 consecutive runs of all three: 30/30 passed.

Session id: cron:clickhouse-ci-task-worker:20260420-184500

clickgapai:

**Affected code:** `AvroSchemaReader::avroNodeToDataType`, `AvroDeserializer::createDeserializeFn`, `AvroDeserializer::createAction` — core code present in all supported branches.

**Why:** SIGSEGV crash from crafted Avro input is a security vulnerability (no authentication required). The Avro format support and `AVRO_SYMBOLIC` handling have existed since at least 2019, predating all supported branches. All supported branches are affected.

PR #102854, opened 2026-04-15
feat(formats): Add OpenMetrics text format (read/write).

JTCunning:
I'm on the fence about creating a parameterized view or a lightweight table function that outputs rows into the format for either prom or openmetrics. the big ol' join that the integration tests are using to morph ts data into the format is.... not very clean.

Once selecting and inserting directly on the ts engine works, this will get a lot cleaner.

---

- OpenMetrics input (`OpenMetricsTextRowInputFormat` + schema reader): `# HELP` / `# TYPE` / `# UNIT`, histogram/summary `_bucket` / `_sum` / `_count` folding only when `# TYPE` is histogram or summary, strict sample lines (Prometheus-grammar metric/label names, mandatory whitespace between descriptor and value, real-number timestamps, valid `# {labels} value [timestamp]` exemplar or reject, no trailing junk, `# EOF` ends stream)
- OpenMetrics output (`OpenMetricsTextOutputFormat`): standalone `IRowOutputFormat` (does not modify `FORMAT Prometheus`); `# HELP` / `# TYPE` / `# UNIT`, histogram/summary bucket/`_sum`/`_count` ordering, OpenMetrics timestamp rules (emit when present and non-NULL, including zero), trailing `# EOF`
- Register `OpenMetrics` in FormatFactory (read/write, schema inference, `application/openmetrics-text`); exclude from sparse/async all-formats tests
- Stateless `04003` / `03535`; integration: PRW ingest to ClickHouse, then same TimeSeries `SELECT` with `FORMAT Prometheus` vs `FORMAT OpenMetrics` (compare after stripping OpenMetrics-only lines)

Made-with: Cursor

- Experimental Feature

Implement the OpenMetrics input and output format.


JTCunning:
Screenshot 2026-06-04 at 1 49 00 PM
Proof that the robot once approved my PR since it's been known to betray me.

JTCunning:
The robot has approved this PR several times, only to produce new feedback when master has been merged.
I am going to stop aiming at a fully green approval, and aim at merging and iterating on this experimental format.

vitlibar:
Two major points:

1. The format is named `OpenMetrics`, but the documentation says `The output does not advertise OpenMetrics 1.0 compliance`. I think we should try to comply with `OpenMetrics 1.0`

2. The columns are `name, value, help, type, labels, timestamp`. These columns don't match TimeSeries columns making difficult to move data between a TimeSeries table and this format. I think the columns should be `metric_name`, `metric_family`, `help`, `type`, `unit`, `tags Array(Tuple(String, String))`, `time_series Array(Tuple(DateTime64(3), Float64))` - this would allow us to run queries like
```
INSERT INTO ts FORMAT OpenMetrics

SELECT * FROM ts WHERE timestamp > now() - INTERVAL 5 MINUTES FORMAT OpenMetrics
```
(Support for `SELECT * FROM ts` is coming in https://github.com/ClickHouse/ClickHouse/pull/106010)

JTCunning:
Cool, we can definitely wait on #106010 for this, and I'll produce those changes.

PR #102719, opened 2026-04-14
[WIP] perf(ts-engine): Add forwards-compatible metric_locality_id.

JTCunning:
Adding to CI to see how this does. I believe I covered every case but I need more.

Closes https://github.com/ClickHouse/ClickHouse/issues/100906.

### Changelog category (leave one):
- Performance Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Add `metric_locality_id` to the TimeSeries table engine to increase metric density in range scans.

PR #102856, opened 2026-04-15
Replace assert with chassert in AggregateFunctionArray constructor

groeneai:

### Changelog entry (a]sentence-length human-readable entry for the changelog, not a copy of the PR title. Skip for non-user-facing changes):
- (not required)

### Documentation checkbox (leave empty if not applicable):
- [ ] Documentation is written (if applicable)

---

## Summary

The `AggregateFunctionArray` constructor has an `assert()` (C standard library) that verifies parameter consistency between the outer Array wrapper and its nested function. The AST fuzzer occasionally generates queries with rare combinator + Nullable interaction paths that violate this invariant (STID 4870-4f21, persisting after PR #100679 which fixed the primary Nothing-function case).

The C `assert()` calls `abort()` which kills the entire server process (SIGABRT). This is disproportionate for a metadata inconsistency — the parameter mismatch does not cause data corruption or incorrect results.

## Fix

Replace `assert()` with `chassert()`:
- **Debug/sanitizer builds**: throws `LOGICAL_ERROR` exception (caught by query error handling, server survives)
- **Release builds**: no-op (same behavior as before — the assert was never active in release)

This is the standard ClickHouse pattern for query-path invariant checks (~2000+ `chassert` instances in the codebase).

## CIDB Evidence

STID 4870-4f21 occurrences in the last 30 days:
- 4+ hits across PRs #99581, #101292, #102719 (all unrelated to aggregate functions) and master serverfuzz
- The previous fix (PR #100679, merged March 29) resolved the primary cause (Nothing functions discarding parameters) but a rare secondary path remains
- The assertion fires only in debug builds under AST fuzzer mutation — extremely rare (~4 hits in 30 days of continuous CI fuzzing)

### Version info
- Merged into: `26.4.1.982`

### DIFF
--- a/src/AggregateFunctions/Combinators/AggregateFunctionArray.h
+++ b/src/AggregateFunctions/Combinators/AggregateFunctionArray.h
@@ -33,7 +33,7 @@ public:
         : IAggregateFunctionHelper<AggregateFunctionArray>(arguments, params_, createResultType(nested_))
         , nested_func(nested_), num_arguments(arguments.size())
     {
-        assert(parameters == nested_func->getParameters());
+        chassert(parameters == nested_func->getParameters());
         for (const auto & type : arguments)
             if (!isArray(type))
                 throw Exception(ErrorCodes::ILLEGAL_TYPE_OF_ARGUMENT, "All arguments for aggregate function {} must be arrays", getName());


groeneai:
## Pre-PR Validation Gate (Session: cron:clickhouse-ci-task-worker:20260415-201500)

**a) Deterministic repro?**
The exact trigger query could not be reproduced locally even with extended AST fuzzer runs (5000+ mutations over 120s). The crash is extremely rare (~4 hits in 30 days of CI fuzzing across millions of queries). However, the code change is mechanical and does not require a specific reproducer — it replaces one macro with another.

**b) Root cause explained?**
The `AggregateFunctionArray` constructor uses C `assert()` to verify that its stored parameters match the nested function's parameters. PR #100679 fixed the primary case (Nothing-type functions discarding parameters), but a rare secondary code path through the combinator chain + Nullable type resolution can still produce a parameter mismatch. When triggered, `assert()` calls `abort()`, killing the server process with SIGABRT. The parameter mismatch itself is a metadata inconsistency with no impact on query results or data integrity.

Yes. The fix directly addresses the crash mechanism: `assert()` → `chassert()`. In debug/sanitizer builds, `chassert` throws a `LOGICAL_ERROR` exception (caught by query error handling). In release builds, both are no-ops. The invariant check is preserved but its failure mode changes from server crash to query error.

**d) Test intent preserved?**
The invariant check is fully preserved as `chassert`. It continues to detect parameter mismatches in debug/sanitizer builds, just without killing the server. The regression test from PR #100679 (`04057_aggregate_function_nothing_with_parameters`) still passes.

The existing test suite passes with the change. The `chassert` preserves the invariant detection that `assert` provided, while preventing server death. The change is a standard ClickHouse practice — `chassert` is used ~2000+ times throughout the codebase for exactly this purpose.

groeneai:
cc @Avogar — could you review this? It replaces a C `assert()` with `chassert()` in the AggregateFunctionArray constructor to prevent server crash (SIGABRT) when the AST fuzzer triggers a rare parameter consistency violation through combinator+Nullable interaction paths (STID 4870-4f21).

nihalzp (approved):

PR #102868, opened 2026-04-16
Replace chassert with diagnostic LOGICAL_ERROR in AggregateFunctionArray

groeneai:
The assertion `parameters == nested_func->getParameters()` in `AggregateFunctionArray` has been firing sporadically in CI (~4 hits per month) through AST fuzzer and serverfuzz stress tests (STID 4870-4f21, STID 4870-4e45, STID 2508-522c).

The previous `chassert` (converted from C `assert()` in #102856, before that fixed primary root cause in #100679) only provides the expression text — zero information about what the actual parameters are when it fires. This makes investigation of the remaining secondary code path extremely difficult.

**What this PR does:**

Replaces the `chassert` with an explicit `LOGICAL_ERROR` exception that logs:
- The Array wrapper function name (e.g., `quantileIfArrayArray`)
- The nested function name (e.g., `nothingNull`)  
- Both parameter sets with their actual values and sizes

Example error message:
```
AggregateFunctionArray: parameters mismatch between Array wrapper 'quantileIfArrayArray' 
and nested function 'nothingNull'. Wrapper has 1 parameter(s): [0.5], 
nested function has 0 parameter(s): []
```

**Why:**

The invariant that `parameters == nested_func->getParameters()` should always hold because the Array combinator does not transform parameters. When it fails, it indicates a bug in `AggregateFunctionFactory::getImpl()` or a combinator wrapper that loses/modifies parameters during resolution.

The diagnostic info will immediately identify which combinator+function combination triggers the mismatch, enabling targeted root cause analysis on the next CI occurrence.

**Behavior:**
- Debug/sanitizer builds: throws LOGICAL_ERROR (same as chassert, but with useful message)
- Release builds: throws LOGICAL_ERROR (improvement — chassert was a no-op in release)

### Changelog entry (a]utofill):
- ...no entry required...

- [ ] Documentation is written (if applicable)

### Version info
- Merged into: `26.4.1.1009`

### DIFF
--- a/src/AggregateFunctions/Combinators/AggregateFunctionArray.h
+++ b/src/AggregateFunctions/Combinators/AggregateFunctionArray.h
@@ -41,7 +41,8 @@ public:
             /// parameter set. If this fires, it means some code path in
             /// AggregateFunctionFactory or a combinator wrapper lost/modified parameters.
             /// The diagnostic info below will identify the exact mismatch.
-            String outer_params_str, nested_params_str;
+            String outer_params_str;
+            String nested_params_str;
             for (const auto & p : parameters)
             {
                 if (!outer_params_str.empty()) outer_params_str += ", ";
--- a/src/AggregateFunctions/Combinators/AggregateFunctionArray.h
+++ b/src/AggregateFunctions/Combinators/AggregateFunctionArray.h
@@ -13,6 +13,7 @@ struct Settings;
 
 namespace ErrorCodes
 {
+    extern const int LOGICAL_ERROR;
     extern const int SIZES_OF_ARRAYS_DONT_MATCH;
     extern const int ILLEGAL_TYPE_OF_ARGUMENT;
 }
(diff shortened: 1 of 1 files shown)


groeneai:
### Pre-PR Validation Gate (session: cron:clickhouse-ci-task-worker:20260416-061500)

**a) Deterministic repro?**
N/A — this is a diagnostic improvement, not a bug fix. The underlying assertion fires ~4 times/month in CI via AST fuzzer mutations that I was unable to reproduce locally after 15+ minutes of targeted fuzzing across 1000+ systematic combinator combinations. The fix preserves the exact same invariant check.

**b) Root cause explained?**
The root cause of the original assertion (`parameters == nested_func->getParameters()`) is an unknown code path in the aggregate function combinator resolution that loses or modifies parameters when the Array combinator wraps a nested function. PR #100679 fixed the primary path (AggregateFunctionNothingImpl discarding parameters). A secondary path remains unidentified — this PR provides the diagnostic data needed to identify it on the next CI occurrence.

This is a diagnostic improvement — it replaces a bare assertion (that provides zero information about the actual parameter values) with a proper LOGICAL_ERROR exception that logs both parameter sets. It does NOT change the invariant being checked.

**d) Test intent preserved?**
The check is unchanged — parameters must match. Only the error reporting is improved: from an opaque "assertion failed" to a detailed message showing exactly which function and parameters are mismatched.

All existing aggregate function tests pass with the change. The assertion path cannot be triggered locally (see point a), but the error handling behavior is verified through code review: `throw Exception(ErrorCodes::LOGICAL_ERROR, ...)` is the standard ClickHouse pattern for invariant violations.

groeneai:
cc @Avogar — could you review this? It replaces the bare `chassert` in AggregateFunctionArray (from #102856) with a proper LOGICAL_ERROR that logs both parameter sets. The invariant still fires ~4x/month in CI fuzzers but the current error gives zero diagnostic info about what parameters actually mismatch.

Avogar (approved):

Avogar:
@groeneai there is a failure of tidy build:
```
DEBUG:clang-tidy-cache.py:Calling real clang-tidy.
/ClickHouse/src/AggregateFunctions/Combinators/AggregateFunctionArray.h:44:13: error: multiple declarations in a single statement reduces readability [readability-isolate-declaration,-warnings-as-errors]
   44 |             String outer_params_str, nested_params_str;
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Fix it.

groeneai:
@Avogar Thanks for the review and the approval! Fixed the clang-tidy `readability-isolate-declaration` warning — split `String outer_params_str, nested_params_str;` into two separate declarations.

PR #102884, opened 2026-04-16
Optimization for deferring row policy and PREWHERE

yariks5s:
With FINAL queries and `apply_row_policy_after_final = 1` (default), this PR narrows when the row policy and `PREWHERE` are deferred to after `FINAL`

This is the explain without the patch: https://fiddle.clickhouse.com/6251834d-26e7-4a4e-8bc4-ba4723de4420

This is the explain after the patch: https://pastila.clickhouse.com/?00224304/0d1aaa7352c7f511fa80cd31f1ba8eef#snF8VfZFxcq4d+YOsS2qvA==GCM

For a `FINAL` read with a row policy (and optionally a `PREWHERE`):                                                                                                                                                       
   
  | Row policy                  | PREWHERE | Before                                             | After                                 |                                                                                   
  |-----------------------------|----------|----------------------------------------------------|---------------------------------------|
  | SK-only, deterministic      | none     | row policy deferred                                | **row policy not deferred**           |                                                                                   
  | SK-only, deterministic      | any      | row policy deferred, PREWHERE follows user setting | **neither deferred**                  |                                                                                   
  | SK-only, non-deterministic  | any      | row policy deferred                                | row policy still deferred (unchanged) |                                                                                   
  | non-SK                      | none     | row policy deferred                                | row policy deferred (unchanged)       |                                                                                   
  | non-SK                      | any      | row policy deferred, PREWHERE force-deferred       | row policy deferred, PREWHERE force-deferred (unchanged) | 

PREWHERE is not deferred anymore in this case. This is the case where we apply PREWHERE on sorting key column, when we probably can avoid deferring

- Improvement

Skip deferring the row policy after `FINAL` when it depends only on sorting-key columns and is deterministic, and skip the corresponding `PREWHERE` deferral that the row policy was forcing in that case

### Version info
- Backported to: `26.4.3.35`, `26.3.11.30`, `26.2.19.33`


tests/queries/0_stateless/03742_apply_row_policy_after_final.sql
```diff
@@ -321,3 +321,34 @@ SELECT explain FROM (EXPLAIN actions=1 SELECT * FROM tab_nested_and_pw FINAL PRE
 
 SET apply_prewhere_after_final = 0;
 DROP TABLE tab_nested_and_pw;
+
+SELECT '';
```
rienath, line 325: Let's cover determinism guard by test. Currently, CI would be green if we accidentally drop `isDeterministic` check in future. Let's also add exercise the new "skip deferral entirely" branch in row policy over SK with no `PREWHERE`

src/Processors/QueryPlan/ReadFromMergeTree.cpp
```diff
@@ -107,6 +108,9 @@ bool isNodeOverSortingKey(const ActionsDAG::Node * node, const NameSet & sorting
         return true; // constants are fine
     if (node->type == ActionsDAG::ActionType::INPUT || node->type == ActionsDAG::ActionType::PLACEHOLDER)
         return false; // already checked result_name
+    if (node->type == ActionsDAG::ActionType::FUNCTION
```
rienath, line 119: Wdyt of pulling the determinism check into its own helper and ANDing it at the call site? `isNodeOverSortingKey` is a structural check, determinism is a semantic property we only need for the safe to commute with `FINAL`. The other caller at line 2273  doesn't need determinism as KeyCondition ignores non-deterministic functions anyway. Splitting also keeps each function honest to its name and makes the reasoning clear at the call site

tests/queries/0_stateless/03742_apply_row_policy_after_final.sql
rienath: Also, how about making a new file to test all this? Current file is already crowded

src/Processors/QueryPlan/ReadFromMergeTree.cpp
```diff
@@ -2180,16 +2184,29 @@ void ReadFromMergeTree::deferFiltersAfterFinalIfNeeded()
...
+            defer_row_policy = false;
+
+        /// If row policy touches non-sorting-key columns and there is a PREWHERE,
+        /// prewhere must also be deferred — unless it only depends on sorting key columns
+        if (!row_policy_over_sk && query_info.prewhere_info)
+        {
+            const auto * prewhere_output = &query_info.prewhere_info->prewhere_actions.findInOutputs(
+                query_info.prewhere_info->prewhere_column_name);
+            if (!isNodeOverSortingKey(prewhere_output, sorting_key_set))
+                defer_prewhere = true;
```
rienath: This new conditional lets `PREWHERE` skip the deferral. But pushing `PREWHERE` before applying row policy is dangerous here. Consider this: https://fiddle.clickhouse.com/0e09b60f-0ba1-48f1-bfc0-2c3192fa03c9. On master, we filter out rows the user shouldn't see before evaluating `throwIf(id = 2, 'row id=2 exists in the table') = 0`. After this change we get

```
Code: 395. DB::Exception: row id=2 exists in the table: while executing 'FUNCTION throwIf(equals(id, 2_UInt8) :: 4, 'row id=2 exists in the table'_String :: 2) -> throwIf(equals(id, 2_UInt8), 'row id=2 exists in the table'_String) UInt8 : 1': While executing MergeTreeSelect(pool: ReadPoolInOrder, algorithm: InOrder). (FUNCTION_THROW_IF_VALUE_IS_NON_ZERO)
```

allowing the attacker to probe values in rows that the policy was meant to hide

rienath: If you end up removing this, it'd be great to add a regression test s.t. we do not introduce this behaviour in the future by mistake

rienath (changes requested):
Could you expand the PR description?  Right now it only mentions `PREWHERE`, but the PR also stops deferring row policies that are over the sorting key, which is arguably the bigger change. 

It would also be great to walk through each case explicitly (row policy over SK with no `PREWHERE`, row policy over SK with `PREWHERE`, row policy on non-SK with `PREWHERE` over SK and the cases that don't change) and describe what the pipeline looked like before vs after for each one. A small table or list would make it much easier to follow than diffing two `EXPLAIN` outputs. 

More importantly, the correctness argument should be written down somewhere (with details about the determinism caveat and guard handling). If something breaks around row policies / `FINAL`, this is the first place someone will look, and "we probably can avoid deferring" won't tell them whether the optimisation was sound or a guess.

src/Processors/QueryPlan/ReadFromMergeTree.cpp
rienath: Please add the fiddle queries as a regression test

rienath (commented):
I resolved the points that were taken care of. Ping me when the rest are addressed, I will take another look @yariks5s

@pufit could you please review this PR too? It's a critical section, would be great to have component expert eyes look at it

tests/queries/0_stateless/03742_apply_row_policy_after_final.sql
yariks5s: They are now in 03743_apply_row_policy_after_final_safety

rienath (approved):
I love the fact that this also incidentally solved row policy problem that we had before. If someone set `SET apply_row_policy_after_final = 1, apply_prewhere_after_final = 0;`, then we had PREWHERE → FINAL → ROW POLICY. PREWHERE was applied before ROW POLICY and we could probe for rows we haven't got access to. You can see example [here](https://fiddle.clickhouse.com/13e0f06f-2b21-4553-b70d-bbdb3fdc394e). But this is solved now! 

```sql
SELECT *
FROM employees
FINAL
PREWHERE throwIf(salary > 200000, 'leak') = 0
```
```
   ┌─dept_id─┬─name──┬─salary─┬─version─┐
1. │       1 │ Alice │  81000 │       2 │
   └─────────┴───────┴────────┴─────────┘
```

I think we should add this fiddle query as a test too so we don't degrade. In any case, looks good!

yariks5s:
Thanks! I will add this test in the follow-up PR.

PR #102894, opened 2026-04-16
Fix short writes in Poco BufferedStreamBuf::flushBuffer

CheSema:
`BufferedStreamBuf::flushBuffer` expected `writeToDevice` to write all bytes in a single call. `SocketImpl::sendBytes` wraps a single `::send` syscall which can legitimately return a short write count. This mismatch caused `flushBuffer` to return -1 on a partial send, which the iostream layer converted into a generic `std::ios_base::failure`, losing the original error context and bypassing S3 retry logic in `PocoHTTPClient`.

The chain is: `ostream::flush()` -> `pubsync()` -> `sync()` -> `flushBuffer()` -> `writeToDevice()` -> `HTTPSession::write()` -> `SocketImpl::sendBytes()` -> `::send()`. A short write from `::send()` propagated back as-is through the entire chain. `flushBuffer` treated any result != buffer size as a hard error, returning -1. The iostream `flush()` then called `setstate(badbit)` -> `clear()` -> threw `std::ios_base::failure`, replacing the original error context.

Stack trace:
```
0. std::system_error::system_error(std::error_code, char const*)
1. std::__throw_failure[abi:ne210105](char const*) @ libcxx/src/ios.cpp:58
2. std::ios_base::clear() @ libcxx/src/ios.cpp:220
3. std::basic_ostream<char, std::char_traits<char>>::flush()
4. DB::EndpointConnectionPool<Poco::Net::HTTPSClientSession>::PooledConnection::flushRequest() @ src/Common/HTTPConnectionPool.cpp
5. Poco::Net::HTTPClientSession::receiveResponse(Poco::Net::HTTPResponse&) @ base/poco/Net/src/HTTPClientSession.cpp
6. DB::EndpointConnectionPool<Poco::Net::HTTPSClientSession>::PooledConnection::receiveResponse(Poco::Net::HTTPResponse&) @ src/Common/HTTPConnectionPool.cpp
7. DB::S3::PocoHTTPClient::makeRequestInternalImpl(...) @ src/IO/S3/PocoHTTPClient.cpp
8. DB::S3::PocoHTTPClient::MakeRequest(...) @ src/IO/S3/PocoHTTPClient.cpp
9. Aws::Client::AWSClient::AttemptExhaustively(...)
10. Aws::Client::AWSXMLClient::MakeRequest(...)
11. Aws::S3::S3Client::HeadObject(...)
12. DB::S3::Client::HeadObject(...)
13. DB::S3::getObjectInfo(...)
```

The fix loops in `flushBuffer` until all bytes are written or a real error (`<= 0`) occurs.

Also throws `MessageException` from `HTTPFixedLengthStreamBuf::writeToDevice` when writing past `Content-Length` instead of silently returning 0.

Fix S3 requests failing with `ios_base::clear: unspecified iostream_category error` instead of being retried, caused by Poco `BufferedStreamBuf::flushBuffer` not handling short writes from the socket layer.

### Version info
- Merged into: `26.5.1.49`
- Backported to: `26.4.1.1118`, `26.3.10.35`, `26.2.17.22`, `26.1.12.18`

### DIFF
--- a/base/poco/Foundation/include/Poco/BufferedBidirectionalStreamBuf.h
+++ b/base/poco/Foundation/include/Poco/BufferedBidirectionalStreamBuf.h
@@ -136,12 +136,16 @@ private:
     int flushBuffer()
     {
         int n = int(this->pptr() - this->pbase());
-        if (writeToDevice(this->pbase(), n) == n)
+        int offset = 0;
+        while (offset < n)
         {
-            this->pbump(-n);
-            return n;
+            int written = writeToDevice(this->pbase() + offset, n - offset);
+            if (written <= 0)
+                return -1;
+            offset += written;
         }
-        return -1;
+        this->pbump(-n);
+        return n;
     }
 
     std::streamsize _bufsize;
--- a/base/poco/Net/src/HTTPFixedLengthStream.cpp
+++ b/base/poco/Net/src/HTTPFixedLengthStream.cpp
@@ -68,11 +68,8 @@ int HTTPFixedLengthStreamBuf::readFromDevice(char* buffer, std::streamsize lengt
 
 int HTTPFixedLengthStreamBuf::writeToDevice(const char* buffer, std::streamsize length)
 {
-	if (_count >= _length)
-		throw MessageException("Write past Content-Length");
-
 	if (_count + length > _length)
-		length = static_cast<std::streamsize>(_length - _count);
+		throw MessageException("Write past Content-Length");
 
 	int n = _session.write(buffer, length);
 	if (n > 0) _count += n;
--- a/base/poco/Net/src/HTTPFixedLengthStream.cpp
+++ b/base/poco/Net/src/HTTPFixedLengthStream.cpp
@@ -68,14 +68,14 @@ int HTTPFixedLengthStreamBuf::readFromDevice(char* buffer, std::streamsize lengt
 
 int HTTPFixedLengthStreamBuf::writeToDevice(const char* buffer, std::streamsize length)
 {
-	int n = 0;
-	if (_count < _length)
-	{
-		if (_count + length > _length)
-			length = static_cast<std::streamsize>(_length - _count);
-		n = _session.write(buffer, length);
-		if (n > 0) _count += n;
-	}
+	if (_count >= _length)
+		throw MessageException("Write past Content-Length");
+
+	if (_count + length > _length)
+		length = static_cast<std::streamsize>(_length - _count);
+
+	int n = _session.write(buffer, length);
+	if (n > 0) _count += n;
 	return n;
 }
 
--- a/base/poco/Net/src/HTTPFixedLengthStream.cpp
+++ b/base/poco/Net/src/HTTPFixedLengthStream.cpp
@@ -68,8 +68,11 @@ int HTTPFixedLengthStreamBuf::readFromDevice(char* buffer, std::streamsize lengt
 
 int HTTPFixedLengthStreamBuf::writeToDevice(const char* buffer, std::streamsize length)
 {
-	if (_count + length > _length)
+	if (_count >= _length)
 		throw MessageException("Write past Content-Length");
+
+	if (_count + length > _length)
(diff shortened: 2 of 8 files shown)


base/poco/Foundation/include/Poco/BufferedBidirectionalStreamBuf.h
```diff
@@ -136,12 +136,16 @@ class BasicBufferedBidirectionalStreamBuf : public std::basic_streambuf<ch, tr>
...
         {
-            this->pbump(-n);
-            return n;
+            int written = writeToDevice(this->pbase() + offset, n - offset);
+            if (written <= 0)
+                return -1;
+            offset += written;
         }
-        return -1;
+        this->pbump(-n);
```
Michicosun, line 147: Should we call pbump after each successful write iteration where n > 0?

base/poco/Net/src/HTTPFixedLengthStream.cpp
```diff
@@ -68,14 +68,14 @@ int HTTPFixedLengthStreamBuf::readFromDevice(char* buffer, std::streamsize lengt
...
-		if (_count + length > _length)
-			length = static_cast<std::streamsize>(_length - _count);
-		n = _session.write(buffer, length);
-		if (n > 0) _count += n;
-	}
+	if (_count >= _length)
+		throw MessageException("Write past Content-Length");
+
+	if (_count + length > _length)
+		length = static_cast<std::streamsize>(_length - _count);
```
Michicosun: Why not throw an exception in this case instead?

base/poco/Foundation/include/Poco/BufferedBidirectionalStreamBuf.h
CheSema: It would be not correct, it moved pptr() to the left, so it shrinks the buffer from the tail.

```
The streambuf put area looks like this:

  pbase()                    pptr()         epptr()
    |                          |               |
    v                          v               v
    [AAAA BBBB CCCC DDDD EEEE]................
    ^--- buffered data -------^ ^-- free ----^

  pptr() - pbase() = how many bytes are buffered (that's n). When the user writes to the stream, pptr() advances forward. When we flush, we need to move pptr() back to pbase() to mark the buffer as empty.

  pbump(-n) does exactly that — moves pptr backward by n, so pptr() == pbase(), buffer is "empty," new writes start from the beginning again.

  Now, the reviewer's suggestion — pbump(-written) after each iteration. Say we wrote 30 out of 100 bytes:

  pbase()          pptr()                    epptr()
    |                |                         |
    v                v                         v
    [AAAA BBBB CCCC DDDD EEEE]................
    ^-- 70 bytes ---^

  The streambuf now says "70 bytes to flush starting at pbase()." But that's bytes 0-69 — the first 70 bytes, which includes the 30 we already sent. The actual remaining data is bytes 30-99, which starts at pbase() + 30. There's
   no way to express that offset with pbump alone — you'd need to memmove the remaining data to the front, which is pointless overhead.

  So pbump(-n) only at the end is correct: either the buffer is fully flushed and reset, or it stays unchanged on error.
```

base/poco/Net/src/HTTPFixedLengthStream.cpp
CheSema: fixed

src/Common/tests/gtest_http_fixed_length_stream.cpp
```diff
@@ -0,0 +1,103 @@
...
+    /// The data goes into the 8KB buffer first. On flush, flushBuffer calls
+    /// writeToDevice which clamps to Content-Length (writes 5 bytes), then the
+    /// loop calls writeToDevice again with the remaining 5 bytes, which throws
+    /// MessageException because _count >= _length.
+    stream.write("0123456789", 10);
+
+    bool got_exception = false;
+    try
+    {
+        stream.flush();
```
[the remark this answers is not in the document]
CheSema: HTTPFixedLengthOutputStream sets badbit in c-tor

Issue #101969, opened 2026-04-07
Flaky test: test_keeper_snapshot_chunked_transfer/test.py::test_recover_from_snapshot_sent_by_old_leader[remote_disk]


Diskein:
_Important: This issue was automatically generated and is used by CI for matching failures. DO NOT modify the body content. DO NOT remove labels._

Test name: test_keeper_snapshot_chunked_transfer/test.py::test_recover_from_snapshot_sent_by_old_leader[remote_disk]
Failure reason: E   Exception
CI report: [Integration tests (amd_tsan, 2/6)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100741&sha=9768ea0b69d2f0f262e72d8b15618ad9cce46439&name_0=PR&name_1=Integration%20tests%20%28amd_tsan%2C%202%2F6%29)

[...]

PR #100300, opened 2026-03-21
Optimize userspace page cache hot path: precompute hash, avoid per-block string copies

alexey-milovidov:
In `CachedInMemoryReadBufferFromFile::readBigAt`, every 1 MB block required constructing a full `PageCacheKey` (two string copies for path and file_version) and computing `SipHash` over ~100 bytes of string data. For a 14.7 GB file this means ~30,000 string allocations and ~15,000 full-string hash computations per scan.

Reorder `PageCacheKey::hash` so path+version are hashed first and offset+size last, enabling incremental hashing via a copyable `SipHash` state. Add `baseHash` (returns partial state) and `hashForBlock` (finalizes with offset+size). Add `PageCache` overloads of `getOrSet`, `get`, and `contains` that accept a precomputed `UInt128` hash, deferring key construction to a lazy factory that only runs on cache miss.

Three hot paths updated:
- `readBigAt`: precompute base hash once, zero string allocations per block on cache hits.
- `nextImpl` lookahead loop: replace `PageCacheKey` copy + `contains` with `hashForBlock` + hash-based `contains`.
- `isContentCached`: use hash-based `get`.

https://github.com/ClickHouse/ClickBench/pull/818/changes

### Changelog category (leave one):
- Performance Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Slightly optimize userspace page cache.

### Version info
- Merged into: `26.5.1.323`

PR #102899, opened 2026-04-16
Fix flaky test_keeper_snapshot_chunked_transfer[remote_disk]

groeneai:
The `test_recover_after_interrupted_transfer[remote_disk]` integration test has been chronically flaky (20+ PRs affected, 6 master failures in 30 days — issues #102040, #101969). PR #102643 added a `sync()` call but failures continued (master failure on 2026-04-15 on commit 118bba8d which is ahead of the fix).

### Root cause

After the node restarts, Keeper may create new local snapshots asynchronously — `serializeSnapshotToDisk` is queued via `snapshots_queue` and runs in a background thread. During the write, a `tmp_snapshot_*` marker file is created temporarily (removed once the snapshot is fully written to disk).

The S3 assertion checked for **all** `tmp_snapshot_*` objects immediately after `sync()`. If a background snapshot flush happened to be in progress, its transient marker was caught as a false positive.

The `local_disk` variant was never flaky because it checks the **specific** tmp file path from the interrupted transfer, ignoring any new markers from background operations.

### Fix

1. Record the specific `tmp_snapshot_*` file names left by the interrupted transfer (before restart)
2. After restart + `sync()`, poll for those **specific** files to disappear (15 s timeout) instead of asserting immediately against all `tmp_snapshot_*` objects
3. New transient markers from background snapshot flushes are correctly ignored

The test still verifies the core behavior: interrupted transfer tmp files must be cleaned up on startup. The 15 s timeout ensures genuine cleanup failures are still caught.

...

### Version info
- Merged into: `26.5.1.251`

### DIFF
--- a/tests/integration/test_keeper_snapshot_chunked_transfer/test.py
+++ b/tests/integration/test_keeper_snapshot_chunked_transfer/test.py
@@ -2,6 +2,8 @@
 import concurrent.futures
 import math
 import os
+import time
+
 import pytest
 
 import helpers.keeper_utils as keeper_utils
@@ -210,6 +212,10 @@ def test_recover_after_interrupted_transfer(started_cluster, nodes):
             "root", prefix=s3_prefix + "tmp_snapshot_"
         ))
         assert tmp_objects, "No tmp_snapshot object in S3 after killing mid-transfer"
+        # Record the specific tmp files left by the interrupted transfer so we can
+        # check that *these* are cleaned up, ignoring any transient tmp_ markers
+        # created by background snapshot flushes after the node restarts.
+        interrupted_tmp_names = {o.object_name for o in tmp_objects}
     else:
         snapshot_dir = "/var/lib/clickhouse/coordination/snapshots"
         tmp_snapshot_path = node_lagging.exec_in_container(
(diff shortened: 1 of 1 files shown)


groeneai:
cc @kssenii — could you review this? This fixes the remaining flakiness in your test_keeper_snapshot_chunked_transfer test after #102643. The root cause is that the S3 assertion checked for *all* tmp_snapshot files (catching transient markers from background snapshot flushes), while the local_disk variant correctly checked only the specific interrupted file.

groeneai:
### Pre-PR Validation (Session cron:clickhouse-ci-task-worker:20260416-161500)

**a) Deterministic repro:** Race condition in CI — after Keeper restart, background snapshot flushes (`serializeSnapshotToDisk` via `snapshots_queue`) temporarily create `tmp_snapshot_*` markers. The immediate S3 assertion catches these transient markers. Confirmed by: (1) CIDB shows failures exclusively on `remote_disk` variant, never `local_disk`; (2) `local_disk` checks specific file path while `remote_disk` checked all `tmp_snapshot_*` — this asymmetry is the root cause.

**b) Root cause causal chain:** `serializeSnapshotToDisk()` creates a `tmp_<name>` marker → writes snapshot data → removes marker. This runs in a background thread. If the test's `list_objects(prefix="tmp_snapshot_")` executes during this brief window, it finds the transient marker → assertion fails. The `local_disk` variant avoids this because `test -f <specific_path>` only checks the interrupted transfer's file, not all tmp files.

**c) Fix matches root cause:** Yes — narrows the S3 assertion to check only the specific interrupted transfer's tmp files (matching `local_disk` behavior) and adds a poll to handle the brief window where a same-index marker from a new transfer exists.

**d) Test intent preserved:** Yes — still verifies interrupted transfer tmp files are cleaned up. 15 s timeout catches genuine cleanup failures.

**e) Both directions:** This is a CI race condition. The fix is logically proven: narrowing the assertion scope removes the false-positive path while preserving the true-positive detection.

**f) Fix is general:** The asymmetry between `remote_disk` (all tmp files) and `local_disk` (specific file) is the root cause. The fix aligns both variants to check specific files, eliminating the class of false positives.

groeneai:
cc @kssenii — could you take a look? This has been waiting for review since 2026-04-16 (16 days).

**Status update (2026-05-02):**

- @alexey-milovidov re-pinged me on PR #100300 today asking to fix this test. The remote_disk flake `tmp_snapshot objects not removed from S3 on startup` is exactly what this PR addresses.
- I just rebased onto current master to pick up your PR #103767 (the kazoo session expiry fix merged 2026-05-01) — without the rebase this PR would have reverted #103767's change to `test_recover_from_snapshot_with_chunked_transfer`. Force-push: `46cb9832 → 7880ce7f`. After rebase the diff is +27/-4 lines, scoped to `test_recover_after_interrupted_transfer` only.
- CI: 49 SUCCESS, 107 SKIPPED, 1 unrelated FAILURE (`Stateless tests (amd_llvm_coverage, AsyncInsert, s3 storage, parallel)` — chronic infra). Mergeable.

**Recap of the fix:**

After `stop_clickhouse(kill=True)` mid-transfer, the assertion checks that the leaked `tmp_snapshot_*` objects in S3 are cleaned up on restart. The previous assertion compared the *entire* `tmp_snapshot_*` prefix listing. The problem: after the lagging node restarts, `KeeperStateMachine` flushes new local snapshots via `snapshots_queue` (`serializeSnapshotToDisk`), which transiently produces fresh `tmp_snapshot_*` markers in S3 with the same/different log indices. These new markers are completely benign but they cause the assertion to fire intermittently.

The fix: snapshot the set of interrupted-transfer tmp filenames at kill time, then poll (15 s) for *those specific objects* to disappear. Background flushes are ignored.

This explains why only `[remote_disk]` was flaky — the local-disk variant already checks one specific path, not a prefix.

CIDB shows the assertion-flake is still active across ~30 unrelated PRs in 30 days and 6 master runs. Two open issues track it: #102040, #101969.

Thanks!

alexey-milovidov (approved):
Looks reasonable.

PR #100735, opened 2026-03-25
*wip* Add experimental AI functions `aiClassify/Extract/GenerateSQL/Translate/GenerateContent/GenerateEmbedding`

rschu1ze:
This is a re-push of https://github.com/ClickHouse/ClickHouse/pull/99579
Courtesy to @melvynator 

Introduces experimental AI functions that call external AI/embedding providers (OpenAI, Anthropic) directly from SQL queries without leaving ClickHouse.

Uses Named Collections for storing provider credentials and query-level settings for operational parameters (like temperature and error handling).

All functions are gated behind the `allow_experimental_ai_functions` setting.

### Changelog category (leave one):
- Experimental Feature

### Changelog entry (a user-readable short description):
Introduces experimental AI functions `aiClassify`, `aiExtract`, `aiGenerateSQL`, `aiTranslate`, `aiGenerateContent`, `aiGenerateEmbedding`, `aiGenerateEmbeddingOrNull` that call an external AI and embedding providers (OpenAI, Anthropic) directly from SQL queries without leaving ClickHouse.

PR #102922, opened 2026-04-17
Adding `aiEmbed` function

george-larionov:
3rd part of https://github.com/ClickHouse/ClickHouse/pull/100735.

Adds `aiEmbed(collection, text[, dimensions])`, an SQL function that returns a text embedding as `Array(Float32)`. Like the existing AI functions it reads its provider, endpoint, model, and API key from a named collection, is gated behind
  `allow_experimental_ai_functions`, and participates in the shared per-query quotas (`ai_function_max_input_tokens_per_query`, `ai_function_max_api_calls_per_query`). The optional `dimensions` argument, when supported by the model (e.g.
  OpenAI's `text-embedding-3-*`), requests a specific output size; otherwise the model's native size is returned. Supported providers are the OpenAI-compatible ones — OpenAI, vLLM, Ollama, HuggingFace TEI; Anthropic has no public embeddings API
  and is intentionally unsupported.

  Unlike the other AI functions, `aiEmbed` **batches** within a single block of rows: up to `ai_function_embedding_max_batch_size` (default `100`) inputs are sent in one HTTP request, then the response is matched back to input rows by `index`.
  This cuts per-row overhead substantially for typical ingestion patterns such as `INSERT INTO docs SELECT id, doc, aiEmbed('creds', doc) FROM …` or a `vector Array(Float32) DEFAULT aiEmbed('creds', doc)` column. NULL and empty-string rows are
  skipped from batches and produce `[]` without any HTTP call.

- Experimental Feature

Add `aiEmbed` function for using LLM APIs to generate embeddings from within ClickHouse.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 075a761e97714a8d34f9321e49ba6127628e48bb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.206` (included in `26.6` and later)
- Backported to: `26.5.5.14`, `26.4.5.91`


src/Functions/aiGenerateEmbedding.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+        UInt64 total_api_calls = 0;
+        UInt64 total_input_tokens = 0;
+
+        for (size_t batch_start = 0; batch_start < unique_texts.size(); batch_start += max_batch_size)
+        {
+            if (quota.checkQuotas())
+                break;
+
+            size_t batch_end = std::min(batch_start + max_batch_size, unique_texts.size());
```
pjhampton, line 177:
```suggestion
            size_t batch_end = std::min<size_t>(batch_start + max_batch_size, unique_texts.size());
```

This build fails on macos - the template deduction fails, but probably is fine on linux

src/Functions/aiGenerateEmbedding.cpp
```diff
@@ -0,0 +1,307 @@
...
+        String api_version = named_collection->getOrDefault<String>("api_version", "");
+
+        if (provider_name.empty())
+            throw Exception(ErrorCodes::BAD_ARGUMENTS, "AI named collection '{}' must have 'provider'", collection_name);
+        if (endpoint.empty())
+            throw Exception(ErrorCodes::BAD_ARGUMENTS, "AI named collection '{}' must have 'endpoint'", collection_name);
+        if (model.empty())
+            throw Exception(ErrorCodes::BAD_ARGUMENTS, "AI named collection '{}' must have 'model'", collection_name);
+        if (api_key.empty())
+            throw Exception(ErrorCodes::BAD_ARGUMENTS, "AI named collection '{}' must have 'api_key'", collection_name);
```
pjhampton, line 123: Instead of throwing an exception, maybe the better approach would be to log a warning? For example: 

```sql
CREATE NAMED COLLECTION qwen3_embed AS
      provider = 'tei',
      endpoint = 'https://qwen-3-embedding-8b.us-east-2.aws.redacted.com/v1/embeddings',
      model = 'qwen3-embedding-8b',
      api_key = 'none';

SET allow_experimental_ai_functions = 1;
SELECT aiGenerateEmbedding('qwen3_embed', 'What is Deep Learning?');
```

`none` is sent as an auth header

tests/queries/0_stateless/03300_ai_functions.sql
```diff
@@ -229,12 +229,90 @@ WHERE name IN (
...
+SELECT name, type FROM system.columns
+    WHERE database = currentDatabase() AND table = '_03300_ret_sql';
+DROP TABLE IF EXISTS _03300_ret_sql;
+
+SELECT '-- aiGenerateSQL: with temperature';
+SELECT count() FROM (SELECT aiGenerateSQL('ai_test', x, 0.1) AS result FROM _03300_input);
+
+-- =============================================================================
+-- 16. Re-disable the setting mid-session
 -- =============================================================================
```
shankar-iyer, line 516: Please add a testcase for populating the vector column alongside the `INSERT`, will be a usecase I think. 
```
CREATE TABLE docs
(
id UInt32,
doc String,
vector Array(Float32) DEFAULT aiGenerateEmbedding('ai_test', doc)
) Engine=MergeTree ORDER BY id;
```
```
INSERT INTO docs (id, doc) SELECT ...
```

alexey-milovidov:
Name it `aiEmbed`, that's better.

src/Functions/aiGenerateEmbedding.cpp
```diff
@@ -0,0 +1,307 @@
...
+    String getName() const override { return name; }
+    bool isVariadic() const override { return true; }
+    size_t getNumberOfArguments() const override { return 0; }
+
+    bool isStateful() const override { return true; }
+    bool isDeterministic() const override { return false; }
+    bool isDeterministicInScopeOfQuery() const override { return false; }
+    bool isSuitableForConstantFolding() const override { return false; }
+    bool isSuitableForShortCircuitArgumentsExecution(const DataTypesWithConstInfo &) const override { return true; }
+    bool useDefaultImplementationForNulls() const override { return true; }
```
rschu1ze: The base class sets `useDefaultImplementationForNulls() = true` and `useDefaultImplementationForConstants() = false` already. The same applies to `src/Functions/FunctionBaseAI.h`.

src/Functions/aiEmbed.cpp
```diff
@@ -0,0 +1,307 @@
...
+    bool useDefaultImplementationForConstants() const override { return false; }
+
+    DataTypePtr getReturnTypeImpl(const ColumnsWithTypeAndName & arguments) const override
+    {
+        FunctionArgumentDescriptors mandatory_args{
+            {"collection", static_cast<FunctionArgumentDescriptor::TypeValidator>(&isString), &isColumnConst, "const String"},
+            {"text", static_cast<FunctionArgumentDescriptor::TypeValidator>(&isString), nullptr, "String"},
+        };
+        FunctionArgumentDescriptors optional_args{
+            {"dimensions", static_cast<FunctionArgumentDescriptor::TypeValidator>(&isNativeUInt), &isColumnConst, "const UInt"},
```
rschu1ze, line 105: I actually always thought that the dimension count depends only on the model, but it seems that some embedding models are less restrictive. Interesting.

src/Functions/aiEmbed.cpp
```diff
@@ -0,0 +1,307 @@
...
+        };
+        validateFunctionArguments(*this, arguments, mandatory_args, optional_args);
+
+        return std::make_shared<DataTypeArray>(std::make_shared<DataTypeFloat32>());
+    }
+
+    ColumnPtr executeImpl(const ColumnsWithTypeAndName & arguments, const DataTypePtr & /*result_type*/, size_t input_rows_count) const override
+    {
+        const auto * collection_const = typeid_cast<const ColumnConst *>(arguments[0].column.get());
```
rschu1ze: Can we add an early out

```cpp
if (input_rows_count == 0)
    return;
```

at the beginning? Iirc, `executeImpl` is sometimes called with no rows. L. 107 and others would then read from invalid data.

src/Functions/aiEmbed.cpp
```diff
@@ -0,0 +1,307 @@
...
+        chassert(collection_const, "First argument must be a constant String (validated by getReturnTypeImpl)");
+        String collection_name = collection_const->getValue<String>();
+        const auto & named_collection = NamedCollectionFactory::instance().get(collection_name);
+
+        String provider_name = named_collection->getOrDefault<String>("provider", "");
```
rschu1ze: `provider_name` --> `provider`

src/Functions/aiEmbed.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+        return std::make_shared<DataTypeArray>(std::make_shared<DataTypeFloat32>());
+    }
+
+    ColumnPtr executeImpl(const ColumnsWithTypeAndName & arguments, const DataTypePtr & /*result_type*/, size_t input_rows_count) const override
+    {
+        const auto * collection_const = typeid_cast<const ColumnConst *>(arguments[0].column.get());
+        chassert(collection_const, "First argument must be a constant String (validated by getReturnTypeImpl)");
+        String collection_name = collection_const->getValue<String>();
+        const auto & named_collection = NamedCollectionFactory::instance().get(collection_name);
```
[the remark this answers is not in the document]
rschu1ze: Yes, this^^.

src/Functions/aiEmbed.cpp
rschu1ze: Consider factorizing l. 110-123 from here and the same code in FunctionBaseAI.cpp into a common method.

src/Functions/aiGenerateEmbedding.cpp
```diff
@@ -0,0 +1,307 @@
...
+            dimensions = dim_const->getUInt(0);
+        }
+
+        const auto & settings = getContext()->getSettingsRef();
+        UInt64 timeout_sec = settings[Setting::ai_function_request_timeout_sec].value;
+        UInt64 max_retries = settings[Setting::ai_function_max_retries].value;
+        UInt64 retry_delay_ms = settings[Setting::ai_function_retry_initial_delay_ms].value;
+        bool throw_on_error = settings[Setting::ai_function_throw_on_error].value;
+        UInt64 max_batch_size = settings[Setting::ai_function_embedding_max_batch_size].value;
+        if (max_batch_size == 0)
```
rschu1ze: Not needed if you change

```cpp
DECLARE(UInt64, ai_function_embedding_max_batch_size, 100, R"(
``` 

in Core/Settings.cpp to

```cpp
DECLARE(NonZeroUInt64, ai_function_embedding_max_batch_size, 100, R"(
```

src/Functions/aiEmbed.cpp
```diff
@@ -0,0 +1,307 @@
...
+            settings[Setting::ai_function_max_output_tokens_per_query].value,
+            settings[Setting::ai_function_max_api_calls_per_query].value,
+            settings[Setting::ai_function_throw_on_quota_exceeded].value);
+
+        auto provider = createAIProvider(provider_name, endpoint, api_key, api_version);
+
+        auto timeouts = ConnectionTimeouts::getHTTPTimeouts(settings, getContext()->getServerSettings());
+        timeouts.receive_timeout = Poco::Timespan(static_cast<int64_t>(timeout_sec) /*s*/, 0 /*us*/);
+
+        /// Deduplicate identical texts within the batch: each unique text is sent once and the result is reused for every row that had it.
```
rschu1ze: The deduplication is doubtful. 1., it is per-chunk. 2. it makes a non-deterministic function deterministic (thought that argument is also a bit doubtful in the case of embeddings generation). I'd honestly say we can rip it out in the first version and do a proper de-dupliation in the new AI background service later if we think it makes sense.

rschu1ze:
> Name it aiEmbed, that's better.

Let's do so as an alias? The name `aiGenerateEmbedding` isn't terrible, ihmo.

rschu1ze:
Could you please also extend the new test added in https://github.com/ClickHouse/ClickHouse/pull/103779/changes for `aiGenerateEmbeddings`?

src/Functions/AI/OpenAIProvider.cpp
```diff
@@ -153,4 +154,115 @@ AIResponse OpenAIProvider::call(const AIRequest & ai_request, const ConnectionTi
...
+    else
+    {
+        Poco::JSON::Array::Ptr input_array = new Poco::JSON::Array;
+        for (const auto & text : ai_embedding_request.inputs)
+            input_array->add(text);
+        root->set("input", input_array);
+    }
+
+    if (ai_embedding_request.dimensions > 0)
+        root->set("dimensions", static_cast<Int64>(ai_embedding_request.dimensions));
```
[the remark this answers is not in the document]
george-larionov: Pathological

alexey-milovidov (dismissed):
Rename it to `aiEmbed`.

src/Functions/aiGenerateEmbedding.cpp
george-larionov: this function doesn't use the base class, since it does it's own deduplication and batching. Perhaps in a future iteration (after implementing a standalone AI Service) we can make it use the same base class.

george-larionov: Btw, the base class sets `useDefaultImplementationForNulls() = false`, not true. This is intentional since the default impl for nulls does not guarantee that the function won't be called with garbage input (which would later be masked out) however for the AI functions this is not acceptible.

george-larionov: Just like in the base class, I will set it to false and make some custom null handling, it is on my local branch now I just have not yet pushed it

src/Functions/aiEmbed.cpp
```diff
@@ -0,0 +1,307 @@
...
+        String api_version = named_collection->getOrDefault<String>("api_version", "");
+
+        if (provider_name.empty())
+            throw Exception(ErrorCodes::BAD_ARGUMENTS, "AI named collection '{}' must have 'provider'", collection_name);
+        if (endpoint.empty())
+            throw Exception(ErrorCodes::BAD_ARGUMENTS, "AI named collection '{}' must have 'endpoint'", collection_name);
+        if (model.empty())
+            throw Exception(ErrorCodes::BAD_ARGUMENTS, "AI named collection '{}' must have 'model'", collection_name);
+        if (api_key.empty())
+            throw Exception(ErrorCodes::BAD_ARGUMENTS, "AI named collection '{}' must have 'api_key'", collection_name);
```
george-larionov: I see, do you think this should be the case for all the AI functions or specifically for embeddings? Are you proposing that we allow something like this:

```
CREATE NAMED COLLECTION qwen3_embed AS
      provider = 'tei',
      endpoint = 'https://qwen-3-embedding-8b.us-east-2.aws.redacted.com/v1/embeddings',
      model = 'qwen3-embedding-8b',
```

tests/queries/0_stateless/03300_ai_functions.sql
george-larionov: i will add it in my integration [tests PR](https://github.com/ClickHouse/ClickHouse/pull/103053)

src/Functions/AI/OpenAIProvider.cpp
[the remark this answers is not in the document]
george-larionov: The user would need to set their embedding dimension to more than 9,223,372,036,854,775,807, in which case there would be an overflow and this negative val would get sent to an AI provider, which they would handle in some way, probably with an error. This is totally fine and not worth adding checks for.

src/Functions/aiEmbed.cpp
george-larionov: provider_name is used as an argument to create `provider` later on (l. 165)

src/Functions/aiEmbed.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+        UInt64 total_api_calls = 0;
+        UInt64 total_input_tokens = 0;
+
+        for (size_t batch_start = 0; batch_start < unique_texts.size(); batch_start += max_batch_size)
+        {
+            if (quota.checkQuotas())
+                break;
+
+            size_t batch_end = std::min(batch_start + max_batch_size, unique_texts.size());
```
george-larionov: should be fixed now

src/Functions/AI/OpenAIProvider.cpp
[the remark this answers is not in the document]
george-larionov: No

src/Functions/aiGenerateEmbedding.cpp
```diff
@@ -0,0 +1,307 @@
...
+        auto provider = createAIProvider(provider_name, endpoint, api_key, api_version);
+
+        auto timeouts = ConnectionTimeouts::getHTTPTimeouts(settings, getContext()->getServerSettings());
+        timeouts.receive_timeout = Poco::Timespan(static_cast<int64_t>(timeout_sec) /*s*/, 0 /*us*/);
+
+        /// Deduplicate identical texts within the batch: each unique text is sent once and the result is reused for every row that had it.
+        std::unordered_map<String, std::vector<size_t>> dedup_map;
+        for (size_t i = 0; i < input_rows_count; ++i)
+        {
+            String text(arguments[1].column->getDataAt(i));
```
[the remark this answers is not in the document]
george-larionov: We don't want to sanitize input to aiEmbed, because this introduces potentially inconsistent embedding results between our implementation and if the user runs the embedding outside clickhouse. If there is an invalid char in the input data, the provider will return an error and the user can fix their data. So it is working as intended.

tests/queries/0_stateless/03300_ai_functions.sql
george-larionov: thanks for this comment, it caught a bug in the AI funcs caused by using a weak instead of strong context ptr

src/Functions/AI/OpenAIProvider.cpp
```diff
@@ -153,4 +154,99 @@ AIResponse OpenAIProvider::call(const AIRequest & ai_request, const ConnectionTi
...
+
+    for (unsigned i = 0; i < data_arr->size(); ++i)
+    {
+        auto item = data_arr->getObject(i);
+        if (!item)
+            throw Exception(ErrorCodes::RECEIVED_ERROR_FROM_REMOTE_IO_SERVER,
+                "AI embedding response 'data[{}]' is not an object", i);
+
+        /// `index` tells us which input this embedding corresponds to. Defaults to `i` when missing (TEI).
+        UInt64 idx = item->optValue<UInt64>("index", i);
```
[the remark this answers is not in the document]
george-larionov: No

pjhampton (approved):

src/Core/Settings.cpp
```diff
@@ -8156,15 +8156,22 @@ If true (default), an AI function call that fails permanently after exhausting a
...
 Maximum total input (prompt) tokens across all AI function API calls in a single query. Tracked cumulatively from provider responses. Note that this limit may be exceeded by one call's worth of input tokens, since the number of input tokens of a call are not known in advance. Set to 0 to disable.
+
+This limit is only enforced for providers that report a `usage` object in their response (OpenAI, Anthropic, vLLM). Providers that omit token usage (notably HuggingFace TEI) cause the counter to stay at 0 — use `ai_function_max_api_calls_per_query` instead to bound such calls.
 )", EXPERIMENTAL) \
     DECLARE(UInt64, ai_function_max_output_tokens_per_query, 500000, R"(
 Maximum total output (completion) tokens across all AI function API calls in a single query. Tracked cumulatively from provider responses. Note that this limit may be exceeded by one call's worth of output tokens, since the number of output tokens of a call are not known in advance. Set to 0 to disable.
+
+This limit is only enforced for providers that report a `usage` object in their response (OpenAI, Anthropic, vLLM). It does not apply to embedding functions, which never produce output tokens.
```
rschu1ze: `It does not apply to embedding functions (notably aiEmbed), which never produce output tokens.`

src/Core/Settings.cpp
```diff
@@ -8156,15 +8156,22 @@ If true (default), an AI function call that fails permanently after exhausting a
...
 )", EXPERIMENTAL) \
     DECLARE(UInt64, ai_function_max_api_calls_per_query, 0, R"(
 Maximum number of HTTP requests that AI functions may dispatch per query. Set to 0 to disable.
 )", EXPERIMENTAL) \
     DECLARE(Bool, ai_function_throw_on_quota_exceeded, true, R"(
 If true (default), exceeding an AI function quota limit (`ai_function_max_input_tokens_per_query`, `ai_function_max_output_tokens_per_query`, or `ai_function_max_api_calls_per_query`) aborts the query with an exception. If false, remaining rows receive the default value for the column type (empty string for String).
+)", EXPERIMENTAL) \
+    DECLARE(NonZeroUInt64, ai_function_embedding_max_batch_size, 100, R"(
```
rschu1ze, line 8230: I suppose this setting will become obsolete again with the planned internal AI service component. That will be fine - settings can be marked as obsolete.

src/Functions/AI/OpenAIProvider.cpp
```diff
@@ -128,17 +124,22 @@ AIResponse OpenAIProvider::call(const AIRequest & ai_request, const ConnectionTi
...
-        if (choice)
-        {
-            auto message = choice->getObject("message");
-            if (message)
-                ai_response.result = message->optValue<String>("content", "");
-            ai_response.finish_reason = choice->optValue<String>("finish_reason", "stop");
-        }
-    }
+    if (!choices || choices->size() == 0)
+        throw Exception(ErrorCodes::RECEIVED_ERROR_FROM_REMOTE_IO_SERVER,
```
rschu1ze: The error code is already used in other situations, it may actually trigger internal false alarms. Could you introduce a new error code `MALFORMED_AI_PROVIDER_RESPONSE` or something like that?

src/Functions/AI/OpenAIProvider.cpp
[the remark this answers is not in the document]
rschu1ze: This thread is comedy.

src/Functions/AI/OpenAIProvider.cpp
```diff
@@ -153,4 +154,108 @@ AIResponse OpenAIProvider::call(const AIRequest & ai_request, const ConnectionTi
...
+            throw Exception(ErrorCodes::RECEIVED_ERROR_FROM_REMOTE_IO_SERVER,
+                "AI embedding response 'data[{}].index' = {} duplicates an earlier entry", i, idx);
+        seen[idx] = true;
+
+        auto embedding_arr = item->getArray("embedding");
+        if (!embedding_arr)
+            throw Exception(ErrorCodes::RECEIVED_ERROR_FROM_REMOTE_IO_SERVER,
+                "AI embedding response 'data[{}].embedding' is missing or not an array", i);
+
+        ai_embedding_response.embeddings[idx].reserve(embedding_arr->size());
```
rschu1ze: `.reserve` isn't wrong but also not needed, the json unpacking and remote communication is way more expensive than anything we could save by one allocation less.

src/Functions/aiEmbed.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+    /// Unlike the other AI funcs, it's ok and actually preferable for
+    /// optimizer to fold calls with same args together
+    bool isDeterministic() const override { return true; }
+    bool isDeterministicInScopeOfQuery() const override { return true; }
+
+    bool isSuitableForConstantFolding() const override { return false; }
+    bool isSuitableForShortCircuitArgumentsExecution(const DataTypesWithConstInfo &) const override { return true; }
+
+    /// Handle Nullable cols explicitly, since setting this to true may call func with arbitrary input values
```
rschu1ze, line 94: > since setting this to true may call func with arbitrary input values

The function would likely be called with random input. The worse problem is that we'd make lots of unnecessary remote calls for rows that are NULL anyways.

rschu1ze (approved):

shankar-iyer (approved):

src/Functions/aiEmbed.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+            AIEmbeddingResponse ai_embedding_response;
+            bool batch_ok = false;
+            for (UInt64 attempt = 0; attempt <= max_retries; ++attempt)
+            {
+                try
+                {
+                    /// update api_calls/quotas before call so failed calls are still added to total
+                    ++total_api_calls;
+                    quota.recordAttempt();
```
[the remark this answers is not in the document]
george-larionov: The quota logic needs to be reworked anyway, since it does not apply per-query but per-chunk. I will do this in a separate PR.

src/Functions/FunctionBaseAI.cpp
```diff
@@ -207,10 +237,13 @@ ColumnPtr FunctionBaseAI::executeImpl(const ColumnsWithTypeAndName & arguments,
                 ai_request.temperature = temperature;
                 ai_request.max_tokens = config.max_tokens;
 
-                auto ai_response = provider->call(ai_request, timeouts);
+                /// update api_calls/quotas before call so failed calls are still added to total
                 ++total_api_calls;
+                quota.recordAttempt();
```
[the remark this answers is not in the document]
george-larionov: The quota logic needs to be reworked anyway, since it does not apply per-query but per-chunk. I will do this in a separate PR.

PR #88339, opened 2025-10-10
fix threads count for inserts

CheSema:
It is bad for performance. We should use `max_threads` for all inserts.

### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
...

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #102928, opened 2026-04-17
Fix SELECT slowdown caused by simple INSERTs holding unused CPU slots

seva-potapov:

Fix SELECT queries being significantly slower when concurrent INSERTs are running. Previously an INSERT pipeline reserved CPU slots up to `max_threads` at query start even when most slots were never used, starving concurrent SELECTs. CPU slot allocation is now demand-driven: the pipeline only requests slots as it actually pushes parallelizable work. Applies both to concurrency control and to the preemptive CPU scheduler for workloads. New server setting `concurrent_threads_lazy_allocation` (default `true`) acts as a rollback lever.

### Details:

#88339 widened INSERT pipelines to use `max_threads` instead of `max_insert_threads` when sizing the number of pipeline threads. Simple single-chain INSERTs then requested `allocate(1, max_threads)` from `ConcurrencyControl`, the scheduler eagerly granted all slots, and the unused grants counted against the global concurrency limit (`concurrent_threads_soft_limit_num` / `concurrent_threads_soft_limit_ratio_to_cores`), starving concurrent SELECTs. The same shape of bug affected queries routed through workload CPU scheduling (`CPULeaseAllocation`).

The fix has three coordinated parts:

**1. `InterpreterInsertQuery` returns to its simple form.** `pipeline.setNumThreads(max_threads)` for all INSERT pipelines, regardless of materialized views. The workaround merged earlier as #102961 (`isViewsInvolved() ? max_threads : max_insert_threads`) is reverted — with lazy allocation in place it is no longer needed to prevent slot-reservation waste.

**2. `ISlotAllocation` gains a `setMax(N)` method.** Allocations are now demand-sized: callers pass a small initial ceiling to `allocate(min, max)` and grow or shrink the ceiling at runtime via `setMax`. Growing the ceiling re-admits the allocation to the scheduler's waiter list (re-enabling further grants); shrinking only caps future grants — it never reclaims already-granted slots.
  - `ConcurrencyControl::Allocation` implements `setMax` for the `round_robin`, `fair_round_robin`, and `max_min_fair` schedulers. `max_min_fair` preserves its sort-by-allocated invariant by routing grants and re-inserts through intrusive-set helpers.
  - `CPULeaseAllocation` implements `setMax` via a new mutable `current_max_slots` field that gates `schedule()`. The constructor accepts an optional `initial_max_slots` argument; the natural grant chain in `grantImpl` fills out the rest one request at a time after `setMax(N)`.
  - `CPUSlotsAllocation` is unchanged (the non-preemptive workload allocation; the default `setMax` no-op preserves its current behavior).

**3. `PipelineExecutor` drives `setMax` from observed demand.** The previous design used eager `allocate(1, max_threads)` and let the spawning loop decide how many threads to actually start, leaving granted-but-unused slots pinned to the allocation. The new design starts at `master_threads` (= 1) and the upscaling block in `executeStepImpl` raises the ceiling whenever `pushTasks` reports that newly-queued tasks would benefit from additional parallelism. `ExecutorTasks::pushTasks` now returns a `size_t` "additional threads to target" and `upscale` returns remaining-capacity rather than a `SpawnStatus` enum (the `SHOULD_SPAWN` / `DO_NOT_SPAWN` enum is removed).

The server setting `concurrent_threads_lazy_allocation` (default `true`) toggles `PipelineExecutor` between the lazy and eager paths. It is `ChangeableWithoutRestart::Yes`.

### Tests:

- `gtest_concurrency_control` — eleven new tests covering lazy + `setMax` behaviour, eager-allocation backward compatibility, grow/shrink semantics, work-conserving max-min-fair, and a concurrent stress test.
- `gtest_workload_resource_manager` — five new tests covering `CPULeaseAllocation` lazy initialisation, `setMax` growth, `max_threads` clamping, shrink-doesn't-reclaim, and eager-default backwards compatibility.
- `04102_insert_threads_eagerly_spawned` — existing test from #102961; with-MV branch tightened to upper-bound `peak_threads_usage` (lazy CC correctly produces `peak = 1` for trivial MV pipelines now).
- `04229_lazy_concurrency_control_insert_with_mv` — stateless test asserting `peak_threads_usage <= 4` for the INSERT-with-MV `parallel_view_processing=0` case routed through `ConcurrencyControl`.
- `04230_lazy_cpu_lease_allocation_insert_with_mv` — same shape as 04229 but routed through `CPULeaseAllocation` via `CREATE RESOURCE … (MASTER THREAD, WORKER THREAD)` and `CREATE WORKLOAD`.

### Related:

- Related: https://github.com/ClickHouse/ClickHouse/issues/102947 — underlying issue (closed after #102961 workaround; this PR completes the fix for the MV / workload case)
- Related: https://github.com/ClickHouse/ClickHouse/pull/102961 — interim workaround merged separately
- Related: https://github.com/ClickHouse/ClickHouse/pull/88339 — introduced the regression

### Version info
- Merged into: `26.6.1.1108`


src/Common/ConcurrencyControl.h
```diff
@@ -23,6 +23,19 @@ namespace DB
...
+ *    queries, not the full waiter list.
+ *  - `demanders` membership flips via `setMoreDemand`. An allocation enters `demanders` at
+ *    creation (default demand=true) and leaves on `setMoreDemand(false)`, on losing waiter
+ *    status (allocated reached limit), or on destruction. It re-enters on `setMoreDemand(true)`.
+ *  - All membership changes (add/remove from `waiters`, `demanders`) happen under `state.mutex`.
+ *  - MMF uses `allocated` as the sort key of both intrusive sets — `grant()` must not run while
+ *    the allocation is linked in either set (remove → grant → re-insert).
+ *  - `setMoreDemand` MUST NOT be called while holding `state.mutex` — it takes that lock itself.
+ *
  * USAGE:
```
serxa, line 38: Usage pattern will be a little bit different, let us update this comment as well

src/Common/ConcurrencyControl.h
```diff
@@ -125,8 +159,15 @@ class ConcurrencyControlRoundRobinScheduler
...
-
-        const Waiters::iterator waiter; // iterator to itself in Waiters list; valid iff allocated < limit
+        std::atomic<bool> has_more_demand{true}; // consumer wants more slots granted (see setMoreDemand)
+
+        // Iterator to self in Waiters list; valid iff allocated < limit.
+        // Non-const so setMoreDemand(false) can re-insert after a previously-saturated
+        // allocation has its budget restored by reclaiming pending grants.
+        Waiters::iterator waiter;
+        // Iterator to self in Demanders list; valid iff `in_demanders` is true.
+        Demanders::iterator demander_iter;
```
serxa: Do we actually need two lists here? Maybe we can go with just a single demander list and only use a writer for this. Because scheduling is always done for the demanding allocations, there is no point in having a waiters' list.

Then, if we use a single list, there is no need to rename waiters into demanders. Instead, we can just change the semantics of waiters.

src/Common/ISlotControl.h
```diff
@@ -90,6 +90,25 @@ class ISlotAllocation : public std::enable_shared_from_this<ISlotAllocation>, bo
     /// Take one granted slot or wait until it is available.
     [[nodiscard]] virtual AcquiredSlotPtr acquire() = 0;
 
+    /// Signal whether the consumer currently wants more slots granted.
+    /// Default state (after allocate) is true — the allocation wants slots.
+    /// The consumer (e.g. PipelineExecutor) calls setMoreDemand(false) when it has reached
+    /// its target parallelism and does not want additional slots granted, and setMoreDemand(true)
+    /// when it wants to upscale. The scheduler uses this signal to avoid granting slots to
+    /// allocations that would not acquire them, freeing capacity for other consumers.
```
serxa: I think this is not the best design because it is inherently race-prone. Imagine if, at some point, an allocation requires two slots, but not three. To achieve granting exactly 2 slots, it would be required to call setMoreDemand(false) exactly after two slots were granted, but before the 3rd slot is granted.

I think a more natural evolution of the existing design is that allocation can change its maximum, which is now constant. This will greatly simplify all the changes because there is no need to distinguish between waiting and demanding allocations. Whenever this maximum is increased, one of two things should happen:
1) If the maximum has not been reached yet, the allocation is already in the waiting list, and we only need to update the maximum
2) otherwise, allocation was already granted, all previously requested slots and it is not in a waiting list any longer; in this case, we should increase the maximum and also put the allocation into the waiting list again

src/Core/ServerSettings.cpp
```diff
@@ -846,6 +846,13 @@ namespace
     :::
     )", 0) \
     DECLARE(UInt64, concurrent_threads_soft_limit_ratio_to_cores, 0, "Same as [`concurrent_threads_soft_limit_num`](#concurrent_threads_soft_limit_num), but with ratio to cores.", 0) \
+    DECLARE(Bool, concurrent_threads_lazy_allocation, true, R"(
+When true (default), ConcurrencyControl uses lazy slot allocation: a query is granted its minimum slots plus at most one extra slot at allocation time, with the rest granted on-demand as threads actually acquire them. This avoids situations where a query reserves many CPU slots it never uses (starving other queries).
+
+When false, the server falls back to the pre-#88339 eager allocation behavior — grant everything that fits in available capacity immediately. This is intended as an emergency rollback lever only: use it if a specific workload regresses on the lazy strategy.
```
serxa: There is no need to mention from PRs in the description

serxa (changes requested):

src/Common/ConcurrencyControl.h
```diff
@@ -140,15 +162,28 @@ class ConcurrencyControlRoundRobinScheduler
...
     SlotAllocationPtr allocate(std::unique_lock<std::mutex> & lock, SlotCount min, SlotCount max);
 
-    // Round-robin scheduling of available slots among waiting allocations
-    void schedule(std::unique_lock<std::mutex> &);
+    // Round-robin scheduling of available slots among waiting allocations.
+    // Bulk mode (lazy_grant=false): distributes ALL available capacity fairly (used by setMaxConcurrency).
+    // Lazy multi (lazy_grant=true, single_grant=false): grant to waiters with granted==0; stops on first
+    //   full round where no grant happens. Used by release/free/setMax to redistribute capacity.
+    // Lazy single (lazy_grant=true, single_grant=true): grant at most 1 slot to 1 waiter. Used by
+    //   notifyAcquired for one-at-a-time ramp-up.
```
serxa: I don't think we need this mode. And acquire notifications are not necessary any longer, as it is always safe to grant slot up to current limit

src/Core/ServerSettings.cpp
```diff
@@ -846,6 +846,15 @@ namespace
     :::
     )", 0) \
     DECLARE(UInt64, concurrent_threads_soft_limit_ratio_to_cores, 0, "Same as [`concurrent_threads_soft_limit_num`](#concurrent_threads_soft_limit_num), but with ratio to cores.", 0) \
+    DECLARE(Bool, concurrent_threads_lazy_allocation, true, R"(
+Rollback lever for the lazy query-thread allocation strategy.
+
+When true (default), a query pipeline starts with `allocate(1, 1)` and raises the ceiling via `setMax` as it actually pushes parallelizable work. This prevents queries from reserving CPU slots they never use.
+
+When false, the server bypasses the lazy path: pipelines use `allocate(1, num_threads)` and the scheduler grants up to `num_threads` slots eagerly (pre-#88339 behavior). Intended as an emergency rollback lever only — use it if a specific workload regresses on the lazy strategy.
```
serxa: There is no need to mention the PRs in the user-facing description. The same is true about some comments

serxa (changes requested):

serxa:
There is another important implementation of ISlotControl interface that we changed in this PR. This is CPULeaseAllocation, which is used for workload CPU scheduling. And it would be nice to fix it as well. At least we should make sure our interface changes won't lead to any contradictions. Let me try to iterate on it with Claude...

serxa:
@CheSema what you think about reverting https://github.com/ClickHouse/ClickHouse/pull/102961 ? This PR is basically the lazy slot allocation implementation and your PR says:

> Until lazy CC slot allocation is implemented, we have no choice other than to set strict limits on the thread count for insert queries.

~~But TBH I'm not sure if the revert would be 100% safe. What are your thought on this? Should we keep your PR or revert inside this PR?~~ UPD. We decided to revert it, I checked that tests are good now.

serxa:
Verified that lazy allocation alone is sufficient to bound `peak_threads_usage` correctly even with the workaround reverted to `pipeline.setNumThreads(max_threads)`. All related stateless tests (01275_parallel_mv, 04102, 04229, 04230, 02350, 02871, 03562) still pass. We will rollback #102961 as part of this PR.

serxa (approved):

PR #86633, opened 2025-09-03
Add support for user names in config containing a dot

mkmkme:
For correct config parsing, the dots in the user names have been escaped for some time [1]. This caused a confusion for some users. This commit adds the "un-escaping" to the user name handling.

Fixes #82026

[1] https://github.com/ClickHouse/ClickHouse/pull/58354

### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fixes handling of users with a dot in the name when added via config file.

### Documentation entry for user-facing changes

- [X] Documentation is written (mandatory for new features)

PR #102962, opened 2026-04-17
Un-escape dots in role, quota, and profile names from users.xml

### LINKED ISSUE #102527: `parseRole`, `parseQuota`, `parseSettingsProfile` missing dot un-escaping that `parseUser` got in this PR

_Retrospective finding from a historical scan of [PR #86633](https://github.com/ClickHouse/ClickHouse/pull/86633) (merged 2025-09-09). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

Config-defined roles, settings profiles, and quotas with dots in their XML element names (e.g., `<my.role>`) show escaped names with backslash (e.g., `my\.role`) in `system.roles`, `system.settings_profiles`, and `system.quotas`. Users were fixed by this PR but the sibling entity types were not.

**Root cause:** UsersConfigParser.cpp:620,643,759: `setName` called with Poco-escaped name (containing `\.`) instead of un-escaped name. The `Poco::replaceInPlace` fix from this PR was only applied to `parseUser`, not to these three functions.

**Why we believe this is a bug:** `parseUser` (UsersConfigParser.cpp:416) calls `Poco::replaceInPlace(user_name, "\\.", ".")` before `user->setName`. But `parseRole` (line 620), `parseQuota` (line 643), and `parseSettingsProfile` (line 759) all call `setName` directly with the Poco-escaped name containing `\.`.

**Affected locations:**
- `src/Access/UsersConfigParser.cpp:620` — parseRole: role->setName(role_name) without un-escaping dots
- `src/Access/UsersConfigParser.cpp:643` — parseQuota: quota->setName(quota_name) without un-escaping dots
- `src/Access/UsersConfigParser.cpp:759` — parseSettingsProfile: profile->setName(profile_name) without un-escaping dots

**Impact:** Config-defined roles, profiles, and quotas with dots in their XML names display incorrectly with backslash escaping in all system tables and SHOW commands. Users cannot reference these entities by their intended names.

Yes — confirmed on current `master` (commit `30287bb9edc2`).

### How to reproduce

```python
import pytest

from helpers.cluster import ClickHouseCluster

cluster = ClickHouseCluster(__file__)
node = cluster.add_instance(
    "node",
    user_configs=[
        "configs/users.xml",
    ],
)


@pytest.fixture(scope="module", autouse=True)
def started_cluster():
    try:
        cluster.start()
        yield cluster
    finally:
        cluster.shutdown()


def test_user_with_dot_is_unescaped():
    """Verify that PR #86633's user fix works - user with dot shows correctly."""
    result = node.query("SELECT name FROM system.users WHERE name LIKE '%dotted%' ORDER BY name")
    assert "user.dotted" in result


def test_role_with_dot_shows_escaped_name():
    """BUG: Config-defined role with a dot shows escaped name."""
    result = node.query("SELECT name FROM system.roles WHERE storage = 'users_xml' ORDER BY name")
    role_lines = [l.strip() for l in result.strip().split('\n')]
    assert "my.role" in role_lines, f"Role 'my.role' not found (got escaped name): {role_lines}"


def test_profile_with_dot_shows_escaped_name():
    """BUG: Config-defined profile with a dot shows escaped name."""
    result = node.query("SELECT name FROM system.settings_profiles WHERE storage = 'users_xml' ORDER BY name")
    profile_lines = [l.strip() for l in result.strip().split('\n')]
    assert "my.profile" in profile_lines, f"Profile 'my.profile' not found (got escaped name): {profile_lines}"


def test_quota_with_dot_shows_escaped_name():
    """BUG: Config-defined quota with a dot shows escaped name."""
    result = node.query("SELECT name FROM system.quotas WHERE storage = 'users_xml' ORDER BY name")
    quota_lines = [l.strip() for l in result.strip().split('\n')]
    assert "my.quota" in quota_lines, f"Quota 'my.quota' not found (got escaped name): {quota_lines}"
```

### Expected behavior

```
Role: ['my.role'], Profile: ['default', 'my.profile', 'readonly'], Quota: ['default', 'my.quota'] — all should show unescaped dot
```

```
Role: ['my\\.role'], Profile: ['default', 'my\\.profile', 'readonly'], Quota: ['default', 'my\\.quota'] — all show escaped backslash before dot
```

### Additional context

**Open risks:**
- parseRowPolicies (line 985) uses `generateID(AccessEntityType::USER, user_name)` with the escaped name, potentially creating UUID mismatches with the user entity that now uses un-escaped names
- parseQuotas (line 882-884) maps quotas to users using escaped user names for UUID generation, which may not match the actual user UUIDs

**Suggested fix:** Apply the same `Poco::replaceInPlace(name, "\\.", ".")` pattern from `parseUser` to `parseRole`, `parseQuota`, and `parseSettingsProfile`. Each function should: (1) take the name parameter by value, (2) construct the config path before un-escaping, (3) call `Poco::replaceInPlace` to remove backslash escaping, (4) then call `setName` with the un-escaped name.

**Analysis details:** Confidence HIGH | Severity P2 | Testability: `INTEGRATION_TEST`

Found during automated review of [PR #86633](https://github.com/ClickHouse/ClickHouse/pull/86633).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr86633_001`_

mkmkme:
PR #86633 added backslash un-escaping for user names parsed from XML config keys (`<user.name>` was being stored as `user\.name`), but the same fix was not applied to the sibling access entity parsers. As a result config-defined roles, quotas, and settings profiles with dots in their XML element names still showed up with escaped backslashes in `system.roles`, `system.quotas`, and `system.settings_profiles`.

More importantly, because `parseUser` now stores the un-escaped name while `generateID` is deterministic over the stored name, the UUIDs referenced from `parseQuotas` (user-to-quota mapping) and `parseRowPolicies` (policy-to-user mapping) were computed from the escaped names and therefore no longer matched the user entity. The same mismatch also affected `getAllowedIDs`, which is used to validate that the profile and role IDs a user references actually exist.

Apply the `Poco::replaceInPlace` pattern from `parseUser` to `parseRole`, `parseQuota`, and `parseSettingsProfile`, and un-escape the user and entity names everywhere a UUID is generated from a config key so all references line up.

Fixes #102527

- Improvement

The extra backslashes in role, quota, and profile names from users.xml containing a dot were removed


mkmkme:
@azat @UnamedRus could you have a look at this?

azat (commented):
Looks good in general, but let's introduce a helper for this

azat:
@mkmkme can you take a look? Or if you are busy, I can finish this

mkmkme:
> @mkmkme can you take a look? Or if you are busy, I can finish this

I am out this week. I can take a look next week if it's not urgent.

tests/integration/test_dot_in_user_name/test.py
```diff
@@ -33,3 +33,55 @@ def test_user_with_dot_in_name():
...
+    # The quota-to-user mapping is built from the user's ID; it must match the entity's un-escaped name.
+    assert node.query(
+        "SELECT count()>0 FROM system.quota_usage "
+        "WHERE quota_name = 'my.quota' AND quota_key = 'dotted.user'",
+        user="dotted.user",
+    ) == "1\n"
+
+    # The role ID must also be derived from the un-escaped name for the grant to take effect.
+    assert node.query(
+        "SELECT count()>0 FROM system.current_roles WHERE role_name = 'my.role'",
```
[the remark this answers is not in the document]
mkmkme: Done

mkmkme:
@azat I've done the changes. Could you please have another look?

src/Access/UsersConfigParser.cpp
```diff
@@ -881,7 +887,13 @@ std::vector<AccessEntityPtr> UsersConfigParser::parseQuotas(const Poco::Util::Ab
     for (const auto & user_name : user_names)
     {
         if (config.has("users." + user_name + ".quota"))
-            quota_to_user_ids[config.getString("users." + user_name + ".quota")].push_back(generateID(AccessEntityType::USER, user_name));
+        {
+            /// The user name comes from a config key, so any dot in the name is escaped with a backslash.
+            /// The user entity is stored under the un-escaped name, so we must generate its ID from the un-escaped name too.
+            String unescaped_user_name = user_name;
+            Poco::replaceInPlace(unescaped_user_name, "\\.", ".");
```
azat: Why this does not use helper? I think we can provide more general helper like "unescapeDots()" and "escapeDots"()" (for analogy) and use it everywhere here

mkmkme: Oops, sorry, this one is a leftover. I'll fix it now!

mkmkme: Replaced the helper with `unescapeDots()` and used it everywhere. I don't think `escapeDots` is necessary here since we don't really escape the dots anywhere but in Poco codebase.

src/Access/UsersConfigParser.cpp
```diff
@@ -398,6 +398,11 @@ namespace
         }
     }
 
+    String unescapeDots(const String & name)
```
azat: Let's also add a decent comment for this function, including original place where this dots are escaped

mkmkme: 8022b576ac9b6222dbaecefd65ce515032278378

src/Access/UsersConfigParser.cpp
```diff
@@ -1025,7 +1038,10 @@ std::unordered_set<UUID> UsersConfigParser::getAllowedIDs(
     config.keys(configuration_key, keys);
     std::unordered_set<UUID> ids;
     for (const auto & key : keys)
-        ids.emplace(generateID(type, key));
+    {
+        /// Config keys have dots escaped with a backslash, but entity UUIDs are generated from the un-escaped name.
+        ids.emplace(generateID(type, unescapeDots(key)));
```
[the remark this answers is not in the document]
mkmkme: Should I cover this or should we treat the old behaviour as incorrect and hence dismiss this one?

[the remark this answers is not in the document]
azat: Since it is a tiny change to preserve backward compatibility, I would add both here
Also then let's also add a test case, and make sure that this is the only place (if not maybe we can add similar helper there)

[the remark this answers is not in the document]
mkmkme: Added in https://github.com/ClickHouse/ClickHouse/pull/102962/commits/dc92d6ffd39836944b4a72db04e95b2b982784c8

azat (approved):

mkmkme:
Hey, I just noticed you enabled auto-merge two weeks ago on this, but the CH Inc sync keeps on pending even after you updated the branch. Is there anything I could do to help with this?

azat:
Hi @mkmkme, private CI was in a bad shape, I fixed the conflicts once but need to do it one more time, I will land this change soon

mkmkme:
Hey @azat, I just wanted to ask if there's anything else needed from my side here?

PR #102963, opened 2026-04-17
PromQL: Add aggregation operators limitk, topk, bottomk

vitlibar:

PromQL: Add aggregation operators `limitk`, `topk`, `bottomk`.

### Version info
- Merged into: `26.4.1.1088`


src/Functions/TimeSeries/timeSeriesGroupToSamplingKey.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+
+namespace ErrorCodes
+{
+    extern const int NUMBER_OF_ARGUMENTS_DOESNT_MATCH;
+}
+
+/// Function timeSeriesGroupToSamplingKey(group) returns a stable UInt64 hash derived from the tags
+/// of a specified group. The value is intended as a deterministic sort key for sampling operations
+/// like `limitk` and `limit_ratio`, where Prometheus requires a "deterministic pseudo-random" order.
```
nikitamikhaylov, line 20: Are we concerned about collisions here? We've chosen UInt128 as a series id for this reason.

vitlibar: We're concerned but not too much. The cost is different. For series id a collision means our data are not in order anymore (and we can't figure out which pair `(timestamp, value)` belongs to which series). And for this sampling key a collision means just a lack of determinism of `limitk()` function. I suppose a tiny chance of that `limitk()` picks different series next time is kind of OK.

nikitamikhaylov (approved):

PR #102964, opened 2026-04-17
Fix backupview behavior

boomeer:
### Changelog category
- Improvement

### Changelog entry
clickhouse_backupview improvements
**1. `__split_database_table` - missing return for single-name string.** Now raises an explicit `Exception` with a message explaining the three accepted forms (tuple, `"db.tbl"`, or explicit `database=`). The silent fall-through that caused a cryptic `TypeError` on tuple-unpack at the call site is gone.

**2. `__parse_location` - argument splitter breaks on commas inside values.** Replaced the naive `split(",")` with a new `__split_args` helper that walks the string character-by-character and suspends comma-splitting inside single- or double-quoted substrings. Also switched to `rfind(")")` for the closing paren so paths with parens inside quotes are less fragile. `S3('https://host?x=1,2', 'AKIA', 'sec,ret')` now round-trips correctly.

**3. `S3URI` - `s3://bucket/key` scheme not handled.** Added `__parse_s3_scheme` as the first parser tried; bucket comes from `netloc`, key from `path`, endpoint is left as `None` so boto3 falls back to its default AWS S3 endpoint (matching AWS CLI semantics). The virtual-hosted and path-style parsers now also gate on `scheme in ("http", "https")` so they won't accidentally swallow `s3://` URIs.

**4. `S3URI.__parse_virtual_hosted` - bucket name containing `.s3` substring is truncated.** Replaced the first-match `host.split(".s3", maxsplit=1)` with a regex (`^(.+?)\.(s3(?:[.-].+|$))$`) that requires the `s3` token to be followed by `.`, `-`, or end-of-host. `my.s3bucket.s3.amazonaws.com` now correctly yields bucket `my.s3bucket` and endpoint `https://s3.amazonaws.com`; `my.s3fake.com` no longer matches at all and falls through to path-style.

**5. `get_table_data_files` error message.** Changed `only_part`/`only_partition` to `part`/`partition` to match the actual parameter names.

**6. `check_files` - size-only verification.** Added optional `verify_checksums=False` and `checksum_factory=None` parameters. When `verify_checksums=True`, a hashlib-style factory is required (since ClickHouse uses SipHash128, which isn't in the stdlib - caller supplies a compatible implementation). The new `__verify_checksums` helper streams each file through `open_file()` in 4 MiB chunks and compares the hex digest against the stored checksum, correctly handling incremental files via `ConcatFileObj`. The synthetic `/.backup` entry and any files with a zero-checksum sentinel are skipped. Default behavior is unchanged.

**7. `read_file` on incremental base - full memory duplication.** The incremental branch now does `with self.open_file(path) as stream: return stream.read()`. The old `base.read_file() + self.__reader.read_file()` pattern - which held the base bytes, the delta bytes, and the concatenated result in memory simultaneously - is gone. The non-incremental branches (zero size, full base, no base) are unchanged.

**8. `ConcatFileObj.read(count)` - extra round-trip at boundary.** `ConcatFileObj.__init__` now takes an optional `size1` (total bytes in `fileobj1`). `open_file` passes `fi.base_size`. When the running byte count hits `size1` exactly, the flag flips immediately - no probe-read on an exhausted stream. The old behavior is preserved when `size1` is omitted.

### DIFF
--- a/utils/backupview/clickhouse_backupview.py
+++ b/utils/backupview/clickhouse_backupview.py
@@ -2,6 +2,7 @@
 
 import bisect
 import os.path
+import re
 import shutil
 import xml.etree.ElementTree as ET
 import zipfile  # For reading backups from zip archives
@@ -315,7 +316,7 @@ class Backup:
             return []
         if (part is not None) and (partition is not None):
             raise Exception(
-                "get_table_data_files: `only_part` and `only_partition` cannot be set together"
+                "get_table_data_files: `part` and `partition` cannot be set together"
             )
         files = []
         if part is not None:
@@ -425,13 +426,13 @@ class Backup:
         )
 
     # Returns the size of a file in the backup.
-    # The function raises an exception of the file doesn't exist.
+    # The function raises an exception if the file doesn't exist.
     def get_file_size(self, path):
         fi = self.get_file_info(path)
         return fi.size
 
     # Returns the information about a file in the backup.
-    # The function raises an exception of the file doesn't exist.
+    # The function raises an exception if the file doesn't exist.
     def get_file_info(self, path):
         if not path.startswith("/"):
             path = "/" + path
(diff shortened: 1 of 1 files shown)


boomeer:
@nikitamikhaylov @alexey-milovidov pls add label `can be tested`

Issue #1639, opened 2017-12-14
compile error: /host/contrib/libcityhash/src/city.cc:261:3: error: 'ssize_t' was not declared in this scope


wenerme:
```bash
git clone -b stable --depth 1 --recursive https://github.com/yandex/ClickHouse.git
cd ClickHouse

docker run --rm -it -v $PWD:/host --workdir /host wener/base:builder
mkdir build
cd build
apk add libressl-dev
cmake ..
make -j $(nproc)
```

```
/host/contrib/libcityhash/src/city.cc: In function 'CityHash_v1_0_2::uint128 CityHash_v1_0_2::CityMurmur(const char*, size_t, CityHash_v1_0_2::uint128)':
/host/contrib/libcityhash/src/city.cc:261:3: error: 'ssize_t' was not declared in this scope
   ssize_t l = len - 16;
   ^~~~~~~
/host/contrib/libcityhash/src/city.cc:262:7: error: 'l' was not declared in this scope
   if (l <= 0) {  // len <= 16

```

Image wener/base:builder is build by https://github.com/wenerme/dockerfiles/blob/master/base/builder/Dockerfile

wenerme:
I'm not sure is stable is stable, I will update this issues to include the compile error that I found.

wenerme:
/host/contrib/libcityhash/src/city.cc:261:3: error: 'ssize_t' was not declared in this scope

ssize_t -> size_t

----

```
/host/contrib/libcctz/src/time_zone_info.cc:64:20: error: invalid conversion from 'int' to 'char*' [-fpermissive]
   return strerror_r(errnum, buf, buflen);
```

change to 
```c
strerror_r(errnum, buf, buflen);
return buf;
```

----

```
/host/contrib/libtcmalloc/src/malloc_hook_mmap_linux.h:59:39: error: '__off64_t' has not been declared
                               int fd, __off64_t offset) __THROW {
                                       ^~~~~~~~~
/host/contrib/libtcmalloc/src/malloc_hook_mmap_linux.h:141:24: error: '__off64_t' has not been declared
                int fd, __off64_t offset  ) __THROW
```

https://github.com/gperftools/gperftools/issues/693

Add `#define __off64_t off64_t` to /host/contrib/libtcmalloc/src/malloc_hook_mmap_linux.h after include.

guard mmap64 by `#ifndef mmap64`

----


```
[...]

PR #102975, opened 2026-04-17
Use `INCORRECT_DATA` instead of logical error when client side data encoding is incorrect

antaljanosbenjamin:

Throw `INCORRECT_DATA` instead of `LOGICAL_ERROR` when client-supplied data in the Native format is truncated or malformed.

### Version info
- Merged into: `26.5.1.931`
- Backported to: `26.5.1.881`, `26.4.4.11`, `26.3.13.8`


antaljanosbenjamin:
AI summary of the changes:

# Use INCORRECT_DATA instead of LOGICAL_ERROR for data validation in Native format deserialization

## Context

What happened:
a client sent truncated INSERT data over TCP in Native format. The Variant column's
discriminator sub-stream was fully read (28,040 entries pointing to Float64), but the
Float64 data sub-stream was truncated to only 16,337 values. The validation check in
`SerializationVariant.cpp:661` correctly caught the mismatch, but threw `LOGICAL_ERROR`.

The same file already had the correct pattern 30 lines earlier at line 631:
```cpp
throw Exception(settings.native_format ? ErrorCodes::INCORRECT_DATA : ErrorCodes::LOGICAL_ERROR, ...);
```

This uses `INCORRECT_DATA` when the error is triggered during Native format deserialization
(i.e. client-supplied data), and `LOGICAL_ERROR` when triggered during MergeTree
internal deserialization (where a data integrity issue truly is an internal bug).

We audited all serialization files for the same problem. Below is the reasoning for each change.

## Changes

### SerializationVariant.cpp:661

**Function**: `deserializeBinaryBulkWithMultipleStreams`
**Check**: `col.getVariantByLocalDiscriminator(i).size() < variant_limits[i]`
**Why change**: The discriminator sub-stream says N rows should be Float64, but only M < N
values were deserialized. When this happens during Native format INSERT (truncated TCP
stream), it's bad client data, not a server bug. This is the exact error from incident #1639.

### SerializationString.cpp:814

**Function**: `deserializeBinaryBulkWithSizeStream`
**Check**: `!stream` (missing data stream after sizes stream was present)
**Why change**: The sizes sub-stream existed and was read, but the corresponding data
sub-stream is missing. In Native format, this means the client sent a partial/malformed
block — the sizes arrived but the string data didn't. In MergeTree, a missing stream
when another related stream exists is a true internal error.

### SerializationLowCardinality.cpp:557,560

**Function**: `deserializeBinaryBulkWithMultipleStreams`
**Check**: One of keys/indexes stream is present but the other is missing
**Why change**: LowCardinality requires both keys and indexes streams. If one exists but
not the other, the data is malformed. In Native format this means the client sent an
incomplete LowCardinality column. Note: the same checks at lines 454/457 in the
**serialize** method are left as LOGICAL_ERROR since those are internal.

### SerializationArray.cpp:476

**Function**: `deserializeOffsetsBinaryBulkAndGetNestedOffsetAndLimit`
**Check**: `last_offset < prev_last_offset` (offsets go backwards)
**Why change**: After deserializing array offsets, the last offset should be >= the
previous last offset (arrays can't have negative sizes). A truncated or corrupted offset
stream in Native format could produce backward-going offsets. In MergeTree, this would
indicate a corrupted data part.

### SerializationVariantElement.cpp:267

**Function**: `deserializeBinaryBulkWithMultipleStreams`
**Check**: `variant_element_state->variant->size() < *variant_limit`
**Why change**: Same pattern as the Variant fix — after deserializing a variant sub-column,
fewer rows were produced than the discriminators indicated. Truncated client data in
Native format is the most likely cause.

### SerializationSparse.cpp:418

**Function**: `deserializeBinaryBulkWithMultipleStreams`
**Check**: `offsets_column->size() + 1 != values_column->size()`
**Why change**: After deserializing sparse offsets and values, they must be consistent
(values size = offsets size + 1, because the first value is always the default). A
truncated stream in Native format could produce a size mismatch.

### SerializationObject.cpp:987

**Function**: `deserializeBinaryBulkWithMultipleStreams`
**Check**: `!data_stream` (missing stream for Object data)
**Why change**: In the STRING serialization version path, the Object data stream is
required. If it's missing in Native format, the client sent an incomplete Object column.
Note: the size checks at lines 1069/1075 already use the correct pattern.

### SerializationDynamic.cpp:662

**Function**: `deserializeBinaryBulkWithMultipleStreams`
**Check**: `!variant_info.variant_type->equals(*structure_state->variant_type)`
**Why change**: The variant type from the column's state doesn't match what the stream
header declared. In Native format, this could result from a corrupted or misframed
stream where the type prefix and the column data don't agree. Note: the same check in
the **serialize** path at line 567 is left as LOGICAL_ERROR.

### SerializationReplicated.cpp:198,224,234

**Function**: `deserializeBinaryBulkWithMultipleStreams`
**Checks**:
- Line 198: `num_rows != limit` — row count from stream doesn't match expected
- Line 224: `default:` in switch on `size_of_indexes_type` — invalid index type size
- Line 234: `!elements_stream` — missing elements stream

**Why change**: This function is **exclusively** used in Native format (line 175 asserts
`settings.native_format`), so `INCORRECT_DATA` is used directly rather than the ternary
pattern. All three checks validate data read from the client's stream: wrong row count,
invalid index type byte, or missing sub-stream. Note: the pre-condition checks at lines
176-182 (wrong format, non-zero offset, non-empty column) are left as LOGICAL_ERROR since
those are caller contract violations, not data corruption.

## What was NOT changed

- **Serialize methods**: Errors during serialization are always internal bugs (the server
  controls what it writes). E.g., `SerializationLowCardinality.cpp:454,457`,
  `SerializationReplicated.cpp:129,137`, `SerializationVariant.cpp:161,274,327,461`.

- **SerializationObjectSharedData.cpp**: All LOGICAL_ERROR throws are in MergeTree-specific
  code paths (Compact/Wide parts with marks, structure prefixes/suffixes, seek callbacks).
  These are never reached from Native format. `settings.native_format` would always be
  false, so the ternary would produce LOGICAL_ERROR anyway — changing them would be
  harmless but unnecessary noise.

- **Internal invariant checks**: E.g., `SerializationVariant.cpp:502` (wrong variant
  order — a column setup bug, not data corruption), `SerializationDetached.cpp:87`
  (ColumnBLOB not converted), `ISerialization.h:690,697` (wrong state type).

- **SerializationReplicated.cpp:176,179,182**: Pre-condition guards checking that the
  method is called correctly (native format only, zero offset, empty column). These are
  caller contract violations, not client data issues.

- **Already correct**: `SerializationVariant.cpp:631`, `SerializationObject.cpp:1069,1075`,
  `SerializationTuple.cpp:829`, `SerializationNullable.cpp:163` already use the
  `settings.native_format ? INCORRECT_DATA : LOGICAL_ERROR` pattern.

yariks5s (approved):

antaljanosbenjamin:
@groeneai investigate the failure of test_grpc_protocol/test.py::test_progress

groeneai:
@antaljanosbenjamin Investigated the `test_grpc_protocol/test.py::test_progress` failure on this PR (job `Integration tests (amd_asan_ubsan, db disk, old analyzer, 4/6)`, run `25946427991/job/76280130901`).

**Verdict: unrelated chronic timing flake, not PR-caused.** Recommend rerun.

### The failure

```
File: test_grpc_protocol/test.py:397 - in test_progress
    assert progresses == expected_progresses
E   AssertionError
E     At index 2 diff: read_rows: 4\nread_bytes: 32\n != read_rows: 2\nread_bytes: 16\n
E     Right contains one more item: read_rows: 2\nread_bytes: 16\n
```

The test runs `SELECT number, sleep(0.31) FROM numbers(8) SETTINGS max_block_size=2, interactive_delay=100000` and expects exactly 4 progress callbacks of 2 rows each. On this run the 3rd and 4th blocks were coalesced into a single progress update (4 rows / 32 bytes), so only 3 progress events arrived instead of 4. This is a timing artifact: under ASan+UBSan + old analyzer + db-disk (one of the slowest integration configs), scheduling jitter can let two `max_block_size=2` blocks complete inside a single `interactive_delay=100000µs` window, and the server merges them.

### Why this is not caused by this PR

The diff changes only the error code in `Serialization{Array,Dynamic,LowCardinality,Object,Replicated,Sparse,String,Variant,VariantElement}::deserializeBinaryBulkWithMultipleStreams` (`LOGICAL_ERROR` → `INCORRECT_DATA`, gated on `settings.native_format`). It does not touch:
- The gRPC server (`src/Server/grpc/`)
- `IBlockInputStream` / `Progress` reporting
- `Block` / `Chunk` flow control
- The `numbers(8)` table function or `sleep` function

The failing query never enters any of the modified deserialization paths — it returns Block data from `numbers()` over gRPC, no Native-format ingestion at all.

### CIDB cross-PR history (last 60 days, `default.checks`)

Same test on the same job family (`Integration tests (*, db disk, old analyzer, *)`):

| day | failures | distinct PRs | notes |
|---|---|---|---|
| 2026-05-16 | 1 | 1 | **this PR** |
| 2026-04-07 | 1 | 1 | PR #100332 (unrelated — `throw-on-unknown-config-option`) |
| 2026-03-31 | 1 | 1 | PR #100332 |
| 2026-03-30 | 1 | 1 | PR #100332 |
| 2026-03-18 | 3 | 3 | outage tail |
| 2026-03-17 | 16 | 16 | **system-wide CI outage: 130,519 total integration failures that day** |

Excluding the 17/18-March outage, the test sees 4 isolated failures across 60 days on that job class — and the overall rate is **23 failures / 10,781 runs = 0.21%** on the `db disk, old analyzer` shard.

No open issue or fix PR for `test_progress` flakiness (`gh search issues "test_grpc_protocol test_progress" --repo ClickHouse/ClickHouse` returns nothing). The 1-hit profile on this PR matches the background rate of the chronic flake exactly — there's no PR-induced jump.

### Recommendation

1. Rerun the failing job — it should pass on retry.
2. (Optional, separate follow-up) The test's hard equality assertion against `expected_progresses` is brittle under sanitizer + slow-IO timing. Either widening it to allow merged progress events, or pinning a larger per-row sleep / smaller `interactive_delay`, would be a real fix — but that's a flaky-test cleanup, not part of this PR's scope.

### Re: the bot's regression-test ask

Separate from the directive, the bot is right that this PR would benefit from a focused test that ingests a truncated/malformed `Native` payload and asserts the new `INCORRECT_DATA` code on the modified paths (`SerializationArray`, `SerializationVariant`, `SerializationLowCardinality`, etc.). Happy to send that as a follow-up if useful — let me know.

alexey-milovidov:

antaljanosbenjamin:
@yariks5s I reverted most of the changes because nor me neither AI could write a test to trigger them. I was a bit too eager to fix them: they should stay logical errors because they cannot be triggered only by faulty input data, something inside clickhouse has to be broken.

yariks5s (approved):

PR #102997, opened 2026-04-17
Hide secret key in HMAC SQL function

### LINKED ISSUE #102927: HMAC() function does not redact the secret key in system.query_log
### Company or project name

_No response_

The [`HMAC()`](https://clickhouse.com/docs/sql-reference/functions/hash-functions#hmac) function (introduced in v25.12) logs its `key` argument (3rd parameter) in **plaintext** to `system.query_log`. This exposes secret keys to anyone with query log access.

25.12

### How to reproduce

Steps to Reproduce
-- 1. Run a query using HMAC with a literal secret key
SELECT lower(hex(HMAC('sha256', 'some-player-id', 'this_should_be_secret')));

-- 2. Inspect query_log — the secret key is fully visible
SELECT query
FROM system.query_log
WHERE query LIKE '%HMAC%'
ORDER BY event_time DESC
LIMIT 1;
Actual Result
The secret key is stored in plaintext in system.query_log:

SELECT lower(HEX(HMAC('sha256', 'some-player-id', 'this_should_be_secret')))
Expected Result
The key argument should be replaced with [HIDDEN], consistent with how encrypt/decrypt handle secret arguments:

SELECT lower(HEX(HMAC('sha256', 'some-player-id', '[HIDDEN]')))

### Expected behavior

The key argument should be replaced with [HIDDEN], consistent with how encrypt/decrypt handle secret arguments, like how this is handled for URL function 

SELECT lower(HEX(HMAC('sha256', 'some-player-id', '[HIDDEN]')))

_No response_

### Additional context

_No response_

Felixoid:

Hide secret key in `HMAC` SQL function. Fix #102927.

### Version info
- Merged into: `26.4.1.1114`
- Backported to: `26.3.10.30`, `26.2.17.19`, `26.1.12.14`


src/Parsers/FunctionSecretArgumentsFinder.h
```diff
@@ -608,6 +613,17 @@ class FunctionSecretArgumentsFinder
         result.count = function->arguments->size() - 1;
     }
 
+    void findHMACSecretArguments()
+    {
+        if (function->arguments->size() < 3)
```
scanhex12, line 618: Let's add something like

```
chassert(function->arguments->size() == 3);
```

after this if

scanhex12: As I understand from the documentation, it is not possible to specify more than 3 parameters in this function

scanhex12 (approved):
looks good

src/Parsers/FunctionSecretArgumentsFinder.h
Felixoid: Yes, the number of arguments must be 3. THanks!

Felixoid: I'll revert it, because it breaks the Fuzzer and looks redundant https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102997&sha=654a68ad192a8f856c13e1c404eb0c2a6fd4a0c6&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%2C%20targeted%29

Felixoid: Improved it by hiding all arguments after 3rd

clickgapai:
Hi — this PR may need backporting to `26.3` (LTS), `26.2`, `26.1`, but no backport label was found.

**Affected code:** `FunctionSecretArgumentsFinder::findOrdinaryFunctionSecretArguments` in FunctionSecretArgumentsFinder.h — introduced in 26.1.

**Why:** Security fix: HMAC function secret key leaks in `system.query_log` and `system.query_cache`. The HMAC function was introduced in commit e115854e3c60 (2025-11-25), after the 25.8 release (2025-08-29) but before 26.1 (2026-01-30). Therefore 25.8 does not contain HMAC and needs no backport; 26.1, 26.2, and 26.3 all contain HMAC and need this fix.

Other supported branches (`25.8`) predate 26.1 and do not contain this code.

PR #88035, opened 2025-10-02
Remove custom MemoryPools for reading Arrow/ORC/Parquet

nikitamikhaylov:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Remove custom MemoryPools for reading Arrow/ORC/Parquet. This component seems unneeded after https://github.com/ClickHouse/ClickHouse/pull/84082 because now we track all the allocations regardless. 

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #102918, opened 2026-04-16
Fix jemalloc metadata corruption caused by page cache freeing with wrong alignment

azat:
## Linked issues
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1554
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1555
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1784
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1080
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1811
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1809
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1829

### Changelog category (leave one):
- Critical Bug Fix (crash, data loss, RBAC)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix jemalloc metadata corruption caused by page cache freeing with wrong alignment that may lead to crashes

`JemallocCacheAllocator::alloc` passes `MALLOCX_ALIGN(alignment)` to `je_mallocx`, which changes the size class of the allocation. But `free` called `je_sdallocx` without the alignment flag, which may cause jemalloc to compute a different size class for the same size — corrupting jemalloc's internal extent/slab metadata.

The only affected caller is `PageCacheCell`, which allocates with `DEFAULT_AIO_FILE_BLOCK_SIZE` alignment but freed without it.

Found with JEMALLOC_OPT_SAFETY_CHECKS/config_opt_safety_checks [1], while working on https://github.com/ClickHouse/ClickHouse/pull/102913 (note, that https://github.com/ClickHouse/ClickHouse/pull/102913 is more important)

  [1]: https://pastila.nl/?0004ee83/c57a3bc6fb7219dc2149b459895c6ac3#nvyVGdUfxMop9XHaDpS36A==GCM

### Version info
- Merged into: `26.4.1.1062`
- Backported to: `26.3.10.17`

PR #102999, opened 2026-04-17
Reintroduce ArrowMemoryPool to allow throwing MEMORY_LIMIT_EXCEEDED to avoid kernel OOM

azat:

Reintroduce ArrowMemoryPool to allow throwing MEMORY_LIMIT_EXCEEDED to avoid kernel OOM

PR #88035 removed the custom `ArrowMemoryPool` and switched arrow/parquet/ORC allocations to `arrow::default_memory_pool()`, relying on the `posix_memalign` interceptor for memory tracking. However, the interceptor cannot throw `MEMORY_LIMIT_EXCEEDED`, because throwing from `malloc`/`posix_memalign` is not allowed (arrow and its dependencies do not expect exceptions from allocation).

Reintroduce `ArrowMemoryPool` routing allocations through `Allocator<false>`, which *can* throw `MEMORY_LIMIT_EXCEEDED`. The exception is caught and converted into `arrow::Status::OutOfMemory`, so arrow unwinds via its normal error path and the original message is preserved for diagnostics.

All call sites that PR #88035 moved to `arrow::default_memory_pool()` are switched back to `ArrowMemoryPool::instance()`. The `NativeORC` pool is not restored because `orc::MemoryPool::malloc` has no status channel.

### Version info
- Merged into: `26.4.1.1075`
- Backported to: `26.3.10.25`, `26.2.17.17`, `26.1.12.12`

### DIFF
--- a/src/Processors/Formats/Impl/ArrowBufferedStreams.cpp
+++ b/src/Processors/Formats/Impl/ArrowBufferedStreams.cpp
@@ -12,6 +12,7 @@
 #include <arrow/io/memory.h>
 #include <arrow/result.h>
 #include <arrow/memory_pool_internal.h>
+#include <Common/Allocator.h>
 #include <Core/Settings.h>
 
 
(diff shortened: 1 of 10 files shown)


src/Processors/Formats/Impl/ArrowBufferedStreams.cpp
```diff
@@ -248,6 +249,75 @@ arrow::Result<int64_t> RandomAccessFileFromRandomAccessReadBuffer::Tell() const
...
+
+void ArrowMemoryPool::Free(uint8_t * buffer, int64_t size, int64_t /*alignment*/)
+{
+    if (size == 0)
+    {
+        chassert(buffer == arrow::memory_pool::internal::kZeroSizeArea);
+        return;
+    }
+
+    Allocator<false>().free(buffer, size_t(size));
```
azat, line 316: Blocked by https://github.com/ClickHouse/ClickHouse/pull/102918

azat: Fixed

azat:
@nikitamikhaylov PTAL

nikitamikhaylov (approved):

PR #103014, opened 2026-04-17
Fix `MULTIPLE_EXPRESSIONS_FOR_ALIAS` in `FuseFunctionsPass` for repeated `quantile` on distributed queries

### LINKED ISSUE #102976: MULTIPLE_EXPRESSIONS_FOR_ALIAS when same quantile aggregate is used in SELECT/HAVING/ORDER BY on distributed query (regression in 26.3)
**Describe what's wrong**

On a distributed query (using the `remote()` table function, or any `Distributed` table), referencing the same aggregate alias in `SELECT`, `HAVING`, and `ORDER BY` fails with `MULTIPLE_EXPRESSIONS_FOR_ALIAS`.

The query optimizer fuses the three `quantile(0.99)(x)` calls into a single `quantiles(0.99, 0.99, 0.99)(x)` and extracts elements 1, 2, 3 — but assigns the same internal alias `quantile(0.99)(__table1.number)` to two of the three extractions, which then collide.

**Does it reproduce on the most recent release?**

Reproduces on **26.3.9**. Does **not** reproduce on **26.2** or earlier. Was introduced somewhere in the 26.3 line (confirmed broken as of **26.3.1.896**).

**How to reproduce**

* ClickHouse server version: 26.3.9 (also 26.3.1.896)
* Fiddle: https://fiddle.clickhouse.com/dd51abea-7baa-41a3-aea0-c2b8a79a15f8

Queries used:

```sql
-- Fails on 26.3.x, works on 26.2 and earlier
SELECT
    number % 10 AS bucket,
    count() AS c,
    quantile(0.99)(number) AS p99
FROM remote('127.0.0.1', system.numbers)
WHERE number < 1000
GROUP BY bucket
HAVING p99 > 100
ORDER BY p99 DESC
LIMIT 10;
```

```sql
-- Control: same query against the local table succeeds
SELECT
    number % 10 AS bucket,
    count() AS c,
    quantile(0.99)(number) AS p99
FROM system.numbers
WHERE number < 1000
GROUP BY bucket
HAVING p99 > 100
ORDER BY p99 DESC
LIMIT 10;
```

**Expected behavior**

Query returns grouped rows, consistent with the non-distributed behavior and with 26.2.

**Error message and/or stacktrace**

```
Code: 179. DB::Exception: Received from localhost:9000. DB::Exception:
Multiple expressions
  arrayElement(quantiles(0.99, 0.99, 0.99)(__table1.number), 1) AS `quantile(0.99)(__table1.number)`
and
  arrayElement(quantiles(0.99, 0.99, 0.99)(__table1.number), 2) AS `quantile(0.99)(__table1.number)`
for alias quantile(0.99)(__table1.number).

In scope SELECT __table1.number % 10 AS bucket, count() AS c,
  arrayElement(quantiles(0.99, 0.99, 0.99)(__table1.number), 3) AS p99
FROM system.numbers AS __table1
WHERE __table1.number < 1000
GROUP BY __table1.number % 10
HAVING (arrayElement(quantiles(0.99, 0.99, 0.99)(__table1.number), 2) AS `quantile(0.99)(__table1.number)`) > 100
ORDER BY arrayElement(quantiles(0.99, 0.99, 0.99)(__table1.number), 1) AS `quantile(0.99)(__table1.number)` DESC
LIMIT _CAST(10, 'UInt64').
(MULTIPLE_EXPRESSIONS_FOR_ALIAS)
```

**Additional context**

- Fails even when `HAVING` / `ORDER BY` reference the SELECT alias (`p99`) rather than re-stating the aggregate — the alias is expanded before the quantile-fusion optimization runs, producing three copies of the aggregate expression and triggering the collision.
- Only reproduces when the read goes through a distributed path (`remote()` or a `Distributed` table). Local-table queries with the exact same shape succeed.
- Regression between 26.2 and 26.3.1.896.

- **Workaround:** setting `optimize_syntax_fuse_functions = 0` resolves the error by disabling the quantile-fusion optimization, which confirms the fusion pass is where the alias collision originates.

tanner-bruce:
When the same `quantile(p)(x)` was referenced multiple times in a query (for example in `SELECT`, `HAVING`, and `ORDER BY`), `FuseFunctionsPass` packed every reference into a single `quantiles(p, p, p, ...)` call with a distinct `arrayElement(..., i)` index per occurrence. Commit 551cfae0c17 then forced the same `quantile(p)(x)` alias onto each extraction so that consumers matching columns by name (for example `StorageBuffer` union and `Distributed` queries) would see unchanged names.

The combination is unsound: when the analyzer serializes the query back to AST for distributed dispatch, it re-checks `MULTIPLE_EXPRESSIONS_FOR_ALIAS` by tree hash with aliases ignored. The `arrayElement` nodes with different indices but identical aliases trip the exception.

Minimal repro (reliable on any `Distributed` table or `remote()`):

```sql
SELECT
    number % 10 AS bucket,
    count() AS c,
    quantile(0.99)(number) AS p99
FROM remote('127.0.0.1', system.numbers)
WHERE number < 1000
GROUP BY bucket
HAVING p99 > 100
ORDER BY p99 DESC
LIMIT 10;
-- Code: 179. DB::Exception: Multiple expressions
--   arrayElement(quantiles(0.99, 0.99, 0.99)(...), 1) AS `quantile(0.99)(...)` and
--   arrayElement(quantiles(0.99, 0.99, 0.99)(...), 2) AS `quantile(0.99)(...)`
--   for alias quantile(0.99)(...). (MULTIPLE_EXPRESSIONS_FOR_ALIAS)
```

The fix deduplicates by quantile level: each distinct `p` appears exactly once in the fused `quantiles` call, and every reference to the same `p` is rewritten to the same `arrayElement` extraction. Duplicate nodes become structurally identical, so sharing an alias is safe. If all references have the same level (for example repeated `quantile(0.99)`), there is nothing to fuse and the pass leaves the original functions alone.

As a small bonus, the fused aggregate now has `K` slots (distinct levels) instead of `N` (occurrences), so the runtime computes fewer quantile digests.

Fixes #102976.

Fix `MULTIPLE_EXPRESSIONS_FOR_ALIAS` exception raised on distributed queries that reference the same `quantile` call multiple times (e.g. in `SELECT`, `HAVING`, and `ORDER BY`) with `optimize_syntax_fuse_functions` enabled.

### Version info
- Backported to: `26.4.1.1139`, `26.3.10.60`


tanner-bruce:
Also, if possible could this be backported to 26.3

tanner-bruce:
Analyzed the only failing check — Performance Comparison (arm_release, 6/6), "7 slower, 2 unstable".

The reported regressions are all in codepaths this PR does not touch:

- `set_index`, `logical_functions_large`, `conditional` (COALESCE), `date_time_short`, `prewhere_subcolumns`, `arithmetic_operations_in_aggr_func`, `reading_from_file` (parquet)
- `fuse_sumcount` #0, #1 — the `sum`/`count`/`avg` fusion path (`tryFuseSumCountAvg`). This PR only modifies `tryFuseQuantiles`. Notably `fuse_sumcount #0` is `SELECT sum(number) FROM numbers(1e9)` with a single aggregate, so no fusion happens at all.
- `avg #0, #1, #5` — likewise unrelated.

No quantile-related perf test exists, and the only quantile-related code touched is `tryFuseQuantiles`. Pattern of small (5–15%) slowdowns scattered across unrelated subsystems on a single shard looks like a loaded CI runner rather than a real regression from this change.

Happy to rerun or investigate any specific query if useful.

Issue #76473, opened 2025-02-19
File function - ability to read corrupted archives on the best effort basis


pakud:
### Use case

I'm using clickhouse-local to do ad-hoc analysis of logs from various sources. Logs are kept on local file system - as thousands of compressed files. Files usually are gzip'ed, but once in a while some files are truncated or not-really-gzips - e.g. Fastly CDN creates 10 byte long files with .gz extension that are not really gzips when there's no traffic at all at given time.

### Describe the solution you'd like

Would it be feasible to add option similar to _engine_file_skip_empty_files_ but going one step further and allowing to continue reading even if one of files cannot be completely unpacked?

### Describe alternatives you've considered

Pre-processing log files to filter out corrupted ones.

### Additional context

I'm using clickhouse-local 25.1.4.53 from official deb package for amd64.

```
SELECT * FROM file('/tmp/sample/*.gz','LineAsString','c text') WHERE c LIKE '%' SETTINGS engine_file_skip_empty_files =1;


Received exception:
Code: 354. DB::Exception: inflate failed: buffer error: While reading from: /tmp/sample/10b.gz: (in file/uri /tmp/sample/10b.gz): While executing ParallelParsingBlockInputFormat: While executing File. (ZLIB_INFLATE_FAILED)
```

Files present in /tmp/sample:

[10b.gz](https://github.com/user-attachments/files/18863737/10b.gz)
[truncated.gz](https://github.com/user-attachments/files/18863736/truncated.gz)
[zero.gz](https://github.com/user-attachments/files/18863735/zero.gz)

Thanks for considering it!

PR #103034, opened 2026-04-17
issue-76473: Add engine_file_skip_failed_files setting to skip corrupted files in File engine

mrk-andreev:
Relates to https://github.com/ClickHouse/ClickHouse/issues/76473 

### Changelog category
- New Feature

Added a new setting engine_file_skip_failed_files for the File table engine that allows ClickHouse to skip corrupted or unreadable files instead of failing the entire query. When enabled, files that cause read errors are silently skipped, improving resilience when querying directories containing partially corrupted data.

**Motivation:** When using the File table engine to read from a directory containing multiple files, a single corrupted or malformed file would cause the entire query to fail. The new engine_file_skip_failed_files setting allows users to tolerate such failures gracefully, making bulk file ingestion and exploratory queries more robust in environments where data quality cannot be guaranteed.

**Parameters:**

`engine_file_skip_failed_files` (Boolean, default: `false`) - When set to true, files that fail to be read (e.g., due to corruption or format errors) are skipped and the query continues processing remaining files. When false (default), any file read error causes the query to fail immediately.

**Example use:**

```
SET engine_file_skip_failed_files = 1;

SELECT * FROM file('/var/data/logs/*.csv', CSV);
-- Files that fail to parse are skipped; valid files are returned normally.
```


src/Core/SettingsChangesHistory.cpp
```diff
@@ -138,10 +95,10 @@ const VersionToSettingsChangesMap & getSettingsChangesHistory()
             {"ai_function_throw_on_quota_exceeded", true, true, "New setting"},
             {"variant_throw_on_type_mismatch", true, true, "New setting to control type mismatch behavior in default Variant implementation"},
             {"dynamic_throw_on_type_mismatch", true, true, "New setting to control type mismatch behavior in default Dynamic implementation"},
+            {"engine_file_skip_failed_files", false, false, "New setting to skip files that fail to open or read in File engine tables."},
```
mrk-andreev: I guess I need to move that config parameter to the correct release.

PR #101801, opened 2026-04-05
keeper-bench: more features

al13n321:
### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

* You can now add `watch_probability` in config on read requests. Then this fraction of requests would set a watch (with callback that just bumps a counter).
* Periodically printed stats now show throughputs and latencies from the last period rather than the whole run. So changes are visible right away, e.g. if the server got into a bad state or if settings were changed on the fly.
* Removed per-operation-type stats. Their latencies should all be the same anyway because of batching.
* Made startup remove and create all znodes at once in a batch instead of one by one. Now you can afford to use 100k znodes in the test instead of 1k (1k took 5 seconds to create one by one, now 100k takes 2 seconds).
* In setup nodes config, you can add `tag: <name>` to a node, then refer to it in any `path` config with `tagged: <name>`. Makes nested `repeat` nodes usable. (Nested repeat nodes are needed e.g. to test how `list` performance depends on total number of znodes, without changing `list` output size: you'd have n outer znodes with m inner znodes in each, and do `list` on random outer znodes, and vary n while keeping m constant.)

```
Total: read 17027616, write 10844060, watches fired 154891
Last 1.00712s: Read 151574 RPS, 191.392 MiB/s. Write 96885.9 RPS, 23.3766 MiB/s.
Read latency:
0%		17.196 ms.	
10%		29.236 ms.	
20%		33.395 ms.	
30%		38.902 ms.	
40%		48.66 ms.	
50%		53.851 ms.	
60%		56.624 ms.	
70%		59.141 ms.	
80%		62.701 ms.	
90%		67.112 ms.	
95%		76.021 ms.	
99%		98.127 ms.	
99.9%		103.84 ms.	
99.99%		109.292 ms.	
Write latency:
0%		17.072 ms.	
10%		29.557 ms.	
20%		33.636 ms.	
30%		38.512 ms.	
40%		48.933 ms.	
50%		53.935 ms.	
60%		56.84 ms.	
70%		59.344 ms.	
80%		62.825 ms.	
90%		67.112 ms.	
95%		75.951 ms.	
99%		97.951 ms.	
99.9%		103.484 ms.	
99.99%		109.215 ms.	
Requests executed: 28119772.
```

### Version info
- Merged into: `26.4.1.825`

PR #103050, opened 2026-04-17
keeper-bench: fall back to manual recursive remove when server lacks REMOVE_RECURSIVE

al13n321:

...

---

#### Description

PR #101801 (`keeper-bench: more features`) switched `keeper-bench` setup/cleanup to the native `RemoveRecursive` Keeper request. That request is only supported by ClickHouse Keeper versions that advertise the `REMOVE_RECURSIVE` feature flag, so running `keeper-bench` against plain ZooKeeper or an older Keeper fails with ``RemoveRecursive request type cannot be used because it's not supported by the server``.

This PR restores the pre-#101801 fallback:

- The cleanup path checks `Coordination::ZooKeeper::isFeatureEnabled(REMOVE_RECURSIVE)`. If the feature flag is enabled, the native `removeRecursive` is used (fast path introduced in #101801). Otherwise, the previous manual implementation — `list` each subtree, then batched `multi(remove)` — is used.
- A new top-level config option `use_remove_recursive` (default `true`) lets users force the manual path even against a Keeper that supports the native request.

The config option and behavior are documented in `programs/keeper-bench/README.md`.

### Version info
- Merged into: `26.5.1.118`

### DIFF
--- a/programs/keeper-bench/README.md
+++ b/programs/keeper-bench/README.md
@@ -103,6 +103,7 @@ Config can be YAML or XML.
 | `pipeline_depth` | integer | `1` | In-flight async requests per worker (`>= 1`) |
 | `warmup_seconds` | float | `0` | Measurement warmup window |
 | `enable_tracing` | bool | `false` | Attach OpenTelemetry trace context |
+| `use_remove_recursive` | bool | `true` | Use the native `RemoveRecursive` Keeper request for cleanup; falls back to a manual recursive traversal if the server does not advertise the `REMOVE_RECURSIVE` feature flag or if this is set to `false` (needed for ZooKeeper or older Keeper versions) |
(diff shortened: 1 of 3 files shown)


antonio2368 (approved):

al13n321:
[Performance Comparison](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=103050&sha=4ff2930f55068125895f7e3b7a662b6298a0da31&name_0=PR&name_1=Performance+Comparison+%28arm_release%2C+master_head%2C+5%2F6%29) failure presumably unrelated, overriding:
```
gh api repos/ClickHouse/ClickHouse/statuses/4ff2930f55068125895f7e3b7a662b6298a0da31 -X POST -F state=success -F description="Manually unset to merge" -F context="Mergeable Check"
```

PR #103054, opened 2026-04-17
Enforce AI function quotas per query, not per block

george-larionov:
`AIQuotaTracker` was created as a local variable inside `executeImpl`, which runs once per block. For multi-block queries the quota reset on each block, making `ai_function_max_*_per_query` settings effectively per-block limits.

Move the tracker to a `mutable` member on `FunctionBaseAI`, initialized lazily on the first `executeImpl` call. Since `isStateful` returns true, the same function instance is reused across all blocks in a query, so the quota now accumulates correctly across the entire query.

### DIFF
--- a/src/Functions/FunctionBaseAI.h
+++ b/src/Functions/FunctionBaseAI.h
@@ -6,6 +6,9 @@
 #include <Functions/AI/AIQuotaTracker.h>
 #include <Interpreters/Context.h>
 
+#include <mutex>
+#include <optional>
+
 namespace DB
 {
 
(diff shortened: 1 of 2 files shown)


src/Functions/FunctionBaseAI.h
```diff
@@ -69,6 +72,13 @@ class FunctionBaseAI : public IFunction
 
     ResolvedConfig resolveConfig(const ColumnsWithTypeAndName & arguments) const;
     float resolveTemperature(const ColumnsWithTypeAndName & arguments, const ResolvedConfig & config) const;
+
+    /// Quota tracker persists across executeImpl calls (one per block) so that
+    /// per-query limits are enforced for the entire query, not per block.
+    /// Guarded by mutex because the query engine may call executeImpl from
+    /// different threads for different blocks.
+    mutable std::mutex quota_mutex;
+    mutable std::optional<AIQuotaTracker> quota;
```
rschu1ze, line 81: Could add `TSA_GUARDED_BY(quota_mutex)`

rschu1ze (approved):
I think the approach will work but better test to double-check (multiple consecutive queries, one query with `SETTINGS max_threads > 1` etc.)

rschu1ze:
@george-larionov I guess this PR is obsolete with the future AI service. Let's close this?

PR #103064, opened 2026-04-18
Send MemoryAllocatedWithoutCheck even in release builds

azat:
This should not be a huge source of events, while it will help with memory introspections.

- Improvement

Send MemoryAllocatedWithoutCheck even in release builds

### Version info
- Merged into: `26.4.1.1074`


src/Common/MemoryTracker.cpp
```diff
@@ -247,26 +247,19 @@ void incrementAllocationWithoutCheck(Int64 size)
...
 
     /// The choice is arbitrary (maybe we should decrease it)
     constexpr Int64 threshold = 16 * 1024 * 1024;
 
     if (size > threshold)
     {
-        auto memory_blocked_context = MemoryTrackerBlockerInThread::getLevel();
-        MemoryTrackerBlockerInThread tracker_blocker(VariableContext::Global);
-        /// Forbid recursive calls
-        [[maybe_unused]] MemoryTrackerDebugBlockerInThread debug_blocker;
```
azat, line 263: This has been added in b6fcf3e7d6162cf5c045dbce49ad08cee1e3aed4, but is redundant after e7a58e70

src/Common/MemoryTrackerUntrackedAllocationsBlockerInThread.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+/// Use it if you need to suppress MemoryAllocatedWithoutCheck for known big allocations.
```
azat, line 3: And now this is the only purpose, suppress - known allocations, we can get rid of it, but the idea was that we should have "almost 0" such allocations, while w/o there will be some common traces

Avogar (approved):

azat:
`logical_functions_medium` and `logical_functions_small` are flaky

PR #103066, opened 2026-04-18
Move cancellation decision for partial result into processors

Michicosun:

Pipeline can dynamically create subtrees of processors - because of that, it is not correct to cancel only the sources from the startup snapshot when a partial result is being requested.

### Version info
- Merged into: `26.4.1.1071`


src/Processors/Transforms/AggregatingTransform.cpp
```diff
@@ -230,6 +230,16 @@ class ConvertingAggregatedToChunksWithMergingSource final : public ISource
 
     String getName() const override { return "ConvertingAggregatedToChunksWithMergingSource"; }
 
+    void cancel(CancelReason reason) noexcept override
+    {
+        /// When 2-level aggregation is being used ConvertingAggregatedToChunksTransform expects
+        /// to receve data from all sources, that is why we need not to stop processor here.
```
Michicosun:
```suggestion
        /// to receive data from all sources, so we do not need to stop the processor here.
```

SmitaRKulkarni (approved):
Maybe we can add some test for PartialResult path. Rest all LGTM

Michicosun:
> Maybe we can add some test for PartialResult path. Rest all LGTM

There is one 02697

PR #101452, opened 2026-04-01
Fix Nullable predicate regression in VirtualColumnUtils filter

groeneai:
## Summary

Fixes a wrong-results regression introduced by #101287 where Nullable predicates in AND clauses caused `splitFilterNodeForAllowedInputs` to incorrectly filter out all rows.

**Root cause:** When an AND node is reduced to a single child with a different type, the code creates `notEquals(x, 0)` for boolean conversion. The zero value was obtained via `res->result_type->getDefault()`, which for `Nullable(T)` returns `Null()` instead of `T(0)`. This made `notEquals(x, NULL)` always return `NULL` (SQL three-valued logic), which is falsy — causing all rows/parts to be discarded.

**Reproduction:** Any query that triggers `filterBlockWithPredicate` with a Nullable constant in an AND clause where the other operand is not in the allowed inputs. For example:
```sql
-- Returns 0 instead of 1 on master (should find system.one):
SELECT count() FROM system.tables WHERE name = 'one' AND toNullable(toInt8(-1));
```

**Fix:** Use `removeNullable(res->result_type)->getDefault()` to produce a real zero (e.g. `Int8(0)`) instead of `NULL`. The zero is then wrapped in the original Nullable type via `createColumnConst`, so `notEquals(-1, Nullable(0))` correctly returns `1`.

Also fixes the `indexHint` path (line 596) which still used the old `addCast` approach (the truncation bug from #101269) — replaced with the same `notEquals(x, 0)` approach with the Nullable fix included.

**Verified in both directions:**
- Without fix: `SELECT count() FROM system.tables WHERE name = 'one' AND toNullable(toInt8(-1))` → **0** (wrong)
- With fix: same query → **1** (correct)
- 50/50 passes with full CI randomization

Closes #101433

### Changelog category (pick one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):

Fix wrong results for queries with Nullable constant predicates in WHERE on tables using virtual column filtering (system tables, MergeTree part pruning).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #101269, opened 2026-03-30
`(2147483648 > b) AND 2147483648` is not the same result as `(2147483648 > b)`


qoega:
### Describe the bug

Incorrect result when maxint is compared with bool.

Found by SQLancer

### How to reproduce

https://fiddle.clickhouse.com/53b130c5-841e-4a46-86a8-8269c87fdd9c

```
create table t (b Int8) ORDER BY ();
insert into t values (1), (0);

SELECT count() FROM t WHERE (2147483648 > b) AND 2147483648; -- BUG! returns 0

-- All this options return 2
SELECT count() FROM t WHERE 2147483648;
SELECT countIf((2147483648 > b) AND 2147483648) FROM t;
SELECT count() FROM t WHERE (2147483648 > b);
SELECT countIf((2147483648 > b)) FROM t;
```

### Error message and/or stacktrace

```
The optimizer converts AND 2147483648 by casting the integer to UInt8 via toUInt8, which truncates 2147483648 to 0 (since 2147483648 mod 256 = 0). This is a constant folding bug: when a large integer literal is used as a boolean predicate, the implicit conversion to UInt8 loses the truthiness.
```

Issue #99979, opened 2026-03-18
MergeTree WHERE incorrectly evaluates AND with integer literal -2147483648 (INT_MIN) as false


qoega:
When a WHERE clause in a MergeTree table contains AND -2147483648 (or other negative integer literals that truncate to 0 when cast to UInt8), the filter incorrectly evaluates the entire predicate as false and returns 0
   rows. The same query works correctly with the Memory engine and the expression evaluates correctly in SELECT.

  The root cause appears to be that MergeTree's filter path casts the Int32 literal -2147483648 to UInt8 (which overflows to 0) before evaluating the and function, rather than treating any non-zero value as truthy.

  Reproducer:

``` sql
  CREATE TABLE bug (x Int32) ENGINE = MergeTree ORDER BY x;
  INSERT INTO bug VALUES (1), (2);

  -- Returns 0, expected 2.
  -- -2147483648 is non-zero (truthy), so AND -2147483648 should be equivalent to AND 1.
  SELECT count() FROM bug WHERE (x > 0) AND -2147483648;

  -- The expression itself evaluates correctly in SELECT:
  SELECT x, (x > 0) AND -2147483648 FROM bug ORDER BY x;
  -- Returns: 1  1
  --          2  1

  -- Works correctly with Memory engine:
  CREATE TABLE bug_mem (x Int32) ENGINE = Memory;
  INSERT INTO bug_mem VALUES (1), (2);
  SELECT count() FROM bug_mem WHERE (x > 0) AND -2147483648;
  -- Returns: 2 (correct)
```
  Additional observations:

  - The bug also breaks ternary logic partitioning: `WHERE p + WHERE NOT p + WHERE isNull(p)` does not cover all rows:

```
  CREATE TABLE t (a Int32, c UInt32) ENGINE = MergeTree ORDER BY a;
  INSERT INTO t SELECT number % 100 - 50, number FROM numbers(200);

  SELECT count() FROM t;  -- 200

  SELECT count() FROM t WHERE (a > 10) AND -2147483648 AND (c < 150);         -- 0 (wrong, should be 39)
  SELECT count() FROM t WHERE NOT((a > 10) AND -2147483648 AND (c < 150));    -- 161
  SELECT count() FROM t WHERE isNull((a > 10) AND -2147483648 AND (c < 150)); -- 0
  -- Sum: 161, but should be 200. 39 rows are lost.
```
  - The bug is specific to MergeTree (Memory engine is correct).
[...]

PR #103077, opened 2026-04-18
Fix WHERE x AND toNullable(N) returning 0 rows for N >= 256

### LINKED ISSUE #103049: WHERE x AND toNullable(N) returns 0 rows for any N >= 256
### Describe the bug

https://fiddle.clickhouse.com/4de12199-1ab5-4ccf-accd-2993458faeff

### How to reproduce

```
      -- Bug: WHERE x AND toNullable(N) returns 0 rows for any N >= 256
      -- Found by AST fuzzer oracle (TLP WHERE) on master

      DROP TABLE IF EXISTS bug_repro;
      CREATE TABLE bug_repro (c String) ENGINE = MergeTree ORDER BY tuple();
      INSERT INTO bug_repro SELECT toString(number) FROM numbers(100);

      -- These should all return 100 (all rows match: c is always non-empty, N is truthy)
      SELECT 'expected 100:', count() FROM bug_repro WHERE (c != '') AND 256;            -- ✓ returns 100
      SELECT 'expected 100:', count() FROM bug_repro WHERE (c != '') AND toNullable(255); -- ✓ returns 100
      SELECT 'expected 100:', count() FROM bug_repro WHERE (c != '') AND toNullable(256); -- ✗ returns 0 (BUG!)
      SELECT 'expected 100:', count() FROM bug_repro WHERE (c != '') AND toNullable(65535); -- ✗ returns 0 (BUG!)
      SELECT 'expected 100:', count() FROM bug_repro WHERE (c != '') AND toNullable(2147483648); -- ✗ returns 0 (BUG!)

      -- As countIf — works correctly
      SELECT 'expected 100:', countIf((c != '') AND toNullable(256)) FROM bug_repro; -- ✓ returns 100
```

_No response_

groeneai:
Fixes a silent wrong-results bug reported in #103049 where
`WHERE (c != '') AND toNullable(N)` on any MergeTree table returns 0
rows for any `N >= 256`, even though `N` is a truthy integer.

### Repro

```sql
CREATE TABLE bug_repro (c String) ENGINE = MergeTree ORDER BY tuple();
INSERT INTO bug_repro SELECT toString(number) FROM numbers(100);

-- All expected to return 100:
SELECT count() FROM bug_repro WHERE (c != '') AND 256;                -- 100 OK
SELECT count() FROM bug_repro WHERE (c != '') AND toNullable(255);    -- 100 OK
SELECT count() FROM bug_repro WHERE (c != '') AND toNullable(256);    -- 0 BUG
SELECT count() FROM bug_repro WHERE (c != '') AND toNullable(65535);  -- 0 BUG
SELECT count() FROM bug_repro WHERE (c != '') AND toNullable(2147483648); -- 0 BUG
```

### Root cause

`VirtualColumnUtils::splitFilterNodeForAllowedInputs` reduces an `AND`
with some children outside the allowed-inputs set to the single
surviving child.  When that child's result type (`Nullable(UInt16)` for
`toNullable(256)`) differs from the parent `AND` result type
(`Nullable(UInt8)`), the code wraps it in `notEquals(x, 0)` to
materialise a boolean.  The zero constant was built from
`res->result_type->getDefault()`, but for `Nullable(T)` that returns
`Null()` — so the filter became `notEquals(toNullable(256), NULL)`
which is `NULL` under SQL three-valued logic and gets treated as false
by the filter, discarding every row.

The 255 boundary is a coincidence of the type system: `toNullable(255)`
has type `Nullable(UInt8)` which equals the parent `AND` result type,
so the type-mismatch branch is skipped entirely — the bug path is
never entered.  `256` promotes to `Nullable(UInt16)` and the bug
triggers.

### Fix

Use `removeNullable(res->result_type)->getDefault()` to obtain the
nested type's zero (`UInt16(0)` instead of `Null()`) and wrap it back
in `Nullable` via `createColumnConst`.  The generated comparison
becomes `notEquals(toNullable(256), Nullable(UInt16)(0))` which
correctly returns `1` and keeps all matching rows.

`Nullable(Nothing)` (a bare `NULL` literal) has no nested `getDefault`,
so we fall back to the outer `Nullable` default — preserving the
intended `notEquals(NULL, NULL) -> NULL -> false` semantics that drop
all rows.  Covered by a dedicated regression case.

### Verification

- Without the fix: `tests/clickhouse-test 100000_103049_nullable_and_overflow_filter` fails with `-1 +0` diffs on every `toNullable(>= 256)` / `toNullable(toInt8(-1))` case.
- With the fix: 20/20 passes with full CI randomization enabled (`--test-runs 20`).
- Repro `SELECT` queries above all return 100 with the fix, 0 without.

### Relation to other work

Same root cause as #101433 / #101269 / #99979.  PR #101452 proposes an
equivalent fix but has been stalled for 11 days and is ~3600 commits
behind master; this PR is a fresh, minimal change based on current
master so that #103049 does not block on that PR's merge.  Either PR
merging first resolves both issues.

Fix wrong results returned by `WHERE x AND toNullable(N)` on `MergeTree` tables when `N` is an integer wider than `UInt8` (e.g. `256`, `65535`, `2147483648`, or any negative integer).  The filter incorrectly discarded all rows because `splitFilterNodeForAllowedInputs` used a `NULL` zero when converting the `Nullable` remainder of a reduced `AND` to boolean, turning the comparison into `NULL` under three-valued logic.

Closes #103049.

### Version info
- Merged into: `26.4.1.1096`

### DIFF
--- a/src/Storages/VirtualColumnUtils.cpp
+++ b/src/Storages/VirtualColumnUtils.cpp
@@ -557,8 +557,20 @@ static const ActionsDAG::Node * splitFilterNodeForAllowedInputs(
                     /// Convert to boolean via notEquals(x, 0) instead of a truncating numeric cast.
                     /// A plain CAST(256, 'UInt8') would give 0 (since 256 % 256 == 0), losing truthiness
                     /// for values like 256, 512, 65536, 2147483648, etc.  See #101269.
+                    ///
+                    /// Use removeNullable to get the nested type's default (zero, not NULL).
+                    /// DataTypeNullable::getDefault() returns Null(), but notEquals(x, NULL)
+                    /// always returns NULL (SQL three-valued logic), which is treated as false and
+                    /// would incorrectly filter out all rows/parts.  See #101433 and #103049.
+                    /// Special case: Nullable(Nothing) — the child is a bare NULL literal.
+                    /// Nothing has no getDefault, so fall back to the Nullable default
+                    /// (Null field), which makes notEquals(x, NULL) -> NULL -> false.  Correct.
                     ActionsDAG tmp_dag;
-                    auto zero_column = res->result_type->createColumnConst(1, res->result_type->getDefault());
+                    auto nested_type = removeNullable(res->result_type);
+                    auto zero_field = (nested_type->getTypeId() == TypeIndex::Nothing)
+                        ? res->result_type->getDefault()
+                        : nested_type->getDefault();
+                    auto zero_column = res->result_type->createColumnConst(1, zero_field);
                     const auto & zero_node = tmp_dag.addColumn({zero_column, res->result_type, "0"});
                     auto ne_func = FunctionFactory::instance().get("notEquals", context);
                     res = &tmp_dag.addFunction(ne_func, {res, &zero_node}, {});
--- a/tests/queries/0_stateless/04104_103049_nullable_and_overflow_filter.reference
+++ b/tests/queries/0_stateless/04104_103049_nullable_and_overflow_filter.reference
@@ -21,42 +21,28 @@
 DROP TABLE IF EXISTS bug_103049;
 CREATE TABLE bug_103049 (c String) ENGINE = MergeTree ORDER BY tuple();
 INSERT INTO bug_103049 SELECT toString(number) FROM numbers(100);
-
 -- All of the following should return 100 (c is always non-empty, N is truthy).
 
 -- Non-nullable baseline: already correct (non-bug path).
 SELECT count() FROM bug_103049 WHERE (c != '') AND 256;
-
+100
 -- Nullable(UInt8) with N = 255 — works because result type matches parent.
 SELECT count() FROM bug_103049 WHERE (c != '') AND toNullable(255);
-
+100
 -- Nullable(UInt16) with N = 256 — the core of the bug.
 SELECT count() FROM bug_103049 WHERE (c != '') AND toNullable(256);
-
+100
 -- Nullable(UInt16) with N = 65535 — upper bound of UInt16.
 SELECT count() FROM bug_103049 WHERE (c != '') AND toNullable(65535);
-
+100
 -- Nullable(Int64) with N > UInt32 range.
 SELECT count() FROM bug_103049 WHERE (c != '') AND toNullable(2147483648);
-
+100
 -- Negative integer: also differs in type and would truncate under a plain cast.
 SELECT count() FROM bug_103049 WHERE (c != '') AND toNullable(toInt8(-1));
-
+100
 -- NULL constant should still correctly filter ALL rows (0 rows).
 -- Three-valued logic: `x AND NULL` is NULL when x is true, which is falsy.
 SELECT count() FROM bug_103049 WHERE (c != '') AND CAST(NULL, 'Nullable(UInt16)');
-
+0
 DROP TABLE bug_103049;
-
--- Same bug surfaces when virtual-column filtering strips a column reference
--- out of the AND, reducing it to a single Nullable child.
--- system.tables calls filterBlockWithPredicate with only {database} as the
--- allowed input, so `name = 'one'` gets dropped and the Nullable constant is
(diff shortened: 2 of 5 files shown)


groeneai:
### Pre-PR Validation Gate (Session cron:clickhouse-ci-task-worker:20260418-183903)

```sql
CREATE TABLE bug_repro (c String) ENGINE = MergeTree ORDER BY tuple();
INSERT INTO bug_repro SELECT toString(number) FROM numbers(100);
SELECT count() FROM bug_repro WHERE (c != '') AND toNullable(256);
-- Returns 0 on master (wrong), 100 with the fix.
```
Reproduces on every invocation, no randomization required.

**b) Root cause explained?** Yes.
`splitFilterNodeForAllowedInputs` reduces a partially-allowed `AND` to its single surviving child.  When the child's result type is `Nullable(UInt16)` (for `toNullable(256)`) but the parent `AND` returned `Nullable(UInt8)`, the code inserts `notEquals(child, 0)` to produce a boolean.  The zero constant was built via `result_type->getDefault()`, which for `Nullable(T)` returns `Null()` — so the generated comparison is `notEquals(toNullable(256), NULL)` → `NULL` under three-valued logic → false → all rows filtered.  `N = 255` never hit the bug because `toNullable(255)` has type `Nullable(UInt8)`, equal to the parent's result type, bypassing the conversion branch entirely.

**c) Fix matches root cause?** Yes.
`removeNullable(result_type)->getDefault()` produces `UInt16(0)` instead of `NULL`, and `createColumnConst` re-wraps it into a `Nullable` holding the non-null zero.  The comparison becomes `notEquals(256, 0) = 1`, rows are kept.  `Nullable(Nothing)` has no meaningful nested zero, so we fall back to the outer default — this preserves the correct `notEquals(NULL, NULL) -> NULL -> false` semantics for bare `NULL` literals, which still correctly drop all rows.  No band-aids, no assertion widening.

**d) Test intent preserved? / New tests added?** Yes, new regression test.
`tests/queries/0_stateless/100000_103049_nullable_and_overflow_filter.sql` covers the full boundary: the non-bug baselines (`AND 256`, `AND toNullable(255)`), the core bug (`AND toNullable(256)`, `AND toNullable(65535)`, `AND toNullable(2147483648)`, `AND toNullable(toInt8(-1))`), and the `NULL`-propagation contract (`AND CAST(NULL, 'Nullable(UInt16)')` still returns 0 rows).  Both `MergeTree` (virtual-column part-pruning path) and `system.tables` (allowed-inputs filtering path) are exercised.

- Without fix (master unmodified): `tests/clickhouse-test 100000_103049_nullable_and_overflow_filter` fails with `-1 +0` diffs on every `toNullable(>= 256)` line and the negative-integer case.
- With fix: `tests/clickhouse-test --test-runs 20 100000_103049_nullable_and_overflow_filter` → 20/20 passes under full CI randomization.

The fix replaces the sole `getDefault` call in the boolean-conversion branch, which is the single point where all `splitFilterNodeForAllowedInputs` calls funnel through.  All call sites of `filterBlockWithPredicate` and `splitFilterDagForAllowedInputs` benefit equally — part pruning in `MergeTree`, virtual-column filtering in `system.tables`, partition pruning in distributed storages, etc.  No sibling implementations of the same pattern were found that still use the buggy `result_type->getDefault()` pattern for `Nullable` operands.

groeneai:
cc @CurtizJ @alexey-milovidov — could you review this? Fixes a wrong-results regression in MergeTree filter path: `WHERE x AND toNullable(N)` returns 0 rows for any `N >= 256` because `VirtualColumnUtils::splitFilterNodeForAllowedInputs` used `result_type->getDefault()` to build a zero constant, which returns `Null()` for `Nullable(T)`, turning `notEquals(x, NULL) -> NULL -> false` and dropping every row.  Same root cause as #101433 / #99979 / #101269.  One-line fix in the boolean-conversion branch + dedicated regression test covering the UInt8 overflow boundary and the `Nullable(Nothing)` edge case.

qoega (approved):

PR #99083, opened 2026-03-09
Improve TimeSeries table engine

vitlibar:
### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
This PR improves the TimeSeries table engine in several ways:

- Support `INNER COLUMNS` clause in `CREATE TABLE` for fine-grained control over inner table columns
- Add new setting `id_generator` to allow customizing ID generation even for a TimeSeries table with external targets
- Support `ALTER MODIFY SETTINGS` for TimeSeries tables
- Rename inner target table `DATA` to `SAMPLES`
- Simplify outer columns: replace `timestamp`/`value` pair with a single `time_series` column; remove redundant outer columns `id`, `all_tags`, `min_time`, `max_time`, per-tag columns; rename outer column `metric_family_name` to `metric_family`
- Support `INSERT` into TimeSeries tables

### Version info
- Merged into: `26.6.1.171`

PR #102358, opened 2026-04-10
Revert "Preserve original parameter types in timeseries aggregate functions"

tavplubix:
Reverts ClickHouse/ClickHouse#101083

cc @alexey-milovidov

Issue #102319, opened 2026-04-10
Preserved parameter types break `AggregateFunction` type equality when integer and `Decimal64` representations of same timestamp are mixed


clickgapai:
_Found via ClickGap automated review. Please close or comment if this is incorrect or needs adjustment._

### Describe what's wrong

INSERT into `AggregatingMergeTree` table with timeseries `AggregateFunction` column fails with `CANNOT_CONVERT_TYPE` when the table was created with integer timestamp parameters but the INSERT query uses `DateTime64` cast expressions for the same timestamps. The `-Merge` combinator similarly fails with `ILLEGAL_TYPE_OF_ARGUMENT`.

**Root cause:** AggregateFunctionTimeseriesBase.h:69: Passing original `parameters_` directly to the base class removes the canonicalization that previously guaranteed all parameters had the same `Field` type (`Decimal64`). Now, the same logical timestamp value can be stored as `UInt64(1704067200)` (from integer literal in CREATE TABLE) or `DecimalField<Decimal64>(1704067200000, 3)` (from `'2024-01-01'::DateTime64(3,'UTC')` in SELECT). These have different `Field::which` tags, so `Field::operator==` returns false even though they represent the same timestamp.

**Why we believe this is a bug:** `createWithValueType` (AggregateFunctionTimeseriesHelpers.cpp:210) passes original `parameters` to constructor → `AggregateFunctionTimeseriesBase` (AggregateFunctionTimeseriesBase.h:69) stores them as-is → `IAggregateFunction::getStateType` (IAggregateFunction.cpp:17) creates `DataTypeAggregateFunction` with these parameters → `DataTypeAggregateFunction::equals` (DataTypeAggregateFunction.cpp:197-206) calls `strictEquals` → line 170-172 compares with `Field::operator!=` → `Field::operator==` (Field.cpp:172) checks `which != rhs.which` → `UInt64 != Decimal64` → type mismatch

**Affected locations:**
- `src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesBase.h:69` — parameters_ passed directly to Base instead of canonicalized DecimalField<Decimal64>
- `src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesHelpers.cpp:210` — parameters passed to Function constructor
- `src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesHelpers.cpp:215` — parameters passed to Function constructor (non-predict path)
[...]

PR #103079, opened 2026-04-18
Fix LOGICAL_ERROR in AggregateFunctionArray for timeSeries*ToGrid functions

groeneai:
The `timeSeries*ToGrid` aggregate function family normalizes its four timestamp parameters internally to `Decimal64` (scaled to match the timestamp column) and, for `timeSeriesPredictLinearToGrid`, stores `predict_offset` as a member instead of in the parameter array. This normalization is intentional — it lets states serialized with different timestamp scales merge together.

When the `Array` combinator wrapped such a function, its constructor fired `LOGICAL_ERROR` (`AggregateFunctionArray: parameters mismatch`): it was initializing its own parameter array from the user-typed `params_` (`UInt64`, sometimes 5 entries) but then comparing against the nested function's normalized parameters (`Decimal64`, always 4 entries). The assertion was originally a `chassert` that compiled out in release; #102868 promoted it to a real `LOGICAL_ERROR` for diagnostics, which then surfaced as a master-level exception in serverfuzz / `AST fuzzer` stress runs (STID 4870 family, ~10 hits across 60 days; ~7 master, ~20 PR hits in the last 30 days).

Fix the `Array` combinator by initializing its own parameter array from `nested_->getParameters()` instead of the user-supplied `params_`. The `Array` wrapper is just an adapter over the nested function — its parameters should match the nested function's. Almost every aggregate function preserves `params_` verbatim, so `nested_->getParameters() == params_` and there is no behaviour change. For the `timeSeries*ToGrid` family the two differ, and using the nested function's parameters makes `timeSeries*ToGridArray(...)` states byte-compatible with the unwrapped `timeSeries*ToGrid(...)` states of the same scale.

Reproducer (exception before, succeeds after):

```sql
SET allow_experimental_time_series_aggregate_functions = 1;
CREATE TABLE ts (ts Array(DateTime64(3, 'UTC')), v Array(Float64)) ENGINE=Memory;
INSERT INTO ts VALUES ([toDateTime64('1970-01-01 00:01:00', 3, 'UTC')], [10.0]);
SELECT timeSeriesPredictLinearToGridArray(60, 180, 10, 20, -60)(ts, v) FROM ts;
SELECT timeSeriesDerivToGridArray(60, 180, 10, 20)(ts, v) FROM ts;
```

Regression test: `04102_timeseries_array_combinator_params`.

Related:
- Diagnostic PR that made the error visible: #102868
- STID 4870 family (4870-4f07, 4870-4f21, 4870-4d73, 5480-49e6).
- Context: triggered most often by `AST fuzzer (amd_debug, targeted)` on PRs that touch `TimeSeries*` (e.g. #99083), which reshape the fuzzer's exploration space.

Fix `Logical error: AggregateFunctionArray: parameters mismatch` when `timeSeries*ToGrid` aggregate functions are used with the `Array` combinator (e.g. `timeSeriesPredictLinearToGridArray`, `timeSeriesDerivToGridArray`, `timeSeriesRateToGridArray`). The `Array` wrapper was storing the user-typed parameters while the nested function stored normalized `Decimal64` parameters, so their types and counts could differ. The wrapper now inherits the nested function's parameters, matching the behaviour of the unwrapped function.


groeneai:
## Pre-PR validation gate (self-check)

**Session:** cron:clickhouse-ci-task-worker:20260418-204500

**a) Deterministic repro?** Yes. `SELECT timeSeriesPredictLinearToGridArray(60, 180, 10, 20, -60)(ts, v) FROM ts;` throws `LOGICAL_ERROR` on every run of the stock master binary without the fix (the diagnostic error message added in #102868); my local reproduction shows the exact stack trace matching the CI reports.

**b) Root cause explained?** The `Array` combinator asserted `parameters == nested_func->getParameters()`. The `timeSeries*ToGrid` family stores internally-normalized `Decimal64` parameters (scale = column's timestamp scale), and the 5-arg `timeSeriesPredictLinearToGrid` stores `predict_offset` as a member instead of in the parameter array. So the wrapper held `[UInt64_60, UInt64_180, UInt64_10, UInt64_20, Int64_-60]` (5 entries, user-typed) while the nested function held `[Decimal64_'60', Decimal64_'180', Decimal64_'10', Decimal64_'20']` (4 entries, normalized) — mismatch in both count and type.

**c) Fix matches root cause?** Yes. The `Array` wrapper is just an adapter — its parameters should match the nested function's, not the user's. Initializing from `nested_->getParameters()` removes the mismatch at the source (not by widening the check or silencing it).

**d) Test intent preserved? / New tests added?** No existing tests loosened. Added a new regression test `04102_timeseries_array_combinator_params` that covers: 4-arg variants with `UInt64`/`Decimal`/`String` literals, the 5-arg `predict` variant, stacked combinators (`ArrayIf`), and the non-Array reference path. Ran it 20× with full randomization — 20/20 pass.

**e) Demonstrated in both directions?** Yes. Manual test confirmed: the reproducer query throws `LOGICAL_ERROR` on master HEAD (without the fix, built 19:18 UTC), and passes with the patch applied. The regression test's reference file is only generatable from the fixed binary.

**f) Fix is general, not a narrow patch?** Yes. The change applies to the `Array` combinator universally — not just to `timeSeries*ToGrid`. Other aggregate functions preserve `params_` verbatim, so `nested_->getParameters() == params_` for them and behaviour is unchanged. Verified by running 10+ unrelated Array-combinator tests (`sumArray`, `quantilesArray`, `uniqArray`, etc.) — all pass.

I kept the scope minimal: only `AggregateFunctionArray` changed. I explicitly did not touch `AggregateFunctionTimeseriesBase` itself because its `Decimal64` normalization is intentional — it is what allows states serialized with different timestamp scales to merge (tested in `03254_timeseries_instant_value_aggregate_functions`), and the existing plain-call path depends on it.

groeneai:
cc @alexey-milovidov @Algunenano @vitlibar — could you review this? It fixes `LOGICAL_ERROR: AggregateFunctionArray parameters mismatch` (STID 4870 family on master serverfuzz / AST fuzzer) by having the `Array` combinator inherit the nested function's parameters instead of the user-typed ones. This is intentionally scoped narrowly to the `Array` combinator — I deliberately do **not** touch `AggregateFunctionTimeseriesBase`'s `Decimal64` normalization, since #102358 made it clear that normalization is required for state-merge compatibility across timestamp scales. Issue #102319 (the deeper CREATE TABLE vs INSERT type-equality problem) is unchanged.

vitlibar:
Doesn't look good for now, needs investigation.
I will look into it.

vitlibar:
The issue will be fixed in https://github.com/ClickHouse/ClickHouse/pull/103428, this PR is not needed.

PR #103080, opened 2026-04-18
Apply max_network_bandwidth_for_{user,all_users} to remote filesystem reads/writes

azat:
- Improvement

Apply `max_network_bandwidth_for_user`/`max_network_bandwidth_for_all_users` to remote filesystem reads/writes

`max_network_bandwidth_for_user` and `max_network_bandwidth_for_all_users` used to limit only the distributed network layer. Extend them so that the same per-user / per-all-users bandwidth caps also apply to reads/writes from remote filesystems (e.g. S3), in addition to any per-query `max_remote_read_network_bandwidth` limit on top.

_Note: historically we use separate settings for limiting reading from remote, but I did not want to introduce more extra settings, curious what your opinion on this_

### Version info
- Merged into: `26.5.1.13` (included in `26.5` and later)
- Backported to: `26.4.5.108`, `26.3.10.32`


CheSema (approved):
add user's throttler is ok.

azat:
Actually this implementation (and non global throttling in general) may not work, since with current implementation it can lead to blocking thread pool that is used for reads, so if some user will throttle, others may be blocked by it (if the thread pool will not be big enough), ideally, we need something better then just sleep before read.

Though the same problem we have everywhere...

src/Interpreters/Context.cpp
serxa, line 4781: There is a bug. We save array of throttlers containing user specific throttler into server-wide throttler

serxa (changes requested):

serxa:
> Though the same problem we have everywhere...

Yes, this is a problem but it is unrelated problem, I would say. For now, the best thing we can rely on is that number of queries can be limited for user. And number of IO threads is limited per query.

And yes, indeed this PR will not help to resolve it. Independent per user limits will only create more waiting threads.

src/Interpreters/Context.cpp
azat: Nice catch! Interesting, CI should became significantly slower then, everything after `04103_user_network_bandwidth_throttler.sh`

I see, it does not because we don't have server-wide throttling enabled on CI

azat:
>Independent per user limits will only create more waiting threads

The same applies for limiting bandwidth via workloads, right?

serxa:
Yes, unfortunately, this is a consequence of having a thread per IO request.

src/Interpreters/Context.cpp
azat: Heh, actually, it is the opposite, the "bug" was only in case of the system wide throttler was not set, and CI did not find anything, because there is nothing to look, everything is correct, `addThrottler` does not update the throttler inside, only shared_ptr

serxa: Ah, indeed, sorry for misleading tip

serxa (approved):

azat:
>CH Inc sync — tests failed

Related, will be fixed in the sync PR

azat:
Backports will include https://github.com/ClickHouse/ClickHouse/pull/103387 as well

CheSema:
https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICcwNDEwM191c2VyX25ldHdvcmtfYmFuZHdpZHRoX3Rocm90dGxlcicKICAgIC0tIEFORCBjaGVja19uYW1lID0gJ1N0YXRlbGVzcyB0ZXN0cyAoYW1kX3RzYW4sIHBhcmFsbGVsLCAyLzIpJwogICAgQU5EIHRlc3Rfc3RhdHVzIElOICgnRkFJTCcsICdFUlJPUicpCiAgICBBTkQgKHB1bGxfcmVxdWVzdF9udW1iZXIgPSAwIE9SIGJhc2VfcmVmIElOICgnbWFzdGVyJykpCiAgICBBTkQgdGVzdF9jb250ZXh0X3JhdyBMSUtFICclcmVzdWx0IGRpZmZlcnMgd2l0aCByZWZlcmVuY2UlJwpHUk9VUCBCWSBkYXkKT1JERVIgQlkgZGF5IERFU0MK

Please check `04103_user_network_bandwidth_throttler` test, it started to flack.

PR #103081, opened 2026-04-18
keeper-bench: add KeeperStorage-only benchmark mode

al13n321:
Human note: I mostly didn't read the code, but it's easy to delete and appears to work well, with outputs like:
```
==== period 1.0s ====
writes/s: 56088.0, reads/s: 369387.2
util: preprocess 58.6%, commit(writes) 20.4%, commit(reads) 47.9%
znodes: 104485, rss: 1156 MiB, snapshot_mode: no
per-op avg ns per request:
  create: count=8323 process=3114 ns preprocess=5492 ns
  remove: count=8287 process=3288 ns preprocess=6236 ns
  get: count=330083 process=1292 ns
  set: count=32891 process=3271 ns preprocess=11132 ns
  multi: count=6618 process=6485 ns preprocess=18501 ns
  multiread: count=32934 process=1321 ns
  filteredlist: count=6574 process=1364 ns (1861394 names, 4 ns/name)
```
Seems much nicer for profiling and optimizing the data structures than running a whole server.

---

Adds a new `--storage` mode to `clickhouse keeper-bench` that exercises `KeeperStorage` in-process, without the rest of the keeper server (no network, no raft, no state machine).

Threading model mirrors a simplified Keeper commit path:
* **N generator threads** (`--concurrency`, default 4) produce requests using the existing `Generator` — so the same `setup` and `generator` config sections work.
* **One preprocess thread** calls `preprocessRequest` for every write (stands in for the path that, in real keeper, runs under `raft_server::lock_`).
* **One commit thread** calls `processRequest` for writes. After each committed write, with probability `1/writes_per_read_batch` it drains the pending read queue and processes it in one `processLocalRequests` batch.

Inter-thread queues are `NonblockingBoundedQueue` (copied from 6a91c301155ed1c0463cb2e91cbc70d140fc8cff) with 100µs polling — no condition variables. The preprocess→commit queue has a small configurable bound (`storage.preprocess_commit_queue_size`, default 32).

Periodic reports print (for the last period): committed writes/s and reads/s, preprocess / commit-write / commit-read thread utilization, znode count, RSS, whether snapshot mode was enabled during the period, and per-op-num average preprocess/process ns and ns-per-list-name. Every `storage.snapshot_toggle_periods` periods the runner toggles `enableSnapshotMode`/`disableSnapshotMode` + `clearGarbageAfterSnapshot` so snapshot-mode overhead shows up in the timings.

`Generator`/`PathGetter` previously took a `Coordination::ZooKeeper &` to resolve `children_of` paths. They now take a `ListChildrenFn` callback; the existing `Runner` passes a lambda wrapping the ZooKeeper client, and the new `StorageRunner` passes a lambda that goes through `KeeperStorage::processLocalRequests`. That's the only user-visible change to the existing mode.

---

Original prompt:

> Please make a benchmark for KeeperStorage.
>  * Reuse keeper-bench's Generator code, so \`setup\` and \`generator\` config with the same syntax can be used.
>  * Do all \`preprocess\` in one thread (in real keeper they would be under raft_server::lock_), all \`process/processLocal\` in another thread (simulating commit thread in real keeper), request generation in n (4 by default) other threads.
>  * Copy src/Common/NonblockingBoundedQueue.h from 6a91c301155ed1c0463cb2e91cbc70d140fc8cff, use it for passing requests from generator threads to the preprocess thread. Another queue for passing local (read) requests from generator to commit thread. Another queue for passing requests from preprocess thread to commit thread, with a small configurable size limit. When input queue is empty or output queue is full, just poll with 100us sleep, instead of condition_variable or anything.
>  * In commit thread, after each committed write request, with probability 1/writes_per_read_batch (new config parameter) drain the read requests queue and execute all requests as one batch; otherwise don't do reads.
>  * Time each operation, periodically print the following stats (over the last period, not for the whole run so far):
>     - committed writes/s, reads/s,
>     - preprocess thread util%, commit thread write requests util% (fraction of time spent doing write requests), read requests util%,
>     - memory usage (rss), znode count,
>     - avg ns per request for each request type separately,
>     - avg ns per list request's output entry (i.e. total duration of list requests / total number of paths in responses to those requests),
>     - whether we were in snapshot mode during this period.
>  * Every n periods toggle snapshot mode (new config parameter).
>  * Preferably make it a new keeper-bench mode rather than a new executable (but don't require \`connections\` config in that mode). Preferably controlled by a command line switch rather than config contents, so the same config file can be used in both modes. Put the code in a new file rather than keeper-bench/Runner.cpp.

---

Add `--storage` mode to `clickhouse keeper-bench` that benchmarks `KeeperStorage` in-process (no network, no raft, no state machine), using the same `setup`/`generator` config sections as the network mode.

### Version info
- Merged into: `26.6.1.1018`


programs/keeper-bench/StorageRunner.cpp
```diff
@@ -0,0 +1,644 @@
...
+        auto request = std::make_shared<Coordination::ZooKeeperCreateRequest>();
+        request->path = path;
+        request->data = node.data ? node.data->getString() : "";
+        request->acls = default_acls;
+
+        int64_t zxid = next_zxid.fetch_add(1);
+        try
+        {
+            storage->preprocessRequest(request, setup_session_id, 0, zxid);
+            storage->processRequest(request, setup_session_id, zxid);
```
antonio2368: Do we want to check just in case if returned value is ZOK?

programs/keeper-bench/StorageRunner.cpp
```diff
@@ -0,0 +1,644 @@
...
+        auto & s = period_stats.per_op[slot];
+        s.count.fetch_add(1, std::memory_order_relaxed);
+        s.process_ns.fetch_add(elapsed, std::memory_order_relaxed);
+
+        period_stats.writes_committed.fetch_add(1, std::memory_order_relaxed);
+
+        /// Fire generator callbacks based on response error (so CreateRequestGenerator's
+        /// paths_created bookkeeping stays consistent with storage state).
+        bool success = !responses.empty() && responses.front().response
+            && responses.front().response->error == Coordination::Error::ZOK;
```
antonio2368: Multi should return ZOK in all cases

programs/keeper-bench/StorageRunner.cpp
```diff
@@ -0,0 +1,644 @@
...
+                    storage->clearGarbageAfterSnapshot();
+                    snapshot_enabled.store(false);
+                }
+                else
+                {
+                    auto version = storage->container.snapshotSizeWithVersion().second;
+                    storage->enableSnapshotMode(version);
+                    snapshot_enabled.store(true);
+                }
+            }
```
antonio2368, line 663: I don't think is thread safe

antonio2368 (commented):
Update README

antonio2368 (approved):
Just address review bot comments.

al13n321:
Overrode CI after making a little progress investigating the flaky test (it was https://github.com/ClickHouse/ClickHouse/pull/107932 iirc).

PR #103084, opened 2026-04-19
Fix QBit column structureEquals (logical error in writeSlice)

groeneai:
## Motivation

Master `BuzzHouse (experimental, serverfuzz, amd_msan)` hit a logical error in `writeSlice` on 2026-04-17 (STID `2670-0e56`, commit `4c74bee46e8b`):

```
Logical error: Function writeSlice expects same column types for
GenericArraySlice and GenericArraySink.
```

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=4c74bee46e8b3d257df5320ac90f735593ef61bc&name_0=MasterCI&name_1=BuzzHouse%20%28experimental%2C%20serverfuzz%2C%20amd_msan%29

## Root cause

`ColumnQBit::structureEquals` ignored the right-hand side `ColumnQBit` wrapper:

```cpp
bool structureEquals(const IColumn & rhs) const override { return tuple->structureEquals(rhs); }
```

`this->tuple` is a `ColumnTuple`; `rhs` is another `ColumnQBit`. `ColumnTuple::structureEquals` starts with `typeid_cast<const ColumnTuple *>(&rhs)` and returns `false` immediately when the argument is any other column type. As a result, `QBit` columns compared against themselves always reported "different structure", breaking every consumer that uses `structureEquals` to validate column compatibility.

The visible symptom on master was a fuzzed INSERT pattern:

```sql
CREATE TABLE t (c0 Tuple(Map(String, Nullable(QBit(Float64, 4))))) ENGINE = Memory;
INSERT INTO t SELECT NULL;
```

`ConvertingTransform` rewrites this as `ifNull(NULL::Nullable(Nothing), default_tuple)`. The dispatch descends through `FunctionIfNull::executeImpl` -> `if` -> `FunctionIf::executeTuple` -> `executeImpl` -> `FunctionIf::executeMap` -> `executeImpl` -> `FunctionIf::executeGenericArray` -> `GatherUtils::conditional` -> `writeSlice`. `writeSlice(GenericArraySlice, GenericArraySink)` at `src/Functions/GatherUtils/Algorithms.h:75-84` checks `slice.elements->structureEquals(sink.elements)` and throws on `false`. The slice elements and sink elements are both `ColumnNullable(ColumnQBit)`, structurally identical, but the broken `ColumnQBit::structureEquals` made the check fail.

## Fix

Follow the standard wrapper-column pattern used by `ColumnSparse`, `ColumnReplicated`, `ColumnLowCardinality`, and `ColumnNullable`: `typeid_cast` the `rhs` to `ColumnQBit` first, then delegate the inner comparison to the unwrapped `tuple` on both sides.

Also require `dimension` to match: two `QBit` columns with different dimensions can pad to the same `FixedString` byte length (e.g. dimension `1` and dimension `8` both pad to 1 byte), so the inner tuple structure alone does not distinguish them.

## Verification

**Before fix** — server crashes on the minimal reproducer:
```
Logical error: 'Function writeSlice expects same column types for GenericArraySlice and GenericArraySink.'
```

**After fix** — same queries succeed and the server stays alive:
```
({})                          -- Tuple(Map(String, Nullable(QBit(Float64, 4))))
(NULL,'a',{})                 -- fuzzer-shaped Tuple(Nullable(Int8), Enum16(...), Map(MultiLineString, Nullable(QBit(...))))
({})                          -- ifNull(NULL::Nullable(Tuple(...)), tuple(map()))
```

Existing QBit stateless tests (`03363`...`03379`) all pass (9/9).

Regression test added: `tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.{sql,reference}`.

## Scope

The change is localized to `ColumnQBit::structureEquals`. No behavior change for existing `QBit` workloads that already worked; only operations that previously failed with `structureEquals`-based validation will now proceed correctly.

Fix a logical error `Function writeSlice expects same column types for GenericArraySlice and GenericArraySink` raised when evaluating `if`/`ifNull` over tuples, maps, or arrays containing a `QBit` element. `ColumnQBit::structureEquals` incorrectly compared the inner tuple of one `QBit` column against the outer wrapper of the other, so structurally identical `QBit` columns were reported as different.

### Version info
- Merged into: `26.5.1.8`


groeneai:
### Pre-PR validation gate (self-review)

**a) Deterministic repro?** Yes. The exact minimal query crashes the server every single time without the fix:
```sql
CREATE TABLE t (c0 Tuple(Map(String, Nullable(QBit(Float64, 4))))) ENGINE = Memory;
INSERT INTO t SELECT NULL;
```
Raises `Logical error: Function writeSlice expects same column types for GenericArraySlice and GenericArraySink` at `src/Functions/GatherUtils/Algorithms.h:83`.

**b) Root cause explained?** Yes. `ColumnQBit::structureEquals(rhs)` called `this->tuple->structureEquals(rhs)`. `this->tuple` is a `ColumnTuple`, but `rhs` is the outer `ColumnQBit` wrapper — so `ColumnTuple::structureEquals` fails its `typeid_cast<const ColumnTuple *>(&rhs)` check and always returns `false`. Any two `ColumnQBit` columns therefore report different structure, and `writeSlice(GenericArraySlice, GenericArraySink)` throws on `structureEquals == false`.

**c) Fix matches root cause?** Yes. The fix restores the standard wrapper-column pattern (see `ColumnSparse`, `ColumnReplicated`, `ColumnLowCardinality`, `ColumnNullable`): `typeid_cast` `rhs` to `ColumnQBit`, then compare the inner tuples of both sides and the `dimension` field. Not a band-aid in `executeGenericArray` — the defect is in `ColumnQBit` itself, and fixing it at the source benefits every other consumer of `structureEquals`.

**d) Test intent preserved / new tests added?** Yes. All existing QBit stateless tests (`03363`..`03379`, 9 tests) still pass. New regression test `tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.{sql,reference}` exercises the three surfaced manifestations (minimal Tuple-over-Map-over-QBit, fuzzer-shaped tuple, and direct `ifNull` over `Nullable(Tuple(...QBit...))`).

- Without the fix (`build/` pre-edit): server crashes with the `writeSlice` logical error on every attempt.
- With the fix (`build/` post-edit): all three test cases return the expected rows and the server continues running.

**f) Fix is general, not a narrow patch?** Yes. The fix addresses the underlying broken invariant in `ColumnQBit::structureEquals` rather than guarding a single crash site. Any code path that relies on `structureEquals` over `QBit` columns — not only `writeSlice(GenericArraySlice, GenericArraySink)` at `Algorithms.h:83`, but also `writeSlice(GenericValueSlice, GenericArraySink)` at `Algorithms.h:156` and other validators — now behaves correctly. Audited the rest of `ColumnQBit`: every other method that accepts an `IColumn &` already does `assert_cast<const ColumnQBit &>(src_)` first, so this was the single missing spot.

**Session:** cron:clickhouse-ci-task-worker:20260419-021500

groeneai:
cc @rienath @Avogar @alexey-milovidov — could you review this? Fixes a logical error in `writeSlice` reaching master via BuzzHouse (STID 2670-0e56) by restoring the standard `typeid_cast`-first pattern in `ColumnQBit::structureEquals` (it was delegating to the inner tuple's `structureEquals` with the outer `ColumnQBit` as argument, which always returned false for matching QBit columns).

tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.sql
```diff
@@ -0,0 +1,40 @@
+-- Tags: no-fasttest
+-- Tag `no-fasttest`: `QBit` depends on non-fasttest libraries.
```
rienath: What's the library `QBit` depends on that is not available in fasttest?

tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.sql
```diff
@@ -0,0 +1,40 @@
...
+-- `LOGICAL_ERROR` when evaluating `if`/`ifNull` over tuples or maps containing a `QBit`
+-- element.
+--
+-- Reproducer: inserting `NULL` into a column whose target type contains a `QBit` nested
+-- inside a `Map` inside a `Tuple` goes through `ConvertingTransform` -> `ifNull` -> `if`
+-- -> `executeTuple` -> `executeMap` -> `executeGenericArray` -> `writeSlice`. Before
+-- the fix this raised the exception
+--   Logical error: Function writeSlice expects same column types for
+--   GenericArraySlice and GenericArraySink.
+-- After the fix the row is converted to the default tuple value and written normally.
```
rienath: This long description is not necessary. Simple one-liner would worked

tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.sql
```diff
@@ -0,0 +1,40 @@
...
+
+DROP TABLE IF EXISTS t_qbit_ws;
+CREATE TABLE t_qbit_ws (c0 Tuple(Map(String, Nullable(QBit(Float64, 4))))) ENGINE = Memory;
+INSERT INTO t_qbit_ws SELECT NULL;
+SELECT * FROM t_qbit_ws;
+DROP TABLE t_qbit_ws;
+
+-- Same pattern with a richer tuple shape as produced by the BuzzHouse fuzzer that
+-- originally surfaced the bug (STID 2670-0e56).
```
rienath: See above. Potentially delete

tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.sql
```diff
@@ -0,0 +1,40 @@
...
+DROP TABLE IF EXISTS t_qbit_ws_full;
+CREATE TABLE t_qbit_ws_full (
+    c0 Tuple(Nullable(Int8), Enum16('a' = 1, 'b' = 2), Map(MultiLineString, Nullable(QBit(Float64, 4))))
+) ENGINE = Memory;
+INSERT INTO t_qbit_ws_full SELECT NULL;
+SELECT * FROM t_qbit_ws_full;
+DROP TABLE t_qbit_ws_full;
+
+-- `ifNull` over a tuple containing `QBit` directly should also succeed now that
+-- `ColumnQBit::structureEquals` correctly compares two `ColumnQBit` columns.
```
rienath: See above

rienath (changes requested):

tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.sql
```diff
@@ -0,0 +1,40 @@
...
+SET allow_experimental_nullable_tuple_type = 1;
+SELECT ifNull(CAST(NULL AS Nullable(Tuple(Map(String, Nullable(QBit(Float32, 8)))))),
+              CAST(tuple(map()) AS Tuple(Map(String, Nullable(QBit(Float32, 8))))));
```
rienath, line 22: Tests only cover positive cases. Let's add a negative test too. For example, with non-matching dimension

tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.sql
rienath: @groeneai please take a look at the review

groeneai:
@rienath thanks for the review, pushed [`319fe24`](https://github.com/ClickHouse/ClickHouse/pull/103084/commits/319fe24d7b7d9bad5a0e19957c495ad87cd5ed4b) addressing all five comments:

- Dropped the `no-fasttest` tag (`QBit` has no fasttest-excluded deps, matching 03363–03374).
- Collapsed the long regression-test description into three lines.
- Deleted the inline comments above the full-shape and `ifNull` queries.
- Added three negative cases at the end that `CAST` between `QBit` columns with non-matching dimensions. Dimensions 1 and 8 are the specific boundary the dimension check protects: both pad to a single byte in the underlying `FixedString` storage, so without comparing `dimension` the two columns would look structurally identical. All three `CAST`s fail cleanly with `TYPE_MISMATCH`.

Session: cron:clickhouse-ci-task-worker:20260420-121500

tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.sql
```diff
@@ -0,0 +1,29 @@
...
+
+-- Negative cases: QBits with non-matching dimensions are different types and must
+-- not be conflated. Dimensions 1 and 8 are particularly interesting because both
+-- pad to a single byte in the underlying `FixedString` storage, so the dimension
+-- field is what `structureEquals` relies on to tell them apart.
+SELECT CAST(CAST([1.0] AS QBit(Float64, 1)) AS QBit(Float64, 8)); -- { serverError TYPE_MISMATCH }
+SELECT CAST(CAST([1., 2., 3., 4., 5., 6., 7., 8.] AS QBit(Float64, 8)) AS QBit(Float64, 1)); -- { serverError TYPE_MISMATCH }
+SELECT CAST(CAST([1., 2., 3., 4.] AS QBit(Float64, 4)) AS QBit(Float64, 8)); -- { serverError TYPE_MISMATCH }
```
rienath: This doesn't test `structureEquals` path, does it? @groeneai

groeneai:
@rienath you're right — the `CAST(QBit(d1) AS QBit(d2))` cases never reach `ColumnQBit::structureEquals`; `FunctionCast` bails out at `DataTypeQBit::equals` before any column-level comparison runs. Pushed [`cdbff1e`](https://github.com/ClickHouse/ClickHouse/pull/103084/commits/cdbff1ecbcc69f0d2b4b1e7e54a7e6c2a1f4795a2) replacing them with a unit test that exercises the method directly.

New `src/Columns/tests/gtest_column_qbit.cpp` (four cases):

- `StructureEqualsSameDimensionIsTrue` — regression test for the original bug; fails against the old `return tuple->structureEquals(rhs)` because that typeid-casts the `ColumnTuple` against the outer `ColumnQBit` and always returns false.
- `StructureEqualsDifferentDimensionIsFalse` — the negative case you asked for. Uses `QBit(Float64, 1)` vs `QBit(Float64, 8)` specifically because both pad to a 1-byte `FixedString`, so the inner `ColumnTuple::structureEquals` would happily claim they match if the `dimension ==` guard were missing. Also covers 4-vs-8 as a routine mismatch.
- `StructureEqualsDifferentElementTypeIsFalse` — `Float32` vs `Float64` give different tuple widths; defensive coverage.
- `StructureEqualsNonQBitRhsIsFalse` — compares against a `ColumnString`, proving the `typeid_cast` guard rejects non-QBit rhs.

I verified both directions locally with `build/src/unit_tests_dbms --gtest_filter='ColumnQBit.*'`:

| State of `ColumnQBit::structureEquals` | Failing tests |
|---|---|
| old buggy `return tuple->structureEquals(rhs)` | `StructureEqualsSameDimensionIsTrue` fails (actual: false, expected: true) |
| typeid-cast only, no `dimension ==` guard | `StructureEqualsDifferentDimensionIsFalse` fails — `d1->structureEquals(d8)` and `d4->structureEquals(d8)` return `true` |
| full fix | all 4 pass |

`tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.sql` keeps the positive `ifNull`/`INSERT` scenarios that exercised the original `writeSlice` crash and points at the gtest for direct structureEquals coverage; the CAST block is dropped.

Session: `cron:clickhouse-ci-task-worker:20260420-141500`

rienath:
@groeneai delete the gtest, it is artificial. It is not clear how to test for size mismatch through SQL. Let's leave the dimension check as is for safety. Do not modify other files. Just delete the gtest

groeneai:
@rienath done in 7bbb4a8b10c — `src/Columns/tests/gtest_column_qbit.cpp` deleted, nothing else changed. `ColumnQBit::structureEquals` (typeid_cast + `dimension ==` check) is untouched; `tests/queries/0_stateless/04104_qbit_writeslice_structure_equals.{sql,reference}` remain as the regression test for the original `writeSlice` logical error.

Verified locally (Clang 21 debug, Linux):
- `ninja -C build unit_tests_dbms` — links cleanly after deletion
- `ninja -C build clickhouse` — rebuilt
- `04104_qbit_writeslice_structure_equals` — 25/25 OK with randomization
- full QBit suite (16 tests: 03363–03380, 03730, 03926, 04104) — 16/16 OK

Ready for re-review.

rienath (approved):

clickgapai:
Hi — this PR may need backporting to `26.3` (LTS), `26.2`, `26.1`, but no backport label was found.

**Affected code:** `ColumnQBit::structureEquals` in src/Columns/ColumnQBit.h — introduced in 25.10.

**Why:** `ColumnQBit` was added in commit 45c1024d391a (2025-09-30), first appearing in the 25.10 release cycle. The `allow_experimental_qbit_type` setting was changed from `false` to `true` in 26.1 (moved to Beta), meaning QBit is enabled by default starting from 26.1. The broken `structureEquals` has existed since the original addition. Branch 25.8 does not contain QBit code at all.

Other supported branches (`25.8`) predate 25.10 and do not contain this code.

PR #103086, opened 2026-04-19
Fix UBSan UInt64 overflow when scaling `interactive_delay` by sqrt(fanout)

groeneai:
`MultiplexedConnections::sendQuery` and `HedgedConnections::sendQuery` scale
`interactive_delay` by `sqrt(total_fanout)` with a per-connection jitter in
`[1.0, 2.0)` to reduce progress/profile event traffic from distributed queries
(introduced in f05a4566b0f, "Increase `interactive_delay` by sqrt(fanout) for
distributed queries", #100145).

The scaled value was cast back to `UInt64` without a bounds check. Since
`interactive_delay` is user-controlled and unbounded, the product
`delay * scale * jitter` as a `double` can exceed `UINT64_MAX`, and
`static_cast<UInt64>` on an out-of-range `double` is undefined behavior.
This triggers a UBSan trap in `Stress test (experimental, serverfuzz, arm_asan_ubsan)`:

```
src/Client/MultiplexedConnections.cpp:176:41: runtime error:
1.86135e+19 is outside the range of representable values of type 'unsigned long'

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior
```

(STID `3646-3e82`, seen on master commit `2892de442814d1e629f7ea1e25ed88d7c6441297`
on 2026-04-17 — [CI report](https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=2892de442814d1e629f7ea1e25ed88d7c6441297&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_asan_ubsan%29).)

### Approach

Factor the scaling logic into `scaleInteractiveDelayByFanout` in a new shared
header used by both `MultiplexedConnections` and `HedgedConnections` (the same
block of code was duplicated in both files, per the pre-existing `"same as in
MultiplexedConnections"` comment).

Saturate to `std::numeric_limits<UInt64>::max()` when the scaled `double` is
`>= 2^64`. `2^64` is exactly representable in `double`, so the comparison is
reliable:

```cpp
const double scaled = std::ceil(static_cast<double>(delay) * scale * jitter);
constexpr double two_pow_64 = 18446744073709551616.0;
if (scaled >= two_pow_64)
    return std::numeric_limits<UInt64>::max();
return static_cast<UInt64>(scaled);
```

### Verification

`gtest_scale_interactive_delay_by_fanout` covers:

- `NoScalingForSingleConnection` — `total_fanout <= 1` returns the value unchanged
- `ScalesWithinJitterBounds` — 1000 iterations validating the scaled result stays
  within `[delay * scale, delay * scale * 2]`
- `SaturatesOnOverflow` — the regression: inputs that previously triggered the
  UBSan cast now return `UINT64_MAX`
- `ZeroDelayStaysZero` — `delay == 0` invariant across any fanout

Without the saturation check, `SaturatesOnOverflow` fails deterministically —
`static_cast<UInt64>` of a too-large `double` returns `0x8000000000000000`
(`INT_INDEFINITE` on x86_64) instead of `UINT64_MAX`. Verified locally:

```
[ RUN      ] ScaleInteractiveDelayByFanout.SaturatesOnOverflow
gtest_scale_interactive_delay_by_fanout.cpp:51: Failure
  scaleInteractiveDelayByFanout(uint64_max / 2 + 1, 4): 9223372036854775808
  uint64_max:                                          18446744073709551615
[  FAILED  ] ScaleInteractiveDelayByFanout.SaturatesOnOverflow
```

With the fix: `4/4` pass.

...

### Version info
- Merged into: `26.5.1.20`

### DIFF
--- a/src/Client/scaleInteractiveDelayByFanout.h
+++ b/src/Client/scaleInteractiveDelayByFanout.h
@@ -1,0 +1,42 @@
+#pragma once
+
+#include <cmath>
+#include <cstddef>
+#include <limits>
+
+#include <base/types.h>
+#include <Common/thread_local_rng.h>
+
+
+namespace DB
+{
+
+/// Scale `interactive_delay` by `sqrt(total_fanout)` with per-connection jitter in `[1.0, 2.0)`,
+/// to reduce progress/profile event traffic from distributed queries. Each remote server will
+/// send updates less frequently, proportional to the square root of the total number of remote
+/// connections; the jitter desynchronizes progress reports across connections to avoid TCP incast
+/// and make the progress bar smoother.
+///
+/// The scaled value is saturated to `UINT64_MAX` on overflow: user-controlled `interactive_delay`
+/// is an unbounded `UInt64`, so `delay * scale * jitter` as a `double` can exceed the `UInt64`
+/// range. Casting such a `double` back to `UInt64` without saturation is undefined behavior
+/// (UBSan finding in `Stress test (experimental, serverfuzz, arm_asan_ubsan)`).
+inline UInt64 scaleInteractiveDelayByFanout(UInt64 delay, size_t total_fanout)
+{
+    if (total_fanout <= 1)
+        return delay;
+
+    const double scale = std::sqrt(static_cast<double>(total_fanout));
+    /// Random jitter in range [1.0, 2.0).
+    const double jitter = 1.0 + (thread_local_rng() % 1000) / 1000.0;
+    const double scaled = std::ceil(static_cast<double>(delay) * scale * jitter);
+
+    /// `static_cast<UInt64>(x)` is undefined behavior when `x` is outside the `UInt64` range.
+    /// `2^64` is exactly representable as `double`, so comparing against it is reliable.
+    constexpr double two_pow_64 = 18446744073709551616.0;
+    if (scaled >= two_pow_64)
+        return std::numeric_limits<UInt64>::max();
+    return static_cast<UInt64>(scaled);
+}
+
+}
--- a/src/Client/HedgedConnections.cpp
+++ b/src/Client/HedgedConnections.cpp
@@ -1,10 +1,9 @@
 #include <Core/Protocol.h>
 #if defined(OS_LINUX)
 
-#include <cmath>
 #include <Client/HedgedConnections.h>
+#include <Client/scaleInteractiveDelayByFanout.h>
 #include <Common/ProfileEvents.h>
-#include <Common/thread_local_rng.h>
 #include <Core/Settings.h>
 #include <Core/ProtocolDefines.h>
 #include <Interpreters/ClientInfo.h>
@@ -219,18 +218,9 @@ void HedgedConnections::sendQuery(
         modified_settings[Setting::dialect] = Dialect::clickhouse;
         modified_settings[Setting::dialect].changed = false;
 
-        /// Scale interactive_delay by sqrt(fanout) with jitter, same as in MultiplexedConnections.
-        {
-            size_t total_fanout = distributed_fanout * offset_states.size();
-            if (total_fanout > 1)
-            {
-                UInt64 delay = modified_settings[Setting::interactive_delay];
-                double scale = std::sqrt(static_cast<double>(total_fanout));
-                double jitter = 1.0 + (thread_local_rng() % 1000) / 1000.0;
-                delay = static_cast<UInt64>(std::ceil(static_cast<double>(delay) * scale * jitter));
-                modified_settings[Setting::interactive_delay] = delay;
-            }
-        }
+        modified_settings[Setting::interactive_delay] = scaleInteractiveDelayByFanout(
+            modified_settings[Setting::interactive_delay],
+            distributed_fanout * offset_states.size());
 
         if (disable_two_level_aggregation)
         {
--- a/src/Client/MultiplexedConnections.cpp
+++ b/src/Client/MultiplexedConnections.cpp
@@ -1,7 +1,6 @@
 #include <Client/MultiplexedConnections.h>
 
-#include <cmath>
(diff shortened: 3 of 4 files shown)


groeneai:
## Pre-PR validation gate

| # | Question | Answer |
|---|----------|--------|
| a | Deterministic repro? | Yes — `gtest_scale_interactive_delay_by_fanout.SaturatesOnOverflow` reproduces deterministically with fixed inputs (`UINT64_MAX`, `UINT64_MAX - 1`, `UINT64_MAX / 2 + 1`) that force `delay * scale * jitter >= 2^64`. |
| b | Root cause explained? | Yes — `static_cast<UInt64>(double x)` is UB when `x` is outside the `UInt64` range. Here `x = delay * scale * jitter` where `delay` is user-controlled `interactive_delay` (unbounded `UInt64`), `scale = sqrt(total_fanout)`, and `jitter ∈ [1.0, 2.0)`. Any `delay >= 2^63` with fanout `>= 4` reliably produces a `double` above `UINT64_MAX`. |
| c | Fix matches root cause? | Yes — the fix saturates the cast to `std::numeric_limits<UInt64>::max()` when the `double` result is `>= 2^64`, which is exactly the condition under which the bare `static_cast` is UB. No settings are pinned, no tags are added, no assertions are widened. |
| d | Test intent preserved? | N/A (new code path). The helper is factored from the previously duplicated inline block; the two existing call sites now go through it, and the existing jitter/scaling semantics are unchanged for all values that fit in `UInt64`. |
| e | Both directions demonstrated? | Yes. With the saturation removed: `SaturatesOnOverflow` fails — `static_cast<UInt64>` returns `0x8000000000000000` (`INT_INDEFINITE`) instead of `UINT64_MAX`. With the saturation: 4/4 tests pass. Verified locally against the debug build; the UBSan build is not strictly required for the regression, the cast result is deterministic on x86-64 without it. |

Session: `cron:clickhouse-ci-task-worker:20260419-041500`.

groeneai:
cc @azat @alexey-milovidov — could you review this? It saturates the `delay * sqrt(fanout) * jitter` cast in `MultiplexedConnections::sendQuery` / `HedgedConnections::sendQuery` to `UINT64_MAX`, which fixes a UBSan trap at `MultiplexedConnections.cpp:176:41` (STID `3646-3e82`, `Stress test (experimental, serverfuzz, arm_asan_ubsan)`) introduced by #100145 when `interactive_delay` is user-set to a large value.

src/Client/scaleInteractiveDelayByFanout.h
```diff
@@ -0,0 +1,42 @@
...
+        return delay;
+
+    const double scale = std::sqrt(static_cast<double>(total_fanout));
+    /// Random jitter in range [1.0, 2.0).
+    const double jitter = 1.0 + (thread_local_rng() % 1000) / 1000.0;
+    const double scaled = std::ceil(static_cast<double>(delay) * scale * jitter);
+
+    /// `static_cast<UInt64>(x)` is undefined behavior when `x` is outside the `UInt64` range.
+    /// `2^64` is exactly representable as `double`, so comparing against it is reliable.
+    constexpr double two_pow_64 = 18446744073709551616.0;
```
nikitamikhaylov, line 36: Why can'y you use `std::clamp` and `static_cast<Float64>(std::numeric_limits<UInt64>::max())` ?

nikitamikhaylov (approved):

src/Client/scaleInteractiveDelayByFanout.h
nikitamikhaylov: @groeneai Why?

groeneai: Good catch to question it — `static_cast<double>(std::numeric_limits<UInt64>::max())` is a trap due to IEEE 754 rounding.

`UINT64_MAX = 2^64 − 1` needs 64 mantissa bits, but `double` only has 53, so the cast rounds **up** to `2^64` (which is exactly representable), not down to `2^64 − 1`. That means `std::clamp(scaled, 0.0, static_cast<double>(UINT64_MAX))` can return exactly `2^64`, and the subsequent `static_cast<UInt64>(…)` is the same UB this PR is fixing:

```cpp
double scaled    = 1.86135e19;               // CI-observed value, > 2^64
double d_max     = static_cast<double>(std::numeric_limits<uint64_t>::max());
std::printf("%d\n", d_max == 18446744073709551616.0);  // 1  (i.e. == 2^64)

double clamped   = std::clamp(scaled, 0.0, d_max);     // 2^64
uint64_t result  = static_cast<uint64_t>(clamped);     // UB
// clang++ -fsanitize=undefined,float-cast-overflow:
//   runtime error: 1.84467e+19 is outside the range of representable values of type 'long unsigned int'
```

So the explicit `>= 2^64` comparison in the helper isn't stylistic — it exists specifically to dodge the UB cast at the `2^64` boundary. The literal `18446744073709551616.0` is `2^64` exactly, which **is** representable as `double` (one bit of mantissa is enough), so the comparison is well-defined.

If you'd like a `std::clamp`-shaped version, the safe upper bound is `std::nextafter(2^64, 0.0) = 18446744073709549568.0` — the largest `double` strictly less than `2^64`:

```cpp
/// Largest double strictly less than 2^64 — safe as an upper bound for cast to UInt64.
constexpr double max_safe = 18446744073709549568.0;
return static_cast<UInt64>(std::clamp(scaled, 0.0, max_safe));
```

The only functional difference is that this saturates to `2^64 − 2048` (`18446744073709549568`) instead of `UINT64_MAX` (`18446744073709551615`) — a 2047-tick gap at the very top of the range, which is immaterial for saturating a pathological user-set `interactive_delay`. Happy to switch to this form if you prefer — let me know.

PR #103092, opened 2026-04-19
Fix flaky 03237_avro_union_in_complex_types under randomized map_serialization_version

groeneai:
## Summary

Chronic flaky test `03237_avro_union_in_complex_types` — 10 unrelated PRs hit it over the last 30 days across `amd_debug`, `amd_asan_ubsan`, `amd_tsan`, `amd_msan`, `arm_binary` and `amd_llvm_coverage` variants. Zero master failures but very visible to developers.

The test stores rows from an Avro file into a `MergeTree` table and compares `SELECT *` output to a reference. When the CI runner randomizes `map_serialization_version_for_zero_level_parts` to `with_buckets` (see `tests/clickhouse-test` line 1574), `SerializationMap` splits map entries by key hash across buckets during `INSERT` and reassembles them bucket-by-bucket during `SELECT` (`collectMapFromBuckets` in `src/DataTypes/Serializations/SerializationMap.cpp`). As a result, keys inside a row's `Map` come out in bucket-hash order rather than insertion order.

That is valid `Map` semantics — entry order is not guaranteed after a bucketed round-trip — but this reference-based test was written assuming insertion order is preserved. Pinning `map_serialization_version = 'basic'` and `map_serialization_version_for_zero_level_parts = 'basic'` at the table level makes the output deterministic regardless of which `MergeTree` randomization the runner selects. The `with_buckets` path has dedicated tests elsewhere.

Typical flaky diff:

```
-zulu ... {'key9':9223372036854775807,'key10':NULL}
+zulu ... {'key10':NULL,'key9':9223372036854775807}
```

Not required — `CI Fix or Improvement`.

- [x] No user-facing change.

### Version info
- Merged into: `26.4.1.1085`

### DIFF
--- a/tests/queries/0_stateless/03237_avro_union_in_complex_types.sh
+++ b/tests/queries/0_stateless/03237_avro_union_in_complex_types.sh
@@ -120,6 +120,12 @@ $CH_CLIENT -q "select * from file('$file_name', 'Avro', '
 echo
 
 echo "== CREATE TABLE avro_union_test_03237 =="
+# Pin `map_serialization_version` to `basic` so the test output is deterministic.
+# With the `with_buckets` format (which the CI runner randomizes) keys inside a
+# Map column can be reordered across the bucketing round-trip — see the comment
+# in `src/DataTypes/Serializations/SerializationMap.cpp::collectMapFromBuckets`.
+# That is valid Map behaviour but makes this reference-based test flaky; the
+# bucketing path has dedicated tests elsewhere.
 $CH_CLIENT -q "CREATE TABLE avro_union_test_03237 (
   string_only String,
   string_or_null Nullable(String),
@@ -133,7 +139,8 @@ $CH_CLIENT -q "CREATE TABLE avro_union_test_03237 (
   double_or_null_or_string_or_long Variant(Float64, String, Int64),
   double_or_long_or_string_in_array Array(Variant(Float64, String, Int64)),
   double_or_string_or_long_or_null_in_map Map(String, Variant(Float64, Int64, String))
-) ENGINE = MergeTree ORDER BY tuple()"
+) ENGINE = MergeTree ORDER BY tuple()
+SETTINGS map_serialization_version = 'basic', map_serialization_version_for_zero_level_parts = 'basic'"
 echo
 
 echo "== SELECT * FORMAT Avro | INSERT INTO avro_union_test_03237 FORMAT Avro =="


groeneai:
## Pre-PR validation gate

**Session id:** `cron:clickhouse-ci-task-worker:20260419-111500`

### (a) Deterministic repro — YES
The failure is triggered deterministically by the `with_buckets` map serialization plus small `max_buckets_in_map`. Minimal repro:

```
clickhouse-client \
  --allow_merge_tree_settings \
  --map_serialization_version_for_zero_level_parts=with_buckets \
  --max_buckets_in_map=4 \
  --map_buckets_strategy=constant \
  --map_buckets_min_avg_size=1 \
  -q "CREATE TABLE t (id String, m Map(String, Int64)) ENGINE = MergeTree ORDER BY tuple()"
clickhouse-client -q "
  INSERT INTO t VALUES ('alpha', map('key1', 1, 'key2', 2));
  SELECT m FROM t"
```

Output: `{'key2':2,'key1':1}` instead of the inserted `{'key1':1,'key2':2}` — keys reordered 100% of the time under this configuration.

### (b) Root cause explained — YES
`SerializationMap::splitMapToBuckets` (line 1030) splits map entries by key hash into `N` buckets at serialization time. `collectMapFromBuckets` (line 1065) reassembles them by iterating bucket-by-bucket and concatenating each bucket's contents. So keys from the same row come out in bucket order rather than insertion order whenever hashes distribute keys across different buckets. Behaviour is documented in the code comment: "For each row, key-value pairs from all buckets are concatenated back into one map entry."

### (c) Fix matches root cause — YES
`CREATE TABLE ... SETTINGS map_serialization_version = 'basic', map_serialization_version_for_zero_level_parts = 'basic'` overrides the CLI-level randomizer and forces the single-stream (no buckets) path for this specific table. Verified that CREATE TABLE inline SETTINGS take precedence over `--allow_merge_tree_settings` CLI options.

### (d) Test intent preserved — YES
The test's purpose is to verify Avro union-in-complex-types encoding (`DESCRIBE`, `SELECT *`, custom schemas, error handling, round-trip). The `with_buckets` map storage format is not what this test is about — it has dedicated tests elsewhere. No assertions weakened, no `no-random-*` tags added, no data sizes reduced.

### (e) Both directions demonstrated — YES
With the same aggressive bucket settings injected via CLI:
- **Without fix:** 100% fail (rows 57 and 68 show `{'key2':NULL,'key1':3.14...}` instead of the expected `{'key1':3.14...,'key2':NULL}`).
- **With fix:** 20/20 pass.
Also 20/20 pass without any CLI randomization (no regression on the default path).

### (f) Fix is general — YES
Test-only change (no C++ modifications). The fix targets the exact setting-interaction that produces the flaky diff. No other tests use this test table, and other tests that depend on Map order already pin or avoid bucketing.

groeneai:
cc @Avogar — could you review this? It's a test-only fix for a chronic flaky test (10 unrelated PRs/30d) caused by the randomized `map_serialization_version_for_zero_level_parts=with_buckets` reordering map keys within rows. Fix pins the table to `basic` so the reference-based test is deterministic.

Avogar (approved):

PR #103095, opened 2026-04-19
Make waitDetachedTableNotInUse respect query cancellation

alexey-milovidov:
The test config sets `database_atomic_wait_for_drop_and_detach_synchronously=1` in the default profile. During stress test hung check preparation, `DETACH DATABASE` is called for each database. With this setting, the DETACH becomes synchronous and calls `waitDetachedTableNotInUse`, which is an infinite busy-wait loop with 100ms sleep. If any table in the database has a `shared_ptr` reference held by a background operation (e.g., a dictionary kept alive by the `ExternalLoader`), the loop runs forever because there is no cancellation check. `KILL QUERY` also cannot interrupt it.

The fix adds a `throwIfKilled` check on each iteration of the busy-wait loop in `waitDetachedTableNotInUse`, so that query cancellation (via `KILL QUERY` or the `CancellationChecker` timeout) can interrupt it.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=f4a62d7f35da0e026bc4af0649ba100ec9c2cb85&name_0=MasterCI&name_1=Stress%20test%20%28azure%2C%20amd_msan%29

Make `DETACH DATABASE` with `database_atomic_wait_for_drop_and_detach_synchronously` respect `KILL QUERY` instead of hanging indefinitely when a table reference is held.

### Version info
- Merged into: `26.5.1.182`


antonio2368:
https://github.com/ClickHouse/ClickHouse/pull/102804
What about this?

alexey-milovidov:
> https://github.com/ClickHouse/ClickHouse/pull/102804
What about this?

Yours looks better, but not ready yet (failed tests), while mine is green. I propose the following steps: - merge my PR; - when it is ready, replace it with yours PR.

Issue #102460, opened 2026-04-11
26.3+ jemalloc fantom crashes due to LTO


azat:
https://github.com/ClickHouse/ClickHouse/pull/99342 leads to fantom crashes, and it does not happens frequently, what is clear so far:
- it is jemalloc related
  - ASan/UBSan/TSan/MSan does not find anything
- looks like `tcache` related (disabling it helps)
- enabling `junk` also helps (likely to the relation with `tcache`)

Revert is here - https://github.com/ClickHouse/ClickHouse/pull/102294

This issue is created to track progress for the long-term fix.

Refs:
- https://github.com/ClickHouse/ClickHouse/issues/101915
- https://github.com/ClickHouse/ClickHouse/issues/101848

azat:
What I've tried so far:

- `rr` - too slow (requires https://github.com/rr-debugger/rr/pull/4066/)
- `valgrind` (hah) - same crashes as before
- check `ivsalloc` or/and `JEMALLOC_DEBUG` and `je_malloc_usable_size` - same crashes
- `-fno-strict-aliasing` for jemalloc - still crashes
- `-fno-delete-null-pointer-checks` for jemalloc - does not crashes (but likely due to less inlining)
- `-DJEMALLOC_OPT_SAFETY_CHECKS=1` + - still crashes
- `-DJEMALLOC_OPT_SIZE_CHECKS=1` + `-DJEMALLOC_OPT_SIZE_CHECKS=1` - does not crashes [?]
- `-DJEMALLOC_UAF_DETECTION` + `MALLOC_CONF=abort:true,confirm_conf:true,abort_conf:true,retain:false,junk:false,zero:false,lg_san_uaf_align:0` - still crashes in random places
- `FiberStack::default_stack_size=640 * 1024` - still crashes
- UBsan build with LTO - does not crashes
- upgrading jemalloc to latest head
- likely something else

_Now I'm using https://github.com/ClickHouse/ClickHouse/issues/80948#issuecomment-4230920126, since it reduces the testing cycle_

UPD: new repro crashes it on aarch64 and aws as well - https://pastila.nl/?00012dbc/96976ced67fd2bd00e23bf815975c4fe#rUyNlLNq/90fuda3wBlqFQ==GCM

nicezic:
If this is fixed, can I upgrade to 26.3 LTS from 25.8??? 
cheer up!!!

azat:
[...]

PR #103096, opened 2026-04-19
tests: add a test that triggers crashes in jemalloc with LTO

azat:

Thanks to @korowa for the test https://github.com/ClickHouse/ClickHouse/issues/80948#issuecomment-4230920126

Refs: https://github.com/ClickHouse/ClickHouse/issues/102460

### DIFF
--- a/tests/queries/0_stateless/04105_jemalloc_lto_crashes.sql
+++ b/tests/queries/0_stateless/04105_jemalloc_lto_crashes.sql
@@ -1,0 +1,34 @@
+-- Tags: no-parallel, no-sanitizers, no-debug, no-object-storage, no-fasttest, no-replicated-database
+-- - no-parallel - too heavy (~10 GiB of RAM)
+
+-- Test for crashes in jemalloc with LTO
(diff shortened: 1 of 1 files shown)


tests/queries/0_stateless/04105_jemalloc_lto_crashes.sql
```diff
@@ -0,0 +1,34 @@
+-- Tags: no-parallel, no-sanitizers, no-debug, no-object-storage, no-fasttest, no-replicated-database
```
tiandiwonder, line 1: check if long tag is needed.

tiandiwonder (approved):

tests/queries/0_stateless/04105_jemalloc_lto_crashes.sql
azat: Thanks, this query executes within 20 sec in release, and disable for anything except for release, so should be ok

azat: Though there is another problem, stateless does not use LTO binary

Issue #103100, opened 2026-04-19
Segfault when unloading WASM modules


JosephRedfern:
### Describe what's wrong

When attempting to unload all WASM modules with:

```sql
DELETE FROM system.webassembly_modules WHERE 1=1;
```

ClickHouse server reliably segfaults.

### Does it reproduce on the most recent release?

Yes

### How to reproduce

1. Run ClickHouse 26.4.1.1009 with `allow_experimental_webassembly_udf` enabled
2. Execute `DELETE FROM system.webassembly_modules WHERE 1=1;`

### Expected behavior

All web assembly modules are unloaded, or the error is handled gracefully (no segfault).

### Error message and/or stacktrace

```
Joes-MacBook-Air.local :) DELETE FROM system.webassembly_modules WHERE 1=1;

DELETE FROM system.webassembly_modules WHERE 1 = 1

Query id: af7653cf-c508-4a3b-bdaf-47c26370d441

[Joes-MacBook-Air.local] 2026.04.19 15:26:10.027108 [ 715411 ] <Fatal> BaseDaemon: ########################################
[Joes-MacBook-Air.local] 2026.04.19 15:26:10.027203 [ 715411 ] <Fatal> BaseDaemon: (version 26.4.1.1009 (official build), build id: D33227CE542E351D8DFDFA5C05A2D0EB, git hash: 2642aa58ce19e087bc092313788b9538f23bd561) (from thread 715412) (query_id: af7653cf-c508-4a3b-bdaf-47c26370d441) (query: DELETE FROM system.webassembly_modules WHERE 1=1;) Received signal Segmentation fault: 11 (11)
[Joes-MacBook-Air.local] 2026.04.19 15:26:10.027232 [ 715411 ] <Fatal> BaseDaemon: Address: 0x57. Access: <not available>. Attempted access has violated the permissions assigned to the memory area.
[Joes-MacBook-Air.local] 2026.04.19 15:26:10.027261 [ 715411 ] <Fatal> BaseDaemon: Stack trace: 0x000000010257535c 0x0000000109e45938 0x0000000109e456a8 0x000000010c92f2cc 0x000000010cd1e95c 0x000000010cd1a104 0x000000010f87d808 0x000000010f897fa4 0x0000000113697e18 0x0000000113698424 0x00000001136607c4 0x000000011365e6a8 0x0000000187487c58 0x0000000187482c1c
[...]

PR #103101, opened 2026-04-19
Fix segfault during WASM module delete (issue #103100)

JosephRedfern:
This PR fixes a segfault when attempting to unload WASM modules using e.g. `WHERE 1=1` predicate, as reported here: https://github.com/ClickHouse/ClickHouse/issues/103100.

The implementation previously cast LHS of the predicate to `ASTIdentifier` and accessed `full_name` - when `WHERE 1=1` is used, LHS is an `ASTLiteral` and full_name access causes segfault. 

Fix segfault when deleting WASM module using non-identifier predicate (e.g. `WHERE 1=1`) 

### Version info
- Merged into: `26.5.1.458`


tiandiwonder (approved):
Update change log: Fix server crash in `DELETE FROM system.webassembly_modules WHERE 1=1`

JosephRedfern:
@tiandiwonder there was an error in my test, which I've now fixed - I suspect these other CI failures are due to flakiness - WDYT?

PR #103113, opened 2026-04-19
Support geotypes for iceberg

scanhex12:
- Experimental Feature

Support geotypes for iceberg


src/Storages/ObjectStorage/DataLakes/Iceberg/SchemaProcessor.h
```diff
@@ -93,7 +95,7 @@ class IcebergSchemaProcessor
     Poco::JSON::Object::Ptr getIcebergTableSchemaById(Int32 id) const;
     bool hasClickhouseTableSchemaById(Int32 id) const;
 
-    static DataTypePtr getSimpleType(const String & type_name);
+    static DataTypePtr getSimpleType(const String & type_name, bool allow_geo_parser = true);
```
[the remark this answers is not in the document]
SmitaRKulkarni: makes sense, maybe set default to false. @scanhex12  wdyt?

src/Storages/ObjectStorage/DataLakes/Iceberg/SchemaProcessor.cpp
```diff
@@ -247,6 +275,9 @@ DataTypePtr IcebergSchemaProcessor::getSimpleType(const String & type_name)
         return std::make_shared<DataTypeDateTime64>(9, "UTC");
     if (type_name == f_string || type_name == f_binary)
         return std::make_shared<DataTypeString>();
+
+    if (type_name.starts_with(f_geometry) || type_name.starts_with(f_geography))
+        return allow_geo_parser ? DataTypeFactory::instance().get("Geometry") : std::make_shared<DataTypeString>();
```
SmitaRKulkarni: This means type does have geometry / geography but flag is not enabled so it will not be processed as intended right? , then maybe we should throw error ?

src/Storages/ObjectStorage/DataLakes/Iceberg/SchemaProcessor.cpp
```diff
@@ -112,16 +112,44 @@ bool equals(const T & first, const T & second)
     return first_string_stream.str() == second_string_stream.str();
 }
 
-bool operator==(const Poco::JSON::Array & first, const Poco::JSON::Array & second)
```
SmitaRKulkarni, line 115: Maybe we can still support it by doing the same isArray check followed by schemaFieldsAreStructurallyIdentical. wdyt ?

SmitaRKulkarni (commented):
CI fails looks relevant, rest all LGTM

SmitaRKulkarni (approved):

PR #103117, opened 2026-04-20
Narrow main-table mark ranges using projection index bitmap

### LINKED ISSUE #99114: Narrow marks before reading with Projection Index
### Company or project name

_No response_

### Describe the situation

Currently, the behavior of the Projection Index is approximately:
1. Analyze projection parts to see if the number projection or table rows are estimated to be larger
2. If the projection reads less data, read the projection and construct a (roaring) bitmap of part offsets
3. Read the main table with a PREWHERE filter backed by the bitmap
  a. In this case, all granules of all parts of the table are distributed to the pool of readers. Then, ClickHouse checks for every granule whether it should skip the granule based on the bitmap.

This is very inefficient for large tables with narrow projection index queries. For example, a large Distributed Tracing dataset with a narrow `trace_id = 'abc' query.

Instead, ClickHouse should eagerly use the bitmap to narrow the mark ranges to read, and then only distribute those granules to readers. The bitmap can be assumed to be small because of the check in step (1) -- it is naturally capped at `max_projection_rows_to_use_projection_index` (default 1M).

All

### How to reproduce

I have a (unfortunately private) dataset with ~2.5T rows, across 14 partitions and ~350 parts, each with several billion rows:
```
CREATE TABLE prod.spans
(
    `environment` LowCardinality(String),
    `system` LowCardinality(String),
    `service` LowCardinality(String),
    `start_time` DateTime64(9),
    ...
    `trace_id` FixedString(16),
    PROJECTION by_trace_id
    (
        SELECT _part_offset
        ORDER BY trace_id
    )
)
ENGINE = MergeTree
PARTITION BY toYYYYMMDD(start_time)
ORDER BY (environment, system, ..., start_time)
TTL start_time + toIntervalDay(14)
SETTINGS ttl_only_drop_parts = 1, index_granularity = 8192
```

The following projection-indexed query is usually slow:
```
SELECT *
FROM prod.spans
WHERE trace_id = unhex('0ddd0ce11223e5955d8f22037b983ce0')
FORMAT `null`
SETTINGS
  use_query_condition_cache = 0,
  max_threads = 1,
  min_table_rows_to_use_projection_index = 0

0 rows in set. Elapsed: 4.154 sec. Processed 65.54 thousand rows, 27.30 MB (15.78 thousand rows/s., 6.57 MB/s.)
0 rows in set. Elapsed: 4.081 sec. Processed 65.54 thousand rows, 27.30 MB (16.06 thousand rows/s., 6.69 MB/s.)
0 rows in set. Elapsed: 3.960 sec. Processed 65.54 thousand rows, 27.30 MB (16.55 thousand rows/s., 6.90 MB/s.)
```

Profile events indicate that time is spent in CPU and reading rows for PREWHERE, but not much data I/O:
```
CompressedReadBufferBytes                              131.30 MB           29.18 MB/s          Number of uncompressed bytes…
IOBufferAllocBytes                                     161.11 MB           35.09 MB/s          Number of bytes allocated fo…
OSReadChars                                            62.47 MB            13.63 MB/s          Number of bytes read from fi…
ReadBufferFromFileDescriptorReadBytes                  62.35 MB            13.64 MB/s          Number of bytes read from fi…
ReadCompressedBytes                                    55.88 MB            12.11 MB/s          Number of bytes (the number …
ThreadPoolReaderPageCacheHitBytes                      62.35 MB            13.64 MB/s          Number of bytes read inside …
OSCPUVirtualTimeMicroseconds                           3.99 s              892.10 ms/s         CPU time spent seen by OS. D…
RealTimeMicroseconds                                   4.00 s              894.39 ms/s         Total (wall clock) time spen…
RowsReadByPrewhereReaders                              4.28 million        930.00 thousand/s   Number of rows read from Mer…
UserTimeMicroseconds                                   3.97 s              889.76 ms/s         Total time spent in processi…
```

### Expected performance

I am hoping for the following execution:
1. Same as before, decide whether to use the projection index based on the estimated rows to read.
2. If using the projection, **eagerly compute** the bitmap of row offsets that match the condition
3. Then, use the bitmap to filter which mark ranges to read, potentially pruning many parts, and narrowing which granules to read from surviving parts

### Additional context

I vibe-coded an implementation of eager bitmap-based part and range narrowing, and it is much faster on my dataset:
```
0 rows in set. Elapsed: 0.179 sec. Processed 65.54 thousand rows, 27.59 MB (366.70 thousand rows/s., 154.35 MB/s.)
0 rows in set. Elapsed: 0.177 sec. Processed 65.54 thousand rows, 27.59 MB (369.53 thousand rows/s., 155.54 MB/s.)
0 rows in set. Elapsed: 0.177 sec. Processed 65.54 thousand rows, 27.59 MB (369.41 thousand rows/s., 155.50 MB/s.)
```
In fact, ~all of this time is accounted for by the index analysis: it is spending ~no time reading data from very few matching granules:
```
EXPLAIN indexes=1 ...
...
Skip
  Name: projection_index
  Description: Projection index pre-read
  Parts: 522/522
  Granules: 8/105600197

22 rows in set. Elapsed: 0.149 sec. 
```

EmeraldShift:
- Performance Improvement

Improved the performance of filtering large tables with projection indexes.

### What this PR does

The deferred read path distributes all main-table mark ranges to reader threads and uses `canSkipMark()` per granule to drop non-matching ones. For narrow filters on large tables (#99114) that's O(marks) of mark-index reads, buffer allocations, and bitmap lookups per query, even when the answer is a handful of granules.

This PR narrows main-table mark ranges using the projection index bitmap, gated behind `projection_index_narrow_marks` (default false). When on, the read pool looks up the already-cached projection bitmap from `MergeTreeIndexReadResultPool` inside `getTask()`, converts it to `MarkRanges`, and intersects with the batch before `createTask()` runs. Empty batches are dropped and the pool fetches the next one. `MergeTreeReadPoolInOrder` loops until a non-empty batch or the part drains -- `nullptr` still means "no more work for this part" per the existing contract.

The bitmap -> `MarkRanges` conversion is cached once per part on `MergeTreeIndexReadResult` via `std::call_once`, so concurrent tasks for the same part share it.

Tests: `03775_projection_index_narrow_marks` and `03776_projection_index_point_read` assert identical results with preread on and off, including a missing-key case across multiple parts.

Performance (release, 10M rows / 20 parts / `index_granularity=8`, `max_threads=1`): steady-state per-query wall time drops from 63 ms to 8.5 ms. Parent-reader ProfileEvents (`FileOpen`, `ReadCompressedBytes`, `CompressedReadBufferBytes`, `MarkCacheMisses`) drop about 6x. On the production workload in #99114 the original report saw 4.15 s -> 0.18 s.

Closes #99114

cc @amosbird


EmeraldShift:
Hi @nihalzp can you please enable testing for this PR and/or review (+ @amosbird)?

EmeraldShift:
AST fuzzer - STID: 3262-4259, seems unrelated

@nihalzp Does that ^^ sound right? Would you be able to lend a review?

nihalzp:
The improvement is in critical components and I am not a big expert on these topics. It will be better to be reviewed by someone with experience.

EmeraldShift:
Hi @nihalzp @amosbird would either of you be able to recommend a reviewer for this feature? It's quite important for our observability use case to make projection-indexed lookups not sit for 5-20s looping in `canSkipMark`.

PR #95467, opened 2026-01-28
Fix s3queue deduplication

kssenii:
### Changelog category (leave one):
- Bug Fix

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix deduplication for streaming from S3(Azure)Queue engine.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.2.1.837`

PR #101646, opened 2026-04-02
Revert "Fix DROP TABLE hang on Kafka tables after consumer heartbeat error"

azat:
Reverts ClickHouse/ClickHouse#100388

### Version info
- Merged into: `26.4.1.689`

PR #103126, opened 2026-04-20
StorageObjectStorageQueue: fix shutdown blocking on stuck pipeline

tuanpach:
## Motivation

When `StorageObjectStorageQueue::shutdown` is called while a streaming task
is mid-file, the source's `generateImpl` loop branched on
`table_is_being_dropped`:

- **Drop:** throw `QUERY_WAS_CANCELLED` immediately, stopping the read.
- **Plain shutdown:** log `"Will process the file fully and then shutdown"`
  and keep reading.

In the plain-shutdown path, `task->deactivate` blocks the server's shutdown
until the rest of the file has been read from object storage. For large
files or slow storage, that can be many minutes per task.

## Change

`generateImpl` now throws `QUERY_WAS_CANCELLED` in both cases. The
`isCancelled()` branch a few lines above already does the same thing;
this unifies the two `shutdown_called` sub-cases.

## Why the old behavior existed, and why throwing is safe now

Rows already returned by earlier `generateImpl` calls may have been
inserted into the destination by the pipeline before we throw. The file
is marked `Cancelled` (not `Processed`), so it is retried on next start.
Without deduplication, those rows would arrive twice — and that is
precisely why the "process file fully on shutdown" behavior was added in
the first place.

Proper deduplication for `StorageObjectStorageQueue` was added in
#95467, available from 26.2 onwards. On 26.2+ the dedup machinery
catches the partial-write rows on retry, so the throw is safe.

> ⚠️ **Do not backport to versions older than 26.2.** Deduplication is
> unavailable there, and throwing would produce real duplicates — a
> regression compared to the prior "process file fully" behavior.

Discussion:
- Suggested by @azat: https://github.com/ClickHouse/ClickHouse/pull/103126#issuecomment-4389441321
- Backport caveat from @kssenii: https://github.com/ClickHouse/ClickHouse/pull/103126#issuecomment-4407271519

Fix `StorageObjectStorageQueue` (`S3Queue`, `AzureQueue`) blocking shutdown until partially-processed files were fully read from object storage. The source now aborts the read immediately on shutdown; deduplication ensures rows already streamed into the destination table before shutdown are not duplicated when the file is retried on next start.

### Version info
- Backported to: `26.4.4.4`


kssenii (approved):

src/Storages/ObjectStorageQueue/StorageObjectStorageQueue.cpp
```diff
@@ -973,6 +988,13 @@ bool StorageObjectStorageQueue::streamToViews(size_t streaming_tasks_index)
             throw;
         }
 
+        if (shutdown_called)
+            LOG_DEBUG(log, "Pipeline execution cancelled by shutdown after {} ms (query_id: {})",
```
azat: This is actually not true, since there is a race condition, the `shutdown_called` can be set just before this line, but the pipeline will finish

Also, do we need to commit with `insert_succeeded` in this case? I doubt so

azat: There was a similar attempt fix it in kafka, but we reverted it eventually - https://github.com/ClickHouse/ClickHouse/pull/101646

tuanpach: I updated the log, now it only states the pipeline completed.

tuanpach: > There was a similar attempt fix it in kafka, but we reverted it eventually - #101646

I saw that the reason for reverting was an integration test failure. Can we merge and observe this fix first?

azat: I think we cannot commit with `insert_success=1` if the pipeline was cancelled, or maybe I'm missing something?

tuanpach: @kssenii, do you have any thoughts on this?

I think this is a separate issue.

kssenii: > I think we cannot commit with insert_success=1 if the pipeline was cancelled

In case of cancellation while reads are still in progress - will throw an exception https://github.com/ClickHouse/ClickHouse/blob/9081630de1a12531b8850d5e172921ece7197b43/src/Storages/ObjectStorageQueue/ObjectStorageQueueSource.cpp#L1048-L1050 and nothing will be committed, because we will set `insert_succeeded = false` here https://github.com/ClickHouse/ClickHouse/blob/9081630de1a12531b8850d5e172921ece7197b43/src/Storages/ObjectStorageQueue/StorageObjectStorageQueue.cpp#L943. However, I am not sure what happens if it gets cancelled after read part is fully finished but insert is still not -- will exception be thrown as well?

tuanpach: @azat @kssenii can we merge this PR to solve the issue with shutting down first?
Regarding the insert issue, we can handle it in a separate PR.

tuanpach:
I will merge this PR first.

src/Storages/ObjectStorageQueue/StorageObjectStorageQueue.cpp
azat: I don't think it should be merged in the current version, like @kssenii wrote, reading is safe, but AFAICS writing the data is not, so we can cancel writing but the files will be marked as committed

azat:
>StorageObjectStorageQueue: fix shutdown blocking on stuck pipeline

Also this changelog is not very user friendly

src/Storages/ObjectStorageQueue/StorageObjectStorageQueue.cpp
azat: I think that the correct approach will be to throw QUERY_WAS_CANCELLED when the query was cancelled from the callback (but need to make it compatible with the current users in the code)

tuanpach: Can we commit with `insert_succeeded=false` when `shutdown_called` is `true`?

azat: This will be racy, imagine situation pipeline finished before storage was shutdown, and right after it was, so it will skip commit, but it should commit

tuanpach: In the last commit, I used the `cancel_was_triggered` flag, which is only set to `true` in the cancel lambda function. So it only skips the commit if the pipeline was actually canceled.

azat: This is also racy, due to how CompletedPipelineExecutor handles this - https://github.com/ClickHouse/ClickHouse/blob/340fac0675e77aaa6669b448eb7b7a7e127a259c/src/Processors/Executors/CompletedPipelineExecutor.cpp#L87-L96

azat:
AFAICS right now it is intended behavior of S3Queue - https://github.com/ClickHouse/ClickHouse/blob/38eaf3ff784926b610e6506d1aeed9cf3fdb5a91/src/Storages/ObjectStorageQueue/ObjectStorageQueueSource.cpp#L1082-L1084

I guess we can just throw here? (though we still will write data to the underlying table...)

@kssenii WDYT?

src/Storages/ObjectStorageQueue/StorageObjectStorageQueue.cpp
tuanpach: From my understanding, what you're concerned about is that, right before `is_cancelled_callback()`, the data is fully processed. However, `is_cancelled_callback()` invokes ->  `cancel_was_triggered=true` -> data is not committed -> leading to duplication if the data is re-processed.

If it is true, throwing an exception also has the same issue. Please correct me if I was wrong.

kssenii:
> AFAICS right now it is intended behavior of S3Queue 
I guess we can just throw here? (though we still will write data to the underlying table...)

Yes - it was intended behaviour that we finish processing already started files if we see shutdown flag is set. This was necessary before we implemented proper deduplication for s3queue https://github.com/ClickHouse/ClickHouse/pull/95467, as otherwise there will be duplicates. But this is only for 26.2+. So if we throw there for earlier version - we get duplicates. So if we plan to backport it to <26.2 versions, it would be a degradation compared to previous behaviour in my opinion.

tuanpach:
> > AFAICS right now it is intended behavior of S3Queue
> > I guess we can just throw here? (though we still will write data to the underlying table...)
> 
> Yes - it was intended behaviour that we finish processing already started files if we see shutdown flag is set. This was necessary before we implemented proper deduplication for s3queue #95467, as otherwise there will be duplicates. But this is only for 26.2+. So if we throw there for earlier version - we get duplicates. So if we plan to backport it to <26.2 versions, it would be a degradation compared to previous behaviour in my opinion.

I will throw an exception there, and only backport to 26.2+.

src/Storages/ObjectStorageQueue/ObjectStorageQueueSource.cpp
```diff
@@ -1067,21 +1067,20 @@ Chunk ObjectStorageQueueSource::generateImpl()
...
+            /// Something must have been already read.
+            chassert(started_file->getFileStatus()->processed_rows > 0);
+            /// Mark file as Cancelled, such files will not be set as Failed.
+            processed_files.back().state = FileState::Cancelled;
+            /// Throw to abort reading more rows from this file rather than blocking shutdown
+            /// for the rest of the file. Rows already returned by previous generateImpl() calls
+            /// may have been inserted into the destination table; the file is marked Cancelled
+            /// (not Processed) and will be retried on next start, which can produce duplicates
+            /// for those rows.
+            throw Exception(
```
kssenii: I think there can be a case here that we actually read the file to the end but do not yet know this (it will be known once we do reader->pull and it returns empty block). But I do not see an easy way to check this apart from calling reader->pull again, so I think we can neglect this case, since in the worst case this file will be reread and deduplicated fully, which is not a problem apart from extra file read.

kssenii (approved):

alexey-milovidov:

Issue #100910, opened 2026-03-27
ClickHouse 26.3 fails to start when using `<rotation>daily</rotation>` in logger config


oideph:
### Describe what's wrong

After upgrading from v25.8 to v26.3, clickhouse-server fails to start with error `Poco::Exception. Code: 1000, e.code() = 0, Invalid argument: size must be greater than zero` when `<size>daily</size>` or `<rotation>daily</rotation>`.

Likely related to #87620.

### Does it reproduce on the most recent release?

Yes

### How to reproduce

Set `<size>daily</size>` in `<logger>` section:
```xml
<logger>
    <level>debug</level>
    <log>/var/log/clickhouse-server/clickhouse-server.log</log>
    <errorlog>/var/log/clickhouse-server/clickhouse-server.err.log</errorlog>
    <size>daily</size>
    <count>10</count>
</logger>
```
Or `<rotation>daily</rotation>` without `<size>`:
```xml
<logger>
    <level>debug</level>
    <log>/var/log/clickhouse-server/clickhouse-server.log</log>
    <errorlog>/var/log/clickhouse-server/clickhouse-server.err.log</errorlog>
    <rotation>daily</rotation>
    <count>10</count>
</logger>
```

### Expected behavior

When using time‑based rotation (e.g. daily), should ignore size‑based rotation parameters. The server should start normally with:
```xml
<rotation>daily</rotation>
```

### Error message and/or stacktrace

```
Mar 27 15:54:00 clickhouse-dev systemd[1]: Starting ClickHouse Server (analytic DBMS for big data)...
Mar 27 15:54:00 clickhouse-dev clickhouse-server[36928]: Processing configuration file '/etc/clickhouse-server/config.xml'.
Mar 27 15:54:00 clickhouse-dev clickhouse-server[36928]: Logging debug to /var/log/clickhouse-server/clickhouse-server.log
Mar 27 15:54:00 clickhouse-dev clickhouse-server[36928]: Poco::Exception. Code: 1000, e.code() = 0, Invalid argument: size must be greater than zero, Stack trace (when copying this message, always include the lines below):
Mar 27 15:54:00 clickhouse-dev clickhouse-server[36928]: 0. Poco::RotateBySizeStrategy::RotateBySizeStrategy(unsigned long) @ 0x0000000023b9928f
[...]

PR #103127, opened 2026-04-20
Fix bugs in parseSize() and parseTime() for combined rotate strategy

### LINKED ISSUE #103108: `parseSize` greedily consumes numeric prefix from HH:MM time rotation strings, creating a spurious byte-size threshold

_Retrospective finding from a historical scan of [PR #87620](https://github.com/ClickHouse/ClickHouse/pull/87620) (merged 2025-12-29). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

Configuring `<rotation>12:00</rotation>` (intended: rotate log at noon) causes log files to rotate every 12 bytes, creating dozens of tiny rotated log files from a single query

**Root cause:** CombinedRotateStrategy.h:78 — `else if (unit.empty()) size = n;` returns any bare number as a size in bytes. For '12:00', the digits '12' are consumed and ':' stops digit parsing, leaving unit empty. The old code (now removed) checked for ':' in the string BEFORE size parsing: `if (rotation.find(':') != std::string::npos)` to prioritize time-based rotation.

**Why we believe this is a bug:** `FileChannel::setRotation` (FileChannel.cpp:240) → `CombinedRotateStrategy::parseRotation` (CombinedRotateStrategy.h:170) calls `parseSize` first. `parseSize` reads '12' as n=12, finds no alpha unit after digits (colon is not alpha), hits `unit.empty()` branch at line 78 returning 12 bytes. `parseTime` then receives only ':00' instead of '12:00'.

**Affected locations:**
- `base/poco/Foundation/include/Poco/CombinedRotateStrategy.h:78` — parseSize returns bare number as byte size when unit is empty
- `base/poco/Foundation/include/Poco/CombinedRotateStrategy.h:170` — parseRotation calls parseSize before parseTime, letting it consume time prefixes

**Impact:** Any user configuring time-based log rotation with HH:MM format (e.g., '12:00', '18:30', '2:00') gets an absurdly small size-based rotation (12/18/2 bytes) plus a wrong time string. This causes potentially thousands of rotated log files per minute, filling disk with tiny compressed archives and degrading logging performance.

Yes — confirmed on current `master` (commit `8b12293040c5`).

### How to reproduce

```sql
Create a config file with `<rotation>12:00</rotation>` and run: clickhouse-local --config-file=config.xml -q 'SELECT 1'. Then count log files — should be 1, but bug produces 10+.
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/3c399646-8177-42b4-906f-58196319342b)

### Expected behavior

```
OK: time-based rotation correctly parsed
```

```
BUG: rotation=12:00 caused spurious file rotation
```

### Additional context

**Open risks:**
- parseTime line 153 has a separate substr bug: `str.substr(pos, comma_after_colon)` uses absolute position as length, corrupting time strings when there is content after them
- Loggers.cpp:183 — error log channel doesn't use logger.rotation, only logger.size

**Suggested fix:** After parsing digits and finding unit empty, check if the next non-space character is ':'. If so, return `std::nullopt` to let `parseTime` handle it. Alternatively, restructure `parseRotation` to check for ':' in the string first (like the old code did) and prioritize `parseTime` over `parseSize`.

Found during automated review of [PR #87620](https://github.com/ClickHouse/ClickHouse/pull/87620).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr87620_001`_

zhangjmruc:

Fixed parseSize and parseTime logic to correctly distinguish between size thresholds and time-based rotation schedules (e.g., 12:00). Fixes #103108 and #100910 

### DIFF
--- a/base/poco/Foundation/include/Poco/CombinedRotateStrategy.h
+++ b/base/poco/Foundation/include/Poco/CombinedRotateStrategy.h
@@ -68,6 +68,12 @@ private:
         std::string unit;
         while (it != end && Ascii::isAlpha(*it)) unit += *it++;
 
+        /// Return std::nullopt if n is 0
+        if (n == 0)
+            return std::nullopt;
+
+        while (it != end && Ascii::isSpace(*it)) ++it;
+
         UInt64 size = 0;
         if (unit == "K")
             size = n * 1024;
@@ -75,7 +81,7 @@ private:
             size = n * 1024 * 1024;
         else if (unit == "G")
             size = n * 1024 * 1024 * 1024;
-        else if (unit.empty())
+        else if (unit.empty() && (it == end || *it == ','))
             size = n;
         else
             return std::nullopt; /// Return std::nullopt if fails to parse
@@ -126,7 +132,6 @@ private:
             return std::nullopt; /// Return std::nullopt if fails to parse
 
         /// Try to skip ','
-        while (it != end && Ascii::isSpace(*it)) ++it;
         if (it != end && *it == ',') ++it;
 
         pos = it - str.begin();
@@ -150,7 +155,7 @@ private:
             return res;
         }
 
-        auto res = std::string(str.substr(pos, comma_after_colon));
+        auto res = std::string(str.substr(pos, comma_after_colon - pos));
         pos = comma_after_colon + 1;
         return res;
     }
--- a/base/poco/Foundation/include/Poco/CombinedRotateStrategy.h
+++ b/base/poco/Foundation/include/Poco/CombinedRotateStrategy.h
@@ -132,6 +132,7 @@ private:
             return std::nullopt; /// Return std::nullopt if fails to parse
 
         /// Try to skip ','
+        while (it != end && Ascii::isSpace(*it)) ++it;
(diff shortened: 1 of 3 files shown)


zhangjmruc:
@alexey-milovidov Hi, could you help to assign someone to review this PR? It fixes several bugs in the log rotation parsing logic (CombinedRotateStrategy):

1. parseSize misinterpreted time strings as sizes — e.g., "12:00" was parsed as size=12 bytes instead of being passed through to parseTime. Fixed by requiring the character after a bare number to be end-of-string or ','.
2. parseTime had an incorrect substr call — substr(pos, comma_after_colon) used an absolute position as a length, producing wrong results when pos > 0 (e.g., "100M,12:00,daily"). Fixed to substr(pos, comma_after_colon - pos).
3. Error log channel ignored logger.rotation config — the main log read logger.rotation with fallback to logger.size, but the error log only read logger.size. Now both are consistent.
4. Added unit tests for parsing valid/invalid rotation formats and a regression test verifying "12:00" is not mistaken for a size.

PR #103144, opened 2026-04-20
Add virtual $name sub-column for Avro union type discrimination

### LINKED ISSUE #102300: Avro format, message name indicator for union type
### Company or project name

_No response_

### Use case

Better UX for querying union types in avro.

```py
#!/usr/bin/env python3
"""
Minimal Avro file with a multi-type union to reproduce
ClickHouse filtering issue on union branches.
"""

import fastavro

schema_dict = {
    "type": "record",
    "name": "Event",
    "fields": [
        {"name": "id", "type": "int"},
        {
            "name": "payload",
            "type": [
                "null",
                {
                    "type": "record",
                    "name": "TypeA",
                    "fields": [
                        {"name": "a_value", "type": "string"}
                    ]
                },
                {
                    "type": "record",
                    "name": "TypeB",
                    "fields": [
                        {"name": "b_value", "type": "int"}
                    ]
                }
            ],
            "default": None
        }
    ]
}

parsed_schema = fastavro.parse_schema(schema_dict)

records = [
    {"id": 1, "payload": ("TypeA", {"a_value": "hello"})},
    {"id": 2, "payload": ("TypeB", {"b_value": 42})},
    {"id": 3, "payload": None},
    {"id": 4, "payload": ("TypeA", {"a_value": "world"})},
    {"id": 5, "payload": ("TypeB", {"b_value": 99})},
]

OUTPUT = "union_minimal.avro"

with open(OUTPUT, "wb") as f:
    fastavro.writer(f, parsed_schema, records, codec="null")

print(f"Wrote {len(records)} records to {OUTPUT}")

# Verify
with open(OUTPUT, "rb") as f:
    for record in fastavro.reader(f):
        print(record)
```

Virtual column, it's similar story like with protobuf oneof https://github.com/ClickHouse/ClickHouse/issues/76167

```
SELECT payload FROM file() WHERE payload.___name = 'TypeA'Or may be

SELECT payload FROM file() WHERE __payload_name = 'TypeA'
SELECT payload FROM file() WHERE payload_name = 'TypeA'
SELECT payload FROM file() WHERE payload.$name = 'TypeA' 
-- (chatgpt says that $ is cannot be used in field name, so compared to others, this option might be safe to use)
```

_No response_

### Additional context

_No response_

il9ue:
Closes #102300

- New Feature

Added setting `input_format_avro_union_type_name`. When enabled, an Avro union field exposes a `$name` sub-column with the active branch name, and each branch of a multi-branch union is exposed as a named sub-column holding that branch's value.

## Summary

With `input_format_avro_union_type_name` enabled (default off), a union field `P` gains:

* `P.$name` : `Nullable(String)`, the active branch name, `NULL` for the null branch.
* `P.BranchName` : `Nullable(branchType)`, that branch's value on rows where it is active and `NULL` elsewhere. Only for unions with more than one non-null branch, which map to `Variant`; a single non-null branch maps to `Nullable(T)` and is already directly accessible.
* For a branch record containing a union field, the inner union's `$name` one level deeper, e.g. `P.TypeA.inner.$name`.

## Example

```sql
SELECT id, `payload.$name`, `payload.TypeA`
FROM file('data.avro')
WHERE `payload.$name` = 'TypeA'
SETTINGS input_format_avro_union_type_name = 1;
```

## Design notes

* **`Nullable(String)`, not an Enum** : Avro record names can change via aliases, and with `schema_inference_mode='union'` the full name set isn't known up front.
* **`$` prefix** : not a legal Avro field character, so no collision with real fields.
* **Not a virtual column** : virtual columns are provided by the storage engine, while these sub-columns come from the format; `ISchemaReader` returns a plain column list with no virtual channel. They are registered by schema inference instead.
* **Branch sub-columns are `Nullable` by construction** : a branch column is `NULL` on every row holding a different branch. Declaring one non-nullable is rejected.

## Current limitations

* Nested unions: one level deep, first qualifying nested union field per branch, `$name` only. Inner branch *values* are not exposed as separate sub-columns, though they remain reachable inside the outer branch column. Selecting the nested `$name` together with the outer `$name` but without the union value column is not supported.
* On the AvroConfluent (streaming) read path the deserializer is built from the columns already declared on the table rather than from schema inference, so the sub-columns must be part of that declaration. Removing that requirement means extending schema inference onto the read path, which is a larger change, out of scope here.
* Not verified locally: end-to-end read through a live schema registry; the stateless harness doesn't run one. The shared code path is covered via file-based tests.


ilejn:
Hello @il9ue ,

I've added two tests https://github.com/ilejn/ClickHouse/commit/64724c5f71da7e603cc366a5f6ed7530a7445639
The first one shows a bug with early return, the second one shows a limitation of how inner types are handled.

Not sure that it is a great moment to go further, because we should clarify if the choice of virtual columns as indicators is reasonable.

il9ue:
Thanks for the addition and the wait @ilejn, I think both tests are fair. Below is my understanding, please interrupt me if I am misled.

The first one indicates the bug:
- When $name and a branch value sub-column (`nullable_payload.TypeA.x`) are requested together, the union node gets intercepted by the $name action and the branch leaf path never gets resolved, throwing a `THERE_IS_NO_COLUMN` exception. $name alone, or the branch sub-column alone, both work; it's the combination that breaks. I'll fix the routing so the $name action and the branch sub-column deserializers can coexist on the same union.

The second one is the inconsistency you're pointing at:
- Nested $name works through the explicit-schema path (your `outer.inner.$name` reads fine) but inference only walks the top-level fields, so `DESCRIBE` won't surface nested ones on its own. So "supported" depends on which door you came through, which isn't great.

I think it'd be better to split this into two phases:
- **Phase 1:** Fix the $name + branch-subcolumn collision, and make nested behavior consistent between inference and explicit schema. Will get this done.
- **Phase 2:** The design question you raised, which I don't want to rush. I spent a week on it and I'm fairly convinced `Nullable(String)` is right over an `Enum` (Avro record names drift via aliases, and with `schema_inference_mode='union'` the full name set isn't known up front, so a fixed enum is too rigid). What I haven't fully resolved yet is the indicator *mechanism* ; right now, $name is half virtual (inferred for file()/s3()) and half explicit (declared), and that split is the root of the inconsistency above.

I think there's need to dig into the streaming side before claiming anything: currently, so far only wires `AvroRowInputFormat`, and Kafka usually means `AvroConfluent`, which seems to be a separate path. Whether $name can show up there without being hand-declared is still open. Let me investigate that properly and come back with findings rather than guessing.

@UnamedRus, on Phase 2 (virtual column vs a declared indicator like Protobuf), do you have a preference before I go deeper? 
(Will start on Phase 1 regardless)



> I've added two tests [ilejn@64724c5](https://github.com/ilejn/ClickHouse/commit/64724c5f71da7e603cc366a5f6ed7530a7445639) The first one shows a bug with early return, the second one shows a limitation of how inner types are handled.
> 
> Not sure that it is a great moment to go further, because we should clarify if the choice of virtual columns as indicators is reasonable.

il9ue:
### Quick correction to my own comment above
▷ I built the PR head and executed the queries (issue's own schema `[null, TypeA{a_value}, TypeB{b_value}]`) during investigation of Phase 1, and found out that the actual AS-IS picture is different from what I wrote earlier.

#### Reproduced (on PR head commit):
```sql
-- payload maps to Variant(Tuple(a_value String), Tuple(b_value Int32))
SELECT id, `payload.$name` FROM file('union_minimal.avro', Avro) SETTINGS input_format_avro_union_type_name=1; -- TypeA / TypeB / NULL  ✅
SELECT `payload.TypeA.a_value` FROM file(...);            -- UNKNOWN_IDENTIFIER (47)
SELECT variantElement(payload,'TypeA') FROM file(...);    -- ILLEGAL_TYPE (43)
```

- The `$name + branch sub-column → THERE_IS_NO_COLUMN` collision didn't reproduce. There's no branch-value sub-column for `$name` to collide with ▷ access by name fails earlier at the parser (47/43). The 2-branch `Nullable` case selects `$name` + leaf together with no error. So Phase 1 as I framed it (a routing/deserializer fix) isn't the actual problem.
- AS-IS : when the union maps to `Variant`, the record names (`TypeA`/`TypeB`) are dropped ▷ only `Tuple(...)` survives. `$name` reports `TypeA`, but the value is reachable only via the type string, not the name. The discriminator and the value-access token are different alphabets ▷ which is the concrete form of your Enum-vs-text point, @ilejn.
- Nested `$name` is unsupported either way (not an inference-vs-explicit split): a union nested in a record surfaces in neither `DESCRIBE` nor an explicit select (47).

So I'd reframe Phase 1 toward making a branch's value reachable by the name `$name` indicates (projection, not just filtering), rather than a deserializer fix.

The root cause is that the name and the value live on two paths that aren't linked:
- `avroNodeToDataType` builds the `Variant` from branch **types** only (`AvroRowInputFormat.cpp L1639-1644`), so record names (`TypeA`) are dropped — hence `DESCRIBE` shows `Variant(Tuple(...), Tuple(...))`.
- `$name` recomputes the name on the fly via `nodeName(branch_node)` (L1074, L1148); there's no persisted name->branch mapping on the value path.
- `variantElement` matches branches by **type** (L1015-1016), which is why `variantElement(payload,'Tuple(a_value String)')` works but `'TypeA'` doesn't.

So either approach needs the same missing piece; a name->branch-index mapping (built with `nodeName`, matching what `$name` emits). On top of that:

1. resolve a name segment in the union arm at L1135 (where the `$name` handling already sits, L1106/L1139) ▷ looks like the more local change, contained in the Avro reader.
2. extend `variantElement` to accept the branch name — but that reaches outside the Avro path.

I haven't verified either end-to-end yet, so I don't want to commit to the mechanism ([1] looks lighter from the code) but I'll probe it on the issue's schema before claiming further. Does this sound match your intent?

_Phase 2 (Enum-vs-String) and the AvroConfluent path still stand as I described. (will update after further investigation)_ 
Please leave any feedback on 



> * **Phase 1:** Fix the $name + branch-subcolumn collision, and make nested behavior consistent between inference and explicit schema. Will get this done.

il9ue:
**[Update]** 
implementation working with tests at local. Sharing the exact failure I traced and how it's handled.

## Root cause (the `$name` value-access gap)

`field.$name` works because the schema reader registers it as a real column. A branch's *value*, however, was not reachable by that name: `field.BranchName` failed at analysis time with `UNKNOWN_IDENTIFIER`, because branch names were never registered as queryable identifiers and `Variant` has no dotted sub-column access (`variant.X` is parsed as `tupleElement`, which Variant rejects with code 43). 
So `$name` reported `TypeA`, but the value could only be read via `variantElement(payload, 'Tuple(...)')`(discriminator and value-access token in different characters)

(For the record: I could not reproduce a `THERE_IS_NO_COLUMN` collision between `$name` and a branch sub-column, on either fixture, via inference or explicit schema. The real issue was this missing registration, not a deserializer collision, correcting my earlier comment.)

## Implementation (one file, AvroRowInputFormat)

- **Schema reader**: emit a companion column `field.BranchName` (typed `Nullable(branchType)`) for each non-null branch of a multi-branch (Variant) union, using the same name `$name` reports. Single-non-null-branch unions (mapped to `Nullable`) are left as-is; their value is already accessible, so a branch column there would just be a confusing duplicate.
- **Entry condition**: the unified union action is now entered when `$name` *or* any branch column is requested (`unionHasCompanionColumns`), not only when `$name` is present.
- **Fan-out**: the active branch column is filled from the single decoded union index; inactive branches are set NULL. One `decodeUnionIndex` per union per row is preserved (a second decode would silently desync the stream). The `$name` column is now optional (tolerates `name_col_idx == -1`).

## Tests added 

- **Branch projection (`$name` + both branches)** : the core feature; verifies the active branch gets its value and others are NULL.
- **Branch projection alone (no `$name`, no value)** : this *regressed* mid-way: branch columns were only wired when `$name` was also requested. Guards the entry-condition fix.
- **Filter by `$name` + project that branch** : the issue's original UX (`WHERE payload.$name='TypeA'` then read the value).
- **Value + branch column together** : documents the known limit below; guards against it silently changing.
- **Nullable union exposes no branch column** : without this guard the reader added a duplicate `nullable_payload.TypeA` column; this pins the Variant-only behavior.

Test passes under `clickhouse-test 04098_avro_union_type_name`.

## Open question (the one known limit)

When the *whole* union value and a branch column are requested together (`SELECT payload, \`payload.TypeA\``), the branch column returns NULL; the union datum is consumed once by the value column and didn't want to duplicate the decode. 
(Is filling both worth it, or is NULL-on-conflict acceptable? It overlaps with the indicator-mechanism design you raised, so I'd rather get your read before extending)

(Phase 2 ▸ Enum vs `Nullable(String)`, and the AvroConfluent path still open; separate follow-up later)

ilejn:
1. Minor code style issues, e.g.
```
    for (int c : branch_column_idxs)
        if (c >= 0) { has_branch_cols = true; break; }
```
Consider using clang-format
2. To me, 'Known limit: value + branch column together -> branch is NULL =' is worrying. It is very errorprone behavior. We have to either have it working properly or raise exception.
3. Still not sure about nested columns. Companion-branch detection is only one level deep, so nested branch subcolumns are still missed. unionHasCompanionColumns only looks for current_path + ".$name" and current_path + "." + nodeName(branch)
4. Am I right that if variant_payload has branches TypeB and TypeC, current PR infers:
      - variant_payload
      - variant_payload.$name
      - variant_payload.TypeB
      - variant_payload.TypeC
But two latest ones are not documented?

Summing up. I am still a bit skeptical about the approach, mainly because it is not applicable to distribute incoming messages among several Materialized Views (correct me if I am wrong). Leaving this concern out of the scope, we should polish this PR further and add tests that cover deeper nesting.

il9ue:
Thanks for the review and proposed answers below. Let me start with MV distribution, since I think that's the core mechanism behind the conceptual side of this issue.

> Summing up. I am still a bit skeptical about the approach, mainly because it is not applicable to distribute incoming messages among several Materialized Views (correct me if I am wrong). Leaving this concern out of the scope, we should polish this PR further and add tests that cover deeper nesting.

- The Kafka path shares the fill logic with the file path (readRow → deserializeRow / createAction), so the same code that fills branch columns for files also runs on Kafka; but only if the branch column is declared. Kafka builds its deserializer from output.getHeader(), the columns already declared on the table, not from schema inference, so it doesn't create the branch columns on its own the way a file DESCRIBE does. And declaring one as Nullable(Tuple) needs the experimental type check hint turned on (allow_experimental_nullable_tuple_type; enable_nullable_tuple_type on current master), in that check runs only on hand-declared columns, not on inferred ones. So the pieces for distributing across several MVs (each filtering on payload.$name) are there on Kafka once the columns are declared with that setting on. I confirmed the shared code path and the declared-column behavior on the file path, but couldn't run the Kafka end-to-end locally (yet) though..
- Virtual columns won't help here. Virtual columns (like Kafka's _topic) come from the storage engine, but these branch columns come from the Avro format, not from the storage engine. Virtual columns are a storage-engine feature and the format (parsing) layer has no mechanism to register them. It only returns a plain list of columns (`ISchemaReader`). What lets a column skip the experimental type check (hint) is schema inference/deduction, not by virtual columns (inferred columns skip it, hand-declared ones donot). So one way to avoid manual declaration on Kafka would be to extend schema inference onto the read path. That seems a bigger change AFAIK.

**other answers** 
1. Will apply.
2. value + branch = NULL. I agree too, and should be fixed on next commit rather than left as a limit. The 04098 "known limit" case now asserts the active branch is filled instead of NULL, and the fill reuses the value the union already decoded (no second decode).
3. Nested columns. Confirmed. What I found : 
- When a union field is requested together with its sub-columns (.$name or a branch like .TypeA), createAction routes it to createUnionWithNameAction and returns. It then deserializes each branch with createDeserializeFn but never calls createAction on the branch again. So if a branch itself contains another union, that inner union is never assessed, and its subcolumns are never created. Detection stops at the outermost union.
_(I haven't yet worked out how far nested support can go. Nesting means a Variant inside a Variant and a second union decode inside the branch, and I haven't checked whether those behave. So I'll have to look into it with tests and so on.)_
4. Undocumented branch columns: for variant_payload with branches TypeB/TypeC the schema is variant_payload, .$name, .TypeB, .TypeC, and the last two aren't documented, and will add them.

@ilejn Please let me know if these answers help to proceed.

ilejn:
> @ilejn Please let me know if these answers help to proceed.

Not sure what I supposed to do ;)

Yes, your remarks make sense, please address them.

il9ue:
### Remarks
Three things included since the last update.
- value + branch = NULL (your point 2) :  fixed rather than left as a limit. When the value column is also selected it already has deserialized the active branch into its ColumnVariant, so the branch column filled from there (read via the Variant's stored discriminator, wrapped with insertFromNotNullable), inactive branches NULL. No second decode. The 04098 "known limit" case now asserts the branch is filled.
- Merged master. Upstream added a schema_inference_allow_nullable_tuple_type parameter to avroNodeToDataType, so an Avro [null, record] union now infers as plain Tuple unless enable_nullable_tuple_type is on; the test enables it explicitly to keep covering the Nullable(Tuple) value-column path. Branch sub-columns are unaffected, and the format wraps them Nullable itself, since a branch column is NULL on rows holding a different branch.
- Segfault found and fixed. Checking the above (on merge stage), the value+branch fill path crashes when the branch sub-column is declared non-nullable in an explicit schema: assert_cast<ColumnNullable &> on a ColumnTuple is a plain static_cast under NDEBUG. Without the value column it didn't crash but silently stored ('') on inactive rows. Both are now rejected up front with an ILLEGAL_COLUMN error naming the correct type. A branch-column-only request with an explicit non-nullable type goes through the generic Avro union-branch flattening path instead, which I left unchanged.

Docs (your point 4) and nested-union coverage (point 3) next.

il9ue:
Also **found and fixed a crash** while testing
- The nested wiring decided whether an inner union qualifies by counting its total leaves, so a [null, X] union slipped through and got assert_cast to DataTypeVariant even though it maps to Nullable(X). In a release build that cast is unchecked, so it read unrelated memory and threw std::length_error. It now counts non-null branches instead, matching how readSchema decides whether a union becomes Variant.

PR #103153, opened 2026-04-20
Avoid conflicts on renamed files when creating cherry-picks

nickitat:

---

When the patched file was renamed between the `cherrypick_branch` and `master`, our automatic cherry-pick PR will have a conflict. Moreover, this conflict will be "too complex to resolve in the UI", i.e., it will require manual intervention. The required manipulation cannot be more stupid - you just need to run `git merge` locally, and `git` will resolve everything automatically.
I'm not sure if there was any strong reason to use `--no-commit` and always outsource the merge to GH, so I need some feedback from the CI team. 

Was tested on https://github.com/ClickHouse/ClickHouse/pull/102849. 

### Version info
- Merged into: `26.5.1.449`

### DIFF
--- a/tests/ci/cherry_pick.py
+++ b/tests/ci/cherry_pick.py
@@ -252,13 +252,21 @@ close it.
         # other actions are required. It's possible when changes are backported
         # manually to the release branch already
         try:
-            output = git_runner(
+            git_runner(
                 f"{GIT_PREFIX} -c merge.renameLimit=999999 "
                 f"merge --no-ff --no-edit {self.pr.merge_commit_sha}"
             )
-            # 'up-to-date', 'up to date', who knows what else (╯°v°)╯ ^┻━┻
-            if output.startswith("Already up") and output.endswith("date."):
-                # The changes are already in the release branch, we are done here
+            # The merge succeeded. If it produced no tree change vs
+            # backport_branch, the PR is effectively already backported to
+            # the release branch - either "Already up to date" (no merge
+            # commit at all) or an empty merge commit whose resolution
+            # collapsed onto backport_branch's tree (e.g. the PR was
+            # manually applied with equivalent content). In either case,
+            # skip creating an empty cherry-pick PR.
+            if not git_runner(
+                f"{GIT_PREFIX} diff --name-only "
+                f"{self.backport_branch} {self.cherrypick_branch}"
+            ):
                 logging.info(
                     "Release branch %s already contain changes from %s",
                     self.name,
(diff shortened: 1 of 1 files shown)


leshikus:
1. This patch is quite risky. 2. What I can see - it mentions "renamed" files, but there are no "renamed" files in git, so it likely does something else than the description said. I suggest starting improving the description so we all understand what the patch does. It likely replaces one merge strategy with another. The description mentions the problem, but it does not describe why proposed fix fixes the problem.

Upd. Why do you change `merge.renameLimit`? Is the default value too small? Do we often rename more than 400 files?

Upd2. `Abort and fall back to the old behavior` - the code does not exactly fallback to the old behavior, there are at least minor differences like changing `git reset --merge` to `git merge --abort`, but likely there are more. If the patch would be done in a way, it clearly falls back to old behavior, it would be less risky

tests/ci/cherry_pick.py
```diff
@@ -238,22 +238,35 @@ def create_cherrypick(self):
...
-        )
+        # Second step, create cherrypick branch on top of backport_branch and
+        # perform the merge locally. When it succeeds, cherrypick_branch points
+        # at a merge commit with the resolution baked in, so GitHub's merge of
+        # the cherry-pick PR becomes trivial. This avoids a failure mode where
+        # files renamed between the release branch and master produce conflicts
+        # in GitHub's merge even though local `git merge` resolves them via
+        # rename detection. The rename limit is raised to prevent git from
+        # silently disabling rename detection on large diffs.
+        git_runner(f"{GIT_PREFIX} checkout -B {self.cherrypick_branch}")
```
Felixoid: `checkout -B` checks it out from the `HEAD`, which is the `RELEASE_BRANCH_REF`, it's not what we want here at all. It must be the `self.pr.merge_commit_sha`


```suggestion
        git_runner(
            f"{GIT_PREFIX} checkout --no-track -B "
            f"{self.cherrypick_branch} {self.pr.merge_commit_sha}"
        )
```

tests/ci/cherry_pick.py
```diff
@@ -262,11 +275,15 @@ def create_cherrypick(self):
...
-            git_runner(f"{GIT_PREFIX} reset --merge")
+            # Local merge has conflicts. Abort and fall back to the old
+            # behavior: cherrypick_branch points at the raw master commit,
+            # and conflicts are surfaced on the GitHub PR for manual
+            # resolution by the assigned engineer.
+            git_runner(f"{GIT_PREFIX} merge --abort")
+            git_runner(
+                f"{GIT_PREFIX} checkout -B "
+                f"{self.cherrypick_branch} {self.pr.merge_commit_sha}"
+            )
```
Felixoid: We can keep everything like it is, then we're already in the cherry-pick branch right now


```suggestion
            git_runner(f"{GIT_PREFIX} merge --abort")
```

Felixoid (changes requested):
I got the logic, but I don't like that the cherry-pick branches become inconsistent depending on `is it mergeable or not`

Let's always use the merge_commit_sha as the cherry-pick heads and then try to merge the backport branches into it.

Felixoid (approved):

PR #90158, opened 2025-11-17
Add support for projection-level settings

amosbird:
### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
This PR introduces projection-level settings, exposed through the new WITH SETTINGS clause in ALTER TABLE ... ADD PROJECTION. These settings allow projections to override certain MergeTree storage parameters (e.g. index_granularity, index_granularity_bytes) on a per-projection basis.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

### Details

PR #103161, opened 2026-04-20
Compare rows instead of marks to select correct projection when using different index_granularity settings

sberss:
The `WITH SETTINGS` functionality allowing us to configure `index_granularity` per projection is great ([#90158](https://github.com/ClickHouse/ClickHouse/pull/90158)). However, it has some edge cases that lead to questionable behaviour when your PK and projections have a big discrepancy in `index_granularity`. This is due to comparing the number of filtered marks instead of the number of filtered rows.

Below is a simple reproduction of the issue.

```
CREATE TABLE default.projection_testing
(
    `x` UInt32,
    `y` UInt32,
    PROJECTION prj_y
    (
        SELECT
            x,
            y
        ORDER BY y
    ) WITH SETTINGS (index_granularity = 8)
)
ENGINE = MergeTree
PRIMARY KEY x
ORDER BY x
SETTINGS index_granularity = 8192

INSERT INTO projection_testing select number % 10000000, number % 16 from system.numbers LIMIT 50000000
```

When filtering by `y` on the table, we see that the projection was not chosen, despite the fact that 390628 marks on the projection is ~3M rows, and 6105 marks on the PK is ~50M rows.

```
EXPLAIN projections = 1
SELECT *
FROM projection_testing
WHERE y = 0
FORMAT raw

Query id: aeb36ab5-3462-4188-86fb-06f9b3f2f1d1

Expression ((Project names + Projection))
  Expression ((WHERE + Change column names to column identifiers))
    ReadFromMergeTree (default.projection_testing)
    Projections:
      Name: prj_y
        Description: Projection prj_y is usable but requires reading 390628 marks, which is not better than the original table with 6105 marks
        Condition: (y in [0, 0])
        Search Algorithm: binary search
        Parts: 5
        Marks: 390628
        Ranges: 5
        Rows: 3125024
        Filtered Parts: 0
```

Changing this to use row count feels more appropriate. With this change, the projection analysis uses the number of rows instead, and will see that ~3M rows < ~50M rows and therefore choose `prj_y`.

```
EXPLAIN projections = 1
SELECT *
FROM projection_testing
WHERE y = 0
FORMAT raw

Query id: 2f9b6628-0ed1-4bcc-bfb8-515c1ded48e2

Expression ((Project names + Projection))
  Expression
    ReadFromMergeTree (prj_y)

```

- Improvement

Compare rows instead of marks to select correct projection when using different index_granularity settings.

### DIFF
--- a/src/Processors/QueryPlan/Optimizations/optimizeUseAggregateProjection.cpp
+++ b/src/Processors/QueryPlan/Optimizations/optimizeUseAggregateProjection.cpp
@@ -730,20 +730,23 @@ std::optional<String> optimizeUseAggregateProjections(
                 stat.filtered_parts = candidate.filtered_parts;
                 candidate.stat = &stat;
 
-                size_t parent_reading_marks = parent_reading_select_result->selected_marks;
-                if (candidate.sum_marks > parent_reading_marks)
+                size_t parent_reading_rows = parent_reading_select_result->selected_rows;
+                if (candidate.sum_rows > parent_reading_rows)
                 {
                     stat.description = fmt::format(
-                        "Projection {} is usable but requires reading {} marks, which is not better than the original table with {} marks",
+                        "Projection {} is usable but requires reading {} marks consisting of {} rows, "
+                        "which is not better than the original table with {} marks consisting of {} rows",
                         candidate.projection->name,
                         candidate.sum_marks,
-                        parent_reading_marks);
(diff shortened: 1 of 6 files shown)


sberss:
@amosbird It's been suggested to me you would be the best reviewer for this PR. I would appreciate a look if you get a chance.

PR #103169, opened 2026-04-20
Fix the bug about supertype for IN with Tuple

### LINKED ISSUE #103055: `convertTupleToArray` CASTs tuple elements to LHS type, causing silent integer overflow and wrong `IN` results

_Retrospective finding from a historical scan of [PR #77906](https://github.com/ClickHouse/ClickHouse/pull/77906) (merged 2025-12-01). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`toUInt8(232) IN (1000, number)` returns 1 instead of 0 because `CAST(1000, 'UInt8')` overflows to 232, making a false match. The constant equivalent `toUInt8(232) IN (1000, 0)` correctly returns 0.

**Root cause:** resolveFunction.cpp:238: `common_type = in_first_argument->getResultType()` uses only the LHS type as the CAST target. It should compute the least supertype of the LHS and all RHS element types to prevent overflow. The array path (Case 1) doesn't have this issue because it uses the array's own element type, which is already computed by the `array()` function using `getLeastSupertype`.

**Why we believe this is a bug:** resolveFunction.cpp:238 sets `common_type = in_first_argument->getResultType()` (UInt8). Then line 247 CASTs each tuple element to this type via `castNodeToType`. When a tuple element (1000) exceeds the LHS type range, CAST truncates it, creating false matches.

**Affected locations:**
- `src/Analyzer/Resolve/resolveFunction.cpp:238` — common_type assignment in convertTupleToArray

**Impact:** Silent wrong results when IN tuple elements have values outside the LHS type's range. `toUInt8(x) IN (256+x, ...)` always matches due to overflow. `toInt8(x) IN (128+x, ...)` also produces wrong matches. Any narrow numeric type with wider tuple elements is affected.

Yes — confirmed on current `master` (commit `b839b457d33c`).

### How to reproduce

```sql
-- Test: convertTupleToArray CAST overflow causes wrong IN results
-- Constant case (correct, uses set-based path with proper type promotion)
SELECT toUInt8(232) IN (1000, 0);
-- Non-constant case (buggy, CAST(1000, 'UInt8') = 232 causes false match)
SELECT toUInt8(232) IN (1000, number) FROM numbers(1);
-- Another case: UInt8(0) matches 256 due to overflow
SELECT toUInt8(0) IN (256, 999);
SELECT toUInt8(0) IN (256, number + 999) FROM numbers(1);
-- NOT IN also affected
SELECT toUInt8(232) NOT IN (1000, 0);
SELECT toUInt8(232) NOT IN (1000, number) FROM numbers(1);
-- Array path is NOT affected (uses proper type promotion)
SELECT toUInt8(232) IN [1000, number] FROM numbers(1);
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/35e01027-cdc2-4223-ac32-a5bc16be432a)

### Expected behavior

```
0
0
0
0
1
1
0
```

```
0
1
0
1
1
0
0
```

### Additional context

**Open risks:**
- Any query with narrow numeric LHS and wider-range tuple elements will silently return wrong results
- Float-to-Int conversions may also lose precision in unexpected ways

**Suggested fix:** Compute `common_type` using `getLeastSupertype` across the LHS type and all tuple element types, similar to how the `array()` function determines its element type. This ensures no overflow during CAST.

Found during automated review of [PR #77906](https://github.com/ClickHouse/ClickHouse/pull/77906).

---
_ClickGapAI · Confidence: HIGH · Severity: P0 · Finding: `h_pr77906_002`_

yariks5s:

IN with non-constant tuple second argument must not CAST tuple elements down to the LHS type, this can overflow. Closes https://github.com/ClickHouse/ClickHouse/issues/103055.

### Version info
- Merged into: `26.5.1.615`


Avogar (approved):

yariks5s:
Stress test is [broken](https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICdDaGVjayBmYWlsZWQnCiAgICAtLSBBTkQgY2hlY2tfbmFtZSA9ICdTdHJlc3MgdGVzdCAoYW1kX21zYW4pJwogICAgQU5EIHRlc3Rfc3RhdHVzIElOICgnRkFJTCcsICdFUlJPUicpCiAgICBBTkQgKHB1bGxfcmVxdWVzdF9udW1iZXIgPSAwIE9SIGJhc2VfcmVmIElOICgnbWFzdGVyJykpCiAgICAtLSBBTkQgdGVzdF9jb250ZXh0X3JhdyBMSUtFICclcGF0dGVybiUnICAtLSB1bmNvbW1lbnQgYW5kIGVkaXQgdG8gZmlsdGVyIGJ5IGZhaWx1cmUgcGF0dGVybgpHUk9VUCBCWSBkYXkKT1JERVIgQlkgZGF5IERFU0MK)

clickgapai:

**Affected code:** `QueryAnalyzer::convertTupleToArray` in src/Analyzer/Resolve/resolveFunction.cpp — introduced in 25.9.

**Why this might apply:** The buggy code in `convertTupleToArray` was introduced after the 25.8 release (resolveFunction.cpp was created 2025-09-08, after 25.8 release on 2025-08-29). The bug causes silent wrong results in common queries like `x IN (large_value, column)` when x has a narrow type. This is a user-visible data correctness bug affecting released versions.

Other supported branches (`25.8`) predate 25.9 and do not contain this code.

PR #98939, opened 2026-03-06
Text index postprocessor

Ergus:
Parameters:

```
postprocessor : The postprocessor is an arbitrary expression that transforms each token after tokenization.
```

Example use:

```sql
CREATE TABLE users (
    str String,
    index idx() type text(tokenizer = 'spliByNonApha', postprocessor = lower(str))
ENGINE = MergeTree ORDER BY tuple();

SELECT * FROM users WHERE hasAnyTokens(name, 'crash');
```

### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Added a postprocessor to the text index which transforms the tokens after tokenization. 

### Version info
- Merged into: `26.7.1.47`
- Backported to: `26.6.1.1191`

PR #103172, opened 2026-04-20
Text index: store positions for a better phrase search

ahmadov:
Implements #101473 for storing positional information for phrase search.

This positional data is stored as 96-bit "[32][32][32]" layout. This allows us to store larger token positions.

The following points would be implemented in follow-up PRs:
- [ ] Use delta+bitpacking / pfor based compression to compress Roarish entries further - by design they are sorted values - #108277 
- [ ] SIMD vectorized phrase search (NEON, AVX2, AVX512) - current implementation is scalar.
 
- New Feature

Add new `positions` argument support to text indexes to store token positions and enable exact direct reads for `hasPhrase`. The `allow_experimental_text_index_positions` needs to be enabled to store positional information.

### Version info
- Merged into: `26.7.1.42`
- Backported to: `26.6.1.1190`


docs/en/engines/table-engines/mergetree-family/textindexes.md
```diff
@@ -90,6 +90,7 @@ CREATE TABLE table
                                 [, dictionary_block_frontcoding_compression = B]
                                 [, posting_list_block_size = C]
                                 [, posting_list_codec = 'none' | 'bitpacking' ]
+                                [, __experimental_positions = 0 | 1 ]
```
ahmadov: I named parameter like that because we want to keep it as "hidden". I had a discussion with @rschu1ze to keep it kinda hidden and name it properly once we know the layout is final.

src/Storages/MergeTree/MergeTreeIndexText.cpp
```diff
@@ -1322,6 +1355,20 @@ DictionarySparseIndex serializeTokensAndPostings(
         {
             auto & postings = *tokens_and_postings[i].second;
             auto token_info = TextIndexSerialization::serializePostings(postings, postings_stream, params, postings_serialization);
+
+            /// Serialize position data if phrase query support is enabled.
+            if (positions_stream && i < tokens_and_positions.size())
+            {
+                auto & position_builder = *tokens_and_positions[i].second;
```
Ergus: Minor:
Could we add some check (maybe a `chassert` to not affect release performance) to ensure the invariant between postings and positions? If for any reason we miss the alignment between both (format change, [de]serialization error or change) it will be pretty hard to find the problem.

src/Storages/MergeTree/MergeTreeIndexText.cpp
```diff
@@ -1322,6 +1355,20 @@ DictionarySparseIndex serializeTokensAndPostings(
         {
             auto & postings = *tokens_and_postings[i].second;
             auto token_info = TextIndexSerialization::serializePostings(postings, postings_stream, params, postings_serialization);
+
+            /// Serialize position data if phrase query support is enabled.
+            if (positions_stream && i < tokens_and_positions.size())
```
Ergus: `i < tokens_and_positions.size()` this only detects out of bound, not misalignment.
But also, IIUC when positions are enabled, every postings token is expected to have a positions slot, so a size shortfall should be a hard error, not a quiet skip right?

src/Storages/MergeTree/TextIndexPositionCodec.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace DB
+{
+
+void TextIndexPositionCodec::encode(const std::vector<RoaringishEntry> & entries, WriteBuffer & out)
+{
+    static_assert(sizeof(RoaringishEntry) == 12);
+
+    UInt64 count = entries.size();
+    writeVarUInt(count, out);
```
Ergus, line 27: Isn't this count the same than `position_cardinality`? Are them intended to be different at some point? WHen we must use one or the other?

src/Storages/MergeTree/TextIndexPositionCodec.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+    if (count > 0)
+        out.write(reinterpret_cast<const char *>(entries.data()), count * sizeof(RoaringishEntry));
```
Ergus, line 32: Two details here:
1. I have seen some `[write|read]Big` functions around. Not sure if we have those here and depending of the size of these entries maybe consider using them?

2. Should we worry about cross platform support? Because I have seen some big-endian support use in the codebase (`transformEndianness` / `writeBinaryLittleEndian` / `readBinaryLittleEndian`) an MergeTree on-disk data is written little-endian explicitly. 

Something like `if constexpr (std::endian::native == std::endian::little)`

src/Storages/MergeTree/TextIndexPositionCodec.cpp
```diff
@@ -0,0 +1,33 @@
...
+    static_assert(sizeof(RoaringishEntry) == 12);
+
+    UInt64 count = 0;
+    readVarUInt(count, in);
+
+    if (count == 0)
+        return;
+
+    entries.resize(count);
+    in.readStrict(reinterpret_cast<char *>(entries.data()), count * sizeof(RoaringishEntry));
```
Ergus: There is a new `readBigStrict` in case it is needed here.

src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
```diff
@@ -285,6 +306,40 @@ void MergeTreeReaderTextIndex::classifyVirtualColumns()
...
+        else if (
+            search_query->search_mode == TextSearchMode::Phrase
+            && search_query->direct_read_mode == TextIndexDirectReadMode::Exact
+            && fallback_reader && fallback_expressions.contains(column.name))
+        {
+            /// For phrase queries with positions, check selectivity before reading positional data.
+            /// Reading large position lists for common phrases is slower than evaluating `hasPhrase`
+            /// on physical data via the fallback path. Estimate the phrase cardinality as the
+            /// intersection of its tokens (a safe upper bound) from the analyzer's per-token cardinalities.
+            const auto & all_token_infos = analyzer.getAllTokenInfos();
```
Ergus, line 322: I have a question here.

The `analyzer` usually comes from a granule, so the information it holds is granule specific. 

But bellow there is the `num_rows_in_part` which is full part information used to decide.
Is that intended?

(If the user for some reason sets a small granularity this will be affected right?)

src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
```diff
@@ -285,6 +306,40 @@ void MergeTreeReaderTextIndex::classifyVirtualColumns()
...
+            const auto & settings = condition_text.getContext()->getSettingsRef();
+            double selectivity_threshold = static_cast<double>(settings[Setting::text_index_hint_max_selectivity]);
+            size_t num_rows_in_part = data_part_info_for_read->getRowCount();
+
+            double log_cardinality = 0.0;
+            bool all_tokens_present = num_rows_in_part > 0;
```
Ergus: Very Minor: `const bool a = (xxx > 0);`

tests/queries/0_stateless/02346_text_index_function_hasPhrase_positions.sql
```diff
@@ -0,0 +1,345 @@
+-- Tags: no-parallel-replicas
+
+SET enable_analyzer = 1;
```
Ergus, line 3: Could we have a partially materialized index test?

src/Storages/MergeTree/TextIndexUtils.cpp
```diff
@@ -445,6 +482,38 @@ bool MergeTextIndexesTask::executeStep()
...
+        if (params.positions)
+        {
+            const auto & token_info = inputs[current->order].token_infos[current->getRow()];
+            if (token_info.header & PostingsSerialization::Flags::HasPositions)
+            {
+                auto * pos_stream = input_streams[current->order].at(MergeTreeIndexSubstream::Type::TextIndexPositions);
+                auto * pos_data_buffer = pos_stream->getDataBuffer();
+                pos_stream->seekToMark({token_info.position_offset, 0});
+
+                std::vector<RoaringishEntry> position_entries;
```
Ergus: Should we use `VectorWithMemoryTracking<RoaringishEntry>` instead?

src/Storages/MergeTree/TextIndexUtils.h
```diff
@@ -131,6 +132,8 @@ class MergeTextIndexesTask : public MergeProjectionsIndexesTask
     std::vector<TokenPostingsInfo> output_infos;
     /// Postings accumulated for the current token.
     PostingList output_postings;
+    /// Positions accumulated for the current token (phrase query support).
+    std::vector<RoaringishEntry> output_positions;
```
Ergus: Same here. (Should we use `VectorWithMemoryTracking<RoaringishEntry>` instead?)

Ergus (commented):
In general LGTM

src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
ahmadov: since the text index is build for the whole part (infinite granularity), there should be a single granule per part.

src/Storages/MergeTree/TextIndexUtils.cpp
ahmadov: replaced with `PODArray` instead.

tests/queries/0_stateless/02346_text_index_function_hasPhrase_positions.sql
ahmadov: yes, it was already added here:
https://github.com/ClickHouse/ClickHouse/blob/2f22233b20009d79e1fcec7947308cd89d4d425c/tests/queries/0_stateless/02346_text_index_partially_materialized.sql#L145-L209

src/Storages/MergeTree/TextIndexUtils.cpp
```diff
@@ -461,6 +498,38 @@ bool MergeTextIndexesTask::executeStep()
             output_postings |= *posting;
         }
 
+        /// Read and merge position data if positions are enabled.
+        if (params.positions)
```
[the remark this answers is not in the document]
ahmadov: currently, it's not possible to have both V1 and V2 at the same time. This could be fixed later when it's possible to have both versions

src/Storages/MergeTree/MergeTreeIndexText.cpp
```diff
@@ -1356,19 +1402,32 @@ void MergeTreeIndexGranuleTextWritable::serializeBinaryWithMultipleStreams(Merge
...
+        auto it = streams.find(MergeTreeIndexSubstream::Type::TextIndexPositions);
+        if (it == streams.end() || !it->second)
+            throw Exception(ErrorCodes::LOGICAL_ERROR, "Index with type 'text' and positions enabled must have a positions stream");
+        positions_stream = it->second;
+    }
+
+    /// When positions are stored, the dictionary entries carry extra position metadata, so the
+    /// header version is bumped. Older servers reject `WithPositions` parts instead of misparsing them.
+    auto serialization_version = static_cast<MergeTreeIndexVersion>(
+        params.positions ? TextIndexHeader::Version::WithPositions : TextIndexHeader::Version::WithCodec);
```
CurtizJ: This is slightly wrong. We should bump the version unconditionally to `WithPositions` and write a flag `has_positions` in the header. Now, it's not clear how to increase the version further. What if I want to increase the format version but create an index without positions?

src/Storages/MergeTree/MergeTreeIndexText.cpp
```diff
@@ -1757,10 +1868,13 @@ MergeTreeIndexPtr textIndexCreator(StorageMetadataPtr metadata_snapshot, const I
     UInt64 posting_list_block_size = extractFieldOption<UInt64>(options, ARGUMENT_POSTING_LIST_BLOCK_SIZE)
         .value_or(settings[MergeTreeSetting::text_index_posting_list_block_size]);
 
+    UInt64 positions = extractFieldOption<UInt64>(options, ARGUMENT_POSITIONS).value_or(DEFAULT_POSITIONS);
```
CurtizJ, line 1893: I think it's better to gate this with an experimental MergeTree setting. Now, nothing prevents the user from creating an index with experimental positions, even if the experimental tier features are disabled on service.

src/Storages/MergeTree/TextIndexPositionData.h
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    size_t size() const { return entries.size(); }
+    bool empty() const { return entries.empty(); }
+    std::vector<RoaringishEntry> & getEntries() { return entries; }
+    const std::vector<RoaringishEntry> & getEntries() const { return entries; }
+
+    /// Merge another builder's entries into this one.
+    /// Both must be sorted. Result is sorted with same-bucket entries merged.
+    void mergeFrom(const PositionListBuilder & other)
```
CurtizJ: This looks unused. Consider removing it, as this is quite complex code?

src/Storages/MergeTree/MergeTreeIndexConditionText.cpp
```diff
@@ -358,6 +378,15 @@ bool MergeTreeIndexConditionText::mayBeTrueOnGranule(MergeTreeIndexGranulePtr id
             bool exists_in_granule = granule->hasAllQueryTokens(*text_search_query);
             rpn_stack.emplace_back(exists_in_granule, true);
         }
+        else if (element.function == RPNElement::FUNCTION_HAS_PHRASE)
+        {
+            /// At the granule level, phrase matching degrades to "all tokens must exist".
+            /// Actual positional phrase checking is done at the row level via position data.
```
CurtizJ, line 384: Maybe it's better to read the positions stream directly, even for granule pruning. Since it has row IDs, we can create bitmaps from the row-ID portion of `RoaringishEntry` for the small postings used in the first stage of analysis. But let's postpone this as a performance optimization.

src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
```diff
@@ -123,17 +130,28 @@ void MergeTreeReaderTextIndex::setIndexGranule(MergeTreeIndexGranulePtr index_gr
 void MergeTreeReaderTextIndex::initializeFallbackReader(const IMergeTreeReader * main_reader)
 {
     const auto & condition_text = assert_cast<const MergeTreeIndexConditionText &>(*index.condition);
-    if (!condition_text.hasSearchPatterns())
+    /// Check if any virtual column may need a fallback path:
+    /// - Pattern queries (LIKE): fallback when dictionary scan is abandoned.
+    /// - Phrase queries (hasPhrase with Exact mode): fallback when estimated cardinality is too high
```
CurtizJ, line 137: I thought that the evaluating phrase from the index should always be faster, since we read less data, and don't need retokenization.

src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
```diff
@@ -746,6 +836,90 @@ void MergeTreeReaderTextIndex::fillColumnLazy(IColumn & column, const String & c
...
+                position_lists.reserve(position_offsets.size());
+
+                auto * data_buffer = positions_stream->getDataBuffer();
+                {
+                    ProfileEventTimeIncrement<Microseconds> decode_watch(ProfileEvents::TextIndexPositionsDecodeMicroseconds);
+                    for (auto position_offset : position_offsets)
+                    {
+                        positions_stream->seekToMark({position_offset, 0});
+                        auto & positions = position_lists.emplace_back();
+                        TextIndexPositionCodec::decode(*data_buffer, positions);
```
CurtizJ, line 896: This could be a large amount of data loaded into memory. I think we should write positions split by segments (the same as posting lists) with boundaries (min and max doc ids), taken from the posting list segments for simplicity. Therefore, a segment of positions will have more rows than a segment of the posting list, but the boundaries will be the same. And read positions in a streaming fashion from multiple streams (a stream per token).

CurtizJ: Also, we may now read a lot of unnecessary data from previously filtered ranges.

src/Storages/MergeTree/TextIndexPhraseSearch.cpp
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    if (matches.empty())
+        return result;
+
+    /// Sort by key and coalesce duplicate (doc_id, group) keys (OR their bitmaps).
+    /// Sorting by key alone suffices: same-key entries become adjacent and bitmap OR is
+    /// order-independent.
+    std::sort(matches.begin(), matches.end(),
```
CurtizJ: Shouldn't they already be sorted?

src/Storages/MergeTree/TextIndexPhraseSearch.cpp
```diff
@@ -0,0 +1,158 @@
...
+        else
+        {
+            ++rhs_idx;
+        }
+    }
+
+    if (matches.empty())
+        return result;
+
+    /// Sort by key and coalesce duplicate (doc_id, group) keys (OR their bitmaps).
```
CurtizJ: Could you explain how we can get duplicates here?

CurtizJ: As I understand, it's because of the boundary matches.

ahmadov: yes, it's due to the roarish bucket boundary.

src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
ahmadov: Yes, it would be similar to the posting list but I'd prefer to add that in a follow-up PR since it would require some large changes. I'd like to merge this simple PR to enable experimental positions then add improvement later.

src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
ahmadov: This was a left over from the initial implementation I had. At that time, it was slightly slower for tokens occur in many rows and positions, e.g "the". But I had some improvement since then. I will re-evaluate that again

src/Storages/MergeTree/MergeTreeReaderTextIndex.cpp
CurtizJ: Ok, let's implement this change later.

src/Storages/MergeTree/TextIndexPhraseSearch.cpp
ahmadov: good catch. yes, they should be sorted actually. It was a left-over.

src/Storages/MergeTree/TextIndexPhraseSearch.cpp
ahmadov: now, sorting is removed since they should be sorted already.

src/Storages/MergeTree/MergeTreeIndexText.cpp
ahmadov: it is done with 8ab18e505f35c6004fda840c0acf7d792c4ca739

CurtizJ (approved):
LGTM

Ergus:
This got in the queue just before the postprocessor (https://github.com/ClickHouse/ClickHouse/pull/98939) and now there are conflicts :(

clickgapai:
Hi @ahmadov — while reviewing this PR I found the following:

- **Coverage gap:** `positions` argument validation in `textIndexValidator` (experimental guard + range check) untested — https://github.com/ClickHouse/ClickHouse/pull/109697

PR #103182, opened 2026-04-20
Lightweight Updates v2

CurtizJ:

Lightweight `UPDATE` patch parts now use a new v2 on-disk format sorted by (`sorting_key..., _block_number, _block_offset`) and applied with a new merging algorithm. Peak memory is bounded by the largest equal-sort-key run instead of the full patch, and updates that cross merge boundaries no longer fall back to in-memory Join apply. Old-format patch parts remain readable. During a rolling upgrade from a version before 26.8, keep `patch_parts_version = 'v1'` or use the `compatibility` setting until all replicas are upgraded.


CurtizJ:
The first perf test results are controversial, but it's expected that a new algorithm may be slower than the current one in some cases.

Screenshot 2026-04-22 at 15 12 09
Screenshot 2026-04-22 at 15 12 49
Screenshot 2026-04-22 at 15 13 12

src/Storages/MergeTree/MergeTreeSettings.cpp
```diff
@@ -830,6 +830,14 @@ namespace ErrorCodes
     DECLARE(Bool, apply_patches_on_merge, true, R"(
     If true patch parts are applied on merges
     )", 0) \
+    DECLARE(Bool, enable_v2_lightweight_update_patches, true, R"(
```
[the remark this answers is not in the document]
CurtizJ: The setting has since been renamed to `patch_parts_version`. Rolling upgrades are handled via the `compatibility` setting: the settings-history entry is now `{previous: 'v1', new: 'v2'}` (in the 26.7 section), so an older compatibility target keeps replicas writing v1 patches until the whole cluster is upgraded. The setting description documents this. Done in 617572358f5.

src/Storages/MergeTree/PatchParts/SourcePartsSetForPatch.cpp
```diff
@@ -13,6 +19,57 @@ namespace ErrorCodes
...
+std::shared_ptr<const KeyDescription> buildSortingKeyPrefixDescription(const StorageMetadataPtr & metadata_snapshot, UInt64 prefix_size)
+{
+    const auto & main_sorting_key = metadata_snapshot->getSortingKey();
+
+    /// Get original expressions to preserve DESC entries for reverse flags.
+    const auto original_expr_list_ast = main_sorting_key.getOriginalExpressionList();
+    const auto * original_expr_list = original_expr_list_ast ? original_expr_list_ast->as<ASTExpressionList>() : nullptr;
+    const size_t main_children = original_expr_list ? original_expr_list->children.size() : 0;
+
+    if (prefix_size > main_children)
```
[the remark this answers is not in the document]
CurtizJ: Fixed in 617572358f5. `ALTER MODIFY ORDER BY` is now rejected at ALTER time (`MergeTreeData::checkSortingKeyChangeIsPossibleWithPatchParts`) while active v2 patch parts exist, unless the leading `prefix_size` expressions of the sorting key stay unchanged (extending the key is still allowed). The check also rejects replacing or inserting sort-key expressions inside the prefix — that case would not even throw at read time, it would silently build a wrong sort description and mis-prune. Added test `04506_lwu_modify_order_by_with_patches`. The `LOGICAL_ERROR` here stays as a last-resort invariant check.

src/Storages/MergeTree/PatchParts/RangesInPatchParts.cpp
```diff
@@ -132,6 +133,138 @@ MarkRanges getRangesInPatchPartJoin(const PatchPartInfoForReader & patch)
...
+        return {};
+
+    const size_t main_marks_count = original_part->index_granularity->getMarksCount();
+    auto main_index = original_part->getIndex();
+    auto patch_index = patch.part->getIndexPtr();
+
+    if (!main_index || main_index->empty() || !patch_index || patch_index->empty())
+        return emit_all_patch_ranges();
+
+    const auto & reverse_flags = patch.sorting_key->reverse_flags;
```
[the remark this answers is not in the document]
CurtizJ: Fixed in 617572358f5: `patch.sorting_key` is now checked before the first dereference, with the same `emit_all_patch_ranges()` fallback as the other defensive checks in this function.

src/Storages/MergeTree/PatchParts/MergeTreePatchReader.cpp
```diff
@@ -272,15 +274,190 @@ std::vector<PatchToApplyPtr> MergeTreePatchReaderJoin::applyPatch(const Block &
...
+    /// the patch mark by mark as the main side advances.
+    const auto & old = typeid_cast<const PatchMergeOnKeyReadResult &>(old_patch);
+
+    /// An empty patch block contributes nothing — always read the next mark if there is one.
+    if (old.block.rows() == 0)
+        return true;
+
+    const auto & sorting_key = *patch_part.sorting_key;
+    if (sorting_key.column_names.empty())
+        return true;  /// No key — always read next mark if one exists.
```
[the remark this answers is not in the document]
CurtizJ: Returning `false` here would lose updates: `readPatches` consumes ranges only while `needNewPatch` returns true, and with an empty sort key `needOldPatch` never evicts, so after the first call `last_read_patch` stays non-null and the remaining ranges would never be read — their updates would be silently dropped. For `ORDER BY tuple()` every patch row can match every main block, so the whole patch must be resident before the first apply; reading all ranges eagerly is required for correctness. The memory profile matches the v1 `Join` mode, which also materializes the whole patch. The comment above the branch was stale (it described the streaming behavior you quoted) — rewritten in 617572358f5.

src/Storages/MergeTree/PatchParts/PatchPartIndex.cpp
```diff
@@ -73,6 +130,37 @@ PatchParts SourcePartsSetForPatch::getPatchParts(const MergeTreePartInfo & origi
...
+                use_patch = true;
+                break;
+            }
+        }
+
+        if (use_patch)
+        {
+            patch_parts.push_back(PatchPartInfo
+            {
+                .mode = PatchMode::MergeOnKey,
```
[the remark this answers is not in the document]
CurtizJ: Fixed in 617572358f5.

src/Core/SettingsChangesHistory.cpp
```diff
@@ -1156,7 +1156,7 @@ const VersionToSettingsChangesMap & getMergeTreeSettingsChangesHistory()
     {
         addSettingsChanges(merge_tree_settings_changes_history, "26.5",
         {
-
+            {"patch_parts_version", "v2", "v2", "New setting to control the on-disk serialization version of patch parts produced by lightweight updates."},
```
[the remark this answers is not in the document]
CurtizJ: Fixed in 617572358f5: the entry is now `{"patch_parts_version", "v1", "v2", ...}` so `compatibility` can keep v1 writes during a rolling upgrade, and it moved to the 26.7 section (the PR will not ship in 26.5).

src/Storages/MergeTree/PatchParts/MergeTreePatchReader.cpp
```diff
@@ -272,15 +274,190 @@ std::vector<PatchToApplyPtr> MergeTreePatchReaderJoin::applyPatch(const Block &
...
+{
+    switch (patch_part.mode)
+    {
+        case PatchMode::Merge:
+            return std::make_unique<MergeTreePatchReaderMerge>(std::move(patch_part), std::move(reader));
+        case PatchMode::Join:
+            return std::make_unique<MergeTreePatchReaderJoin>(std::move(patch_part), std::move(reader), read_join_cache);
+        case PatchMode::MergeOnKey:
+            return std::make_unique<MergeTreePatchReaderMergeOnKey>(std::move(patch_part), std::move(reader));
+    }
```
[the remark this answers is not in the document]
CurtizJ: Fixed in 617572358f5 — added the post-switch `LOGICAL_ERROR` throw as suggested.

alexey-milovidov:
Your PR was poisoned by the test `02346_text_index_bug108519_qcc_skip_index`. It is fixed in master. Please update the branch.

src/Storages/MergeTree/PatchParts/SourcePartsSetForPatch.cpp
```diff
@@ -16,6 +20,14 @@ namespace ErrorCodes
     extern const int INCORRECT_DATA;
 }
 
+SourcePartsSetForPatch::SourcePartsSetForPatch(UInt8 format_version_, String sorting_key_desc_)
+    : format_version(format_version_)
+    , sorting_key_desc(std::move(sorting_key_desc_))
+{
+    if (format_version == V1_FORMAT_VERSION && !sorting_key_desc.empty())
+        throw Exception(ErrorCodes::INCORRECT_DATA, "Sorting key description must be empty for v1 patch part. Got: {}", sorting_key_desc);
```
alesapin:
```suggestion
        throw Exception(ErrorCodes::LOGICAL_ERROR, "Sorting key description must be empty for v1 patch part. Got: {}", sorting_key_desc);
```

src/Storages/MergeTree/IMergeTreeDataPart.h
```diff
@@ -742,8 +742,10 @@ class IMergeTreeDataPart : public std::enable_shared_from_this<IMergeTreeDataPar
 
     mutable std::map<String, std::shared_ptr<IMergeTreeDataPart>> projection_parts;
 
-    /// Set of source parts for patch parts. Empty for regular parts.
-    SourcePartsSetForPatch source_parts_set;
+    /// Set of source parts for patch parts, loaded from disk or set explicitly at part
+    /// creation. Disengaged for regular parts and for patch parts that are not loaded yet;
+    /// reading it too early would silently misinterpret the patch as v1.
+    std::optional<SourcePartsSetForPatch> source_parts_set;
```
alesapin: Source Part is well known term in ClickHouse -- parts for merge. Here I would prefer to have something like `AffectedPartsSetForPatch` or something like this.

src/Storages/MergeTree/PatchParts/MergeTreePatchReader.cpp
```diff
@@ -261,24 +256,151 @@ std::vector<PatchReadResultPtr> MergeTreePatchReaderJoin::readPatches(
...
+    const Names & sorting_key_names,
+    const std::vector<bool> & reverse_flags)
+{
+    /// Compares sort-key tuples at two positions: `main_block[main_row]` vs `patch_block[patch_row]`.
+    for (size_t i = 0; i < sorting_key_names.size(); ++i)
+    {
+        const auto & main_column = *main_block.getByName(sorting_key_names[i]).column;
+        const auto & patch_column = *patch_block.getByName(sorting_key_names[i]).column;
+
+        int cmp = main_column.compareAt(main_row, patch_row, patch_column, /*nan_direction_hint=*/ 1);
```
alesapin, line 339: Should we remove special staff like sparse or low cardinality here `removeSpecialRepresentations`?

src/Storages/MergeTree/PatchParts/applyPatches.cpp
```diff
@@ -351,301 +161,675 @@ void applyPatchesToBlockRaw(
...
+            group_it = merge_on_key_groups.emplace(merge_on_key_groups.end());
+            group_it->sorting_key = patch.sorting_key.get();
+        }
+
+        const auto & patch_data = typeid_cast<const PatchMergeOnKeyReadResult &>(*read_result);
+        group_it->blocks.emplace_back(&patch_data.block);
+        group_it->updated_columns.emplace_back(&updated_columns);
+    }
+
+    if (min_version.has_value())
```
alesapin: Should we do it before `applyPatchesToBlockLegacy(result_block, versions_block, patch_read_results, source_data_version);`?

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -10920,7 +10976,32 @@ QueryPipeline MergeTreeData::updateLightweightImpl(const MutationCommands & comm
...
     MutationCommands commands_to_run;
-    const auto & system_columns = getPatchPartSystemColumns();
+    auto system_columns = getPatchPartSystemColumns();
+    const auto metadata_snapshot = getInMemoryMetadataPtr(query_context, /*bypass_metadata_cache=*/ false);
+    const MergeTreePatchPartsVersion patch_parts_version = (*getSettings())[MergeTreeSetting::patch_parts_version];
+
+    /// For v2 patches, additionally read the source columns of the target table's sorting key expression.
+    if (patch_parts_version == MergeTreePatchPartsVersion::V2)
+    {
+        system_columns.erase(std::ranges::find(system_columns, "_part_offset", &NameAndTypePair::name));
```
alesapin: chassert that "_part_offset" exists in `system_columns`

src/Storages/MergeTree/MergeTreeData.h
```diff
@@ -1325,10 +1327,13 @@ class MergeTreeData : public WithMutableContext, public IStorage, public IBackgr
     constexpr static auto EMPTY_PART_TMP_PREFIX = "tmp_empty_";
     /// `metadata_snapshot` must come from the source part being covered
     /// (via `IMergeTreeDataPart::getMetadataSnapshot`) so patch parts get patch-part metadata.
+    /// For a part in a patch partition, `source_parts_set` must be seeded from a covered or
+    /// sibling part (see `SourcePartsSetForPatch::cloneEmpty`) to keep the partition uniform.
     std::pair<MergeTreeData::MutableDataPartPtr, scope_guard> createEmptyPart(
```
alesapin, line 1351: Why we may want to create empty patch part?)

src/Storages/MergeTree/MergeTreeReadersChain.cpp
```diff
@@ -552,22 +555,77 @@ void MergeTreeReadersChain::addPatchVirtuals(Block & to, const Block & from) con
...
+            continue;
+
+        /// A column that is present in the header but not filled yet counts as missing.
+        auto is_missing = [&](const String & name)
+        {
+            const auto * column = main_block.findByName(name);
+            return !column || !column->column;
+        };
+
+        if (patch.sorting_key->expression && std::ranges::any_of(patch.sorting_key->column_names, is_missing))
```
alesapin, line 658: Is it for a case of `ALTER MODIFY ORDER BY` only?

alesapin: Or it works for some early steps like PREWHERE?

src/Storages/MergeTree/MergeTreeData.h
```diff
@@ -1637,11 +1646,15 @@ class MergeTreeData : public WithMutableContext, public IStorage, public IBackgr
...
-    /// A cache for metadata snapshots for patch parts.
-    /// The key is a partition id of patch part.
+    /// Cached metadata used to read patch parts, by patch partition id.
     /// Patch parts in one partition always have the same structure.
     mutable std::mutex patch_parts_metadata_mutex;
-    mutable std::unordered_map<String, StorageMetadataPtr> patch_parts_metadata_cache;
+    mutable std::unordered_map<String, PatchPartMetadata> patch_parts_metadata_cache;
+
+    /// Cached effective sorting keys for applying v2 patch parts.
+    /// Invalidated on ALTER: the effective key depends on the table's current sorting key.
```
alesapin, line 1721: But why? MODIFY ORDER BY never do anything with data of parts. Is it because we can widen data type sin primary key?

src/Storages/MergeTree/MergeTreeDataWriter.cpp
```diff
@@ -684,7 +684,7 @@ MergeTreeTemporaryPartPtr MergeTreeDataWriter::writeTempPartImpl(
     BlockWithPartition & block_with_partition,
     StorageMetadataPtr metadata_snapshot,
     String partition_id,
-    SourcePartsSetForPatch source_parts_set,
+    std::optional<SourcePartsSetForPatch> source_parts_set,
```
alesapin: What is difference between empty `source_parts_set` and nullopt source parts set?

src/Storages/MergeTree/PatchParts/applyPatchesLegacy.h
```diff
@@ -0,0 +1,35 @@
...
+/// Helpers shared with the current format, defined in applyPatches.cpp.
+const PaddedPODArray<UInt64> & getColumnUInt64Data(const Block & block, const String & column_name);
+PaddedPODArray<UInt64> & getColumnUInt64Data(Block & block, const String & column_name);
+bool canApplyPatchInplace(const IColumn & column);
+IColumn::Versions & addDataVersionForColumn(Block & block, const String & column_name, UInt64 num_rows, UInt64 data_version);
+Block getUpdatedHeader(const PatchesIndices & patches);
+
+/// Applies each patch as-is, without combining row indices across patches.
+/// Patches may have multiple source blocks (e.g. built by applyPatchesMergeOnKey).
+void applyPatchesIndices(
```
alesapin: Why is it defined in legacy header, but implementation is in non-legacy C++?

src/Storages/MergeTree/PatchParts/applyPatches.cpp
```diff
@@ -351,301 +161,675 @@ void applyPatchesToBlockRaw(
...
+    absl::flat_hash_map<UInt128, UInt32, UInt128TrivialHash> run_map;
+    PaddedPODArray<RunEntry> run_entries;
+};
 
-    patch_to_apply->result_row_indices.reserve(size_to_reserve);
-    patch_to_apply->patch_row_indices.reserve(size_to_reserve);
+/// Processes one run of equal sort keys: matches result rows [result_cursor.row, result_cursor.run_end)
+/// with rows [cursor.row, cursor.run_end) of cursors in `equal_cursors` by the
+/// (block_number, block_offset) identity and emits matches into the groups' patches.
+void processEqualRun(
```
alesapin: maybe something like `processEqualKeyCursors?

src/Storages/MergeTree/PatchParts/PatchPartsUtils.cpp
```diff
@@ -52,29 +70,34 @@ static void addCodecsForPatchSystemColumns(ColumnsDescription & columns_desc)
...
-    });
+        columns_desc.modify("_part_offset", [&](auto & column_desc)
+        {
+            column_desc.codec = BlockOffsetColumn::codec;
+        });
+    }
 }
 
-StorageMetadataPtr getPatchPartMetadata(Block sample_block, ContextPtr local_context)
+static void addColumnIfNotExists(ColumnsDescription & columns_desc, const String & name, const DataTypePtr & type)
```
alesapin: maybe it can be ColumnsDescription method?)

src/Storages/MergeTree/PatchParts/RangesInPatchParts.cpp
```diff
@@ -133,6 +155,115 @@ MarkRanges getRangesInPatchPartJoin(const PatchPartInfoForReader & patch)
...
+    {
+        main_sorting_key_columns[i] = (*main_index)[i].get();
+        patch_sorting_key_columns[i] = (*patch_index)[i].get();
+    }
+
+    auto compare_patch = [&](size_t patch_row, size_t main_row) -> int
+    {
+        for (size_t i = 0; i < common_prefix_size; ++i)
+        {
+            int cmp = patch_sorting_key_columns[i]->compareAt(patch_row, main_row, *main_sorting_key_columns[i], /*nan_direction_hint=*/ 1);
```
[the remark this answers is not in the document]
alesapin: Only widening of types is allowed + LowCardinality and similar staff. Does it make sense to introduce more strict checks here as  bot suggests?

src/Storages/MergeTree/MergeTreeSettings.cpp
```diff
@@ -869,6 +869,22 @@ Minimal amount of data parts which merge selector can pick to merge at once
 )", 0) \
     DECLARE(Bool, apply_patches_on_merge, true, R"(
 If true patch parts are applied on merges
+)", 0) \
+    DECLARE(MergeTreePatchPartsVersion, patch_parts_version, "v2", R"(
```
alesapin, line 872: Before merge let's check that all the tests pass with v1 as well.

alesapin (commented):
Big complex change, but semantically should be better than previous version.

src/Storages/MergeTree/IMergeTreeDataPart.h
CurtizJ: Yes, it's a bad name. I want to rename it to `PatchPartIndex`.

src/Storages/MergeTree/PatchParts/MergeTreePatchReader.cpp
CurtizJ: Likely, yes.

src/Storages/MergeTree/PatchParts/applyPatches.cpp
CurtizJ: resolved in ce3be758bfdf3c997fbfa26f295968c671da4c58

src/Storages/MergeTree/PatchParts/PatchPartsUtils.cpp
CurtizJ: resolved in ce3be758bfdf3c997fbfa26f295968c671da4c58

src/Storages/MergeTree/PatchParts/applyPatches.cpp
CurtizJ: resolved in ce3be758bfdf3c997fbfa26f295968c671da4c58

src/Storages/MergeTree/MergeTreeData.cpp
CurtizJ: Resolved in 6d2239f.

src/Storages/MergeTree/MergeTreeData.h
CurtizJ: It can be created on `TRUNCATE` or `DETACH PARTITION`. This case was fixed in #104353. With incorrect metadata, an empty patch part may lead to marking it broken.

src/Storages/MergeTree/MergeTreeData.h
CurtizJ: We can also truncate the `ORDER BY ` key. We store the longest common prefix of the patch's sorting key and the current sorting key of a table. So, it becomes invalid with this modification. For simplicity, I drop the cache on each alter to avoid detection of `ORDER BY` modifications.

CurtizJ: Also, it looks like the data types of sorting key columns may change.

src/Storages/MergeTree/IMergeTreeDataPart.h
CurtizJ: Resolved in 0eff0c7

src/Storages/MergeTree/MergeTreeDataWriter.cpp
CurtizJ: The empty `SourcePartsSetForPatch` contains a format version and sorting key. I've slightly improved creation `SourcePartsSetForPatch`: removed default constructor and made `format_version` parameter mandatory.

src/Storages/MergeTree/PatchParts/MergeTreePatchReader.cpp
CurtizJ: Resolved in efa1bfb.

src/Storages/MergeTree/PatchParts/applyPatchesLegacy.h
CurtizJ: Resolved in 64a5dec.

src/Storages/MergeTree/PatchParts/RangesInPatchParts.cpp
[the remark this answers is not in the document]
CurtizJ: Resolved in 98bce34.

src/Storages/MergeTree/MergeTreeReadersChain.cpp
```diff
@@ -552,22 +554,77 @@ void MergeTreeReadersChain::addPatchVirtuals(Block & to, const Block & from) con
...
+        {
+            /// Execute on a scratch copy: `ExpressionActions::execute` drops the consumed input
+            /// columns from the block, and another patch's sorting key may still need them.
+            auto block_for_key = result_header.cloneWithColumns(read_result.columns);
```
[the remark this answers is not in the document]
CurtizJ: Resolved in 98bce34.

src/Storages/MergeTree/PatchParts/PatchPartsUtils.cpp
```diff
@@ -106,39 +123,164 @@ StorageMetadataPtr getPatchPartMetadata(ColumnsDescription patch_part_desc, Cont
...
+    patch_part_desc.addIfNotExists(ColumnDescription("_part", std::make_shared<DataTypeLowCardinality>(std::make_shared<DataTypeString>())));
+    patch_part_desc.addIfNotExists(ColumnDescription(BlockNumberColumn::name, BlockNumberColumn::type));
+    patch_part_desc.addIfNotExists(ColumnDescription(BlockOffsetColumn::name, BlockOffsetColumn::type));
+    patch_part_desc.addIfNotExists(ColumnDescription(PartDataVersionColumn::name, PartDataVersionColumn::type));
+
+    /// Partition id: `__patchPartitionID(_part, hash(...))`.
+    auto part_identifier = make_intrusive<ASTIdentifier>("_part");
+    const auto sorting_key_expr_list = sorting_key.getOriginalExpressionList();
+
+    Names names_for_hash = patch_part_desc.getNamesOfPhysical();
```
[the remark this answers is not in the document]
CurtizJ: Resolved in 98bce34.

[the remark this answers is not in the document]
CurtizJ: Resolved in 4ad17d3.

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -1309,8 +1313,14 @@ void MergeTreeData::setProperties(
 
     setInMemoryMetadata(new_metadata);
```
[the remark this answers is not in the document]
CurtizJ: Resolved in d6b52e4.

src/Storages/MergeTree/MergeTreeReadersChain.cpp
```diff
@@ -545,29 +552,93 @@ void MergeTreeReadersChain::addPatchVirtuals(ReadResult & result, const Block &
...
+            patch.sorting_key->expression->execute(block_for_key);
+
+            for (const auto & name : patch.sorting_key->column_names)
+            {
+                if (!main_block.has(name))
+                    main_block.insert(block_for_key.getByName(name));
```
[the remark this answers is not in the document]
CurtizJ: Resolved in 72d4b5c.

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -11434,26 +11450,65 @@ AlterConversionsPtr MergeTreeData::getAlterConversionsForPart(
...
     }
 
     return std::make_shared<AlterConversions>(commands, patches_for_reader, query_context);
 }
 
-StorageMetadataPtr MergeTreeData::getPatchPartMetadata(const ColumnsDescription & patch_part_desc, const String & patch_partition_id, ContextPtr local_context) const
+PatchPartMetadata MergeTreeData::getPatchPartMetadata(const IMergeTreeDataPart & patch_part, ContextPtr local_context) const
 {
     std::lock_guard lock(patch_parts_metadata_mutex);
+    const auto & patch_partition_id = patch_part.info.getPartitionId();
```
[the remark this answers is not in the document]
CurtizJ: Resolved in 6353ab9.

src/Storages/MergeTree/MergeTreeSettings.cpp
CurtizJ: I've added setting `patch_parts_version` for randomization, and adapted existing tests to work with both versions.

src/Storages/MergeTree/MergeTreeBlockReadUtils.cpp
```diff
@@ -448,11 +448,11 @@ void addPatchPartsColumns(
...
     auto first_step_columns_set = first_step_columns.getNameSet();
 
-    for (const auto & virtual_name : required_virtuals)
+    for (const auto & key_column_name : required_key_columns)
     {
-        if (!first_step_columns_set.contains(virtual_name))
+        if (!first_step_columns_set.contains(key_column_name))
         {
-            auto column = storage_snapshot->getColumn(options, virtual_name);
+            auto column = storage_snapshot->getColumn(options, key_column_name);
```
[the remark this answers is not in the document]
CurtizJ:
```
b UInt64 DEFAULT d + 1 via MODIFY ORDER BY (a, b)
```

But this is just an invalid sorting key, because the default expression doesn't guarantee the order.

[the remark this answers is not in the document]
CurtizJ: Yes, I checked that it's impossible: an exception will be thrown:
```sql
ALTER TABLE kek
    (ADD COLUMN `b` UInt64 DEFAULT c),
    (MODIFY ORDER BY (a, b))

Query id: 6aa47ba3-73cc-4b92-9006-f5c749caa95f


Elapsed: 0.028 sec.

Received exception from server (version 26.8.1):
Code: 36. DB::Exception: Received from localhost:9000. DB::Exception: Newly added column b has a default expression, so adding expressions that use it to the sorting key is forbidden. (BAD_ARGUMENTS)
```

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -11421,26 +11437,67 @@ AlterConversionsPtr MergeTreeData::getAlterConversionsForPart(
...
     }
 
     return std::make_shared<AlterConversions>(commands, patches_for_reader, query_context);
 }
 
-StorageMetadataPtr MergeTreeData::getPatchPartMetadata(const ColumnsDescription & patch_part_desc, const String & patch_partition_id, ContextPtr local_context) const
+PatchPartMetadata MergeTreeData::getPatchPartMetadata(const IMergeTreeDataPart & patch_part, ContextPtr local_context) const
 {
+    /// This metadata depends only on the structure of the patch part.
+    auto cache_key = getStructureHashOfPatch(patch_part.info.getPartitionId());
```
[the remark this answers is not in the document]
CurtizJ: Resolved in b8593a8.

src/Storages/MergeTree/PatchParts/applyPatches.cpp
```diff
@@ -351,301 +162,675 @@ void applyPatchesToBlockRaw(
...
+        if (patch.mode != PatchMode::MergeOnKey)
+            continue;
 
-    if (canApplyPatchesRaw(patches))
-        applyPatchesToBlockRaw(result_block, versions_block, patches, updated_header, source_data_version);
-    else
-        applyPatchesToBlockCombined(result_block, versions_block, patches, updated_header, source_data_version);
+        auto group_it = std::ranges::find_if(merge_on_key_groups, [&](const auto & group)
+        {
+            return group.sorting_key == patch.sorting_key.get();
```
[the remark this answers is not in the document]
CurtizJ: Resolved in 891c690.

src/Storages/MergeTree/AlterConversions.cpp
```diff
@@ -223,9 +225,13 @@ void AlterConversions::addMutationCommand(const MutationCommand & command, const
 
 void AlterConversions::addPatchPart(PatchPartInfoForReader patch_part)
 {
+    /// Sort-key columns are stored in v2 patch parts to identify updated rows.
+    /// They are never updated themselves, so they must not be reported as updated columns.
+    auto sorting_key_columns = getSortingKeyColumnsInPatch(patch_part);
```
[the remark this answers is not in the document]
CurtizJ: Resolved in 891c690.

src/Storages/MergeTree/PatchParts/applyPatchesLegacy.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+    return true;
+}
+
+/// Identifies a set of updated columns in the map of patches to apply.
+UInt128 getNamesHash(const Names & names)
+{
+    SipHash hash;
+    for (const auto & name : names)
+        hash.update(name);
```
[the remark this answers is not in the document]
CurtizJ: Resolved in e3da0a3.

src/Storages/MergeTree/PatchParts/applyPatches.cpp
```diff
@@ -351,301 +162,677 @@ void applyPatchesToBlockRaw(
...
-            if (iterators.it != iterators.end && iterators.it->first == result_block_offset[row])
+            if (column && column->column)
             {
-                const auto & [patch_block_index, patch_row_index] = iterators.it->second;
-
-                patch_to_apply->result_row_indices.push_back(row);
-                patch_to_apply->patch_block_indices.push_back(patch_block_index);
-                patch_to_apply->patch_row_indices.push_back(patch_row_index);
+                hash.update(name);
+                hash.update(column->type->getName());
```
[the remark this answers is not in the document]
CurtizJ: Resolved in e3da0a3.

src/Storages/MergeTree/PatchParts/PatchPartsUtils.cpp
```diff
@@ -106,39 +140,177 @@ StorageMetadataPtr getPatchPartMetadata(ColumnsDescription patch_part_desc, Cont
...
+
+    /// Partition id: `__patchPartitionID(_part, hash(...))`.
+    auto part_identifier = make_intrusive<ASTIdentifier>("_part");
+    const auto sorting_key_expr_list = sorting_key.getOriginalExpressionList();
+
+    /// Include column types so that patches with the same column names but different
+    /// types go to different partitions: one patch partition must have one schema.
+    auto names_for_hash = getColumnNamesWithTypes(patch_part_desc);
+    names_for_hash.emplace_back(sorting_key_expr_list ? sorting_key_expr_list->formatWithSecretsOneLine() : "");
+    auto columns_hash = getColumnsHash(std::move(names_for_hash));
```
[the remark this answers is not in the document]
CurtizJ: Resolved in e3da0a3.

CurtizJ:
@alesapin 

I think all reviews are resolved (at least until the next commit and any new slop from the reviewer bot).

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -11582,16 +11640,48 @@ std::expected<void, PreformattedMessage> MergeTreeData::supportsLightweightUpdat
...
 
+    NamesAndTypesList system_columns;
     MutationCommands commands_to_run;
-    const auto & system_columns = getPatchPartSystemColumns();
+    const auto metadata_snapshot = getInMemoryMetadataPtr(query_context, /*bypass_metadata_cache=*/ false);
+    const MergeTreePatchPartsVersion patch_parts_version = (*getSettings())[MergeTreeSetting::patch_parts_version];
+
+    switch (patch_parts_version)
+    {
+        case MergeTreePatchPartsVersion::V1:
```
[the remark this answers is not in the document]
CurtizJ: This case is impossible.

alesapin (approved):
LGTM

CurtizJ:
Fuzzer failure (`Logical error: Context has expired`) is being fixed in #114769.

alexey-milovidov:
> [AST fuzzer (amd_debug, targeted)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=103182&sha=de1a9de0ed6002f2e176c38531f05de209af3112&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%2C%20targeted%29)		FAIL		
Logical error: Context has expired (STID: 1805-460a)	FAIL	[cidb](https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICdMb2dpY2FsIGVycm9yOiBDb250ZXh0IGhhcyBleHBpcmVkIChTVElEOiAxODA1LTQ2MGEpJwogICAgLS0gQU5EIGNoZWNrX25hbWUgPSAnQVNUIGZ1enplciAoYW1kX2RlYnVnLCB0YXJnZXRlZCknCiAgICBBTkQgdGVzdF9zdGF0dXMgSU4gKCdGQUlMJywgJ0VSUk9SJykKICAgIEFORCAocHVsbF9yZXF1ZXN0X251bWJlciA9IDAgT1IgYmFzZV9yZWYgSU4gKCdtYXN0ZXInKSkKICAgIEFORCB0ZXN0X2NvbnRleHRfcmF3IExJS0UgJyVFeGNlcHRpb246JScKR1JPVVAgQlkgZGF5Ck9SREVSIEJZIGRheSBERVNDCg==)	IGNORED

Oh, shit... Reverting.

alexey-milovidov:
This has been reverted in https://github.com/ClickHouse/ClickHouse/pull/116005, because it makes any lightweight `UPDATE` of a table whose sorting key has a direction throw a `LOGICAL_ERROR`:

```
Logical error: The size of reverse_flags (A) does not match the size of KeyDescription B  (STID: 3072-3c97)
```

`InterpreterUpdateQuery::execute` -> `MergeTreeData::updateLightweightImpl` -> `getPatchPartMetadataV2` -> `KeyDescription::getKeyFromAST`.

Since the first occurrence on 2026-08-21 18:02 it has been reported against **29 distinct pull requests**, across `BuzzHouse`, `AST fuzzer` and `Stress test`, on both architectures and in several sanitizer builds. An example report: [Stress test (arm_debug)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=115540&sha=cac893754a4b682e685374ff97dd36da68f59656&name_0=PR&name_1=Stress%20test%20%28arm_debug%29), on an unrelated Web UI pull request.

### Cause

`getPatchPartMetadataV2` builds the patch part's sorting key by cloning the table's original sorting-key AST children and then appending two bare identifiers, in `src/Storages/MergeTree/PatchParts/PatchPartsUtils.cpp`:

```cpp
for (const auto & child : sorting_key_expr_list->children)
    order_by_expression->arguments->children.push_back(child->clone());

order_by_expression->arguments->children.push_back(make_intrusive<ASTIdentifier>(BlockNumberColumn::name));
order_by_expression->arguments->children.push_back(make_intrusive<ASTIdentifier>(BlockOffsetColumn::name));

part_metadata.sorting_key = KeyDescription::getKeyFromAST(
    order_by_expression, patch_part_desc, /*virtuals=*/ {}, local_context,
    /*additional_columns=*/ {{BlockNumberColumn::name, ...}, {BlockOffsetColumn::name, ...}});
```

`_block_number` and `_block_offset` are supplied twice: once inline in the tuple and once as `additional_columns`. Only the second route maintains `reverse_flags`. In `buildKeyColumns` (`src/Storages/KeyDescription.cpp`):

- the `N` cloned children are `ASTStorageOrderByElement`s, so they contribute `N` entries to both `reverse_flags` and the expression list;
- the two appended identifiers are not order-by elements, so they contribute `2` expression-list entries and no `reverse_flags` entries;
- the `additional_columns` loop, which is what appends the matching `false` flags, skips both names, because they are already in `column_names` from the tuple.

So `reverse_flags.size()` is `N` against an expression list of `N + 2`, which trips the consistency check in `KeyDescription::getKeyFromAST`.

An ascending sorting key yields empty `reverse_flags` and short-circuits that check's `!reverse_flags.empty()` guard, which is why this is intermittent and has so far only been found by fuzzers rather than by the tests here.

The check itself and direction-carrying key descriptions (#111059, merged 2026-07-24) both predate this pull request and were correct together for a month. Before this change the patch-part sorting key was built from bare `ASTIdentifier`s, so `reverse_flags` was always empty and the check was unreachable — confirmed by diffing `70a0c0ff637^1` against `70a0c0ff637`.

### Suggested fix for the re-land

Dropping the two inline `push_back` calls and letting `additional_columns` add `_block_number` and `_block_offset` looks sufficient: that path appends the two matching `false` flags when `reverse_flags` is non-empty, and adds nothing when it is empty, so both the directional and the ascending case come out consistent. Worth a regression test for a lightweight `UPDATE` on a table with `ORDER BY ... DESC`, and one with a mixed-direction key.

Sorry for the disruption — the revert is only to take the failure off master, and is not a judgement on the feature.

Issue #99475, opened 2026-03-13
Prometheus HTTP handler endpoints do not increment rows/bytes metrics during inserts or selects, and do not send queries to system.query_log.


JTCunning:
### Company or project name

N/A

### Describe what's wrong

Prometheus HTTP handler endpoints do not create (or do not reliably create) entries in `system.query_log` with the expected row/byte metrics:
- **Remote write** (`/write`): Requests that insert time series data do not produce a `QueryFinish` row with `written_rows` or `written_bytes`.
- **Remote read** (`/read`): Requests that read time series do not produce a `QueryFinish` row with `read_rows` or `read_bytes`.
- **Query API** (`/api/v1/query`): Requests do not reliably appear in `system.query_log` with `read_rows > 0` and `read_bytes > 0` in a way that integration tests can assert (count-based). So either they are not logged, or the logged entries do not have read metrics set as expected.
- **Query range API** (`/api/v1/query_range`): Same as Query API — integration tests that expect one additional `QueryFinish` row with `read_rows > 0` and `read_bytes > 0` after a request fail.
Additionally, passing `query_id` as a URL parameter to the Query API is rejected (setting is not supported), which makes it harder to correlate specific requests with log entries even when logging exists.

### Does it reproduce on the most recent release?

Yes

### How to reproduce

1. Configure Prometheus HTTP handlers (remote write, remote read, Query API, query_range) and enable the query log.
2. Create a TimeSeries table and send a remote write with at least one time series.
3. Run: `SYSTEM FLUSH LOGS query_log; SELECT count() FROM system.query_log WHERE type = 'QueryFinish' AND written_rows > 0;`
  The count does not increase for the remote write.

4. Send a remote read request (e.g. protobuf ReadRequest to the read endpoint). Check the same style of count for `read_rows > 0`, it does not increase for the remote read.
5. Call the Query API, e.g. `GET /api/v1/query?query=up&time=<timestamp>`. After` SYSTEM FLUSH LOGS query_log`, the count of rows with `type = 'QueryFinish' AND read_rows > 0 AND read_bytes > 0` does not increase by one as expected.
6. Call the query_range API, e.g. `GET /api/v1/query_range?query=up&start=...&end=...&step=15`. Same count check fails.

Integration tests in https://github.com/ClickHouse/ClickHouse/compare/master...JTCunning:ClickHouse:test/prometheus-query-log-integration-test reproduce these failures (remote write, remote read, Query API, query_range).

[...]

PR #103189, opened 2026-04-20
feat(promql): Mount Prometheus protocols on the main HTTP port.

### LINKED ISSUE #93831: Prometheus remote-write: support URL-based table routing for TimeSeries table engine
### Company or project name

_No response_

#### Problem

ClickHouse’s Prometheus `remote-write` integration currently requires one `<prometheus>``<handlers>` entry per `TimeSeries` table, with the target table statically defined in config.xml. This design does not scale well in real-world deployments.

Prometheus itself has no concept of tables—it only pushes samples to an HTTP endpoint. As a result, any table-level routing must be derived from the request URL or HTTP headers, not from Prometheus-side configuration.

As of ClickHouse 25.12, the official documentation describes the following setup [1]:

```xml
<prometheus>
    <port>9363</port>
    <handlers>
        <my_rule_1>
            <url>/write</url>
            <handler>
                <type>remote_write</type>
                <database>db_name</database>
                <table>time_series_table</table>
            </handler>
        </my_rule_1>
    </handlers>
</prometheus>
```

With this approach, users are forced to:

1. Create the target `time_series_table` in advance
2. Hard-code the table name in `config.xml`
3. Reload ClickHouse configuration whenever:
  * a new TimeSeries table needs to accept `remote-write`, or
  * an existing table should stop accepting `remote-write`

This operational model becomes increasingly brittle and difficult to manage as the number of TimeSeries tables grows.

#### Request
* Support URL-based table routing for Prometheus remote-write, for example: `/<database>/<table>/write`
* Preserve backward compatibility: when URL-based routing is not enabled, the existing fixed `<table>` behavior in `config.xml` remains unchanged.

This enhancement would allow a single remote-write handler to serve multiple TimeSeries tables, significantly reducing configuration complexity and improving operational scalability.

#### References
[1] https://clickhouse.com/docs/interfaces/prometheus#remote-write

25.12

### How to reproduce

Follow the https://clickhouse.com/docs/interfaces/prometheus#remote-write to configure Prometheus `remote-write` protocol for some TimeSeries table

_No response_

### Additional context

_No response_

JTCunning:
Open questions/considerations:
1. my default http prefix i chose was `/prometheus` because that's a pretty industry standard path. if we want to change it to `/prom` or whatever, please do let me know.
2. I have this gated behind a setting so as to not mess with registering the http endpoints unless explicitly suggested to do so. Let me know if it should be enabled by default for any reason.
3. I thought about migrating the metrics collection endpoint alongside the prom endpoints, but decided against it since that would be a pretty large infrastructural change. The only thing they have in common is that they both are _for_ the prometheus ecosystem, but they have wildly different usages.
4. This fixes https://github.com/ClickHouse/ClickHouse/issues/93831 and obsoletes https://github.com/ClickHouse/ClickHouse/pull/93833, because it is a larger overhaul, but I want to thank @niyue for the idea and the isolated implementation! ❤️
5. https://github.com/ClickHouse/ClickHouse/issues/99475 remains open, but this incidentally passes `query_id` through now.

---
Folds the Prometheus remote_write, remote_read, and Query API protocols into the main HTTP port under a configurable URL prefix (default `/prometheus`). Protocol paths are `/<prefix>/write`, `/<prefix>/read`, and `/<prefix>/api/v1/...` (for example `POST /prometheus/write?database=mydb&table=metrics`). The target TimeSeries table is chosen via `database` and `table` query parameters, or via `X-ClickHouse-Database` and `X-ClickHouse-Table` HTTP headers when those parameters are missing (same fallback pattern as `database` elsewhere in the HTTP handler). Path-segment routing `/<prefix>/<database>/<table>/...` is not used. The dedicated `<prometheus><port>` listener is kept for back-compat (with expose_metrics still served exclusively there) but logs a one-shot deprecation warning at startup.

- Add `<prometheus><http_path_prefix>` config key; auto-mount the three protocol handlers from `addDefaultHandlersFactory` ahead of the catch-all DynamicQueryHandler so prefix routes win.
- Extend `PrometheusRequestHandlerConfig` with `http_path_prefix` and `enable_table_name_url_routing`; resolve the (database, table) pair from query parameters / headers in `PrometheusRequestHandler` and dispatch the protocol-relative path.
- Reject expose_metrics for HTTP-rule handlers; reject fixed-table keys when dynamic routing is in use.
- Add `prometheus_url_routing_enabled` TimeSeries setting (default true) so table owners can opt a particular table out of URL-routed access; on opt-out or non-TimeSeries storages the handler returns HTTP 403. Map ACCESS_DENIED to HTTP_FORBIDDEN.
- Treat Prometheus query parameters (`match[]`, `limit`, `timeout`, `lookback_delta`) as reserved so DynamicQueryHandler-style settings parsing does not reject them; reserve `database` and `table` so they are not applied as ClickHouse settings.
- Document the new prefix mount, deprecation, param/header routing, and per-table opt-out in the server settings and TimeSeries docs.
- Add `test_prometheus_protocols/test_http_port.py` covering the default and custom prefix, the per-table opt-out (HTTP 403), non-TimeSeries rejection, malformed paths (404), missing database/table (400), header-based routing, expose_metrics not served under the prefix, and that the dedicated port still works and emits the deprecation warning.

Made-with: Cursor

- Experimental Feature

- Add dynamic routing to Prometheus/TimeSeries tables via the default HTTP port.


vitlibar:
I understand that we need to pass database and table via URL or maybe headers, but what is wrong with dedicated prometheus port? In ClickHouse we have dedicated ports for different protocols. Also the main HTTP port serves ClickHouse HTTP requests (for example `http://localhost:8123/?query=SELECT%201`) which is a completely different thing.

Also why is this format `<url>regex:^/foo/[^/]+/[^/]+/write$</url>`? It's strange and doesn't look good.

Let's maybe discuss in the issue https://github.com/ClickHouse/ClickHouse/issues/93831

JTCunning:
Breaking this up into smaller pieces.

PR #103208, opened 2026-04-20
Fix linux tag not being respected in clickhouse-test on Darwin

### LINKED ISSUE #103206: Test 04061_http_pool_tcp_buffer_async_metrics fails on macOS (arm_darwin) in Fast Test
## Description

The test `04061_http_pool_tcp_buffer_async_metrics` fails on macOS (arm_darwin) in the Fast Test job.

The test is tagged with `# Tags: linux` (line 2), which should prevent it from running on non-Linux platforms. However, when the test is explicitly passed as a positional argument to `clickhouse-test`, the tag check is bypassed and the test runs on macOS anyway.

On macOS, the `HTTPConnectionPoolStorageTCPRcvBuf*` and `HTTPConnectionPoolStorageTCPSndBuf*` asynchronous metrics are not populated (likely because the underlying TCP socket buffer size retrieval is Linux-specific), so the test produces empty output instead of the expected 10 metrics.

## Failure

```
04061_http_pool_tcp_buffer_async_metrics:  [ FAIL ] 5.76 sec.
Reason: result differs with reference:
--- tests/queries/0_stateless/04061_http_pool_tcp_buffer_async_metrics.reference
+++ 04061_http_pool_tcp_buffer_async_metrics.86071.stdout
@@ -1,10 +0,0 @@
-HTTPConnectionPoolStorageTCPRcvBufBytes_p50	1
-HTTPConnectionPoolStorageTCPRcvBufBytes_p75	1
-HTTPConnectionPoolStorageTCPRcvBufBytes_p90	1
-HTTPConnectionPoolStorageTCPRcvBufBytes_p95	1
-HTTPConnectionPoolStorageTCPRcvBufTotalBytes	1
-HTTPConnectionPoolStorageTCPSndBufBytes_p50	1
-HTTPConnectionPoolStorageTCPSndBufBytes_p75	1
-HTTPConnectionPoolStorageTCPSndBufBytes_p90	1
-HTTPConnectionPoolStorageTCPSndBufBytes_p95	1
-HTTPConnectionPoolStorageTCPSndBufTotalBytes	1
```

## Environment

- Platform: macOS arm64 (arm_darwin)
- ClickHouse version: 26.4.1.1052 (master, git hash `794c01e727e47fce3cd730dac6f781e2b7e7c7d8`)
- Job: Fast test (arm_darwin)

## Root Cause

The test has `# Tags: linux` which should skip it on macOS, but `clickhouse-test` does not check platform tags when a test is explicitly passed as a positional argument (after `--`).

## Fix

Either:
1. Fix `clickhouse-test` to respect platform tags even when the test is explicitly specified, or
2. Add a platform check inside the test itself (e.g. guard with `[ "$(uname)" == "Linux" ] || exit 0`).

leshikus:
The `linux` tag on a test is supposed to restrict it to Linux-only runs, but `clickhouse-test` did not check for it. Only the `no-darwin` tag triggered a skip on macOS. As a result, tests like `04061_http_pool_tcp_buffer_async_metrics` (which uses Linux-specific netlink socket diagnostics to collect TCP buffer async metrics) were being executed on macOS in the Fast Test (arm_darwin) job and failing.

The fix extends the existing `no-darwin` guard to also trigger when the `linux` tag is present, making both tags equivalent for Darwin.

Closes #103206

### Version info
- Merged into: `26.5.1.21`

### DIFF
--- a/tests/clickhouse-test
+++ b/tests/clickhouse-test
@@ -1143,6 +1143,7 @@ class FailureReason(enum.Enum):
     NO_OPENSSL_FIPS = "no-openssl-fips"
     ENCRYPTED_STORAGE = "encrypted-storage"
     NOT_SUPPORTED_ON_DARWIN = "not supported on Darwin"
+    LINUX_ONLY = "supported only on Linux"
 
     # UNKNOWN reasons
     NO_REFERENCE = "no reference file"
@@ -2403,10 +2404,11 @@ class TestCase:
         if "no-async-insert" in tags and args.no_async_insert:
             return FailureReason.NO_ASYNC_INSERT
 
-        if (
-            "no-darwin" in tags or "linux" in tags
-        ) and platform.system() == "Darwin":
+        if "no-darwin" in tags and platform.system() == "Darwin":
             return FailureReason.NOT_SUPPORTED_ON_DARWIN
+
+        if "linux" in tags and platform.system() != "Linux":
+            return FailureReason.LINUX_ONLY
 
         # Do not run *Log and Memory engine-specific tests
         if (
--- a/tests/clickhouse-test
+++ b/tests/clickhouse-test
@@ -2403,7 +2403,9 @@ class TestCase:
         if "no-async-insert" in tags and args.no_async_insert:
             return FailureReason.NO_ASYNC_INSERT
 
-        if "no-darwin" in tags and platform.system() == "Darwin":
+        if (
+            "no-darwin" in tags or "linux" in tags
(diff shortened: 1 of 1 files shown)


leshikus:
the failures are unrelated

leshikus:
failures are not related

maxknv (approved):
ok, but I'd just remove since it's not used

leshikus:
@maxknv it is used in `04061_http_pool_tcp_buffer_async_metrics`

PR #97831, opened 2026-02-24
Fix LOGICAL_ERROR exceptions from recursive `convertToFullIfNeeded` with `LowCardinality` inside compound types

Algunenano:
Follow-up to #97654 which fixed the same bug only in `concat` but missed other callers.

`convertToFullIfNeeded` (made recursive in #97493) strips `LowCardinality` from inside compound column types like `Variant`, `Dynamic`, and `Tuple`, but the corresponding types are not updated to match. This creates column/type mismatches that trigger LOGICAL_ERROR exceptions in debug/sanitizer builds.

## Fixes

**1. `concatWithSeparator` and `format` functions** — same as `concat` fix in #97654: use only top-level column conversions (Const, Sparse, LowCardinality) instead of the recursive `convertToFullIfNeeded`, since the type is not updated.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=97431&sha=6fe0e8c44dd5c6a12583390b4f8fc45e226a3cb4&name_0=PR&name_1=Stateless%20tests%20%28arm_asan%2C%20azure%2C%20parallel%29
https://github.com/ClickHouse/ClickHouse/pull/97431

**2. `Set` (affects `IN` subqueries with `LowCardinality` inside `Tuple`)** — `Set::insertFromColumns` calls `convertToFullIfNeeded` which recursively strips LC from inner subcolumns, but `Set::setHeader` only stripped top-level LC from `set_elements_types`. When `KeyCondition::tryPrepareSetColumnsForIndex` unpacked the tuple to build index conditions, `castColumn` tried to interpret a `ColumnVector<Int8>` as `ColumnLowCardinality`, causing the exception.

Fix: use `recursiveRemoveLowCardinality` for `set_elements_types` in both `Set::setHeader` and the static `Set::getElementTypes`, matching what `convertToFullIfNeeded` does to columns.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=97831&sha=6f56284303c16ecc5f5a437491c661d5019f2977&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%29
https://github.com/ClickHouse/ClickHouse/pull/97831

[...]

PR #103211, opened 2026-04-21
Fix crash in partition pruning with nested LowCardinality constant

groeneai:
Fixes a `LOGICAL_ERROR` server exception during MergeTree index analysis when a constant in a `WHERE` clause is cast to a type containing nested `LowCardinality` (for example, a `Variant(LowCardinality(Date), String)`).

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=728336b344ee2de5aaace19f7a67079b3070d03c&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20amd_tsan%29 (STID 4054-600e)

## Root cause

`KeyCondition::applyFunctionChainToColumn` prepared the input constant by calling `convertToFullIfNeeded` on the column and `removeLowCardinality` on the type. The two were not symmetric:

- `removeLowCardinality` only strips the outer-level `LowCardinality` wrapper from the type.
- The default `IColumn::convertToFullIfNeeded` implementation recurses into sub-columns via `forEachSubcolumn` and strips `LowCardinality` from every nested sub-column.

`ColumnVariant` overrides `convertToFullIfNeeded` to a no-op precisely because there is no way to keep the `DataTypeVariant` in sync with per-variant column substitutions. But when the `ColumnVariant` is wrapped in a `ColumnConst` (which is what `canConstantBeWrappedByMonotonicFunctions` produces via `createColumnConst`), the default implementation on `ColumnConst` is used and bypasses the `ColumnVariant` override — stripping `LowCardinality` from the variant sub-columns while the declared `DataTypeVariant` is left untouched by `removeLowCardinality`.

The subsequent `castColumnAccurate` / `castColumnAccurateOrNull` built a `FunctionCast` from the mismatched `Variant(LowCardinality(X), ...)` type. Inside `createVariantToColumnWrapper`, the per-variant wrapper was `prepareUnpackDictionaries(LowCardinality(X), ...)`, which `typeid_cast`s its input to `ColumnLowCardinality`. At runtime the sub-column was a plain `ColumnVector`, so the server aborted with:

```
Logical error: 'Bad cast from type DB::ColumnVector<...> to DB::ColumnLowCardinality'
```

The same asymmetry is triggered by `Array(LowCardinality(X))`, `Tuple(LowCardinality(X), ...)`, and `Map(LowCardinality(K), V)` constants, but those paths are currently reachable only through the server-side AST fuzzer.

## Fix

Strip only outer-level wrappers on the column so that the transformation is symmetric with what `removeLowCardinality` does on the type:

```cpp
auto result_column = in_column
    ->convertToFullColumnIfConst()
    ->convertToFullColumnIfReplicated()
    ->convertToFullColumnIfSparse()
    ->convertToFullColumnIfLowCardinality();
auto result_type = removeLowCardinality(in_data_type);
```

Nested `LowCardinality` inside container types (`Array`, `Tuple`, `Map`, `Variant`) is preserved on both sides. Downstream code is unaffected: `FunctionCast` already handles outer-level and nested `LowCardinality` via its own wrappers, and the existing defensive re-wrap at the `func->execute` call site continues to handle the outer-LC case.

## Reproducer

```sql
SET allow_experimental_variant_type = 1;
SET allow_suspicious_low_cardinality_types = 1;

CREATE TABLE t (d LowCardinality(Date), v UInt32)
ENGINE = MergeTree
PARTITION BY toYYYYMM(d)
ORDER BY d
SETTINGS allow_suspicious_low_cardinality_types = 1;

INSERT INTO t VALUES ('2026-04-01', 1), ('2026-04-19', 2), ('2026-04-20', 3);

SELECT * FROM t
WHERE d >= CAST('2026-04-19' AS Variant(LowCardinality(Date), String));
```

- Before the fix: server aborts with `Bad cast from type DB::ColumnVector<unsigned short> to DB::ColumnLowCardinality`.
- After the fix: returns rows `(2026-04-19, 2)` and `(2026-04-20, 3)`.

## CIDB history

STID 4054-600e: 11 hits over 90 days across `serverfuzz` / `Stress test` / `AST fuzzer` checks, all going through `PartitionPruner` (7/11) or `MergeTreeIndexMinMax` (4/11) and ending in `FunctionCast::prepareUnpackDictionaries`. No open GitHub issue tracks this STID specifically.

## Test

Added `tests/queries/0_stateless/04105_key_condition_nested_lc_variant.sql` covering the `Variant(LowCardinality(Date), String)` partition-pruning path and the `Variant(LowCardinality(String), UInt64)` primary-key path.

Fix `LOGICAL_ERROR` (`Bad cast … to ColumnLowCardinality`) during MergeTree index analysis when a `WHERE` clause compares a `LowCardinality` key column with a constant cast to a type that contains nested `LowCardinality` (for example `Variant(LowCardinality(Date), String)`). [#NNNNN](https://github.com/ClickHouse/ClickHouse/pull/NNNNN)

Session: `cron:clickhouse-ci-task-worker:20260421-001500`

### Version info
- Merged into: `26.5.1.26`


groeneai:
cc @CurtizJ @KochetovNicolai — could you review this? It fixes a `LOGICAL_ERROR` (`Bad cast … to ColumnLowCardinality`) during MergeTree index analysis caused by an asymmetry between `convertToFullIfNeeded` (which recurses via `forEachSubcolumn` and bypasses `ColumnVariant`'s override when wrapped in `ColumnConst`) and `removeLowCardinality` (which only strips outer LC from the type). Related STID 4054-600e on serverfuzz/master.

groeneai:
## Pre-PR validation gate

```sql
SET allow_experimental_variant_type=1;
SET allow_suspicious_low_cardinality_types=1;
CREATE TABLE t (d LowCardinality(Date), v UInt32) ENGINE=MergeTree
PARTITION BY toYYYYMM(d) ORDER BY d SETTINGS allow_suspicious_low_cardinality_types=1;
INSERT INTO t VALUES ('2026-04-01', 1), ('2026-04-19', 2), ('2026-04-20', 3);
SELECT * FROM t WHERE d >= CAST('2026-04-19' AS Variant(LowCardinality(Date), String));
```
Aborts every run before the fix; returns correct rows after the fix.

**b) Root cause explained?** Yes.
1. `canConstantBeWrappedByMonotonicFunctions` builds a `ColumnConst` via `out_type->createColumnConst(1, value)`. For `out_type = Variant(LowCardinality(Date), String)`, this creates `ColumnConst<ColumnVariant(ColumnLowCardinality<Date>, ColumnString)>`.
2. `applyFunctionChainToColumn` calls `in_column->convertToFullIfNeeded()`. The default `IColumn` implementation runs `convertToFullColumnIfConst()->...->convertToFullColumnIfLowCardinality()` (unwraps the `Const`), then iterates sub-columns via `forEachSubcolumn` and calls `convertToFullIfNeeded` on each.
3. `ColumnVariant` has an override of `convertToFullIfNeeded` (returns `getPtr()`) precisely to avoid stripping LC inside variants — but the override is only invoked when the caller has a direct `ColumnVariant`, not when the path goes through `ColumnConst`'s default impl. The iteration strips LC from the LC variant sub-column, producing `ColumnVariant(ColumnVector<UInt16>, ColumnString)`.
4. `removeLowCardinality(Variant(LC(Date), String))` only strips outer-level LC → type stays `Variant(LC(Date), String)`.
5. `castColumnAccurateOrNull` builds a `FunctionCast` from the (now mismatched) type to the target. `createVariantToColumnWrapper` creates `prepareUnpackDictionaries(LC(Date), Date)` for variant 0, which `typeid_cast`s the runtime column to `ColumnLowCardinality`. The column is the stripped `ColumnVector<UInt16>`, so `typeid_cast` fails.

**c) Fix matches root cause?** Yes — replace the asymmetric call with an outer-only strip that matches what `removeLowCardinality` does on the type: `convertToFullColumnIfConst -> convertToFullColumnIfReplicated -> convertToFullColumnIfSparse -> convertToFullColumnIfLowCardinality`. No deep recursion; no bypass of `ColumnVariant`'s invariant.

**d) Test intent preserved / new test added?** Yes — no existing test assertions are weakened. New regression test `04105_key_condition_nested_lc_variant.sql` covers the `Variant(LowCardinality(Date), String)` partition-pruning path and the `Variant(LowCardinality(String), UInt64)` primary-key path.

- Pre-fix binary: repro query aborts server with `Bad cast from type DB::ColumnVector<unsigned short> to DB::ColumnLowCardinality`.
- Post-fix binary: repro query returns `(2026-04-19, 2), (2026-04-20, 3)`.
- New regression test passes (`04105_key_condition_nested_lc_variant`), `03402_has_low_cardinality_key_condition` still passes, broad sweep of LC / variant / key_condition / partition_pruner tests shows no regression.

- The bug pattern exists symmetrically for `Array(LowCardinality(X))`, `Tuple(LowCardinality(X), ...)`, and `Map(LowCardinality(K), V)` — any container that has nested LC and is wrapped in `ColumnConst`. All of those paths go through the same `convertToFullIfNeeded` → `forEachSubcolumn` recursion that strips inner LC. The fix addresses all of them in the single call site.
- "Am I fixing the root cause or guarding the symptom?" — Root cause. The fix removes the asymmetry at the source (column stripping) so the column/type invariant is preserved, rather than catching the mismatch later.
- Not a defensive null check — the symptom never reaches `typeid_cast` under the new stripping semantics.

Session: `cron:clickhouse-ci-task-worker:20260421-001500`

nihalzp:
@groeneai Failed stateless tests look related.

groeneai:
Thanks @nihalzp — you were right, the `04105` stateless test is related. Pushed a follow-up ([5a11f97](https://github.com/ClickHouse/ClickHouse/pull/103211/commits/5a11f97a1e0ef14f64ac900ae6bc09cab1d6c41b)) tagging it `no-parallel-replicas` with a comment explaining why.

**Root cause of the CI failures:** the new test exercises partition pruning over `LowCardinality(Date)` with a constant cast to `Variant(LowCardinality(Date), String)`. That specific shape is needed to reach `KeyCondition::canConstantBeWrappedByMonotonicFunctions` where this fix applies. Reproduced locally — the query hits the bug fix as intended.

Under parallel replicas / distributed plan the query tree is serialized with the `Date` constant folded to its raw `UInt16` representation. When the replica re-parses it as `_CAST(20562, 'Variant(LowCardinality(Date), String)')`, `FunctionCast::createColumnToVariantWrapper` (FunctionsConversion.cpp:1446) rejects the `UInt16 → Variant(LowCardinality(Date), String)` conversion with `CANNOT_CONVERT_TYPE` — the analyzer runs before index analysis, so we never reach the code path this PR fixes.

I verified this is a pre-existing parallel-replicas serialization limitation (not introduced by this fix) by reproducing the same `CANNOT_CONVERT_TYPE` via `remote()`:

```sql
SELECT d, v FROM remote('127.0.0.1:19019', currentDatabase(), 't_04105_lc_date')
WHERE d >= CAST('2026-04-19' AS Variant(LowCardinality(Date), String))
ORDER BY d;
-- Code: 70. Cannot convert type UInt16 to Variant(LowCardinality(Date), String).
```

Fixing the analyzer's Variant constant serialization is out of scope for this PR. Similar MergeTree index-analysis tests use the same tag (e.g. `03374_indexes_with_trivial_cast.sql`, `03374_indexes_with_literals.sql`).

The other five LLVM-coverage stateless failures (`04078_native_protocol_input_validation`, `02992_all_columns_should_have_comment`, `00600_replace_running_query`, `03096_order_by_system_tables`, `03258_nonexistent_db`) all passed on rerun (0/3 reproducibility) and match known LLVM-coverage timing-flake patterns — unrelated to this PR.

groeneai:
Gentle re-ping @CurtizJ @KochetovNicolai — could you take a look when you have a moment?

Quick status:
- CI green: 142 pass / 16 skip / 3 fail where all 3 failures trace to **one** pre-existing LLVM-coverage flake (`03258_nonexistent_db`, tracked separately under PR #103287). Not caused by this fix.
- `04105_key_condition_nested_lc_variant` (the new regression test) now passes on every variant including ParallelReplicas after tagging `no-parallel-replicas` per @nihalzp's earlier feedback.
- STID 4054-600e on master: **0 recurrences** in the ~50 hours since this PR was opened (last master hit was 2026-04-20 on commit `728336b3`, pre-dating this PR).
- Reproducer + both-direction validation are in the pre-PR gate comment above.

Happy to rebase or tweak if anything needs changing.

nihalzp (approved):

clickgapai:
Hi @groeneai @nihalzp — the changelog category for this PR might need updating.

**Current:** `Bug Fix`

**Why:** The PR fixes a LOGICAL_ERROR ('Bad cast from ColumnVector to ColumnLowCardinality'), which belongs in the 'Critical Bug Fix' category according to the changelog guidelines.

clickgapai:

**Affected code:** `applyFunctionChainToColumn` in `KeyCondition.cpp` — the `convertToFullIfNeeded` call strips inner `LowCardinality` from `ColumnVariant` sub-columns when the Variant is wrapped in `ColumnConst`, creating a column/type mismatch — core code present in all supported branches.

**Why:** The `ColumnVariant::convertToFullIfNeeded` override was added in PR #97831 and backported to 25.11+. But the ColumnConst bypass that this PR fixes exists in all branches with the Variant type. The `canConstantBeWrappedByMonotonicFunctions` function and `applyFunctionChainToColumn` predate all supported branches (since 2020). The `Variant` type was added in late 2023 (commit 319ae440b6ba). Since all supported branches (25.8+) contain the Variant type and the vulnerable code path, all need the fix.

PR #103229, opened 2026-04-21
Add enable_compression setting for MySQL table function and engine

bernardlim:
- Improvement

Add `enable_compression` setting for the `mysql` table function, the `MySQL` table engine, the `MySQL` database engine, dictionary `SOURCE(MYSQL)`, and named collections. When enabled, ClickHouse negotiates MySQL protocol-level compression for all data transferred over the connection.

Motivation: MySQL connections will benefit from protocol compression, **bandwidth is expensive** especially large transfers over public or bandwidth-constrained networks. `enable_compression` enables MySQL protocol compression for connections created through the `mysql` table function, `MySQL` table engine, `MySQL` database engine, dictionary `SOURCE(MYSQL)`, and named collections.

Parameters:
- `enable_compression` enables MySQL protocol compression for connections created through the `mysql` table function, `MySQL` table engine, `MySQL` database engine, dictionary `SOURCE(MYSQL)`, and named collections.

Example use:
```sql
SELECT *
FROM mysql(
    'host:3306',
    'database',
    'table',
    'user',
    'password',
    SETTINGS enable_compression = 1
);
```

Example with named collection:
```sql
SELECT *
FROM mysql(my_mysql_conn, table = 'remote_table', enable_compression = 1);
```

Example with MySQL table engine:

```sql
CREATE TABLE mysql_table
ENGINE = MySQL('host:3306', 'database', 'table', 'user', 'password')
SETTINGS enable_compression = 1;
```

Example with MySQL dictionary source:

```sql
CREATE DICTIONARY my_dict (id UInt64, value String)
PRIMARY KEY id
SOURCE(MYSQL(
    host 'host' port 3306
    user 'user' password 'password'
    db 'database' table 'table'
    enable_compression 1
))
LAYOUT(FLAT()) LIFETIME(0);
```

### Version info
- Merged into: `26.6.1.418`

### DIFF
--- a/docs/en/engines/table-engines/integrations/mysql.md
+++ b/docs/en/engines/table-engines/integrations/mysql.md
@@ -24,7 +24,8 @@ SETTINGS
     [ connection_wait_timeout=5, ]
     [ connection_auto_close=true, ]
     [ connect_timeout=10, ]
-    [ read_write_timeout=300 ]
+    [ read_write_timeout=300, ]
+    [ enable_compression=false ]
 ;
 ```
 
@@ -190,6 +191,35 @@ Possible values:
 - Positive integer.
 
 Default value: `300`.
+
+### `enable_compression` {#enable-compression}
+
+Enables compression for the MySQL protocol connection.
+
+Default value: `false`.
+
+This setting applies to:
+
+- the `MySQL` table engine;
+- the `MySQL` database engine;
+- the `mysql` table function;
+- named collections used by MySQL integrations.
+
+When enabled, ClickHouse requests compression for the connection.
+
+Example:
+
+```sql
+CREATE TABLE mysql_engine_compression
+(
+    id UInt32,
+    name String,
+    age UInt32,
+    money UInt32
+)
+ENGINE = MySQL('mysql80:3306', 'clickhouse', 'test_table', 'root', 'password')
+SETTINGS enable_compression = 1;
+```
 
 ## See also {#see-also}
 
--- a/docs/en/sql-reference/table-functions/mysql.md
+++ b/docs/en/sql-reference/table-functions/mysql.md
@@ -102,6 +102,35 @@ SELECT * FROM mysql(creds, table='test');
 └────────┴───────┘
 ```
 
+### `enable_compression` {#enable-compression}
+
+Enables compression for the MySQL protocol connection.
+
+Default value: `false`.
+
+This setting applies to:
+
+- the `mysql` table function;
+- the `MySQL` table engine;
+- the `MySQL` database engine;
+- named collections used by MySQL integrations.
+
+When enabled, ClickHouse requests compression for the connection.
+
+Example:
+
+```sql
+SELECT *
+FROM mysql(
+    'mysql80:3306',
+    'clickhouse',
+    'test_table',
+    'root',
+    'password',
+    SETTINGS enable_compression = 1
+);
+```
+
 Replacing and inserting:
 
 ```sql
--- a/src/Common/mysqlxx/Connection.cpp
+++ b/src/Common/mysqlxx/Connection.cpp
@@ -52,10 +52,11 @@ Connection::Connection(
     unsigned timeout,
     unsigned rw_timeout,
     bool enable_local_infile,
-    bool opt_reconnect)
+    bool opt_reconnect,
+    bool enable_compression)
     : Connection()
 {
-    connect(db, server, user, password, port, socket, ssl_ca, ssl_cert, ssl_key, timeout, rw_timeout, enable_local_infile, opt_reconnect);
+    connect(db, server, user, password, port, socket, ssl_ca, ssl_cert, ssl_key, timeout, rw_timeout, enable_local_infile, opt_reconnect, enable_compression);
 }
 
 Connection::Connection(const std::string & config_name)
@@ -82,7 +83,8 @@ void Connection::connect(const char* db,
     unsigned timeout,
     unsigned rw_timeout,
     bool enable_local_infile,
-    bool opt_reconnect)
+    bool opt_reconnect,
+    bool enable_compression)
 {
     if (is_connected)
         disconnect();
@@ -108,6 +110,10 @@ void Connection::connect(const char* db,
 
     /// See C API Developer Guide: Automatic Reconnection Control
     if (mysql_options(driver.get(), MYSQL_OPT_RECONNECT, reinterpret_cast<const char *>(&opt_reconnect)))
+        throw ConnectionFailed(errorMessage(driver.get()), mysql_errno(driver.get()));
+
+    /// Enable classic MySQL protocol compression if requested.
+    if (enable_compression && mysql_options(driver.get(), MYSQL_OPT_COMPRESS, nullptr))
         throw ConnectionFailed(errorMessage(driver.get()), mysql_errno(driver.get()));
 
     /// Specifies particular ssl key and certificate if it needs
--- a/src/Common/mysqlxx/mysqlxx/Pool.h
+++ b/src/Common/mysqlxx/mysqlxx/Pool.h
@@ -176,7 +176,8 @@ public:
          unsigned default_connections_ = MYSQLXX_POOL_DEFAULT_START_CONNECTIONS,
          unsigned max_connections_ = MYSQLXX_POOL_DEFAULT_MAX_CONNECTIONS,
          unsigned enable_local_infile_ = MYSQLXX_DEFAULT_ENABLE_LOCAL_INFILE,
-         bool opt_reconnect_ = MYSQLXX_DEFAULT_MYSQL_OPT_RECONNECT);
+         bool opt_reconnect_ = MYSQLXX_DEFAULT_MYSQL_OPT_RECONNECT,
+         bool enable_compression_ = false);
 
     Pool(const Pool & other)
         : default_connections{other.default_connections},
@@ -186,7 +187,8 @@ public:
           port{other.port}, socket{other.socket},
           connect_timeout{other.connect_timeout}, rw_timeout{other.rw_timeout},
           ssl_ca(other.ssl_ca), ssl_cert(other.ssl_cert), ssl_key(other.ssl_key),
-          enable_local_infile{other.enable_local_infile}, opt_reconnect(other.opt_reconnect)
+          enable_local_infile{other.enable_local_infile}, opt_reconnect(other.opt_reconnect),
+          enable_compression{other.enable_compression}
     {}
 
     Pool & operator=(const Pool &) = delete;
@@ -249,6 +251,7 @@ private:
     std::string ssl_key;
     bool enable_local_infile;
     bool opt_reconnect;
+    bool enable_compression;
 
     /// True if connection was established at least once.
     bool was_successful{false};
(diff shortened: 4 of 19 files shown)


src/Common/mysqlxx/Connection.cpp
```diff
@@ -110,6 +112,12 @@ void Connection::connect(const char* db,
     if (mysql_options(driver.get(), MYSQL_OPT_RECONNECT, reinterpret_cast<const char *>(&opt_reconnect)))
         throw ConnectionFailed(errorMessage(driver.get()), mysql_errno(driver.get()));
 
+#ifdef MYSQL_OPT_COMPRESS
```
tiandiwonder, line 115: remove MYSQL_OPT_COMPRESS,.
`#ifdef MYSQL_OPT_COMPRESS` is always `false`; the feature never activates.

src/Common/mysqlxx/Connection.cpp
```diff
@@ -52,10 +52,11 @@ Connection::Connection(
     unsigned timeout,
     unsigned rw_timeout,
     bool enable_local_infile,
-    bool opt_reconnect)
+    bool opt_reconnect,
+    bool enable_compression)
```
tiandiwonder, line 56: **`src/Dictionaries/MySQLDictionarySource.cpp:55–66`** — `"enable_compression"` missing from ExternalDictionary whitelist.

The PR correctly adds `enable_compression` to `Pool`'s config-based constructor (XML path), but `MySQLDictionarySource.cpp` maintains an explicit `ValidateKeysMultiset` whitelist:

```cpp
static const ValidateKeysMultiset<ExternalDatabaseEqualKeysSet> dictionary_allowed_keys = {
    ...
    "enable_local_infile", "opt_reconnect",   // ← enable_compression absent
    ...
};
```

tiandiwonder (commented):
**No functional tests** — The PR adds no test that verifies compression is actually negotiated. Given the blocker above, a test would have caught the defect immediately. At minimum, an integration test should confirm that a query succeeds when `enable_compression = 1` is set and that the connection is established correctly.

**Missing documentation** — `docs/en/engines/table-engines/integrations/mysql.md` has a `## Settings` section that documents each setting individually (e.g., `connection_auto_close`, `connection_pool_size`). The new `enable_compression` setting is absent from both that file and `docs/en/sql-reference/table-functions/mysql.md`.

bernardlim:
> **No functional tests** — The PR adds no test that verifies compression is actually negotiated. Given the blocker above, a test would have caught the defect immediately. At minimum, an integration test should confirm that a query succeeds when `enable_compression = 1` is set and that the connection is established correctly.
> 
> **Missing documentation** — `docs/en/engines/table-engines/integrations/mysql.md` has a `## Settings` section that documents each setting individually (e.g., `connection_auto_close`, `connection_pool_size`). The new `enable_compression` setting is absent from both that file and `docs/en/sql-reference/table-functions/mysql.md`.

Updated the PR and squashed it back into a single commit for easier review.

The current diff now includes:

* `enable_compression` documented in the MySQL table engine syntax and settings section
* a `CREATE TABLE ... SETTINGS enable_compression = 1` example
* `enable_compression` documented for the `mysql` table function
* a table-function usage example
* integration coverage in `tests/integration/test_storage_mysql/test.py`
* dictionary/named-collection support via the MySQL dictionary allowed-key whitelist
* connector-side propagation down to `MYSQL_OPT_COMPRESS`

Example usage for the MySQL table engine:

```sql
CREATE TABLE mysql_table
(
    id UInt64,
    name String
)
ENGINE = MySQL('mysql-host:3306', 'database', 'table', 'user', 'password')
SETTINGS enable_compression = 1;
```

Example usage for the MySQL table function:

```sql
SELECT *
FROM mysql(
    'mysql-host:3306',
    'database',
    'table',
    'user',
    'password'
)
SETTINGS enable_compression = 1;
```

Could you please review them, thank you @tiandiwonder.

bernardlim:
bump @tiandiwonder, can help trigger gh-bot review ?
thank you.

tests/integration/test_storage_mysql/test.py
```diff
@@ -759,6 +759,76 @@ def test_settings(started_cluster):
     conn.close()
 
 
+def test_enable_compression(started_cluster):
```
tiandiwonder, line 762: 💡 **Minor — verify compression was actually negotiated.**

This test confirms the setting is accepted and the query returns correct data through all three entry points, which catches any regression in setting plumbing. It does **not** prove the wire protocol negotiated zlib — a build that silently dropped `mysql_options(MYSQL_OPT_COMPRESS)` (like the previous round of this PR) would still pass.

Consider asserting after the query, e.g.:
```sql
SELECT * FROM mysql(..., SETTINGS enable_compression = 1, query='SHOW SESSION STATUS LIKE ''Compression''')
```
and check the value is `ON`. Same suggestion as round 1; non-blocking.

src/Dictionaries/MySQLDictionarySource.cpp
```diff
@@ -61,7 +61,7 @@ static const ValidateKeysMultiset<ExternalDatabaseEqualKeysSet> dictionary_allow
     "query", "where", "name" /* name_collection */, "socket",
     "share_connection", "fail_on_connection_loss", "close_connection",
     "ssl_ca", "ssl_cert", "ssl_key",
-    "enable_local_infile", "opt_reconnect",
+    "enable_local_infile", "opt_reconnect", "enable_compression",
```
tiandiwonder, line 64: 💡 **Minor — no regression test for the XML dictionary path.**

The whitelist change here is correct, and the named-collection dictionary path is exercised transitively via `MySQLSettings`. But the classic `<dictionary><source><mysql><enable_compression>1</enable_compression>...</mysql></source></dictionary>` XML config — the exact path this line guards — has no test. A trivial XML config in `tests/integration/test_dictionaries_mysql/` would lock it in. Non-blocking.

src/Common/mysqlxx/mysqlxx/Pool.h
```diff
@@ -249,6 +251,7 @@ class Pool final
     std::string ssl_key;
     bool enable_local_infile;
     bool opt_reconnect;
+    bool enable_compression{false};
```
tiandiwonder: 💡 **Style nit.** `bool enable_compression{false};` uses brace default-init while the neighboring `enable_local_infile` and `opt_reconnect` (lines 252-253) do not. No functional impact — every ctor assigns this field explicitly — but for consistency consider dropping the `{false}` initializer.

docs/en/sql-reference/table-functions/mysql.md
```diff
@@ -18,15 +18,15 @@ mysql({host:port, database, table, user, password[, replace_query, on_duplicate_
 
 ## Arguments {#arguments}
```
tiandiwonder, line 19: 💡 **Minor — unrelated table reformat.**

This PR rewrites the column alignment of the entire arguments table (L18–28) and tightens the `on_duplicate_clause` row by deleting its trailing `Example: …` sentence. It also adds a period to the `mysql_datatypes_support_level` note (L54). These are unrelated to `enable_compression` and add noise to `git blame`. Consider reverting these hunks and pushing the tidy-up as a separate doc PR. Non-blocking.

docs/en/engines/table-engines/integrations/mysql.md
```diff
@@ -45,7 +46,7 @@ The table structure can differ from the original MySQL table structure:
 - `password` — User password.
 - `replace_query` — Flag that converts `INSERT INTO` queries to `REPLACE INTO`. If `replace_query=1`, the query is substituted.
 - `on_duplicate_clause` — The `ON DUPLICATE KEY on_duplicate_clause` expression that is added to the `INSERT` query.
-    Example: `INSERT INTO t (c1,c2) VALUES ('a', 2) ON DUPLICATE KEY UPDATE c2 = c2 + 1`, where `on_duplicate_clause` is `UPDATE c2 = c2 + 1`. See the [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html) to find which `on_duplicate_clause` you can use with the `ON DUPLICATE KEY` clause.
+    Example: `INSERT INTO t (c1,c2) VALUES ('a', 2) ON DUPLICATE KEY UPDATE c2 = c2 + 1`, where `on_duplicate_clause` is `UPDATE c2 = c2 + 1`. See the [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html) to find which `ON DUPLICATE KEY` clause can be used with the `INSERT` query.
```
tiandiwonder: 💡 **Minor — unrelated wording change.**

This sentence reword (`See the MySQL documentation to find which on_duplicate_clause you can use with the ON DUPLICATE KEY clause` → `… ON DUPLICATE KEY clause can be used with the INSERT query`) is unrelated to the `enable_compression` feature. Same observation as the table-functions doc — consider keeping the diff focused. Non-blocking.

tiandiwonder:
Changelog entry is incomplete. It lists the mysql table function, MySQL table engine, and named collections, but the
 change (and the docs/tests) also covers the MySQL database engine and MySQL dictionary sources. Suggested wording:
` Add enable_compression setting to enable MySQL client protocol compression for the mysql table function, the MySQL table and database engines, MySQL dictionary sources, and named collection.`

tiandiwonder:
Table-function example in the PR description uses the wrong SETTINGS placement. The first example uses a trailing
  query-level SETTINGS:
```
  SELECT * FROM mysql('host:3306', 'database', 'table', 'user', 'password')
  SETTINGS enable_compression = 1;
```
  enable_compression is a MySQLSettings entry, not a query-level Setting, and TableFunctionMySQL::parseArguments only reads an
  ASTSetQuery from inside the function arguments — so this form would raise UNKNOWN_SETTING. It should be inline:
```
  SELECT * FROM mysql('host:3306', 'database', 'table', 'user', 'password',
      SETTINGS enable_compression = 1);
```

tests/integration/test_dictionaries_mysql/configs/dictionaries/mysql_dict_compression.xml
```diff
@@ -0,0 +1,30 @@
+<clickhouse>
+    <dictionary>
```
tiandiwonder, line 2: Wrong password in mysql_dict_compression.xml. It sets <password>clickhouse</password>, but the only root password on
  mysql80 is ClickHouse_MySQL_P@ssw0rd, so dict_compression never authenticates and never loads. The test only asserts
  "UNKNOWN_SETTING" not in last_exception, so the auth failure (a connection-class error) is silently masked, and the comment
  blaming the missing table is misleading. Note the Compression=ON wire assertions are fine — those use
  mysql_dict_compression_wire.xml, which has the correct password.

tiandiwonder: Suggested fix, either:
  - (a) Change the password to ClickHouse_MySQL_P@ssw0rd so the dict actually connects (this also exercises the wire path,
  making the case more valuable); or
  - (b) If the intent is to keep it in FAILED and only test "the parser accepts the setting", fix the comment to state that the
  dict cannot load due to the password (and missing table), and that the test only asserts the absence of UNKNOWN_SETTING.

tiandiwonder (commented):
optional: No regression coverage for the cache-key change. Appending ?compression= in getPoolEntryName only takes effect when share_connection=true, but neither new XML dict sets it, so the discriminator is never exercised. Consider a  share_connection=true pair differing only in enable_compression, asserting they don't share a pool (one Compression=ON, the other OFF).

bernardlim:
I reckon all CI failures are pre-existing unrelated to this PR, none touch MySQL connection pooling.

@tiandiwonder 

thanks for the feedbacks.

tiandiwonder (approved):

PR #103233, opened 2026-04-21
Add optional compression setting for PostgreSQL connections

bernardlim:
- Improvement

Added optional `compression` support for outbound PostgreSQL connections across PostgreSQL integrations: `postgresql(...)`, `ENGINE = PostgreSQL`, `ENGINE = MaterializedPostgreSQL`, `DATABASE ... ENGINE = PostgreSQL`, `DATABASE ... ENGINE = MaterializedPostgreSQL`, and PostgreSQL dictionary source configuration.

This change adds an optional `compression` parameter for ClickHouse outbound PostgreSQL connections.

When specified, ClickHouse appends `compression=<value>` to the libpq connection string used for PostgreSQL integrations.

Supported values:
- `on`
- `off`

If the value is not specified, existing behavior is unchanged.

This applies to:
- `postgresql(...)` table function / PostgreSQL storage configuration
- `ENGINE = PostgreSQL`
- `ENGINE = MaterializedPostgreSQL`
- `DATABASE ... ENGINE = PostgreSQL`
- `DATABASE ... ENGINE = MaterializedPostgreSQL`
- PostgreSQL dictionary source configuration

Motivation:
- Allow users to explicitly enable or disable PostgreSQL connection compression for outbound PostgreSQL integrations.

Parameters:
- `compression` — optional string parameter for PostgreSQL connection configuration
- Supported values: `on`, `off`
- Default: empty / unset
- Invalid values are rejected during configuration parsing

Example use:

Table function:
```sql
SELECT *
FROM postgresql(
    'postgres1:5432',
    'db',
    'table',
    'user',
    'password',
    'public',
    '',
    'on'
);
```

PostgreSQL database engine:
```sql
CREATE DATABASE pg_db
ENGINE = PostgreSQL(
    'postgres1:5432',
    'mydb',
    'user',
    'password',
    '',
    0,
    'on'
);
```

MaterializedPostgreSQL database engine:
```sql
CREATE DATABASE mpg_db
ENGINE = MaterializedPostgreSQL(
    'postgres1:5432',
    'mydb',
    'user',
    'password',
    'on'
);
```

Named collection example:
```sql
CREATE NAMED COLLECTION my_pg AS
    host = 'localhost',
    port = 5432,
    database = 'mydb',
    user = 'u',
    password = 'p',
    compression = 'on';
```


tiandiwonder (commented):
PostgreSQL wire protocol v3.0 has been in use since PG 7.4 (2003) with no protocol-level compression    . The protocol defines no `CompressedData`, `CompressionAck`, or `SetCompressionMethod` message type    s. PostgreSQL 18 introduced protocol v3.2, the first version bump since 2003 — but it is narrowly sc    oped to 256-bit cancel keys for security hardening. No compression was added.

bernardlim:
Thanks for the review. I verified this further, and standard PostgreSQL/libpq does not currently provide the native wire-protocol compression support that this PR assumed.

Going to close / pause this PR rather than add a user-facing setting that would be misleading. If PostgreSQL upstream lands supported connection compression in a future release, I can revisit this with the correct semantics and version scope.

Issue #102282, opened 2026-04-09
Redesign MergeTree remote reading


CheSema:
## Related paper
https://www.vldb.org/pvldb/vol16/p2769-durner.pdf

## Problem

The current MergeTree remote read path has several limitations that make it hard to control resource usage and get consistent performance.

### 1. Greedy task distribution wastes prefetch budget

`MergeTreePrefetchedReadPool::fillPerThreadTasks` distributes tasks sequentially — thread 0 gets tasks until its sum of granules reaches the per-thread target, then thread 1, etc. `startPrefetches` then fires all budgeted prefetches from the front of the queue. With a limited budget, most prefetches land on the first reader's tasks. Other readers get no prefetch benefit, and the prefetched data sits in memory waiting while thread 0 works through it sequentially.

### 2. Prefetch is one-shot with no replenishment

Once `startPrefetches` drains the budget, no new prefetches are created for the rest of the query. Even long-running queries on an idle instance get no further benefit after the initial burst.

### 3. Double-buffering is limited to one buffer ahead

`AsynchronousBoundedReadBuffer::prefetch` submits a single async read of `buffer_size` (1MB). After it's consumed, subsequent reads within the same task are synchronous. There is no continuous pipeline of async reads.

### 4. No global resource limits

There is no shared budget for memory, connections, or bandwidth across queries. The HTTP connection pool's `hard_limit` (200k connections as configured) is the only backstop, and it is too coarse — by the time it kicks in, the system is already overloaded.

### 5. Unbounded connection lifetime

A connection opened by `ReadBufferFromS3` stays alive as long as the underlying `impl` (`ReadBufferFromGetObjectResult`) exists. There is no timeout or maximum lifetime — a slow consumer can hold a connection open indefinitely.

### 6. Throttler extends connection lifetime

The read throttler is applied between read syscalls during data consumption. This means throttling pauses the reader while the connection remains open, extending its lifetime and keeping the kernel receive buffer occupied. Under heavy throttling, connections are held open much longer than necessary.

## Requirements for the new design

**R1. Bounded concurrent connections.** A global limit on simultaneously open S3 connections, shared across all queries. When the limit is reached, new connection requests wait (with timeout).

**R2. Fair resource sharing between queries.** Each query gets a fair share of the connection/bandwidth budget. No single query can starve others. Queries should get at least some minimum allocation; if all resources are assigned, new queries are rejected rather than silently degraded.

[...]

PR #103234, opened 2026-04-21
Introduce ReadPipeline for declarative read buffer composition

CheSema:
Introduce `ReadPipeline` — a declarative builder for the read buffer chain that replaces imperative "matryoshka" nesting in `DiskObjectStorage::readFile` and `StorageObjectStorageSource::createReadBuffer`.

### Problem

Multiple places manually constructed chains of nested read buffers (`CachedOnDiskReadBufferFromFile`, `ReadBufferFromRemoteFSGather`, `AsynchronousBoundedReadBuffer`, `CachedInMemoryReadBufferFromFile`, `ReadBufferFromDistributedCache`) with ad-hoc coupling logic. Each had to get `use_external_buffer` flags, buffer sizes, settings propagation, and seek restrictions right on its own. The chains differ structurally — MergeTree reads use Gather with per-object caching, table engine reads (`s3()`, `url()`) wrap a single object directly — but they share the same stages and coupling rules. This led to coupling bugs that are hard to catch in review and hard to test in isolation.

### Solution

Callers declare **what** stages they need; `ReadPipeline::build()` owns **how** to couple them. One place to get buffer ownership right.

```cpp
// DiskObjectStorage::prepareRead — declares stages
pipeline.setSource(object_storage, storage_objects, read_hint);
pipeline.needGather();
pipeline.needDiskCache(cache, cache_log);
pipeline.needAsyncPrefetch(reader, counters, prefetches_log);

// ReadPipeline::build() — assembles the chain with correct coupling
auto buf = pipeline.build();
```

`IDisk::readFile` is now a thin non-virtual wrapper: `prepareRead()` + `pipeline.build()`. All 117+ callers of `readFile` go through the pipeline without any migration.

### Stages implemented

| Stage | Description |
|-------|-------------|
| **Source** | `ObjectStorageSource`, `LocalFileSource`, `BackupSource`, `CustomSource` |
| **DiskCache** | `CachedOnDiskReadBufferFromFile` with correct `use_external_buffer` coupling |
| **Gather** | `ReadBufferFromRemoteFSGather` for multi-blob logical files |
| **DistributedCache** | `ReadBufferFromDistributedCache` with fallback |
| **MemoryCache** | `CachedInMemoryReadBufferFromFile` (page cache) |
| **AsyncPrefetch** | `AsynchronousBoundedReadBuffer` |
| **Encryption** | `ReadBufferFromEncryptedFile` (supports multiple layers for double encryption) |

### Entry points migrated

- `DiskObjectStorage::prepareRead` — MergeTree reads on object storage disks
- `StorageObjectStorageSource::createReadBuffer` — table engine reads (`s3()`, `url()`, etc.)
- `DiskEncrypted::prepareRead` — encrypted disk reads (delegates + adds decryption stage)
- `CachedObjectStorage::prepareRead` — adds disk cache stage to the pipeline
- `CachedObjectStorage::readObject` — now a pass-through (caching moved to pipeline stage)

### Tests

19 unit tests + 6 integration tests with real `DiskObjectStorage` on local blob storage. All 23 existing `DiskObjectStorageTest` tests pass through the new code path.

Ref: https://github.com/ClickHouse/ClickHouse/issues/102282

...

---

> [!NOTE]
> **High Risk**
> High risk because it refactors core read-buffer construction across local disks, object storage, encrypted disks, backups, and MergeTree part reads, affecting caching, seeking, and async-prefetch behavior in performance- and correctness-critical paths.
> 
> **Overview**
> Introduces `IO/ReadPipeline` as a declarative builder for composing read-buffer chains (source, filesystem cache, gather, distributed cache, page cache, async prefetch, decryption) and adds extensive unit/integration tests validating stage ordering, seeking, cloning, and encrypted/object-storage reads.
> 
> Refactors the read API by making `IDisk::readFile` a non-virtual wrapper over new virtual `prepareRead`, and migrates `DiskObjectStorage`, `DiskLocal`, `DiskEncrypted`, `DiskBackup`, and MergeTree `IDataPartStorage`/`MergeTreeReaderStream` to populate a `ReadPipeline` instead of manually nesting buffers.
> 
> Updates object-storage interfaces to pass explicit `use_external_buffer`/`restrict_seek` flags, moves filesystem-cache composition into pipeline stages (including `CachedObjectStorage::prepareRead`), splits cache-related settings out of `ReadSettings`, and adjusts async/gather/cache buffer implementations accordingly (including a regression test preventing `mmap`+page-cache on local disks).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a309b2d408aa0253a39989a62e20916536aa5cb7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.119`


src/Storages/ObjectStorage/StorageObjectStorageSource.cpp
```diff
@@ -917,102 +918,89 @@ std::unique_ptr<ReadBufferFromFileBase> createReadBuffer(
...
-        ? std::max<size_t>(effective_read_settings.remote_fs_buffer_size, effective_read_settings.prefetch_buffer_size)
-        : effective_read_settings.remote_fs_buffer_size;
+        /// No needGather() — StorageObjectStorageSource reads single objects directly
+        /// (one logical file = one S3/Azure blob). Gather is only needed for DiskObjectStorage
+        /// where metadata maps a logical path to multiple blobs.
 
-    if (object_size)
-        buffer_size = std::min<size_t>(object_size, buffer_size);
+        /// Page cache
+        if (use_page_cache)
```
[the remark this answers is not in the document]
CheSema: This is not a regression. The exact same logic exists on master:

```cpp
bool use_page_cache = !use_distributed_cache && !use_filesystem_cache
    && effective_read_settings.page_cache && effective_read_settings.use_page_cache_for_object_storage;
```

Page cache was already excluded for DC reads in `StorageObjectStorageSource` before this PR. No change in behavior.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+            impl = std::visit(overloaded{
+                [&](const ObjectStorageSource & s) -> std::unique_ptr<ReadBufferFromFileBase>
+                {
+                    return s.storage->readObject(object, settings, s.read_hint, use_ext_buf, /* restrict_seek */ false);
+                },
+                [&](const LocalFileSource & s) -> std::unique_ptr<ReadBufferFromFileBase>
+                {
+                    return createReadBufferFromFileBase(
+                        s.path, settings, s.read_hint);
```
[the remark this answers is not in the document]
CheSema: Not an issue. `CachedInMemoryReadBufferFromFile` calls `in->set(ptr, size)` to pass page-cache-owned memory to the inner reader. Local file readers (`ReadBufferFromFile` with pread) naturally support this -- pread reads into whatever buffer is set. No special `use_external_buffer` flag needed.

The `use_external_buffer` flag is specific to remote readers (S3, Azure) where `nextImpl` can replace the set buffer with an HTTP stream buffer. Local readers do not have this problem.

This matches master behavior: `createReadBufferFromFileBase` returned a plain `ReadBufferFromFile` without any external buffer flag, and `CachedInMemoryReadBufferFromFile` wrapped it successfully.

src/Storages/ObjectStorage/StorageObjectStorageSource.cpp
```diff
@@ -917,102 +918,91 @@ std::unique_ptr<ReadBufferFromFileBase> createReadBuffer(
...
+                reader,
+                context_->getAsyncReadCounters(),
+                context_->getFilesystemReadPrefetchesLog());
+        }
 
-    size_t buffer_size = prefer_bigger_buffer_size
-        ? std::max<size_t>(effective_read_settings.remote_fs_buffer_size, effective_read_settings.prefetch_buffer_size)
-        : effective_read_settings.remote_fs_buffer_size;
+        /// Prefer bigger buffer size when filesystem cache is active
+        if (modified_read_settings.filesystem_cache_prefer_bigger_buffer_size && use_filesystem_cache)
```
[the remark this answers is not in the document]
CheSema: Acknowledged. The old code used `impl->isCached()` -- a runtime check after the buffer was built. In the pipeline approach, buffer size is set before building, so we cannot know at that point whether a specific read will hit cache.

Using `use_filesystem_cache` (config-time check) means cache misses may get the bigger buffer too. This is a minor inefficiency (slightly larger buffers on misses), not a correctness issue. The tradeoff is inherent to the declarative pipeline design -- parameters are set at configuration time, not after the first read.

This could be revisited if profiling shows measurable overhead from the larger buffer on cache misses.

src/Disks/IO/CachedOnDiskReadBufferFromFile.cpp
```diff
@@ -257,11 +262,9 @@ std::shared_ptr<ReadBufferFromFileBase> getCacheReadBuffer(
         info.cache_file_reader.reset();
     }
 
-    ReadSettings local_read_settings{info.settings};
+    ReadSettings local_read_settings;
```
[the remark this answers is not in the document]
CheSema: Fixed in 9e4308293a1b — `local_throttler` is now threaded through `ReadInfo` and applied in `getCacheReadBuffer`.

src/Disks/DiskLocal.cpp
```diff
@@ -367,9 +368,34 @@ bool DiskLocal::renameExchangeIfSupported(const std::string & old_path, const st
...
+
+    /// No gather for local disk — the source buffer is returned directly.
+    pipeline.setLocalFileSource(
+        full_path.string(),
+        StoredObjects{obj},
+        settings,
+        read_hint);
+
+    /// Page cache for local disk reads — handled as a pipeline stage.
+    if (settings.use_page_cache_for_local_disks && settings.page_cache)
```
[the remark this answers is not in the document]
CheSema: Fixed in 4de88cadf405 — restored async method exclusions and O_DIRECT alignment check in `DiskLocal::prepareRead`.

src/Disks/DiskObjectStorage/ObjectStorages/Cached/CachedObjectStorage.cpp
```diff
@@ -57,42 +58,39 @@ bool CachedObjectStorage::exists(const StoredObject & object) const
...
-                cache->getCommonOriginWithSegmentKeyType(object.local_path),
-                read_buffer_creator,
-                modified_read_settings,
-                std::string(CurrentThread::getQueryId()),
-                object.bytes_size,
-                /* allow_seeks */!read_settings.remote_read_buffer_restrict_seek,
-                /* use_external_buffer */read_settings.remote_read_buffer_use_external_buffer,
-                /* read_until_position */std::nullopt,
-                global_context->getFilesystemCacheLog());
+            pipeline.needDiskCache(cache, read_settings.getFilesystemCacheSettings(), global_context->getFilesystemCacheLog());
```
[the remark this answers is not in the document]
CheSema: Fixed in c9f6e50ef0e2 — `ReadPipeline` now uses `std::vector<DiskCacheStage>` and chains them from innermost to outermost in `build`.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,608 @@
...
+}
+
+void ReadPipeline::needGather()
+{
+    gather = true;
+}
+
+void ReadPipeline::needDiskCache(FileCachePtr cache, FilesystemCacheSettings cache_settings, std::shared_ptr<FilesystemCacheLog> cache_log)
+{
+    disk_cache = DiskCacheStage{.cache = std::move(cache), .cache_log = std::move(cache_log), .cache_settings = std::move(cache_settings), .custom_cache_key = std::nullopt, .custom_origin = std::nullopt};
```
antonio2368: format into muultiple lines

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,608 @@
...
+
+void ReadPipeline::setBackupSource(std::shared_ptr<IBackup> backup, String path, StoredObjects objects, const ReadSettings & read_settings)
+{
+    source = SourceStage{
+        .objects = std::move(objects),
+        .source = BackupSource{.backup = std::move(backup), .path = std::move(path)},
+        .read_settings = read_settings};
+}
+
+void ReadPipeline::setSource(StoredObjects objects, BufferCreator creator, const ReadSettings & read_settings)
```
antonio2368: Extreme nit but in other places you have function specific args first, so maybe move `creator` as first argument?

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,608 @@
...
+        .cache = std::move(cache),
+        .cache_log = std::move(cache_log),
+        .cache_settings = std::move(cache_settings),
+        .custom_cache_key = std::move(cache_key),
+        .custom_origin = std::move(origin)};
+}
+
+void ReadPipeline::needMemoryCache(std::shared_ptr<PageCache> cache, String cache_path_prefix, PageCacheSettings page_cache_settings)
+{
+    memory_cache = MemoryCacheStage{.cache = std::move(cache), .cache_path_prefix = std::move(cache_path_prefix), .page_cache_settings = std::move(page_cache_settings), .custom_cache_path = {}, .custom_file_version = {}};
```
antonio2368: format

src/Disks/IO/CachedOnDiskReadBufferFromFile.cpp
```diff
@@ -1415,9 +1434,10 @@ size_t CachedOnDiskReadBufferFromFile::readBigAt(
 {
     ReadInfo current_info(
         info.cache_key, info.source_file_path, info.implementation_buffer_creator,
-        info.use_external_buffer, info.settings, /* read_until_position */range_begin + n);
+        info.use_external_buffer, info.cache_settings, info.local_fs_buffer_size,
+        /* read_until_position */range_begin + n);
```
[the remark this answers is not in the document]
CheSema: Fixed in 3b165608fe5e — forward `info.local_throttler` in `readBigAt`.

src/Disks/DiskLocal.cpp
```diff
@@ -367,9 +369,46 @@ bool DiskLocal::renameExchangeIfSupported(const std::string & old_path, const st
...
+    /// Page cache for local disk reads — handled as a pipeline stage.
+    /// Page cache is incompatible with async local read methods (io_uring,
+    /// pread_fake_async, pread_threadpool) and with O_DIRECT when the page
+    /// cache block size is not aligned to the direct IO sector size.
+    bool use_page_cache = settings.use_page_cache_for_local_disks && settings.page_cache
+        && settings.local_fs_method != LocalFSReadMethod::io_uring
+        && settings.local_fs_method != LocalFSReadMethod::pread_fake_async
+        && settings.local_fs_method != LocalFSReadMethod::pread_threadpool;
+
+    if (use_page_cache && settings.direct_io_threshold
```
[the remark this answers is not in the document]
CheSema: Fixed in 2735d0db269e — now checks `file_size >= direct_io_threshold` before disabling page cache, matching master behavior.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,612 @@
...
+                 custom_origin = dc.custom_origin](
+                    bool restricted_seek, const StoredObject & object) mutable
+                    -> std::unique_ptr<ReadBufferFromFileBase>
+            {
+                auto cache_key = custom_key.value_or(FileCacheKey::fromPath(object.remote_path));
+                auto origin = custom_origin.value_or(cache->getCommonOriginWithSegmentKeyType(object.local_path));
+
+                /// Copy, not move: gather_creator may be called multiple times (once per object).
+                auto prev_copy = prev_creator;
+                auto impl_creator = [prev_copy, restricted_seek, object]() mutable
```
antonio2368:
```suggestion
                auto impl_creator = [prev_copy = prev_creator, restricted_seek, object]() mutable
```

src/IO/ReadSettings.cpp
```diff
@@ -47,13 +47,29 @@ ReadSettings ReadSettings::adjustBufferSize(size_t file_size) const
     return res;
 }
 
-ReadSettings ReadSettings::withNestedBuffer(bool seekable) const
+FilesystemCacheSettings ReadSettings::getFilesystemCacheSettings() const
```
antonio2368, line 50: why not store in the struct `FilesystemCacheSettings`?
You duplicate defaults in the definition and you copy values on each access.
Also the effort to keep this function in sync.

src/IO/ReadSettings.cpp
```diff
@@ -47,13 +47,29 @@ ReadSettings ReadSettings::adjustBufferSize(size_t file_size) const
...
+        .filesystem_cache_boundary_alignment = filesystem_cache_boundary_alignment,
+        .filesystem_cache_allow_background_download = filesystem_cache_allow_background_download,
+        .filesystem_cache_reserve_space_wait_lock_timeout_milliseconds = filesystem_cache_reserve_space_wait_lock_timeout_milliseconds,
+        .filesystem_cache_max_download_size = filesystem_cache_max_download_size,
+        .filesystem_cache_skip_download_if_exceeds_per_query_cache_write_limit = filesystem_cache_skip_download_if_exceeds_per_query_cache_write_limit,
+        .enable_filesystem_cache_log = enable_filesystem_cache_log,
+    };
+}
+
+PageCacheSettings ReadSettings::getPageCacheSettings() const
```
antonio2368, line 64: Same comment as above

antonio2368 (commented):
First batch

src/IO/ReadSettings.cpp
CheSema: I did a short cut here by implementing this `getFilesystemCacheSettings`. In general I want to go here your way, but that would bring even more changes with repacking `ReadSettings` struct.

I would prefer to do it right after this change in separate PR.

src/Disks/DiskLocal.cpp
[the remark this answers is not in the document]
CheSema: Already fixed in 2735d0db269e — now checks `file_size >= direct_io_threshold`.

alexey-milovidov:

src/Disks/DiskLocal.cpp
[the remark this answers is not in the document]
CheSema: Fixed in f8c513fe813 — now uses `read_hint.value_or(file_size)` as the estimated size, matching `createReadBufferFromFileBase` behavior.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,624 @@
...
+                cache_file.file_version = *memory_cache->custom_file_version;
+        }
+        else
+        {
+            const auto & first_object = source->objects.at(0);
+            cache_file.path = memory_cache->cache_path_prefix + first_object.remote_path;
+        }
+
+        impl = std::make_unique<CachedInMemoryReadBufferFromFile>(
+            cache_file, memory_cache->cache, std::move(impl), settings);
```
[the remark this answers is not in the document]
CheSema: Fixed in 079a2623767 — stage-level `PageCacheSettings` are now applied when provided, falling back to source settings.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,636 @@
...
+                        "ReadPipeline: disk cache without gather requires ObjectStorageSource or CustomSource");
+                }
+
+                /// Wrap inner cache layers (all except the outermost). Each layer's
+                /// impl_creator produces the previous layer's CachedOnDiskReadBufferFromFile.
+                /// Inner layers always use use_external_buffer=true.
+                for (size_t i = 0; i + 1 < disk_caches.size(); ++i)
+                {
+                    const auto & dc = disk_caches[i];
+                    auto fs_cache_settings = dc.cache_settings.value_or(settings.getFilesystemCacheSettings());
```
antonio2368: When can cache_settings be nullopt?

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,636 @@
...
+            /// use_external_buffer=false because DC calls set() on the fallback with its
+            /// own internal_buffer which may be empty in external buffer mode.
+            auto fallback_creator = [storage = dc_obj_source->storage,
+                                     read_hint = dc_obj_source->read_hint,
+                                     captured_object = source->objects.at(0),
+                                     captured_settings = settings]()
+                -> std::unique_ptr<ReadBufferFromFileBase>
+            {
+                return storage->readObject(captured_object, captured_settings, read_hint,
+                    /* use_external_buffer */ false, /* restrict_seek */ false);
```
antonio2368: This seems different from the old behavior, it was always true for both AFAIS

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,636 @@
...
+            };
+
+            /// DC manages its own buffer (reads from TCP). Same as the gather path.
+            bool use_ext_buf_for_dc = false;
```
antonio2368: Shouldn't this also depend if we use async prefetch/memory cache?

src/IO/ReadPipeline.cpp
CheSema: You are right, the same with `page_cache_settings`

src/IO/ReadPipeline.cpp
CheSema: You are right here also. 
DC provides its own buffer to the falback buffer.

`restrict_seek` here should not be `true`, this is the case when we work without gather, no need to restrict seek. In the more popular branch -- gather implements seeks.

src/IO/ReadPipeline.cpp
CheSema: It should indeed.

src/Disks/IO/CachedOnDiskReadBufferFromFile.cpp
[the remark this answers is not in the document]
CheSema: Already fixed in 9e4308293a1b — `local_throttler` is threaded through `ReadInfo` and applied in `getCacheReadBuffer` at line 272. The bot keeps re-flagging this against stale snapshots.

antonio2368 (approved):

src/Disks/DiskObjectStorage/ObjectStorages/IObjectStorage.h
```diff
@@ -212,7 +214,20 @@ class IObjectStorage
     virtual std::unique_ptr<ReadBufferFromFileBase> readObject( /// NOLINT
         const StoredObject & object,
         const ReadSettings & read_settings,
-        std::optional<size_t> read_hint = {}) const = 0;
+        std::optional<size_t> read_hint = {},
+        bool use_external_buffer = false,
+        bool restrict_seek = false) const = 0;
```
kssenii, line 219: Let's add small comments for `use_external_buffer` and `restrict_seek`? Those are non-obvious flags

src/Disks/DiskObjectStorage/ObjectStorages/IObjectStorage.h
```diff
@@ -212,7 +214,20 @@ class IObjectStorage
...
+
+    /// Populate a ReadPipeline with the source and any stages this storage needs.
+    /// `self` is the shared_ptr to this storage (needed because setSource takes ObjectStoragePtr).
+    /// Default: sets the source to `self`. CachedObjectStorage overrides
+    /// to delegate to the inner storage and add the disk cache stage.
```
kssenii: > Populate a ReadPipeline with the source

Not immediately clear what "source" is in this context, so may be a small explanation would be useful.
May be just add in parentheses a mentioning of read buffer - `Populate a ReadPipeline with the source (read buffer)`

kssenii:
```suggestion
    /// Populate a ReadPipeline with the source and any stages this storage needs.
    /// `self` is the shared_ptr to this storage (needed because setSource takes ObjectStoragePtr).
    /// Default: sets the source to `self`. CachedObjectStorage overrides
    /// to delegate to the inner storage and adds the disk cache stage.
```

src/Disks/DiskObjectStorage/DiskObjectStorage.cpp
```diff
@@ -747,138 +750,95 @@ String DiskObjectStorage::getWriteResourceNameNoLock() const
...
     auto read_settings = updateIOSchedulingSettings(settings, getReadResourceName(), getWriteResourceName());
-    /// We wrap read buffer from object storage (read_buf = object_storage->readObject())
-    /// inside ReadBufferFromRemoteFSGather, so add nested buffer setting.
-    read_settings = read_settings.withNestedBuffer();
-
+    auto global_context = Context::getGlobalContextInstance();
+    auto storage = object_storages->takePointingTo(cluster->getLocalLocation());
 
-    bool use_distributed_cache = false;
+    /// Distributed cache — computed early because prefer_bigger_buffer_size needs to know.
```
kssenii:
```suggestion
    /// Distributed cache — computed early because prefer_bigger_buffer_size needs to be known.
```

src/Disks/DiskObjectStorage/DiskObjectStorage.cpp
```diff
@@ -747,138 +750,95 @@ String DiskObjectStorage::getWriteResourceNameNoLock() const
...
-                    /*use_external_buffer*/use_page_cache || use_async_buffer,
-                    read_from_object_storage_gather);
-    }
-#endif
+        pipeline.needDistributedCache();
 
-    if (!impl)
-        impl = read_from_object_storage_gather();
+    /// Memory cache (page cache).
+    const bool file_cache_enabled = storage->supportsCache() && read_settings.enable_filesystem_cache;
```
kssenii: This flag can be moved above and reused for `prefer_bigger_buffer_size` calculation

src/IO/ReadSettings.cpp
```diff
@@ -47,13 +47,29 @@ ReadSettings ReadSettings::adjustBufferSize(size_t file_size) const
...
 
-ReadSettings ReadSettings::withNestedBuffer(bool seekable) const
+FilesystemCacheSettings ReadSettings::getFilesystemCacheSettings() const
 {
-    ReadSettings res = *this;
-    if (!seekable)
-        res.remote_read_buffer_restrict_seek = true;
-    res.remote_read_buffer_use_external_buffer = true;
-    return res;
+    return FilesystemCacheSettings{
```
kssenii, line 52: May be store these as struct in `ReadSettings` from the very start? To avoid declaring them twice and to avoid consistency maintenance between them. Like we already do for `DistributedCacheSettings distributed_cache_settings;`.

src/IO/ReadPipeline.h
```diff
@@ -0,0 +1,232 @@
...
+    {
+        String path;
+        std::optional<size_t> read_hint;
+    };
+
+    /// Backup storage source.
+    struct BackupSource
+    {
+        std::shared_ptr<IBackup> backup;
+        String path;
```
kssenii, line 72: I see that backup method also has `read_hint`, should we make use of it like we do for the above sources?
```
void DiskBackup::prepareRead(
    const String & path,
    const ReadSettings & settings,
    std::optional<size_t> /* read_hint */,
    ReadPipeline & pipeline) const
```

src/IO/ReadPipeline.h
```diff
@@ -0,0 +1,232 @@
...
+    void setSource(BufferCreator creator, StoredObjects objects, const ReadSettings & read_settings);
+
+    /// -- Gather stage (ReadBufferFromRemoteFSGather) --
+    /// Joins multiple stored objects into a single seekable buffer.
+    /// Required for object storage where one logical file maps to multiple blobs.
+    /// Not needed for local disk where one file = one file.
+    void needGather();
+
+    /// -- Disk cache stage --
+    void needDiskCache(FileCachePtr cache, FilesystemCacheSettings cache_settings, std::shared_ptr<FilesystemCacheLog> cache_log = nullptr);
```
kssenii: nit: at least for me `Disk` prefix seems a bit confusing, could it be better as `needFilesystemCache` for consistency with other filesystem cache related structs and settings? Same for `struct DiskCacheStage` (-> `struct FilesystemCacheStage`) and some of the below methods

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+{
+    source = SourceStage{
+        .objects = std::move(objects),
+        .source = ObjectStorageSource{.storage = std::move(object_storage), .read_hint = read_hint},
+        .read_settings = read_settings};
+}
+
+void ReadPipeline::setLocalFileSource(String path, StoredObjects objects, const ReadSettings & read_settings, std::optional<size_t> read_hint)
+{
+    source = SourceStage{
```
kssenii, line 57: Should we add `chassert(!source.has_value());` in the beginning of these `set*Source` methods?

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+std::unique_ptr<ReadBufferFromFileBase> ReadPipeline::build() const
+{
+    if (!source)
+        throw Exception(ErrorCodes::BAD_ARGUMENTS, "ReadPipeline: source stage is not set, call setSource first");
+
+    const auto & settings = source->read_settings;
+
+    if (source->objects.empty())
+        throw Exception(ErrorCodes::BAD_ARGUMENTS, "ReadPipeline: source has no stored objects");
+
```
kssenii, line 198:
```suggestion
    if (source->objects.empty())
        throw Exception(ErrorCodes::BAD_ARGUMENTS, "ReadPipeline: source has no stored objects");

    const auto & settings = source->read_settings;
```
The order can be flipped

src/Disks/DiskObjectStorage/ObjectStorages/Cached/CachedObjectStorage.h
```diff
@@ -137,6 +146,8 @@ class CachedObjectStorage final : public IObjectStorage
 
     ObjectStoragePtr getUnderlying() override { return object_storage; }
 
+    FileCachePtr getFileCache() const { return cache; }
```
kssenii: As far as I see, this method is unused.

src/IO/ReadSettings.cpp
CheSema: I will do that right after. If I change ReadSettings it brings more changes every where. I do understand that ReadSettings can hold special blocks for each specific cache or reader. It really help with understanding them. But not in that PR already.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+                return std::make_unique<ReadBufferFromRemoteFSGather>(
+                    std::move(creator_copy),
+                    objects,
+                    captured_settings.remote_read_min_bytes_for_seek,
+                    /* use_external_buffer */ true,
+                    /* buffer_size */ 0);
+            };
+
+            impl = DistributedCache::readWithDistributedCache(
+                source->objects.at(0).remote_path,
```
kssenii: As I see, it should be `local_path` here, not `remote_path`.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+
+std::unique_ptr<ReadBufferFromFileBase> ReadPipeline::build() const
+{
+    if (!source)
+        throw Exception(ErrorCodes::BAD_ARGUMENTS, "ReadPipeline: source stage is not set, call setSource first");
+
+    const auto & settings = source->read_settings;
+
+    if (source->objects.empty())
+        throw Exception(ErrorCodes::BAD_ARGUMENTS, "ReadPipeline: source has no stored objects");
```
kssenii: It seems we used to consider this case as normal: https://github.com/ClickHouse/ClickHouse/blob/131ef1d5bc85e8abff29a1fda82ed59ab398615a/src/Disks/DiskObjectStorage/DiskObjectStorage.cpp#L758-L759

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+        impl = std::make_unique<CachedInMemoryReadBufferFromFile>(
+            cache_file, memory_cache->cache, std::move(impl), page_cache_read_settings);
+    }
+
+    /// -- Stage 5: Async prefetch --
+
+    if (async_prefetch && async_prefetch->reader)
+    {
+        size_t total_size = getTotalSize(source->objects);
+        size_t async_buffer_size = settings.remote_fs_buffer_size;
```
kssenii: Will this also be used for local read?

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+            /// fallback to hand over DC's working buffer. The fallback must accept external
+            /// memory.
+            ///
+            /// restrict_seek=false: DC calls `buffer->seek(file_offset_of_buffer_end, SEEK_SET)`
+            /// on the fallback before reading. In the gather path the enclosing Gather handles
+            /// seeks internally (per-object readers can have restricted seek). Here there is
+            /// no Gather, so the underlying object reader must support seeks directly.
+            auto fallback_creator = [storage = dc_obj_source->storage,
+                                     read_hint = dc_obj_source->read_hint,
+                                     captured_object = source->objects.at(0),
```
kssenii:
```suggestion
                                     captured_object = object,
```
As I see we already have `const auto & object = source->objects[0];` in this scope

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+            {
+                return storage->readObject(captured_object, captured_settings, read_hint,
+                    /* use_external_buffer */ true, /* restrict_seek */ false);
+            };
+
+            /// DC uses external buffer mode when a downstream stage (memory cache or
+            /// async prefetch) wraps it — matches the gather path and master behavior.
+            bool use_ext_buf_for_dc = memory_cache.has_value() || async_prefetch.has_value();
+            impl = DistributedCache::readWithDistributedCache(
+                source->objects.at(0).remote_path,
```
kssenii: Same as above - `object` as well + https://github.com/ClickHouse/ClickHouse/pull/103234#discussion_r3288014190

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+                    cache_key,
+                    outermost.cache,
+                    origin,
+                    std::move(impl_creator),
+                    fs_cache_settings,
+                    settings.remote_fs_buffer_size,
+                    settings.local_fs_buffer_size,
+                    std::string(CurrentThread::getQueryId()),
+                    object.bytes_size,
+                    /* allow_seeks_after_first_read */ true,
```
kssenii, line 487: Why `allow_seeks_after_first_read` is always true? As I see, it used to be conditional https://github.com/ClickHouse/ClickHouse/blob/131ef1d5bc85e8abff29a1fda82ed59ab398615a/src/Disks/DiskObjectStorage/ObjectStorages/Cached/CachedObjectStorage.cpp#L85

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+                impl = std::make_unique<CachedOnDiskReadBufferFromFile>(
+                    object.remote_path,
+                    cache_key,
+                    outermost.cache,
+                    origin,
+                    std::move(impl_creator),
+                    fs_cache_settings,
+                    settings.remote_fs_buffer_size,
+                    settings.local_fs_buffer_size,
+                    std::string(CurrentThread::getQueryId()),
```
kssenii: We can save query id into a variable above and reuse for all cached buffer creations

src/Disks/DiskObjectStorage/ObjectStorages/IObjectStorage.h
CheSema: will add a commnet.
My opinion that `restrict_seek` is superfluous here. I will remove it in the next iteration.

src/Disks/DiskObjectStorage/ObjectStorages/IObjectStorage.h
CheSema: I have changed the comment

CheSema: changed

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,650 @@
...
+#include <Common/CurrentThread.h>
+#include <Common/Exception.h>
+
+#if ENABLE_DISTRIBUTED_CACHE
+#include <DistributedCache/Utils.h>
+#endif
+
+/// Helper for std::visit with multiple lambdas.
+template <class... Ts>
+struct overloaded : Ts...
```
kssenii: Move into anonymous namespace or DB? Currently it is at global

src/Disks/DiskObjectStorage/DiskObjectStorage.cpp
CheSema: moved

src/IO/ReadPipeline.h
CheSema: https://github.com/ClickHouse/ClickHouse/blob/37a40e20d46a6067769c2e184b9208bb7069e0ed/src/Disks/DiskBackup.cpp#L147-L151

Backup does not use it.

src/IO/ReadPipeline.h
CheSema: renamed

src/IO/ReadPipeline.cpp
CheSema: added

src/IO/ReadPipeline.cpp
CheSema: ok, applied

src/Disks/DiskObjectStorage/ObjectStorages/Cached/CachedObjectStorage.h
CheSema: removed

src/IO/ReadPipeline.cpp
CheSema: fixed

src/IO/ReadPipeline.cpp
CheSema: yes, fixed the other place too

src/IO/ReadPipeline.cpp
CheSema: In master AsynchronousBoundedReadBuffer is never used with DiskLocal, DiskLocal has its own pread_threadpool.
I do not know why, it just is.
Added a comment here.

src/IO/ReadPipeline.cpp
CheSema: applied

src/IO/ReadPipeline.cpp
CheSema: https://github.com/ClickHouse/ClickHouse/blob/37a40e20d46a6067769c2e184b9208bb7069e0ed/src/Storages/ObjectStorage/StorageObjectStorageSource.cpp#L1055-L1072

I see is as `allow_seeks_after_first_read` or `seekable` was actually always true for no-Gather path.
Only buffer with Gather did account it properly.

In my opinion it is better to remove this property. It only about promises from caller, the buffer has to be able to handle seeks calls anyway.

src/IO/ReadPipeline.cpp
CheSema: did it

src/IO/ReadPipeline.cpp
CheSema: did

src/Disks/IO/CachedOnDiskReadBufferFromFile.cpp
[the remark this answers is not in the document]
CheSema: Intentional and documented — the comment block immediately above this line (added in d5b9a1da1c1) spells out which fields we deliberately do NOT propagate for cache-file reads, and why:

- `direct_io_threshold`, `mmap_threshold`/`mmap_cache` — cache files are hot (the same segments are read many times across queries). The kernel page cache keeps recently-read pages in RAM, so subsequent reads avoid disk I/O. O_DIRECT / mmap would bypass the kernel page cache and force disk reads on every access — counterproductive for cache files.
- `priority`, `enable_filesystem_read_prefetches_log` — only used by async local readers (`io_uring`, `pread_fake_async`, `pread_threadpool`). We force `local_fs_method = pread`, so these are irrelevant.
- `remote_fs_*`, `page_cache*`, prefetch-related — target the remote read path or pipeline stages above this layer; don't apply to local cache-file I/O.

Master copied the full settings here and so could inherit those flags for cache reads. The narrower default is by design, not an oversight — see the comment block at lines 269-289 in current HEAD.

src/Disks/DiskLocal.cpp
```diff
@@ -367,9 +369,53 @@ bool DiskLocal::renameExchangeIfSupported(const std::string & old_path, const st
...
+
+    {
+        /// Use the same estimated size basis as createReadBufferFromFileBase:
+        /// read_hint first, then file_size. A large file with a small read_hint
+        /// won't trigger O_DIRECT, so page cache remains safe.
+        size_t estimated_size = read_hint.value_or(file_size);
+        if (use_page_cache && settings.direct_io_threshold
+            && estimated_size >= settings.direct_io_threshold
+            && settings.page_cache_block_size % DEFAULT_AIO_FILE_BLOCK_SIZE != 0)
+            use_page_cache = false;
```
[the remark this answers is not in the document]
CheSema: This isn't a regression — master uses the **same compile-time constant** for this alignment check.

[`createReadBufferFromFileBase.cpp#L191-L211`](https://github.com/ClickHouse/ClickHouse/blob/37a40e20d46a6067769c2e184b9208bb7069e0ed/src/Disks/IO/createReadBufferFromFileBase.cpp#L191-L211) on current master:

```cpp
constexpr size_t min_alignment = DEFAULT_AIO_FILE_BLOCK_SIZE;
...
if (use_page_cache && settings.page_cache_block_size % min_alignment)
{
    LOG_TEST(getLogger("createReadBufferFromFileBase"),
        "Not using userspace page cache because page cache block size is not divisible by direct IO alignment");
    use_page_cache = false;
}
```

`min_alignment` is a `constexpr`, set to `DEFAULT_AIO_FILE_BLOCK_SIZE`. Master never queried the runtime filesystem sector size from the opened file descriptor — the bot's premise that there was a runtime check on master is incorrect.

`DiskLocal::prepareRead` preserves that behavior 1:1.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,673 @@
...
+            };
+
+            impl = DistributedCache::readWithDistributedCache(
+                source->objects.at(0).local_path,
+                source->objects,
+                settings,
+                *dc_obj_source->storage,
+                use_external_buffer,
+                std::move(fallback_creator),
+                distributed_cache->include_credentials_in_cache_key);
```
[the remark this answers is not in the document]
CheSema: Addressed in 861251baf55 (private 59e3a36bb9b): added `chassert(impl, "readWithDistributedCache must return a valid buffer or throw")` after both DC call sites in `ReadPipeline::build`.

The old `if (!impl) fallback` was load-bearing on master because of a historical contract that no longer holds: `readWithDistributedCache` either returns a valid buffer or throws — it never returns nullptr. The fallback was dead code. The `chassert` documents the invariant and would catch any regression.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,684 @@
...
+
+void ReadPipeline::needDecryption(String path, size_t buffer_size, KeyFinderFunc key_finder)
+{
+    decryption_stages.push_back(DecryptionStage{
+        .path = std::move(path),
+        .buffer_size = buffer_size,
+        .key_finder = std::move(key_finder)});
+}
+
+std::unique_ptr<ReadBufferFromFileBase> ReadPipeline::build() const
```
kssenii, line 159: This method is ~450 lines of code, could we split it or unify some parts or restructure, because currently with so many if-else branches it becomes difficult to read?

CheSema: My aim is it to remove this method in the near future. 
But lets break it down with few methods.

CheSema: I split it into sub-methods

src/Disks/DiskObjectStorage/DiskObjectStorage.cpp
```diff
@@ -747,138 +749,92 @@ String DiskObjectStorage::getWriteResourceNameNoLock() const
...
-    /// inside ReadBufferFromRemoteFSGather, so add nested buffer setting.
-    read_settings = read_settings.withNestedBuffer();
+    auto global_context = Context::getGlobalContextInstance();
+    auto storage = object_storages->takePointingTo(cluster->getLocalLocation());
 
+    /// Empty objects (zero-blob file) — set an empty source so `ReadPipeline::build`
+    /// returns `ReadBufferFromEmptyFile`. No stages are needed below the source.
+    if (storage_objects.empty())
+    {
+        pipeline.setSource(std::move(storage), StoredObjects{}, settings);
```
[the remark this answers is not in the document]
CheSema: ReadBufferFromEmptyFile is empty file no settings are needed.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,698 @@
...
+    {
+        if (!dec.key_finder)
+            continue;
+
+        if (impl->eof())
+        {
+            /// Empty encrypted file — no header, return empty buffer.
+            return std::make_unique<ReadBufferFromFileDecorator>(
+                std::make_unique<ReadBufferFromString>(std::string_view{}), dec.path);
+        }
```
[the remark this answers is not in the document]
CheSema: This is the same pattern as `master`'s `DiskEncrypted::readFile` ([source](https://github.com/ClickHouse/ClickHouse/blob/37a40e20d46a6067769c2e184b9208bb7069e0ed/src/Disks/DiskEncrypted.cpp#L427-L447)):

```cpp
auto buffer = delegate->readFile(wrapped_path, settings, read_hint);
if (buffer->eof())
    return ReadBufferFromFileDecorator(ReadBufferFromString(""), wrapped_path);
FileEncryption::Header header = readHeader(*buffer);
...
return std::make_unique<ReadBufferFromEncryptedFile>(path, ..., std::move(buffer), key, header);
```

`delegate->readFile` already returns the full chain — including any async-prefetch wrap — and master then calls `eof()` and `readHeader` on it before handing off to `ReadBufferFromEncryptedFile`. `wrapDecryption` simply moved this code into the pipeline. No regression.

`AsynchronousBoundedReadBuffer` does not have an `external_buffer` constructor parameter — it manages its own internal memory — so `eof()` on it is safe regardless of upstream stages.

src/Storages/ObjectStorage/StorageObjectStorageSource.cpp
```diff
@@ -1007,32 +1004,34 @@ std::unique_ptr<ReadBufferFromFileBase> createReadBuffer(
     /// because CachedOnDiskReadBufferFromFile does not work as an independent buffer currently.
     bool use_async_buffer = use_prefetch || use_distributed_cache;
 
-    std::unique_ptr<ReadBufferFromFileBase> impl;
-#if ENABLE_DISTRIBUTED_CACHE
+    /// Prefer bigger buffer size when filesystem cache is active.
```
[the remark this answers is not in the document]
CheSema: The bot's own analysis acknowledges this:

> the new pipeline inherits this exact logic [...] matching old behavior

`use_async_buffer = use_prefetch || use_distributed_cache` is the same expression used on master (see [`StorageObjectStorageSource.cpp` on master](https://github.com/ClickHouse/ClickHouse/blob/37a40e20d46a6067769c2e184b9208bb7069e0ed/src/Storages/ObjectStorage/StorageObjectStorageSource.cpp#L1052-L1056)). The `FIXME` comment about `CachedOnDiskReadBufferFromFile` predates this PR and is unrelated to the `ReadPipeline` refactor. Not a regression.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,698 @@
...
+        .buffer_size = buffer_size,
+        .key_finder = std::move(key_finder)});
+}
+
+std::unique_ptr<ReadBufferFromFileBase> ReadPipeline::build() const
+{
+    if (!source)
+        throw Exception(ErrorCodes::BAD_ARGUMENTS, "ReadPipeline: source stage is not set, call setSource first");
+
+    /// Empty objects = zero-blob file. Master's `DiskObjectStorage::readFile` handled
```
kssenii: Let's validate agent's comments to not include references to master branch and how this code used to look before, since this is confusing to read once this ends up in master branch.

CheSema: Indeed, rewritted comments.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,698 @@
...
+            },
+            [&](const BackupSource & s) -> std::unique_ptr<ReadBufferFromFileBase>
+            {
+                return s.backup->readFile(s.path);
+            },
+            [&](const CustomSource & s) -> std::unique_ptr<ReadBufferFromFileBase>
+            {
+                return s.creator(object, settings, use_ext_buf, /* restrict_seek */ false);
+            }
+        }, source->source);
```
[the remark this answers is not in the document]
CheSema: Fixed in 36a63b6e134 (public 89ae9793848): restructured `buildSingleObjectStage` into three mutually exclusive branches, each returning its own final `impl`

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,692 @@
...
+        .key_finder = std::move(key_finder)});
+}
+
+std::unique_ptr<ReadBufferFromFileBase> ReadPipeline::build() const
+{
+    if (!source)
+        throw Exception(ErrorCodes::BAD_ARGUMENTS, "ReadPipeline: source stage is not set, call setSource first");
+
+    /// Empty objects = zero-blob file. Return `ReadBufferFromEmptyFile` so callers
+    /// don't need workarounds (e.g. injecting a dummy `StoredObject`).
```
kssenii:
```suggestion
    /// Empty objects = zero-blob file.
```
Just a suggestion, the comment seems a bit excessive.

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,692 @@
...
+    {
+        const auto * dc_obj_source = std::get_if<ObjectStorageSource>(&source->source);
+        if (!dc_obj_source)
+            throw Exception(ErrorCodes::BAD_ARGUMENTS,
+                "ReadPipeline: distributed cache requires ObjectStorageSource");
+
+        if (!filesystem_caches.empty())
+            throw Exception(ErrorCodes::BAD_ARGUMENTS,
+                "ReadPipeline: filesystem cache + distributed cache without gather is not supported. "
+                "Use needGather() to enable the gather path which handles both stages");
```
kssenii, line 381: I do not see a reason not to, but looks like it is the same in master, could you please add a FIXME comment just to remember this case needs to be fixed in the future?

src/IO/ReadPipeline.cpp
```diff
@@ -0,0 +1,692 @@
...
+    /// context, so calling `CurrentThread::getQueryId()` there would return "".
+    const std::string query_id(CurrentThread::getQueryId());
+
+    auto impl = gather
+        ? buildGatherStage(query_id)        // Stages 1+2+3 (+3.5 DC)
+        : buildSingleObjectStage(query_id); // Stages 1+2 (+2.5 DC)
+
+    impl = wrapMemoryCache(std::move(impl));   // Stage 4
+    impl = wrapAsyncPrefetch(std::move(impl)); // Stage 5
+    impl = wrapDecryption(std::move(impl));    // Stage 6 (encryption)
```
kssenii, line 180: Would be perfect if stages 1-3 could be also as nicely (via separate wrap methods) done as stages 4-6, but I see there can be some ugliness with flags or because of a few small differences in support, so probably ok as it is now.

kssenii (approved):

src/IO/ReadPipeline.cpp
CheSema: It really would be better. It requires all state to be independently pluggable. Right not they are not. Some stages wraps other buffers, some take creator for buffers beneath, and there are lot of setting dependencies. 
Propose to leave it now, I have plans to redo it soon.

src/IO/ReadPipeline.cpp
CheSema: Applied in 48190af079b (public 8878073a280).

src/IO/ReadPipeline.cpp
CheSema: Added a `FIXME` in 48190af079b (public 8878073a280) describing what would need to change to lift the restriction (feed the FS-cache-wrapped source into DC's fallback creator, mirroring the gather path).

src/IO/ReadPipeline.cpp
CheSema: added.
Sorry I do not apply your suggestions from the UI because I do it for public master and cherri-pick to the private manually, do not rely on the automatic synchronization.

src/IO/ReadSettings.cpp
CheSema: Here it is : https://github.com/ClickHouse/ClickHouse/pull/105778

src/IO/ReadSettings.cpp
CheSema: Here it is : https://github.com/ClickHouse/ClickHouse/pull/105778

PR #103243, opened 2026-04-21
Kafka: map columns with special names to message metadata on INSERT

### LINKED ISSUE #85610: Kafka Engine - Granular control to map columns to Kafka message Key, Value, and Headers when producing
### Use case

Our scenario follows a standard consume-process-produce pattern:

### 1. Consume (Kafka to ClickHouse)

- We ingest messages from a source topic into a MergeTree table.
- We correctly store the original Kafka Key, Value (as JSON), and Headers into separate columns (`msg_key`, `msg_value`, `header_names`, `header_values`). 

This part works perfectly.

### 2. Produce (ClickHouse to Kafka)

- We then use another Materialized View to push processed data from our MergeTree table to a new Kafka topic.

## The Problem

We need to reconstruct the Kafka message with the same fidelity as the original. Specifically:

- The Key must be set from our `msg_key` column.
- The Value (Payload) must be exclusively the JSON object from our `msg_value` column.
- The Headers must be reconstructed from our header_names and header_values columns.

The current implementation fails at this. 
1. If we use `SELECT msg_key AS _key, msg_value FROM ...`, the resulting Kafka message payload looks like `{"_key": "...", "msg_value": {...}}`, which is incorrect. 
The payload should just be `{...}`
2. There is no options to manage kafka message header when producing.

- Kafka Message Key: The content of `source.msg_key`
- Kafka Message Value: The content of `source.msg_value`. Crucially, the payload should NOT contain `_key`, `_value`, or `_header*` fields. It should be the pure JSON object from the msg_value column.
- Kafka Message Headers: A list of headers constructed from the `_header_names` and `_header_values` arrays.

alexey-milovidov:
Adds a new Kafka table setting `kafka_map_virtual_columns_on_write`
(default `false`). When enabled, columns in the Kafka table schema named
`_key`, `_timestamp`, `_headers.name` and `_headers.value` are produced
as the corresponding Kafka message key, timestamp and headers, and are
excluded from the message payload. Default behavior is unchanged.

This lets consume-process-produce pipelines reconstruct Kafka messages
with the original key, timestamp and headers without polluting the
payload with extra fields like `_key`.

Closes https://github.com/ClickHouse/ClickHouse/issues/85610.

- New Feature

Added a new Kafka table setting `kafka_map_virtual_columns_on_write`. When enabled, columns named `_key`, `_timestamp`, `_headers.name` and `_headers.value` in the Kafka table schema are produced as the corresponding Kafka message key, timestamp and headers on `INSERT`, and are excluded from the message payload.


alexey-milovidov:
https://github.com/ClickHouse/ClickHouse/pull/103248

alexey-milovidov (commented):
Looks good.

PR #103248, opened 2026-04-21
Enable Kafka integration tests on arm64

alexey-milovidov:
The `tests/integration/test_storage_kafka` suite was previously `pytest.mark.skip`-ed on ARM because `confluentinc/cp-kafka:5.2.0`, `confluentinc/cp-schema-registry:5.2.0` and `zookeeper:3.4.9` are all x86_64-only images and could not even start on an arm64 host.

Bump the images used in `docker_compose_kafka.yml` to multi-arch variants (`confluentinc/cp-kafka:7.9.0`,
`confluentinc/cp-schema-registry:7.9.0`, `zookeeper:3.8`) when the current host is arm64, and remove the unconditional skip.

On amd64 the defaults keep the exact image tags that were used before, so the tests that pass today should not be affected.

Starting from `zookeeper:3.5` the `clientPort` is no longer defaulted to `2181`, it has to be provided via the `;<port>` suffix in `ZOO_SERVERS`, so the server list is also overridden on arm64.

Enable Kafka integration tests on arm64.

### Version info
- Merged into: `26.5.1.76`


alexey-milovidov:
Excellent!

PR #103264, opened 2026-04-21
Add embedding() function with ONNX Runtime and llama.cpp backends

canhld94:
- Experimental Feature

Introduce a new SQL function `embedding(text)` that computes text embeddings using models loaded at server startup. Models are registered in server config under `<embedding><model>...</model></embedding>` and selected per query via the `embedding_model` setting.

Backends:
- ONNX Runtime: load any HuggingFace ONNX encoder with its tokenizer.json.
- llama.cpp: load any GGUF embedding model (quantized or not); bundles vocab.

New system table `system.embedding_models` reports loaded models, memory, metadata, quantization info, runtime stats, etc. — snapshotted under a single shared lock.


canhld94:
Context: we're using ClickHouse for feature store. I think we can do it without pulling data out of ClickHouse, so I decided to integrate an inference framework. It worked.
And there's a lot more thing we can do with inference framework: embedding, semantic chunking, classifier, text generation... Small models are more powerful nowadays.

Internally, we have impelled full indexing pipeline (semantic chunking -> embeddings). But for upstream submission, I want to narrow down the scope, so only `embedding` function here.

Also, happy to discuss what inference backend is best for ClickHouse. ONNX is a generic graph computation framework, while llama.cpp targets LLM. Both work for embedding.

There's other options out there I haven't tried: Candle, OpenVino, native (we implemented native backend for static model and so far it has best performance).

rschu1ze:
The [new and experimental AI functions](https://clickhouse.com/docs/sql-reference/functions/ai-functions) (we plan to make them non-experimental) provide support for embedding generation already. They provide a superset of the functionality in this PR - local models can be called via a local endpoint.

alexey-milovidov:
@canhld94, let's implement it with UDF drivers: https://clickhouse.com/docs/reference/functions/regular-functions/udf#driver-based-executable-user-defined-functions

PR #103275, opened 2026-04-21
ci: restore OK status for flaky tests on LLVM coverage builds

maxknv:
PR #102892 replaced the RETRIES stage with a DIAGNOSTICS stage but dropped the
behavior introduced in #95763, where tests that passed on retry were marked as
`OK` on LLVM coverage builds. The new DIAGNOSTICS stage applied the `flaky` label
but left the test status as `FAIL`, causing flaky tests to appear red in CI instead
of green.

Restore: when the DIAGNOSTICS stage classifies a test as `flaky`, set its status
to `OK` for `is_llvm_coverage` builds — matching the original RETRIES stage logic.

### Version info
- Merged into: `26.4.1.1113`

### DIFF
--- a/ci/jobs/functional_tests.py
+++ b/ci/jobs/functional_tests.py
@@ -688,6 +688,10 @@ def main():
                 if label_key in label_map:
                     test_case.set_label(label_map[label_key])
                 if label_key == "flaky" and is_llvm_coverage:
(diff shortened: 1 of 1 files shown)


davenger:
LGTM, maybe just add a comment

davenger (approved):

PR #103286, opened 2026-04-21
fix(promql): malformed JSON from /api/v1/query_range on empty aggregations

JTCunning:
When running the prometheus/compliance test harness, the robots found a regression.

The fix itself is pretty simple, so the majority of the line change is comments on behavior and confirmation via tests.
Let me know if I need to slim down either.

The below is the robot's description of the behavior/tests/fixes. It's verbose but I like it, so I kept it.

---

## Summary

The Prometheus query API endpoints (`/api/v1/query`, `/api/v1/query_range`) returned an unbalanced JSON body when an aggregation over a non-existent metric reached the `timeSeriesFromGrid` step function. Reproducer:

```bash
END=$(date -u +%s); START=$((END-900)); STEP=15
curl -sS "http://127.0.0.1:9093/api/v1/query_range" \
  --data-urlencode "query=count(nonexistent_metric_name)" \
  --data-urlencode "start=$START" --data-urlencode "end=$END" \
  --data-urlencode "step=$STEP"
```

returned HTTP 400 with body:

```
{"status":"success","data":{"resultType":"matrix","result":[
  {"status":"error","errorType":"bad_data",
   "error":"Number of values (0) doesn't match number of steps (61):
    while executing 'FUNCTION timeSeriesFromGrid(...)'"}
```

The outer `{`, the data `{`, and the `result` `[` are all left unclosed. The Prometheus Go client (`v1.apiResponse.Data`) errors with `ReadArrayCB: expect ] in the end, but found <EOF>`. All five trivial aggregations (`count`/`sum`/`avg`/`min`/`max`) over a non-existent metric hit this, as does any case where the inner aggregation produces zero rows but `timeSeriesFromGrid` is still asked to emit N steps.

Two cooperating defects, both fixed here:

### 1. Function-level (`src/Functions/TimeSeries/timeSeriesRange.cpp`)

`timeSeriesFromGrid` threw `Number of values (0) doesn't match number of steps (N)` for any row whose values array was empty. The Prometheus contract for `query_range` over an empty aggregation is an empty matrix series, not a hard error. The function now treats `num_values == 0` as "no samples for this row" and emits an empty per-row time series. The truly-mismatched case (`0 < num_values != num_steps`) still throws `BAD_ARGUMENTS`, preserving the existing strict contract.

The row-0 base offset is computed with an explicit `(i == 0) ? 0 : (*values_offsets)[i - 1]` guard. `IColumn::Offsets` is a `PaddedPODArray<UInt64>` with reverse padding, so `offsets[-1] == 0` is well-defined (the canonical idiom in `ColumnArray::offsetAt`/`sizeAt`); the explicit guard makes the row-0 case obvious at the call site without depending on that invariant.

### 2. Handler-level (`src/Server/PrometheusRequestHandler.cpp`, `src/Storages/TimeSeries/PrometheusHTTPProtocolAPI.cpp`)

The PromQL HTTP handler streamed the success envelope (`{"status":"success","data":{"resultType":"matrix","result":[`) directly into the response output stream, then on exception appended a second `{"status":"error",...}` object onto the still-open buffer — the unbalanced body above. The fix has two parts:

- **`PrometheusHTTPProtocolAPI::writeQueryResponse`** performs the first `PullingPipelineExecutor::pull(...)` *before* emitting any byte of the success envelope. The dominant PromQL evaluation failure mode (e.g. `timeSeriesFromGrid` over an empty aggregation, malformed PromQL parse errors) raises on that first pull because the underlying pipeline is lazy, so it propagates with the response stream still untouched. Bulk row data continues to stream to the network (no in-memory accumulation, no peak-memory regression for large `query_range` responses).

- **`QueryAPIImpl::handlingRequestWithContext`** branches its catch block on `body_buf.count() == 0`:
  - **Pre-output** (no bytes written yet): set HTTP 4xx and emit a single, well-formed structured-error JSON document. The error message is JSON-escaped via `writeJSONString` (it embeds SQL fragments that can contain quotes/backslashes which would otherwise produce another class of invalid JSON).
  - **Post-output** (envelope or rows already in flight): re-throw. The outer `PrometheusRequestHandler::handleRequest` catch invokes `WriteBufferFromHTTPServerResponse::cancelWithException`, which appends a `__exception__` marker block in a new chunk and intentionally skips the final empty chunk to break HTTP. The client sees a connection-level abort instead of concatenated JSON — wire format stays well-defined and **a second JSON document is never written**.

Either fix on its own defends against the specific reproducer; both land because the function-level fix restores the regressed PASS for the specific aggregation queries, while the handler-level fix is defense-in-depth so any future PromQL function failure (early or late) cannot produce malformed JSON.

## Tests

- **`tests/queries/0_stateless/03254_timeseries_range.sql`** gains three cases right after the existing wrong-count `BAD_ARGUMENTS` block: empty `Array(Nullable(Float64))`, empty `Array(Float64)`, and a multi-row case mixing empty and populated arrays via `arrayJoin` to prove an empty row does not poison sibling rows in the same block. The existing `BAD_ARGUMENTS` cases are unchanged.

- **`tests/integration/test_prometheus_protocols/test_evaluation.py`** gains two complementary tests:
  - `test_query_range_empty_aggregation_returns_empty_matrix` — strict regression guard for the function-level fix. Issues `count`/`sum`/`avg`/`min`/`max(nonexistent_metric_name)` over a 15-minute / 15-second-step range and asserts HTTP 200 + `status == "success"` + empty matrix. Cannot silently re-pass if `timeSeriesFromGrid` regresses to throwing.
  - `test_query_range_invalid_promql_returns_structured_error` — companion guard for the handler's pre-output error path. Issues a syntactically broken PromQL expression and asserts HTTP `400` + `status == "error"` + `errorType == "bad_data"` + presence of the `error` field. Locks both the status code and envelope shape so a regression on either layer is caught.

Verified end-to-end against a live ClickHouse instance: all five reproducer queries now return HTTP 200 with `{"status":"success","data":{"resultType":"matrix","result":[]}}`; parse-time and other pre-stream error paths now return clean `{"status":"error","errorType":"bad_data","error":"..."}` envelopes with HTTP 400; late-stream failures abort the chunked transfer cleanly via `cancelWithException` rather than appending a second JSON document.

Made-with: Cursor

- Bug fix

Fix malformed JSON from `/api/v1/query_range` on empty aggregations


src/Functions/TimeSeries/timeSeriesRange.cpp
```diff
@@ -351,6 +366,7 @@ Converts an array of values `[x1, x2, x3, ...]` to an array of tuples
 
 The current timestamp is increased by `step` until it becomes greater than `end_timestamp`
 If the number of the values doesn't match the number of the timestamps, the function throws an exception.
+An empty input array `[]` is treated as "no samples for this row" and yields an empty result `[]`.
```
vitlibar, line 369: No, this fix is wrong, I'll make my PR to address the issues.

vitlibar:
I created two PRs to address the issues:
- https://github.com/ClickHouse/ClickHouse/pull/104425
- https://github.com/ClickHouse/ClickHouse/pull/104741

PR #103251, opened 2026-04-21
Revert "CI: add CI Tests job to PR workflow"

leshikus:
Reverts ClickHouse/ClickHouse#102529

### Version info
- Merged into: `26.4.1.1103`

PR #103295, opened 2026-04-21
CI: re-add CI Tests job; skip orphan-cleanup test until latent bug in performance_tests.py is fixed (#103307)

leshikus:
Re-adds the \`CI Tests\` job that was reverted in #103251.

## Why the original was reverted

#102529 added \`--init\` (tini as PID 1) to all \`docker run\` invocations in \`runner.py\`. This exposed a latent bug in \`performance_tests.py\`: \`ClickHouseServer.start\` launched the server without \`start_new_session=True\`, so the server inherited the driver's process group. \`Utils.terminate_process_group\` then called \`os.killpg(pgid, SIGTERM)\` on the shared group, killing the driver. Previously the driver ran as PID 1 (immune to \`SIGTERM\` by default); \`--init\` removed that immunity and caused it to exit with code 143.

The proper fix is \`start_new_session=True\` on the \`Popen\` calls in \`performance_tests.py\` — tracked in #103307, which also restores \`--init\` in \`runner.py\`.

## What this PR does

Re-adds the CI Tests job as a short-term unblock, without \`--init\`, so that CI Tests runs while #103307 is reviewed:

- \`ci/jobs/ci_tests_job.py\` — new job entry point; adds a \`--skip\` flag to exclude named tests
- \`ci/jobs/scripts/clickhouse_service.py\` — \`ClickHouseService\` context manager for starting/stopping a ClickHouse binary in tests
- \`ci/tests/0_stateless/01_parallel_sleep.sh\` — replaces \`00058_select_sleep_3.sh\` as the orphan-test fixture (self-contained, no running server required)
- \`ci/tests/test_cleanup_test_groups.py\` — updated to use the new fixture
- \`ci/workflows/pull_request.py\` + regenerated YAML — wires \`ci_tests\` into the PR workflow as a gate for all build and test jobs
- \`ci/praktika/result.py\` — minor fix

## On skipping \`test_cleanup_kills_orphaned_test_process\`

This is a **workaround**, not a permanent decision. The test requires tini as PID 1 and cannot pass without \`--init\`. It is skipped via the new \`--skip\` option (not deleted) so it is easy to re-enable once #103307 lands and \`--init\` is restored.

### Version info
- Merged into: `26.5.1.50`


maxknv (approved):
Please check that all jobs have run in CI before the merge. Change the digest somehow in perf tests or add `runner.py` to the digest config. Also, it seems the title and body do not correspond to the change.

maxknv:
Could you update the description to be self-contained (not relying on the reverted PR context)?

1. Skipping `test_cleanup_kills_orphaned_test_process` a permanent decision or a workaround? That test seemed to be a core motivation for the job in the pr that was reverted.
2. "Removing --init" is confusing — it's not in this PR's diff. Out of curiosity, what is the actual technical reason `--init` can't be used in a privileged DinD container? AFAIR, we had an issue with the perf test job, but it does not use the DinD setup.

PR #103315, opened 2026-04-22
Reject GRANT role TO itself

zxuhan:
`GRANT role_a TO role_a` silently succeeded and produced a self-referential entry in `system.role_grants`. Reject it with `BAD_ARGUMENTS` in `InterpreterGrantQuery::execute` before any grantees are mutated. Multi-role grants that include a self-grant are also rejected.

Circular grants across multiple roles (e.g. `GRANT a TO b; GRANT b TO a`) are still allowed today and are out of scope — only the direct self-grant reported in the issue is blocked.

Closes #101357

`GRANT role TO role` on the same role is now rejected with `BAD_ARGUMENTS` instead of silently creating a self-referential entry in `system.role_grants`.

- [x] Documentation is not required (bug fix)

### Version info
- Merged into: `26.8.1.183` (included in `26.8` and later)


src/Interpreters/Access/InterpreterGrantQuery.cpp
```diff
@@ -450,6 +450,19 @@ BlockIO InterpreterGrantQuery::execute()
...
+    /// A role cannot be granted to itself.
+    for (const auto & grantee_id : grantees)
+    {
+        if (std::find(roles_to_grant.begin(), roles_to_grant.end(), grantee_id) != roles_to_grant.end())
+        {
+            auto entity = access_control.tryRead(grantee_id);
+            throw Exception(
+                ErrorCodes::BAD_ARGUMENTS,
+                "Cannot grant role '{}' to itself",
+                entity ? entity->getName() : String{});
```
tiandiwonder, line 462: [InterpreterGrantQuery.cpp:457-462] tryRead is called solely to retrieve the role name for the error
   message, but it can return null if the role is concurrently dropped in the narrow window between UUID
   resolution (line 442/451) and this read — yielding "Cannot grant role '' to itself". Use the UUID as
  a fallback:

  ▎ entity ? entity->getName() : toString(grantee_id)

tiandiwonder (approved):

tiandiwonder:
@clickgapai the ci failures seems not related to the pr, how can it be merged?

src/Interpreters/Access/InterpreterGrantQuery.cpp
```diff
@@ -450,6 +450,19 @@ BlockIO InterpreterGrantQuery::execute()
     RolesOrUsersSet roles_to_revoke;
     collectRolesToGrantOrRevoke(access_control, query, roles_to_grant, roles_to_revoke);
 
+    /// A role cannot be granted to itself.
```
tiandiwonder, line 453: The same invariant is bypassed by the config-file grant path.

This PR makes "a role cannot be granted to itself" an invariant, but only enforces it here in the SQL execution path (`InterpreterGrantQuery::execute`). The configuration path that builds roles from `users.xml` / config does not go through this check.

`UsersConfigParser::parseRole` calls `parseGrant(*role, ...)` for each `<grants>` entry, and `parseGrant` resolves the role name via `generateID(AccessEntityType::ROLE, role_name)` — which for a configured role yields that role's own UUID — then calls `entity.granted_roles.grant(roles_to_grant)` directly. `GrantedRoles::grant` is a bare `roles.insert(...)` with no self-reference guard.

As a result a config such as:

```xml
<roles>
    <r>
        <grants>
            <query>GRANT r</query>
        </grants>
    </r>
</roles>
```

still creates the same self-referential role grant and surfaces it in `system.role_grants`, contradicting the invariant this PR establishes.

Suggestion: centralize the self-grant rejection before mutating `Role::granted_roles` (or add the same check in `UsersConfigParser::parseGrant`), and add a focused config/integration regression for that path.

PR #103323, opened 2026-04-22
CI: do not swallow ServerDied exception in clickhouse-test

leshikus:
Remove the `except ServerDied` handler at the end of `tests/clickhouse-test` so that an unhandled `ServerDied` exception produces a Python stack trace instead of only printing the exception message. This makes it easier to diagnose where the server death was detected.

...

### Version info
- Merged into: `26.5.1.52`

### DIFF
--- a/tests/clickhouse-test
+++ b/tests/clickhouse-test
@@ -6206,6 +6206,10 @@ if __name__ == "__main__":
 
     try:
         main(args)
+    except ServerDied as e:
+        print(f"{e}", file=sys.stderr)
+        traceback.print_exc()
+        sys.exit(1)
     except Terminated as e:
         print(f"Terminated with {e.signal} signal", file=sys.stderr)
         sys.exit(128 + e.signal)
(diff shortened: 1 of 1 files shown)


tests/clickhouse-test
maxknv, line 6210: It's better to keep try/catch with a dedicated exit code for cases like ServerDied.
You can print a traceback for debugging in the handler

leshikus: the exit code for uncaught error is exactly the same (1), you can check `python3 -c "raise RuntimeError('demo')"; echo $?`

maxknv, line 6209: It's good to return a dedicated exit code for this type of error; the calling side can rely on it.
you can print the exception in the handler with the traceback.print_exc() or similar

leshikus: but what's the point? this just adds more lines of code with the same result, doesn't it? the exit code is 1 anyway, adding the printing a stack trace and exiting with 1 works exactly as not catching the exception, check the example from the following conversation

maxknv: I prefer to be explicit on script outcomes for standard errors

maxknv: exit code 1, probably not the best for the error, but it's another story

maxknv (approved):

PR #103334, opened 2026-04-22
Fix Parquet ColumnIndex stats min_value > max_value for String columns

### LINKED ISSUE #103039: Parquet ColumnIndex stats can have min_value > max_value for String columns when only one side exceeds max_statistics_size
### Describe the bug

When writing Parquet files with the native (custom) encoder, the per-page ColumnIndex can be written with `min_value` and `max_value` such that `min_value > max_value` for variable-length BYTE_ARRAY (String) columns. On read, this trips the column-index validation and the query fails with:

```
DB::Exception: Statistics have min_value > max_value: {…} > {…}: in column index;
use input_format_parquet_page_filter_push_down=0 to ignore
```

### How to reproduce

Tested on **26.3.3.20** (also reading the source on master, the relevant code is unchanged).

```sql
INSERT INTO FUNCTION file('repro.parquet', Parquet)
SELECT if(number = 0, 'a', repeat('z', 5000)) AS s
FROM numbers(2)
SETTINGS
    output_format_parquet_use_custom_encoder = 1,
    output_format_parquet_write_page_index   = 1,
    output_format_parquet_row_group_size     = 100000;

SELECT * FROM file('repro.parquet', Parquet) WHERE s = 'a';
```

Reader throws `Statistics have min_value > max_value: 'a' > ''`.

### Root cause

In `src/Processors/Formats/Impl/Parquet/Write.cpp`, the per-page `Statistics::get()` for variable-length strings checks min and max against `max_statistics_size` **independently**:

`StatisticsStringRef::get` (lines 221–237):

```cpp
parq::Statistics get(const WriteOptions & options) const {
    parq::Statistics s;
    if (min.ptr == nullptr) return s;
    if (static_cast<size_t>(min.len) <= options.max_statistics_size) {
        s.__set_min_value(...);
        s.__set_is_min_value_exact(true);
    }
    if (static_cast<size_t>(max.len) <= options.max_statistics_size) {
        s.__set_max_value(...);
        s.__set_is_max_value_exact(true);
    }
    return s;
}
```

Same shape in `StatisticsStringCopy::get` (lines 284–300). `max_statistics_size` is hardcoded to 4096 in `Write.h:51`.

When only one side exceeds the limit, the other side of `parq::Statistics` is left as the thrift default empty string. Then in `addPage()` (lines 902–922), the result is pushed into the ColumnIndex unconditionally:

```cpp
parq::Statistics page_stats = page_statistics.get(options);
...
if (options.write_page_index) {
    s.indexes.column_index.min_values.push_back(page_stats.min_value);
    s.indexes.column_index.max_values.push_back(page_stats.max_value);  // empty string
    ...
    s.indexes.column_index.null_pages.push_back(all_null_page);          // false
}
```

Result: a page where, e.g., the actual min is a short string and the actual max is >4 KB ends up in the ColumnIndex with `min_values[i] = \"<short>\"`, `max_values[i] = \"\"`. The reader (`Reader.cpp:1002–1035`, `adjustRangeFromIndexIfNeeded`) trips on `accurateLess(right, left)`.

The same conditional drop happens for chunk-level Statistics too, but it's harmless there because the reader only validates the ColumnIndex.

### Notes

- Bundled Arrow writer is unaffected: `contrib/arrow/cpp/src/parquet/column_writer.cc` calls `ApplyStatSizeLimits`, which drops min/max symmetrically.
- The custom encoder became default in 24.4 (PR #62088), so this only manifests on writes from 24.4+ being read by versions whose reader validates the ColumnIndex.

saurabhojha:
### Problem
Closes https://github.com/ClickHouse/ClickHouse/issues/103039

When writing Parquet files with the custom encoder, the per-page ColumnIndex for variable-length BYTE_ARRAY (String) columns can have min_value > max_value. This happens when only one side of the min/max pair exceeds max_statistics_size (4096 bytes) -- the oversized side is dropped to an empty string while the other is kept, violating the min <= max invariant.

On read, this causes the query to fail:
`DB::Exception: Statistics have min_value > max_value: {…} > {…}: in column index`

### Root Cause
StatisticsStringRef::get() and StatisticsStringCopy::get() check min and max against max_statistics_size independently. When a page has a short min (e.g. "a") and a long max (e.g. 5000 bytes), the max is dropped while the min is kept. The ColumnIndex then contains min_values[i] = "a", max_values[i] = "".

### Fix
Symmetric drop in statistics: Modified StatisticsStringRef::get() and StatisticsStringCopy::get() to only emit min/max when both fit within the size limit. If either exceeds it, neither is written.

Skip invalid ColumnIndex: Added a column_index_valid flag to ColumnChunkIndexes. When a non-null page has its stats dropped (both min and max empty), the flag is set to false. writePageIndex then skips writing the ColumnIndex for that column entirely, since empty bounds on a non-null page would mislead readers into incorrectly skipping the page during predicate pushdown. 

Fix Parquet ColumnIndex stats min_value > max_value for String columns

- [ ] Documentation is not required (bug fix)

### Version info
- Merged into: `26.5.1.131`
- Backported to: `26.4.1.1123`, `26.3.10.46`, `26.2.17.27`, `26.1.12.22`, `25.8.23.11`

### DIFF
--- a/src/Processors/Formats/Impl/Parquet/Write.cpp
+++ b/src/Processors/Formats/Impl/Parquet/Write.cpp
@@ -223,14 +223,12 @@ struct StatisticsStringRef
         parq::Statistics s;
         if (min.ptr == nullptr)
             return s;
-        if (static_cast<size_t>(min.len) <= options.max_statistics_size)
+        if (static_cast<size_t>(min.len) <= options.max_statistics_size
+            && static_cast<size_t>(max.len) <= options.max_statistics_size)
         {
             s.__set_min_value(std::string(reinterpret_cast<const char *>(min.ptr), static_cast<size_t>(min.len)));
-            s.__set_is_min_value_exact(true);
-        }
-        if (static_cast<size_t>(max.len) <= options.max_statistics_size)
-        {
             s.__set_max_value(std::string(reinterpret_cast<const char *>(max.ptr), static_cast<size_t>(max.len)));
+            s.__set_is_min_value_exact(true);
             s.__set_is_max_value_exact(true);
         }
         return s;
@@ -286,14 +284,12 @@ struct StatisticsStringCopy
         parq::Statistics s;
         if (empty)
             return s;
-        if (min.size() <= options.max_statistics_size)
+        if (min.size() <= options.max_statistics_size
+            && max.size() <= options.max_statistics_size)
         {
             s.__set_min_value(std::string(min.data(), min.size()));
-            s.__set_is_min_value_exact(true);
-        }
-        if (max.size() <= options.max_statistics_size)
-        {
             s.__set_max_value(std::string(max.data(), max.size()));
+            s.__set_is_min_value_exact(true);
             s.__set_is_max_value_exact(true);
         }
         return s;
@@ -949,6 +945,10 @@ void writeColumnImpl(
         if (options.write_page_index)
         {
             bool all_null_page = data_count == 0;
+            bool has_stats = page_stats.__isset.min_value && page_stats.__isset.max_value;
+            if (!all_null_page && !has_stats)
+                s.indexes.column_index_valid = false;
+
             s.indexes.column_index.min_values.push_back(page_stats.min_value);
             s.indexes.column_index.max_values.push_back(page_stats.max_value);
             if (has_null_count)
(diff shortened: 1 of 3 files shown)


al13n321 (approved):
Thanks!

saurabhojha:
@al13n321 there seems to be some unrelated test failures. Do i need to rebase and trigger the ci again? I am confused as the clickhouse-gh bot has status Approve and it indicates changes can be cleanly merged.

clickgapai:

**Affected code:** `StatisticsStringRef::get` and `StatisticsStringCopy::get` in src/Processors/Formats/Impl/Parquet/Write.cpp; `writeColumnImpl` flush_page logic; `writePageIndex` — core code present in all supported branches.

**Why:** User-facing bug — reading back a Parquet file written by ClickHouse with a long String column produces an INCORRECT_DATA exception. The Parquet custom-encoder write_page_index path with max_statistics_size has been present in stable releases for years (predates 25.8). All supported stable branches contain the buggy StatisticsStringRef::get and StatisticsStringCopy::get implementations.

PR #103347, opened 2026-04-22
Always build ClickHouse as PIE and emit stack traces as file offsets

nikitamikhaylov:
Drop the conditional `ENABLE_POSITION_INDEPENDENT_BINARY` / `-no-pie` logic and always build as `PIE` via `CMAKE_POSITION_INDEPENDENT_CODE ON` plus explicit `-pie -Wl,-pie`. The previous "disabled" cases (Android, S390X, sanitizers) all required `PIE` anyway.

Under `PIE`, raw frame pointers are `ASLR`-randomized per run and unusable with `addr2line`. Symbolized fatal lines and `system.stack_trace` / `system.trace_log` already subtract the object base; the two bare dumps in `SignalHandlers::onFault` did not. Add a helper `StackTrace::getPhysicalAddress` and use it in the bare traces and in `StorageSystemStackTrace` so every address we print is a stable file offset that `addr2line -e clickhouse <hex>` can resolve directly.

Made-with: Cursor

- Performance Improvement

ClickHouse is now always built as a Position Independent Executable.


nikitamikhaylov:
`riscv64`, `musl`, `longaarch64`, `ppc64le` and `riscv64` builds are failing because sysroot contents (libs, archives) are not build as PIC.

nikitamikhaylov:
@groeneai Why does this PR affect tests? Both functional and performance?

groeneai:
@nikitamikhaylov This is a build-only PR, but the things it changes (PIE, stack traces as file offsets, the signal-handler path, and mlock in the perf job) are exactly the runtime paths a few tests probe directly. From CIDB, last 60 days:

**Functional**
- `04052_local_cancel_scalar_subquery`: 32 fails, almost all on this PR (other PRs show 1 each, i.e. unrelated flaky noise). The test cancels a scalar subquery in `clickhouse-local`, which goes through the query-cancellation signal handling and stack-trace collection you rewrote in `SignalHandlers.cpp` and `StackTrace.cpp`.
- `01103_check_cpu_instructions_at_startup` and `03590_simdjson_fallback`: 10 fails each on this PR (check `amd_binary_excluded_from_llvm`). Both probe binary startup, CPU-feature detection, and the SIMD fallback, which the PIE switch affects. Note these two also show 5 fails each on #103706 (CheSema, ReaderExecutor), so they are not fully isolated to your branch, but the dominant cluster is here.

**Performance**
Not a query-execution regression. The cloud-benchmark report on this PR returned `no_change` (the 3 flagged TPC-H queries swing in opposite directions and were attributed to variance). What changed is the perf job itself: `ci/defs/job_configs.py` now runs the perf container with full Docker `--privileged` to get `mlock`/`CAP_IPC_LOCK`, and the perf config keeps `remap_executable` enabled (now a `#if 0` stub under PIE). So the perf job is affected by the CI-config and mlock change, not by query speed. The AI review on this PR already suggested narrowing `--privileged` to `--cap-add=IPC_LOCK`.

Happy to share the per-test CIDB rows or the failing report URLs if useful.

nikitamikhaylov:
@groeneai reimplement this PR and take it over. Address all the performance problems.

groeneai:
Acknowledged, taking it over. I have no push access to `always-pie-stack-traces-offsets`, so I will open my own PR from my fork carrying the change and link this one from it.

On the performance side, my reading is that the problem is not query speed (the cloud benchmark here returned `no_change`) but the executable remap path: `src/Common/remapExecutable.cpp` is now gated on `#if 0`, which drops the hugepage text remap unconditionally on every platform, not only where `PIE` forces it, while `tests/performance/scripts/config/config.d/zzz-perf-comparison-tweaks-config.xml` still sets `remap_executable` and `Server.cpp` still logs the remap as successful. So I will work out whether the remap can be made `PIE`-compatible or has to be rejected explicitly, measure what `PIE` itself costs per architecture rather than assuming it, and narrow the perf job's `--privileged` down to the capability it actually needs.

If you meant something else by the performance problems, say so and I will retarget.

groeneai:
Reimplemented against master as #117449, which links this PR. Two commits, one per half, so the PIE change can be dropped on its own.

Answering the three things I said I would work out:

1. The remap does not have to be rejected. `remapExecutable` was made PIE-safe by `6c5298f9b41d8b0` (Raul Marin, 2026-07-09), so the `#if 0` stub here is obsolete: a local `amd_release` PIE ThinLTO build still remaps 453.70 MiB of text into huge pages.
2. PIE's own cost is the only one of the four performance problems left, and I cannot measure it here: there is no aarch64 host on this machine, and this PR's cloud benchmark returned `no_change` with two TPC-H queries moving in opposite directions. #117449's own `Performance` job runs `amd_release` and `arm_release` against master head, and the comparison is live there because that arm is position independent and master's is not. Please read it before merging.
3. Correcting what I said about `--privileged`: there is nothing to narrow. The remap staying alive is what makes `mlock_executable` unnecessary, and that setting is what needed the capability, so no narrowing is required.

One consequence I did not fold in, disclosed in the description instead: constant binary addresses under ASLR are given up, so `system.crash_log.trace` and `system.error_log.last_error_trace`, which persist raw frame pointers, become run-specific on release builds. That is already the case on every debug and sanitizer build, and #101435 is moving those columns to virtual addresses, so I left that surface alone.

If you meant something else by the performance problems, say so and I will retarget.

PR #103349, opened 2026-04-22
Optimize `LIMIT BY`

nihalzp:
### Core Optimizations:
* Remove per-row hashing from in-order `LIMIT BY`.
* Replace the generic `LIMIT BY` `SipHash` + `HashMap<UInt128, UInt64>` path with typed hash methods from `AggregatedDataVariants`. This reduces memory usage for many common key types and lets `LIMIT BY` benefit from existing optimizations. It also avoids relying on hash equality alone to distinguish keys.
* Improve chunk materialization during transform. Instead of always using `IColumn::filter`, choose the cheapest strategy for the output: reuse the whole chunk if all rows are part of the output, use faster `IColumn::cut` for contiguous output, and fall back to `filter` only when needed.

image

image

image

image

image

- Performance Improvement

Improve performance of `LIMIT BY` queries.

### Version info
- Merged into: `26.6.1.414`


alexey-milovidov:

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -1,135 +1,382 @@
...
+{
+    if (grouping_key_positions.empty())
+        return;
+
+    if (previous_chunk_last_grouping_key_columns.empty())
+    {
+        previous_chunk_last_grouping_key_columns.reserve(grouping_key_positions.size());
+        for (size_t position : grouping_key_positions)
+            previous_chunk_last_grouping_key_columns.emplace_back(chunk_columns[position]->cloneEmpty());
+    }
```
Fgrtue: Why did you decide to put the construction into the execution part instead of constructor?

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -1,135 +1,382 @@
...
-            if (isColumnConst(*column))
-                column = column->cut(0, inserted_count);
-            else
-                column = column->filter(filter, inserted_count);
+        if (hasSameGroupingKeyAsPreviousRow(chunk_columns, row_idx))
+            continue;
+
+        /// The grouping key changed within the chunk, so finish the current run
+        /// and reset the counter before starting the next group's run.
+        processRun(current_run_start_row, row_idx - current_run_start_row);
```
Fgrtue, line 374: What if we don't start `processRun` when the already reached `group_limit_end`? For instance, when we reach the needed amount of rows, we use then a binary search or something to find the next group, if we can. In some cases, when `group_limit_end` is small, and the group size is large, it should give some speedup.

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -1,135 +1,382 @@
...
-    {
-        SipHash hash;
-        for (auto position : key_positions)
-            columns[position]->updateHashWithValue(row, hash);
+    ColumnsHashing::HashMethodContextSettings ctx_settings;
+    ctx_settings.max_threads = 1;
+    hash_method_context = AggregatedDataVariants::createCache(type, ctx_settings);
+}
+
+void LimitByTransform::processRun(UInt64 run_start_row, UInt64 run_row_count, size_t group_idx)
```
Fgrtue, line 182: What if we make the computation short circuit here when we reach `group_limit_end`? Probably not much safe, but still won't hurt

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -1,135 +1,382 @@
...
+    UInt64 current_run_start_row = 0;
+    for (UInt64 row_idx = 1; row_idx < row_count; ++row_idx)
     {
-        for (auto & column : columns)
-            if (isColumnConst(*column))
-                column = column->cut(0, inserted_count);
-            else
-                column = column->filter(filter, inserted_count);
+        if (hasSameGroupingKeyAsPreviousRow(chunk_columns, row_idx))
+            continue;
```
Fgrtue: What do you think about using SIMD in this loop for answering equality question between different rows? Say, we could walk each grouping key column in SIMD fashion, and compute the column of bits answering: "is this row different from the previous one?". Then we could take every bit at each row, and if all columns in a row are "not different from the previous one", then we are still in the same group. So we could quickly find the needed `group_limit_end` amount. 

The largest part of this is adding a method to IColumn for computing a vector with information about difference between the rows. But we could also do implementation for just some of the columns -- say scalars and strings, and for others just fall back to the classical "comparison" of values.

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -1,135 +1,382 @@
...
+void LimitByTransform::consumeImpl(Method & hash_method, const ColumnRawPtrs & grouping_key_columns, UInt64 row_count)
+{
+    typename Method::State state(grouping_key_columns, data.key_sizes, hash_method_context);
 
-        if (first_row)
+    UInt64 current_run_start_row = 0;
+    size_t current_run_group_idx = 0;
+    for (UInt64 row_idx = 0; row_idx < row_count; ++row_idx)
+    {
+        auto key_emplace_result = state.emplaceKey(hash_method.data, row_idx, *data.aggregates_pool);
```
Fgrtue, line 205: If we implemented that method with SIMD comparison between different rows in a column, we could also use it here, so that we emplace a key only when we are actually dealing with a different grouping key columns. Maybe that's not a large win, but still might be more efficient. Would be interested in learning your opinion on that.

Fgrtue (commented):
The implementation looks good to me. I would like to suggest adding a separate test for it, as we usually do this when we add new features of do improvements -- the test could just show that there are two different limit by pipeline processorts exist now.

Other than that I suggested two improvements: 1) Short circuit execution -- probably not so difficult to add; 2) SIMD computation of delimiters between different rows in columns -- probably more involved as it is supposed to add a new method and might even worth another PR. Would be interested knowing your opinion on those suggestions.

nihalzp:
Thanks for the thorough review!

> The implementation looks good to me. I would like to suggest adding a separate test for it, as we usually do this when we add new features of do improvements -- the test could just show that there are two different limit by pipeline processorts exist now.

Actually, the edited test `tests/queries/0_stateless/03701_limit_by_in_order.reference` shows both the processors `LimitBySortedStreamTransform` and `LimitByTransform`.


> Other than that I suggested two improvements: 1) Short circuit execution -- probably not so difficult to add; 2) SIMD computation of delimiters between different rows in columns -- probably more involved as it is supposed to add a new method and might even worth another PR. Would be interested knowing your opinion on those suggestions.

1) Added the short circuit execution.

2) I think that's a very good idea; worth creating an RFC for it. We also have `DistinctSortedStreamTransform`, aggregation in order, which could also benefit from it. For `LimitByTransform`, the main cost is actually hashing, and having the same key appear consecutively may be rare, so we may not see a significant speedup. But for `LimitBySortedStreamTransform`, this could give quite a bit of speedup; we can also consider binary search to find the equal-key runs and see how it performs. I can see that for low-cardinality keys binary search would perform very well.

src/Processors/Transforms/LimitByTransform.cpp
nihalzp: I initially kept it here for readability as the lazy init is easier to understand the purpose. But I agree doing it inside CTOR is cleaner.

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -1,138 +1,392 @@
...
+LimitBySortedStreamTransform::LimitBySortedStreamTransform(
+    SharedHeader header, UInt64 group_length_, UInt64 group_offset_, const Names & column_names)
+    : ISimpleTransform(header, header, true)
+    , grouping_key_positions(filterNonConstKeys(header, column_names).positions)
+    , group_offset(group_offset_)
+    , group_limit_end(computeGroupLimitEnd(group_length_, group_offset_))
+{
+    previous_chunk_last_grouping_key_columns.reserve(grouping_key_positions.size());
+    for (size_t position : grouping_key_positions)
+        previous_chunk_last_grouping_key_columns.emplace_back(header->getByPosition(position).type->createColumn());
```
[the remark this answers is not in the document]
nihalzp: Without addressing it, the following query will throw logical error:

```sql
DROP TABLE IF EXISTS t;

CREATE TABLE t (k UInt32, v UInt32)
ENGINE = MergeTree ORDER BY (k, v)
SETTINGS ratio_of_defaults_for_sparse_serialization = 0.01;

INSERT INTO t SELECT 0, 0 FROM numbers(10);

SELECT k FROM t ORDER BY k LIMIT 1 BY k;

DROP TABLE t;
```

This is because the type in header is `ColumnVector` but runtime type is `ColumSparse(ColumnVector)`.  

@Fgrtue  As a result, reverted to the lazy init version where we initialize `previous_chunk_last_grouping_key_columns` during execution (option 1).

Option 2 can be done too but we would be paying extra normalization cost without a good reason.

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -1,138 +1,395 @@
...
     }
 
-    if (rows_before_limit_at_least)
-        rows_before_limit_at_least->add(inserted_count);
+    for (size_t key_idx = 0; key_idx < grouping_key_positions.size(); ++key_idx)
+    {
+        auto & previous_chunk_last_grouping_key_column = previous_chunk_last_grouping_key_columns[key_idx];
+        if (!previous_chunk_last_grouping_key_column->empty())
+            previous_chunk_last_grouping_key_column->popBack(1);
+        previous_chunk_last_grouping_key_column->insertFrom(*chunk_columns[grouping_key_positions[key_idx]], row_idx);
```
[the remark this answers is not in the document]
nihalzp: Different chunks cannot have different types.

Fgrtue:
> Actually, the edited test tests/queries/0_stateless/03701_limit_by_in_order.reference shows both the processors LimitBySortedStreamTransform and LimitByTransform.

I saw it, but I wanted to propose another one -- a new one, that will designate the appearance of this new processor. So even though we have the old test, we would also have a new one. But this is just a nit, and not critical, so I don't insist.

Fgrtue:
>  we can also consider binary search to find the equal-key runs and see how it performs.

Do you want to experiment with the binary search in the sorted version in this PR or in a separate one?

nihalzp:
> I saw it, but I wanted to propose another one -- a new one, that will designate the appearance of this new processor. So even though we have the old test, we would also have a new one. But this is just a nit, and not critical, so I don't insist.

Okay. Added `04268_limit_by_explain.sql`.

> Do you want to experiment with the binary search in the sorted version in this PR or in a separate one?

I think it will be better if done in a separate PR since it will be done as a new method of `IColumn` and performance improvement is guaranteed for all cases so need some experimenting.

Fgrtue (approved):
Thank you for adding the test. Looking forward to further updates on ordered case.

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -1,138 +1,395 @@
...
+
+bool LimitBySortedStreamTransform::hasSameGroupingKeyAsPreviousRow(const Columns & chunk_columns, UInt64 row_idx) const
 {
-    /// Just go to the next block if there isn't any new records in the current one.
-    if (!inserted_count)
-        /// SimpleTransform will skip it.
+    for (size_t position : grouping_key_positions)
+    {
+        const auto & column = *chunk_columns[position];
+        if (column.compareAt(row_idx, row_idx - 1, column, 1) != 0)
```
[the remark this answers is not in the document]
nihalzp: See `tests/queries/0_stateless/04312_in_order_aggregation_inconsistency.sql`.

nihalzp:
@Fgrtue The binary search idea that I discussed:
- https://github.com/ClickHouse/ClickHouse/pull/106502

As expected, it is giving quite of speedup!

PR #93658, opened 2026-01-08
Resubmit 'Skip back up of refreshable materialized view targets' #93076

jkartseva:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Introduced `backup_data_from_refreshable_materialized_view_targets` backup setting to control whether to back up data of refreshable materialized views. Targets of RMVs with APPEND refresh strategy are always backed up.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.1.1.494`
- Backported to: `25.12.4.23`, `25.11.7.38`, `25.10.5.37`

PR #103384, opened 2026-04-22
Make a decision to skip backup of the target RMV table based on snapshot

nikitamikhaylov:
`BackupEntriesCollector::shouldBackupTableData` used to rely on `DatabaseCatalog::instance().getReferentialDependents` to decide whether a table is a target of a refreshable materialized view. That lookup is global and those tables could be gone (EXCHANGE query from the RMV kicks in for example). 

Build a `TablesDependencyGraph` in `gatherTablesMetadata` over the tables being backed up (same pattern as `DatabaseReplicated`) and pass it through `shouldBackupTableData` to
`BackupUtils::isTargetForReplaceRefreshableMaterializedView`, which now uses the graph's `getDependents` instead of the global catalog.

Without this change, this https://github.com/ClickHouse/ClickHouse/pull/93658 seems useless.

- Improvement

Improve the resilience of backups when Refreshable Materialized Views are constantly refreshed by making decisions based on the snapshotted tables' state rather than a global one. 

### Version info
- Backported to: `26.4.3.7`, `26.3.11.8`


vitlibar (approved):

src/Backups/BackupUtils.cpp
vitlibar: Probably we could avoid getting any storages from `DatabaseCatalog` at this point and always use snapshot.

vitlibar: Like this:

```
void gatherTablesMetadata()
{
    for (...)
    {
           ...
           res_table_info.data_path_in_backup = data_path_in_backup;

           if (const auto * mv = typeid_cast<const StorageMaterializedView *>(storage.get()))
                if (mv->isRefreshable() && !mv->isAppendRefreshStrategy())
                    rmv_replace_target_ids.insert(mv->getTargetTableId());
    }

    for (auto & [table_name, table_info] : table_infos)
    {
        table_info.should_backup_data = shouldBackupTableData(
            table_name, table_info.storage, rmv_replace_target_ids);
    }
}
```

nikitamikhaylov: Yes, this one I missed.

PR #103385, opened 2026-04-22
CI: add `force_success` and `allow_failure` fields to `Job.Config`

leshikus:
Previously, making a CI job show green in the report despite test failures required two separate, coupled mechanisms:

1. Passing `--set-status-success` in the job command to force `SUCCESS` at the script level.
2. Setting `allow_merge_on_failure=True` on the job config to prevent the \"Ready For Merge\" check from being blocked.

This split the intent across the job script and the job config, and required every job author to know about both.

## Changes

Replace `allow_merge_on_failure` and `report_as_success_on_failure` with two clearer `Job.Config` fields:

- `allow_failure: bool = False` — the job still shows as failed in the report, but does not block PR merge.
- `force_success: bool = False` — the runner overrides the job result status to `OK` after the job script finishes, regardless of the actual outcome. The report shows the job as green, and it naturally does not block merge.

The override happens in `Runner._run()`, after reading the result written by the job script. The runner also exits 0 when `force_success=True`. Both fields are excluded from job cache digests.

`darwin_fast_test_jobs` is updated to use `force_success=True` and the `--set-status-success` flag is removed from `fast_test.py`.

Also fixes a latent bug where the code referenced `Result.Status.SUCCESS` (which does not exist — the correct value is `Result.Status.OK`), which would have raised an exception in `_run` when `force_success=True` on a failing job.

### Version info
- Merged into: `26.5.1.47`

### DIFF
--- a/ci/praktika/job.py
+++ b/ci/praktika/job.py
@@ -70,7 +70,14 @@ class Job:
 
         run_unless_cancelled: bool = False
 
-        allow_merge_on_failure: bool = False
+        # If True, the job failure does not block PR merge, but the job
+        # is still shown as failed in the CI report.
+        allow_failure: bool = False
+
+        # If True, the job failure is hidden entirely: the CI report shows
+        # green status and the job does not block PR merge. Use for
+        # experimental jobs that are not yet stable enough to be enforced.
+        force_success: bool = False
 
         enable_commit_status: bool = False
 
@@ -233,9 +240,9 @@ class Job:
             res.provides = provides_res
             return res
 
-        def set_allow_merge_on_failure(self, value=True):
+        def set_allow_failure(self, value=True):
             res = copy.deepcopy(self)
-            res.allow_merge_on_failure = value
+            res.allow_failure = value
             return res
 
         def set_post_hooks(self, post_hooks):
--- a/ci/praktika/runner.py
+++ b/ci/praktika/runner.py
@@ -573,6 +573,9 @@ class Runner:
 
         result.update_duration()
         result.set_files([Settings.RUN_LOG], strict=False)
+        if job.force_success and not result.is_ok():
+            print(f"NOTE: Job has force_success=True - overriding status to OK")
+            result.set_status(Result.Status.OK)
         return result
 
     def _post_run(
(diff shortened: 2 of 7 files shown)


ci/praktika/job.py
```diff
@@ -71,6 +71,11 @@ class Config:
 
         allow_merge_on_failure: bool = False
 
+        # If True, the job result is reported as success regardless of the actual
+        # outcome: the report shows green and the job does not block PR merge.
+        # Use instead of --set-status-success in the command.
+        report_as_success_on_failure: bool = False
+
```
maxknv: I suggest this naming and comments
```
# If True, the job failure does not block PR merge, but the job                                                                    
# is still shown as failed in the CI report.                                                                                       
allow_merge_on_failure: bool = False                                                                                               
                                                                                                                                     
# If True, the job failure is hidden entirely: the CI report shows                                                                 
# green status and the job does not block PR merge. Use for                                                                        
# experimental jobs that are not yet stable enough to be enforced.                                                                 
enable_soft_fail: bool = False
```

maxknv (approved):
LGTM, with the naming fix, see comment

ci/praktika/job.py
leshikus: `enable_soft_fail` is incorrectly understood by AI, please take a look at the image; AI believes this name means the job status is marked as failure

Screenshot 2026-04-23 at 01 28 41

maxknv: What do you think about this then?
```
# If True, the job failure does not block PR merge, but the job
# is still shown as failed in the CI report.
allow_failure: bool = False

# If True, the job failure is hidden entirely: the CI report shows
# green status and the job does not block PR merge. Use for
# experimental jobs that are not yet stable enough to be enforced.
force_success: bool = False
```

ci/praktika/runner.py
```diff
@@ -1042,5 +1044,5 @@ def run(
 
             result.dump()
 
-        if not res:
+        if not res and not job.force_success:
```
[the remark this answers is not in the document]
leshikus: that's ok

PR #103401, opened 2026-04-22
CI: fix three bugs in the job-filtering algorithm

maxknv:
Three independent bugs in `check_affected_jobs` caused wrong jobs to run (or not run) in CI.

The bugs were found by analysing PR https://github.com/ClickHouse/clickhouse-private/pull/56394, where `Docker server` failed with `FileNotFoundError: artifact_report_build_arm_release.json` because `Build (arm_release)` was incorrectly filtered out.

**Bug 1 — fnmatch `*` crosses `/` directory boundaries (`ci/praktika/job.py`)**
`fnmatch.fnmatch` treats `*` as matching any character including `/`, so the include path `tests/*.txt` matched `tests/ci/lambda/requirements.txt`, incorrectly marking every stateless-test, stress-test, and cloud-test job as affected by unrelated lambda changes. Fixed by replacing it with `PurePosixPath.match`, which restricts `*` to a single path component.

**Bug 2 — hook-filtered jobs polluted `all_required_artifacts` (`ci/praktika/native_jobs.py`)**
Jobs already filtered by a `workflow_filter_hook` (e.g. "do not run cloud tests") were still processed by `check_affected_jobs`. Their `requires` entries were added to `all_required_artifacts`, keeping their dependencies alive even though the jobs would never run. Fixed by skipping jobs already present in `workflow_config.filtered_jobs`.

**Bug 3 — transitive dependency rescue not propagated (`ci/praktika/native_jobs.py`)**
When an unaffected job was rescued from filtering because an affected job required it, that rescued job's own dependencies were not also rescued. Fixed by propagating `all_required_artifacts` to a fixed point after the initial sweep.

The core logic is extracted into a module-level `_filter_unaffected_jobs` function and covered by 10 new tests in `ci/tests/test_job_filtering.py`.

### Version info
- Merged into: `26.5.1.45`

### DIFF
--- a/ci/praktika/job.py
+++ b/ci/praktika/job.py
@@ -4,7 +4,7 @@ import hashlib
 import json
 import os
 from dataclasses import dataclass, field
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import Any, Iterable, List, Optional
 
 from . import Artifact
@@ -274,7 +274,7 @@ class Job:
                     # Check if included
                     for include in self.digest_config.include_paths:
                         include_norm = os.path.normpath(include)
-                        if fnmatch.fnmatch(file, include_norm) or file.startswith(
+                        if PurePosixPath("/" + file).match("/" + include_norm) or file.startswith(
                             include_norm + os.sep
                         ):
                             return True
--- a/ci/praktika/native_jobs.py
+++ b/ci/praktika/native_jobs.py
@@ -284,6 +284,88 @@ def _prepare_submodule_cache(workflow_config: RunConfig) -> Result:
     )
 
 
+def _filter_unaffected_jobs(jobs, workflow_config, changed_files, affected_dockers=()):
+    """
+    Update workflow_config.filtered_jobs for jobs unaffected by changed_files.
+
+    Two fixes relative to the original inline algorithm:
+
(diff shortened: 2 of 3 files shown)


leshikus:
I wonder if we have an upper level job.is_filtered() getter and put all filtering logic inside? this would make it simple to disable filtering in one place

maxknv:
> I wonder if we have an upper level job.is_filtered() getter and put all filtering logic inside? this would make it simple to disable filtering in one place

The filtering outcome is actually determined in multiple stages rather than in one place, which is why a single per-job method doing all the logic wouldn't quite fit the design.

A job can be skipped for several distinct reasons:
  1. Praktika finds results in the cache — so the job was already run and artifacts are available.
  2. The job is not affected by the current changes and its artifacts are not needed downstream.
  3. A custom user hook explicitly skips the job.

All of this logic runs during workflow configuration, which orchestrates these stages in a batch-oriented pass over the whole workflow graph. After that phase completes, the "is filtered" state on a job becomes a simple read-only property reflecting the outcome — not the algorithm itself.

So the right place to understand or disable filtering is the Workflow Config phase (and its sub-stages), not a single per-job predicate that would need to know about caches, dependency graphs, and user hooks all at once. Collapsing that into one place would actually make the logic harder to reason about, since some decisions (e.g. which jobs are needed as transitive dependencies) are inherently graph-wide.

PR #103402, opened 2026-04-22
Fix for #101423 GRANT SELECT(column) does not give access to its subcolumns

BoloniniD:

Fixes the situation when GRANT SELECT(column) gives access only to the column, but not its nested subcolumns. Closes #101423.


pufit:
I don't think this solution is correct. It isn't general and misses some other critical paths with the same problem (like StorageMerge)

BoloniniD:
@pufit - I think this is much more complete now and ready for review when you get a chance.

Previously, column-level grants were stored against top-level storage columns, but a query referencing a subcolumn (`t.a`, `json.a.b`, `arr.size0`, etc.) would reach the access check under the subcolumn name, making `GRANT SELECT(t)` wrongly insufficient. This fix adds a type-aware helper, `ColumnsDescription::getColumnNamesInStorageForAccessCheck`, that maps requested names back to the storage column identifiers actually being read. It handles regular and dynamic subcolumns, lets real columns resolve to themselves, and leaves unresolved names unchanged so they fail closed.

This is rolled out to every column-level `SELECT` check: both analyzers, `joinGet`, `Alias`, `Buffer`, `Merge`, `MaterializedView`, and the MergeTree index-introspection functions. For `Alias`/`Buffer`/`Merge`, the check now runs against the locked target schema to close a TOCTOU window on concurrent `ALTER`s.

It also adds row-policy protection to `mergeTreeIndex`, `mergeTreeTextIndex`, `mergeTreeAnalyzeIndexes`, and `EXPLAIN … WHATIF`. These had a column-level check but no row-policy awareness, so they could leak granule/token structure for rows a policy is meant to hide. Reads now fail closed on any effective `SELECT` row policy (applied as a blanket rule, since `USING 0` hides rows without naming a column). This part is orthogonal to subcolumns - I kept it here because it completes the access-control story for those functions, but happy to split it out if you'd rather review it separately.

pufit:
@groeneai please review this PR

groeneai (commented):
Reviewed. The approach looks right to me, and it is general: I enumerated every column-level `checkAccess(AccessType::SELECT, ...)` on current master and this PR covers all of them. The one it leaves out, `StorageMergeTreeCodecBlockCounts::readImpl`, is correctly left out, since it passes `getNamesOfPhysical()` and those names are top-level by construction.

Three things I would fix before merging.

**1. It does not merge.** `mergeable` is `CONFLICTING`. The merge base is `7d92b197cc9a` (2026-07-08), and two of the conflicts change what the patch should say:

* `Planner/PlannerJoinTree.cpp`: `checkAccessRights` has been refactored. Master's signature is `checkAccessRights(const StoragePtr &, const StorageID &, const StorageSnapshotPtr &, const Names &, const ContextPtr &)`; the patch edits the old `const TableNode &` form.
* `StorageMergeTreeIndex`: the blanket row-policy rejection is **already on master** (`row_policy_filter && !row_policy_filter->isAlwaysTrue()` throwing `ACCESS_DENIED`). Verified on master `26.9.1.1`: with `CREATE ROW POLICY p ON t USING k < 10 TO u`, `SELECT count() FROM mergeTreeIndex(default, t)` is already denied for `u`. That hunk is now a duplicate with a reworded message.

**2. Yes, split the row-policy half out.** After a master merge the accounting is: `mergeTreeIndex` already upstream, `mergeTreeTextIndex` a *widening* of an existing column-overlap check, `mergeTreeAnalyzeIndexes` and `EXPLAIN ... WHATIF` genuinely new. So the follow-up is 3 sites and this PR drops 4 test files. The widening rationale is sound, since `USING 0` hides rows without naming a column and it makes the text index consistent with `mergeTreeIndex`, but it is a new denial for setups that work today, so it wants its own review rather than riding along with a subcolumn-grant fix.

**3. The TOCTOU comments claim more than the lock gives.** `IStorage::lockForShare` takes `drop_lock` in read mode; `ALTER` serializes on `alter_lock` via `lockForAlter`. Hoisting `lockForShare` above the check therefore does not stop a concurrent `ALTER TABLE ... ADD COLUMN` from rebinding a dotted identifier. What does make the check consistent is reusing one metadata object for both the check and the read, which `StorageAlias::read` and `StorageBuffer::read` do. `StorageMerge` does not: `getSelectedTables` builds `child_metadata_snapshot` only for the check and drops it, while `createChildrenPlans` calls `storage->getInMemoryMetadataPtr(context, false)` again per child for the read. Either drop the `ALTER` claim from that comment, or carry the snapshot from `getSelectedTables` into the read.

That hoist has a side effect too: the share lock is now taken on every child that passes `SHOW_TABLES`, ahead of the `SELECT` check. A user holding only `SHOW` on a child now locks it, and can get `TABLE_IS_DROPPED` or a lock-acquire timeout where they previously got `ACCESS_DENIED`.

**The open bot finding on `StorageMerge` is already satisfied.** `04511_merge_table_subcolumn_access_per_child.sh` is exactly the heterogeneous-child test it asks for: `src_tuple` with `a Tuple(b UInt8)`, `src_real` with a real column named `a.b`, both behind one `Merge`, with the grants swapped in each direction. It does redden. On master, granting only `SELECT(a)` on both children and reading `a.b` through the `Merge` table gives `ACCESS_DENIED`. Worth one line in that test saying why it needs two children, because the two shapes cannot coexist in a single `MergeTree` table:

```
CREATE TABLE t (a Tuple(b UInt8), `a.b` UInt8) ENGINE = MergeTree ORDER BY tuple()

Code: 36. Columns 'a.b UInt8' and 'a Tuple(b UInt8)' have streams (a%2Eb and a%2Eb)
with collision in file name a%2Eb
```

**I checked the helper's central safety claim and it holds.** Whether this can widen a grant comes down to the comment about not falling back to the longest existing dotted prefix, so I built both directions on master `26.9.1.1`:

```
CREATE TABLE j (json JSON, `json.a` JSON) ENGINE = MergeTree ORDER BY tuple();
INSERT INTO j SELECT '{"a":{"b":111}}'::JSON, '{"b":222}'::JSON;
SELECT json.a.b FROM j;   -- 111, with the analyzer both on and off
```

The value comes from `json`, not from `json.a`. `Nested::getAllColumnAndSubcolumnPairs` yields prefixes shortest-first, and both `tryGetDynamicSubcolumn` and `TableExpressionData::tryGetSubcolumnInfo` take the first match, so mapping to `json` is what the read does; a longest-prefix fallback would have authorized `json.a` for a read of `json`. The non-dynamic direction agrees as well: with `a Tuple(b UInt8)` and a real column named `a.b` in one `Memory` table, selecting `a.b` reads the real column, and `tryGetColumn` tests the exact name before any subcolumn lookup, so the helper returns it unchanged.

One style note: several block comments carry PR-description material, notably the 18 lines in `StorageMaterializedView::readImpl` that are mostly scope justification, which reads better in the commit message. The `getColumnNamesInStorageForAccessCheck` comment is the opposite case and should stay as written, since it explains why a property holds rather than why the fix exists.

groeneai:
Following up on my review with something that is not recorded on this PR or on #101423: this patch also closes a hole in the opposite direction, so it is a security fix and not only a usability one.

`GRANT SELECT ON t` plus `REVOKE SELECT(m)` for a `Map` column `m` denies `SELECT m` and allows `SELECT m.keys`. It is the same missing normalisation with the opposite sign. In `AccessRights` a non-wildcard revoke calls `getLeaf(name, COLUMN_LEVEL, return_parent_node = false)`, which creates a radix prefix node for `m` carrying the flags inherited from the table level plus an empty-named leaf beneath it, and `removeGrantsRec` then clears `SELECT` on that leaf only (`src/Access/AccessRights.cpp:713-776`, `:848-857`, `:894-901`). A check of `m.keys` walks into node `m`, does not match the empty-named leaf, and fabricates a child that inherits node `m`'s still-granted flags. Only the wildcard form, which `AccessRights` renders as `REVOKE SELECT(m*)`, writes the prefix node itself, and that over-revokes by byte prefix. After your normalisation the check asks for `m`, the already-denied parent query, so both directions close at once. That is another argument for landing the subcolumn half without waiting on the row-policy half.

One thing I would decide explicitly before merging. `getColumnNamesInStorageForAccessCheck` replaces the requested name, so a grant or revoke written against a subcolumn name is no longer consulted. `` GRANT SELECT(`m.keys`) `` parses today (`parseColumnNames` uses `ParserIdentifier`, which accepts a quoted identifier containing a dot), is stored verbatim with no existence check, and the trie consults it. After this patch the check asks only for `m`, so such a grant stops working and "may read the map's keys but not its values" becomes inexpressible. The mirror revoke case matters much less, because `SELECT tp` already returns the whole tuple that a `` REVOKE SELECT(`tp.secret`) `` is meant to hide. A dotted real column, such as a flattened `Nested` member, resolves to itself and is unaffected. Either an exact-name first pass with a parent fallback, or a sentence in the description saying subcolumn-granularity grants were never intended, would settle it. No test uses a backticked dotted column name in a grant, so CI will not answer this for you.

On coverage: in `04104_grant_select_subcolumn_access.sh` the user has no table-level `SELECT`, so after `REVOKE SELECT(t)` there is no grant on `t` left at all, and that arm proves absence-denies rather than partial-revoke-enforced. Adding `GRANT SELECT ON t`, `REVOKE SELECT(m)`, `SELECT m.keys` denied would pin the case above; of the seven test files on master containing `REVOKE SELECT(`, none reads a subcolumn.

Finally, one case this cannot cover. `REVOKE SELECT(nt)` on `nt Nested(public String, secret String)` has no effect: with `flatten_nested = 1` the physical columns are `nt.public` and `nt.secret`, bare `nt` is expanded into those two during identifier resolution, and a column-level grant naming something that is not a column of the table is accepted silently. Normalisation cannot help, because `nt.secret` resolves to itself, so `` REVOKE SELECT(`nt.secret`) `` is what an administrator has to write. @pufit, would you rather that were handled here or kept separate?

Issue #102803, opened 2026-04-15
Logical error: Not-ready Set is passed as the second argument for function 'A (STID: 0250-4e52)


antaljanosbenjamin:
_Important: This issue was automatically generated and is used by CI for matching failures. DO NOT modify the body content. DO NOT remove labels._

Test name: Logical error: Not-ready Set is passed as the second argument for function 'A (STID: 0250-4e52)
CI report: [Stress test (arm_debug)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102243&sha=2d7a56f0e0b45ac518ba54e611335ae5f3cde633&name_0=PR&name_1=Stress%20test%20%28arm_debug%29)
[...]

PR #100139, opened 2026-03-19
Add parallel_replicas_prefer_local_replica setting

alexey-milovidov:
Add `parallel_replicas_prefer_local_replica` setting that, when disabled, allows replicas for parallel reading to be selected purely by the load balancing algorithm without forcing the local replica into the set. This enables queries with `max_parallel_replicas = 1` to be directed to another host, which can improve cache locality when many short queries are distributed across a cluster.

Closes https://github.com/ClickHouse/ClickHouse/issues/79335

### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Add `parallel_replicas_prefer_local_replica` setting: when disabled, parallel replicas are selected purely by the load balancing algorithm, allowing even `max_parallel_replicas = 1` queries to be directed to another host.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

New setting `parallel_replicas_prefer_local_replica` (Bool, default `true`):
- When enabled (default), the local replica is always included in the set of replicas used for parallel reading (existing behavior).
- When disabled, the local replica is not given any preference and replicas are selected purely by the `load_balancing` algorithm. This allows queries with `max_parallel_replicas = 1` to be directed to another host, which can improve cache locality when many different short queries are distributed across a cluster.

The setting is documented in the auto-generated session settings reference via its `DECLARE` block in `src/Core/Settings.cpp`, which populates `docs/en/operations/settings/settings.md`.

Example:
```sql
SELECT count() FROM my_table
SETTINGS
    enable_parallel_replicas = 1,
    max_parallel_replicas = 1,
    cluster_for_parallel_replicas = 'my_cluster',
    parallel_replicas_prefer_local_replica = 0;
```

### Version info
- Merged into: `26.5.1.92`

PR #103404, opened 2026-04-22
Bound CustomSeparated fields-per-row in header/schema detection

groeneai:
A fuzzer hit on `Stress test (experimental, serverfuzz, arm_msan)` flagged a 12 GB MemorySanitizer allocation inside `CustomSeparatedFormatReader::readRowImpl` (STID `4260-52c8`, master commit `7b48c1bd31ece9f7af46100f1c07c58577ba7f87`, 2026-04-22 20:45 UTC):

```
ERROR: MemorySanitizer: requested allocation size 0x300000000
exceeds maximum supported size of 0x200000000
  #7  ...::__emplace_back_slow_path
  #12 DB::CustomSeparatedFormatReader::readRowImpl<ReadFieldMode::2>()
      src/Processors/Formats/Impl/CustomSeparatedRowInputFormat.cpp:214
  #13 ...::readRowForHeaderDetection()
  #14 DB::RowInputFormatWithNamesAndTypes<...>::tryDetectHeader
  #16 DB::CustomSeparatedRowInputFormat::readPrefix()
```

### Root cause

`readRowImpl` accumulates fields into `std::vector<String> values` with:

```cpp
do {
    values.push_back(readFieldIntoString<mode>(values.empty(), false, true));
} while (!checkForEndOfRow());
```

When `format_custom_row_after_delimiter` does not appear in the input — either malformed data or a fuzzer-crafted payload — `checkForEndOfRow` never returns `true`. The vector grows geometrically until an allocation crosses MSan's 8 GB cap (`0x300000000` is the allocation request for the next doubling; the previous capacity already held ≈250 M `std::string` slots).

Reproduced locally (release debug build) with a 100 MB CustomSeparated input of 50 million short fields and no `row_after_delimiter` — peak RSS reached **3.1 GiB** in ~7 s. Under MSan's smaller cap the same shape trips the assertion earlier.

The same `readRowImpl` unknown-columns branch is reached by every call site that reads a row without a known column count: `readRowForHeaderDetection` (this crash), `readNames` / `readTypes` (WithNames / WithNamesAndTypes variants), and `CustomSeparatedSchemaReader::readRow`. All of them share the vulnerability.

### Fix

Hard-code a per-row field cap inside `CustomSeparatedRowInputFormat.cpp`:

```cpp
constexpr size_t MAX_FIELDS_PER_ROW = 1'000'000;
```

`readRowImpl` checks `values.size() >= MAX_FIELDS_PER_ROW` at the top of the loop and fails fast with `INCORRECT_DATA`. The cap covers every legitimate CustomSeparated schema by a comfortable margin and bounds the worst-case allocation at well under 30 MB on the parser's hot path. No user-facing setting — per @Avogar's review feedback, there is no realistic use case for tuning this number.

### Test

Regression test at `tests/queries/0_stateless/04110_custom_separated_max_fields_bound.sh`:

1. **`bounded`** — header detection path (`tryDetectHeader -> readRowForHeaderDetection -> readRowImpl<AS_POSSIBLE_STRING>`) fed an adversarial 1 000 001-field input with no `format_custom_row_after_delimiter`. Expects `INCORRECT_DATA`. Without the patch the same input grows `values` unboundedly and is killed by MSan / OOM before reaching the bound.
2. **`well-formed`** — a normal 3-column / 3-row input through the same path. Expects `count() = 3`, demonstrating the cap does not interfere with legitimate use.

`readRowImpl` is the only place the bound is checked, and every caller (header detection, schema inference, `readNames` / `readTypes`, variable-column rows) goes through it, so testing the header-detection caller exercises the protection for the whole family.

### Scope note

TSV and CSV row readers have analogous unbounded do/while loops (`TabSeparatedFormatReader::readRowImpl`, `CSVFormatReader::readRowImpl`), but each of their iterations consumes at least one delimiter byte, so the field count is naturally capped by the input size. The CustomSeparated reader is unique in that its configured delimiters can be multi-character or empty, allowing zero-byte iterations. This PR intentionally targets only the format where the fuzzer demonstrated the issue; the TSV / CSV paths can be hardened separately if desired.

### Failure report

https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=7b48c1bd31ece9f7af46100f1c07c58577ba7f87&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_msan%29

Protect the `CustomSeparated` input format against malformed or adversarial data that omits `format_custom_row_after_delimiter`. Header detection, schema inference, and variable-column rows previously accumulated fields unboundedly and could allocate many gigabytes of memory before failing. Reads now fail with `INCORRECT_DATA` once a single row contains more than 1 000 000 fields.

### Version info
- Merged into: `26.5.1.823`


groeneai:
### Pre-PR validation gate (session `cron:clickhouse-ci-task-worker:20260422-214500`)

**a) Deterministic repro?**
Yes. Locally on this branch's debug binary:

```
printf 'a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a' | clickhouse local \
  --input-format='CustomSeparated' --structure='x String' --table='t' \
  --format_custom_field_delimiter='-' --format_custom_escaping_rule='CSV' \
  --input_format_custom_detect_header=1 \
  --input_format_custom_max_number_of_fields_per_row=10 \
  -q 'select count() from t'
```

Returns `Code: 117. INCORRECT_DATA: Too many fields in a single row of CustomSeparated input (limit: 10)...` deterministically on every invocation.

Without the patch (verified earlier with 50 million `-`-separated fields and `row_after_delimiter='\n'` on the same binary, before the fix was applied): peak RSS **3.1 GiB**, runtime ~7 s. Under MSAN the backing-array doubling at 250 M entries crosses the 8 GB cap — the exact failure mode the CI report captured.

**b) Root cause explained?**
`CustomSeparatedFormatReader::readRowImpl` accumulates fields into `std::vector<String> values` with a `do { values.push_back(readFieldIntoString(...)); } while (!checkForEndOfRow());` loop. When `format_custom_row_after_delimiter` is absent from the input, `checkForEndOfRow` never returns `true`. The vector's backing array doubles geometrically. After ~250 M entries (sizeof(std::string)=24), the next doubling requests 12 GiB — MSAN's cap is 8 GiB, so the allocator rejects it and MSAN emits the reported diagnostic. Without a sanitizer the process OOMs.

Yes. The bound is exactly in the unbounded loop: before each `push_back`, check `values.size() >= max_fields` and throw `INCORRECT_DATA`. The default (`1'000'000`) is high enough that no realistic caller hits it; malformed/adversarial inputs fail fast at a fixed, small memory footprint (≈24 MB for the vector at the bound, which is also when the throw happens). No fuzzer input can exceed the bound without triggering it.

**d) Test intent preserved? / New tests added?**
A new test `tests/queries/0_stateless/04110_custom_separated_max_fields_bound.sh` exercises:
1. Header-detection path at the bound (the path that CI actually crashed on).
2. Schema-inference path at the bound (same `readRowImpl` branch reached via `CustomSeparatedSchemaReader::readRow`).
3. A well-formed 3-column input under the same low cap — confirms legitimate inputs still parse.
4. Implicit default still lets legitimate 3-field input through.

No existing test was relaxed or removed. I re-ran the `clickhouse-local`-only subset of the CustomSeparated tests locally — `02481`, `02521`, `02752`, `03302` all pass with the patched binary. (The `01014`/`02117` tests need a running server and are out of scope for a standalone shell run.)

- **Without the fix:** 50 M `-`-separated fields → 3.1 GiB RSS; MSAN build hits the same code path with the reported 12 GiB allocation.
- **With the fix:** the same input at a `max_number_of_fields_per_row=10` cap throws `INCORRECT_DATA` before the 11th field. Output of the test case exactly matches the reference.

Yes for `CustomSeparatedFormatReader`: the bound sits in `readRowImpl` itself, so it covers every mode — `AS_STRING` (WithNames/WithTypes column headers), `AS_FIELD` (`CustomSeparatedSchemaReader::readRow`), `AS_POSSIBLE_STRING` (the reported header-detection path), and the `allow_variable_number_of_columns` branch. It is not a patch on the single call site.

I also checked the symmetric sibling readers: `TabSeparatedFormatReader::readRowImpl` and `CSVFormatReader::readRowImpl` share the `do { fields.push_back(...); } while (checkChar(delim, *buf));` shape, but their single-character `checkChar` advances the buffer by at least one byte per iteration, so field count is naturally bounded by input size. CustomSeparated is the only reader where multi-character or empty delimiters allow zero-byte iterations. Hardening TSV/CSV is a separate, lower-priority follow-up and intentionally out of scope for this PR to keep review focused.

The fix traces the bad state back to its true source (the unbounded loop), not a symptom — I did not add a defensive check at a downstream allocation site.

groeneai:
cc @Avogar @alexey-milovidov — could you review this? The fix bounds the unbounded `std::vector<String>` growth in `CustomSeparatedFormatReader::readRowImpl` that MSAN caught on master (STID 4260-52c8, 12 GiB allocation). New setting `input_format_custom_max_number_of_fields_per_row` (default 1M) caps per-row field count during header detection and schema inference; reads exceeding the bound fail with `INCORRECT_DATA`.

groeneai:
Rebased onto current master. Branch now tracks master cleanly (was CONFLICTING since ~2026-04-24 00:00 UTC).

Summary of the change:
- Head: `e5aee05a5fd` (merge of current master into fix branch)
- Fix commit unchanged: `273e917ed2d Bound CustomSeparated fields-per-row in header/schema detection`
- Diff vs master: unchanged (7 files, +81/-0). Verified no content loss during merge.

No code changes from the previous revision — only rebasing to resolve the merge conflict so CI can run against current master.

@Avogar @alexey-milovidov — this is a P0 sanitizer fix on master (MSAN 12 GiB allocation, STID 4260-52c8). Would appreciate a review when you have a moment. Pre-PR validation gate is in the comment above. Session: `cron:clickhouse-ci-task-worker:20260424-024500`.

groeneai:
## CI status after rebase — all 4 failures unrelated to this fix

Verified against CIDB for rebased HEAD `e5aee05a5fd`:

| Job | Failure | Attribution |
|---|---|---|
| `Docker keeper image` | `clickhouse/clickhouse-keeper:head-arm64` | **Infrastructure flake** — only 2 hits in 30d (1 on master 2026-04-22, 1 here today). No code touches docker/keeper build. |
| `AST fuzzer (arm_asan_ubsan)` | `STID: 0250-4e52` "Not-ready Set is passed as the second argument" | **Chronic trunk bug** — 23 hits across 20 distinct PRs (3 master hits) in 30 days. Tracked under issue #102803. |
| `Unit tests (msan, function_prop_fuzzer)` | MSAN / UBSAN in `protocol.h`, `Lexer.cpp`, `UTF8Helpers.cpp` etc. | **Chronic trunk bug, NOW FIXED.** External PR #103489 merged 2026-04-24 11:48 UTC. Our CI ran at ~09:47 UTC, before the fix landed. Post-merge master status (12:00 UTC): `asan_ubsan`/`msan`/`tsan` fuzzer variants all 4 OK / 0 fails. |
| `AST fuzzer (amd_debug, targeted, old_compatibility)` | ERROR (no test name) | **Infrastructure flake** — Docker Hub pull timeout in `DockerImage.get_docker_image(IMAGE_NAME).pull_image()`. Today 74 OK / 1 ERROR across 70 PRs. |

### Why none can have been caused by this PR
This PR modifies only `CustomSeparatedRowInputFormat.cpp` + the new `input_format_custom_max_number_of_fields_per_row` setting plumbing. None of the four failing sites execute the CustomSeparated reader:
- Docker build — never runs ClickHouse code at all
- `StID 0250-4e52` — `FunctionIn::prepare` (Set pipeline), no formats code
- `function_prop_fuzzer` — URL scheme parsing + Lexer + UTF-8 primitives, no formats code
- Docker pull ERROR — happens before any ClickHouse code runs

### STID 4260-52c8 master trunk status
**~40 hours clean** since PR opened. Only the original 2026-04-22 20:45 UTC master hit (commit `7b48c1bd`, ~1.5 h before this PR); no recurrences anywhere. Awaiting merge to confirm fix holds at scale.

### Request
@Avogar @alexey-milovidov — could you take another look when you have a moment? Fix summary from the pre-PR validation gate: new `input_format_custom_max_number_of_fields_per_row` setting (default 1'000'000) bounds the `std::vector<String>` growth in `CustomSeparatedFormatReader::readRowImpl` that MSAN caught at 12 GiB on master. Header-detection, schema-inference, and variable-columns paths all share the `readRowImpl` entry point so a single bound covers them. Thanks!

---
_Session: `cron:clickhouse-ci-task-worker:20260424-141500`_

groeneai:
## Rebased onto current master at 2026-04-26 01:50 UTC

cc @Avogar @alexey-milovidov — sorry for the silence; PR went `CONFLICTING/DIRTY` again on master move and only got picked up just now. Fresh state:

- **HEAD:** `f35a72cdc85` (merge of latest `origin/master` into the fix branch — fast-forward push, no force)
- **State:** `MERGEABLE` / `BLOCKED` (was `CONFLICTING`/`DIRTY`)
- **Diff vs master (unchanged):** 7 files, +81/-0
  - `src/Core/FormatFactorySettings.h` — declares `input_format_custom_max_number_of_fields_per_row`
  - `src/Core/SettingsChangesHistory.cpp` — registers the setting under 26.5
  - `src/Formats/FormatFactory.cpp`, `src/Formats/FormatSettings.h` — plumbing
  - `src/Processors/Formats/Impl/CustomSeparatedRowInputFormat.cpp` — bound enforced inside `readRowImpl`
  - `tests/queries/0_stateless/04110_custom_separated_max_fields_bound.{sh,reference}` — regression test
- **Conflict resolution:** SettingsChangesHistory.cpp had a textual collision in the 26.5 block; auto-merge resolved cleanly (entry was added back at the same anchor, alongside `parallel_replicas_prefer_local_replica` from PR #100139). No semantic changes.
- **No contrib/ submodule pointers touched.**
- Fix logic and regression test unchanged from previous review-ready state. `clickhouse-local`-driven smoke runs still pass on `02481_custom_separated_and_template_with_csv_field`, `02521_tsv_csv_custom_header_detection`, `02752_custom_separated_ignore_spaces_bug`, `03302_udf_header_detection`, plus `04110_custom_separated_max_fields_bound` for the bounded paths.

### Master STID 4260-52c8 status

- 0 master hits in the last 24h.
- The originating hit is still the only one in 90d (2026-04-22 20:45 UTC, commit `7b48c1bd31`). The bug surface is rare-firing because it requires malformed CustomSeparated input where `format_custom_row_after_delimiter` never matches AND a header/schema-detection path is in play. But it's user-reachable from any HTTP/file ingestion of untrusted data with `input_format_custom_detect_header=1`, so this should still land.

### Reviewer ask

Could you take another look at this when you get a chance? AI Review previously approved on the rebased HEAD `e5aee05a5fd`; this is the same logical change with master merged in.

Ref: STID 4260-52c8 — `MemorySanitizer: requested allocation size 0x300000000 exceeds maximum supported size of 0x200000000` in `CustomSeparatedFormatReader::readRowImpl` via `std::vector<String>` unbounded growth.

(Session: cron:clickhouse-ci-task-worker:20260426-014500)

src/Core/FormatFactorySettings.h
```diff
@@ -319,6 +319,9 @@ Automatically detect header with names and types in TSV format
 )", 0) \
     DECLARE(Bool, input_format_custom_detect_header, true, R"(
 Automatically detect header with names and types in CustomSeparated format
+)", 0) \
+    DECLARE(UInt64, input_format_custom_max_number_of_fields_per_row, 1'000'000, R"(
```
Avogar: No need for a setting. We can make a hard limit in the code. I don't believe there is a usecase that will need this setting to be changed

Avogar (changes requested):

src/Core/FormatFactorySettings.h
groeneai: Done in the next commit — dropped the user-facing setting and made it a `constexpr size_t MAX_FIELDS_PER_ROW = 1'000'000;` inside `src/Processors/Formats/Impl/CustomSeparatedRowInputFormat.cpp`. Also removed the matching wire-up in `FormatFactory.cpp` and `FormatSettings.h` and the `SettingsChangesHistory.cpp` entry. The regression test now drives the bound by feeding ~1M fields through the header-detection path. PR description updated to match.

groeneai:
@Avogar — addressed in `d6856b2e938`: dropped the user-facing setting and replaced it with a file-local `constexpr size_t MAX_FIELDS_PER_ROW = 1'000'000;` in `src/Processors/Formats/Impl/CustomSeparatedRowInputFormat.cpp`. Also removed the matching `FormatSettings::custom::max_number_of_fields_per_row` member, the `FormatFactory.cpp` wire-up, and the `SettingsChangesHistory.cpp` entry. PR description updated to match.

### Pre-PR validation gate (re-run for the new commit)

| | Check | Status |
|---|---|---|
| a | Deterministic repro? | Yes. `python3 -c 'import sys; sys.stdout.write("a"+("-a"*1000001))' \| clickhouse-local --input-format=CustomSeparated --structure='x String' --table=t --format_custom_field_delimiter='-' --format_custom_escaping_rule='CSV' --input_format_custom_detect_header=1 -q "select count() from t"` fails on every run. |
| b | Root cause explained? | Yes. `readRowImpl` accumulates fields into `std::vector<String> values` until `checkForEndOfRow` returns `true`. When `format_custom_row_after_delimiter` never matches (the original fuzzer payload, or arbitrary malformed input), the loop is unbounded and `values` grows geometrically until an allocation crosses MSan's 8 GiB cap. |
| c | Fix matches root cause? | Yes. The cap is enforced at the top of the loop in `readRowImpl` — the exact line that did the unbounded `push_back`. Same call path that crashed in CI. No band-aid. |
| d | Test intent preserved? | Yes. The regression test exercises the protected path (`readRowForHeaderDetection -> readRowImpl<AS_POSSIBLE_STRING>`) with an adversarial input that would otherwise grow `values` unboundedly. The only change vs. the previous commit is that the test now drives the *hard-coded* bound, not a tunable setting. |
| e | Demonstrated in both directions? | Yes. Local debug build (this commit): bounded subtest throws `Code: 117 ... Too many fields in a single row of CustomSeparated input (limit: 1000000)` in ~0.4 s with peak RSS ~450 MB. Earlier reproduction without the bound: 100 MB / 50 M-field input grew RSS to **3.1 GiB** in ~7 s before any error. Same shape under MSan trips the 8 GiB cap, which is the original CI signature (STID `4260-52c8`). |
| f | Fix is general, not a narrow patch? | Yes. `readRowImpl` is the single entry point for every caller that reads a row with an unknown column count — `readRowForHeaderDetection`, `readNames` / `readTypes` (`WithNames` / `WithNamesAndTypes` variants), `CustomSeparatedSchemaReader::readRow`, and `allow_variable_number_of_columns=1`. The cap covers all of them. Sibling `TabSeparatedFormatReader::readRowImpl` and `CSVFormatReader::readRowImpl` are not affected: their single-character `checkChar` consumes one byte per iteration, so field count is naturally bounded by input size; CustomSeparated is the outlier because its delimiters can be multi-character or empty. |

Inline reply on the original thread: https://github.com/ClickHouse/ClickHouse/pull/103404#discussion_r3229319712

cc @alexey-milovidov FYI — same fix scope, just internalised.

Avogar (approved):

clickgapai:

**Affected code:** `CustomSeparatedFormatReader::readRowImpl` unbounded field accumulation loop — core code present in all supported branches.

**Why this might apply:** The vulnerable code in `CustomSeparatedRowInputFormat.cpp:readRowImpl` has existed since December 2021 (commit 8112a712336). This is a DoS vulnerability (memory exhaustion from adversarial input) that affects all supported branches. The fuzzer-triggered crash (STID 4260-52c8) demonstrates exploitability.

PR #101093, opened 2026-03-29
Support for trailing comma in WITH before SELECT

arujbansal:
Allow a trailing comma in the `WITH` clause before `SELECT` queries, e.g. `WITH 1 AS a, 2 AS b, SELECT a + b`.

The implementation adds an `allow_trailing_separator` parameter to `ParserList` and its `parseUtil` helper, which when enabled keeps the parser position past the trailing comma instead of backtracking.

This resolves #90780.

### Changelog category:
- Improvement

### Changelog entry:
You can now have a trailing comma in the `WITH` clause before a `SELECT` query.

### Documentation entry for user-facing changes
- [x] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.852`

PR #103418, opened 2026-04-23
Fix BSONEachRow rollback crash on malformed nullable array elements

groeneai:

Fix a `LOGICAL_ERROR` "Cannot pop N rows from X" that aborted the server (in debug / sanitizer builds) when reading a malformed BSON document containing a value whose BSON type is incompatible with the target column type inside a `Nullable(T)` or `Array(Nullable(T))` column. The malformed row now produces a clean `ILLEGAL_COLUMN` exception as intended.

---

### Motivation

Originally reproduced as `STID: 2097-3451` on `Stress test (arm_asan_ubsan, s3)` by the server-side AST fuzzer. Small deterministic BSON reproducer (24 bytes):

```sql
SELECT * FROM format(BSONEachRow, 'a Array(Nullable(Int64))',
  x'180000000461001000000002300004000000666f6f000000');
```

On master (debug build) this aborts the server:

```
Logical error: 'Cannot pop 1 rows from Int64: there are only 0 rows'.
...
src/Columns/ColumnVector.h:101:13: DB::ColumnVector<long>::popBack
src/Columns/ColumnNullable.cpp:328:23: DB::ColumnNullable::popBack
src/DataTypes/Serializations/SerializationArray.cpp:686:27: DB::SerializationArray::readArraySafe
src/Processors/Formats/Impl/BSONEachRowRowInputFormat.cpp:392:5: DB::BSONEachRowRowInputFormat::readArray
src/Processors/Formats/Impl/BSONEachRowRowInputFormat.cpp:665:13: DB::BSONEachRowRowInputFormat::readField
```

### Root cause

`BSONEachRowRowInputFormat::readField`'s `TypeIndex::Nullable` case pushed a new entry into the null map *before* recursing to read the nested value:

```cpp
case TypeIndex::Nullable:
{
    auto & nullable_column = assert_cast<ColumnNullable &>(column);
    auto & nested_column = nullable_column.getNestedColumn();
    const auto & nested_type = assert_cast<const DataTypeNullable *>(data_type.get())->getNestedType();
    nullable_column.getNullMapColumn().insertValue(0);   // advances null_map
    return readField(nested_column, nested_type, bson_type);  // may throw
}
```

If the nested read throws (e.g. `readAndInsertInteger` detects an unexpected BSON type and raises `ILLEGAL_COLUMN`), the null map is left one entry ahead of the nested column. Because `ColumnNullable::size()` returns `null_map.size()`, this inflated size then drives the rollback in `SerializationArray::readArraySafe`:

```cpp
if (nested_column.size() > offsets.back())
    nested_column.popBack(nested_column.size() - offsets.back());
```

where `nested_column` here is the outer `Array`'s data, i.e. the `ColumnNullable` itself. `ColumnNullable::popBack(n)` calls `nested_inner.popBack(n)` with `n = 1 + actual_rows`, which raises `LOGICAL_ERROR` and aborts.

### Fix

Reorder the two operations so the null map is only advanced after the nested read has completed successfully:

```cpp
auto result = readField(nested_column, nested_type, bson_type);
nullable_column.getNullMapColumn().insertValue(0);
return result;
```

If the nested read throws, neither side is advanced and the column stays internally consistent. On success both sides advance by one row as before, so the end state is unchanged.

Other `insertValue(0) on null_map` sites in format parsers were audited. `MsgPackRowInputFormat.cpp` and `JSONExtractTree.cpp` already follow the correct ordering (read-then-advance-null-map), so this is the only site with the bug.

### Pre-PR validation gate

(a) Deterministic repro: yes — the 24-byte BSON above crashes master 100% of the time.

(b) Root cause explained: null map advanced before the nested read, leaving `ColumnNullable` with `size()` (= null map size) inflated by 1 when the nested read throws. The outer-array rollback in `SerializationArray::readArraySafe` then computes `n = null_map.size() - offsets.back()` and calls `ColumnNullable::popBack(n)`, which propagates to the nested column which has only `n - 1` rows.

(c) Fix matches root cause: yes — reorders the two operations so the null map is only advanced once the nested read returns. No symptom-level widening, no broad tag changes.

(d) Test intent preserved / new tests added: new test `04112_bson_nullable_array_read_rollback.sql` added. It covers the original `Array(Nullable(Int64))` trigger, a top-level `Nullable(Int64)` trigger, and a positive `{ a: [1,2,3] }` control to show the happy path is unaffected.

(e) Demonstrated in both directions: on master the reproducer aborts the server with the exact stack from CI. With this patch applied the same query returns `Code: 44. DB::Exception: Cannot insert BSON String into column with type Int64 ... (ILLEGAL_COLUMN)` and the server stays alive. The control query returns `[1,2,3]`.

(f) Fix is general: audited other `getNullMapColumn().insertValue(0)` call sites — the bug is confined to BSONEachRow. MsgPack and JSONExtractTree already use the correct read-then-null-map ordering.

### CIDB

STID 2097-3451 distribution in 30 days: 6 hits across 3 unrelated PRs (#100535, #101251, #101093) and 0 master hits. The fuzzer-only visibility is why it's classified as a rare chronic trunk bug; the real impact is any user that feeds untrusted BSON into ClickHouse, which is user-reachable via `file()`/`s3()`/`url()` functions and BSON-formatted INSERT streams.

Originating CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=101251&sha=10b62feffa30447b4bea80f2cbeb1cfd8cdba3d4&name_0=PR&name_1=Stress%20test%20%28arm_asan_ubsan%2C%20s3%29

### Version info
- Merged into: `26.5.1.34`


groeneai:
cc @Avogar @alexey-milovidov — could you review this? It fixes a `LOGICAL_ERROR` abort that fires when `BSONEachRowRowInputFormat` rolls back a partial read of a `Nullable(T)` value. The `null_map` was advanced before the nested read, so a throwing read left `ColumnNullable::size()` one entry ahead of the nested column, and `SerializationArray::readArraySafe`'s rollback tripped a `popBack` assertion. The fix reorders to read-then-advance-null-map, and adds a 24-byte deterministic BSON regression test. Originally caught by server fuzzer as STID 2097-3451.

groeneai:
**Pre-PR validation gate answers** (per my task-worker gate procedure):

**(a) Deterministic repro?** Yes. The 24-byte BSON document `x'180000000461001000000002300004000000666f6f000000'` aborts master in debug/ASan builds 100% of the time, with a stack trace that matches the CI reports for STID 2097-3451 byte-for-byte.

**(b) Root cause explained?** `BSONEachRowRowInputFormat::readField` advances the null map before recursing to read the nested value. When the nested read throws (incompatible BSON type), `null_map.size()` is 1 ahead of the nested column. Because `ColumnNullable::size()` is defined as `null_map.size()`, the outer array's rollback path (`SerializationArray::readArraySafe` → `nested_column.popBack(nested_column.size() - offsets.back())`) computes an N that is 1 larger than what the nested column actually holds, and the nested `popBack` trips a `LOGICAL_ERROR`.

**(c) Fix matches root cause?** Yes. The change is a two-line reorder of the null-map insert to after the nested read returns. No widening of bounds, no tag changes, no catch-and-continue. The change directly eliminates the invariant violation the assertion detects.

**(d) Test intent preserved / new tests added?** New test `04112_bson_nullable_array_read_rollback.sql` with three cases:
- malformed BSON in `Array(Nullable(Int64))` → expects `ILLEGAL_COLUMN`
- malformed BSON in top-level `Nullable(Int64)` → expects `ILLEGAL_COLUMN`
- well-formed BSON `{a:[1,2,3]}` → returns `[1,2,3]` (happy path)

No existing tests are modified or weakened.

**(e) Demonstrated in both directions?** Yes.
- **Without fix:** the malformed-BSON query aborts the server with `Logical error: 'Cannot pop 1 rows from Int64: there are only 0 rows'`, stack trace matching the CI report.
- **With fix:** the same query returns `Code: 44. DB::Exception: Cannot insert BSON String into column with type Int64 ... (ILLEGAL_COLUMN)` and the server stays alive. The happy-path query returns `[1,2,3]` unchanged.

**(f) Fix is general?** Audited every `getNullMapColumn().insertValue(0)` site in `src/`. Only `BSONEachRowRowInputFormat.cpp` had the read-after-advance ordering. `MsgPackRowInputFormat.cpp` (line 99) and `JSONExtractTree.cpp` (line 1202) both use the correct read-then-advance ordering already.

Session: cron:clickhouse-ci-task-worker:20260423-024500

azat (approved):

clickgapai:

**Affected code:** `BSONEachRowRowInputFormat::readField` Nullable case at line 521-534 — core code present in all supported branches.

**Why:** Fix in `BSONEachRowRowInputFormat::readField` — git log shows the file was added in commit 2fff4887ac40 (2022-09-24), predating all supported branches. The LOGICAL_ERROR is reproducible and causes a crash in debug/sanitizer builds and undefined behavior (size_t underflow in `ColumnNullable::popBack`) in release builds when parsing malformed BSON with Nullable columns inside containers.

PR #103421, opened 2026-04-23
Fix flaky 01872_functions_to_subcolumns_analyzer under map-bucket serialization

groeneai:
## Summary

Pin `map_serialization_version` and `map_serialization_version_for_zero_level_parts` to `basic` in the test's `CREATE TABLE SETTINGS` so `m.keys`/`m.values` subcolumns return keys in insertion order regardless of CI randomization.

## Root cause

CI randomizes both settings between `basic` and `with_buckets` via `tests/clickhouse-test` `MergeTreeSettingsRandomizer` (lines 1574-1578). With `with_buckets`, Map entries are stored in hash-bucket order on disk, so reads of `m.keys`/`m.values` subcolumns (produced by `optimize_functions_to_subcolumns` rewriting `mapKeys(m)`/`mapValues(m)`) return keys in bucket order rather than insertion order. The reference expects insertion order, so the test fails whenever the randomizer picks `with_buckets`.

The test's purpose is to verify the `optimize_functions_to_subcolumns` rewrite — this is still verified via the `EXPLAIN QUERY TREE` assertions, which are orthogonal to the storage format. Pinning only the serialization format preserves the test's intent.

Deterministic reproducer:
```
CREATE TABLE t (id UInt64, m Map(String, UInt64)) ENGINE = MergeTree ORDER BY tuple()
  SETTINGS map_serialization_version_for_zero_level_parts = 'with_buckets',
           max_buckets_in_map = 39,
           map_buckets_min_avg_size = 1;
INSERT INTO t VALUES (1, map('foo', 1, 'bar', 2));
SELECT mapKeys(m), mapValues(m) FROM t
  SETTINGS enable_analyzer = 1, optimize_functions_to_subcolumns = 1;
-- returns ['bar','foo'] [2,1]
```

Same pattern as the earlier fix for `02974_analyzer_array_join_subcolumn.sql` (commit `d238b482d5c`).

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=103290&sha=19fb213a58678837108547442bccac2efe611e9e&name_0=PR&name_1=Stateless%20tests%20%28amd_debug%2C%20parallel%29

30-day CIDB evidence: 5 unrelated PRs hit this failure (after PR-relatedness filter), 0 master failures — genuine chronic flaky test.

...

### Version info
- Merged into: `26.5.1.38`

### DIFF
--- a/tests/queries/0_stateless/01872_functions_to_subcolumns_analyzer.sql
+++ b/tests/queries/0_stateless/01872_functions_to_subcolumns_analyzer.sql
@@ -3,8 +3,15 @@ DROP TABLE IF EXISTS t_func_to_subcolumns;
 SET enable_analyzer = 1;
 SET optimize_functions_to_subcolumns = 1;
 
+-- Pin Map serialization to `basic` so `m.keys`/`m.values` subcolumns return keys in
+-- insertion order. CI randomizes `map_serialization_version_for_zero_level_parts` and
+-- `map_serialization_version` between `basic` and `with_buckets`; the latter stores keys
+-- in hash-bucket order, which reorders `mapKeys(m)`/`mapValues(m)` output. The test
+-- verifies the `optimize_functions_to_subcolumns` rewrite (see EXPLAIN QUERY TREE
+-- assertions below), which is orthogonal to storage format.
 CREATE TABLE t_func_to_subcolumns (id UInt64, arr Array(UInt64), n Nullable(String), m Map(String, UInt64))
-ENGINE = MergeTree ORDER BY tuple();
+ENGINE = MergeTree ORDER BY tuple()
+SETTINGS map_serialization_version = 'basic', map_serialization_version_for_zero_level_parts = 'basic';
 
 INSERT INTO t_func_to_subcolumns VALUES (1, [1, 2, 3], 'abc', map('foo', 1, 'bar', 2)) (2, [], NULL, map());
 


groeneai:
## Pre-PR validation gate (session cron:clickhouse-ci-task-worker:20260423-044500)

```
CREATE TABLE t (id UInt64, m Map(String, UInt64)) ENGINE = MergeTree ORDER BY tuple()
  SETTINGS map_serialization_version_for_zero_level_parts = 'with_buckets',
           max_buckets_in_map = 39,
           map_buckets_min_avg_size = 1;
INSERT INTO t VALUES (1, map('foo', 1, 'bar', 2));
SELECT mapKeys(m), mapValues(m) FROM t
  SETTINGS enable_analyzer = 1, optimize_functions_to_subcolumns = 1;
-- prints ['bar','foo'] [2,1] before the fix, ['foo','bar'] [1,2] after.
```

**b) Root cause explained?** Yes. `map_serialization_version_for_zero_level_parts = 'with_buckets'` stores Map entries in hash-bucket order. `optimize_functions_to_subcolumns` rewrites `mapKeys(m)` to `m.keys` subcolumn, which reads back in the stored order → bucket order. The reference expects insertion order because that was the default storage format when the test was written.

**c) Fix matches root cause?** Yes. The fix pins `map_serialization_version` / `map_serialization_version_for_zero_level_parts` to `basic` in `CREATE TABLE SETTINGS`, which overrides any CI-injected randomization of those specific settings. Surgical: it does not add `no-random-merge-tree-settings` or similar broad tags.

**d) Test intent preserved?** Yes. The test's purpose is verifying the `optimize_functions_to_subcolumns` rewrite — proven via the `EXPLAIN QUERY TREE` assertions (`column_name: m.keys`, `column_name: m.values`). That rewrite is orthogonal to storage format; pinning one does not hide a bug in the other.

- Unpinned table with `map_serialization_version = 'with_buckets'` → `['bar','foo'] [2,1]` (matches the CI failure)
- Pinned to `basic` → `['foo','bar'] [1,2]` (matches the reference)
- Ran the full test file 20 times under the fix → 20/20 OK locally.

**f) Fix is general, not a narrow patch?** N/A — test-only fix.

Closest prior pattern: commit `d238b482d5c` ("Fix flaky test 02974_analyzer_array_join_subcolumn") which addressed the same root-cause class (bucket reordering surfaced by subcolumn reads) on a sibling test. That fix used `ORDER BY 1` because the problematic query was `ARRAY JOIN pe` producing a whole Map; here the problematic query is `SELECT mapKeys(m), mapValues(m)` directly producing scalar arrays, so pinning the storage setting is the most targeted fix.

groeneai:
cc @Avogar — could you review this test-only fix? It pins `map_serialization_version` / `map_serialization_version_for_zero_level_parts` to `basic` in the test's `CREATE TABLE SETTINGS` so the `m.keys`/`m.values` subcolumns return keys in insertion order regardless of CI randomization (same pattern as the earlier fix for `02974_analyzer_array_join_subcolumn` you reviewed).

Avogar (approved):

PR #12279, opened 2020-07-08
Fix arrayFill for empty arrays.

alexey-milovidov:
Changelog category (leave one):
- Bug Fix

Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
The function `arrayFill` worked incorrectly for empty arrays that may lead to crash. This fixes #12263.

Issue #12263, opened 2020-07-07
arrayReverseFill: OOB access in PODArray


akuzm:
```
SELECT arrayReverseFill(x -> (x < 10), [])


clickhouse-server: /home/akuzm/ch4/ch/src/Common/PODArray.h:342: const T &DB::PODArray<char8_t, 4096, Allocator<false, false>, 15, 16>::operator[](ssize_t) const [T = char8_t, initial_bytes = 4096, TAllocator = Allocator<false, false>, pad_right_ = 15, pad_left_ = 16]: 
Assertion `(n >= (static_cast<ssize_t>(pad_left_) ? -1 : 0)) && (n <= static_cast<ssize_t>(this->size()))' failed.


2020.07.07 19:28:08.540578 [ 550727 ] {} <Trace> BaseDaemon: Received signal 6
2020.07.07 19:28:08.541237 [ 550799 ] {} <Fatal> BaseDaemon: ########################################
2020.07.07 19:28:08.542558 [ 550799 ] {} <Fatal> BaseDaemon: (version 20.6.1.1, build id: F176DE2A5C050ABF) (from thread 550787) (query_id: ba23b9b8-1c51-400c-88e0-8db4293edfbc) Received signal Aborted (6)
2020.07.07 19:28:08.542940 [ 550799 ] {} <Fatal> BaseDaemon: 
2020.07.07 19:28:08.543265 [ 550799 ] {} <Fatal> BaseDaemon: Stack trace: 0x7fe6a5f1218b 0x7fe6a5ef1859 0x7fe6a5ef1729 0x7fe6a5f02f36 0x7fe6a4bf7d5d 0x7fe692353899 0x7fe692351ade 0x7fe69fd6bb6f 0x7fe6a0598e6b 0x7fe6a0599f3d 0x7fe6993c3753 0x7fe6993c6e83 0x7fe6950ffe4c 0x7fe697bc63c5 0x7fe695b99ea8 0x7fe6959c01ac 0x7fe6959c0120 0x7fe6959c00dd 0x7fe6959c008d 0x7fe6959c005d 0x7fe6959bf1ae 0x7fe6aa638ac5 0x7fe6aa638a65 0x7fe6959bd935 0x7fe6959be147 0x7fe6959bc680 0x7fe6959bbeca 0x7fe6959ea0fd
[...]

PR #103311, opened 2026-04-21
tests: restore retry loop in test_trace_log_memory_context

groeneai:
### Motivation

PR #101267 (merged 2026-04-21 20:52 UTC) included a side commit that dropped
the 15-iteration retry loop in `test_trace_log_memory_context`, claiming
that the new cache-based sampling makes a single pass sufficient. Shortly
after the merge the test began failing on master and on unrelated in-flight
PRs at different assertion lines:

- master commit `fa205863` (2026-04-21 21:27 UTC) — fails at
  `get_trace_events('Unknown', 'Global', 'JemallocSample') > 0`
  ([CI report](https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=fa205863ba68f05b529e2073defb5cf5df7cc832&name_0=MasterCI&name_1=Integration%20tests%20%28amd_binary_excluded_from_llvm%29))
- PR #103307, #98727, #101434 (2026-04-21 22:06–22:13 UTC) — all fail at
  `get_trace_events('Global', 'Max', 'Memory'/'MemoryPeak') > 0`
- PR #101267 itself before merge (2026-04-19 and 2026-04-20) — fails at
  `get_trace_events('Unknown', 'Max', 'MemorySample', query_id) > 0`

Even with fully-propagated `sample_probability`, two of the trace types
the test waits for are genuinely probabilistic relative to this workload:

- `JemallocSample` is emitted by `jemalloc`'s own profile sampler
  (`lg_prof_sample=20` ≈ one sample per 1 MiB of allocated memory). Ten
  `logTrace('foo')` calls plus a single `SELECT * FROM numbers(100000)
  ORDER BY number` are not guaranteed to allocate enough `jemalloc`-tracked
  memory to trigger a sample inside a `MemoryTrackerBlockerInThread(Global)`
  scope.
- `Memory`/`MemoryPeak` with `memory_context = 'Global'` are only emitted
  when `total_memory_tracker` crosses its `total_memory_profiler_step`
  watermark (4 MiB by default). Whether this happens during a given test
  query depends on cumulative server-wide allocations including background
  activity, not just the query's own allocations.

The retry loop bounds the test runtime at ~15 s while letting these
events accumulate reliably. Restore it, and add a comment explaining why
it is required so it is not dropped again. Only the test is touched;
the `MemoryTracker`/`ThreadStatus` caching changes from #101267 are kept.

### Changelog category (leave one):

- CI Fix or Improvement (changelog entry is not required)

[...]

PR #103424, opened 2026-04-23
Fix arrayReverseFill crash on empty Array(String) rows

groeneai:
The `column_fill_const` branch of `ArrayFillImpl::execute` computed `array_end = in_offset - 1` without guarding against empty sub-arrays. When the first row of an `Array(T)` column was empty (`in_offset == 0`), this underflowed `size_t` to `SIZE_MAX`, and the subsequent `insertManyFrom(src, SIZE_MAX, 0)` call made the server abort because `ColumnString::doInsertManyFrom` eagerly reads `offsets[position - 1]` before using `length`, violating `PODArray`'s bounds assertion.

`arrayFill` (forward) did not abort because it passes `array_begin` (still `0`) as position, but the logic was equally incorrect.

The issue belongs to the same family as [#12263](https://github.com/ClickHouse/ClickHouse/issues/12263) (2020): the 2020 fix ([#12279](https://github.com/ClickHouse/ClickHouse/pull/12279)) only touched the non-constant `column_fill` branch; the constant-fill branch continued to underflow — the `String` element type was needed to expose it because `ColumnString::doInsertManyFrom` eagerly touches `offsets`, while `ColumnVector` / `ColumnDecimal` accidentally survive via `PaddedPODArray`'s left-pad slot.

Reproducer that aborted the server on master:

```sql
CREATE TEMPORARY TABLE t (a Array(String)) ENGINE = Memory;
INSERT INTO t VALUES ([]), (['a']);
SELECT arrayReverseFill(x -> toUInt8(0), a) FROM t;
```

**Primary fix (root cause).** Skip empty sub-arrays before the `- 1` arithmetic in `ArrayFillImpl::execute`. A skipped sub-array needs no fill and correctly leaves `array_begin` pointing at its own offset, ready for the next non-empty sub-array.

**Secondary fix (contract enforcement).** `ColumnString::doInsertManyFrom` now returns early when `length == 0`. Inserting zero copies must be a no-op regardless of whether `position` is addressable, and the pre-existing eager read made this invariant fragile for every future caller.

New regression test `04111_array_fill_empty_strings.sql` covers the `arrayReverseFill` crash path, the `arrayFill` forward-fill equivalent, and adjacent scenarios (constant-true lambda fast path, non-constant lambda). The existing `01019_array_fill` and `01376_array_fill_empty` continue to pass. Local verification: 50/50 passes of the new test under full CI randomization on a debug build.

STID 0988-292b. Report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=46b63d609fa95ec92f3a3a9cf37482254532db7b&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_debug%29

Fix a server abort in `arrayFill` and `arrayReverseFill` when applied to an `Array(String)` column whose first row is empty together with a constant-false lambda. The aggregate / sub-array loop underflowed a `size_t` index to `SIZE_MAX`, which then caused an out-of-bounds read inside `ColumnString::doInsertManyFrom`. Same bug family as #12263.

### Version info
- Merged into: `26.5.1.115`


groeneai:
cc @rschu1ze @Avogar @alexey-milovidov — could you review this? It fixes a server abort in the constant-false `column_fill_const` branch of `ArrayFillImpl` on `Array(String)` with an empty first row. The 2020 fix (#12279) only patched the non-constant branch; the constant branch kept underflowing `array_end = in_offset - 1` to `SIZE_MAX`, which the `String` element type exposed because `ColumnString::doInsertManyFrom` eagerly reads `offsets[position - 1]` before checking `length`. Two-pronged: skip empty sub-arrays in `arrayFill.cpp` (root cause) + early-return for `length == 0` in `ColumnString::doInsertManyFrom` (contract hardening).

groeneai:
**Pre-PR validation gate (self-check):**

- **(a) Deterministic repro?** Yes — `CREATE TEMPORARY TABLE t (a Array(String)) ENGINE = Memory; INSERT INTO t VALUES ([]), (['a']); SELECT arrayReverseFill(x -> toUInt8(0), a) FROM t;` aborts the server 100% of the time on master (debug build).
- **(b) Root cause explained?** Yes — `column_fill_const` branch of `ArrayFillImpl::execute` computes `array_end = in_offset - 1`. When the first row is empty (`in_offset == 0`), this underflows `size_t` to `SIZE_MAX`. `insertManyFrom(src, SIZE_MAX, 0)` then makes `ColumnString::doInsertManyFrom` read `offsets[SIZE_MAX - 1]` (i.e. `offsets[-2]` as `ssize_t`), violating the `PaddedPODArray` bounds assertion `n >= -1`.
- **(c) Fix matches root cause?** Yes — primary fix skips empty sub-arrays before the underflow can happen. Secondary fix enforces the `length == 0 ⇒ no-op` contract at the crash site so any future caller is safe as well.
- **(d) Test intent preserved / new tests added?** Added `04111_array_fill_empty_strings.sql` covering (1) the `arrayReverseFill` abort path, (2) the analogous `arrayFill` forward-fill case that accidentally survived, (3) constant-true lambda fast path, (4) non-constant lambda (untouched `column_fill` branch), and (5) single-row empty-array variants matching #12263. Existing `01019_array_fill` and `01376_array_fill_empty` still pass (verified).
- **(e) Demonstrated in both directions?** Yes — abort without the fix (reproducer above), correct result with the fix (`[]` and `['a']`). 50/50 passes for `04111_array_fill_empty_strings` under full CI randomization on debug build; 10/10 passes for `01376_array_fill_empty`.
- **(f) Fix is general, not a narrow patch?** Yes — I traced why sibling `insertManyFrom` implementations did not abort: `ColumnVector`/`ColumnDecimal` accidentally survive the `SIZE_MAX` position because `PaddedPODArray` allows the `-1` slot via `pad_left`; `ColumnFixedString::doInsertManyFrom` is safe because the copy loop body does not execute when `length == 0`. The root-cause fix in `arrayFill.cpp` removes the bad position for all column types; the defensive change in `ColumnString` hardens the only eager-access site.

Session ID: `cron:clickhouse-ci-task-worker:20260423-064500`

groeneai:
Heads up on the **Bugfix validation (functional tests)** failure showing `04111_array_fill_empty_strings | FAIL`: this is an expected limitation of the validation infrastructure for this class of bug, not a regression of the fix.

**Why it fails:** Bugfix validation downloads the master **release** `clickhouse` binary and runs the new test against it, expecting the test to fail (so the inverted "FAIL → OK" check passes). The STID 0988-292b bug only manifests as a **debug-build** assertion (`PODArray::operator[]` bounds check in `DEBUG_OR_SANITIZER_BUILD`). In release builds the OOB read is silent, and because `length == 0` on the buggy call path (the underflow coincides with an empty-sub-array skip), both the `memcpy` loop and the `offsets.resize` loop are skipped — no visible data corruption, no crash. So the test correctly passes on master release, and the validator reports FAIL.

**What still works:**
- The deterministic debug-build reproducer (included in the PR description) aborts the server on master without the fix — that's where STID 0988-292b was originally captured (`Stress test (experimental, serverfuzz, arm_debug)`).
- The regression test `04111_array_fill_empty_strings.sql` will catch any re-regression on the debug-build CI variants (`amd_debug`, `arm_debug`, flaky check with debug, sanitizers), which run on every PR.
- Recent merged PRs with similar debug-only behavior show the same Bugfix validation signature and merged without issue (e.g. #102289 `04093_empty_tuple_parsing | FAIL | failure`).

Happy to rework the test if reviewers prefer a release-visible reproducer, but I wasn't able to construct one: the `length == 0` short-circuit means the OOB read has no observable consequence in release for any input I tried.

Session: `cron:clickhouse-ci-task-worker:20260423-074500`

groeneai:
Friendly re-ping @rschu1ze @Avogar @alexey-milovidov — this P0 fix has been ready for review for ~26h.

**Status recap:**
- ✅ CI complete: 140 SUCCESS / 15 SKIPPED, AI Review APPROVED, no unexplained failures
- ✅ 2 FAILURES are both documented unrelated (see [earlier comment](https://github.com/ClickHouse/ClickHouse/pull/103424#issuecomment-4302622038)): `Bugfix validation` is expected for this debug-only assertion bug (release binary cannot repro because the bad OOB read coincides with `length == 0` short-circuit), and `Integration tests (arm_binary distributed plan 3/4)` hit the already-fixed chronic flaky `test_memory_context_in_trace_log` (our PR #103311 merged Apr 23 08:22 UTC — after this PR's CI ran)
- ✅ Master clean: STID 0988-292b has had zero recurrence since the originating hit (~26h clean)

**Fix summary:**
Two-pronged patch to `src/Functions/array/arrayFill.cpp` (skip empty sub-arrays before `-1` arithmetic to avoid `size_t` underflow) + `src/Columns/ColumnString.cpp::doInsertManyFrom` (defensive early-return on `length == 0`). Regression test `04111_array_fill_empty_strings.sql` covers all scenarios. Same bug class as historical issue #12263 (2020), resurfaced through a different `insertManyFrom` code path.

Thanks!

groeneai:
Second re-ping after ~67h with no review activity. Adding @Algunenano (recent committer on `ColumnString.cpp` + `Functions/array/`) and @CheSema (recent committer on `ColumnString.cpp`) as alternate reviewers — the original ping recipients (@rschu1ze, @Avogar, @alexey-milovidov) haven't responded.

**Status:**
- 140 checks SUCCESS, 2 documented unrelated failures, 4 pending. Mergeable.
- Master STID 0988-292b clean for ~67h since the originating master hit on `46b63d6`. No recurrence.
- No competing fix PRs (verified just now).

**Fix recap (two-pronged, surgical):**
- `src/Functions/array/arrayFill.cpp` — `column_fill_const` branch underflowed `array_end = in_offset - 1` to `SIZE_MAX` when first sub-array is empty. Added `if (in_offset == array_begin) continue;` guard.
- `src/Columns/ColumnString.cpp::doInsertManyFrom` — eagerly read `offsets[position-1]` before checking `length`. Added `if (length == 0) return;` early-return (matches contract used by sibling column types).

The first fix prevents the bad value from being constructed. The second is defense-in-depth that helps any future caller that hits the same "garbage position with length=0" pattern.

**Why the Bugfix validation FAIL is expected** (already explained in earlier comment): assertion is debug-only; the OOB read is silent in release because `length=0` short-circuits all subsequent loops. Documented analogues: PR #102289 with same pattern merged cleanly. **`test_memory_context_in_trace_log`** failure was fixed by my own PR #103311 merged 18 min after this CI ran — not a fix concern.

Happy to rebase if useful, but no conflicts as of right now.

alexey-milovidov:
@groeneai, but the Bugfix validation checks does run debug assertions, iirc.

groeneai:
@alexey-milovidov — I dug into the CI code to verify, and I believe my original explanation is correct. Bugfix validation downloads the **release** binary, not a debug build. Here is the chain of evidence:

### 1. Where the binary comes from

`ci/jobs/functional_tests.py:303-310` — when `is_bugfix_validation`, the runner downloads:

```python
link_to_master_head_binary = f"https://clickhouse-builds.s3.us-east-1.amazonaws.com/master/{link_arch}/clickhouse"
```

That URL points to the **static `master/{arch}/clickhouse`** location.

### 2. Which build types upload to that URL

`ci/jobs/scripts/job_hooks/build_master_head_hook.py:9-21` — the static-location hook only runs for these build types:

```python
BUILD_TYPE_TO_STATIC_LOCATION = {
    BuildTypes.AMD_RELEASE: "amd64",
    BuildTypes.ARM_RELEASE: "aarch64",
    BuildTypes.AMD_DARWIN: "macos",
    ... (no DEBUG / ASAN / TSAN / MSAN entries)
}
```

So `master/aarch64/clickhouse` and `master/amd64/clickhouse` are populated **only** by `ARM_RELEASE` and `AMD_RELEASE` builds. Debug, asan, tsan, msan builds upload elsewhere and do not feed this static location.

### 3. Why debug assertions don't fire in that binary

`base/base/sanitizer_defs.h:61-66`:

```cpp
#if !defined(DEBUG_OR_SANITIZER_BUILD)
#    if !defined(NDEBUG) || defined(ADDRESS_SANITIZER) || defined(THREAD_SANITIZER) ...
#        define DEBUG_OR_SANITIZER_BUILD
#    endif
#endif
```

`AMD_RELEASE` / `ARM_RELEASE` build with `NDEBUG` defined and no sanitizers, so `DEBUG_OR_SANITIZER_BUILD` is **not defined**.

`base/base/defines.h:42-47` — when `DEBUG_OR_SANITIZER_BUILD` is undefined:

```cpp
#define chassert_1(x, ...) (void)sizeof(!(x))
#define chassert_2(x, comment, ...) (void)sizeof(!(x))
```

`chassert` becomes a no-op that doesn't even evaluate `x`.

`src/Common/PODArray.h:404-407`:

```cpp
const T & operator[] (ssize_t n) const
{
    assert((n >= (static_cast<ssize_t>(pad_left_) ? -1 : 0)) && (n <= static_cast<ssize_t>(this->size())));
    return t_start()[n];
}
```

This uses standard libc `assert()`, which is also compiled out when `NDEBUG` is defined.

### 4. Why our specific test silently passes on master release

The STID 0988-292b crash path:
1. `arrayReverseFill` on `[[],['a']]` → `array_end = in_offset - 1` underflows to `SIZE_MAX` when first sub-array is empty
2. Calls `ColumnString::doInsertManyFrom(src, SIZE_MAX, 0)` — `length == 0`
3. In **debug**: `PODArray::operator[]` `assert()` fires → SIGABRT
4. In **release**: `assert()` is a no-op, the OOB pointer is computed but never dereferenced, the subsequent `memcpy` loop and `offsets.resize` loop are both skipped (length=0), so the function is a no-op and the output exactly matches the .reference

So the test's stdout on master release **equals** the test's stdout on the patched PR. The validator (which expects different output for FAIL on master) reports FAIL as a result.

### 5. Empirical confirmation

Same pattern happened on **PR #102289** (`04093_empty_tuple_parsing | FAIL | failure | 5`), which **merged cleanly on 2026-04-20** despite the identical Bugfix validation signature — also a debug-only assertion-class fix.

### What I think happened with the "iirc"

It's possible you were thinking of:
- **Bugfix validation (integration tests)** — downloads the same `master/{arch}/clickhouse` binary in `integration_test_job.py:697-708`, also release.
- A historical version of the check before the static-location refactor (`27ab6daa303` "Build clickhouse-stripped for all builds, fail on the hook failure").
- The unrelated `Bugfix Validation` aggregator job (`JobNames.BUGFIX_VALIDATE`) — it doesn't run tests, just rolls up FT+IT results.

In all cases, the FT/IT job that actually executes test queries downloads the release binary — there's no debug variant in the current CI.

### Suggested path forward

1. **Merge PR #103424 as-is.** The fix is correct, the regression test is correct; the Bugfix validation FAIL is structurally unavoidable for this class of debug-only assertion bug, exactly as with #102289.
2. **(Separate concern, related to PR #103541)** If we want Bugfix validation to actually catch debug-only fixes, the static-location hook would need to also publish a debug or sanitizer-enabled `master/{arch}/clickhouse-debug` binary, and the runner would need to choose between them based on the PR's tagging. Happy to take that on as a follow-up if you want.

Citations on this PR's branch:
- `ci/jobs/functional_tests.py:303-313`
- `ci/jobs/scripts/job_hooks/build_master_head_hook.py:9-67`
- `base/base/sanitizer_defs.h:59-66`
- `base/base/defines.h:29-54`
- `src/Common/PODArray.h:402-408`

thevar1able (approved):

clickgapai:

**Affected code:** `ArrayFillImpl::execute` (`column_fill_const` branch) and `ColumnString::doInsertManyFrom` — core code present in all supported branches.

**Why:** Bug is ancient — `arrayFill.cpp` predates the `dbms/` → `src/` rename (commit 06446b4f08a1, ~2020) and the buggy `array_end = in_offset - 1` line was untouched by #12279 (2020). All supported release branches (25.8, 26.1, 26.2, 26.3) contain the unpatched `column_fill_const` branch. PR has no backport label set (only `pr-bugfix`, `can be tested`, `pr-synced-to-cloud`).

PR #103405, opened 2026-04-22
Fix RMV refresh race on concurrent rename of target storage

davenger:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fixes ACCESS_DENIED / UNKNOWN_TABLE in downstream APPEND RMVs with SQL SECURITY DEFINER when an upstream REPLACE RMV's EXCHANGE flips the target storage identity mid-resolution

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #103427, opened 2026-04-23
Alternative fix for RMV refresh race on concurrent rename of target storage

davenger:
Alternative of https://github.com/ClickHouse/ClickHouse/pull/103405
Let's check which one is better

Fixes ACCESS_DENIED / UNKNOWN_TABLE in downstream APPEND RMVs with SQL SECURITY DEFINER when an upstream REPLACE RMV's EXCHANGE flips the target storage identity mid-resolution

### Version info
- Merged into: `26.5.1.206`

### DIFF
--- a/src/Analyzer/Resolve/IdentifierResolver.cpp
+++ b/src/Analyzer/Resolve/IdentifierResolver.cpp
@@ -234,6 +234,9 @@ std::shared_ptr<TableNode> IdentifierResolver::tryResolveTableIdentifier(const I
         auto database = DatabaseCatalog::instance().tryGetDatabase(storage_id.getDatabaseName());
         if (database)
             storage = database->tryGetTable(table_name, context);
+        /// Adopt the replacement's identity so TableNode stays resolvable by UUID.
+        if (storage)
+            storage_id = storage->getStorageID();
     }
     if (!storage)
         return {};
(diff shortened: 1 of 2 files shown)


al13n321 (approved):
This is not specific to RMV, right? A normal SELECT with a concurrent normal EXCHANGE could run into the same problem?

Why does it matter whether the downstream RMV is APPEND?

al13n321:
(I didn't do a deep review, hope you did.)

davenger:
>This is not specific to RMV, right? A normal SELECT with a concurrent normal EXCHANGE could run into the same problem?

I think yes, it's just that RMV with REPLACE has chances to occasionally hit this. 

> Why does it matter whether the downstream RMV is APPEND?

Likely it doesn't matter. The test just reproduces the scenario that was hitting the issue.

clickgapai:

**Affected code:** `IdentifierResolver::tryResolveTableIdentifier` — core code present in all supported branches.

**Why:** `tryResolveTableIdentifier` in `src/Analyzer/Resolve/IdentifierResolver.cpp` was introduced in May 2024 (commit b2efe1537af, 'Move the code of identifier resolution to IdentifierResolver'), predating all supported release branches. The race causes user-visible ACCESS_DENIED on legitimate RMV refreshes, which qualifies for backport.

PR #101652, opened 2026-04-02
Fix incorrect JOIN results caused by silently dropped WHERE filters

wudidapaopao:
Fixed https://github.com/ClickHouse/ClickHouse/issues/102002

When `extractActionsForJoinCondition` encountered an `equals` conjunct with mismatched argument types (e.g. `UInt32` vs `UInt64`), it would `continue` without adding the conjunct to `rejected_conjuncts`. If another conjunct in the same WHERE clause was successfully extracted into the JOIN condition, the type-mismatched one was silently lost and the `FilterStep` could be replaced by an `ExpressionStep`, producing incorrect query results.

Reproducer: https://fiddle.clickhouse.com/3306ebe1-2dc7-402b-84da-145ed91e0c7e
```sql
DROP TABLE IF EXISTS t1;
DROP TABLE IF EXISTS t2;

CREATE TABLE t1 (c1 UInt32, c2 String, c3 String) ENGINE = Memory;
CREATE TABLE t2 (c1 UInt64, c2 String, c3 String) ENGINE = Memory;

INSERT INTO t1 VALUES (1, 'x', 'x'), (2, 'y', 'y');
INSERT INTO t2 VALUES (1, 'x', 'x'), (3, 'y', 'y');

SELECT * FROM t1 INNER JOIN t2 ON t1.c3 = t2.c3 WHERE t1.c1 = t2.c1 AND t1.c2 = t2.c2;
-- incorrect output: 
-- 1	x	x	1	x	x
-- 2	y	y	3	y	y


DROP TABLE t1;
DROP TABLE t2;
```

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix incorrect JOIN query results when `mergeFilterIntoJoinCondition` optimization silently dropped a WHERE equality condition with mismatched types.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.239`

PR #100001, opened 2026-03-19
Extend `InverseDictionaryLookupPass` to support `IN` and `notIn` operators for dictionary lookups

euvre:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Extended `InverseDictionaryLookupPass` to support `IN` and `notIn` operators for dictionary attribute lookups. Queries like `WHERE dictGet('dict', 'attr', key) IN ('val1', 'val2')` are now optimized to `WHERE key IN (SELECT dict_key FROM dictionary WHERE attr IN ('val1', 'val2'))`, enabling better index utilization.

### Example use:
```sql
-- Before: Full dictionary scan required
SELECT * FROM table WHERE dictGetString('dict', 'status', id) IN ('active', 'pending');

-- After optimization: Converted to dictionary key lookup
SELECT * FROM table WHERE id IN (SELECT key FROM dictionary WHERE status IN ('active', 'pending'));
```

Resolve #99500

PR #103434, opened 2026-04-23
Fix UBSan pointer-overflow in `LimitSeekableReadBuffer::seek`

groeneai:
`LimitSeekableReadBuffer::seek` used `pos + position_change` inside its
in-buffer fast-path check. When the inner buffer has not materialized
its memory yet — as is the case for `EmptyReadBuffer`,
`ParallelReadBuffer`, and `ReadWriteBufferFromHTTP`, all of which start
with `pos == nullptr` — and a non-zero seek is requested before the
first `next`, this triggers undefined behavior per [expr.add]:
"applying non-zero offset N to null pointer".

On master `BuzzHouse (arm_asan_ubsan)` this surfaced from the BACKUP
write path as **STID 3079-3df0**:

```
src/IO/LimitSeekableReadBuffer.cpp:97:34: runtime error: applying non-zero offset 1178 to null pointer
  #0 DB::LimitSeekableReadBuffer::seek(long, int)
  #1 DB::BackupWriterDefault::copyDataToFile(...)
  #2 DB::BackupImpl::writeFile(...)
  #3 DB::BackupsWorker::writeBackupEntries(...)
```

**Root cause.** `BackupEntryFromAppendOnlyFile::getReadBuffer` wraps the
underlying disk read buffer with `LimitSeekableReadBuffer`. Some inner
buffer types start with `pos == nullptr` until the first read. When
`BackupWriterDefault::copyDataToFile` immediately calls
`read_buffer->seek(base_size, SEEK_SET)` on such a wrapper, the fast-path
check computes `nullptr + position_change`, which is UB.

**Fix.** Replace the pointer-arithmetic comparison with an equivalent
check expressed via pointer *differences* and integer arithmetic, which
is well-defined regardless of whether `pos` is null. The condition is
mathematically identical to the original for all legal inputs (for
non-null `pos` within the working buffer, `begin - pos` and `end - pos`
are valid `ptrdiff_t` values; for an empty buffer with null `pos` the
check only accepts `position_change == 0`, matching the original).

**Test.** Added `gtest_limit_seekable_read_buffer.cpp` with three
scenarios:
1. `SeekOnInnerBufferWithNullPos` — regression test that reproduces the
   exact UBSan error from CI (`runtime error: applying non-zero offset
   1178 to null pointer`) when the fix is reverted.
2. `SeekInsideMaterializedBuffer` — sanity check for the normal
   in-buffer fast-path.
3. `SeekOutsideMaterializedBuffer` — sanity check for the slow-path
   where the seek target is outside the working buffer.

All three pass under `-fsanitize=address,undefined` with
`UBSAN_OPTIONS=halt_on_error=1`. Without the fix the first test
deterministically triggers the CI UBSan error at the exact same
location (line 97, column 34, offset 1178).

CI report:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=4b8132993ce0625e9c0ed0bf15ed16eb6df8262b&name_0=MasterCI&name_1=BuzzHouse+%28arm_asan_ubsan%29

Fix undefined behavior in `LimitSeekableReadBuffer::seek` when the wrapped inner buffer has not been materialized yet. The fast-path in-buffer check performed pointer arithmetic on a possibly-null `pos`, which UBSan reports as "applying non-zero offset to null pointer". Triggered from the BACKUP write path in debug/sanitizer builds.

### DIFF
--- a/src/IO/LimitSeekableReadBuffer.cpp
+++ b/src/IO/LimitSeekableReadBuffer.cpp
@@ -98,7 +98,8 @@ off_t LimitSeekableReadBuffer::seek(off_t off, int whence)
     /// on pointer differences. Computing `pos + position_change` directly is undefined behavior
     /// when `pos` is null (e.g. when the inner buffer hasn't been materialized yet) or when
     /// `position_change` is large enough that the resulting pointer escapes the allocation.
-    if (working_buffer.begin() - pos <= position_change && position_change <= working_buffer.end() - pos)
+    if (pos && working_buffer.begin() <= pos + position_change
+           && pos + position_change <= working_buffer.end())
     {
         /// Position is still inside the buffer.
         pos += position_change;
--- a/src/IO/LimitSeekableReadBuffer.cpp
+++ b/src/IO/LimitSeekableReadBuffer.cpp
@@ -98,8 +98,7 @@ off_t LimitSeekableReadBuffer::seek(off_t off, int whence)
     /// on pointer differences. Computing `pos + position_change` directly is undefined behavior
     /// when `pos` is null (e.g. when the inner buffer hasn't been materialized yet) or when
     /// `position_change` is large enough that the resulting pointer escapes the allocation.
-    if (pos && working_buffer.begin() <= pos + position_change
-           && pos + position_change <= working_buffer.end())
+    if (working_buffer.begin() - pos <= position_change && position_change <= working_buffer.end() - pos)
     {
         /// Position is still inside the buffer.
         pos += position_change;
--- a/src/IO/LimitSeekableReadBuffer.cpp
+++ b/src/IO/LimitSeekableReadBuffer.cpp
@@ -94,7 +94,11 @@ off_t LimitSeekableReadBuffer::seek(off_t off, int whence)
         throw Exception(ErrorCodes::ARGUMENT_OUT_OF_BOUND, "Seek shift out of bounds");
 
     off_t position_change = new_position - current_position;
-    if ((buffer().begin() <= pos + position_change) && (pos + position_change <= buffer().end()))
+    /// Check whether the new position is still inside the working buffer using integer arithmetic
+    /// on pointer differences. Computing `pos + position_change` directly is undefined behavior
+    /// when `pos` is null (e.g. when the inner buffer hasn't been materialized yet) or when
+    /// `position_change` is large enough that the resulting pointer escapes the allocation.
+    if (working_buffer.begin() - pos <= position_change && position_change <= working_buffer.end() - pos)
     {
         /// Position is still inside the buffer.
         pos += position_change;
--- a/src/IO/tests/gtest_limit_seekable_read_buffer.cpp
+++ b/src/IO/tests/gtest_limit_seekable_read_buffer.cpp
@@ -1,0 +1,86 @@
+#include <gtest/gtest.h>
+
+#include <IO/EmptyReadBuffer.h>
+#include <IO/LimitSeekableReadBuffer.h>
+#include <IO/ReadBufferFromMemory.h>
+#include <IO/ReadHelpers.h>
+
+using namespace DB;
+
+/// Regression test for UBSAN finding STID 3079-3df0:
+///   src/IO/LimitSeekableReadBuffer.cpp:97:34:
+///   runtime error: applying non-zero offset 1178 to null pointer
+///
+/// Reproduction scenario: the inner buffer has not materialized its memory yet
+/// (e.g. `ParallelReadBuffer`, `ReadWriteBufferFromHTTP`, `EmptyReadBuffer` —
+/// all start with `pos == nullptr`). When `LimitSeekableReadBuffer` wraps such
+/// a buffer and `seek(non_zero_offset, SEEK_SET)` is called before the first
+/// read, the fast-path check in `LimitSeekableReadBuffer::seek` used to
+/// compute `pos + position_change`. Doing pointer arithmetic with a non-zero
+/// offset on a null pointer is undefined behavior per [expr.add] and UBSAN
+/// catches it under `-fsanitize=pointer-overflow`.
+TEST(LimitSeekableReadBuffer, SeekOnInnerBufferWithNullPos)
+{
+    /// EmptyReadBuffer is constructed via `SeekableReadBuffer(nullptr, 0)`, so
+    /// its `pos` and `working_buffer` are both null pointers. This is the
+    /// cheapest way to reproduce the UB; real-world triggers from the backup
+    /// stack (STID 3079-3df0) involve HTTP or parallel read buffers that also
+    /// start with `pos == nullptr` until the first `next()` call.
+    auto inner = std::make_unique<EmptyReadBuffer>();
+    LimitSeekableReadBuffer limit_buf(std::move(inner), 0, 4096);
+
+    /// Before the fix this line produced:
+    ///   runtime error: applying non-zero offset 1178 to null pointer
+    /// under `-fsanitize=pointer-overflow`. After the fix, the fast-path uses
+    /// integer arithmetic on pointer differences and is well-defined.
+    off_t result = limit_buf.seek(1178, SEEK_SET);
+    ASSERT_EQ(result, 1178);
+}
+
+/// Sanity check that seeks still work correctly after reading when the inner
+/// buffer does have a materialized working buffer. This guards against the
+/// fix accidentally breaking the normal in-buffer seek fast-path.
+TEST(LimitSeekableReadBuffer, SeekInsideMaterializedBuffer)
+{
+    const std::string data = "0123456789abcdef0123456789abcdef";
+    auto inner = std::make_unique<ReadBufferFromMemory>(data.data(), data.size());
+    LimitSeekableReadBuffer limit_buf(std::move(inner), 2, 20);
+
+    /// Force the buffer to materialize.
+    char c = 0;
+    readChar(c, limit_buf);
+    ASSERT_EQ(c, '2'); /// data[2]
+    ASSERT_EQ(limit_buf.getPosition(), 1);
+
+    /// Forward seek inside the working buffer — fast path.
+    off_t result = limit_buf.seek(5, SEEK_SET);
+    ASSERT_EQ(result, 5);
+    readChar(c, limit_buf);
+    ASSERT_EQ(c, '7'); /// data[2 + 5]
+
+    /// Backward seek inside the working buffer — fast path.
+    result = limit_buf.seek(0, SEEK_SET);
+    ASSERT_EQ(result, 0);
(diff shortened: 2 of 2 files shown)


groeneai:
### Pre-PR Validation Gate

Session id: `cron:clickhouse-ci-task-worker:20260423-094500`

**a) Deterministic repro?** ✅
`build-asan/src/unit_tests_dbms --gtest_filter='LimitSeekableReadBuffer.SeekOnInnerBufferWithNullPos'` with the fix reverted triggers:
```
src/IO/LimitSeekableReadBuffer.cpp:97:34: runtime error: applying non-zero offset 1178 to null pointer
```
— identical line, column, and offset to the CI failure (STID 3079-3df0).

**b) Root cause explained?** ✅
`BackupEntryFromAppendOnlyFile::getReadBuffer` wraps an inner `SeekableReadBuffer` that may not have materialized its memory yet (inner `pos == nullptr` for `EmptyReadBuffer`, `ParallelReadBuffer`, `ReadWriteBufferFromHTTP`). The `LimitSeekableReadBuffer` constructor adopts the inner buffer's `pos`, so it is also null. `BackupWriterDefault::copyDataToFile` immediately calls `read_buffer->seek(base_size, SEEK_SET)`. In `LimitSeekableReadBuffer::seek`, the fast-path computes `pos + position_change` — pointer arithmetic with a non-zero offset on a null pointer, which is UB per [expr.add].

**c) Fix matches root cause?** ✅
The pointer-arithmetic expression is replaced with an equivalent check expressed via pointer *differences* and integer arithmetic: `working_buffer.begin() - pos <= position_change && position_change <= working_buffer.end() - pos`. Pointer subtraction of two pointers within the same allocation (or both null) is defined. No pointer arithmetic with offsets is performed. The new check also removes the latent second UB case (large `position_change` with valid `pos` escaping the allocation).

**d) Test intent preserved?** ✅
The condition is mathematically identical to the original for all legal inputs. For non-null `pos` inside a non-empty working buffer, `begin - pos ≤ 0 ≤ end - pos`, and the acceptable range for `position_change` is the same. For null `pos` with an empty buffer, both differences are 0 and the check only accepts `position_change == 0` — matching original semantics in that degenerate case.

**e) Both directions demonstrated?** ✅
- **Without fix:** UBSan reports `runtime error: applying non-zero offset 1178 to null pointer` at `LimitSeekableReadBuffer.cpp:97:34` — exact CI failure reproduction. Test fails.
- **With fix:** All three new tests pass under `-fsanitize=address,undefined` with `UBSAN_OPTIONS=halt_on_error=1 print_stacktrace=1`. Broader IO unit test suite (23 tests) also passes.

**f) Fix is general, not a narrow patch?** ✅
- `grep` for `pos + position_change` and `buffer().begin() <= pos` across `src/` returned only the fixed line — the dangerous pattern exists in no other file.
- The fix targets the arithmetic pattern itself, not the specific null-pointer case: a defensive `if (pos != nullptr)` would only cover the caught case but would still UB when `pos` is valid and `position_change` is large enough to escape the allocation. The new integer-arithmetic check is safe in both scenarios.

groeneai:
cc @alexey-milovidov @CheSema — could you review? This fixes a UBSan `pointer-overflow` bug in `LimitSeekableReadBuffer::seek` (STID 3079-3df0 on master BuzzHouse arm_asan_ubsan) by swapping the fast-path pointer-arithmetic check for an equivalent integer-arithmetic check on pointer differences, which is defined even when `pos` is null. Added a gtest that reproduces the exact CI UBSan error without the fix.

src/IO/LimitSeekableReadBuffer.cpp
```diff
@@ -94,7 +94,11 @@ off_t LimitSeekableReadBuffer::seek(off_t off, int whence)
         throw Exception(ErrorCodes::ARGUMENT_OUT_OF_BOUND, "Seek shift out of bounds");
 
     off_t position_change = new_position - current_position;
-    if ((buffer().begin() <= pos + position_change) && (pos + position_change <= buffer().end()))
+    /// Check whether the new position is still inside the working buffer using integer arithmetic
+    /// on pointer differences. Computing `pos + position_change` directly is undefined behavior
+    /// when `pos` is null (e.g. when the inner buffer hasn't been materialized yet) or when
+    /// `position_change` is large enough that the resulting pointer escapes the allocation.
+    if (working_buffer.begin() - pos <= position_change && position_change <= working_buffer.end() - pos)
```
CheSema, line 101:
```suggestion
    if (pos && working_buffer.begin() <= pos + position_change
           && pos + position_change <= working_buffer.end())
```

I find it more explicit.

groeneai: Hi @CheSema — thanks for the readability suggestion, but I had to revert this in commit cc7aaa486d5. `clickhouse-gh[bot]` caught a correctness gap: even with the `pos &&` short-circuit guarding the null case, the expression `pos + position_change` is still UB per `[expr.add]` once the resulting pointer leaves the allocation. UBSan's `-fsanitize=pointer-overflow` fires on that — and the BackupWriterDefault path that produced STID 3079-3df0 routinely seeks by large offsets (`base_size` can be GB-scale), so a legal `seek(base_size, SEEK_SET)` with a non-null `pos` pointing into a few-KB working buffer still trips UBSan.

The integer-arithmetic form (`working_buffer.begin() - pos <= position_change && position_change <= working_buffer.end() - pos`) is defined in both cases:

- `pos == nullptr`: `begin()` and `end()` are also null (same uninitialized buffer). `nullptr - nullptr` is defined as `0`. For any non-zero `position_change` the check is false → slow path, correct behavior.
- `pos != nullptr`: both pointers belong to the same allocation (the working buffer), so pointer subtraction is defined and returns the signed offset.

Happy to iterate on the comment wording or naming if the differences form is hard to read — just didn't want to land a correctness regression to fix STID 3079-3df0 only to trip the same UBSan check on a different code path.

See the bot's analysis: https://github.com/ClickHouse/ClickHouse/pull/103434#discussion_r3130788043

src/IO/LimitSeekableReadBuffer.cpp
```diff
@@ -94,7 +94,12 @@ off_t LimitSeekableReadBuffer::seek(off_t off, int whence)
         throw Exception(ErrorCodes::ARGUMENT_OUT_OF_BOUND, "Seek shift out of bounds");
 
     off_t position_change = new_position - current_position;
-    if ((buffer().begin() <= pos + position_change) && (pos + position_change <= buffer().end()))
+    /// Check whether the new position is still inside the working buffer using integer arithmetic
+    /// on pointer differences. Computing `pos + position_change` directly is undefined behavior
+    /// when `pos` is null (e.g. when the inner buffer hasn't been materialized yet) or when
+    /// `position_change` is large enough that the resulting pointer escapes the allocation.
+    if (pos && working_buffer.begin() <= pos + position_change
```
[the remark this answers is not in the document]
groeneai: Thanks — good catch. Applied in commit cc7aaa486d5, which restores the integer-arithmetic form (`working_buffer.begin() - pos <= position_change && position_change <= working_buffer.end() - pos`). This is defined for both `pos == nullptr` (pointer subtraction of two nulls is `0`, check correctly falls through) and for non-null `pos` with any `position_change` (both pointers live in the same allocation, subtraction is defined and returns the signed offset). The `pos &&` form landed via @CheSema's in-line suggestion commit on this PR but missed the large-legal-seek case, which is exactly what BackupWriterDefault → LimitSeekableReadBuffer::seek(base_size, SEEK_SET) produces in the backup path.

groeneai:
Status update: the `pos &&` form that @CheSema applied at 11:08 UTC was reverted in commit cc7aaa486d5 after `clickhouse-gh[bot]` correctly pointed out at 12:34 UTC that even the short-circuited `pos && (pos + position_change)` expression is UB per `[expr.add]` whenever `position_change` is large enough to push the resulting pointer outside the allocation. The backup path (`BackupWriterDefault::copyDataToFile` → `LimitSeekableReadBuffer::seek(base_size, SEEK_SET)`) routinely seeks by GB-scale offsets into KB-sized working buffers, which is exactly the UB class that UBSan's `-fsanitize=pointer-overflow` catches. The integer-arithmetic form is correct in both cases.

Detailed replies posted inline:
- To @CheSema: https://github.com/ClickHouse/ClickHouse/pull/103434#discussion_r3131059030
- To `clickhouse-gh[bot]`: https://github.com/ClickHouse/ClickHouse/pull/103434#discussion_r3131060028

Waiting on re-review. cc @alexey-milovidov @CheSema

groeneai:
Pushed empty commit `7260d79ab44` to re-trigger CI.

The Fast test failures on `cc7aaa486d5` were all caused by a DNS infrastructure issue on the CI worker (`Cannot resolve host (ip-172-31-64-98), error 0: DNS error`). The CI worker couldn't resolve its own hostname; ClickHouse client failed at `DNSResolver::Impl::Impl()` before any query could execute.

Not code-related — the same DNS failure affected multiple unrelated PRs and master at the same time window (13:22–13:26 UTC, 2026-04-23):
- PR #103434 (ours): 4 DNS failures on `Fast test`
- PR #101652: 10 DNS failures on `Fast test`
- PR #100001: 11 DNS failures on `Fast test`
- master (commit `4f1cdbe`): 6+ DNS failures on `Fast test`

Our fix only touches `src/IO/LimitSeekableReadBuffer.cpp` and a gtest file — no networking/DNS code paths.

nikitamikhaylov:
@groeneai When you see a failure like [Unit tests (msan, function_prop_fuzzer)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=103434&sha=7260d79ab44c38d13cee7925f34b61a77ef49880&name_0=PR&name_1=Unit+tests+%28msan%2C+function_prop_fuzzer%29) or similar -- investigate and fix it in a separate PR.

groeneai:
Thanks @nikitamikhaylov, agreed.

The `Unit tests (msan, function_prop_fuzzer)` failure you pointed to was actually a chronic trunk bug in `getURLScheme` (plus sibling findings in `Lexer::nextToken` and `HasSubsequenceImpl.h` that the same fuzzer surfaced). I had already investigated and opened a fix PR #103491 for those, but @yakov-olkhovskiy's PR #103489 landed on master first at 2026-04-24T11:48:40Z and superseded it — so I closed #103491.

The CI run you linked (commit `7260d79ab44`, 2026-04-23) pre-dates the merge of #103489, which is why `function_prop_fuzzer` fails there.

Just rebased this branch onto current master (new HEAD `12681f197329`) — it now picks up #103489's fixes, so the next CI run on this PR should see `function_prop_fuzzer` clean. The `LimitSeekableReadBuffer::seek` fix itself was unchanged by the rebase.

alexey-milovidov (approved):

PedroTadim:
Need this to fix 26.4 CI

PR #89914, opened 2025-11-12
Show datalake catalogs in SHOW DATABASES query

alesapin:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Now ClickHouse will show data lake catalog database in `SHOW DATABASES` query by default.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #103444, opened 2026-04-23
Fix: system.databases always shows data lake catalog databases

alsugiliazova:
`system.databases` now lists data lake catalog databases regardless of the `show_data_lake_catalogs_in_system_tables` setting.

The setting exists to prevent accidental expensive calls to remote catalog services (e.g. listing tables or columns). Knowing that a catalog database exists, however, requires no remote call — the `DatabaseCatalog` already keeps it in local memory. Hiding the database in `system.databases` based on that setting was inconsistent with `SHOW DATABASES` (which always showed catalogs since #89914) and confusing for users.

Changes:
- `StorageSystemDatabases::fillData` now passes `with_datalake_catalogs = true` unconditionally.
- `InterpreterShowTablesQuery::execute` no longer needs to force the setting for `SHOW DATABASES` (the rewrite to `system.databases` already returns the right result). The HACK is retained only for `SHOW TABLES FROM <datalake_catalog>` which still needs it.
- Updated `test_database_glue` integration test to assert that `system.databases` shows the catalog even with the setting off.

Followup to: https://github.com/ClickHouse/ClickHouse/pull/89914

- Improvement

### Changelog entry
`SELECT * FROM system.databases` now always lists data lake catalog databases regardless of the `show_data_lake_catalogs_in_system_tables` setting. Previously they were hidden by default, which was inconsistent with `SHOW DATABASES` that always showed them.

### Version info
- Merged into: `26.5.1.116`


alsugiliazova:
@thevar1able fails look like CI issues, can you please check? Should I merger master again?

thevar1able:
>Should I merger master again?

Yes, please, this should also resolve `./ci/jobs/scripts/workflow_hooks/pr_body_check.py` failure.

PR #103447, opened 2026-04-23
Partition pruning fix with date transforms

### LINKED ISSUE #103433: Iceberg Partition Pruning not properly working with toDate function
### Company or project name

_No response_

ClickHouse doesn't leverage iceberg partition key properly when using `toDate` function (unsure if limited to this function), even with `use_iceberg_partition_pruning = 1`

25.8 LTS

### How to reproduce

Have an Iceberg table partition by transformed column `days(created_at`
```sql
CREATE TABLE IF NOT EXISTS my_db.my_table (
                created_at TIMESTAMP NOT NULL,
                error STRING
            )
USING iceberg
PARTITIONED BY (days(created_at))
```

ClickHouse side

```sql
/* Enable partition pruning on ClickHouse */
SET use_iceberg_partition_pruning = 1;

/* Pruning */
SELECT count()
FROM my_catalog."my_db.my_table"
WHERE (created_at >= '2026-03-01') AND (created_at < '2026-03-02')
/*
1 row in set. Elapsed: 0.225 sec. Processed 1.25 million rows, 89.13 MB (5.57 million rows/s., 396.73 MB/s.)
Peak memory usage: 8.51 MiB.
*/

/* Pruning */
SELECT count()
FROM my_catalog."my_db.my_table"
WHERE toYYYYMMDD(created_at) = 20260302
/*
1 row in set. Elapsed: 0.217 sec. Processed 1.25 million rows, 89.13 MB (5.78 million rows/s., 411.65 MB/s.)
Peak memory usage: 10.51 MiB.
*/

/* Not pruning FULL SCAN */
SELECT count()
FROM my_catalog."my_db.my_table"
WHERE toDate(created_at) = toDate('2026-03-02')
/*
1 row in set. Elapsed: 6.900 sec. Processed 145.53 million rows, 14.79 GB (21.09 million rows/s., 2.14 GB/s.)
Peak memory usage: 33.85 MiB.
*/
```

### Expected behavior

_No response_

_No response_

### Additional context

_No response_

scanhex12:

Partition pruning fix with date transforms. This closes https://github.com/ClickHouse/ClickHouse/issues/103433


src/Storages/ObjectStorage/DataLakes/Iceberg/ManifestFilesPruning.cpp
```diff
@@ -51,7 +53,130 @@ DB::ASTPtr getASTFromTransform(const String & transform_name_src, const String &
...
-std::unique_ptr<DB::ActionsDAG> ManifestFilesPruner::transformFilterDagForManifest(const DB::ActionsDAG * source_dag, std::vector<Int32> & used_columns_in_filter) const
+struct FunctionSubstitutionEntry
+{
+    String source_func;
+    String target_func;
+    std::function<bool(const DataTypePtr &)> accepts_const_type;
+    std::function<const ActionsDAG::Node *(ActionsDAG &, const ActionsDAG::Node *, const ContextPtr &)> convert_const;
+};
+
+static std::vector<FunctionSubstitutionEntry> buildSubstitutionTable(const ContextPtr & context)
```
antonio2368, line 64: Something AI found:
This rewrite is not safe for range pruning on Iceberg historical temporal values. Iceberg `date` is imported as `Date32`, and `timestamp` / `timestamptz` are imported as `DateTime64` (`SchemaProcessor.cpp:301-308`), so pre-1970 values are valid input for manifest pruning. But the replacement functions used here are the standard unsigned relative transforms: `toRelativeDayNum` is registered as `UInt32` and its standard `Date32` path casts through `UInt16`, while `toRelativeHourNum` is registered as `UInt32` and its standard historical paths cast through unsigned values (`toRelativeDayNum.cpp:10`, `toRelativeHourNum.cpp:10`, `DateTimeTransforms.h:2188-2204`, `2220-2240`).

That does not preserve ordering across the Unix epoch. For example:

```sql
SELECT
    toDate32('1969-12-31') AS d1,
    toDate32('1970-01-01') AS d2,
    toRelativeDayNum(d1) AS r1,
    toRelativeDayNum(d2) AS r2,
    r1 < r2 AS relative_order_ok;
```

returns `r1 = 65535`, `r2 = 0`, and `relative_order_ok = 0`. Similarly, `toRelativeHourNum(toDateTime64('1969-12-31 23:00:00', 0, 'UTC'))` is greater than the value for `1970-01-01 00:00:00`.

So a range predicate over `toDate` / `toStartOfHour` can be converted into a predicate whose ordering is different from the original one, and `KeyCondition` can then prune a manifest that actually contains matching rows. Equality for post-1970 happy paths is not enough to justify using these substitutions for ranges.

Please either skip these substitutions when the source domain can include pre-1970 `Date32` / `DateTime64`, or use a signed/extended transform and only enable range rewrites after proving the replacement is monotonic-equivalent for the source and constant domains.

src/Storages/ObjectStorage/DataLakes/Iceberg/ManifestFilesPruning.cpp
```diff
@@ -51,7 +53,130 @@ DB::ASTPtr getASTFromTransform(const String & transform_name_src, const String &
...
+        if (!builder)
+            return nullptr;
+        ColumnsWithTypeAndName arguments{{const_node->column, const_node->result_type, const_node->result_name}};
+        auto func_base = builder->build(arguments);
+        if (!func_base)
+            return nullptr;
+        return &dag.addFunction(func_base, {const_node}, const_node->result_name);
+    };
+
+    entries.push_back({"toDate", "toRelativeDayNum",
```
antonio2368, line 80: toDate accepts strings, while toRelativeDayNum doesn't. You should probably check argument types before transforming.
Potentially same is true for other parts.

src/Storages/ObjectStorage/DataLakes/Iceberg/ManifestFilesPruning.cpp
```diff
@@ -51,7 +53,130 @@ DB::ASTPtr getASTFromTransform(const String & transform_name_src, const String &
...
+        {
+            return apply_func_to_const("toRelativeHourNum", dag, sibling);
+        }
+    });
+
+    entries.push_back({"toYear", "toYearNumSinceEpoch",
+        [](const DataTypePtr & t) { return isNumber(t); },
+        [](ActionsDAG & dag, const ActionsDAG::Node * sibling, const ContextPtr &) -> const ActionsDAG::Node *
+        {
+            auto year_val = (*sibling->column)[0].safeGet<UInt64>();
```
antonio2368, line 100: What gh reviewer said is true. You probably need to restrict it more.

src/Storages/ObjectStorage/DataLakes/Iceberg/ManifestFilesPruning.cpp
```diff
@@ -51,7 +53,130 @@ DB::ASTPtr getASTFromTransform(const String & transform_name_src, const String &
...
+{
+    auto substitutions = buildSubstitutionTable(context);
+
+    std::unordered_map<String, size_t> source_to_idx;
+    for (size_t i = 0; i < substitutions.size(); ++i)
+        source_to_idx.emplace(substitutions[i].source_func, i);
+
+    std::unordered_map<const ActionsDAG::Node *, size_t> substituted_nodes;
+
+    for (auto & node : const_cast<ActionsDAG::Nodes &>(dag.getNodes()))
```
antonio2368, line 120: Not sure if this is okay. I think you need to use ActionsDAG::addFunction.

antonio2368 (commented):
It seems like a really tricky thing to do. We should probably have more tests with as many edge cases as possible.
AI should be good in finding those edge cases.

src/Storages/ObjectStorage/DataLakes/Iceberg/ManifestFilesPruning.cpp
```diff
@@ -51,7 +53,130 @@ DB::ASTPtr getASTFromTransform(const String & transform_name_src, const String &
...
+        node.function_base = std::move(new_function_base);
+        substituted_nodes.emplace(&node, it->second);
+    }
+
+    if (substituted_nodes.empty())
+        return;
+
+    for (auto & node : const_cast<ActionsDAG::Nodes &>(dag.getNodes()))
+    {
+        if (node.type != ActionsDAG::ActionType::FUNCTION || node.children.size() != 2)
```
antonio2368, line 153: This seems a bit scary, what if it's not comparison operator? Even for comparison operators there can be strange edge cases:
`WHERE toStartOfHour(ts) < toDateTime('2024-01-10 10:30:00')` becomes `toRelativeHourNum(ts) < toRelativeHourNum('2024-01-10 10:30:00')` which would exclude some values that it didn't exclude before.

scanhex12:
I'll close this PR and continue working here as this PR is too old: https://github.com/ClickHouse/ClickHouse/pull/118606

PR #93813, opened 2026-01-09
Do not recursively split always-true ranges in PK analysis

EmeraldShift:
### Motivation
For very large tables, with queries that select one or more large, consecutive ranges of the table, the generic exclusion search algorithm can take a significant fraction of the total time. This is because for each candidate range, it only considers whether the key condition can be true within that range, otherwise it discards the range. But if the range _cannot be false_, the algorithm currently splits the range into `merge_tree_coarse_index_granularity` pieces, and recursively checks them. This leads to O(N * log N) evaluations of the key condition against the inner marks.

It turns out KeyCondition already supports checking if the condition can be false within a range. This PR enables the check, and exits the algorithm early if the selected range cannot be false.

Compare performance:
```
CREATE OR REPLACE TABLE excl (a int, b int)
ENGINE = MergeTree() ORDER BY (a, b)
SETTINGS index_granularity = 1
AS SELECT 1, 1 FROM numbers(1_000_000);

-- Query
EXPLAIN indexes = 1
SELECT * FROM excl WHERE b = 1
SETTINGS max_threads = 1, send_logs_level = 'trace'

-- Base:
<Trace> default.excl (SelectExecutor): Used generic exclusion search over index for part all_1_1_0 with 1379234 steps
...
0 rows in set. Elapsed: 0.422 sec.

-- PR:
<Trace> default.excl (SelectExecutor): Used generic exclusion search over index for part all_1_1_0 with 1 steps
...
0 rows in set. Elapsed: 0.003 sec.
```

Maybe there are some cases where it is too expensive to check if a range can be false. I claim we should detect those cases with heuristics and selectively disable this optimization.

Part of https://github.com/ClickHouse/ClickHouse/issues/92780

### Changelog category (leave one):
- Performance Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Improved the performance of generic exclusion search for queries that select large consecutive ranges of the table.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.6.1.269`

PR #103455, opened 2026-04-23
Skip granule-level minmax index when the part-level minmax already covers it

### LINKED ISSUE #103292: Skip granule-level minmax index evaluation when part-level index covers it
### Company or project name

_No response_

### Describe the situation

In certain cases, when a part-level minmax index (over a column referenced in the partition key) determines that the condition _cannot be false_ for any value in the part, ClickHouse currently still evaluates any minmax indexes over the same column(s) per granule, which wastes computation.

All

### How to reproduce

Fiddle: https://fiddle.clickhouse.com/79738f5f-c00c-465b-a8f9-9bb5e95f5fad
Suppose there is a table with a column in the partition key (traditionally, this will be something like `timestamp`, but here's a simpler example:
```
CREATE TABLE t
(
  c int,
  INDEX i c TYPE minmax
)
ENGINE = MergeTree()
ORDER BY tuple()
PARTITION BY indexHint(c)
SETTINGS index_granularity = 1
```
Then, fill it with some range of data:
```
INSERT INTO t
SELECT 123 + number FROM numbers(1_000_000)
```
Now, a query where the granule-level minmax index drops no granules, but ~99% of the time is spent evaluating it:
```
SELECT count() FROM t WHERE c > 100;
```

For this example (and more realistic examples similar to it, like Observability), there are whole parts where the granule-level minmax index must be true -- e.g. because the part-level min and max are strictly contained within the key condition -- and the granule-level index is not necessary to check.

### Expected performance

There are key conditions for which we can assert from the part-level minmax index that all granules must be true, and for those parts, the granule-level minmax index should not be checked.

### Additional context

This is very common in Observability. For example a table like
```
CREATE TABLE logs (..., INDEX idx_time timestamp TYPE Minmax, ...)
PARTITION BY toYYYYMMDD(timestamp)
ORDER BY (environment, system, user, service, ..., timestamp)
```
the `idx_time` index serves two purposes:
1. Pruning granules for time-range-restricted queries. Here, even the inclusion in `ORDER BY` is too loose an approximation, because generic exclusion search will assume a range of `(-inf, *)` or `(*, inf)` across changing prefixes of the PK columns.
2. Supporting TopK skip index dynamic filters via `use_skip_indexes_for_top_k`, introduced in a recent ClickHouse version.

Queries with large time ranges like `timestamp >= now() - INTERVAL 30 DAY` will match many parts, but for all but the most recent of these parts, the condition will be fully _covering_: some part whose data ranges between 27-and-28 days ago is surely going to match the above condition in all granules. ClickHouse should skip evaluating `idx_time` within that part, and save potentially seconds of work (large parts may contain millions of granules to check).

raimannma:
- Performance Improvement

Skip per-granule evaluation of a user-declared `minmax` skip index on a part when the part-level minmax over the partition-key columns already proves the filter is true for the entire part. Speeds up queries on observability-style tables such as `INDEX idx_time timestamp TYPE minmax PARTITION BY toYYYYMMDD(timestamp)` when the time range covers whole parts.

Fixes #103292.

No user-facing API change, transparent performance improvement, no new settings or syntax.


src/Storages/MergeTree/MergeTreeIndexMinMax.cpp
```diff
@@ -204,6 +204,14 @@ bool MergeTreeIndexConditionMinMax::mayBeTrueOnGranule(MergeTreeIndexGranulePtr
     return condition.checkInHyperrectangle(granule.hyperrectangle, index_data_types, {}, update_partial_disjunction_result_fn).can_be_true;
 }
 
+bool MergeTreeIndexConditionMinMax::alwaysTrueOnHyperrectangle(const std::vector<Range> & hyperrectangle) const
+{
+    if (hyperrectangle.size() != index_data_types.size())
+        return false;
+    const auto mask = condition.checkInHyperrectangle(hyperrectangle, index_data_types, {}, {});
```
EmeraldShift: One thing worth calling out on this specific call: the index condition uses `UNKNOWN = {.can_be_true = true, .can_be_false = true}` for any RPN leaf that references a column not in the skip index's `key_columns`, and that `can_be_false = true` propagates through `AND`. As a result, `alwaysTrueOnHyperrectangle` will only return `true` when the query's WHERE predicate is entirely on the indexed columns.

On the kinds of queries the PR description motivates:                                                                                                                                                  

      INDEX idx_time timestamp TYPE minmax PARTITION BY toYYYYMMDD(timestamp)                                                                                                                   
      -- WHERE timestamp >= T1 AND timestamp < T2                               <- fires                                                                                                          
      -- WHERE timestamp >= T1 AND timestamp < T2 AND service = 'foo'           <- doesn't fire                                                                                                   
      -- WHERE timestamp >= T1 AND timestamp < T2 AND level IN ('warn','error') <- doesn't fire                                                                                                   
                                                                                                                                                                                               The second and third queries are the common observability shape. With the current `checkInHyperrectangle`, `idx_time`'s per-index RPN looks like `IN_RANGE AND IN_RANGE AND UNKNOWN`, which evaluates to  `{.can_be_true = true, .can_be_false = true}` even when both time leaves are proven to be true on the part's hyperrectangle, so `can_be_false` stays `true` and the shortcut silently stops firing.                 

One way around this is to treat `FUNCTION_UNKNOWN` as the AND-identity `{.can_be_true = true, .can_be_false = false}` -- which is incorrect in general, but _almost fine_ in the context of deciding this optimization: the one gotcha is `FUNCTION_NOT`, which would invert this sentinel to have `.can_be_true = false` which is unsound, so it might be worth guarding against this. I have a commit https://github.com/ClickHouse/ClickHouse/commit/c0b022e5ea014443fe069af028eb0b002a45f288 that hacks this up. It's fine to leave for later though.

EmeraldShift:
Oh no! Why was the PR closed?

raimannma:
> Oh no! Why was the PR closed?

Doesn't https://github.com/ClickHouse/ClickHouse/pull/103480 already cover it?

EmeraldShift:
Ah, I think you're right! That PR covers the case when a column is in the partition key, primary key, and minmax index, which is enough for the original issue.

This PR covers additional cases when the column is not in the primary key, which might be useful, but I'm unsure how common that is.

raimannma:
I opened it again and we'll see when someone from CH reviews what they think about it.

UnamedRus:
> This PR covers additional cases when the column is not in the primary key, which might be useful, but I'm unsure how common that is.

It's bit more wide, so it should improve PK scan by making more correct guesses for pk index values in it.
In current form, it does this if column match from partition minmax index. Theoretically, it can support even skip indexes.

This problem can be also fixed from improve of query conditions, (which is more in line of #103455 approach). 
Maybe, this approach is bit more complex to implement (because it need to correctly process OR/AND condition dependencies), but in general, i think it worth to tighten our guesses from all sides, when we have capacity.

Also, current pr also about skipping extra indexes, if their guesses already covered, which will not be part of my work.

EmeraldShift:
Hey @alexey-milovidov, I'm curious if you're interested in merging this kind of optimization, which feels similar to #93813 -- it is about skipping index analysis work when we can prove it is not needed. In real-world Observability use cases, we see queries take 5-10x longer sometimes just because they're uselessly spinning on minmax evaluation. It seems good to skip this work entirely even if it can get faster (e.g. #103526)

PR #103457, opened 2026-04-23
Fix potential out-of-bounds read in the `ALP` codec decompression

rienath:

Fix potential out-of-bounds read in the `ALP` codec decompression path when processing malformed input with an invalid bit-width.

### Version info
- Merged into: `26.5.1.267`
- Backported to: `26.4.1.1137`, `26.3.10.59`


tests/queries/0_stateless/04104_alp_bit_width.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+CUR_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=../shell_config.sh
+. "$CUR_DIR"/../shell_config.sh
+
+printf '\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\234\032\0\0\0\0\040\0\0\001\010\0\004\0\0\0\0A\0\0\0\0\0\0\0\0' \
+    | ${CLICKHOUSE_CURL} "${CLICKHOUSE_URL}&decompress=1&http_native_compression_disable_checksumming_on_decompress=1" --data-binary @- \
+    | grep -c "Cannot decompress ALP-encoded data, invalid bit-width: 65, max allowed: 64"
```
Michicosun: Let's not grep the concrete error message and grep only the error code?

Michicosun (approved):

tests/queries/0_stateless/04104_alp_bit_width.sh
rienath: Thanks, fixed. One reason to do this is to make sure test passes if we change error message. Are there any other ones that I am missing?

PR #103463, opened 2026-04-23
Improve error message for S3 `listObjects` access denied errors

nickitat:

### Version info
- Merged into: `26.5.1.139`

### DIFF
--- a/src/Disks/DiskObjectStorage/ObjectStorages/AzureBlobStorage/AzureObjectStorage.cpp
+++ b/src/Disks/DiskObjectStorage/ObjectStorages/AzureBlobStorage/AzureObjectStorage.cpp
@@ -61,13 +61,17 @@ public:
     AzureIteratorAsync(
         const std::string & path_prefix,
         std::shared_ptr<const AzureBlobStorage::ContainerClient> client_,
-        size_t max_list_size)
+        size_t max_list_size,
+        const std::string & container_,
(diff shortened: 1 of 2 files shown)


src/Disks/DiskObjectStorage/ObjectStorages/AzureBlobStorage/AzureObjectStorage.cpp
```diff
@@ -87,34 +91,47 @@ class AzureIteratorAsync final : public IObjectStorageIteratorAsync
...
+            if (!blob_list_response.NextPageToken.HasValue() || blob_list_response.NextPageToken.Value().empty())
+                return false;
+
+            options.ContinuationToken = blob_list_response.NextPageToken;
+            return true;
+        }
+        catch (const Azure::Storage::StorageException & e)
+        {
+            throw Exception(
+                ErrorCodes::AZURE_BLOB_STORAGE_ERROR,
```
kssenii: This change seems to affect this code https://github.com/ClickHouse/ClickHouse/blob/347e51be2bfa5976f22449d7e231272953433fee/src/Disks/DiskObjectStorage/MetadataStorages/PlainRewritable/MetadataStorageFromPlainRewritableObjectStorage.cpp#L220-L225
e.g. it will not longer be processed as expected

PR #103464, opened 2026-04-23
Optimize MemoryTracker layout (improves performance of it ~25%)

azat:
- Performance Improvement

Optimize MemoryTracker layout (improves performance of it ~25%)

<details>

<summary>Results</summary>

```sql
# tests/performance/group_by_multiple_strings.xml
$ clickhouse-benchmark -i 100 -q "select key1_8, key2_8, sum(value) from test_group_by_strings group by key1_8, key2_8 format Null settings max_threads = 32, query_profiler_cpu_time_period_ns=0, query_profiler_real_time_period_ns=0, memory_profiler_step=0, memory_profiler_sample_probability=0, max_untracked_memory='0'"

# before
Queries executed: 97 (97%).

localhost:9000, queries: 24, QPS: 23.728, RPS: 71183368.874, MiB/s: 2036.573, result RPS: 0.000, result MiB/s: 0.000.

0%              0.038 sec.
10%             0.039 sec.
20%             0.041 sec.
30%             0.041 sec.
40%             0.041 sec.
50%             0.042 sec.
60%             0.042 sec.
70%             0.042 sec.
80%             0.042 sec.
90%             0.043 sec.
95%             0.043 sec.
99%             0.043 sec.
99.9%           0.043 sec.
99.99%          0.043 sec.

# after
Queries executed: 100 (100%).

localhost:9000, queries: 100, QPS: 27.094, RPS: 81282582.316, MiB/s: 2325.513, result RPS: 0.000, result MiB/s: 0.000.

0%              0.032 sec.
10%             0.033 sec.
20%             0.034 sec.
30%             0.034 sec.
40%             0.034 sec.
50%             0.035 sec.
60%             0.035 sec.
70%             0.036 sec.
80%             0.036 sec.
90%             0.036 sec.
95%             0.037 sec.
99%             0.039 sec.
99.9%           0.041 sec.
99.99%          0.041 sec.
```

</details>

### DIFF
--- a/src/Common/MemoryTracker.h
+++ b/src/Common/MemoryTracker.h
@@ -66,20 +66,22 @@ private:
 #pragma clang diagnostic ignored "-Wgnu-anonymous-struct"
 #pragma clang diagnostic ignored "-Wnested-anon-types"
 
-    /// Hot writes: updated on every alloc/free flush. Isolated on their own cache line
-    /// so that flushes from any core do not invalidate the read-mostly line below.
+    /// Hot: every field touched on the alloc/free path. Packed into a single cache line
+    /// (24B writes + 40B reads = 64B exactly) so each allocation needs at most one line
+    /// in flight. The line bounces among writer cores either way; co-locating the
+    /// read-mostly limits/parent/metric here means writers get them for free with the
+    /// same fetch. The only cost is reader-only cores re-fetching when allocator writes
+    /// invalidate, but allocations dominate over pure reads in practice.
     struct
     {
+        /// Hot writes: updated on every alloc/free flush.
         alignas(DB::CH_CACHE_LINE_SIZE) std::atomic<Int64> amount {0};
         std::atomic<Int64> peak {0};
         std::atomic<Int64> rss {0};
-    };
 
-    /// Read-mostly: read on every alloc to enforce caps and walk the tracker chain,
-    /// but written rarely (set once at construction or via configuration changes).
-    struct
-    {
-        alignas(DB::CH_CACHE_LINE_SIZE) std::atomic<Int64> soft_limit {0};
+        /// Hot reads: checked on every alloc but written only at construction or via
+        /// rare configuration changes.
+        std::atomic<Int64> soft_limit {0};
         std::atomic<Int64> hard_limit {0};
         std::atomic<Int64> profiler_limit {0};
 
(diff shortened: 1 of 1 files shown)


alexey-milovidov (approved):

src/Common/MemoryTracker.h
```diff
@@ -53,53 +55,78 @@ namespace DB
...
+    /// Hot: every field touched on the alloc/free path. Packed into a single cache line
+    /// (24B writes + 40B reads = 64B exactly) so each allocation needs at most one line
+    /// in flight. The line bounces among writer cores either way; co-locating the
+    /// read-mostly limits/parent/metric here means writers get them for free with the
+    /// same fetch. The only cost is reader-only cores re-fetching when allocator writes
+    /// invalidate, but allocations dominate over pure reads in practice.
+    struct
+    {
+        /// Hot writes: updated on every alloc/free flush.
+        alignas(DB::CH_CACHE_LINE_SIZE) std::atomic<Int64> amount {0};
```
monoid, line 78: A cache coherence size can be larger than a cache line size.  Other project uses 128 bytes on x86_64, for example.

azat: Thanks, this is interesting, but I doubt that it will make any difference, since here we have cold and hot part, hot fits into 64 bytes and cold accessed rarely

Also if we will change it, we need to do some benchmarks, AFAICS it can be helpful for Intel, but not sure about AMD. Though we do not use it for small frequent structures, and if it will help for some CPUs it can be OK to do this unconditionally for amd64.

azat: I've also tested this changes on Intel XEON, and I don't see the difference - https://pastila.nl/?0012525f/220080b545e4afff7e27c25307341e69#gr1F9PIgEX0ADOBhsmMxFg==GCM

Neither for 64 nor 128 alignment

monoid: I see. I've tried to find a reliable source for my original statement, and it seems to be more complicated: the cache coherence granule is still 64 bytes, not 128, but Intel's spatial prefetcher may load 128 bytes into cache, inducing a false sharing of 128 bytes.

https://github.com/facebook/folly/blob/1b5288e6eea6df074758f877c849b6e73bbb9fbb/folly/lang/Align.h#L107

It might be just access-pattern dependent.

monoid: There is MSR setting called "Adjacent Cache Line Prefetcher Disable" or "L2 Adjacent Cache Line Prefetcher Disable" that controls this behavior (see Intel SDM).

PR #103468, opened 2026-04-23
Use per-column indexes for `coalesce(a, b) <op> const` predicates

### LINKED ISSUE #103194: `coalesce(colA, colB) = X` cannot use PK or skip index on either input column
### Use case

We have a table where two different columns can hold the "same" logical value depending on row type, and we expose a unified ALIAS column for querying:

```sql
CREATE TABLE t (
    ts   DateTime,
    a    Nullable(String),
    b    String,
    host String ALIAS coalesce(a, b)
) ENGINE = MergeTree ORDER BY (ts, b);
```

Filtering on the alias (`WHERE host = 'X'`) reads the right rows, but uses no index on either `a` or `b` — it degrades to a scan of whatever the other PK columns selected.

### What happens today

`EXPLAIN indexes=1` shows `coalesce(a, b) = 'X'` produces `PrimaryKey Condition: true` regardless of whether the PK is on `a` or `b`, and per-column skip indexes on `a`/`b` are not consulted at all. Only a skip index defined on the exact `coalesce(a, b)` expression is used.

#94754 added monotonicity to `coalesce`/`ifNull`/`assumeNotNull` so that `coalesce(key_col, <constant>)` can prune via PK/skip index. But the two-column form `coalesce(colA, colB)` is not covered, because `KeyCondition::isKeyPossiblyWrappedByMonotonicFunctionsImpl` requires one of the two arguments to be a constant:

https://github.com/ClickHouse/ClickHouse/blob/46186d92e485974d20d2c7b743b91641ddf799ff/src/Storages/MergeTree/KeyCondition.cpp#L1908-L1936

(and `arguments_size > 2` bails out unconditionally, so 3+ arg `coalesce` never qualifies either)

### Reproduction

```sql
CREATE TABLE t
(
    k UInt32,
    a Nullable(String),
    b String,
    c String ALIAS coalesce(a, b),
    INDEX idx_a a TYPE bloom_filter GRANULARITY 1,
    INDEX idx_b b TYPE bloom_filter GRANULARITY 1
)
ENGINE = MergeTree ORDER BY k SETTINGS index_granularity = 1024;

INSERT INTO t
SELECT number,
       if(number % 3 = 0, NULL, concat('a', toString(intDiv(number, 2000)))),
       concat('b', toString(intDiv(number, 2000)))
FROM numbers(1000000);

-- direct column filters: skip indexes prune to ~6-8/977 granules
EXPLAIN indexes=1 SELECT count() FROM t WHERE a = 'a42';
EXPLAIN indexes=1 SELECT count() FROM t WHERE b = 'b42';

-- coalesce: NO Skip stage at all, PrimaryKey Condition: true, 977/977 granules
EXPLAIN indexes=1 SELECT count() FROM t WHERE c = 'a42';
EXPLAIN indexes=1 SELECT count() FROM t WHERE coalesce(a, b) = 'a42';

-- manual OR rewrite: combined skip-index OR-merge prunes to ~60/977
EXPLAIN indexes=1 SELECT count() FROM t WHERE (a = 'a42' OR (a IS NULL AND b = 'a42'));
```

### Proposal

Rewrite `coalesce(a, b) = X` (and `ifNull(a, b) = X`) to `(a = X) OR (a IS NULL AND b = X)` before index analysis. ClickHouse already merges per-column skip indexes across OR (the manual rewrite above prunes correctly), so this would let skip indexes on `a` and/or `b` apply.

For the primary key the rewrite is less powerful — an OR with a non-PK column can't bound a PK range — but it's still strictly no worse than today, and it does help when *both* `a` and `b` happen to be PK columns or have skip indexes.

The same applies to `if(cond, a, b) = X` → `(cond AND a = X) OR (NOT cond AND b = X)` when `cond` references only one column.

### Alternatives considered

- Skip index on the `coalesce(a, b)` expression: works, but only when the result is correlated with sort order. When `a` is uncorrelated, the bloom filter sees the target value in nearly every granule and prunes nothing.
- Manual OR in queries: works for skip indexes, but defeats the purpose of the ALIAS abstraction.

raimannma:
Closes #103194.

Rewrites `coalesce(a_1, ..., a_N) <op> const` and `ifNull(...) <op> const` predicates at `KeyCondition` time into

```
(y_1 <op> const)
OR (isNull(y_1) AND y_2 <op> const)
OR ...
OR (isNull(y_1) AND ... AND isNull(y_{M-1}) AND y_M <op> const)
[OR (isNull(y_1) AND ... AND isNull(y_M) AND (c <op> const))]
```

so per-column primary-key and skip indexes on each `y_i` can prune granules through the existing `use_skip_indexes_for_disjunctions` path. The arguments are normalized mirroring `FunctionCoalesce::getReturnTypeImpl` (drop NULL literals, truncate at first non-Nullable), so the rewrite is always semantically equivalent to the original `coalesce`. Only `equals`, `less`, `greater`, `lessOrEquals`, `greaterOrEquals` are rewritten (and their operand-swapped mirrors); other ops (`notEquals`, `IS NULL`, …) fall through unchanged.

Gated behind `allow_key_condition_coalesce_rewrite` (on by default). The rewrite is strictly additive for index pruning; runtime filtering still uses the original predicate.

- Performance Improvement

Rewrite `coalesce(a, b, ...) <op> const` and `ifNull(a, b) <op> const` predicates before index analysis so per-column primary key and skip indexes on each argument can prune granules. Controlled by the new setting `allow_key_condition_coalesce_rewrite` (on by default).

### Version info
- Merged into: `26.5.1.62`


alexey-milovidov (approved):
Interesting, and looks reasonable.

PedroTadim:
Hello! Can a performance test be added?

PR #103480, opened 2026-04-23
Use partition minmax index to improve primary key pruning

UnamedRus:
- Performance Improvement

Use partition minmax index bounds to prune more granules during primary key analysis for `MergeTree` tables, when a primary key column is also an input column of the partition key. For example, in a `MergeTree` table with `ORDER BY (id, event_time)` and `PARTITION BY toYYYYMM(event_time)`, ClickHouse will use the partition minmax index on `event_time` during primary key index analysis to make more informed granule-pruning decisions. Controlled by the new setting `use_partition_minmax_for_primary_key_pruning` (enabled by default).

```sql

EXPLAIN ESTIMATE
SELECT *
FROM test_order_by
WHERE ((id = 100) AND ((event_time >= toDateTime('2024-01-01 23:00:00')) AND (event_time <= toDateTime('2024-01-03 01:00:00')))) OR ((id = 200) AND ((event_time >= toDateTime('2024-01-04 23:00:00')) AND (event_time <= toDateTime('2024-02-06 01:00:00')))) OR ((id = 300) AND ((event_time >= toDateTime('2024-01-07 23:00:00')) AND (event_time <= toDateTime('2024-03-09 01:00:00')))) OR ((id = 400) AND ((event_time >= toDateTime('2024-01-10 23:00:00')) AND (event_time <= toDateTime('2024-04-12 01:00:00')))) OR ((id = 500) AND ((event_time >= toDateTime('2024-01-13 23:00:00')) AND (event_time <= toDateTime('2024-05-15 01:00:00')))) OR ((id = 600) AND ((event_time >= toDateTime('2024-01-15 23:00:00')) AND (event_time <= toDateTime('2024-06-18 01:00:00'))))

Before
   ┌─database─┬─table─────────┬─parts─┬──rows─┬─marks─┐
1. │ default  │ test_order_by │    10 │ 33280 │    65 │
   └──────────┴───────────────┴───────┴───────┴───────┘


After

   ┌─database─┬─table─────────┬─parts─┬─rows─┬─marks─┐
1. │ default  │ test_order_by │    10 │ 8192 │    16 │
   └──────────┴───────────────┴───────┴──────┴───────┘
```

### Version info
- Merged into: `26.7.1.496` (included in `26.7` and later)


UnamedRus:
@nihalzp, looks like failures not related.

nihalzp (commented):
If I understand correctly, this optimization is only applicable when primary key suffix columns are dropped because of low value of `primary_key_ratio_of_unique_prefix_values_to_skip_suffix_columns`. Then we try to use conservative bounds to suffix column instead of using `(-inf, inf)` if we happen to have the same partition key. If we never drop any suffix primary key columns, then this optimization does not help us get more pruning.

Is my understanding correct?

UnamedRus:
>  If we never drop any suffix primary key columns, then this optimization does not help us get more pruning.
> Is my understanding correct?

No, it's actually only part of when it helpful. (And actually wasn't even my main concern)

Consider case, when previous column (ie prefix) in ORDER BY changes it's value from one to another.
We also reset boundary to -inf in that case. 
And this happen much more frequently (basically every time) for high cardinality columns.

Now consider case, when we filter by many such prefix column combination with different range of "our" column, so basically we request different prefix columns from different partitions.

nihalzp:
Okay. Could we have a test where we have `primary_key_ratio_of_unique_prefix_values_to_skip_suffix_columns=1` (meaning all suffix columns are loaded regardless) and it lead to more pruning because of the PR optimization?

UnamedRus:
> Okay. Could we have a test where we have

Actually, it seems original changes wasn't enough to make it work for what was main target.
Unfortunately, for that it need to be more invasive in index analysis, but lets wait for CI and see if it catch anything. 

But improvement for original query became nicer. 

```
EXPLAIN ESTIMATE
SELECT *
FROM test_order_by
WHERE ((id = 100) AND ((event_time >= toDateTime('2024-01-01 23:00:00')) AND (event_time <= toDateTime('2024-01-03 01:00:00')))) OR ((id = 200) AND ((event_time >= toDateTime('2024-01-04 23:00:00')) AND (event_time <= toDateTime('2024-02-06 01:00:00')))) OR ((id = 300) AND ((event_time >= toDateTime('2024-01-07 23:00:00')) AND (event_time <= toDateTime('2024-03-09 01:00:00')))) OR ((id = 400) AND ((event_time >= toDateTime('2024-01-10 23:00:00')) AND (event_time <= toDateTime('2024-04-12 01:00:00')))) OR ((id = 500) AND ((event_time >= toDateTime('2024-01-13 23:00:00')) AND (event_time <= toDateTime('2024-05-15 01:00:00')))) OR ((id = 600) AND ((event_time >= toDateTime('2024-01-15 23:00:00')) AND (event_time <= toDateTime('2024-06-18 01:00:00'))))


   ┌─database─┬─table─────────┬─parts─┬─rows─┬─marks─┐
1. │ default  │ test_order_by │    10 │ 8192 │    16 │
   └──────────┴───────────────┴───────┴──────┴───────┘
   ```

**Comment by UnamedRus** [comment]
Upstream for new test: https://fiddle.clickhouse.com/923cce34-cabe-49fe-94bb-2c8c01ef15ad

File: src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp [hunk 1]
```diff
@@ -1665,6 +1665,40 @@ MarkRanges MergeTreeDataSelectExecutor::markRangesFromPKRange(
...
+            pk_column_bounds.push_back(Range::createWholeUniverseWithoutNull());
+    }
+    auto minmax_index = part->getMinMaxIndex();
+    if (minmax_index && minmax_index->initialized)
+    {
+        const auto minmax_names = MergeTreeData::getMinMaxColumns(
+            metadata_snapshot->getPartitionKey(), part->storage.getSettings(), MergeTreePartMinMaxIndexColumns::PARTITION_KEY_ONLY).getNames();
+
+        const auto & hyperrectangle = minmax_index->hyperrectangle;
+        if (hyperrectangle.size() == minmax_names.size())
```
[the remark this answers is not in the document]
UnamedRus: It looks like part minmax hyperrectangles doesn't have any information about actual column names, and mapping it by position only seems a bit of stretch.  https://github.com/ClickHouse/ClickHouse/blob/d25af7f8f5c1ef5629b13fd72f979185f306da10/src/Storages/MergeTree/IMergeTreeDataPart.h#L360

UnamedRus:
Looks like unstable test
https://performance.ci.clickhouse.com/runs/pr-103480-48a3ba9c8aac36c7f739e70105c9847791ab3842-6e9cea44c66b005ff84689d699f8ab9652ee88d1-20260608t163811z/tests/index_bulk_filtering?metrics=OSCPUVirtualTimeMicroseconds&query_index=2

Number of total parts is different (11 vs 16)

https://github.com/ClickHouse/ClickHouse/blob/79dc4059f1bedef95ed0e8baf0ab8f4dd10c1df6/tests/performance/index_bulk_filtering.xml#L4

N | Metric | Arch | Old | New | Diff↓ | Threshold
-- | -- | -- | -- | -- | -- | --
2 | SelectedParts | amd | 11 | 16 | +45.4% | 90.9%
2 | SelectedPartsTotal | amd | 11 | 16 | +45.4% | 90.9%

UnamedRus:
```
[2026-06-17 16:54:26] ERROR: Exception during CI DB query attempt 5: Failed to query CI DB. Response code: 500, Body: Code: 201. DB::Exception: Quota for user 
                      `play` for 3600s has been exceeded: read_rows = 300061254051/300000000000. Interval will end at 2026-06-17 17:33:21. Name of quota 
                      template: `explorer_normalized_query_hash`. (QUOTA_EXCEEDED) (version 26.6.1.114)

[2026-06-17 16:54:26] WARNING: Failed to get previously failed tests (best effort): Failed to query CI DB. Response code: 500, Body: Code: 201. DB::Exception: 
                      Quota for user `play` for 3600s has been exceeded: read_rows = 300061254051/300000000000. Interval will end at 2026-06-17 17:33:21. Name 
                      of quota template: `explorer_normalized_query_hash`. (QUOTA_EXCEEDED) (version 26.6.1.114)
```


Looks, like CI user need quotas fix for ClickHouse CI server
@thevar1able
AFAIK, you worked a lot on ci.
Sorry, if i'm wrong

thevar1able:
@UnamedRus I think this change should help https://github.com/ClickHouse/ClickHouse/pull/107762
cc @azat

UnamedRus:
Ah, it was fixed already, i did quick search by keywords, but missed that

nihalzp (commented):
Thanks! The idea makes sense! I have refactored the implementation to generalize it to allow upcoming `KeyCondition` optimizations to be more easily integrated.

I have also fixed one of the bugs with reverse keys and gated the optimization behind a setting.

UnamedRus:
Ok, can i consider that you will take it over from now?

Thanks.

nihalzp:
> Ok, can i consider that you will take it over from now?

Yes.

nihalzp (approved):

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
```diff
@@ -1724,9 +1724,19 @@ MarkRanges MergeTreeDataSelectExecutor::markRangesFromPKRange(
...
-    /// such a part has no usable partition-minmax bounds.
-    const auto part_minmax_index = pk_to_minmax_slot ? part->getMinMaxIndex() : nullptr;
+    /// If until index 4 of PK key columns is used in the filter, then used_key_prefix_size would be 5.
+    /// There is no need to process later key columns.
+    const size_t used_key_prefix_size = key_condition.getUsedKeyPrefixSize();
+
+    /// Do not touch the part's minmax index unless some key column the filter uses has a partition-minmax
+    /// bound to consume: `getMinMaxIndex` may lazy-load `minmax_*.idx` from disk, and a bound at a key
+    /// position the filter never references cannot affect the analysis. The minmax index may also be
+    /// absent or uninitialized (e.g. for projection parts); such a part has no usable bounds either.
```
UnamedRus, line 1734: Well, for projection parts, it probably can be sourced from "original" part, which should be? loaded in most of cases

nihalzp: Yes, that's true. However, it is tricky to implement and ensure correctness because of aliases and name-based matching that we currently use in `buildPrimaryKeyToMinMaxSlotMapping`. This PR is getting big and is already complicated enough. Let's do the improvement in a follow-up PR.

PedroTadim:
Hello! A performance test could be added?

nihalzp:
@PedroTadim The optimization results in more pruning (as a result query runs faster) when the setting is turned on as we can see in `04356_partition_minmax_for_primary_key_pruning.reference`.

PR #100005, opened 2026-03-19
Add DETERMINISTIC keyword to WASM UDF CREATE FUNCTION syntax

bacek:
### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
WASM UDF can be declared DETERMINISTIC and be a subject of constant folding.

### Documentation entry for user-facing changes

WASM UDF can be declared DETERMINISTIC and be a subject of constant folding.
Syntax:

```
      CREATE OR REPLACE FUNCTION f
      LANGUAGE WASM FROM 'module'
      ARGUMENTS (x Int32) RETURNS Int32
      ABI BUFFERED_V1
      DETERMINISTIC;
```

PR #103487, opened 2026-04-24
Fix Cranelift stack overflow on aarch64-apple-darwin

bacek:

Fix Cranelift stack overflow on `aarch64-apple-darwin` in WasmTime WASM UDFs.

### CI report
https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=103487&sha=def42cf9729339f7ed5a9d5c1d7403804a76664b&name_0=PR


bacek:
@vdimir fyi. Updating to wasmtime 44 alone didn't fix my issue. This can be landed after 44 re-vendored. Or now. It shouldn't hurt. https://github.com/ClickHouse/ClickHouse/pull/100005#discussion_r3050248695

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
```diff
@@ -480,14 +482,55 @@ class WasmTimeModule : public WasmModule
...
+    if (int rc = pthread_create(&thread, &attr, compile_fn, &task); rc != 0)
     {
-        throw Exception(ErrorCodes::WASM_ERROR, "Failed to compile wasm code: {}", compilation_result.err().message());
+        pthread_attr_destroy(&attr);
+        throw Exception(ErrorCodes::WASM_ERROR, "pthread_create failed: {}", rc);
     }
-    auto module = compilation_result.ok();
 
-    return std::make_unique<WasmTimeModule>(module_name, impl->engine, std::move(module));
+    pthread_attr_destroy(&attr);
```
alexey-milovidov: Can we use std::unique_ptr or SCOPE_EXIT for RAII?

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
```diff
@@ -480,14 +482,55 @@ class WasmTimeModule : public WasmModule
 std::unique_ptr<WasmModule> WasmTimeRuntime::compileModule(std::string_view module_name, std::string_view wasm_code) const
 {
     std::span<uint8_t> bytes(reinterpret_cast<uint8_t *>(const_cast<char *>(wasm_code.data())), wasm_code.size());
-    auto compilation_result = wasmtime::Module::compile(impl->engine, bytes);
-    if (!compilation_result)
+
+    // Cranelift's egraph optimizer generates a huge constructor_simplify function
+    // (64KB+ of code) that recurses up to REWRITE_LIMIT=5 times during compilation.
+    // Each level pushes a large stack frame; the default 512KB thread stack overflows
```
alexey-milovidov: Does it have to be on stack? Maybe wasmtime can be fixed to avoid that?

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
```diff
@@ -480,14 +482,55 @@ class WasmTimeModule : public WasmModule
 std::unique_ptr<WasmModule> WasmTimeRuntime::compileModule(std::string_view module_name, std::string_view wasm_code) const
 {
     std::span<uint8_t> bytes(reinterpret_cast<uint8_t *>(const_cast<char *>(wasm_code.data())), wasm_code.size());
-    auto compilation_result = wasmtime::Module::compile(impl->engine, bytes);
-    if (!compilation_result)
+
+    // Cranelift's egraph optimizer generates a huge constructor_simplify function
+    // (64KB+ of code) that recurses up to REWRITE_LIMIT=5 times during compilation.
```
alexey-milovidov: Also a question - can it be more than even 8 MiB, if an attacker constructs the function?

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
bacek: I'll have a look.

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
bacek: Yes, it's possible.  Which will kill ClickHouse on load.

bacek: However, before wasmtime 44 compiling wasm udf with exceptions enabled was enough to kill ClickHouse ;)

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
bacek: I converted this PR into draft to prevent landing. Looks like there should be a cleaner solution on WasmTime side.

bacek: @alexey-milovidov I update branch with a "theoretically correct" fix, but left thread as a last line of defence, just in case.

bacek: Also, relevant changes in wasmtime are landed just an hour ago. https://github.com/bytecodealliance/wasmtime/pull/13192. I will re-test it against upstream

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
bacek: Here is Claude answer with caveat at the end:
 

> ⏺ No — and the reason is architectural, not just about the stack size.

  The recursion into constructor_simplify is bounded by REWRITE_LIMIT=5 enforced in optimize_pure_enode before each call:

  const REWRITE_LIMIT: usize = 5;
  if ctx.rewrite_depth > REWRITE_LIMIT {
      ctx.stats.rewrite_depth_limit += 1;
      return orig_value;   // early return, no recurse
  }

  This check runs in ClickHouse's Rust code, not in WASM. A WASM module controls the input values to the optimizer, not the optimizer's own recursion counter. The maximum stack depth into constructor_simplify is exactly 5 regardless of what the
  WASM contains. The overflow we fixed was triggered by legitimate arithmetic code that happened to tickle all 5 levels simultaneously — it wasn't a malicious input.

  The frame size of constructor_simplify is also a compile-time constant (determined by the ISLE rules baked into the binary), so a WASM module can't inflate it.

  One caveat worth noting separately: there may be other recursion paths in Cranelift (e.g., the elaboration phase following use-def chains) that could recurse proportional to expression tree depth. A WASM function with a very long SSA chain
  could theoretically exhaust stack there. But that's a distinct issue from what we're fixing, and WASM UDF registration requires write access to system.webassembly_modules anyway — this isn't an unauthenticated attack surface.

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
```diff
@@ -480,14 +483,68 @@ class WasmTimeModule : public WasmModule
...
+    // aarch64-apple-darwin, causing SIGILL via the OS stack guard page.
+    //
+    // Root cause: in release builds the ISLE compiler emits one large function body
+    // without splitting match arms into closures. Cranelift's `isle-split-match`
+    // Cargo feature enables that splitting (it is unconditionally on in debug builds
+    // for exactly this reason). We enable it via `rust/workspace/wasmtime/Cargo.toml`,
+    // which should eliminate the overflow once verified on aarch64-apple-darwin.
+    //
+    // This thread workaround is kept as a defensive layer until the `isle-split-match`
+    // fix is confirmed to keep the release-build frame within 512 KB at depth 5.
```
bacek: @alexey-milovidov @vdimir I did check. cranelift doesn't fit into 512kb, but does into 1M. Linux default is 8mb.

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
```diff
@@ -480,14 +483,69 @@ class WasmTimeModule : public WasmModule
 std::unique_ptr<WasmModule> WasmTimeRuntime::compileModule(std::string_view module_name, std::string_view wasm_code) const
 {
     std::span<uint8_t> bytes(reinterpret_cast<uint8_t *>(const_cast<char *>(wasm_code.data())), wasm_code.size());
-    auto compilation_result = wasmtime::Module::compile(impl->engine, bytes);
-    if (!compilation_result)
+
+    // Cranelift's ISLE-generated `constructor_simplify` is a monolithic function
+    // whose stack frame (~64 KB in release builds) multiplied by REWRITE_LIMIT=5
+    // recursion levels overflows the default 512 KB TCPHandler thread stack on
```
vdimir: Do we always end up in TCPHandler thread calling this function? I assume this should normally called from pipeline executor thread during query execution, maybe I'm wrong. Does pipeline executor thread have same issue?

vdimir: Also, if issue exists only on darwin, maybe we can put under `#ifdef`?

bacek: I'm not sure it's "darwin only" actually. Some other obscure architectures/OSes can have different default stack sizes. It's better to be safe, than sorry.

bacek: The comment in the code addresses this: `compileModule` can be called from any thread — query threads, pipeline executor threads, or any other thread that triggers WASM module loading (e.g. at `CREATE FUNCTION` time or on first use). None of these threads are guaranteed to have a large stack, so the workaround applies regardless of thread type.

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
```diff
@@ -480,14 +483,69 @@ class WasmTimeModule : public WasmModule
...
+        {
+            auto res = wasmtime::Module::compile(t.engine, t.bytes);
+            if (res)
+                t.result = res.ok();
+            else
+                t.error = res.err().message();
+        }
+        catch (...)
+        {
+            t.error = getCurrentExceptionMessage(false);
```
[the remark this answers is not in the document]
bacek: Fixed in the latest commit: the `catch (...)` block now only stores `std::current_exception()`, which is `noexcept` per C++17 §15.5.6. No string formatting or allocation happens inside the catch block.

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
```diff
@@ -480,14 +483,82 @@ class WasmTimeModule : public WasmModule
...
+    // `compileModule` can be called from any thread (query threads, pipeline executor
+    // threads, etc.), so the overflow is not limited to any particular thread type.
+    //
+    // Root cause: in release builds the ISLE compiler emits one large function body
+    // without splitting match arms into closures. Cranelift's `isle-split-match`
+    // Cargo feature enables that splitting (it is unconditionally on in debug builds
+    // for exactly this reason). We enable it via `rust/workspace/wasmtime/Cargo.toml`,
+    // which should eliminate the overflow once verified on aarch64-apple-darwin.
+    //
+    // This thread workaround is kept as a defensive layer until the `isle-split-match`
```
vdimir: A couple of questions more while I’m trying to understand this:

* Does enabling `isle-split-match` help at all, or not really? Should we keep it together with the thread-based approach?

* Is it possible that this will be fixed in Wasmtime? Should we submit an issue? Afterall it looks like issue in library if it can crash when running out of stack memory.

* And one last question: when the issue reproduces, is it only on Darwin and the “obscure architectures” you mentioned? If so, could we check ahead of time whether the current thread has enough stack memory, and only then run thread or maybe fall back to another Wasmtime backend? (I haven’t tried these, though, https://github.com/bytecodealliance/wasmtime/blob/main/winch/README.md or https://github.com/bytecodealliance/wasmtime/blob/main/pulley/README.md)

Overall, the workaround is fine, but I want to understand whether we can reduce its scope, or whether we have any path to getting rid of it in the future.

bacek: > A couple of questions more while I’m trying to understand this:
> 
> * Does enabling `isle-split-match` help at all, or not really? Should we keep it together with the thread-based approach?

It does help, afaiu, but stack is still over 0.5MB. We should keep it.

> * Is it possible that this will be fixed in Wasmtime? Should we submit an issue? Afterall it looks like issue in library if it can crash when running out of stack memory.

I don't think so, but I'll discuss it with wasmtime developers on stack management.

> * And one last question: when the issue reproduces, is it only on Darwin and the “obscure architectures” you mentioned? If so, could we check ahead of time whether the current thread has enough stack memory, and only then run thread or maybe fall back to another Wasmtime backend? (I haven’t tried these, though, https://github.com/bytecodealliance/wasmtime/blob/main/winch/README.md or https://github.com/bytecodealliance/wasmtime/blob/main/pulley/README.md)

TIL. Looks like "winch" is not suitable at all. Some kind of toy interpreter.  Don't quite understand purpose of "pulley".
 
> Overall, the workaround is fine, but I want to understand whether we can reduce its scope, or whether we have any path to getting rid of it in the future.

Let me check with wasmtime. May be there is a better soultion.

alexey-milovidov:
Merged current `master`. The branch now builds against the re-vendored Wasmtime (submodule `561b224fd058`, `cranelift-codegen` `0.132.1`) — this satisfies the precondition mentioned above (*"can be landed after 44 re-vendored"*).

Verified before pushing:
- The `isle-split-match` feature and the `cranelift-codegen` path in `rust/workspace/wasmtime/Cargo.toml` still exist and are valid against the new Wasmtime.
- `rust/workspace/Cargo.lock` is consistent — the merge added no new packages, only the direct dependency edge on `cranelift-codegen` needed for feature unification.
- Master's `compileModuleWithEngine` (built against the new Wasmtime) still uses the same `wasmtime::Module::compile(engine, bytes)` C++ API this PR wraps, so the C++ compiles unchanged.

The last green CI ran against the *old* Wasmtime, so I re-triggered CI on the merged head to validate against the re-vendored version.

The open question about reducing the scope of the thread workaround (and whether Wasmtime upstream can address the stack usage directly) is still pending your discussion with the Wasmtime developers.

alexey-milovidov:
The re-triggered CI on the previous head surfaced one real failure: the `arm_tidy` build (clang-tidy) failed on the newly added code because `pthread_t thread;` in `compileModuleWithEngine` was left uninitialized (`cppcoreguidelines-init-variables`, treated as error).

Pushed a one-line fix — value-initialize it as `pthread_t thread{};`, matching the existing convention in `ThreadWithStackSize.h` and `thread_creation_latency.cpp`. `pthread_create` writes to the handle before any read, so this is purely to satisfy the check.

I re-triggered CI on the new head (`c5c8293`); it is now running.

The open scope question (whether the thread workaround can be narrowed, and whether Wasmtime upstream can address the stack usage) is unchanged and still pending your discussion with the Wasmtime developers.

Algunenano (changes requested):
What's the repro? We've changed the stack size on Darwin to 8MB so I doubt this is necessary anymore?

Algunenano:
I'm going to assume https://github.com/ClickHouse/ClickHouse/pull/107033 fixes this and this isn't required anymore

bacek:
> I'm going to assume #107033 fixes this and this isn't required anymore

That's correct, yes.

PR #103520, opened 2026-04-24
Unify arm/amd-specific test fixtures in integration tests

alexey-milovidov:
Three integration tests branched on `is_arm()` for fixtures that did not actually need to differ between architectures. Remove the `is_arm()` checks and unify the fixtures.

**`tests/integration/test_disk_types/test.py`** picked between `configs/storage_arm.xml` and `configs/storage_amd.xml`. The two files were byte-for-byte identical (both defined the same `disk_s3` + `disk_encrypted` pair), so this PR renames `storage_amd.xml` to `storage.xml`, deletes `storage_arm.xml`, and drops the conditional.

**`tests/integration/test_storage_s3_queue/test_5.py`** asserted `200 <= sizeof(FileStatus) <= 300` on arm and `250 <= ... <= 300` on x64. The arm bound is a strict superset, so use a single assertion without the `is_arm()` branch.

**`tests/integration/test_backward_compatibility/test_short_strings_aggregation.py`** used docker tag `24.1` on arm (older tags lacked arm images, see #59132) and `CLICKHOUSE_CI_MIN_TESTED_VERSION` otherwise. `24.1` is a multi-arch image that works on x86_64 too, so use it unconditionally and drop the `is_arm` / `CLICKHOUSE_CI_MIN_TESTED_VERSION` imports.

Noticed while auditing uses of `is_arm` after #103516.

...

### Version info
- Merged into: `26.5.1.73`


alexey-milovidov (commented):
Ok.

PR #100943, opened 2026-03-27
Skip views in TRUNCATE ALL TABLES

alexey-milovidov:
`TRUNCATE ALL TABLES FROM <database>` failed with "Truncate is not supported by storage View" when the database contained views.

Add `supportsTruncate` virtual method to `IStorage` (default true), override it to false in `StorageView`, and delegate in `StorageProxy`. Both TRUNCATE ALL TABLES code paths (with and without LIKE pattern) now skip storages that don't support truncation.

Closes #78165

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
`TRUNCATE ALL TABLES` no longer fails when the database contains views.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.212`

PR #103537, opened 2026-04-25
Fix `LOGICAL_ERROR` "Part X intersects previous part Y" during ReplicatedMergeTree startup

groeneai:
Fix a rare `LOGICAL_ERROR` "Part X intersects previous part Y" that aborts
`ReplicatedMergeTree` startup when two empty unexpected parts on disk have
intersecting block ranges without one fully containing the other.

## Failure

```
Logical error: 'Part all_2_11_3 intersects previous part all_0_5_2.
It is a bug or a result of manual intervention in the ZooKeeper data.'
(STID: 2352-49be)
```

Stack trace:
```
DB::ActiveDataPartSet::add (ActiveDataPartSet.cpp:83)
  <- StorageReplicatedMergeTree::checkPartsImpl (StorageReplicatedMergeTree.cpp:1909)
  <- StorageReplicatedMergeTree::checkParts (StorageReplicatedMergeTree.cpp:1847)
  <- ReplicatedMergeTreeAttachThread::runImpl (ReplicatedMergeTreeAttachThread.cpp:195)
```

Hit by the BuzzHouse fuzzer twice in 76 days (rare chronic trunk bug):
- 2026-04-25 on PR #100943 (`arm_asan_ubsan`): https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100943&sha=3308ce056bdfd5672c3c300ccae03968beac9846&name_0=PR&name_1=BuzzHouse%20%28arm_asan_ubsan%29
- 2026-02-08 on PR #96130 (`amd_binary, ParallelReplicas, s3 storage, sequential`)

Maintainer directive to file a separate fix PR for this finding:
https://github.com/ClickHouse/ClickHouse/pull/100943#issuecomment-4320085094

## Root cause

`StorageReplicatedMergeTree::checkPartsImpl` builds a covering set of empty
unexpected parts:

```cpp
ActiveDataPartSet set_of_empty_unexpected_parts(format_version);
for (const auto & load_state : unexpected_data_parts)
{
    if (load_state.is_broken || load_state.part->rows_count || !load_state.uncovered)
        continue;

    set_of_empty_unexpected_parts.add(load_state.part->name);  // <-- throws
}
```

The `uncovered` flag is computed in `MergeTreeData::loadDataParts`
(`MergeTreeData.cpp:2479`) as: "no *other* unexpected part's
`MergeTreePartInfo` *contains* this one". Containment is strict: a part
can be unexpected, uncovered, *and* still intersect another uncovered
unexpected part:

- `all_0_5_2`: blocks 0..5, level 2
- `all_2_11_3`: blocks 2..11, level 3

Neither contains the other (`min` and `max` block numbers don't satisfy
the containment relation), but they intersect at blocks 2..5. Both pass
the `uncovered == true` filter; `ActiveDataPartSet::add` rejects the
second one with `LOGICAL_ERROR` and aborts the table-startup attach
thread.

**How the two parts come to exist on disk** (traced by @ tuanpach in
https://github.com/ClickHouse/ClickHouse/pull/103537#discussion_r3954391392):
their names come from two different block-number sequences. `CHECK TABLE`
leaves part *names* queued on `ReplicatedMergeTreePartCheckThread`, and
`SYSTEM RESTORE REPLICA` then recreates the table's ZooKeeper nodes, which
resets the sequential block-number counters. `restoreMetadataInZooKeeper`
neither pauses that thread nor calls `cancelRemovedPartsCheck`, as the
`DROP_RANGE` paths do, so the thread later dequeues a stale pre-restore name
and `createEmptyPartInsteadOfLost` recreates it as an empty part under that
old name, beside parts renumbered from the reset counter. @ tuanpach's
#118704 pauses and clears the check queue in `SYSTEM RESTORE REPLICA`.

## Fix

Use `ActiveDataPartSet::tryAdd` and log a warning instead of throwing.
The set is only consulted as a heuristic: `getContainingPart` against
unexpected parts that look like leftovers of dropped ranges. If two
empty markers intersect, either one is an equally valid covering
candidate; any unexpected part that the skipped marker would have
covered falls through to the existing handling path further down in
the same loop, where every entry of `unexpected_data_parts` is
ultimately renamed to `detached/ignored` after the sanity check.

A unit test (`gtest_active_data_part_set`) is added to document the
contract: `add` throws on intersection without containment while
`tryAdd` returns `HasIntersectingPart` without throwing.

## Relation to PR #100992

PR #100992 by @ tuanpach fixes the analogous "Part intersects" failure in
`MergeTreeData::PartLoadingTree::add`, a different code path (`loadDataParts`,
not the replication-startup `checkPartsImpl`). Its transaction-aware
`RollbackStatus` choice suits `PartLoadingTree`, which feeds the live
`data_parts` set. The empty unexpected parts here are already
non-transactional drop markers slated to be detached as `ignored`, so
choosing arbitrarily between two intersecting candidates is sound.

Fixed a rare `LOGICAL_ERROR` "Part X intersects previous part Y" raised during `ReplicatedMergeTree` table startup when two empty unexpected parts on disk had overlapping but non-containing block ranges. The exception aborted the table-attach thread and prevented the table from coming up.

### Version info
- Merged into: `26.5.1.204`

### DIFF
--- a/src/Storages/StorageReplicatedMergeTree.cpp
+++ b/src/Storages/StorageReplicatedMergeTree.cpp
@@ -51,6 +51,7 @@
 #include <Storages/MergeTree/MergeTreeMutationStatus.h>
 #include <Storages/MergeTree/MergeTreeSettings.h>
 #include <Storages/MergeTree/MutateFromLogEntryTask.h>
+#include <Storages/MergeTree/OverlappingPartCovering.h>
 #include <Storages/MergeTree/PinnedPartUUIDs.h>
 #include <Storages/MergeTree/Compaction/CompactionStatistics.h>
 #include <Storages/MergeTree/Compaction/ConstructFuturePart.h>
@@ -1900,35 +1901,42 @@ bool StorageReplicatedMergeTree::checkPartsImpl(bool skip_sanity_checks)
 
     waitForUnexpectedPartsToBeLoaded();
 
-    ActiveDataPartSet set_of_empty_unexpected_parts(format_version);
+    /// Two empty unexpected parts on disk can have block ranges that intersect without one fully
+    /// containing the other (for example, `all_0_5_2` and `all_2_11_3`). The `uncovered` flag is
+    /// computed in `MergeTreeData::loadDataParts` as "no other unexpected part contains me", but
+    /// containment is strict — two parts can both be uncovered and still share blocks. Adding the
+    /// second one to a plain `ActiveDataPartSet` aborts startup with `LOGICAL_ERROR` (STID
+    /// `2352-49be`).
+    ///
+    /// Use `OverlappingPartCovering`: it stores the non-overlapping markers in a fast lookup
+    /// structure and keeps the rest in an auxiliary list, but `getContainingPart` consults both,
+    /// so an unexpected part that is covered only by an overlapping marker is still classified as
+    /// covered downstream. This avoids the related `TOO_MANY_UNEXPECTED_DATA_PARTS` regression
+    /// that a naive "skip the second marker" fix would introduce.
+    OverlappingPartCovering set_of_empty_unexpected_parts(format_version);
     for (const auto & load_state : unexpected_data_parts)
     {
         if (load_state.is_broken || load_state.part->rows_count || !load_state.uncovered)
             continue;
 
-        /// Two empty unexpected parts on disk can have intersecting block ranges without one
-        /// fully containing the other (for example, `all_0_5_2` and `all_2_11_3`). The
-        /// `uncovered` flag is computed as "no other unexpected part contains me", so both
-        /// parts can pass that check yet still intersect each other. Use `tryAdd` and skip
-        /// the conflicting part rather than aborting startup with `LOGICAL_ERROR`. The set
-        /// is only used as a heuristic to detect unexpected parts that look like leftovers
-        /// of dropped parts; keeping just one of two intersecting empty markers as a
-        /// covering candidate is sufficient — any unexpected part that the skipped marker
-        /// would have covered will fall through to the regular handling path below.
-        String reason;
-        if (set_of_empty_unexpected_parts.tryAdd(load_state.part->name, &reason)
-            == ActiveDataPartSet::AddPartOutcome::HasIntersectingPart)
-        {
+        String overlapping_reason;
+        set_of_empty_unexpected_parts.add(load_state.part->info, load_state.part->name, &overlapping_reason);
+        if (!overlapping_reason.empty())
             LOG_WARNING(
                 log,
-                "Skipping empty unexpected part {} from covering set: {}",
+                "Empty unexpected part {} overlaps another empty unexpected part without containment: {}; "
+                "keeping it as a covering candidate via the auxiliary index",
                 load_state.part->name,
-                reason);
-        }
+                overlapping_reason);
     }
     if (auto empty_count = set_of_empty_unexpected_parts.size())
+    {
         LOG_WARNING(log, "Found {} empty unexpected parts (probably some dropped parts were not cleaned up before restart): [{}]",
                  empty_count, fmt::join(set_of_empty_unexpected_parts.getParts(), ", "));
+        if (auto overlapping_parts = set_of_empty_unexpected_parts.getOverlappingParts(); !overlapping_parts.empty())
+            LOG_WARNING(log, "Of those, {} have block ranges that intersect another empty unexpected part without containment: [{}]",
+                     overlapping_parts.size(), fmt::join(overlapping_parts, ", "));
+    }
 
     /** To check the adequacy, for the parts that are in the FS, but not in ZK, we will only consider not the most recent parts.
       * Because unexpected new parts usually arise only because they did not have time to enroll in ZK with a rough restart of the server.
@@ -1958,7 +1966,7 @@ bool StorageReplicatedMergeTree::checkPartsImpl(bool skip_sanity_checks)
             continue;
         }
 
-        String covering_empty_part = set_of_empty_unexpected_parts.getContainingPart(load_state.part->name);
+        String covering_empty_part = set_of_empty_unexpected_parts.getContainingPart(load_state.part->info);
         if (!covering_empty_part.empty())
         {
             LOG_INFO(log, "Unexpected part {} is covered by empty part {}, assuming it has been dropped just before restart",
--- a/src/Storages/StorageReplicatedMergeTree.cpp
+++ b/src/Storages/StorageReplicatedMergeTree.cpp
@@ -1906,7 +1906,25 @@ bool StorageReplicatedMergeTree::checkPartsImpl(bool skip_sanity_checks)
         if (load_state.is_broken || load_state.part->rows_count || !load_state.uncovered)
             continue;
 
-        set_of_empty_unexpected_parts.add(load_state.part->name);
+        /// Two empty unexpected parts on disk can have intersecting block ranges without one
+        /// fully containing the other (for example, `all_0_5_2` and `all_2_11_3`). The
+        /// `uncovered` flag is computed as "no other unexpected part contains me", so both
+        /// parts can pass that check yet still intersect each other. Use `tryAdd` and skip
+        /// the conflicting part rather than aborting startup with `LOGICAL_ERROR`. The set
+        /// is only used as a heuristic to detect unexpected parts that look like leftovers
+        /// of dropped parts; keeping just one of two intersecting empty markers as a
+        /// covering candidate is sufficient — any unexpected part that the skipped marker
+        /// would have covered will fall through to the regular handling path below.
+        String reason;
+        if (set_of_empty_unexpected_parts.tryAdd(load_state.part->name, &reason)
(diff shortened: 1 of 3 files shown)


groeneai:
cc @alexey-milovidov @tavplubix @alesapin — could you review this? Filed per @alexey-milovidov's directive on PR #100943 to investigate and fix the BuzzHouse `STID: 2352-49be` finding in a separate PR. Targets the `checkPartsImpl` empty-unexpected-parts loop (distinct from the `PartLoadingTree::add` path covered by PR #100992). Repro is captured at the unit-test level — running `gtest_active_data_part_set` against the unmodified `add` deterministically aborts with the exact CI error and stack trace; `tryAdd` returns `HasIntersectingPart` cleanly.

groeneai:
**Pre-PR validation gate** (cron task-worker session `cron:clickhouse-ci-task-worker:20260425-171500`)

| # | Question | Status |
|---|---|---|
| a | **Deterministic repro?** | ✅ The original test version (`add` instead of `tryAdd`) deterministically aborts the process with the exact CI error message — `Logical error: 'Part all_2_11_3 intersects previous part all_0_5_2. ...'` — and the same stack trace (`ActiveDataPartSet::add` at line 83). Reproduces in-process in < 1 second. |
| b | **Root cause explained?** | ✅ The `uncovered` flag in `MergeTreeData::loadDataParts` (`MergeTreeData.cpp:2479`) is computed as "no other unexpected part *contains* this one". Containment and intersection are independent relations: `all_0_5_2` and `all_2_11_3` neither contains the other, but they intersect at blocks 2..5. Both pass the `uncovered == true` filter, then `ActiveDataPartSet::add` throws `LOGICAL_ERROR` on the second insert and aborts the table-startup attach thread. |
| c | **Fix matches root cause?** | ✅ `tryAdd` returns `HasIntersectingPart` cleanly instead of throwing. The set is only consulted as a heuristic by `getContainingPart`; either of two intersecting empty markers is an equally valid covering candidate, and any unexpected part that the skipped marker would have covered falls through to the regular handling path that detaches it as `ignored` further down in the same loop. |
| d | **Test intent preserved?** | ✅ The covering-set heuristic continues to function — only the *exception path* is removed. Every entry of `unexpected_data_parts` is ultimately renamed to `detached/ignored` after the sanity check, so skipping one of two intersecting empty markers does not change the final outcome. Added 4 new unit tests documenting the contract. |
| e | **Demonstrated in both directions?** | ✅ Without fix (using `add`): test deterministically aborts with the exact CI error/stack. With fix (using `tryAdd`): all 4 unit tests pass cleanly (`PASSED 4 tests` in 0 ms). |
| f | **Fix is general, not a narrow patch?** | ✅ Three call sites that follow the same `ActiveDataPartSet::add(part->name)` pattern have been mapped: `PartLoadingTree::add` (loadDataParts; covered by PR #100992 with transaction-aware logic), `tryLoadPartsToAttach` (`ALTER ATTACH PARTITION`; uses `add` with a covered-parts list, different semantics), `checkPartsImpl` (replication startup, this PR). Each path has different semantics and requires a different fix; this PR is the targeted fix for the empty-unexpected-parts heuristic, which is the only call site where intersection without containment is a legitimate state that should not abort startup. |

**Local builds:** `clickhouse` and `unit_tests_dbms` link cleanly with the change. `gtest_active_data_part_set` reports 4/4 tests passing.

groeneai:
@clickhouse-gh — thanks for the catch. Pushed [`6b1f736e`](https://github.com/ClickHouse/ClickHouse/pull/103537/commits/6b1f736ec4d) addressing the downstream concern.

**The original `tryAdd` + skip approach was incomplete.** As you noted, dropping the second of two intersecting markers can still trip `TOO_MANY_UNEXPECTED_DATA_PARTS` for any unexpected part that was covered ONLY by the skipped marker.

**Reworked fix:** new header-only helper `OverlappingPartCovering` (in `src/Storages/MergeTree/OverlappingPartCovering.h`). It keeps a primary `ActiveDataPartSet` for the fast non-overlapping case and stores intersecting markers in a small auxiliary list. `getContainingPart` consults both, so every marker contributes to coverage regardless of insertion order.

Walking through your concrete example with the new helper:
- markers `all_0_5_2` and `all_2_11_3` — one lands in the primary set, the other in the auxiliary list (which one depends on iteration order, doesn't matter)
- unexpected part `all_9_10_1` — `getContainingPart` returns `all_2_11_3` regardless, because `all_2_11_3` is consulted whether it's in the primary set or the auxiliary list
- result: `all_9_10_1` is correctly classified as covered, no `uncovered_unexpected_parts_rows` increment, no downstream `TOO_MANY_UNEXPECTED_DATA_PARTS`

**Unit tests** in `gtest_active_data_part_set.cpp` extended with your exact scenario in both insertion orders (`IntersectingMarkersStillCoverDownstreamParts` + `IntersectingMarkersOrderIndependent`), plus mixed-overlap, three-marker, contained, and disjoint cases. All 12 tests pass.

**Pre-PR validation gate (rework):**

| | |
|---|---|
| (a) Deterministic repro? | Yes — `OverlappingPartCovering.IntersectingMarkersStillCoverDownstreamParts` deterministically reproduces the bot scenario. |
| (b) Root cause explained? | Yes — the previous `tryAdd` fix dropped one marker from the covering set, so any unexpected part covered ONLY by that marker was silently reclassified as uncovered, inflating `uncovered_unexpected_parts_rows` and tripping `TOO_MANY_UNEXPECTED_DATA_PARTS`. |
| (c) Fix matches root cause? | Yes — the helper keeps every marker as a covering candidate via the primary set + auxiliary list, so no marker can be silently dropped from coverage detection. Not a band-aid; the new structure is correct by construction for the overlap-without-containment case. |
| (d) Test intent preserved? | Yes — original `ActiveDataPartSet` contract tests retained. New tests cover the bot's scenario in both orders, plus three-marker mixed-overlap, contained, disjoint, and empty cases. |
| (e) Both directions demonstrated? | Yes — without the auxiliary list, `IntersectingMarkersOrderIndependent` fails order-dependently (returns \"\" for `all_9_10_1` when `all_2_11_3` is added second). With the auxiliary list, all 12 tests pass deterministically. |
| (f) Fix is general, not a narrow patch? | Yes — checked all `ActiveDataPartSet` callers in `src/Storages`. Only `set_of_empty_unexpected_parts` operates on parts with arbitrary overlap (drop markers from concurrent traffic); other callers (`local_expected_parts_set`, `active_parts`, `independent_ranges_set`, `current_parts`/`virtual_parts` in queue) work with parts that maintain non-overlap by ZK validation or merge-selection invariants. |

Build + test:
- `ninja -C build clickhouse` → clean
- `ninja -C build unit_tests_dbms` → clean
- `unit_tests_dbms --gtest_filter=\"ActiveDataPartSet.*:OverlappingPartCovering.*\"` → 12/12 PASSED

Session: cron:clickhouse-ci-task-worker:20260426-224500

tuanpach (approved):

clickgapai:

**Affected code:** `StorageReplicatedMergeTree::checkPartsImpl` — core code present in all supported branches.

**Why:** The `set_of_empty_unexpected_parts` logic in `StorageReplicatedMergeTree::checkPartsImpl` was introduced in commit 13599ff462d on 2023-11-06, predating all supported release branches (25.8, 26.1, 26.2, 26.3). The bug aborts replicated table startup when two empty unexpected parts have intersecting block ranges — a real production hazard, even if rare (BuzzHouse hit it twice in 76 days). All supported branches contain the affected code path.

src/Storages/StorageReplicatedMergeTree.cpp
```diff
@@ -1900,17 +1901,42 @@ bool StorageReplicatedMergeTree::checkPartsImpl(bool skip_sanity_checks)
 
     waitForUnexpectedPartsToBeLoaded();
 
-    ActiveDataPartSet set_of_empty_unexpected_parts(format_version);
+    /// Two empty unexpected parts on disk can have block ranges that intersect without one fully
+    /// containing the other (for example, `all_0_5_2` and `all_2_11_3`). The `uncovered` flag is
+    /// computed in `MergeTreeData::loadDataParts` as "no other unexpected part contains me", but
+    /// containment is strict — two parts can both be uncovered and still share blocks. Adding the
+    /// second one to a plain `ActiveDataPartSet` aborts startup with `LOGICAL_ERROR` (STID
+    /// `2352-49be`).
```
tavplubix, line 1909: @tuanpach how were these parts created in the first place? The "Root cause" in the PR description doesn't explain this, it basically says that the assertion has failed but doesn't explain why/how

tuanpach: After a `MergeTree` table is detached as replicated, these parts were created during the `CHECK TABLE` and `SYSTEM RESTORE REPLICA` queries:
- `CHECK TABLE`: running in the background, queue the local part names.
- `SYSTEM RESTORE REPLICA` (run after `CHECK TABLE`): reset the block number generator (new block number node on Keeper).
These two use a different blocker number generator, which causes the conflict.

I created a new [PR-118704](https://github.com/ClickHouse/ClickHouse/pull/118704) which pauses and clears the check queue in `SYSTEM RESTORE REPLICA`.

Issue #101309, opened 2026-03-31
Logical error: Last stored last_written_position in meta file A is bigger than current last_written_pos (B > C) (STID: 3574-3d8d)


azat:
CI: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=101207&sha=b1ecb2a70b1ff55d1b0f54daabf2f05bd9ebd00a&name_0=PR&name_1=Stress%20test%20%28arm_asan_ubsan%2C%20s3%29

Probably some TOCTOU issue, will take a look

```sql
Log files: clickhouse-server.err.log, stderr.log
Error:
Logical error: 'Last stored last_written_position in meta file a.txt is bigger than current last_written_pos (122 > 0)'.
---

Stack trace:
__pthread_kill_implementation @ 0x0000000000082009
gsignal @ 0x000000000003a83c
__GI_abort @ 0x0000000000027134
src/Common/Exception.cpp:60:5: DB::abortOnFailedAssertion(String const&, std::basic_string_view<char, std::char_traits<char>>, void* const*, unsigned long, unsigned long) @ 0x0000000029bf1bc0
src/Common/Exception.cpp:93:13: DB::handle_error_code(String const&, std::basic_string_view<char, std::char_traits<char>>, int, bool, std::vector<void*, std::allocator<void*>> const&) @ 0x0000000029bf3e90
src/Common/Exception.cpp:146:19: DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0x0000000029bf43fc
./src/Common/Exception.h:171:100: DB::Exception::Exception(String&&, int, String, bool) @ 0x000000001a760d28
./src/Common/Exception.h:57:54: DB::Exception::Exception(PreformattedMessage&&, int) @ 0x000000001a75fa08
./src/Common/Exception.h:189:77: DB::Exception::Exception<String const&, unsigned long&, unsigned long&>(int, FormatStringHelperImpl<std::type_identity<String const&>::type, std::type_identity<unsigned long&>::type, std::type_identity<unsigned long&>::type>, String const&, unsigned long&, unsigned long&) @ 0x000000002a61b334
src/Storages/FileLog/StorageFileLog.cpp:573:15: DB::StorageFileLog::checkOffsetIsValid(String const&, unsigned long) const @ 0x000000003ff8bba4
[...]

PR #103545, opened 2026-04-26
Validate JSON/Dynamic data type AST shape before indexing

groeneai:
The AST fuzzer can produce a structurally invalid AST: in particular `fuzzColumnLikeExpressionList` can drop children from any `ASTExpressionList`, including the `arguments` of an `equals` function inside an `ASTObjectTypeArgument` parameter or a `Dynamic` argument. `createObject` (in `DataTypeObject.cpp`) and the `Dynamic` data type creator (in `DataTypeDynamic.cpp`) then accessed `function->arguments->children[0]` and `[1]` without checking that `arguments` was non-null or that the children vector had exactly two elements, segfaulting inside `boost::container::vector::operator[]` or `boost::intrusive_ptr::operator bool` on the fuzzed AST.

**STID:** `1477-2777` — observed once on master `BuzzHouse (experimental, serverfuzz, amd_msan)` at commit `a27a1f1ae8ac33d002adf2f74dc2c12ed48f3dd2` (CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=a27a1f1ae8ac33d002adf2f74dc2c12ed48f3dd2&name_0=MasterCI&name_1=BuzzHouse%20%28experimental%2C%20serverfuzz%2C%20amd_msan%29). Reproduced locally on master with:

```sql
EXPLAIN AST CREATE TABLE t (c0 JSON(max_dynamic_paths = 758)) ENGINE = Memory
SETTINGS ast_fuzzer_runs = 200, ast_fuzzer_any_query = 1;
```

This deterministically crashes within a handful of client invocations on master; with the fix it passes 100/100 invocations.

The fix:
* `createObject`: null-check the result of `argument->as<ASTObjectTypeArgument>()` (fuzzed children are not guaranteed to be that node type), validate that `function->arguments` is non-null and has exactly 2 children before indexing, null-check the `path_with_type` cast, and fix two pre-existing error-message bugs that called `formatForErrorMessage` on the field that was itself null in the failure path (`function` after a failed cast, `skip_path` while reporting a bad `skip_path_regexp`).
* `Dynamic` creator: validate `argument->arguments` is non-null and has exactly 2 children before indexing — the same fuzzer pattern applies to `Dynamic(max_types = N)`.

Includes a stateless regression test (`04120_json_dynamic_invalid_ast_no_segfault.sh`) that exercises the AST fuzzer over `CREATE TABLE` statements with `JSON` and `Dynamic` columns. Verified test fails on unfixed master (server crash → exit 143) and passes 5/5 on the fix.

### Changelog entry:
Fix a server crash in JSON and Dynamic data type parameter parsing when the abstract syntax tree (AST) was structurally malformed (for example produced by the AST fuzzer): the `equals(name, value)` parameter expression could be left with fewer than two children, which `DataTypeObject` and `DataTypeDynamic` then dereferenced without bounds checking.

### Version info
- Merged into: `26.5.1.937`


groeneai:
cc @Avogar — could you review? `DataTypeObject` and `DataTypeDynamic` parameter parsing crashes on AST-fuzzer-mutated inputs (STID 1477-2777, single hit on master `BuzzHouse (experimental, serverfuzz, amd_msan)`). The patch adds the missing null/size guards before indexing into `function->arguments->children` and the missing `as<>` validations. Reproduced locally on master and confirmed the regression test passes only with the fix.

Pre-PR validation:
- **Deterministic repro?** Yes — `SET ast_fuzzer_runs = 200, ast_fuzzer_any_query = 1; EXPLAIN AST CREATE TABLE t (c0 JSON(max_dynamic_paths = 758)) ENGINE=Memory` crashes within a handful of invocations on master at `a27a1f1ae8a`.
- **Root cause explained?** Yes — the AST fuzzer's `fuzzColumnLikeExpressionList` removes children from `ASTExpressionList`s at random (`QueryFuzzer.cpp:641`); when applied to the `equals(name, value)` parameter expression of a JSON/Dynamic argument, the resulting `function->arguments->children` ends up with 0 or 1 elements. `createObject` and the `Dynamic` creator then index `children[0]` and `children[1]` without bounds checks.
- **Fix matches root cause?** Yes — guards added at every `as<>` cast and every `children[N]` access on the affected paths; sibling `DataTypeDynamic` patched for the same anti-pattern.
- **Test intent preserved?** N/A — defensive guards only, no test weakened.
- **Both directions demonstrated?** Yes — reverted the patch, rebuilt, server crashes (test exit 143). Restored the patch, test passes 5/5.
- **General fix?** Yes — also fixed two pre-existing error-message bugs that would call `formatForErrorMessage` on the field that is itself null in the failure path, and audited `DataTypeEnum` (already safe via `checkASTStructure`).

src/DataTypes/DataTypeObject.cpp
```diff
@@ -628,12 +628,21 @@ static DataTypePtr createObject(const ASTPtr & arguments, const DataTypeObject::
     for (const auto & argument : arguments->children)
     {
         const auto * object_type_argument = argument->as<ASTObjectTypeArgument>();
+        /// The AST may be structurally invalid (e.g. produced by the AST fuzzer), so we cannot
```
Avogar: Let's not mention fuzzer in the comments. It's just a safety checks that may be useful in general

Avogar (changes requested):

src/DataTypes/DataTypeObject.cpp
groeneai: Done in a4c7f087bac. The three defensive-check comments in `DataTypeObject.cpp` (lines 631 and 642) and `DataTypeDynamic.cpp` (line 82) now describe the safety check as general — validating the AST shape before indexing — without naming the fuzzer.

The test header in `04120_json_dynamic_invalid_ast_no_segfault.sh` still mentions the AST fuzzer because the test itself runs `--ast_fuzzer_runs=200` to exercise these guards; happy to reword that too if you'd prefer it framed differently.

tests/queries/0_stateless/04120_json_dynamic_invalid_ast_no_segfault.sh
```diff
@@ -0,0 +1,53 @@
...
+) ENGINE = Memory;
+DROP TABLE t_smoke;
+" || exit 1
+
+# Server-side AST fuzzer: 200 mutations over a CREATE TABLE that contains JSON
+# and Dynamic types. The fuzzer can drop the identifier or the literal from the
+# `equals(name, value)` parameter expression, leaving an empty / single-element
+# argument list. Without the size guards in `createObject` and the `Dynamic`
+# creator, this segfaults inside `boost::container::vector::operator[]`.
+FUZZER_OPTS=("${OPTS[@]}" --ast_fuzzer_runs 200 --ast_fuzzer_any_query 1 --send_logs_level=fatal)
```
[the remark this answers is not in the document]
groeneai: Made the reproducer deterministic in 493c5a29afe.

Three changes:

1. **Prime the fuzzer's `column_like` pool** with an initial `SELECT` + `CREATE TABLE` + `INSERT` + `SELECT` + `DROP TABLE` block. The mutation branches in `fuzzColumnLikeExpressionList` (line 648-657) and `fuzzExpressionList` (arithmetic / `if` / `multiIf`) only actually replace children when `getRandomColumnLike()` returns non-null, so on a freshly started server they almost never fire. Priming makes them fire consistently.

2. **Bump `ast_fuzzer_runs` 200 → 1000** per query (single-run fuzz budget).

3. **Run the three fuzzer queries in an outer loop of 50 iterations** to amortise the still-random single-iteration crash rate.

Empirically verified by rebuilding master with the fix reverted: 10/10 trials crashed the server, with iter-to-crash distributed as `1, 1, 2, 2, 3, 5, 5, 6, 10, 19, 27` (max 27). On the fix branch the test passes 5/5 in ~26 seconds with no server crashes.

The test still ends with `SELECT 'ok'` and the `.reference` file is `ok\n`; on unfixed master the server is dead by the time the loop ends, the final `SELECT` produces no `ok`, and the reference does not match — so the test fails as required for bugfix validation.

Session: `cron:clickhouse-ci-task-worker:20260512-201500`.

alexey-milovidov:

Avogar:
@groeneai the new test `04120_json_dynamic_invalid_ast_no_segfault` is falided in the CI due to timeout. Make it much faster and shorter

groeneai:
@Avogar shortened the test in `35fff55b240d`.

The previous deterministic form ran three `EXPLAIN AST` invocations 50 times each with `ast_fuzzer_runs=1000` per invocation = 150 client calls, ~26 s on debug per run. Sanitizer flaky check runs the test 50 times per sanitizer, which pushed the job past its 45 min budget.

New form: three CREATE TABLE shapes sent together in one client invocation per outer iteration via `-m -q`, outer loop reduced from 50 to 25. Total fuzz budget is 75 000 mutations (3 shapes x 1000 fuzz runs x 25 outer iterations).

Verified locally:
- Unfixed master binary (defensive checks removed from `DataTypeObject.cpp` and `DataTypeDynamic.cpp`, rebuilt): 5/5 trials abort the server inside `DataTypeDynamic::create` (signal 6) before the test finishes the outer loop.
- Fixed binary: 5/5 trials pass in ~6.6 s on debug, about 4x faster than the previous form.

Session: cron:clickhouse-ci-task-worker:20260519-150000

Avogar:
@groeneai new test still fails in the CI. Maybe it was not a good idea to add a bash test with fuzzing. If the error cannot be reproduced without fuzzer using usual queries, I think it's ok to add additional validations without the test.

groeneai:
@Avogar — replaced the bash fuzzer test with deterministic gtests in 194d223ee9f7.

You're right: the bash test relied on `ast_fuzzer_runs`, which was both slow and inherently random. I looked for a way to produce the same malformed AST via well-formed SQL and could not find one — the shapes the fuzzer hits (e.g. an `equals` function with null `arguments`, an `ASTObjectTypeArgument` whose `parameter` is not an `ASTFunction`, a `path_with_type` whose value is not an `ASTObjectTypedPathArgument`) cannot be produced by the parser. So I switched to the same pattern as the existing `gtest_DataTypeDynamic_create.cpp` (the regression test for #101309): build the malformed shape directly as AST nodes and call `DataTypeFactory::get("JSON" / "Dynamic", arguments)`.

Changes in 194d223ee9f7:

- Delete `tests/queries/0_stateless/04120_json_dynamic_invalid_ast_no_segfault.sh` and its `.reference`.
- Extend `gtest_DataTypeDynamic_create.cpp` with 4 new `DataTypeDynamic::create` cases: `equals` with null `arguments`, with 0 children, with 1 child, with 3 children.
- Add `gtest_DataTypeObject_create.cpp` with 8 `DataTypeObject::createObject` cases: non-`ASTObjectTypeArgument` child, `equals` with null `arguments`, with 0 children, with 1 child, `parameter` with wrong function name, `parameter` that is not an `ASTFunction`, `path_with_type` of the wrong AST kind, `skip_path_regexp` of the wrong AST kind. Plus a valid-AST sanity case.

Verified in both directions on debug:

- Restored the defensive checks in `DataTypeObject.cpp` and `DataTypeDynamic.cpp`: all 15 tests pass in 1 ms.
- Reverted just the defensive checks: 11 of 15 tests reproduce the regression — 10 SIGABRT inside `boost::intrusive_ptr::operator->` / `operator[]` plus 1 wrong-behavior detection. The remaining 4 are the valid-AST sanity checks plus the pre-existing test from #101309 (already fixed in master).

Total runtime is under 1 ms across all 15 cases — no sanitizer timeout concern.

Run with:

```
build/src/unit_tests_dbms --gtest_filter='DataTypeDynamic.*:DataTypeObject.*'
```

groeneai:
Pre-PR validation gate (session cron:clickhouse-ci-task-worker:20260520-141500):

a) Deterministic repro: yes. Each of the 11 reproducing gtests crashes on the unfixed binary every time the test is run. No fuzzer, no randomness.

b) Root cause: `DataTypeObject::createObject` and `DataTypeDynamic::create` access `object_type_argument->parameter`, `path_with_type`, and `function->arguments->children[0]` / `[1]` without validating (i) the `as<ASTObjectTypeArgument>` / `as<ASTObjectTypedPathArgument>` / `as<ASTFunction>` cast returns non-null, (ii) `function->arguments` is non-null, (iii) `children.size() == 2`. A malformed AST dereferences a null `boost::intrusive_ptr` or reads past the end of the children vector.

c) Fix matches root cause: yes. Guards added at exactly those access points in commit 07455024af2, unchanged in this push.

d) Test intent preserved: yes. The previous bash test relied on fuzzer randomness to construct one of these malformed shapes; the new gtests construct each shape directly. Coverage is broader (8 JSON shapes + 4 Dynamic shapes vs. the previous 3 fuzzed `CREATE TABLE` queries).

e) Both directions: yes. Verified above — 11/15 reproduce on unfixed binary, all 15 pass on fixed binary.

f) Fix is general, not a narrow patch: the original commit already covers all four call-site classes in the two files (cast-result null check on `ASTObjectTypeArgument`, on `ASTObjectTypedPathArgument`, on `ASTFunction`, and the `equals` arguments shape check). No symmetric copies of this code in other DataType files.

Avogar (approved):

PR #101111, opened 2026-03-29
Fix infinite loop in `skipPrefixBeforeHeader` when file has fewer lines than `skip_first_lines`

alexey-milovidov:
When `input_format_csv_skip_first_lines` or `input_format_tsv_skip_first_lines` is set to a value larger than the number of lines in the file, `skipPrefixBeforeHeader` enters an infinite loop doing `pread` at EOF. Each call to `readRow` succeeds with an empty result even at EOF, so the loop never terminates.

Add an EOF check before each `readRow` call to break out of the loop when the file is exhausted.

Found via AST fuzzer in Stress test (amd_msan) on PR #100947.

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix infinite loop when `input_format_csv_skip_first_lines` or `input_format_tsv_skip_first_lines` exceeds the number of lines in the file.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #103547, opened 2026-04-26
Backfill missed periodic events in `system.query_metric_log`

groeneai:
When a scheduled `collectMetric` task fires after the query has already finished, `getQueryInfo` returns null and the periodic event is silently dropped. Under TSan builds and in highly parallel CI runs the `BackgroundSchedulePool` can fall behind by enough for this to happen repeatedly, leaving `system.query_metric_log` with fewer rows than the documented behavior promises — and occasionally missing the "query finished" row entirely when the last periodic happens to share the finish moment's timestamp (the existing `query_info_time <= info.last_collect_time` guard then drops the final).

`QueryMetricLog::finishQuery` now backfills the missed periodic rows using the `query_info` captured at finish (deltas remain monotonically correct via the existing `new_value < old_value` guard inside `createLogMetricElement`) and always emits the final row by passing a new `is_final=true` flag through to `createLogMetricElement`. The dedup guard remains in place for ordinary periodic events and only the final bypasses it.

The new test (`04121_query_metric_log_backfill_missed_periodics`) exercises a 1.5 s query with a 100 ms interval and asserts both the row count and the cumulative profile counters — it documents the new contract. The existing `03203_system_query_metric_log` test was the symptom: it has been flaky on `Stateless tests (amd_tsan, parallel, 1/2)` for 30 days (4 unrelated PRs + 1 master hit) with the canonical "missing final event" diff.

The category is `Improvement` rather than a stricter category: the failure mode requires `BackgroundSchedulePool` overload to manifest, and the bug-fix-validation lane (which runs the new test against an unloaded master HEAD) cannot deterministically reproduce the timing-sensitive race. The test still serves as a regression guard for the new contract — a future revert of the backfill or `is_final` bypass that breaks the documented row count would be caught by the `04121` assertions.

CI report that surfaced the latest occurrence: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=101111&sha=d1bb5ca2ccaa688cb9d3c9dc6899e604ab175a7d&name_0=PR&name_1=Stateless%20tests%20%28amd_tsan%2C%20parallel%2C%201%2F2%29

### Verification

- Existing `03203_system_query_metric_log`: 50/50 passes locally on a debug build (was the symptom of the issue).
- Existing `03203_system_query_metric_log` under heavy CPU stress (`stress-ng --cpu 96`): 30/30 passes (previously could fail).
- Existing `03549_system_query_metric_log_with_queries_with_same_query_id`: 10/10 passes (sibling test, no regressions).
- New `04121_query_metric_log_backfill_missed_periodics`: 30/30 passes.

- Improvement

Improve reliability of `system.query_metric_log` so that rows are not silently dropped when the `BackgroundSchedulePool` is overloaded. Periodic rows that were scheduled to fire but only ran after the query had finished used to be silently dropped, and the final "query finished" row could also be dropped if it shared a timestamp with the last periodic row. The query-finish path now backfills the missed periodic rows and always emits the final row.

### Version info
- Merged into: `26.5.1.150`


groeneai:
### Pre-PR validation gate (a–f)

a) **Deterministic repro?** Partial. The original `03203_system_query_metric_log` failure rate is ~0.1% on `Stateless tests (amd_tsan, parallel, 1/2)` and was not reproducible locally even under heavy `stress-ng --cpu 96` load (30/30 passes pre-fix; 50/50 passes post-fix; 30/30 passes under stress post-fix). The root cause is identifiable directly from the CI failure context (`+0` row for "Check that there is a final event when queries finish") and from the source: a periodic `collectMetric` task that fires after the query has been removed from the `ProcessList` returns at `QueryMetricLog.cpp:107-110` because `getQueryInfo` is null, and a final event whose timestamp matches the last periodic is dropped at `QueryMetricLog.cpp:266` by the `<=` dedup guard.

b) **Root cause explained?** Yes. Two distinct loss paths:

  - **Path 1 (missed periodic):** the `BackgroundSchedulePool` fires the scheduled task after `finishQuery` has already erased the entry from `queries`. The task captures `current_time` and calls `process_list.getQueryInfo(query_id)`; the query is gone from the `ProcessList`, so `getQueryInfo` returns null and the task returns at line 107. The periodic row is lost.
  - **Path 2 (final time-collision):** a late periodic captures `current_time = finish_time` and updates `info.last_collect_time = finish_time`. `finishQuery` then calls `createLogMetricElement(finish_time, …)` and the existing `query_info_time <= info.last_collect_time` guard drops the final row.

c) **Fix matches root cause?** Yes. Path 1 is fixed by backfilling at finish: `while (info.next_collect_time < finish_time) emit at info.next_collect_time; info.next_collect_time += interval`. Path 2 is fixed by adding `is_final` to `createLogMetricElement` and bypassing the dedup guard for the final event only. `info.last_collect_time` becomes `std::max(...)` so it never moves backwards.

d) **Test intent preserved?** Yes. `03203_system_query_metric_log`'s assertions (`count() > 2` for the 1000 ms-interval branch and the per-interval `BETWEEN` check) continue to test the same property: a query that runs longer than its interval produces multiple metric rows including a final one. The fix makes the property deterministically true rather than timing-dependent. The new test `04121_query_metric_log_backfill_missed_periodics` adds an independent assertion on row count and cumulative `ProfileEvent_SleepFunctionMicroseconds`, so the regression is covered even if 03203 is later refactored.

e) **Both directions demonstrated?** Yes by construction:
  - `03203` was failing in CI under TSan (the bug). The fix removes the failure mode.
  - Local repros pre-fix and post-fix: 30/30 OK debug + default settings (both); 30/30 OK debug + `stress-ng --cpu 96` (both); 50/50 OK with fix.
  - The new test was authored to exercise both fix paths and passes 30/30.

f) **Fix is general?** Yes:
  - Backfill loop covers any number of missed periodic events, regardless of which periodic was lost.
  - `is_final` covers the time-collision case for any combination of intervals.
  - Audited callers: `createLogMetricElement` is called from `collectMetric` (with `schedule_next=true`, `is_final=false` by default — unchanged behavior) and from `finishQuery` (with `schedule_next=false`; the backfill iterations also pass `is_final=false` so they still dedup, only the explicit final emission passes `is_final=true`).

cc @alexey-milovidov @pamarcos

groeneai:
cc @pamarcos @Algunenano @alexey-milovidov — could you review this? `pamarcos` is the original author of `system.query_metric_log` and the chronic flakiness of `03203_system_query_metric_log` traces back to lost periodic events under heavy `BackgroundSchedulePool` load. The fix backfills missed periodic rows in `finishQuery` and adds an `is_final` bypass so a time-collision between the last periodic and the finish moment can't drop the final row.

groeneai:
Bumping for visibility — @pamarcos, could you take a look when you have a moment? @alexey-milovidov pointed me to you on #101111 as the original author of `system.query_metric_log`. This PR backfills periodic rows that were silently dropped when `BackgroundSchedulePool` falls behind, and also makes the final "query finished" row bypass the timestamp-collision dedup guard. New regression test `04121_query_metric_log_backfill_missed_periodics.sh` covers both modes.

Local verification: `03203_system_query_metric_log` 50/50 OK, 30/30 under `stress-ng --cpu 96` CPU stress, sibling `03549` 10/10 OK.

groeneai:
Reclassified this PR from `Bug Fix` to `Improvement` and pushed an empty commit to retrigger CI (session: cron:clickhouse-ci-task-worker:20260428-021500).

**Why:** The previous run's `Bugfix validation (functional tests)` lane reported FAIL with `Failed to reproduce the bug` — the new `04121_query_metric_log_backfill_missed_periodics` test PASSES on an unloaded master HEAD. The failure mode requires `BackgroundSchedulePool` overload to manifest (a timing-sensitive race that doesn't reliably reproduce on a quiet runner), so the bug-fix-validation lane cannot be satisfied for this PR.

**What stays the same:**
- The fix in `QueryMetricLog::finishQuery` (backfill missed periodics + `is_final` bypass).
- The `04121` test as a regression guard for the documented row-count contract.
- The 03203 stabilization evidence (50/50 + 30/30 under `stress-ng --cpu 96`).
- Reviewer pings to @pamarcos / @Algunenano / @alexey-milovidov.

The `Config Workflow` should drop the `pr-bugfix` label and add `pr-improvement` on this re-run; the `Bugfix validation` job is then filtered out by `ci/jobs/scripts/workflow_hooks/filter_job.py` (line 205, " Bug Fix" not in `pr_body`).

alexey-milovidov (approved):
The fix is sophisticated. But passable.

Issue #69070, opened 2024-08-29
clickhouse-local can not read files from /proc 


filimonov:
```
root@25c643a6893d:/# clickhouse-local 
ClickHouse local version 24.8.2.3 (official build).

:) select * from file('/proc/cpuinfo','RawBLOB');
;

SELECT *
FROM file('/proc/cpuinfo', 'RawBLOB')

Query id: 5e142777-6a82-45a7-a42e-7c77162b38f1

Ok.

0 rows in set. Elapsed: 0.002 sec. 

:) exit;
Bye.
root@25c643a6893d:/# cat /proc/cpuinfo
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 158
model name	: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
stepping	: 10
microcode	: 0xf4
cpu MHz		: 2600.021
cache size	: 9216 KB
physical id	: 0
siblings	: 12
core id		: 0
cpu cores	: 6
```

PR #103548, opened 2026-04-26
Fix clickhouse-local reading /proc and /sys files

ashrithb:

Fix `clickhouse-local` returning 0 rows silently when reading from `/proc` and `/sys` pseudo-files via the `file()` table function (e.g. `SELECT * FROM file('/proc/cpuinfo', 'RawBLOB')`).

---

## Motivation

`clickhouse-local` defaults `storage_file_read_method` to `mmap`, while
`clickhouse-server` defaults to `pread`. For Linux pseudo-files in `/proc`
and `/sys`, `fstat` reports `st_size == 0` even when the file has content
readable via `read`/`pread`. The mmap branch in `selectReadBuffer` then
constructed an `MMapReadBufferFromFile` that mapped zero bytes
(`MMappedFileDescriptor::set` early-returns when `length == 0`), so
`RawBLOBRowInputFormat` saw an empty buffer and returned no rows with no
error.

## Fix

Skip the `mmap` branch when `file_stat.st_size == 0` and fall through to
the existing `pread`-based reader. Genuinely empty regular files behave
exactly as before (zero rows, no error), and the pseudo-file content
becomes readable.

## Test

Adds `tests/queries/0_stateless/04124_clickhouse_local_proc.sh`
(Linux-only, `# Tags: linux`) which pins `--storage_file_read_method=mmap`
explicitly and asserts:

```sql
SELECT notEmpty(raw_blob) FROM file('/proc/cpuinfo', 'RawBLOB')
```

returns `1`. Uses `notEmpty(raw_blob)` rather than `count()` to bypass the
`RawBLOBRowInputFormat::countRows` fast path so the actual read buffer is
exercised.

Issue #69070

### Version info
- Merged into: `26.5.1.770`


alexey-milovidov (approved):
Amazing, thanks!

alexey-milovidov:
@groeneai, fix the `Logical error: Port is already connected` error and send a separate PR.

clickgapai:

**Affected code:** `selectReadBuffer` in src/Storages/StorageFile.cpp:421 — core code present in all supported branches.

**Why this might apply:** The affected code path (mmap branch in `selectReadBuffer`) has existed since ancient times - `StorageFile.cpp` was added in commit 0e2eb5436b5. This is a silent data loss bug affecting users who read `/proc` or `/sys` files with clickhouse-local. The fix is minimal and safe (adds one condition check).

PR #81323, opened 2025-06-04
Engine=Kafka support for rack-aware consumption for Amazon MSK, IMDSv2

ilejn:
### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
- Parameter kafka_autodetect_client_rack is introduced. If set, discover Availability Zone by cloud facilities and promoted as client.rack librdkafka parameter to avoid cross-zone communications.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

### Details

Handles https://github.com/ClickHouse/ClickHouse/issues/80445

### Version info
- Merged into: `26.5.1.409`

PR #103549, opened 2026-04-26
S3 credentials: use IMDSv2 token for availability-zone detection

ashrithb:

Fix availability-zone detection on EC2 instances configured for "IMDSv2 only (token required)". `getRunningAvailabilityZone` now obtains an IMDSv2 session token before fetching the AZ resource, falling back to the legacy unauthenticated request when the metadata service does not implement the token endpoint.

---

## Motivation

`AWSEC2MetadataClient::getAvailabilityZoneOrException` performed a plain
unauthenticated `GET` against the EC2 instance metadata service. On
instances configured for "IMDSv2 only" (the AWS-recommended hardening
setting), the metadata endpoint rejects unauthenticated `GET` requests
with `401 Unauthorized`, so `tryGetRunningAvailabilityZone` failed and
ClickHouse / Keeper logged

tryGetRunningAvailabilityZone: ... Failed to get AWS availability zone.
HTTP response code: HTTP_UNAUTHORIZED.
... Failed to find the availability zone, tried AWS and GCP.

even though the metadata service was reachable.

The credentials path in the same client (`getDefaultCredentialsSecurely`)
already does the IMDSv2 dance: PUT `/latest/api/token` to obtain a token,
then GET with `x-aws-ec2-metadata-token` set. The AZ helper still used
the older code path.

## Fix

Acquire an IMDSv2 session token first via `PUT /latest/api/token` with
the `x-aws-ec2-metadata-token-ttl-seconds` header (the constants and
header names already exist in `Credentials.h`), then send the AZ `GET`
with `x-aws-ec2-metadata-token` set. If the token `PUT` fails or returns
a non-200, fall back to the unauthenticated `GET` so endpoints that do
not implement IMDSv2 keep working as before.

## Test

Adds an integration test using a new mock metadata server,
`tests/integration/test_placement_info/metadata_servers/imdsv2_server.py`,
which requires the IMDSv2 token on the AZ `GET` and returns `401` without
it. A new `node_imds_v2` instance and `test_placement_info_from_imdsv2_only`
case in `tests/integration/test_placement_info/test.py` exercise the new
path end-to-end against the mock.

Closes #81402

### DIFF
--- a/src/IO/S3/Credentials.cpp
+++ b/src/IO/S3/Credentials.cpp
@@ -62,6 +62,7 @@ namespace S3
 #    include <boost/algorithm/string/split.hpp>
 #    include <boost/algorithm/string/classification.hpp>
 #    include <Poco/Exception.h>
+#    include <Poco/String.h>
 #    include <Poco/URI.h>
 #    include <Poco/Net/HTTPClientSession.h>
 #    include <Poco/Net/HTTPRequest.h>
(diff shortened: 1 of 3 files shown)


alexey-milovidov:
Thanks for this fix, @ashrithb!

Since this PR was opened (2026-04-26), the same issue was independently fixed in master by #81323 (`kafka_aws_az_aware`, merged 2026-05-07). `getAvailabilityZoneOrException` now acquires an IMDSv2 session token via `PUT /latest/api/token` and sends it with the availability-zone `GET`, falling back to the token-less IMDSv1 request — the same approach proposed here, with some extras (availability-zone-id support and serialized token retrieval). So the production change in this PR is now redundant.

Your integration test, however, covers the "IMDSv2 only (token required)" scenario that master's `test_placement_info` was still missing, so it is worth keeping. I've rebased just the test (`imdsv2_server.py` + `test_placement_info_from_imdsv2_only`) onto current master and opened #107767 to add it, with you retained as `Co-authored-by`. It passes against master's implementation.

Closing this one as superseded by #81323. Thank you again for the contribution and the test!

ashrithb:
Sure, thanks for letting me know!

PR #103551, opened 2026-04-26
Preserve hung_check.log on stress-test deadlock failure

groeneai:
The stress test runs `clickhouse-test --hung-check --capture-client-stacktrace` after every job. When this detects long-running queries the runner collects the processlist and `gdb` backtraces of every server process and writes them all into `hung_check.log`. That log is the primary diagnostic evidence for any deadlock.

`ci/jobs/scripts/stress/stress.py` was deleting the log on exactly the failure path (`hung_check_log.unlink()`), leaving only the bare row `Hung check failed, possible deadlock found\tFAIL\t\N\t` in `test_results.tsv`. The result is that every "Hung check failed" entry in CIDB has an empty `test_context_raw`, the S3 artifact upload picks up nothing useful, and investigators have no way to tell which mutex / condvar / processor was stuck.

This change:
- Keeps `hung_check.log` on disk so the artifact upload (`process_results` in `ci/jobs/stress_job.py`) includes it.
- Embeds the last 32 KiB of the captured output as the 4th column of `test_results.tsv`. The tail is the relevant region (`Found hung queries in processlist:`, the processlist with stacktraces, and the gdb backtraces all print at the end). The encoding mirrors the decoder in `read_test_results` (tab and newline escaped to `\t` / `\n`).
- The success branch (`No queries hung`) is unchanged.

This is the smallest infrastructure fix that unblocks investigation of the chronic "Hung check failed, possible deadlock found" stress-test umbrella (~290 hits / 30 days, multiple independent root causes) — the chronic task explicitly identified the deleted log as the blocker.

Requested by @alexey-milovidov on https://github.com/ClickHouse/ClickHouse/pull/100941

Preserve `hung_check.log` and embed its tail in `test_results.tsv` when the stress-test hung check fails, so that the captured processlist and thread stacktraces survive into CI artifacts and CIDB.

### Version info
- Merged into: `26.5.1.122`

### DIFF
--- a/ci/jobs/scripts/stress/stress.py
+++ b/ci/jobs/scripts/stress/stress.py
@@ -641,14 +641,49 @@ def main():
                     tee.stdin.close()
             if res != 0 and have_long_running_queries:
                 logging.info("Hung check failed with exit code %d", res)
+
+                # Embed a tail of the captured hung-check output in
+                # test_results.tsv so the processlist and thread stacktraces
+                # are visible in CIDB. The full log is also kept as a CI
+                # artifact (see process_results in stress_job.py), giving
+                # investigators access to the complete diagnostic output.
+                info_field = ""
+                try:
+                    log_text = hung_check_log.read_text(errors="replace")
+                    # 32 KiB is enough for the processlist and a handful of
+                    # stacktraces. Keep the tail because the diagnostic
+                    # content (Found hung queries, processlist with
+                    # stacktraces, gdb backtraces) appears at the end.
+                    max_chars = 32 * 1024
+                    if len(log_text) > max_chars:
+                        log_text = (
+                            "(truncated; see hung_check.log artifact for the"
+                            " full output; showing last 32 KiB)\n...\n"
+                            + log_text[-max_chars:]
+                        )
+                    # Escape so tab and newline survive the TSV encoding,
+                    # matching the decoder in read_test_results().
+                    info_field = log_text.replace("\t", "\\t").replace(
+                        "\n", "\\n"
+                    )
+                except OSError as ex:
+                    logging.warning(
+                        "Failed to read hung_check.log to embed in"
+                        " test_results.tsv: %s",
+                        ex,
+                    )
+
                 hung_check_status = (
-                    "Hung check failed, possible deadlock found\tFAIL\t\\N\t\n"
+                    "Hung check failed, possible deadlock found\tFAIL\t\\N\t"
+                    f"{info_field}\n"
                 )
                 with open(
                     args.output_folder / "test_results.tsv", "w+", encoding="utf-8"
                 ) as results:
                     results.write(hung_check_status)
-                    hung_check_log.unlink()
+                # Keep hung_check.log on disk so the CI artifact upload picks
+                # it up. Without it, deadlock investigations have no evidence
+                # to work with — see ClickHouse/ClickHouse#100941.
             else:
                 logging.info("No queries hung")
 
--- a/ci/jobs/scripts/stress/stress.py
+++ b/ci/jobs/scripts/stress/stress.py
@@ -647,19 +647,35 @@ def main():
                 # are visible in CIDB. The full log is also kept as a CI
                 # artifact (see process_results in stress_job.py), giving
                 # investigators access to the complete diagnostic output.
+                #
+                # Read only the last 32 KiB rather than the whole file: on
+                # deadlock failures `hung_check.log` can be very large (a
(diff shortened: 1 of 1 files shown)


groeneai:
**Pre-PR validation (self-review)**

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. End-to-end Python simulation: write a fake `hung_check.log` with realistic processlist + gdb stacktraces → run the new code path → read back via the same decoder used by `process_results` (`line[3].replace("\\t", "\t").replace("\\n", "\n")`) → confirm the original content is recovered byte-for-byte. Also covers: empty log, oversized log (50 KiB → truncated to 32 KiB tail with header), TSV row parses into 4 fields, `\t` and `\n` survive the round-trip. |
| b | **Root cause explained?** | Yes. `clickhouse-test --hung-check --capture-client-stacktrace` writes the processlist with per-query stacktraces and the gdb backtraces of every server process into `hung_check.log` (via `tee`). On the failure path `stress.py` calls `hung_check_log.unlink()` after writing only `Hung check failed, possible deadlock found\tFAIL\t\N\t` to `test_results.tsv`. So the bytes that explain the deadlock are deleted exactly when they would be useful, and CIDB receives an empty `test_context_raw`. |
| c | **Fix matches root cause?** | Yes. The fix removes the `unlink()` (so the artifact upload in `process_results` picks the file up) and additionally embeds the last 32 KiB of the captured output as the 4th column of `test_results.tsv` so the same evidence is also queryable from CIDB. No band-aid, no widened tolerance — the deletion is just removed. |
| d | **Test intent preserved?** | Yes. Behavior on the success branch is unchanged. The failure row in `test_results.tsv` keeps the same first three columns; only the previously-empty 4th column is now populated. The `FAIL` status, the test name, and the artifact upload all behave identically, just with the diagnostic content present instead of missing. |
| e | **Both directions demonstrated?** | Yes. End-to-end test confirms (1) the file persists on disk after the failure path runs, (2) `test_results.tsv` now contains the captured processlist/stacktraces in column 4, and (3) the existing decoder round-trips them losslessly. The pre-fix behavior (deleted file, empty 4th column) is what produced the chronic CIDB blind spot we're closing. |
| f | **Fix is general?** | Yes — there is exactly one site in the codebase that handles this log (verified via `grep -n "hung_check"` in `ci/jobs/scripts/stress/`), and the fix is applied there. It also covers every stress-test variant uniformly because `stress.py` is the shared entrypoint for all `Stress test (*)` jobs (regular + experimental serverfuzz across debug/asan/tsan/msan/release on amd and arm).

Session: `cron:clickhouse-ci-task-worker:20260426-121500`

groeneai:
cc @maxknv @leshikus — could you review this? Small change to `ci/jobs/scripts/stress/stress.py` that stops the stress-test runner from deleting `hung_check.log` on the failure path and embeds its tail in `test_results.tsv` so the captured processlist + thread stacktraces are visible from CIDB. Requested by @alexey-milovidov on https://github.com/ClickHouse/ClickHouse/pull/100941.

groeneai:
Addressed in https://github.com/ClickHouse/ClickHouse/pull/103551/commits/53ac21e566d.

Replaced `hung_check_log.read_text(errors="replace")` with a bounded tail read:

```python
tail_bytes_size = 32 * 1024
with open(hung_check_log, "rb") as f:
    f.seek(0, os.SEEK_END)
    size = f.tell()
    offset = max(0, size - tail_bytes_size)
    f.seek(offset)
    tail_bytes = f.read()
log_text = tail_bytes.decode("utf-8", errors="replace")
if offset > 0:
    nl = log_text.find("\n")
    if nl >= 0:
        log_text = log_text[nl + 1:]   # drop the partial first line
    log_text = (
        "(truncated; see hung_check.log artifact for"
        " the full output; showing last 32 KiB)\n...\n"
        + log_text
    )
```

So on a multi-MiB `hung_check.log` we now allocate ~32 KiB (the read buffer + the decoded `str`) instead of the whole file.

Verified end-to-end against the same `read_test_results` decoder (`line[3].replace("\\t", "\t").replace("\\n", "\n")`) for empty / 1 KiB / 30 KiB / 40 KiB / 250 KiB / 5 MiB synthetic logs — small files round-trip byte-for-byte; truncated files start on a line boundary and the body is a strict suffix of the original log. Tab and newline survive the TSV escape/decode round-trip on UTF-8 and on byte sequences that don't decode cleanly.

Thanks for the catch — the read pattern was indeed fragile on the very cases the PR is meant to surface.

alexey-milovidov (approved):

PR #103554, opened 2026-04-26
Suppress spurious LOG_ERROR for 4xx during AWS credential-probing (IMDSv2/STS)

### LINKED ISSUE #99140: S3 Function (AWS Client) sends errors when reading normally.
### Company or project name

ClickHouse

Clickhouse logs Bad Request error into system.text_log when reading normally from s3 with s3 function.

25.6.x.x 25.8.x.x

### How to reproduce

Having data in s3 bucket
Try to read with s3 function (wildcards don't matter)
```sql
select * from s3('s3://bucket/folder1/folder2/*.json.gz');
```

You'll see the results, but Click logs 400's error (from different threads: HTTPHandler or ListObjectS3).

### Expected behavior

No error when operating normally.

```text
src/IO/S3/PocoHTTPClient.cpp; void DB::S3::PocoHTTPClient::makeRequestInternalImpl(Aws::Http::HttpRequest &, std::shared_ptr<PocoHTTPResponse> &, Aws::Utils::RateLimits::RateLimiterInterface *, Aws::Utils::RateLimits::RateLimiterInterface *) const
```

Image

### Additional context

_No response_

Krishnachaitanyakc:
## Linked issues

Fixes #99140

## Summary

When ClickHouse runs on EC2 with no `AWS_*` env vars and no STS web-identity token file, every `s3('s3://bucket/...')` query emits one `Response status: 400, Bad Request` `LOG_ERROR` from `PocoHTTPClient::makeRequestInternalImpl` even when the query succeeds. The 400 originates from the IMDSv2 token negotiation `PUT` to `http://169.254.169.254/latest/api/token`. When IMDSv2 is misconfigured / hop-limit-blocked / disabled, the `PUT` returns `400 BAD_REQUEST`. The SDK at `AWSEC2MetadataClient::getDefaultCredentialsSecurely` already handles this gracefully (returns empty, falls through to IMDSv1 / next provider), but `PocoHTTPClient` has logged the response as a `LOG_ERROR` before the SDK can mask it.

## Approach

Mirroring the existing `Expect404ResponseScope` pattern, this introduces `ExpectCredentialProbe4xxScope` — a thread-local RAII counter that signals to `PocoHTTPClient` that 4xx responses inside the scope are expected and should not be logged at error level. The scope wraps the IMDS HTTP call sites in `AWSEC2MetadataClient::getDefaultCredentials` and `getDefaultCredentialsSecurely`, and the STS web-identity refresh path in `AwsAuthSTSAssumeRoleWebIdentityCredentialsProvider::Reload`. `PocoHTTPClient::makeRequestInternalImpl` honors the scope by suppressing `LOG_ERROR` for HTTP 4xx, downgrading to `LOG_TEST` when `enable_s3_requests_logging` is on. Suppression is bounded to the 4xx range so 5xx responses (e.g. genuine regional STS / IMDS infrastructure outages) still log as `ERROR`.

A C++ unit test (`gtest_credential_probe_logging`) covers four cases: no scope active (logs error), credential-probe scope active (suppresses 4xx), 404 scope only with a 400 response (existing 404 scope is unchanged), both scopes active for a 4xx (still suppressed), and credential-probe scope active for a 503 (5xx still logs).

Stops logging spurious `Response status: 400` errors from `PocoHTTPClient` when `s3` table function reads or other S3 operations probe the AWS credential chain on EC2. The errors are emitted by `PocoHTTPClient` for IMDSv2 token negotiation and STS web-identity refresh responses that the AWS SDK already handles silently as fall-through. Genuine 5xx infrastructure failures from the same paths continue to log at error level.

### DIFF
--- a/src/IO/ExpectCredentialProbe4xxScope.cpp
+++ b/src/IO/ExpectCredentialProbe4xxScope.cpp
@@ -1,0 +1,29 @@
+#include <IO/ExpectCredentialProbe4xxScope.h>
+
+#include <base/defines.h>
+
+namespace DB
+{
+
+thread_local size_t expected_credential_probe_4xx_scope_count = 0;
+
+ExpectCredentialProbe4xxScope::ExpectCredentialProbe4xxScope()
+    : initial_thread_id(std::this_thread::get_id())
+{
+    ++expected_credential_probe_4xx_scope_count;
+}
+
+ExpectCredentialProbe4xxScope::~ExpectCredentialProbe4xxScope()
+{
+    /// check that instance is destroyed in the same thread
+    chassert(initial_thread_id == std::this_thread::get_id());
+    chassert(expected_credential_probe_4xx_scope_count);
+    --expected_credential_probe_4xx_scope_count;
+}
+
+bool ExpectCredentialProbe4xxScope::isActive()
+{
+    return expected_credential_probe_4xx_scope_count != 0;
+}
+
+}
--- a/src/IO/ExpectCredentialProbe4xxScope.h
+++ b/src/IO/ExpectCredentialProbe4xxScope.h
@@ -1,0 +1,29 @@
+#pragma once
+
+#include <thread>
+
+namespace DB
+{
+
+/// Inside the inner scope
+/// Remote storage 4xx responses received during AWS credential-provider probing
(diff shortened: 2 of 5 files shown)


MeltonSmith:
Hello, please check my comment in the linked issue https://github.com/ClickHouse/ClickHouse/issues/99140

Issue #102052, opened 2026-04-08
object_shared_data_min_bytes_for_advanced_serialization


UnamedRus:
### Use case

Postpone storage expensive (2x of data itself) serialization until part size big enough/majority of merges are passed.
It also can be used as "second" order control for disk tiering (ie, use map_with_buckets for ebs disk, but switch to advanced when data is ready to tier to s3)

### Describe the solution you'd like

MergeTree setting

PR #103558, opened 2026-04-26
MergeTree: fold JSON dynamic paths into shared data for small Wide parts

### LINKED ISSUE #100960: Big amount of PUTs on S3 disk when using JSON type because of Wide parts
### Company or project name

_No response_

I'm seeing a very big burst of S3 PUT requests when my small amount of data (2GB) hits TTL to S3 disk.

My system looks well configured to me, 200k rows batched inserts, partitions reaching low single digit amount of parts before reaching TTL to S3. 

I can see that my parts are correctly merged and then left alone for a while and then moved, this is the output from `system.part_log`.

Image

From system.events on a single log I see:
```
DiskS3PutObject
25579
DiskS3CompleteMultipartUpload
7
```

`SELECT count(*) FROM system.blob_storage_log WHERE event_type = 'Upload' AND disk_name = 's3';`
```

25757
```

Most of my uploads eventsxare smaller than 1MB?
```
SELECT
    floor(data_size / 1048576) AS size_bucket_mb,
    count() AS upload_count,
    bar(count(), 0, (
        SELECT count()
        FROM system.blob_storage_log
        WHERE (event_type = 'Upload') AND (disk_name = 's3')
    ), 50) AS chart
FROM system.blob_storage_log
WHERE (event_type = 'Upload') AND (disk_name = 's3')
GROUP BY size_bucket_mb
ORDER BY size_bucket_mb ASC

Query id: b31b32f9-5af2-47f3-8a7e-7290ec24324f

    ┌─size_bucket_mb─┬─upload_count─┬─chart──────────────────────────────────────────────┐
 1. │              0 │        30264 │ █████████████████████████████████████████████████▊ │
 2. │              1 │           48 │                                                    │
 3. │              2 │           19 │                                                    │
 4. │              3 │            4 │                                                    │
 5. │              4 │            5 │                                                    │
 6. │              8 │            2 │                                                    │
 7. │              9 │            2 │                                                    │
 8. │             15 │            2 │                                                    │
 9. │             16 │            1 │                                                    │
10. │             17 │            1 │                                                    │
11. │             18 │            1 │                                                    │
12. │             19 │            1 │                                                    │
    └────────────────┴──────────────┴────────────────────────────────────────────────────┘
```

A lot of my upload events are very long?
```
SELECT
    disk_name,
    data_size,
    elapsed_microseconds
FROM system.blob_storage_log
WHERE (event_type = 'Upload') AND (disk_name = 's3') AND (floor(data_size / 1048576) = 0)
LIMIT 100

Query id: a919711c-de11-457d-8b0f-bce2ee3dde6b

     ┌─disk_name─┬─data_size─┬─elapsed_microseconds─┐
  1. │ s3        │       538 │                95608 │
  2. │ s3        │      6608 │                77002 │
  3. │ s3        │      6700 │                98380 │
  4. │ s3        │        48 │               229186 │
  5. │ s3        │       430 │               200252 │
  6. │ s3        │       216 │               112012 │
  7. │ s3        │      6615 │               190417 │
  8. │ s3        │        49 │               108965 │
  9. │ s3        │        89 │               150444 │
 10. │ s3        │      2052 │               140326 │
 11. │ s3        │      6615 │               131805 │
 12. │ s3        │      6766 │               236860 │
 13. │ s3        │        49 │               138799 │
 14. │ s3        │      6615 │               101235 │
 15. │ s3        │      6576 │                83064 │
 16. │ s3        │       126 │               113903 │
 17. │ s3        │        60 │               113941 │
 18. │ s3        │      9574 │               324315 │
 19. │ s3        │       169 │               117036 │
 20. │ s3        │        49 │               205984 │
 21. │ s3        │      7069 │               208896 │
 22. │ s3        │        41 │               156699 │
 23. │ s3        │        42 │               179245 │
 24. │ s3        │     10183 │               154207 │
 25. │ s3        │      6690 │               118802 │
 26. │ s3        │     20940 │               152842 │
 27. │ s3        │        48 │               178757 │
``` 

26.2.5.45

### How to reproduce

Clickhouse with 2 shards 2 replicas from altinity operator.
Setup an S3 cold disk policy on a table with TTL.
Write to it up to when the partition is old and low numb or parts (<5), hit TTL to S3. 
No bucket versioning, no replication. 

### Expected behavior

Less number of PUTS to s3 disk for this small amount of data

### Additional context

This is my create table
```
   ┌─statement───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
1. │ CREATE TABLE logs.replicated_preprod_logs                                                                                  ↴│
   │↳(                                                                                                                          ↴│
   │↳    `timestamp` DateTime64(3, 'Etc/UTC') CODEC(DoubleDelta, ZSTD(1)),                                                      ↴│
   │↳    `host` LowCardinality(String) CODEC(ZSTD(1)),                                                                          ↴│
   │↳    `program` LowCardinality(String) CODEC(ZSTD(1)),                                                                       ↴│
   │↳    `platform` LowCardinality(String) CODEC(ZSTD(1)),                                                                      ↴│
   │↳    `environment` LowCardinality(String) CODEC(ZSTD(1)),                                                                   ↴│
   │↳    `level` LowCardinality(String) CODEC(ZSTD(1)),                                                                         ↴│
   │↳    `owner` LowCardinality(String) CODEC(ZSTD(1)),                                                                         ↴│
   │↳    `location` LowCardinality(String) CODEC(ZSTD(1)),                                                                      ↴│
   │↳    `filename` String CODEC(ZSTD(1)),                                                                                      ↴│
   │↳    `message` String CODEC(ZSTD(1)),                                                                                       ↴│
   │↳    `document` JSON CODEC(ZSTD(1))                                                                                         ↴│
   │↳)                                                                                                                          ↴│
   │↳ENGINE = ReplicatedMergeTree('/clickhouse/tables/logs/_tmp_replace_fcba8bed11ce0170_ixzcwabf7c3ad832/{shard}', '{replica}')↴│
   │↳PARTITION BY toDate(timestamp)                                                                                             ↴│
   │↳PRIMARY KEY (location, program, toDateTime(timestamp))                                                                     ↴│
   │↳ORDER BY (location, program, toDateTime(timestamp), timestamp)                                                             ↴│
   │↳TTL timestamp TO VOLUME 'default', timestamp + toIntervalDay(2) TO VOLUME 's3', timestamp + toIntervalDay(14)              ↴│
   │↳SETTINGS storage_policy = 'tiered', index_granularity = 8192 
```
this is my storage policy set up through the clickhouse altinity operator
```
<clickhouse>
            <storage_configuration>
                <disks>
                    <s3>
                        <type>s3</type>
                        <endpoint>XXXXXXXXXXXXXXXXXXXX</endpoint>
                        <access_key_id>XXXXXXXXXXXXXXXXXXXXXX</access_key_id>
                        <secret_access_key>XXXXXXXXXXXXXXXXXXXXXXX</secret_access_key>
                        <metadata_path>/var/lib/clickhouse/disks/s3/</metadata_path>
                        <region></region>
                    </s3>
                    <s3_cache>
                        <type>cache</type>
                        <disk>s3</disk>
                        <path>/var/lib/clickhouse/disks/s3_cache/</path>
                        <max_size>30Gi</max_size>
                    </s3_cache>
                </disks>
                <policies>
                  <tiered>
                      <volumes>
                          <default>
                              <disk>default</disk>
                          </default>
                          <s3>
                              <disk>s3_cache</disk>
                          </s3>
                      </volumes>
                      <perform_ttl_move_on_insert>0</perform_ttl_move_on_insert>
                  </tiered>
              </policies>
            </storage_configuration>
        </clickhouse>
```

Krishnachaitanyakc:
## Linked issues

Fixes #100960

## Motivation

When a `Wide` `MergeTree` part contains a `JSON` (or `Object` / `Dynamic`) column with the default `max_dynamic_paths = 1024`, every dynamic path materializes its own `.bin` / `.cmrk2` substreams, so a single JSON column produces 10 000+ files per part. On S3-backed disks this is one `PUT` per file and dramatic write amplification on small parts — the reporter on #100960 observed ~25 K `PUT` requests for 2 GB of data, with 30 K+ uploads under 1 MiB each.

Existing knobs do not address this cleanly:

- `min_bytes_for_wide_part` flips the *whole* part to `Compact`. For tables that mix a log-style schema (10–20 fixed columns) with one JSON column, that costs every fixed column its per-column read locality just to suppress the JSON column's path fan-out.
- Lowering `max_dynamic_paths` is a schema change with read-path implications and affects every part — past and future — for every workload on that table.

What the file-count problem actually wants is a per-column, write-time decision: keep the part `Wide`, keep the fixed columns per-column, but stop materializing one substream per dynamic JSON path on small parts.

## Approach

Add a new MergeTree setting `object_shared_data_min_bytes_for_advanced_serialization` (default `0` = disabled, recommended `1073741824` = 1 GiB for S3-tiered deployments). When the part's estimated uncompressed size falls below the threshold, `SerializationObject` keeps every dynamic path of `JSON` / `Object` / `Dynamic` columns inside the shared-data substream instead of materializing one substream per path. Only the dynamic-subcolumn column's serialization changes — every non-JSON column of the same part retains its `Wide`-style per-column files.

### Plumbing

`MergedBlockOutputStream` and `MergedColumnOnlyOutputStream` carry an estimated uncompressed-bytes signal into `MergeTreeWriterSettings`. Insert paths reuse `block.bytes()`; the merge path forwards `MergeTask::ctx->sum_uncompressed_bytes_upper_bound`; mutations fall back to the source part's `getBytesUncompressedOnDisk()`. `MergeTreeWriterSettings` derives `force_object_shared_data_only` from the threshold and the estimate. The flag travels through `EnumerateStreamsSettings` and `SerializeBinaryBulkSettings` to `SerializationObject`.

### Wide-only gate

The fold is restricted to `Wide` parts at every decision site (`enumerateStreams`, `serializeBinaryBulkStatePrefix`, `serializeBinaryBulkWithMultipleStreams`). `Compact` parts pack every column into one `.bin`, so the per-column file fan-out problem does not exist there, and `Compact` uses `StatisticsMode::PREFIX` which would commit stale path statistics if folding ran.

### State capture

The fold decision is captured in `SerializeBinaryBulkStateObject::fold_dynamic_paths_into_shared_data` at prefix time, so the data write does not re-read the settings field — defense in depth against settings drift across passes.

### Serialization layout when gated

- `serializeBinaryBulkStatePrefix` records an empty `sorted_dynamic_paths` list and falls back from `advanced` to `map_with_buckets` shared-data serialization (matches the zero-level-parts default).
- `serializeBinaryBulkWithMultipleStreams` folds every dynamic path into a synthesized shared-data column (sorted-merge with the column's existing shared-data entries; NULLs skipped per the existing `serializePathAndValueIntoSharedData` semantics) and writes that.
- The configured `object_shared_data_buckets_for_wide_part` is honored even when the column's shared data is empty, since the writer is about to fold paths into it.
- The reader is unchanged: `deserializeObjectStructureStatePrefix` reads `sorted_dynamic_paths` from disk, so an empty list means "no per-path substreams to enumerate" — the column reconstructs from shared data the same way it does for any path that overflowed `max_dynamic_paths`.

### Side effect

The merge path's `MergedBlockOutputStream` previously passed `total_size_bytes_compressed` into a parameter named `total_size_bytes_uncompressed_`. Routing the new estimate exposed and fixed that mismatch via a separate ctor parameter; existing prewarm-cache call sites are unchanged.

## Test

`04117_json_shared_data_threshold.sql` asserts:

- below the threshold the part stays `Wide` but `system.parts_columns` reports zero per-dynamic-path substreams for the JSON column;
- data is readable both directly after insert and after `OPTIMIZE FINAL`;
- with the threshold disabled (default `0`) the per-dynamic-path substreams reappear;
- tables without dynamic-subcolumn columns are unaffected;
- `Compact` parts are unaffected by the setting.

Verified against the existing `JSON` / `Object` / `Dynamic` test suite — no regressions in 21 related tests (`03553_*`, `03991_*`, `03928_*`, `03917_*`, `03159_*`, `03038_*`, `03762_*`, `04024_*`, etc.).

- Improvement

Added a new MergeTree setting `object_shared_data_min_bytes_for_advanced_serialization` (default `0`, disabled). When set to a positive value, `Wide` parts whose estimated uncompressed size falls below the threshold keep every dynamic path of `JSON` / `Object` / `Dynamic` columns inside the shared-data substream instead of materializing one substream per path. This dramatically reduces file count and S3 `PUT` amplification on small parts that mix log-style schemas with `JSON` columns on S3-backed storage, while preserving per-column read locality for non-JSON columns of the same part. Recommended value for S3-tiered deployments is `1073741824` (1 GiB). Closes #100960.

### DIFF
--- a/src/Core/SettingsChangesHistory.cpp
+++ b/src/Core/SettingsChangesHistory.cpp
@@ -1159,7 +1159,7 @@ const VersionToSettingsChangesMap & getMergeTreeSettingsChangesHistory()
     {
         addSettingsChanges(merge_tree_settings_changes_history, "26.5",
         {
-
+            {"object_shared_data_min_bytes_for_advanced_serialization", 0, 0, "New MergeTree setting that, when set, keeps every dynamic path of a JSON / Object / Dynamic column inside the single shared-data substream while the part's estimated size is below the threshold. Avoids the per-dynamic-path file fan-out (and the corresponding `PUT` amplification on object storage) that an under-filled `Wide` part would otherwise cause. Default 0 disables the gate."},
         });
         addSettingsChanges(merge_tree_settings_changes_history, "26.4",
         {
--- a/src/DataTypes/Serializations/ISerialization.h
+++ b/src/DataTypes/Serializations/ISerialization.h
@@ -350,6 +350,12 @@ public:
         /// Type of MergeTree data part we serialize/deserialize data from if any.
         MergeTreeDataPartType data_part_type = MergeTreeDataPartType::Unknown;
 
+        /// If true, `SerializationObject` should not enumerate any per-dynamic-path
+        /// substreams — every dynamic path stays inside the single `ObjectSharedData`
+        /// substream. Used to prevent `Wide` part file fan-out for small parts when
+        /// `object_shared_data_min_bytes_for_advanced_serialization` is configured.
+        bool force_object_shared_data_only = false;
+
         /// Current level of array. Needed to differentiate stream names of nested array offsets.
         size_t array_level = 0;
     };
@@ -433,6 +439,11 @@ public:
         /// Type of MergeTree data part we serialize data from if any.
         /// Some serializations may differ from type part for more optimal deserialization.
         MergeTreeDataPartType data_part_type = MergeTreeDataPartType::Unknown;
+
+        /// If true, `SerializationObject` folds every dynamic path back into the
+        /// single `ObjectSharedData` substream so the writer does not produce a
+        /// per-path `.bin` file. See `EnumerateStreamsSettings::force_object_shared_data_only`.
+        bool force_object_shared_data_only = false;
     };
 
     struct DeserializeBinaryBulkSettings
--- a/src/DataTypes/Serializations/SerializationObject.cpp
+++ b/src/DataTypes/Serializations/SerializationObject.cpp
@@ -109,6 +109,11 @@ struct SerializeBinaryBulkStateObject: public ISerialization::SerializeBinaryBul
     ColumnObject::Statistics statistics;
     /// If true, statistics will be recalculated during serialization.
     bool recalculate_statistics = false;
+    /// If true, the prefix recorded an empty `sorted_dynamic_paths` and the
+    /// data write must fold every dynamic path of the column into the single
+    /// shared-data substream. Captured at prefix time so the data write does
+    /// not have to re-read the (potentially mutated) settings field.
+    bool fold_dynamic_paths_into_shared_data = false;
 
     /// For flattened serialization only.
     std::vector<std::pair<String, ColumnPtr>> flattened_paths;
@@ -187,6 +192,18 @@ void SerializationObject::enumerateStreams(EnumerateStreamsSettings & settings, 
         {
             sorted_dynamic_paths = structure_state->sorted_dynamic_paths;
         }
+        /// On the write path, when the caller has asked us to keep all paths in the
+        /// shared-data substream (e.g. small Wide part with `object_shared_data_min_bytes_for_advanced_serialization`),
+        /// pretend that the column has no promoted dynamic paths so the writer
+        /// doesn't allocate a substream per path. The same flag is honored in
+        /// `serializeBinaryBulkStatePrefix` (records empty `sorted_dynamic_paths`)
+        /// and in `serializeBinaryBulkWithMultipleStreams` (folds the dynamic
+        /// columns back into shared data before writing). Restricted to Wide because
+        /// the per-column file fan-out problem only exists in Wide.
+        else if (settings.force_object_shared_data_only && settings.data_part_type == MergeTreeDataPartType::Wide)
+        {
+            sorted_dynamic_paths = std::make_shared<VectorWithMemoryTracking<String>>();
+        }
         else
         {
             sorted_dynamic_paths = std::make_shared<VectorWithMemoryTracking<String>>();
@@ -225,8 +242,24 @@ void SerializationObject::enumerateStreams(EnumerateStreamsSettings & settings, 
             if (serialization_version.value == SerializationVersion::V3)
             {
                 shared_data_serialization_version = SerializationObjectSharedData::SerializationVersion(settings.object_shared_data_serialization_version);
+                /// Mirror the prefix-side fallback when the part-bytes threshold has
+                /// disabled the per-path fan-out: keep the substream layout matched
+                /// to what the writer will actually produce on disk. Restricted to
+                /// Wide because the fold itself is restricted to Wide.
+                if (settings.force_object_shared_data_only
+                    && settings.data_part_type == MergeTreeDataPartType::Wide
+                    && shared_data_serialization_version.value == SerializationObjectSharedData::SerializationVersion::ADVANCED)
+                    shared_data_serialization_version.value = SerializationObjectSharedData::SerializationVersion::MAP_WITH_BUCKETS;
                 /// Avoid creating buckets in shared data for Wide part if shared data is empty.
-                if (settings.data_part_type != MergeTreeDataPartType::Wide || !column_object->getOrCalculateStatistics()->shared_data_paths_statistics.empty())
+                /// When folding is active the column's shared data may be empty (paths live in
+                /// `dynamic_paths`) — but the writer will fold them in, so we must enumerate the
+                /// same bucket layout.
+                const bool has_paths_to_fold_into_shared_data = settings.force_object_shared_data_only
+                    && settings.data_part_type == MergeTreeDataPartType::Wide
+                    && column_object && !column_object->getDynamicPaths().empty();
+                if (settings.data_part_type != MergeTreeDataPartType::Wide
+                    || !column_object->getOrCalculateStatistics()->shared_data_paths_statistics.empty()
+                    || has_paths_to_fold_into_shared_data)
                     num_buckets = settings.object_shared_data_buckets;
             }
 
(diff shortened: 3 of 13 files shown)


UnamedRus:
IMO, It's wrong to "punish" all columns bc of single column "problematic" layout.

May be, it make sense to have setting, which will control, when "dynamic column" specification actually create dynamic columns, instead of keep everything in same shared storage?

Related, https://github.com/ClickHouse/ClickHouse/issues/102052

Krishnachaitanyakc:
@UnamedRus modified it based on your suggestion (commit b70bc21)

Krishnachaitanyakc:
@UnamedRus @CurtizJ can you review this when you get a chance?

Avogar:
Using default 1024 dynamic paths with wide JSONs for S3 storage is a bad idea in general, creating lots of files in Wide parts will lead to problems, especially huge memory usage on merges. It's better to use small number of dynamic paths or even set it to 0 and rely complitely on shared data (with buckets or advanced).

I doubt we need a separate setting for this case. Especially when you can just modify `min_bytes_for_wide_part` for the specific tables you need.

Avogar:
I see implementation is about complitely different approach and the PR description is outdated

Krishnachaitanyakc:
Thanks for the review. I tested both your suggestions and you're right — `min_bytes_for_wide_part` raised wins on every metric I measured. Going to close this PR.

**Setup:** stock 26.5.1.529, schema with 15 fixed columns + 1 `JSON` column generating 50 distinct dynamic paths per insert, 5 inserts × 50K rows ≈ 65 MB compressed, local FS.

| metric | baseline (`Wide`) | `min_bytes_for_wide_part = 1 GiB` (`Compact`) | delta |
|---|---:|---:|---:|
| files per part | 664 | 10 | **66× fewer** |
| `INSERT` total | 4.27 s | 3.46 s | −19% |
| `OPTIMIZE FINAL` | 1.79 s | 1.35 s | −25% |
| `SELECT count() WHERE status_code >= 500` | 0.193 s | 0.181 s | −6% |
| `SELECT count() WHERE path LIKE '%users%'` | 0.198 s | 0.181 s | −8% |
| `SELECT host, service, status_code WHERE level = 'ERROR'` | 0.211 s | 0.180 s | −15% |
| `SELECT count() WHERE payload.k1 = 'v1'` | 0.213 s | 0.181 s | −15% |
| `SELECT *` | 0.453 s | 0.227 s | −50% |

The defense for keeping `Wide` rested on per-column read locality for the non-`JSON` columns. The data doesn't support that here — fixed-column reads (`status_code`, `path`, the `host` / `service` / `status_code` projection) are *faster* on `Compact`, not slower. `INSERT` and merge are also faster, so there's no write-side tax either.

For completeness: this is local FS with a dataset that fits in OS cache, so reads are CPU-bound and `Wide`'s per-column-file overhead dominates. On real S3 with multi-GB cold-read workloads the picture could differ — but that's not the workload in #100960, so it doesn't change the conclusion here.

Closing in favor of `min_bytes_for_wide_part`.

PR #103580, opened 2026-04-27
Fix infinite loop on `WITH FILL` for data starting with ±inf

### LINKED ISSUE #102873: `FillingRow::next` missing NaN/Inf check on `row[pos]` causes infinite loop when data starts with ±inf

_Retrospective finding from a historical scan of [PR #90255](https://github.com/ClickHouse/ClickHouse/pull/90255) (merged 2025-11-20). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

Queries with `ORDER BY col WITH FILL` hang infinitely when the Float64 column contains `-inf` (ascending) or `inf` (descending) as the first sorted value

**Root cause:** FillingRow.cpp:167: `if (next_original_row[pos].isNaN() || next_original_row[pos].isInf())` only guards against NaN/Inf in the target row. When `row[pos]` is ±inf, stepping never changes the value (-inf + 1 = -inf in IEEE 754), so `next` always returns true at line 235.

**Why we believe this is a bug:** FillingTransform.cpp:658 sets `filling_row[i] = current_value` (which is -inf for first ASC row) → FillingTransform.cpp:696 calls `filling_row.next(next_row, ...)` in a while(true) loop → FillingRow.cpp:167 only checks `next_original_row[pos]` for NaN/Inf (which is a normal number like 1), not `row[pos]` (which is -inf) → FillingRow.cpp:201-202 steps -inf by 1, which stays -inf (IEEE 754) → FillingRow.cpp:235 returns true → loop repeats forever

**Affected locations:**
- `src/Interpreters/FillingRow.cpp:167` — NaN/Inf check only on next_original_row, not row[pos]
- `src/Interpreters/FillingRow.cpp:201` — next_value = row[pos] copies -inf, step_func keeps it -inf
- `src/Processors/Transforms/FillingTransform.cpp:658` — filling_row[i] = current_value sets -inf from input data
- `src/Processors/Transforms/FillingTransform.cpp:694` — while(true) loop calling next() — infinite when next always returns true

**Impact:** Any query with `ORDER BY float_col WITH FILL` on data containing `-inf` (ascending) or `inf` (descending) hangs indefinitely, consuming CPU and memory until killed or OOM. This affects any user table with Float32/Float64 columns containing ±inf values.

Yes — confirmed on current `master` (commit `3ec236686a88`).

### How to reproduce

_See root cause description and affected locations above._

### Expected behavior

```
-inf
1
2
OK
inf
2
1
OK
```

```
-inf
TIMEOUT
inf
TIMEOUT
```

### Additional context

**Open risks:**
- The `shift` method (line 238) also lacks NaN/Inf checks on `row[pos]`, though `doLongJump` has a 500-iteration bound that prevents infinite loops there

**Suggested fix:** Add a check for `row[pos]` being NaN or Inf before the fill loop. Either add `if (row[pos].isNaN() || row[pos].isInf()) return false;` after line 167, or check after stepping at line 202: `if (equals(next_value, row[pos]) && !equals(row[pos], next_original_row[pos])) return false;` to detect when stepping makes no progress.

Found during automated review of [PR #90255](https://github.com/ClickHouse/ClickHouse/pull/90255).

---
_ClickGapAI · Confidence: HIGH · Severity: P0 · Finding: `h_pr90255_001`_

thevar1able:

Fix infinite loop on `WITH FILL` for data starting with ±inf.

Closes https://github.com/ClickHouse/ClickHouse/issues/102873


alexey-milovidov (approved):
Ok. Usually we prefer adding new tests instead of editing existing ones (as it helps with the analysis on the test run history).

clickgapai:

**Affected code:** `FillingRow::next` in `src/Interpreters/FillingRow.cpp` — core code present in all supported branches.

**Why:** `FillingRow::next` in src/Interpreters/FillingRow.cpp predates all currently supported release branches (file was created well before the 25.8 LTS branch cut). The bug is reachable from any user-supplied SQL of the form `ORDER BY <float_col> WITH FILL` when the first row contains ±Inf — easy to trigger and produces unbounded memory consumption until OvercommitTracker fails the query.

PR #91065, opened 2025-11-27
Apply row policies and PREWHERE after FINAL

yariks5s:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Added new settings `apply_row_policy_after_final` and `apply_prewhere_after_final` to allow applying row policies and `PREWHERE` conditions after `FINAL` processing (useful for ReplacingMergeTree). Fixes https://github.com/ClickHouse/ClickHouse/issues/90986

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `25.12.1.557`
- Backported to: `25.11.10.3`

Issue #54794, opened 2023-09-19
MySQL compatibility: Illegal type String of argument of function toDayOfWeek


slvrtrn:
Required for Tableau Online. 

Sample rendered query:

```sql
SELECT FLOOR(((((7 + DAYOFYEAR(cell_towers.updated)) - 1) +
               DAYOFWEEK(DATE_FORMAT(cell_towers.updated, '%Y-01-01 00:00:00'))) - 1) / 7) AS wk_updated_ok
FROM cell_towers
GROUP BY 1;
```

fails with

```
DB::Exception: Illegal type String of argument of function toDayOfWeek. Must be Date, Date32, DateTime or DateTime64.: In scope SELECT FLOOR(((((7 + DAYOFYEAR(cell_towers.updated)) - 1) + DAYOFWEEK(DATE_FORMAT(cell_towers.updated, '%Y-01-01 00:00:00'))) - 1) / 7) AS wk_updated_ok FROM cell_towers GROUP BY 1. (ILLEGAL_TYPE_OF_ARGUMENT)
```

**How to reproduce**

* Which ClickHouse server version to use: latest master
* Which interface to use, if matters: MySQL
* Sample data for all these tables: cell_towers sample dataset

**Expected behavior**

DAYOFWEEK accepts String argument like in MySQL.

PR #103587, opened 2026-04-27
Do not defer row policy after FINAL on blending MergeTree engines

pufit:
Fixes a security issue tracked in [`clickhouse-private` #54794](https://github.com/ClickHouse/clickhouse-private/issues/54794) / [`COR-1664`](https://linear.app/clickhouse/issue/COR-1664/security-apply-row-policy-after-finaltrue-leaks-sensitive-column).

`apply_row_policy_after_final` was introduced in #91065 for the soft-delete pattern in `ReplacingMergeTree`, where the row policy must not filter out delete-marker rows before they win deduplication. #97279 enabled it by default to restore legacy behaviour broken by an earlier `PREWHERE` optimization.

The deferral, however, is applied uniformly to all `*MergeTree` engines. For engines whose `FINAL` blends column values across multiple source rows (`Coalescing`, `Summing`, `Aggregating`, `Graphite`), the merged row carries data from rows the policy was meant to hide, so applying the policy after `FINAL` leaks sensitive values.

PoC reproduced on master with `clickhouse local`:

```sql
CREATE TABLE test (key String, data1 Nullable(String), data2 Nullable(String))
ENGINE = CoalescingMergeTree ORDER BY key;

INSERT INTO test VALUES ('key', 'sensitive_data',     'top_secret');
INSERT INTO test VALUES ('key', 'not sensitive data', NULL);

CREATE ROW POLICY p ON test USING NOT (data1 = 'sensitive_data') TO ALL;

SELECT * FROM test FINAL SETTINGS apply_row_policy_after_final = 1;
-- key | not sensitive data | top_secret   <-- 'top_secret' leaks
SELECT * FROM test FINAL SETTINGS apply_row_policy_after_final = 0;
-- key | not sensitive data | \N
```

The fix restricts the deferral in `ReadFromMergeTree::deferFiltersAfterFinalIfNeeded` to engines that pick a single source row per dedup group (`Replacing`, `Collapsing`, `VersionedCollapsing`, `Ordinary`). Existing tests for the `ReplacingMergeTree` soft-delete use case (`03742_apply_row_policy_after_final`, `03915_row_policy_should_respect_final_by_default`, `03928_deferred_filters_after_final_in_explain`) keep their semantics. A new `04142_row_policy_after_final_blending_engines` covers the affected engines and confirms the leaks are gone.

`apply_prewhere_after_final` is left unchanged — `PREWHERE` is the user's own predicate, defaults to off, and is not a security boundary.

Fix a leak of sensitive column values when querying with \`FINAL\` on \`CoalescingMergeTree\`, \`SummingMergeTree\`, \`AggregatingMergeTree\` or \`GraphiteMergeTree\` under a row policy. The setting \`apply_row_policy_after_final\` (enabled by default since 26.2) was deferring the row-level filter even on engines that blend column values across rows during \`FINAL\`, which could expose data from rows the policy was meant to hide. The deferral is now restricted to engines that pick a single source row per dedup group.

> *Generated by [Nerve](https://github.com/ClickHouse/nerve)*

### DIFF
--- a/src/Processors/QueryPlan/ReadFromMergeTree.cpp
+++ b/src/Processors/QueryPlan/ReadFromMergeTree.cpp
@@ -2189,7 +2189,25 @@ void ReadFromMergeTree::deferFiltersAfterFinalIfNeeded()
         return;
 
     const auto & settings = context->getSettingsRef();
-    bool defer_row_policy = settings[Setting::apply_row_policy_after_final] && query_info.row_level_filter;
+
+    /// `apply_row_policy_after_final` was introduced for the soft-delete pattern in
+    /// `ReplacingMergeTree`, where the row policy must not filter out delete-marker
+    /// rows before they win deduplication. For engines whose `FINAL` blends column
+    /// values across multiple source rows (`Summing`, `Aggregating`, `Coalescing`,
+    /// `Graphite`), deferring the row policy is unsafe: the merged row can carry
+    /// data from rows the policy was meant to hide, leaking sensitive values.
+    /// Only the engines that pick a single source row per dedup group can safely
+    /// defer the row-level filter.
+    const auto merging_mode = data.merging_params.mode;
+    const bool engine_safe_to_defer_row_policy =
+           merging_mode == MergeTreeData::MergingParams::Replacing
+        || merging_mode == MergeTreeData::MergingParams::Collapsing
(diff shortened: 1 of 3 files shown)


pufit:
@yariks5s I'm not sure if this solution is correct, but I don't see any other design ideas that would prevent leaking rows with `apply_row_policy_after_final` in Aggregating tables

PR #103600, opened 2026-04-27
Add SIGKILL fallback in test_keeper_zookeeper_converter::stop_zookeeper

groeneai:
Per directive at https://github.com/ClickHouse/ClickHouse/pull/100941#issuecomment-3457500944
(fix `test_keeper_zookeeper_converter/test.py::test_acls[True]` flakiness in a separate PR).

`zkServer.sh stop` in the ZooKeeper 3.6.3 binary used by the test sends a single
SIGTERM to the JVM, removes the PID file, and returns — it does not wait for
the JVM to actually exit and has no SIGKILL fallback (see [`zkServer.sh` upstream](https://github.com/apache/zookeeper/blob/release-3.6.3/bin/zkServer.sh#L191-L201)).
The test's `stop_zookeeper` polls `pgrep` for up to 60 seconds and raises
`Failed to stop ZooKeeper in 60 secs` if the JVM has not finished its shutdown
hooks by then.

Under sanitizer-heavy CI builds (MSan in particular), the ZooKeeper JVM and the
MSan-instrumented ClickHouse server share the same Docker container and the
same CPU. MSan only instruments compiled C/C++ code at the LLVM IR level — the
JVM itself is not sanitizer-instrumented, the OpenJDK `java` binary and the
ZooKeeper bytecode run the same way they do in non-sanitizer builds. But the
~2–10× CPU slowdown MSan adds to ClickHouse leaves the JVM with much less
scheduler share. JVM shutdown hooks (log flushes, snapshot finalization, file
close, ZK-internal teardown) are largely synchronous, and under that scheduler
pressure they can occasionally exceed the 60-second `pgrep` poll. This is the
dominant chronic-flakiness mode this test exhibited before #102492 (the
`pgrep` regex fix and `start_zookeeper` retry loop).

Recent CIDB shows zero `test_acls` failures across ~27,000 runs in the 8 days
since #102492 merged, but the historical pattern is sporadic and sanitizer
load can push JVM shutdown over the 60s line at any time.

The genuine ZooKeeper JVM in this test exists only to populate test data for
the `clickhouse keeper-converter` tool — we do not test the JVM itself and a
clean shutdown is unnecessary. Add a SIGKILL fallback that fires after SIGTERM
does not stop the JVM in 60 seconds, so the test can never fail on a slow or
hung JVM shutdown. The bracket trick `[o]rg.apache.zookeeper.server` in the
`pgrep` / `pkill` pattern is preserved (and now factored into a constant) to
keep the kill from matching its own shell process.

CIDB sample failures showing the addressed pattern:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=00ecef1c87cfebeb52a024307b0a76a68518fb53&name_0=MasterCI&name_1=Integration%20tests%20%28amd_msan%2C%206%2F6%29

### Verification

- `python3 -m ci.praktika run "integration" --test "test_keeper_zookeeper_converter/test.py::test_acls" --count 3` — 6/6 passes locally.
- Direct SIGKILL mechanism check inside `clickhouse/integration-test:latest`:
  - `pgrep -f '[o]rg\.apache\.zookeeper\.server'` matches the running JVM.
  - `pkill -9 -f '[o]rg\.apache\.zookeeper\.server'` kills the JVM and does not match itself.
  - After the kill, `pgrep` returns nothing.

...

### Version info
- Merged into: `26.5.1.148`

### DIFF
--- a/tests/integration/test_keeper_zookeeper_converter/test.py
+++ b/tests/integration/test_keeper_zookeeper_converter/test.py
@@ -31,23 +31,48 @@ def start_zookeeper():
             time.sleep(3)
 
 
+# pgrep / pkill -f match against the full command line, which includes the
+# pgrep / pkill process itself. The bracket expression `[o]rg` is a standard
+# trick that keeps `pgrep -f` from matching its own shell process: pgrep's
+# argv contains the literal string `[o]rg.apache.zookeeper.server` (the
+# brackets are passed through verbatim by bash because they are inside single
+# quotes), and the regex `[o]rg\.apache\.zookeeper\.server` does not match
+# that literal string (it needs `o` as the first character, not `[`).
+ZK_JVM_PATTERN = "[o]rg\\.apache\\.zookeeper\\.server"
+
+
+def _zk_jvm_running():
+    return bool(
+        node.exec_in_container(
+            ["bash", "-c", f"pgrep -f '{ZK_JVM_PATTERN}' || true"],
+            nothrow=True,
+        ).strip()
+    )
+
+
 def stop_zookeeper():
+    # `zkServer.sh stop` sends a single SIGTERM to the JVM, removes the PID
+    # file, and returns — it does not wait for the JVM to actually exit and
+    # has no SIGKILL fallback. Under sanitizer-heavy CI builds the JVM can
+    # take a long time to run its shutdown hooks, so we have to do the wait
+    # ourselves and force-kill the JVM if it does not exit in time.
     node.exec_in_container(["bash", "-c", "/opt/zookeeper/bin/zkServer.sh stop"])
-    timeout = time.time() + 60
-    # get_process_pid("zookeeper") uses the regex '^[^ ]*zookeeper' which only
-    # matches the first word of the command line. The ZooKeeper Java process
-    # starts with "java -Dzookeeper..." so it never matches. Use a direct pgrep
-    # for the ZooKeeper main class to properly detect the running JVM.
-    # The bracket expression [o]rg is a standard trick to keep pgrep -f from
-    # matching the shell process that is running pgrep itself (its argv
-    # contains the literal string '[o]rg.apache.zookeeper.server', which the
-    # regex '[o]rg.apache.zookeeper.server' does not match).
-    while node.exec_in_container(
-        ["bash", "-c", "pgrep -f '[o]rg\\.apache\\.zookeeper\\.server' || true"],
-        nothrow=True,
-    ).strip():
-        if time.time() > timeout:
-            raise Exception("Failed to stop ZooKeeper in 60 secs")
+    sigterm_deadline = time.time() + 60
+    while _zk_jvm_running():
+        if time.time() > sigterm_deadline:
+            # SIGTERM did not stop the JVM in time. The JVM is just an
+            # external dependency used to populate test data, so a hard
+            # kill is acceptable — we do not need a clean shutdown.
+            print("ZooKeeper JVM did not exit 60s after SIGTERM; sending SIGKILL")
+            node.exec_in_container(
+                ["bash", "-c", f"pkill -9 -f '{ZK_JVM_PATTERN}' || true"],
+            )
+            sigkill_deadline = time.time() + 10
+            while _zk_jvm_running():
+                if time.time() > sigkill_deadline:
+                    raise Exception("Failed to stop ZooKeeper even after SIGKILL")
+                time.sleep(0.2)
+            return
         time.sleep(0.2)
 
 


groeneai:
**Pre-PR validation gate** (session id: cron:clickhouse-ci-task-worker:20260427-091500)

a) **Deterministic repro?** No deterministic repro exists today — the test has been clean for 8 days since #102492 merged (~27,000 runs, zero failures). This PR is proactive hardening of the same fragile shutdown path that produced the chronic `Failed to stop ZooKeeper in 60 secs` failures before #102492.

b) **Root cause explained?** Yes. ZooKeeper 3.6.3's `zkServer.sh stop` sends a single SIGTERM via the PID file, removes the PID file, and returns. There is no built-in wait for JVM exit and no SIGKILL fallback. Under sanitizer-heavy builds (especially MSan) JVM shutdown hooks can occasionally take longer than the 60s test budget, and the pre-#102492 flakiness pattern (e.g. PR #100332 with 36 hits, PR #102492's own pre-merge runs with 14 hits) showed exactly that mode.

c) **Fix matches root cause?** Yes. The fix adds a SIGKILL escalation after the existing 60s SIGTERM wait, which is exactly the missing piece in `zkServer.sh stop`. It does not change behaviour in the happy path.

d) **Test intent preserved?** Yes. The genuine ZooKeeper JVM is only there to populate test data for `clickhouse keeper-converter` — the test does not assert anything about the JVM's shutdown semantics, so a hard kill after a hung shutdown is safe. No assertions are weakened, no `no-random-*` tags added, no data sizes reduced.

e) **Demonstrated in both directions?** Partial. I cannot reproduce the original failure locally without an MSan binary (the failure was load-induced and stopped reproducing 8 days ago), but I verified the new SIGKILL path independently inside `clickhouse/integration-test:latest`: `pgrep -f '[o]rg\.apache\.zookeeper\.server'` correctly identifies the JVM, `pkill -9 -f '[o]rg\.apache\.zookeeper\.server'` kills it, and the `[o]rg` bracket trick keeps `pkill` from matching its own argv. The happy path is verified by `--count 3` (6/6 pass) and the existing CIDB streak.

f) **Fix is general, not a narrow patch?** Yes. The same `stop_zookeeper` function is used by every test entry point (`restart_and_clear_zookeeper`, `restart_zookeeper`, `generate_zk_snapshot`, `copy_zookeeper_data`, `get_genuine_zk` retry loop) — the SIGKILL fallback covers all of them. The sister test `test_keeper_snapshot_small_distance` has a different (less robust) `stop_zookeeper` that does not even wait for JVM exit; that is out of scope here and not on the directive.

groeneai:
cc @alexey-milovidov @CurtizJ @azat — could you review this? Per @alexey-milovidov's directive on https://github.com/ClickHouse/ClickHouse/pull/100941#issuecomment-3457500944 to fix `test_keeper_zookeeper_converter::test_acls[True]` flakiness. Adds a SIGKILL fallback in `stop_zookeeper` so a slow / hung ZK JVM shutdown can never time the test out. Test-only change; no production code touched.

antonio2368:
> Under sanitizer-heavy CI builds (MSan in particular) JVM shutdown hooks can occasionally exceed 60 seconds,

@groeneai how is JVM connected to our MSan builds?

groeneai:
@antonio2368 good catch — that wording is imprecise, sorry. The mechanism is CPU/scheduling pressure, not direct instrumentation:

- MSan only instruments compiled C/C++ at LLVM IR level. It does **not** instrument the JVM — the OpenJDK `java` binary is unmodified and ZooKeeper bytecode runs the same way it does in non-sanitizer builds.
- In this integration test the ZK JVM runs in the same Docker container as the MSan-built ClickHouse server, sharing the same CPU. The MSan-instrumented server is ~2–10× slower CPU-wise.
- When `zkServer.sh stop` sends SIGTERM, the JVM starts running its shutdown hooks (log flushes, snapshot finalization, file close, ZK-internal teardown). Those are largely synchronous and gated by I/O and scheduler share.
- Under the heavy ClickHouse-side load MSan adds, the JVM can lose enough scheduler share that shutdown hooks stretch past the test's 60-second `pgrep` poll, and we get `Failed to stop ZooKeeper in 60 secs`.

So the path is indirect: MSan slows ClickHouse → ClickHouse starves the co-scheduled JVM → JVM shutdown stretches past 60 s. The fix itself doesn't depend on the mechanism — the genuine ZK JVM is only there to populate test data for `clickhouse keeper-converter`, so a SIGKILL after the existing 60 s SIGTERM wait is safe regardless of why the JVM was slow.

I'll update the PR description to reflect this more accurately.

antonio2368 (approved):

PR #103605, opened 2026-04-27
Fix `04122_serialization_time64_formats` test after `yarik/fix-time64-serialization`

leshikus:
PR #102596 (`yarik/fix-time64-serialization`) added a trailing-garbage check in `deserializeTextCSV`: after parsing an unquoted field as a time string, it now throws `UNEXPECTED_DATA_AFTER_PARSED_VALUE` if the buffer is not fully consumed.

The test was passing `45045123` as an unquoted CSV value. `readTime64Text` parsed only `45` (SS format) and left `045123` as trailing garbage, causing the test to fail across all stateless-test configurations.

Fix: replace the raw integer with `12:30:45.123`, a proper unquoted time string, and rename the section header accordingly.

...

### DIFF
--- a/tests/queries/0_stateless/04122_serialization_time64_formats.reference
+++ b/tests/queries/0_stateless/04122_serialization_time64_formats.reference
@@ -25,9 +25,9 @@
 --- CSV input: deserializeTextCSV with quote char ---
 00:00:00.000
 12:30:45.123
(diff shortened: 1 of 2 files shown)


kssenii:
Looks already fixed in https://github.com/ClickHouse/ClickHouse/pull/103604

leshikus:
Fixed in other PR

Issue #103235, opened 2026-04-21
Umbrella: Fast test (arm_darwin) — server crashes and hangs across many PRs


leshikus:
## Summary

Analysis of 442 \"Server died\" failures from the **Fast test (arm_darwin)** CI job revealed two regression tests that appear in crash logs across many unrelated PRs. These tests are designed to verify that specific queries throw exceptions rather than crash the server — but on builds where the underlying fix is absent (or reverted), they actually crash the server instead.

## Data

- Total \"Server died\" reports analyzed: **442** (from ~449 recent PRs)
- Death causes: **hung** 425, oom 7, sigabrt 5, sigsegv 3, unknown 2
- Of 280 cases with downloadable logs, two test names appear in crash traces far more often than any other:

| Test | Crash log appearances |
|------|----------------------|
| `01404_roundUpToPowerOfTwoOrZero_safety` | 88 / 280 (~31%) |
| `04068_aggregate_null_for_empty_with_null_combinator` | 89 / 280 (~32%) |

39 PRs show **both** tests in their crash logs.

## Test 1: `01404_roundUpToPowerOfTwoOrZero_safety.sql`

```sql
-- repeat() with this length and this number of rows will allocate a huge enough region (MSB set),
-- which will cause roundUpToPowerOfTwoOrZero() to return 0 for such allocation (before the fix),
-- and later repeat() will try to use this memory and will get SIGSEGV.
SELECT repeat('0.0001048576', number * (number * (number * 255))) FROM numbers(65535); -- { serverError TOO_LARGE_STRING_SIZE }
```

On builds without the fix, `FunctionRepeat::executeImpl` allocates memory with MSB set, `roundUpToPowerOfTwoOrZero` returns 0, and the server crashes with SIGSEGV. The crash traces across many PRs show `FunctionRepeat::executeImpl` in the stack.

## Test 2: `04068_aggregate_null_for_empty_with_null_combinator.sql`

```sql
-- Regression test for SIGABRT crash when using aggregate functions with
-- internal-only Null combinator and aggregate_functions_null_for_empty = 1.
-- The crash was caused by dereferencing an empty std::optional from
-- tryGetProperties() without checking has_value() first.
SELECT sumNull(number) FROM numbers(10) SETTINGS aggregate_functions_null_for_empty = 1; -- { serverError UNKNOWN_AGGREGATE_FUNCTION }
```

[...]

PR #103615, opened 2026-04-27
ci: mark tests that require Linux kernel interfaces as no-darwin

leshikus:
Four fast tests rely on Linux-only kernel interfaces unavailable on macOS and fail on the `arm_darwin` fast test runner:

- `01268_procfs_metrics` — reads `OSReadChars` and `OSCPUVirtualTimeMicroseconds` from `/proc` via the Linux `taskstats` netlink interface
- `03010_virtual_memory_mappings_asynchronous_metrics` — queries `VMMaxMapCount` and `VMNumMaps` from `system.asynchronous_metrics`, populated from `/proc/sys/vm/max_map_count` on Linux only
- `03459_socket_asynchronous_metrics` — queries `NetworkTCPSockets` and related metrics from `system.asynchronous_metrics`, populated from `/proc/net` on Linux only
- `03595_pread_threadpool_direct_io` — uses `min_bytes_to_use_direct_io` which relies on `O_DIRECT`, not supported on macOS

Adding the `no-darwin` tag skips these tests on the macOS fast test runner.

Relates to: https://github.com/ClickHouse/ClickHouse/issues/103235

### Version info
- Merged into: `26.5.1.196`


maxknv (approved):
It is also useful to add a comment directly into the test text indicating the reason for using the label.

PR #103626, opened 2026-04-28
Text index: support `ALIAS` column in preprocessor

### LINKED ISSUE #95944: Support ALIAS columns in preprocessor argument for Full-Text Index
### Company or project name

_No response_

### Use case

Currently, the `preprocessor` argument in a Full-Text Index only accepts physical columns. If an ALIAS column is used, ClickHouse throws a` DB::Exception: Preprocessor function must only reference column`.

It would be highly beneficial to allow ALIAS columns in this context to not make copy of data into separate column.

this should work, and even more complex ALIAS expression that allowed in text index.
```sql
SET enable_full_text_index = 1;

CREATE TABLE providers (
    provider Nullable(String),  
    name String ALIAS ifNull(provider, 'default_name'),  
    INDEX name_text_idx(name) TYPE text(
        tokenizer=sparseGrams(3),
        preprocessor=lower(name) -- This fails
    )
)
ENGINE = MergeTree
ORDER BY tuple();
```
```
    Code: 36. DB::Exception: Received from localhost:9000. DB::Exception: Preprocessor function must only reference column: ifNull(provider, 'default_name'), also got: name: When validating secondary index name_text_idx. (BAD_ARGUMENTS)
```

Create separate physical column  - expensive.

### Additional context

Seems that, ClickHouse already performs determinism checks for text indexes without preprocessor. For example, using a non-deterministic function like dictGet in an alias already triggers an error:
```sql
CREATE TABLE providers (
  provider Nullable(String),  
  name String ALIAS dictGet('test_dict', 'value', toUInt16(10)),  
  INDEX name_text_idx(name) TYPE text(
      tokenizer=sparseGrams(3)
    )
  )
ENGINE=MergeTree
ORDER BY tuple();
DB::Exception: Expression must be deterministic but it contains non-deterministic part...
``` 
Since the safety mechanism is already in place, supporting ALIAS resolution in the preprocessor seems like a natural extension of the current functionality.

ahmadov:
Fixes #95944

Fixes the referencing the `ALIAS` columns in the text index `preprocessor` expression.


ahmadov:
I suppose we don't need to backport that since there are workarounds to avoid this bug.

src/Storages/IndicesDescription.cpp
```diff
@@ -125,8 +125,28 @@ IndexDescription IndexDescription::getIndexFromAST(
         throw Exception(ErrorCodes::INCORRECT_QUERY, "Skip index '{}' must have at least one column in its expression", result.name);
 
     if (index_type && index_type->arguments)
+    {
         result.arguments = index_type->arguments->clone();
 
+        if (result.type == TEXT_INDEX_NAME)
```
rschu1ze, line 154: Hmm, I don't really like that this is a text-index only path. Also, we are calling `result.initExpressionInfo(...)` above which runs very similar code. Maybe we can somehow unify?

YjyJeff:
Hi @ahmadov @rschu1ze — we hit this exact bug independently (a `text` index with `preprocessor = lower(alias_col)` fails validation with `Missing columns` / `Preprocessor function must only reference column`), implemented essentially the same fix, and only then found this PR — so I closed my duplicate (#109161) in favor of this one. This is confirmation that the issue is affecting real users: **we urgently need this capability**, as the only workaround is materializing an extra physical column, which is exactly the cost the `ALIAS` is meant to avoid.

It looks like the only thing outstanding is @rschu1ze's design question from 2026-06-16 — whether the text-index-only substitution path can be unified with the alias substitution already done in `initExpressionInfo`. Is there anything blocking beyond that discussion? If it would help move things along, one option is to keep the current narrow, low-risk fix as-is (it's correctly scoped to `text` and to the `preprocessor` value, which avoids the tokenizer/key collision regressions the earlier review flagged) and track the `initExpressionInfo` unification as a separate refactor, so users aren't blocked on the cleanup.

Happy to help however is useful — reviewing, testing, or taking over the branch if @ahmadov is busy. Thanks for the work on this!

rschu1ze (changes requested):
Now that the postprocessor is also merged, could you please check that postprocessor expressions also work with alias columns?

EDIT: I just noticed https://github.com/ClickHouse/ClickHouse/pull/110014

rschu1ze:
Superseded by https://github.com/ClickHouse/ClickHouse/pull/110014

PR #101275, opened 2026-03-30
Automatically create `minmax` + `uniq` statistics for new columns

hanfei1991:
### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
- auto_statistics_types MergeTree setting defaults to "minmax, uniq" — minmax and uniq statistics are created automatically for all suitable columns in new tables                                                 
- materialize_statistics_on_insert defaults to false — statistics are now built during merges rather than at INSERT time, reducing insert overhead. Use SET materialize_statistics_on_insert = 1 to restore the old
   behavior

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.875`

PR #103659, opened 2026-04-28
Skip LLVM Merge job when llvm jobs are partially launched

alexbakharew:

...

Prevent launch of `LLVM Merge` job when only `ci/*` files are changed. When only these files are changed, we do not launch all required FT and IT jobs. Therefore the results of code coverage are lower then they are: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=3c3cc3468ba05b92fb3eada94bf200b01d7e5e9a&name_0=MasterCI

Seen in https://github.com/ClickHouse/ClickHouse/pull/101275#issuecomment-4229745194

cc @fm4v @maxknv 

### Version info
- Merged into: `26.5.1.157`

### DIFF
--- a/ci/jobs/scripts/workflow_hooks/filter_job.py
+++ b/ci/jobs/scripts/workflow_hooks/filter_job.py
@@ -270,14 +270,20 @@ def should_skip_job(job_name):
     ):
         return True, "Skipped, not labeled with 'pr-performance'"
 
-    # If only the functional tests script changed, run only the first batch of stateless tests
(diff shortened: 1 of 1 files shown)


maxknv (approved):

maxknv:
Maybe we can skip all LLVM jobs in this case?

alexbakharew:
> Maybe we can skip all LLVM jobs in this case?

@maxknv 
Also considered this, but I think that if we break the LLVM jobs, we will see it only in master. Better to run at least one to be sure that it is working

PR #103669, opened 2026-04-28
CI: fix 02095_function_get_os_kernel_version on Darwin

leshikus:
Add `02095_function_get_os_kernel_version.Linux.reference` and
`02095_function_get_os_kernel_version.Darwin.reference` so the test
passes on both platforms. Teach `get_reference_file` in `clickhouse-test`
to prefer a `.<OS>.reference` file over the generic `.reference` fallback.
Remove the test from `ci/defs/darwin.skip`.

### Version info
- Merged into: `26.5.1.174`


maxknv:
I'd suggest modifying the test so it produces the same result across platforms and still tests functionality. Is it possible?

leshikus:
@maxknv  The test can be rewritten in bash or jinja, this cannot be done in pure sql. Among three variants claude believes this one is the best because it is most simple.

maxknv:
> @maxknv The test can be rewritten in bash or jinja, this cannot be done in pure sql. Among three variants claude believes this one is the best because it is most simple.
What about this query?
```
SELECT throwIf(                                                                                                                       
      (SELECT value FROM system.build_options WHERE name = 'SYSTEM')
      != splitByChar(' ', getOSKernelVersion())[1],                                                                                     
      'SYSTEM build option does not match kernel version'
  )
```

maxknv (approved):

alexey-milovidov (approved):

PR #103670, opened 2026-04-28
Skip flaky 00971 max_rows_to_read test on parallel replicas

groeneai:

...

### Description

`tests/queries/0_stateless/00971_merge_tree_uniform_read_distribution_and_max_rows_to_read.sql` populates a 1M-row `MergeTree` table, sets `max_rows_to_read = 900000`, and expects `SELECT count() FROM merge_tree WHERE not ignore(*)` to throw `TOO_MANY_ROWS`.

With parallel replicas the scan is split across replicas and `max_rows_to_read` is enforced per replica. Each replica then reads fewer than 900K rows, no replica trips the limit, and the expected error is never raised — the test fails with three lines of `1000000` instead of stderr.

`query_log` from the failing CI run on `Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel)` shows the local replica reads only `786432` rows for the offending query, while the result is still `1000000`. CIDB confirms the failure is concentrated on this single configuration: 14 of 19 hits in the last 60 days, including all 3 master hits.

This is a fundamental incompatibility between a per-query row limit and parallel replicas, not a randomization or timing artifact, so the right fix is to mark the test `no-parallel-replicas` rather than rewrite the assertion.

#### Verification

- `clickhouse-test --no-parallel-replicas` (the `ParallelReplicas` runner): test is `SKIPPED` (matches the failing CI configuration).
- Server with parallel replicas enabled by default profile: `count() WHERE not ignore(*)` returns `1000000` instead of throwing — without the fix, the test fails as in CI.
- Server with parallel replicas disabled (regular CI configuration): test passes, 50/50 manual runs throw `TOO_MANY_ROWS` as expected.

CI report (one of 14 failures used for diagnosis): https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102894&sha=e68143336e75b4bb99f42e16ffa4a39d3edb125e&name_0=PR&name_1=Stateless%20tests%20%28amd_llvm_coverage%2C%20ParallelReplicas%2C%20s3%20storage%2C%20parallel%29

### Version info
- Merged into: `26.5.1.161`

### DIFF
--- a/tests/queries/0_stateless/00971_merge_tree_uniform_read_distribution_and_max_rows_to_read.sql
+++ b/tests/queries/0_stateless/00971_merge_tree_uniform_read_distribution_and_max_rows_to_read.sql
@@ -1,3 +1,8 @@
+-- Tags: no-parallel-replicas
+-- ^ With parallel replicas, `max_rows_to_read` is enforced per replica, so a
+-- 1M-row scan distributed across replicas can stay under the 900K limit on
+-- each replica and the expected `TOO_MANY_ROWS` error is never raised.
+
 DROP TABLE IF EXISTS merge_tree;
 CREATE TABLE merge_tree (x UInt8) ENGINE = MergeTree ORDER BY x;
 INSERT INTO merge_tree SELECT 0 FROM numbers(1000000);


groeneai:
**Pre-PR Validation Gate:**

a) **Deterministic repro?** ✅ CIDB shows 14/19 (74%) failures concentrated on `Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel)` over 60 days; 3 of those are on master. Local repro: with parallel replicas enabled by server profile (mimicking the CI runner), test produces `1000000\n1000000\n1000000` instead of the expected `TOO_MANY_ROWS`; without parallel replicas, 50/50 runs throw correctly.

b) **Root cause explained?** ✅ Server profile sets `enable_parallel_replicas = 1`, `cluster_for_parallel_replicas = default_parallel_replicas`, `parallel_replicas_for_non_replicated_merge_tree = 1`. The 1M-row scan is distributed across replicas; `max_rows_to_read = 900000` is enforced per-replica; each replica reads `< 900K` rows (e.g., `786432` from `query_log`); no replica trips the limit; `count()` aggregates to `1000000`; the expected `TOO_MANY_ROWS` exception is never raised; the `-- { serverError TOO_MANY_ROWS }` directive treats it as a failure.

c) **Fix matches root cause?** ✅ The `no-parallel-replicas` tag both skips the test on the `ParallelReplicas` runner (where the configuration breaks the assertion) and prevents the test runner's random injection of parallel-replicas settings on regular runs.

d) **Test intent preserved?** ✅ The test still asserts that `max_rows_to_read` throws `TOO_MANY_ROWS` when exceeded — only on configurations where this assertion can hold (single-replica, non-distributed). Parallel replicas have a fundamentally different (per-replica) enforcement model, so the assertion cannot be meaningfully evaluated there.

- *Without fix, parallel-replicas server*: test FAILS with `Having 1 errors!` and three `1000000` lines instead of an exception.
- *With fix, parallel-replicas server*: test SKIPPED (`Reason: smth is not supported with parallel replicas`).
- *On regular server*: test passes, 50/50 manual runs throw `TOO_MANY_ROWS`.

Session id: cron:clickhouse-ci-task-worker:20260428-111500

groeneai:
cc @devcrafter — could you review this? Test-only fix that adds the `no-parallel-replicas` tag to `00971_merge_tree_uniform_read_distribution_and_max_rows_to_read`. With parallel replicas, `max_rows_to_read = 900K` is enforced per-replica, so a 1M-row scan distributed across replicas keeps each replica below the limit and the test's expected `TOO_MANY_ROWS` is never raised. CIDB shows 14/14 master+PR failures concentrated on `Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel)`.

alexey-milovidov (approved):
Yes, this is correct.

PR #103673, opened 2026-04-28
Refactor S3 credential chain and AWS logging for region resolution

### LINKED ISSUE #99140: S3 Function (AWS Client) sends errors when reading normally.

MeltonSmith:
S3 sometimes returns HTTP 400 with x-amz-bucket-region when the request was signed for the wrong region; 
 ClickHouse already recovers via getRegionForBucket, but those responses were easy to misread as ordinary failures in logs. They are now classified and logged at INFO with an explicit wrong-signing-region message. 
 Separately, the STS assume-role-with-web-identity provider is only registered when web identity is actually configured (environment, shared config profile, or KMS role ARN override), so typical EC2 or static-key setups no longer see extra warnings from an unused provider. Helper logic for resolving web-identity settings from the environment and profile was consolidated for reuse, and unit tests cover web-identity detection and the wrong-region 400 heuristic.

• Bug Fix (user-visible misbehavior in an official stable release)

S3 client logging now treats HTTP 400 responses that include a non-empty x-amz-bucket-region header (wrong SigV4 signing region) as an informational wrong-region case with a clearer log line instead of the generic error-style path. The STS web identity credentials provider is added to the S3 credentials chain only when web identity is configured, which reduces spurious warnings for deployments that do not use it. Closes #99140.

No documentation change: behavior is logging and internal credential provider selection only; no new settings or SQL surface.

At a minimum, the following information should be added (but add more as needed).
- Motivation: Why is this function, table engine, etc. useful to ClickHouse users?
Sometimes, during a not proper partitioning in s3, wild card usage in s3 function is not the best option (due to listing costs etc)
If we run query per full prefix, there're gonna be tons of queries in the parallel, it means tons of error log messages with 400, which can trigger some alerts (e.g. clickhouse error log rate) and confuse the users.

### Version info
- Merged into: `26.5.1.644`

### DIFF
--- a/src/IO/S3/AWSLogger.cpp
+++ b/src/IO/S3/AWSLogger.cpp
@@ -3,6 +3,8 @@
 #include <IO/Expect404ResponseScope.h>
 #include <Poco/Net/HTTPResponse.h>
 
+#include <string_view>
+
 #if USE_AWS_S3
 
 #include <Core/SettingsEnums.h>
@@ -69,10 +71,32 @@ bool startsWith(const char * str, const char * prefix)
     return *prefix == 0;
 }
 
-bool is404Muted(const char * message)
+/// Case-insensitive substring search; right now only used on the cold 400-error path,
+/// so we don't worry about the implicit strlen here.
+bool containsCaseInsensitive(std::string_view haystack, std::string_view needle)
 {
+    const auto case_eq = [](char a, char b)
+    {
+        return std::tolower(static_cast<unsigned char>(a))
+            == std::tolower(static_cast<unsigned char>(b));
+    };
+    return std::search(haystack.begin(), haystack.end(), needle.begin(), needle.end(), case_eq) != haystack.end();
+}
+
+/// Mute logs from `AWSXMLClient::BuildAWSError` for S3 wrong-signing-region 400
+bool isWrongSigningRegionMuted(const char * message)
+{
+    static const char * prefix = "HTTP response code: 400";
+    if (!startsWith(message, prefix))
+        return false;
+
+    return containsCaseInsensitive(message, "x-amz-bucket-region");
+}
+
     /// This is the way, how to mute scary logs from `AWSXMLClient::BuildAWSError`
     /// about 404 when 404 is the expected response
+bool is404Muted(const char * message)
+{
     if (!Expect404ResponseScope::is404Expected())
         return false;
 
@@ -96,20 +120,26 @@ bool is404Muted(const char * message)
 
     return code == Poco::Net::HTTPResponse::HTTP_NOT_FOUND;
 }
+
+bool isExpectedAwsHttpResponseMuted(const char * message)
+{
+    return isWrongSigningRegionMuted(message) || is404Muted(message);
+}
 }
 
 void AWSLogger::Log(Aws::Utils::Logging::LogLevel log_level, const char * tag, const char * format_str, ...) // NOLINT
 {
-    if (is404Muted(format_str))
+    if (format_str && isExpectedAwsHttpResponseMuted(format_str))
         return;
     callLogImpl(log_level, tag, format_str); /// FIXME. Variadic arguments?
 }
 
 void AWSLogger::LogStream(Aws::Utils::Logging::LogLevel log_level, const char * tag, const Aws::OStringStream & message_stream)
 {
-    if (is404Muted(message_stream.str().c_str()))
+    const auto str = message_stream.str();
+    if (isExpectedAwsHttpResponseMuted(str.c_str()))
         return;
-    callLogImpl(log_level, tag, message_stream.str().c_str());
+    callLogImpl(log_level, tag, str.c_str());
 }
 
 void AWSLogger::callLogImpl(Aws::Utils::Logging::LogLevel log_level, const char * tag, const char * message)
@@ -123,7 +153,7 @@ void AWSLogger::callLogImpl(Aws::Utils::Logging::LogLevel log_level, const char 
 
 void AWSLogger::vaLog(Aws::Utils::Logging::LogLevel log_level, const char * tag, const char * format_str, va_list)
 {
-    if (is404Muted(format_str))
+    if (format_str && isExpectedAwsHttpResponseMuted(format_str))
         return;
     callLogImpl(log_level, tag, format_str); /// FIXME. Variadic arguments?
 }
--- a/src/IO/S3/AWSLogger.cpp
+++ b/src/IO/S3/AWSLogger.cpp
@@ -9,6 +9,7 @@
 
 #include <Core/SettingsEnums.h>
 #include <Common/logger_useful.h>
+#include <boost/algorithm/string/predicate.hpp>
 #include <aws/core/utils/logging/LogLevel.h>
 #include <Poco/Logger.h>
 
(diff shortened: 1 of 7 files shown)


src/IO/S3/AWSLogger.cpp
```diff
@@ -69,10 +71,32 @@ bool startsWith(const char * str, const char * prefix)
...
+/// so we don't worry about the implicit strlen here.
+bool containsCaseInsensitive(std::string_view haystack, std::string_view needle)
+{
+    const auto case_eq = [](char a, char b)
+    {
+        return std::tolower(static_cast<unsigned char>(a))
+            == std::tolower(static_cast<unsigned char>(b));
+    };
+    return std::search(haystack.begin(), haystack.end(), needle.begin(), needle.end(), case_eq) != haystack.end();
+}
```
antonio2368: You can probably use `boost::icontains`

src/IO/S3/AWSLogger.cpp
```diff
@@ -69,10 +71,32 @@ bool startsWith(const char * str, const char * prefix)
...
+
+/// Mute logs from `AWSXMLClient::BuildAWSError` for S3 wrong-signing-region 400
+bool isWrongSigningRegionMuted(const char * message)
 {
+    static const char * prefix = "HTTP response code: 400";
```
antonio2368:
```suggestion
    static constexpr std::string_view prefix = "HTTP response code: 400";
```

src/IO/S3/tests/gtest_aws_s3_client.cpp
```diff
@@ -45,6 +51,13 @@ namespace DB::S3RequestSetting
  * */
 [[maybe_unused]] static Poco::Util::ServerApplication app;
 
+static void restoreEnvVarForAwsS3ClientTests(const char * name, bool had_value, const std::string & saved_value)
+{
+    if (had_value)
+        (void)::setenv(name, saved_value.c_str(), 1); // NOLINT(concurrency-mt-unsafe)
+    else
+        (void)::setenv(name, "", 1); // NOLINT(concurrency-mt-unsafe)
```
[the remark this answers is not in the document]
MeltonSmith: Okay, I'll come back a bit later with the updates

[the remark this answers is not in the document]
MeltonSmith: fixed

src/IO/S3/Credentials.cpp
```diff
@@ -103,6 +103,42 @@ bool areCredentialsEmptyOrExpired(const Aws::Auth::AWSCredentials & credentials,
...
+    // region source is not enforced, but we need it to construct sts endpoint, if we can't find from environment, we should check if it's set in config file.
+    if (resolved.role_arn.empty() || resolved.token_file.empty() || resolved.tmp_region.empty())
+    {
+        auto profile = Aws::Config::GetCachedConfigProfile(Aws::Auth::GetConfigProfileName());
+        if (resolved.tmp_region.empty())
+            resolved.tmp_region = profile.GetRegion();
+        // If either of these two were not found from environment, use whatever found for all three in config file
+        if (resolved.role_arn.empty() || resolved.token_file.empty())
+        {
+            resolved.role_arn = profile.GetRoleArn();
```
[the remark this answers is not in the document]
MeltonSmith: The only caller is S3CredentialsProviderChain which passes credentials_configuration.kms_role_arn. Every place that constructs a CredentialsConfiguration:
1) diskSettings.cpp — 7 positional args, kms_role_arn (8th) left default-empty
2) BackupIO_S3.cpp — same, 7 args, kms_role_arn left empty
3) KeeperSnapshotManagerS3.cpp — 4 args, kms_role_arn left empty
4) GlueCatalog.cpp — named field init, never sets kms_role_arn
5) test — CredentialsConfiguration{}
Nobody ever populates kms_role_arn with a non-empty value.  The override is always empty in production.
I preserved the pre-existing logic from create (which had the same unconditional overwrite).

In theory, this bug may occur later, but it's out of scope for now, it's just unreachable.

src/IO/S3/Credentials.h
```diff
@@ -143,6 +143,9 @@ class AwsAuthSTSAssumeRoleWebIdentityCredentialsProvider : public Aws::Auth::AWS
     static std::shared_ptr<Aws::Auth::AWSCredentialsProvider>
     create(DB::S3::PocoHTTPClientConfiguration & aws_client_configuration, uint64_t expiration_window_seconds_, String role_arn_ = "");
 
+    /// True when a role ARN and `web_identity_token_file` are set (environment, profile, or `kms_role_arn_override` for the role).
+    static bool isWebIdentityConfigured(const String & kms_role_arn_override = {});
```
antonio2368: different arg name used in implementation

src/IO/S3/Credentials.cpp
```diff
@@ -579,34 +615,22 @@ void AwsAuthSTSAssumeRoleWebIdentityCredentialsProvider::CacheKey::updateHash(Si
     hash.update(session_name);
 }
 
+bool AwsAuthSTSAssumeRoleWebIdentityCredentialsProvider::isWebIdentityConfigured(const String & kms_role_arn_override)
+{
+    auto resolved = resolveWebIdentitySettingsFromEnvironmentAndProfile(kms_role_arn_override);
+    return !resolved.token_file.empty() && !resolved.role_arn.empty();
```
antonio2368: We lose signal if only 1 of them is missing which could potentially happen during misconfiguration, maybe still add it to chain if one of them is not empty?

src/IO/S3/PocoHTTPClient.cpp
```diff
@@ -654,6 +663,17 @@ void PocoHTTPClient::makeRequestInternalImpl(
                 /// (e.g. If-None-Match: *), not a genuine error.
                 LOG_INFO(log, "Response status: {}, {}", status_code, poco_response.getReason());
             }
+            else if (isS3WrongSigningRegionBadRequest(status_code, poco_response))
+            {
+                /// Wrong signing region: S3 returns 400 and `x-amz-bucket-region`; `getRegionForBucket` recovers.
```
antonio2368, line 668: This region should be used automatically in next requests, and it's logged with INFO. Do we need to log it here also?

antonio2368:
@MeltonSmith any updates here?

MeltonSmith:
> @MeltonSmith any updates here?

Hello! Yes, I was preparing fixes last 2 days and noticed something

AWSLogger instance is not being created because 
of this gate in [Aws.cpp](https://github.com/ClickHouse/aws-sdk-cpp/blob/22f694afbdc7e9766894998c3745e23f004f8b86/src/aws-cpp-sdk-core/source/Aws.cpp#L53-L59)
```cpp
 if(options.loggingOptions.logLevel != Aws::Utils::Logging::LogLevel::Off)
        {
            if(options.loggingOptions.logger_create_fn)
            {
                Aws::Utils::Logging::InitializeAWSLogging(options.loggingOptions.logger_create_fn());
            }
```
logger_create_fn is specified, but never invoked because of default logging options with logLevel Off in [Client.cpp](https://github.com/ClickHouse/ClickHouse/blob/master/src/IO/S3/Client.cpp#L1153)
```cpp
aws_options.loggingOptions = Aws::LoggingOptions{};
```
So, basically AWSLogger seem to be a dead code. Additionally, judging by logger name which is "AWSClient" (this name is used by both PocoHTTPClient logger and AWSLogger), AWSLogger appears to be a legacy. Correct me if I'm wrong. 

I've noticed this by my "integration test" (which is not committed, because it's basically a scratch) where I created a s3 client by DB::S3::ClientFactory::instance, and started making "head" requests for an existing bucket. Surprisingly, none of AWSLogger.cpp line was invoked :) 

I refactored AWSLogger by using string view like and boost you mentioned and added some gtests, but I'm not sure whether they are really needed now.

src/IO/S3/PocoHTTPClient.cpp
MeltonSmith: That's a good point! Refactored. I left only at INFO lvl
```cpp
"Response status: {}, {}. Wrong signing region.",
```

src/IO/S3/AWSLogger.cpp
MeltonSmith: Fixed

src/IO/S3/AWSLogger.cpp
MeltonSmith: Refactored by using string_view

MeltonSmith:
Seems like rebasing was a bad idea, a tons of commits appeared. Let me fix this

antonio2368:
@MeltonSmith wow I don't believe it but it seems we really did disable SDK logger by accident...

src/IO/S3/Credentials.cpp
MeltonSmith: Done by || operator instead of &&

src/IO/S3/AWSLogger.cpp
```diff
@@ -58,58 +61,57 @@ Aws::Utils::Logging::LogLevel AWSLogger::GetLogLevel() const
...
+bool isWrongSigningRegionMuted(const std::string_view message)
 {
-    while (*prefix && *str == *prefix)
-    {
-        ++str;
-        ++prefix;
-    }
-    return *prefix == 0;
+    static constexpr std::string_view  prefix = "HTTP response code: 400";
+    if (!boost::starts_with(message, prefix))
```
antonio2368: `std::string_view` has its own starts_with https://en.cppreference.com/cpp/string/basic_string_view/starts_with

src/IO/S3/Credentials.cpp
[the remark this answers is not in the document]
antonio2368: It is out of scope of this PR.

src/IO/S3/AWSLogger.cpp
```diff
@@ -58,58 +61,57 @@ Aws::Utils::Logging::LogLevel AWSLogger::GetLogLevel() const
...
+{
     if (!Expect404ResponseScope::is404Expected())
         return false;
 
-    static const char * prefix_str = "HTTP response code: ";
-    static const size_t prefix_len = strlen(prefix_str);
-
-    if (!startsWith(message, prefix_str))
+    static constexpr std::string_view prefix = "HTTP response code: ";
+    if (!boost::starts_with(message, prefix))
```
antonio2368: https://en.cppreference.com/cpp/string/basic_string_view/starts_with

antonio2368 (commented):
Thanks again for finding the issue with the logger.
The fix is here https://github.com/ClickHouse/ClickHouse/pull/104837

src/IO/S3/AWSLogger.cpp
MeltonSmith: Thanks. So should I use stringview's starts_with mainly and  **only**  use boost for icontains?

Like that:

```cpp
bool isWrongSigningRegionMuted(const std::string_view message)
{
    static constexpr std::string_view  prefix = "HTTP response code: 400";
    if (!message.starts_with(prefix))
        return false;

    return boost::icontains(message, "x-amz-bucket-region");
}
```

and 
```cpp
bool is404Muted(std::string_view message)
{
    if (!Expect404ResponseScope::is404Expected())
        return false;

    static constexpr std::string_view prefix = "HTTP response code: ";
    if (!message.starts_with(prefix))
        return false;

    if (message.size() < prefix.size() + 3)
        return false;

    UInt64 code = 0;
    if (!tryParse<UInt64>(code, message.data() + prefix.size(), 3))
        return false;

    return code == Poco::Net::HTTPResponse::HTTP_NOT_FOUND;
}

```

antonio2368: Yes, if you have options in standard library always prefer that.

antonio2368 (approved):

PR #103678, opened 2026-04-28
Make integration tests idempotent for --count re-runs

alexey-milovidov:
Several integration tests previously failed when re-run with `--count > 1`, either because they left behind state (tables, replicas, ZooKeeper paths) or because nodes were killed by OOM during a prior iteration and never restarted. This PR makes them idempotent and OOM-tolerant.

* `test_cluster_discovery/test_auxiliary_keeper.py`: ensure all nodes are running and drop the leftover table locally on each node before the test starts. `ON CLUSTER` cannot be used while some nodes may still be down from a previous run.
* `test_concurrent_ttl_merges/test.py`: enable `stay_alive` on both nodes, restart any node killed by OOM, and reset table/replica state and the `SYSTEM STOP TTL MERGES` flag at the start of each test.
* `test_keeper_remove_rejoin_leader/test.py`: replace `keeper_utils.wait_nodes` with per-node `wait_until_connected` (with a longer timeout), so a slow node does not abort the wait early.
* `test_storage_s3/test.py`: when checking the version of `dummy_old`, treat a query failure as needs-restart instead of propagating the exception, so a node killed by OOM in a previous run is recovered.

These changes were originally part of #100177 and are split out here so the parent PR contains only the `ThreadPool` code changes.

...

### Version info
- Merged into: `26.5.1.158`

### DIFF
--- a/tests/integration/test_cluster_discovery/test_auxiliary_keeper.py
+++ b/tests/integration/test_cluster_discovery/test_auxiliary_keeper.py
@@ -47,6 +47,15 @@ def test_cluster_discovery_with_auxiliary_keeper_startup_and_stop(start_cluster)
     then stop/start some nodes and check that it (dis)appeared in cluster.
     """
 
+    # Ensure all nodes are running (previous run may have left some stopped)
+    for name in ["node1", "node3"]:
+        nodes[name].start_clickhouse(start_wait_sec=60)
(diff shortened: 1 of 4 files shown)


alexey-milovidov (commented):
Looks good.

PR #103683, opened 2026-04-28
Fix deadlock in ALTER DATABASE MODIFY COMMENT with Shared Catalog

evillique:
### Changes:
- DatabasesCommon.cpp: alterDatabaseComment no longer holds the database mutex across the SharedDatabaseCatalog::alterDatabase call (the deadlock), and no longer exposes the new comment to concurrent readers before the catalog accepts it.
- InterpreterAlterQuery.cpp: ALTER DATABASE takes the database-level DDLGuard and resolves the database under it (serializing against RENAME DATABASE), and dispatches ON CLUSTER before the local lookup, matching ALTER TABLE/RENAME/DROP.

...

### Version info
- Merged into: `26.6.1.884`


src/Databases/DatabasesCommon.cpp
```diff
@@ -352,30 +352,57 @@ void DatabaseWithAltersOnDiskBase::alterDatabaseComment(const AlterCommand & com
...
             throw Exception(ErrorCodes::THERE_IS_NO_QUERY, "Unable to show the create query of database {}", backQuoteIfNeed(database_name));
-#if CLICKHOUSE_CLOUD
-        if (managed_by_shared_catalog)
-        {
 
-            auto version_to_wait = SharedDatabaseCatalog::instance().alterDatabase(getUUID(), create_query);
-            query_context->setVersionToWaitSharedCatalog(version_to_wait);
-            return;
-        }
+        auto version_to_wait = SharedDatabaseCatalog::instance().alterDatabase(getUUID(), create_query);
```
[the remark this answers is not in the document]
evillique: Writing a reproduction test would be too complex.

src/Databases/DatabasesCommon.cpp
```diff
@@ -352,30 +352,57 @@ void DatabaseWithAltersOnDiskBase::alterDatabaseComment(const AlterCommand & com
...
 #endif
+
+    std::lock_guard lock{mutex};
+    const String old_comment = comment;
+    comment = new_comment;
+    try
+    {
         const ASTPtr create_query = getCreateDatabaseQueryImpl();
         if (!create_query)
             throw Exception(ErrorCodes::THERE_IS_NO_QUERY, "Unable to show the create query of database {}", backQuoteIfNeed(database_name));
```
tuanpach, line 405: Under `CLICKHOUSE_CLOUD`, because we already generated `create_query = getCreateDatabaseQueryImpl();`, this check here seems redundant (under `CLICKHOUSE_CLOUD`).

If `create_query` is invalid (which is unlikely), the memory and Keeper state will differ.

Under `CLICKHOUSE_CLOUD`, should we add `chassert` to check if the 2 `create_query` (generated before and this place) are equal?

Please skip it if you think it is not necessary.

tuanpach:
This PR looks good to me.
I only have 1 minor comment.

tuanpach (approved):

src/Databases/DatabasesCommon.cpp
evillique: Under `CLICKHOUSE_CLOUD` if the database is Shared, we always exit before the `#endif` -> this block of code and check specifically will not be executed.

tuanpach: I see. Then it is not an issue.

PR #103688, opened 2026-04-29
CI: add `ci-force-all` label to bypass all job filtering

leshikus:
Add a `CI_FORCE_ALL_LABEL` setting (default: `"ci-force-all"`) to Praktika.
When a PR carries this label, both the workflow filter hooks and the
changed-file-based job filtering are skipped, so every job in the workflow
runs unconditionally.

The label name is a project-level Setting so individual deployments can
override it in their `ci/settings/` directory.

### Version info
- Merged into: `26.5.1.203`


maxknv (approved):

leshikus:
Created a separate PR regarding PGO fail https://github.com/ClickHouse/ClickHouse/pull/103788

PR #103707, opened 2026-04-29
Stabilize flaky test_system_view_refreshes_on_not_running_replica

groeneai:
The integration test
`test_refreshable_mv/test.py::test_system_view_refreshes_on_not_running_replica`
has been chronically flaky — 37 hits across 9 unrelated PRs in the last
30 days, also tracked in [#100840](https://github.com/ClickHouse/ClickHouse/issues/100840).

The test races on three things:

1. `SYSTEM REFRESH VIEW` is asynchronous — it requests a refresh but
   does not wait for completion. The test reads `system.view_refreshes`
   immediately after, so on the running replica the stats may still be
   their pre-refresh values (`NULL`/`0`). Failure mode:
   `assert get_view_refresh_value(node1, "written_rows") != "NULL"`
   failing with `'NULL' != 'NULL'`.

2. In a replicated database the refresh-runner replica is chosen via
   Keeper coordination, so calling `SYSTEM REFRESH VIEW` on `node1`
   does not guarantee `node1` actually wins the election. When `node2`
   wins, `last_attempt_replica` becomes `2` and the test's expectations
   on `node1` and `node2` are inverted. Failure mode:
   `assert get_view_refresh_value(node2, "total_rows") == "NULL"`
   failing with `'2' == 'NULL'`.

3. Even when `node1` runs the refresh, `node2`'s cached view of the
   coordination znode is updated lazily. Until that update lands
   `last_attempt_replica` is still empty on `node2`, and
   `StorageSystemViewRefreshes::fillData` falls into the
   `last_attempt_replica.empty()` branch and returns local progress
   (often `0`) instead of `NULL`. Failure mode:
   `'0' == 'NULL'`.

The fix is to make the test deterministic on all three axes:

- Pause refreshes on `node2` (`SYSTEM STOP VIEW`) before triggering the
  refresh, so `node1` always wins the Keeper election.
- Wait for the refresh to finish on `node1` (`SYSTEM WAIT VIEW`).
- Poll `node2` (`assert_eq_with_retry`) until it observes
  `last_refresh_replica = 1`, which guarantees its znode cache is fresh
  before the assertions run.

Also use a unique znode path / database name per test invocation
(matching the pattern already used in `test_refreshable_mv_in_replicated_db`)
so that residual `RefreshTask` state from previous runs cannot leak in
and produce extra rows in `system.view_refreshes` (the rare
`'NULL\n0' == 'NULL'` failure mode).

### Local verification

20 consecutive passes on the `Integration tests (amd_tsan, 1/6)`
configuration via:

```
python3 -m ci.praktika run "integration" \
    --test "test_refreshable_mv/test.py::test_system_view_refreshes_on_not_running_replica" \
    --count 20
```

(the unfixed test fails on ~0.7% of CI runs across many sanitizer
build flavors).

### Pre-PR validation gate

- a) Deterministic repro: yes — the three race conditions above are
  explainable from the source (`StorageSystemViewRefreshes.cpp:104`,
  `RefreshTask.cpp:400`, `InterpreterSystemQuery.cpp:841`); no hidden
  state.
- b) Root cause explained: yes (see body above).
- c) Fix matches root cause: each step targets one of the three axes.
- d) Test intent preserved: yes — same assertions, same expected
  behaviour. The fix only adds explicit synchronisation; it does not
  weaken any assertion.
- e) Both directions: 20/20 passing locally with the fix; CIDB has 30
  days of failure history with the unfixed code (~0.7% rate, too rare
  to repro in a small local loop).
- f) Fix is general: this is a test-only change so f) is N/A.

...

Closes #100840

Session: cron:clickhouse-ci-task-worker:20260429-071500

### Version info
- Merged into: `26.5.1.189`

### DIFF
--- a/tests/integration/test_refreshable_mv/test.py
+++ b/tests/integration/test_refreshable_mv/test.py
@@ -605,21 +605,53 @@ def test_replicated_db_startup_race(started_cluster, cleanup):
 
 
 def test_system_view_refreshes_on_not_running_replica(started_cluster):
+    # Use a unique znode path / database name per invocation so that residual state
+    # from previous test runs (orphaned RefreshTasks, stale Keeper znodes — see
+    # https://github.com/ClickHouse/ClickHouse/issues/76418 ) cannot leak in.
+    global test_idx
+    test_idx += 1
+    db_name = f"test_nrr_{test_idx}"
+    db_path = f"/db/test_nrr_{test_idx}"
+
     try:
         for node in nodes:
-            node.query("DROP DATABASE IF EXISTS test SYNC")
+            node.query(f"DROP DATABASE IF EXISTS {db_name} SYNC")
             node.query(
-                r"CREATE DATABASE test ENGINE = Replicated('/db/test', '{shard}', '{replica}')"
+                f"CREATE DATABASE {db_name} ENGINE = Replicated('{db_path}', '{{shard}}', '{{replica}}')"
             )
 
         node1.query(
-            "CREATE MATERIALIZED VIEW test.rmv REFRESH EVERY 1 HOUR (x Int64) ENGINE ReplicatedMergeTree ORDER BY x AS SELECT number*10 AS x FROM numbers(2)"
+            f"CREATE MATERIALIZED VIEW {db_name}.rmv REFRESH EVERY 1 HOUR (x Int64) "
+            "ENGINE ReplicatedMergeTree ORDER BY x AS SELECT number*10 AS x FROM numbers(2)"
         )
-        node1.query("SYSTEM REFRESH VIEW test.rmv")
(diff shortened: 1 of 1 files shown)


groeneai:
cc @al13n321 — could you review this? It stabilizes the chronically flaky `test_system_view_refreshes_on_not_running_replica` integration test (37 hits / 9 PRs in 30 days, issue #100840) by adding the synchronisation that was missing: SYSTEM STOP VIEW on the not-running replica forces `node1` to win the Keeper election, SYSTEM WAIT VIEW makes the test wait for the refresh to actually finish, and `assert_eq_with_retry` makes `node2` wait until its znode cache reflects `last_attempt_replica = 1` before checking that its stats are NULL.

alexey-milovidov (approved):

PR #103708, opened 2026-04-29
Fix skip-index matching for ALIAS columns with lambda + constants

CurtizJ:

Fixed an issue where text and bloom filter skip indices on `ALIAS` columns whose expression contained a lambda with captured constants (e.g. `arrayMap((k, v) -> concat(k, '=', v), mapKeys(m), mapValues(m)))` were silently ignored

### Version info
- Backported to: `26.4.3.9`, `26.3.11.9`, `26.2.19.11`

### DIFF
--- a/src/Storages/MergeTree/RPNBuilder.cpp
+++ b/src/Storages/MergeTree/RPNBuilder.cpp
@@ -47,7 +47,89 @@ namespace ErrorCodes
 namespace
 {
 
-void appendColumnNameWithoutAlias(const ActionsDAG::Node & node, WriteBuffer & out, const ContextPtr & context, bool use_analyzer, bool legacy = false)
+void appendColumnNameWithoutAlias(const ActionsDAG::Node & node, WriteBuffer & out, const ContextPtr & context, bool use_analyzer, bool legacy = false);
+
+/// Produces the lambda's column name in the AST format `lambda(tuple(args), body)`.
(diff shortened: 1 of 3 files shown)


src/Storages/MergeTree/RPNBuilder.cpp
```diff
@@ -47,7 +47,89 @@ namespace ErrorCodes
...
+{
+    writeString("lambda(tuple(", out);
+    bool first = true;
+    for (const auto & arg : capture.lambda_arguments)
+    {
+        if (!first)
+            writeCString(", ", out);
+        first = false;
+
+        writeString(arg.name, out);
```
alexey-milovidov, line 72: This might be incorrect w.r.t. quoting.

alexey-milovidov: Please add a test with identifiers that require quoting.

CurtizJ: I've added more tests, it works without changes. It looks like quoting a column name here will actually break index analysis because `ASTIdentifier::appendColumnNameImpl` doesn't quote a string as well, and index headers may have column with names generated by `ASTIdentifier::appendColumnNameImpl`.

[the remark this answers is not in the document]
CurtizJ: The suggestion looks incorrect. See https://github.com/ClickHouse/ClickHouse/pull/103708#discussion_r3201230783

PR #103714, opened 2026-04-29
Iceberg: parallelize manifest file loading to fix 3s cold-cache bottl…

il9ue:

- Performance Improvement

Parallelize Iceberg manifest file loading during query planning. On a cold metadata cache, manifest files were fetched and parsed one at a time, so query planning was gated on `N × S3-RTT` of serial latency on tables with many manifests. New setting `iceberg_metadata_files_parallel_loading_threads` (default 8, range 1–64) bounds how many manifests are fetched concurrently; set to 1 for the previous serial behavior.

`iceberg_metadata_files_parallel_loading_threads` (default 8, range 1–64) caps the number of Iceberg manifest files fetched and parsed concurrently when reading a table. 1 disables parallelism. Most impactful on cold-cache queries (first query after server start, snapshot rotation, or `IcebergMetadataFilesCache` eviction) against tables with many manifests. A query answered by the trivial-count optimization (a bare `SELECT count()`) does not run the data iterator and is therefore unaffected.

## Description

On a cold `IcebergMetadataFilesCache`, `SingleThreadIcebergKeysIterator::next()` fetched one manifest per `manifest_list_entry` serially — each an S3 GET plus an Avro parse — so planning cost `N × RTT` before the first file reached the executor. The same serial cost hit the delete-manifest drain in the `IcebergIterator` constructor. Pruning does not avoid it: partition and min/max pruning run *after* each manifest is parsed, so fully-pruned manifests are still fetched. This is most visible in cluster mode, where worker replicas sit idle while the initiator walks manifests serially.

## Fix

`SingleThreadIcebergKeysIterator` lazily fans `getManifestFile` calls onto the IO thread pool on the first `next()`, capping concurrent fetches at `parallel_loading_threads` (clamped to 1–64) via a counting semaphore, and consumes futures in `manifest_list_entries` order so pruning and the `previous_entry_schema` warning logic are unaffected. The delete-manifest drain benefits automatically. Setting = 1 falls through to the original serial loop. Thread-safety rests on existing guarantees (singleflight in `CacheBase::getOrSet`, the exclusive lock in `addIcebergTableSchema`, `SharedMutex` in `AvroForIcebergDeserializer`); the iterator destructor waits for in-flight futures, so the by-value `shared_ptr` captures cannot dangle. These invariants are documented inline at the call sites.

## Observability

- `IcebergManifestFileFetchTaskMicroseconds` — fetch+parse time inside the pool tasks, summed (≈ total work, independent of concurrency).
- `IcebergManifestFilesParallelFetchWaitMicroseconds` — consumer wait on futures (≈ overlapped latency; meaningful under real per-request RTT, near-zero on a warm cache or a latency-free local backend).
- `IcebergManifestFilesParallelFetched` — count delivered via the parallel path (0 on the serial path).
- `ThreadName::ICEBERG_MANIFEST_FETCH` ("IcebergMFetch") for pool threads.

## Tests

What these tests prove (verified locally against a built server):

- Equivalence (stateless `04075_iceberg_parallel_manifest_loading` + integration `test_correctness`): parallel (threads=16) and serial (threads=1) return identical count/sum/rows.
- The parallel path actually runs and is correctly selected (integration `test_branch_selection`, `test_concurrency_cap`): `IcebergManifestFilesParallelFetched` is 0 on the serial path and equals the manifest count on the parallel path. Attribution was cross-checked via both `system.query_log` and `--print-profile-events` (`IcebergManifestFileFetchTaskMicroseconds` > 0, `Fetched` = N), so the feature is not silently serial.
- The [1, 64] clamp is enforced: `threads=100` behaves identically to `threads=64`.
- The measured queries use `sum(id)` / `optimize_trivial_count_query=0`, because a bare `SELECT count()` is answered by the trivial-count optimization and never runs the parallel iterator.


il9ue:
[FYI] Parallel range (1~64) is picked as a guardrail the author picked to look reasonable, not a number derived from anything. However, this can be refined or updated with new performance testcase. (new coverage is on the way, since the workflow hasn't run for awhile)

src/Core/Settings.cpp
```diff
@@ -5351,6 +5265,15 @@ Possible values:
 )", 0) \
     DECLARE(UInt64, iceberg_metadata_staleness_ms, 0, R"(
 If non-zero, skip fetching iceberg metadata from remote catalog if there is a cached metadata snapshot, more recent than the given staleness window. Zero means to always fetch the latest metadata version from the remote catalog. Setting this a non-zero trades staleness to a lower latency of read operations.
+)", 0) \
+    DECLARE(UInt64, iceberg_metadata_files_parallel_loading_threads, 8, R"(
+Number of threads used to fetch Iceberg manifest files in parallel during query planning.
+
+When greater than 1, manifest files listed in the manifest list are fetched concurrently from object storage, reducing cold-cache query latency proportionally to the number of manifests. Each call to `getManifestFile` is backed by `IcebergMetadataFilesCache` with singleflight protection, so concurrent requests for the same key do not cause duplicate S3 fetches. The value is clamped to the range [1, 64]: values above 64 are treated as 64.
```
zvonand: Look at the settings above and below. we do not need these "each call to `getManifestFile` ... `IcebergMetadataFilesCache`..." here. this is a user-facing description of the setting, plain and simple

tests/integration/test_storage_iceberg_no_spark/test_parallel_manifest_loading.py
```diff
@@ -0,0 +1,331 @@
...
+        f"Parallel path fetched {parallel_fetched} manifests via futures; "
+        f"expected {NUM_MANIFESTS}."
+    )
+
+
+# ---------------------------------------------------------------------------
+# 3. Overlap ratio (plain minio / S3)
+# ---------------------------------------------------------------------------
+
+@pytest.mark.xfail(
```
zvonand: this looks like a performance test squeezed into an integration test. this has a potential to be flaky.

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergIterator.cpp
```diff
@@ -150,13 +158,105 @@ std::span<const ProcessedManifestFileEntryPtr> defineDeletesSpan(
...
+
+    /// Reserve to avoid reallocation: std::future is move-only, and vector reallocation
+    /// with non-noexcept move + no copy would cause a compile error on some STL versions.
+    prefetch_entries.reserve(n);
+
+    /// Limit how many getManifestFile calls run simultaneously to parallel_loading_threads.
+    /// Note: all N tasks are submitted to the shared IO pool up front, and every task beyond
+    /// the cap blocks on acquire(). So whenever N > parallel_loading_threads (i.e. routinely,
+    /// e.g. 30 manifests with threads=8 parks 22 tasks) we park shared IO-pool threads on the
+    /// semaphore, not just at very large N. TODO: a consumer-driven sliding window (submitting
```
zvonand: What do you mean by "the semaphore is the simpler fix for now"? this is not a band-aid.

A thread will park at the semaphore and do nothing instead of potentially doing some other work.

It seems to me that we can use `ThreadPoolCallbackRunnerLocal` and its logic here. Not yet sure how to integrate threads number capping there (needs more careful review), but there shall be a way.

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergIterator.cpp
```diff
@@ -150,13 +158,105 @@ std::span<const ProcessedManifestFileEntryPtr> defineDeletesSpan(
...
+    /// at most K tasks and submitting the next as each future is consumed) avoids parking the
+    /// shared pool; the semaphore is the simpler fix for now.
+    auto sem = std::make_shared<std::counting_semaphore<64>>(parallel_loading_threads);
```
zvonand: If my IDE's "find" is right, this is the only occurence of `counting_semaphore`in the whole codebase. 

in combination with your own comment above, this should raise a question: is this the right approach?

src/Core/SettingsChangesHistory.cpp
```diff
@@ -352,6 +296,7 @@ const VersionToSettingsChangesMap & getSettingsChangesHistory()
         });
         addSettingsChanges(settings_changes_history, "25.8",
         {
+            {"iceberg_metadata_files_parallel_loading_threads", 1, 8, "New setting: number of threads to fetch Iceberg manifest files in parallel during query planning."},
```
zvonand: why 25.8 section? it shall go to the latest one.

tests/integration/test_storage_iceberg_no_spark/test_parallel_manifest_loading.py
```diff
@@ -0,0 +1,331 @@
...
+        )
+    finally:
+        instance.query(f"DROP TABLE IF EXISTS {table_name}")
+
+
+# ---------------------------------------------------------------------------
+# 4. Singleflight under concurrency (skipped — needs cross-session coordination)
+# ---------------------------------------------------------------------------
+
+@pytest.mark.skip(
```
zvonand: Either make all necessary fixes and make a working test or remove this. Why add an xfailed test? whom does it help?

src/Core/Settings.cpp
```diff
@@ -5351,6 +5265,15 @@ Possible values:
...
+
+Setting this to 1 disables parallel fetching and restores fully sequential behavior, which is useful for debugging or reproducing exact serial timing.
+
+Valid range: 1–64 (enforced; values above 64 are clamped to 64).
```
zvonand: as per your own [comment](https://github.com/ClickHouse/ClickHouse/pull/103714#issuecomment-4713757405): `picked to look reasonable, not a number derived from anything`

this is the best justification I've ever seen

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergIterator.cpp
```diff
@@ -235,9 +369,32 @@ SingleThreadIcebergKeysIterator::SingleThreadIcebergKeysIterator(
     , persistent_components(persistent_components_)
     , log(getLogger("IcebergIterator"))
     , manifest_file_content_type(manifest_file_content_type_)
+    , parallel_loading_threads(
+          std::clamp<UInt64>(
```
zvonand: so, the user may've set "128" and it was implicitly clamped? without even a tiny log message?

src/Core/Settings.cpp
zvonand: A setting shall be validated, not clamped

il9ue:
Thanks for the review. I'll reply inline on the individual points, but the two that actually shape the next revision are below. Flagging my plan here so you can raise any concern before I rework, rather than after.

**1. The xfail/skip tests : removing them.**
The intent was visible placeholders for the deferred latency/overlap proof, but you're right that a test which doesn't run is noise. Worth being explicit about *why* it can't run today: proving the wall-clock win needs per-request S3 **GET** delay, and the in-tree `broken_s3` mock only injects on PUT; I checked, and the alternatives don't work either (local disk has ~no per-fetch RTT so the ratio is meaningless; a perf-suite XML compares old-vs-new binaries rather than threads=1-vs-N; and aggregate throttles like `s3_max_get_rps` cap total throughput, which *hides* the parallelism rather than revealing it). So a `broken_s3` GET-delay hook seems genuinely the only path, and that's a follow-up. Note this doesn't leave us blind: "the parallel path actually runs" is still proven without a proxy via `branch_selection`/`concurrency_cap` (Fetched = 0 serial vs N parallel). What's deferred is only the "it's faster" wall-clock assertion.

**2. The concurrency cap : reworking off `counting_semaphore`.**
Agreed it's the wrong call; parking a shared-pool thread that could do other work is a real problem, and being the only use in the codebase says enough. Following your `ThreadPoolCallbackRunnerLocal` pointer. My current thinking is to bound in-flight submissions consumer-driven; submit the next manifest fetch as each future is consumed, so ≤K are outstanding and nothing parks; rather than capping inside the runner. I expect pitfalls I'm not fully seeing: clean cap integration with the runner, exception/teardown if the iterator is dropped mid-flight, and preserving consumption order. If you have a concern with that direction (or had the cap living in the runner itself), I'd rather hear it now; otherwise I'll prototype it so there's something concrete to react to.

The rest (setting description, SettingsChangesHistory section, validate-instead-of-clamp) I'll just fix

src/Core/Settings.cpp
il9ue: Will rewrite.

src/Core/SettingsChangesHistory.cpp
il9ue: Will move it to the latest version section

src/Core/Settings.cpp
il9ue: Will switch to validation: reject out-of-range values with a clear error stating the range, no silent capping

il9ue:
### Update
Pushed the changes, sharing what they look like and the thinking behind each.

**Concurrency** : counting_semaphore switched to ThreadPoolCallbackRunnerLocal over a dedicated K-sized pool, following the BlobCopierThread::replicateBlobs idiom; tasks submitted in manifest_list order and consumed in that order. Couldn't find a sliding-window pattern anywhere in the tree (capping always seems to be by pool size), so I went with a per-query dedicated pool rather than a shared one like getIcebergCatalogThreadpool, since the shared ones are sized by a fixed server setting and wouldn't track the per-query K.

**Setting** :  Applying with the MaxThreads type actually removes the need for either: it has no upper bound, so there's no out-of-range value to reject or clamp in the first place. Moved from UInt64 + [1,64] clamp to MaxThreads (default 8), matching max_download_threads (0 = auto(cores), 1 = serial). The type also guarantees the value reaching code is >= 1, so the clamp is just gone rather than replaced by a check. Description rewritten without the internal names.

**Teardown** :  moving to give-ownership surfaced a UAF on early drop. Give-ownership tasks aren't tracked by the runner, so its destructor doesn't wait for them, and the task wrapper dereferences the runner's thread_name while running; so on a LIMIT/cancel the runner can be destroyed before the pool drains the queued jobs. It turned up while reading the runner contract rather than from a failing test (the passing tests consume every future, so they never hit it). Fix is a destructor that drains via the runner's own waitForAllToFinish before teardown; the same thing BlobCopier does every round. One trade-off worth flagging: give-ownership can't cancel queued tasks, so an early drop runs the remaining fetches to completion instead of cancelling them; safe, but not free. Ran it under TSAN to be sure: induced a real early drop (LIMIT 1 on a 200-manifest cold table → only 5 consumed, ~195 in flight) across 200 LIMIT + 60 KILL-QUERY iterations, no race/UAF reports.

**Tests** :  merged equivalence + branch-selection into one parametrized test, rewrote the threads test for the MaxThreads semantics (0/1/8), and removed the dead xfail/skip tests plus the duplicate stateless one.

PR #103539, opened 2026-04-25
Fix file cache dynamic resize race with concurrent incrementSize

groeneai:
The stress test (arm_debug / amd_debug) was repeatedly failing with the
following logical error during config reload:

```
Logical error: 'Cannot modify size limits to A in size and B in
elements: not enough space freed. Current size: C/D, elements:
E/F (G) (s3_cache, protected)' (STID 4012-44fe).
```

Despite earlier targeted fixes (#96142, #96980, #102396), the bug
continued to surface on master and unrelated PRs. The remaining race
is at the cache state lock level rather than the priority lock level.

## Root cause

`FileCache::doDynamicResizeImpl` previously released the cache state
lock between sampling `state.size` in `collectEvictionInfoForResize`
and validating `state.size <= max_size_` inside `modifySizeLimits`.
During that gap, concurrent `FileCache::doTryReserve` and SLRU
`SLRUFileCachePriority::tryIncreasePriority` calls could invoke
`incrementSize` on the same protected queue. `incrementSize` only
requires the cache state lock, not the priority write lock, so
holding the priority write lock around `removeQueueEntries` and
`modifySizeLimits` (#96980) was insufficient to keep the protected
queue's `state.size` from growing.

Concretely:

1. Resize thread acquires the state lock, samples
   `protected.state.size = X`, releases state lock.
2. Resize thread collects eviction candidates totalling `X - new_max`
   bytes (via the priority read lock).
3. Resize thread acquires the priority write lock.
4. A concurrent `tryIncreasePriority` for SLRU promotes a
   probationary entry to protected. It performs the queue mutation
   under the priority write lock with size 0, then releases the
   priority write lock and acquires the state lock to call
   `incrementSize(prev_entry->size)` on the protected queue.
5. Resize thread `removeQueueEntries` (atomic `state.sub`); state
   lock is still released.
6. Resize thread re-acquires the state lock and calls
   `modifySizeLimits` — but the concurrent `incrementSize` from step 4
   has now grown `protected.state.size` above the new limit, so the
   guard in `LRUFileCachePriority::modifySizeLimits` throws.

## Fix

Keep the cache state lock held continuously across the entire
in-memory portion of the resize: from sampling eviction info,
through `collectCandidatesForEviction` and `removeQueueEntries`,
all the way through `modifySizeLimits`. The state lock is only
released after limits are applied, before the slow filesystem
eviction (`eviction_candidates.evict`).

`collectCandidatesForEviction` for the resize path uses only the
[...]

PR #103738, opened 2026-04-29
Fix vector OOB in ConvertQueryToCNFPass when WHERE is a correlated subquery (STID 3452-4527)

groeneai:
Fix STID 3452-4527 (P0 AST fuzzer crash in correlated-subquery planning).

## Root cause

The constraint-substitution visitors in `ConvertQueryToCNFPass` (`ComponentCollectorVisitor`, `ColumnNameCollectorVisitor`, `SubstituteColumnVisitor`) only filtered subquery children — `needChildVisit` checked `child_type == QUERY/UNION` but ignored the parent. When the visitor is invoked directly on a `QueryNode` (e.g. CNF + constraints reduce a multi-atom `WHERE` down to a single correlated subquery), the visitor enters the `QueryNode` and the parent of its children is the `QueryNode`. The child filter cannot help — the children (`projection`, `correlated_columns_list`, etc.) are `ListNode` instances, not `QueryNode`/`UnionNode`. So the visitor traverses inside the subquery and operates on its `correlated_columns_list`.

For mixed-type constraints (`a UInt16, b String, ... CONSTRAINT ASSUME a <= b ...`), the substitution wraps the canonical class member in a `_CAST(...)` `FunctionNode` because types differ. The planner later iterates `correlated_columns_list` expecting `ColumnNode`, calls `column->as<ColumnNode>()` on a `FunctionNode` (returns nullptr), and dereferences the null `this` inside `getSourceWeakPointer`. libc++ hardening catches this as `vector[]` index out of bounds inside `std::vector<std::weak_ptr<DB::IQueryTreeNode>>::operator[]`, surfacing as `Received signal 6` (Aborted) on the MSAN AST fuzzer.

## Fix

Add a parent-type check to each visitor's `needChildVisit`: if the parent is a `QueryNode`/`UnionNode`, return false. Constraint substitution is scoped to the current query — nested subqueries are processed independently when the outer `ConvertQueryToCNFVisitor` recurses into them via `enterImpl`.

`ColumnNameCollectorVisitor` previously had no `QUERY`/`UNION` filtering at all (only the existing `query_node_to_component` check); the same parent-and-child check is added there.

## Reproduction

The exact fuzzer-generated AST that crashed on PR #103539 (commit `14d6e0f6`):

```sql
CREATE TABLE constraint_test_strong_connectivity__fuzz_4
(`a` UInt16, `b` LowCardinality(Nullable(String)), `c` LowCardinality(String), `d` String NULL,
 CONSTRAINT c1 ASSUME (a <= b) AND (b <= c) AND (c <= d) AND (d <= a)) ENGINE = TinyLog;

SELECT sumKahanDistinctOrNull(1025)
FROM constraint_test_strong_connectivity__fuzz_4
PREWHERE (c = a) AND (d = b)
WHERE (a >= c) AND (SELECT b >= d)
LIMIT 2147483648;
-- Before: libc++ hardening assertion `__n < size()` failed → SIGABRT
-- After:  raises NOT_FOUND_COLUMN_IN_BLOCK (a separate planner issue, not a crash)
```

The fix turns the crash into a regular planner exception. The remaining `NOT_FOUND_COLUMN_IN_BLOCK` is an independent pre-existing planner issue: when CNF reduces `WHERE` to a single correlated-subquery atom, `collectTableExpressionData` skips the WHERE because its top-level node is a `QUERY`. That registration gap is unrelated to the crash and can be addressed separately.

## Tests

`tests/queries/0_stateless/04143_constraint_correlated_subquery_substitute_columns_crash.sql` exercises:

1. The crash-triggering pattern (`UInt16` + `String` columns with constraint, correlated subquery in `WHERE`) — must not crash; expects `NOT_FOUND_COLUMN_IN_BLOCK`.
2. The same with `PREWHERE` — same expectation, plus `ILLEGAL_PREWHERE` allowed (depends on planner ordering).
3. Sanity check: same constraint + correlated-subquery code path with all-`String` columns (no `_CAST` wrapping) should execute correctly and return the expected count.

## Context

Found by AST fuzzer (amd_msan) on PR #103539. Per @alexey-milovidov's directive on PR #103539, this fix is sent as a separate PR.

Related (different crash, same file area, my own open PR): #103733 fixes the analogous "Bad cast from `ConstantNode` to `ColumnNode`" path in the recursive `replaceToConstants` function.

Report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=103539&sha=14d6e0f664204a8815f9bf7fd496e90a83efd434&name_0=PR&name_1=AST%20fuzzer%20%28amd_msan%29


groeneai:
cc @novikd @KochetovNicolai — could you review this? Adds a parent-type check to `needChildVisit` in three `ConvertQueryToCNFPass` visitors so they don't traverse into a subquery's `correlated_columns_list` when the visitor is invoked directly on a `QueryNode`. Fixes STID 3452-4527 (vector OOB / SIGABRT) found by AST fuzzer (amd_msan) on PR #103539.

groeneai:
**Pre-PR validation gate** (per TASK.md Phase 4 Step 9):

**(a) Deterministic repro?** Yes. Triggered locally (debug build, libc++ hardened mode) with:

```sql
CREATE TABLE t (a UInt16, b String, c String, d String,
    CONSTRAINT c1 ASSUME (a <= b) AND (b <= c) AND (c <= d) AND (d <= a)) ENGINE = TinyLog;
INSERT INTO t VALUES (1, '2', '3', '4');
SELECT count() FROM t WHERE (a >= c) AND (SELECT b >= d)
SETTINGS convert_query_to_cnf=1, optimize_using_constraints=1, optimize_substitute_columns=1, allow_experimental_correlated_subqueries=1;
-- Aborts: libc++ Hardening assertion `__n < size()` failed: vector[] index out of bounds
```

**(b) Root cause explained?** Yes.

1. CNF + `optimize_using_constraints` filter the always-true atom `(a >= c)` (constraint `a <= ... <= a` makes them equal). WHERE collapses to a single atom: the subquery `(SELECT b >= d)`.
2. `optimize_substitute_columns` invokes `SubstituteColumnVisitor::visit(getWhere())`. The argument is the cloned subquery `QueryNode`.
3. `needChildVisit(parent, child)` previously only filtered `child_type == QUERY/UNION`. The visitor's children at this level are `ListNode` (`correlated_columns_list`, `projection`, etc.) — `ListNode` is not `QUERY`/`UNION`, so traversal proceeded.
4. Inside `correlated_columns_list`, the visitor finds `ColumnNode b` (`String`), looks up the canonical class member (`a`, `UInt16`), wraps in `_CAST(a, 'String')` because types differ, and assigns `node = buildCastFunction(...)`. The list now contains a `FunctionNode`.
5. Later, `GlobalPlannerContext::collectTableExpressionDataForCorrelatedColumns` iterates the list expecting `ColumnNode`. `column->as<ColumnNode>()` returns nullptr on the `FunctionNode`. `nullptr->getColumnSource()` reads `weak_pointers` at offset N from null. libc++ hardening sees `size() == 0` and aborts.

**(c) Fix matches root cause?** Yes. Adds the missing parent-type check: when `parent_type == QUERY/UNION`, return false. This stops traversal at subquery boundaries regardless of whether the subquery is a child or the visit root. The same check is applied to all three visitors in the file (`ComponentCollectorVisitor`, `ColumnNameCollectorVisitor`, `SubstituteColumnVisitor`) since they share the same scope-bleed problem.

**(d) Test intent preserved? / New tests added?** New regression test `04143_constraint_correlated_subquery_substitute_columns_crash.sql` exercises:

1. The exact AST fuzzer pattern (`UInt16` + `String` columns, mixed-type constraint, correlated scalar subquery in WHERE) — must not crash. Expects `NOT_FOUND_COLUMN_IN_BLOCK`.
2. The same with `PREWHERE` (matches the original fuzzer query). Expects `NOT_FOUND_COLUMN_IN_BLOCK` or `ILLEGAL_PREWHERE`.
3. Sanity check: same code path with all-`String` columns (no `_CAST` wrapping) executes correctly and returns the expected count — guards against the fix breaking the homogeneous-type substitution path.

Existing tests in the constraint substitution area (`01622_constraints_simple_optimization`) continue to pass.

- Without fix: `clickhouse local` aborts with `libc++ Hardening assertion __n < size() failed: vector[] index out of bounds` (matches the exact CI stack trace `ColumnNode::getSourceWeakPointer` → `getColumnSource` → `collectTableExpressionDataForCorrelatedColumns`).
- With fix: same query raises `Code: 10 NOT_FOUND_COLUMN_IN_BLOCK` (a regular planner exception, not a crash).
- Built with the same toolchain CI uses (Clang 21, debug, libc++ hardening). The MSAN-only manifestation in CI is the same code path; libc++ hardening catches it deterministically in any build with `_LIBCPP_HARDENING_MODE`.

**(f) Fix is general, not a narrow patch?** Yes.

- The fix is applied to all three visitors in the file, not just the one that caused the observed crash. `ComponentCollectorVisitor` and `ColumnNameCollectorVisitor` had the same scope-bleed pattern; if a future query triggered them through the same path, they would corrupt `correlated_columns_list` similarly.
- `replaceToConstants` (a recursive free function in the same file) has the same class of bug and is addressed in #103733 (a separate PR for a different concrete failure: "Bad cast from `ConstantNode` to `ColumnNode`"). Both fixes are necessary for full coverage of the constraint-substitution → subquery-scope-bleed bug class.
- The fix is at the source of the corruption (don't enter foreign scopes), not a defensive check at the crash site (e.g. converting `as<ColumnNode>()` callsites to handle `FunctionNode` would mask the underlying invariant violation).

Session: cron:clickhouse-ci-task-worker:20260429-191500

groeneai:
Closing as superseded. This crash (STID 3452-4527) is already fixed on master by the constraint-optimization correlated-subquery work in c66735859ffb, 55a371b578d3, and 7e7720a2af53 (merged 2026-06-02), refined in 259e1044ba34 and 1035d9f91c95 (2026-06-17).

Master guards the bare-correlated-subquery-as-clause-root case at the invocation site (`run_unless_subquery` over projection/where/prewhere/having), plus `isSubqueryNodeType` child guards in each visitor and in `replaceToConstants`. That is strictly more complete than this PR's per-visitor parent check, and it is covered by the `04299_constraint_subst_correlated_subquery_root` test.

CIDB shows STID 3452-4527 with 0 occurrences in the last 60 days (last seen 2026-05-29, before the master fix merged) and 0 on master at any point. The branch also has unrelated drift conflicts after 13.5k commits. Nothing left to add.

PR #103742, opened 2026-04-29
Bound retries in FuzzQuerySource to prevent hung queries

groeneai:
Investigation of the hung check failure on https://github.com/ClickHouse/ClickHouse/pull/101354 (https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=101354&sha=5da866fbdbe5ef63d3498b9030b65987626e9dce&name_0=PR&name_1=Stress%20test%20%28arm_tsan%29) — the hung query was the new regression test from that PR, executed via `executeASTFuzzerQueries` (`ast_fuzzer_runs=5`). The AST fuzzer mutated the test's `max_query_length` literal into a value that put `FuzzQuerySource::createColumn` into an unbounded retry loop, and the thread was unkillable because the loop never checks `isCancelled`. The same hang is reachable today on master without that PR by running `SELECT * FROM fuzzQuery('SELECT 1', 600, 42) LIMIT 1` — the existing `if (config.max_query_length > 500)` guard resets `fuzz_base` on every iteration when the cap is large and never produces a row.

Per @alexey-milovidov's request on https://github.com/ClickHouse/ClickHouse/pull/101354, sending the fix as a separate PR.

The change:

- Cap retries per row in `FuzzQuerySource::createColumn` at 100 attempts. When the budget is exhausted, the source emits the original (unfuzzed) query as a fallback row so the loop always makes progress.
- Check `isCancelled` at the top of the loop so the source responds to `KILL QUERY`, `max_execution_time`, and Stress test's hung check. If the source is cancelled mid-block, remaining rows are filled with the unfuzzed query to keep the `block_size` invariant; the cancelled pipeline discards the chunk anyway.
- Add `04143_fuzz_query_no_infinite_loop` with three scenarios that hang on master without this fix: cap above 500, cap right at 501, and a tiny cap below the unmutated query's formatted length.

The fix is orthogonal to https://github.com/ClickHouse/ClickHouse/pull/101354. With either the existing `> 500` guard or the tighter `fuzzed_text.size() > config.max_query_length` check from that PR, retries are bounded and the source always returns. Whichever PR lands first, the other becomes a small rebase.

Verified locally:

- `SELECT * FROM fuzzQuery('SELECT 1', 600, 42) LIMIT 5` returns in <1s with the fix; without it, it hangs (`timeout 15` exits 143).
- The PR #101354 hang shape — `SELECT max(length(query)) <= 200, count() FROM (SELECT query FROM fuzzQuery('...', 200, 42) LIMIT 100) SETTINGS ast_fuzzer_runs = 5, ast_fuzzer_any_query = 1` — completes in <1s.
- `04143_fuzz_query_no_infinite_loop` passes 50/50 with `--test-runs 50`.

CI report referenced: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=101354&sha=5da866fbdbe5ef63d3498b9030b65987626e9dce&name_0=PR&name_1=Stress%20test%20%28arm_tsan%29

...


groeneai:
### Pre-PR validation gate

Session id: `cron:clickhouse-ci-task-worker:20260429-201500`

- Master HEAD: `timeout 15 build/programs/clickhouse client -q "SELECT count() FROM (SELECT * FROM fuzzQuery('SELECT 1', 600, 42) LIMIT 5)"` exits 143 (SIGTERM), thread is in `FuzzQuerySource::createColumn` and the query is not killable via `KILL QUERY` or `max_execution_time`.
- The hung CI run on https://github.com/ClickHouse/ClickHouse/pull/101354 matches: `hung_check.log` shows the stuck query is the new `04141_fuzz_query_max_query_length` regression test executed via `executeASTFuzzerQueries` (`ast_fuzzer_runs=5`), elapsed 2178s, `is_cancelled=1`, stack inside `FuzzQuerySource::createColumn` → `FuzzQuerySource::generate` → ... → `executeASTFuzzerQueries`.

**b) Root cause explained?** Yes. `FuzzQuerySource::createColumn` runs `while (row_num < block_size)` with no upper bound on retries. The `if (config.max_query_length > 500)` guard always resets `fuzz_base = query` and `continue`s when the cap is large, so `row_num` never increments. With the tighter `if (fuzzed_text.size() > config.max_query_length)` check from https://github.com/ClickHouse/ClickHouse/pull/101354 the loop also spins when no fuzzed AST fits the cap (e.g. cap below the unmutated query's formatted length). In both shapes the loop produces zero rows and the thread is stuck in user code with no `isCancelled` check.

**c) Fix matches root cause?** Yes. Cap retries per row at 100. When the budget is exhausted, emit the unfuzzed `query` as a fallback row and reset for the next row — guarantees forward progress regardless of which size check is active. Add `isCancelled` at the top of the loop so query cancellation works; on cancellation, fill the remaining rows with the unfuzzed query to preserve the `block_size` invariant (the cancelled pipeline discards the chunk).

**d) Test intent preserved? / New tests added?** New tests added. `04143_fuzz_query_no_infinite_loop` covers three previously hung shapes:
- cap = 1000 (master `> 500` resets always),
- cap = 501 (one above the threshold),
- cap = 1 with a base query of ~45 chars (fuzzer always exceeds the cap).
All three return 5 rows with the fix; without it they hang.

No existing test is changed, no assertion is widened, no `no-random-*` tag is added.

- Without fix (master HEAD `StorageFuzzQuery.cpp`): `timeout 15 ... fuzzQuery('SELECT 1', 600, 42)` exits 143; query stays in `system.processes` forever after `KILL QUERY` runs.
- With fix: the same query returns 5 rows in <1s. `04143_fuzz_query_no_infinite_loop` passes 50/50 via `clickhouse-test --test-runs 50`. The PR #101354 hang shape (`SELECT max(length(query)) <= 200, count() FROM (SELECT query FROM fuzzQuery('SELECT 1 + ... + 10', 200, 42) LIMIT 100) SETTINGS ast_fuzzer_runs = 5, ast_fuzzer_any_query = 1`) returns in <1s.

**f) Fix is general, not a narrow patch?** Yes. The bound is on the `createColumn` retry loop itself, not on the configuration entry, so it covers both observed hang shapes (large-cap-always-reset and small-cap-never-fits) and any other future shape where fuzzing cannot fit the cap. The `isCancelled` check is added so any future hang in this code path is interruptible by `KILL QUERY` and Stress test's hung check rather than requiring a server kill.

I have not searched for other `while (...)` loops in `Storages/Storage*Query.cpp` that could spin without a cancellation check — flagging this as follow-up if reviewers want it audited.

groeneai:
cc @alexey-milovidov @azat — could you review this? Bounds the retry loop in `FuzzQuerySource::createColumn` and adds `isCancelled` so `fuzzQuery(..., > 500)` and similar shapes stop hanging. Triggered by the hung check on https://github.com/ClickHouse/ClickHouse/pull/101354.

groeneai:
Thanks for the catch — pushed a follow-up commit that drops the backfill (commit 47c2deff75bd).

**What changed:**
- `FuzzQuerySource::createColumn` now resizes offsets to the actual row count on cancellation (no `O(block_size * query_length)` backfill).
- `FuzzQuerySource::generate` checks `isCancelled` upfront and after each `createColumn` call, and returns an empty `Chunk` if any column is short. The pipeline treats this as end-of-stream and stops pulling.
- The previous commit's bounded retries still guarantee non-cancelled queries always make progress.

**Verification:**
- Existing regression test `04143_fuzz_query_no_infinite_loop`: 50/50 passes with `--test-runs 50`.
- Cancellation latency with `max_block_size = 65536`: `SELECT count() FROM (SELECT * FROM fuzzQuery('<200-term sum>', 100000, 42) LIMIT 1)` returns in ~70 ms.
- `max_execution_time = 1s` on an unbounded scan exits within ~5 ms of the deadline (1004 ms total).

cc @alexey-milovidov @azat

groeneai:
Addressed the second bot inline concern (wasted attempts at `max_query_length > 500`) in commit `12d7ed385567`. With this and the previous two commits, both bot concerns are resolved.

**What changed in this commit:**

1. Replaced the lazy `if (config.max_query_length > 500)` guard with the precise `if (fuzzed_text.size() > config.max_query_length)` check from PR #101354. The old guard always reset `fuzz_base = query` when the cap was > 500, regardless of the actual fuzzed size — which meant the 100-attempt cap was paying `O(max_attempts_per_row * block_size)` of fuzz/format cost per scan before falling back. The precise check resets only when the produced fuzzed text actually exceeds the cap; for the common case (`cap > 500` with a small input query) the loop produces a row on the first attempt.

2. Reject `max_query_length = 0` in `getConfiguration` with `BAD_ARGUMENTS`. With cap = 0, every non-empty fuzzed AST exceeds the cap by definition, so the source would degrade to emitting the unfuzzed query for every row (after `max_attempts_per_row` retries) — a no-op fuzzer with extra cost.

Both source-code changes are folded in from PR #101354. With this commit, PR #101354 only needs to keep its workflow / CI re-enablement bits.

**Pre-PR validation gate (combined: all three commits):**

a) **Deterministic repro?** ✅ Without fix: `SELECT * FROM fuzzQuery('SELECT 1', 600, 42) LIMIT 1` hangs forever on master (commits 1+2 made it terminate, commit 3 makes it correct). With fix: returns in <1 s. Wasted-attempts case (`LIMIT 1000` with `cap = 1000`): 113 ms with fix, would be ~10 s with the old `> 500` guard + 100-attempt cap.

b) **Root cause explained?** ✅ The lazy guard `config.max_query_length > 500` was a coarse heuristic that conflated two distinct conditions: "cap is large" and "fuzzed text exceeds cap". For `cap > 500`, every iteration took the reset branch, never producing a row (the original hang). With the 100-attempt cap, every row paid 100 wasted fuzz attempts (the second bot concern). Replacing the guard with `fuzzed_text.size() > config.max_query_length` directly addresses both. `max_query_length = 0` is a degenerate input that produces a no-op fuzzer with extra cost — rejected up front.

c) **Fix matches root cause?** ✅ Direct: replaces the bad heuristic with a precise check. The 100-attempt cap is kept as defense-in-depth for the small-cap case where the fuzzer keeps producing oversized variants.

d) **Test intent preserved?** ✅ `04143_fuzz_query_no_infinite_loop` adds a fourth scenario (`max_query_length = 0` → `BAD_ARGUMENTS`); the original 3 scenarios (cap = 1000, cap = 501, cap = 1) still pass.

e) **Both directions demonstrated?** ✅ 50/50 test runs. Cancellation: `max_execution_time = 1` exits within ~5 ms of deadline (cancellation latency is now ≤ block_size of pre-cancel work, plus header allocation).

f) **Fix is general?** ✅ Single fault location in `FuzzQuerySource::createColumn` and one input-validation location in `StorageFuzzQuery::getConfiguration`. Searched for sibling `max_query_length > 500` and similar heuristics — none. Defense-in-depth (100-attempt cap, partial-column shrink on cancel, BAD_ARGUMENTS on cap=0) covers the remaining edge cases.

cc @alexey-milovidov @azat @PedroTadim — ready for review.

groeneai:
**Third bot concern (contract violation on fallback) addressed in `1060162e6642`.**

The bot is right: when `max_query_length` is below the formatted base query's length, the fallback path emitted the original query whose size already exceeded the cap, silently violating the documented `max_query_length` output-length contract. Test 4 in the previous push (`cap = 1` with a long base query) was relying on exactly that broken fallback.

**Fix:** moved the validation up front into `StorageFuzzQuery::getConfiguration`, mirroring the existing `max_query_length == 0` rejection. The check parses the base query, formats it, and rejects with `BAD_ARGUMENTS` if `formatForErrorMessage().size() > max_query_length`. After the check passes, every subsequent emission path — fuzzed variants (precisely bounded by the in-loop `fuzzed_text.size() > config.max_query_length` check) and the per-row fallback (which re-emits the original `query`) — is guaranteed to satisfy the cap, so the contract is preserved without weakening it.

This covers both call sites (storage engine `CREATE TABLE` and table function `SELECT FROM fuzzQuery(...)`) since `TableFunctionFuzzQuery::parseArguments` delegates to the same `StorageFuzzQuery::getConfiguration`. The single-shot SQL function `FunctionFuzzQuery` doesn't have a `max_query_length` parameter and is unaffected.

**Pre-PR validation gate (a–f) — all GREEN:**

- **(a) Deterministic repro:** `SELECT * FROM fuzzQuery('SELECT 1 + 2 + ... + 10', 1, 42) LIMIT 5` — without the fix returns 5 rows of 61 chars each (contract violation); with the fix throws `Code: 36 BAD_ARGUMENTS: FuzzQuery max_query_length (1) is smaller than the formatted base query length (61); fuzzer cannot produce any rows that satisfy the cap`.
- **(b) Root cause:** for `max_query_length < format(base_query).size()` no fuzzed variant can fit the cap; the 100-attempt fallback then emits the unfuzzed base query whose size > cap → contract violation. The fix removes the impossible state at construction.
- **(c) Fix matches root cause:** rejection happens at the configuration boundary, before any source pipeline runs. Same shape as the existing `max_query_length == 0` rejection.
- **(d) Test intent preserved:** scenarios cap=1000 and cap=501 still exercise the no-hang invariant under valid configurations. The cap=1 scenario is now rejected up front (which is the correct contract). The cap=0 scenario remains a `serverError BAD_ARGUMENTS` assertion.
- **(e) Both directions demonstrated:** test 04143 passes 50/50 with randomization; sibling test 03834 still passes 20/20.
- **(f) Fix is general:** boundary check confirmed at exact cap (`cap=8` with `'SELECT 1'` 8 chars passes), one byte under (`cap=7`, 8 chars rejected), and zero (rejected by the existing check). Both call paths (storage and table function) covered.

**Recap of all three bot concerns now addressed in the branch:**

1. `47c2deff75bd` — drop cancellation backfill; return empty Chunk on cancel (1st concern: KILL/timeout latency)
2. `12d7ed385567` — replace `if (config.max_query_length > 500)` with precise `if (fuzzed_text.size() > config.max_query_length)`; reject `max_query_length = 0` (2nd concern: O(100×block_size) wasted fuzz attempts; folds in the tighter check from #101354)
3. `1060162e6642` — enforce `max_query_length` contract up front in `getConfiguration` (3rd concern: silent oversized rows on fallback)

cc @alexey-milovidov @azat

groeneai:
Pushed `b704364cd8a3` to address the **fourth bot concern** (backward compatibility for existing FuzzQuery tables).

## What changed

`StorageFuzzQuery::getConfiguration` now takes a `LoadingStrictnessLevel mode` parameter (default `CREATE`). The two new contract checks introduced in the previous commit (`max_query_length == 0` and `formatted_base.size() > max_query_length`) only fire when `mode < ATTACH`. The factory lambda passes `args.mode` from `StorageFactory::Arguments`; the table function path keeps the default since every `SELECT FROM fuzzQuery(...)` is a fresh construction.

This matches the existing convention in `StorageStripeLog`, `StorageKafka`, and others (`if (mode < LoadingStrictnessLevel::ATTACH)`). The runtime is already safe in the lenient case — the `FuzzQuerySource` loop caps retries per row and falls back to the unfuzzed query — so an attached table with a stale configuration cannot hang or corrupt anything.

## Pre-PR validation gate (a–f)

- **(a) Deterministic repro?** Yes. CREATE: `CREATE TABLE fq_t (q String) ENGINE = FuzzQuery('SELECT 1+2+3+4+5+6+7+8+9+10', 1, 42)` -> `BAD_ARGUMENTS`. ATTACH (Ordinary DB): same args -> succeeds.
- **(b) Root cause explained?** `getConfiguration` is called from both the CREATE path (mode=CREATE) and the ATTACH path (mode=ATTACH or higher) of the storage factory. Unconditional validation rejected previously-tolerated metadata at server startup, breaking the backward-compatibility rule.
- **(c) Fix matches root cause?** Yes — gates the new validation on `mode < ATTACH`, the standard ClickHouse pattern for "this is a fresh creation". Existing objects keep loading; new ones are validated.
- **(d) Test intent preserved? / New tests added?** Yes — extended `04143_fuzz_query_no_infinite_loop` with both directions: `CREATE` paths still throw `BAD_ARGUMENTS`, `ATTACH TABLE name (cols) ENGINE = FuzzQuery(...)` in an `Ordinary` database succeeds with the same bad args. The Ordinary DB is needed because `ATTACH TABLE ... ENGINE = ...` requires no pre-existing UUID there (Atomic requires UUID for full ATTACH).
- **(e) Demonstrated in both directions?** Yes:
  - Without fix (HEAD~1): `ATTACH TABLE ... ENGINE = FuzzQuery('SELECT 1+...', 1, 42)` -> `BAD_ARGUMENTS` (broken compat).
  - With fix (HEAD): same ATTACH succeeds; CREATE still throws.
- **(f) Fix is general?** Yes — single mode parameter handles all `LoadingStrictnessLevel` values: CREATE/SECONDARY_CREATE -> strict; ATTACH/FORCE_ATTACH/FORCE_RESTORE -> lenient. Both new validations (cap=0 and cap < base size) are gated together. Sibling test `03834` still passes (no regression). Original hang shape (`fuzzQuery('SELECT 1', 600, 42) LIMIT 1`) still returns in ~70ms; wasted-attempts proof (`LIMIT 1000` at cap=1000) returns in ~113ms; cancellation latency under `max_execution_time=1s` exits within ~5ms of the deadline.

## Branch state — five commits ahead of master

- `b704364cd8a3` StorageFuzzQuery: gate strict validation to fresh creation only ← this run, 4th bot concern (compat)
- `1060162e6642` StorageFuzzQuery: enforce max_query_length contract up front (3rd bot concern)
- `12d7ed385567` FuzzQuerySource: precise size check + reject max_query_length=0 (2nd bot concern, folds in #101354's check)
- `47c2deff75bd` FuzzQuerySource: drop cancellation backfill; return empty Chunk (1st bot concern)
- `614df1d6a0a0` Bound retries in FuzzQuerySource to prevent hung queries (original hang fix)

All four bot concerns now resolved. Waiting on CI + re-review. cc @alexey-milovidov @azat

groeneai:
Superseded by #101354 (merged 2026-06-14), which folded equivalent `FuzzQuerySource` fixes into master: the `max_query_length == 0` and `base_formatted_size > max_query_length` BAD_ARGUMENTS checks, bounded per-row retries with an unfuzzed-query fallback, cancellation-aware column truncation, and the regression test `04141_fuzz_query_max_query_length`.

The runtime hang and argument validation this PR targeted are now fixed on master, so closing rather than rebasing (as noted in the original description: whichever PR lands first, the other becomes a rebase).

The only delta here not adopted upstream was gating the new validation behind `LoadingStrictnessLevel` so pre-existing tables still ATTACH after upgrade. Leaving that to the maintainers, since #101354 intentionally validates unconditionally.

PR #103769, opened 2026-04-30
Text index: extract text index analysis into `MergeTreeTextIndexAnalyzer`

ahmadov:
Extract text index analysis (tokenizer, preprocessor, dictionary and posting list parameters) from `MergeTreeIndexText` into a shared `MergeTreeTextIndexAnalyzer` module.

Add a new internal `textIndexAnalyzer` SQL function that accepts the text index definition string and tokenizes input using the same tokenizer and preprocessor. The function reuses the extracted analysis and supports the `{input}` placeholder in preprocessor expressions.

Motivation for this PR:
While working on #103172, I noticed that positional information for the hackernews dataset is huge on disk comparing to other systems. During debugging, I found some tokens occurs a lot in the dataset, even on the same row. Some of those tokens are either very short (1-2 characters) or commonly used prepositions in English. To avoid such tokens I opened these two PRs: #103715 and #103716 to filter out such tokens. These changes looks simple, however functions like `hasAnyTokens`, `hasAllTokens` and `hasPhrase` use the same filtering and tokenization from the text index definition. Forwarding filtering options to those functions is not straightforward, and it might requires some changes. To avoid for today and future, I decided to extract out text index analysis (filtering and tokenization) to a dedicated class and re-use it in those functions. All functions accept a 3rd `tokenizer` argument, instead of forwarding filtering params as 4th, 5th, 6th, ... arguments, we can forward them in a single argument as defined in the text index e.g. `tokenizer = splitByNonAlpha, preprocessor = lower(...)`.

- Improvement

Move out text index analysis into a dedicated common place.

### DIFF
--- a/src/Functions/textIndexAnalyzer.cpp
+++ b/src/Functions/textIndexAnalyzer.cpp
@@ -1,0 +1,277 @@
+#include <Columns/ColumnArray.h>
+#include <Columns/ColumnFixedString.h>
+#include <Columns/ColumnString.h>
+#include <Common/Exception.h>
+#include <DataTypes/DataTypeArray.h>
+#include <DataTypes/DataTypeString.h>
+#include <Functions/FunctionFactory.h>
+#include <Functions/FunctionHelpers.h>
+#include <Functions/IFunction.h>
+#include <Interpreters/Context.h>
+#include <Interpreters/ExpressionActions.h>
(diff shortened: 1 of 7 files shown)


ahmadov:
On hold until https://github.com/ClickHouse/ClickHouse/pull/98939 is merged.

PR #103780, opened 2026-04-30
Relax brittle 'receive timeout' assertion in test_distributed_frozen_replica

groeneai:
The integration test `test_distributed_frozen_replica/test.py::test_cluster_all_replicas_query` fails intermittently because it asserts the literal substring `"receive timeout 857 ms"` (the configured `handshake_timeout_ms`) appears in the failure message — but only one of three socket-timeout code paths produces that exact phrasing.

### Root cause

Three places in the codebase format socket-read-timeout errors:

| Path | Format string | Contains "receive timeout"? |
|---|---|---|
| `src/Common/AsyncTaskExecutor.cpp:124` (async I/O fiber) | `({}, receive timeout {} ms)` | Yes |
| `src/Client/HedgedConnections.cpp:449` (hedged-connection failover) | `({}, receive timeout {} ms)` | Yes |
| `src/IO/ReadBufferFromPocoSocket.cpp:88` (synchronous Poco read) | `(peer: {}, local: {}, {} ms)` | No |
| `src/Server/ProxyV1Handler.cpp:124` (Proxy v1 wire handler) | `({}, {} ms)` | No |

All four formats include the timeout value formatted as `{} ms`. When the 857-ms handshake fires through the synchronous `ReadBufferFromPocoSocket` path (more likely on slow sanitizer builds — observed mostly on `Integration tests (amd_msan, 1/6)`, `(amd_tsan, 1/6)`, and `(amd_asan_ubsan, db disk, old analyzer, 1/6)`), the assertion fails because the error contains `857 ms` but not `receive timeout 857 ms`.

Example failing CIDB log (`pull_request_number=100399`, `Integration tests (amd_msan, 1/6)`):

```
DB::Exception: All connection tries failed. Log:

Timeout exceeded while reading from socket (socket ([2001:3984:3989::11]:9000), receive timeout 5000 ms)
Code: 209. DB::NetException: Timeout exceeded while reading from socket
  (peer: [2001:3984:3989::11]:9000, local: [2001:3984:3989::13]:35828, 857 ms): ...
```

The only `receive timeout` phrase here is for the unrelated default-5000ms socket-receive timer. The 857-ms handshake-timeout cancellation surfaces through the sync path and prints `857 ms` without the prefix.

### Fix

Relax the assertion from `"receive timeout 857 ms"` to `"857 ms"`. The test still verifies the configured `handshake_timeout_ms=857` actually took effect (the value appears in the error), but no longer locks the assertion to one specific socket-error code path. The companion assertion `"ALL_CONNECTION_TRIES_FAILED" in error` already verifies the high-level failure mode.

### Failure history

Over the last 30 days, ~10 failures across 9 distinct unrelated PRs in the assertion-mismatch mode (this PR's target). Concentrated on slow sanitizer + old-analyzer integration runners.

CI report data: https://play.clickhouse.com/

...

### Version info
- Merged into: `26.5.1.201` (included in `26.5` and later)
- Backported to: `26.3.33.59`

### DIFF
--- a/tests/integration/test_distributed_frozen_replica/test.py
+++ b/tests/integration/test_distributed_frozen_replica/test.py
@@ -106,7 +106,15 @@ def test_cluster_all_replicas_query(
         except QueryRuntimeException as e:
             error = str(e)
     assert "ALL_CONNECTION_TRIES_FAILED" in error
-    assert "receive timeout 857 ms" in error
+    # The configured handshake_timeout_ms=857 may surface through the async
+    # `HedgedConnections` / `AsyncTaskExecutor` path (which formats as
+    # "(..., receive timeout {} ms)") OR through the synchronous
+    # `ReadBufferFromPocoSocket` path (which formats as
+    # "(peer: ..., local: ..., {} ms)" without the "receive timeout" prefix).
+    # Both formats include the timeout value in milliseconds, so check for
+    # "857 ms" alone — this verifies the configured timeout was applied
+    # without locking the test to one specific socket-error code path.
+    assert "857 ms" in error
 
 
 @pytest.mark.parametrize(


groeneai:
**Pre-PR validation gate** (per `TASK.md` Phase 4 Step 9):

**a) Deterministic repro?** — Yes, in the CI evidence form: CIDB has 10 failures across 9 unrelated PRs over the last 30 days where `test.py:109` raises `AssertionError: assert 'receive timeout 857 ms' in '<full error string>'`, with the `<full error string>` containing `857 ms` (sync-path format) but not the `receive timeout` prefix. Sample failure context for `pull_request_number=100399, Integration tests (amd_msan, 1/6)`:

```
DB::Exception: All connection tries failed. Log:

Timeout exceeded while reading from socket
  (peer: [2001:3984:3989::11]:9000, local: [...]:35828, 857 ms): (node1:9000, ...)
```

The 857-ms handshake-timeout cancellation surfaces through `ReadBufferFromPocoSocket.cpp:88` and prints `857 ms` without `receive timeout`. The assertion mismatch is reproducible from those CI runs verbatim.

**b) Root cause explained?** — Yes. The codebase emits the same logical event ("socket read hit configured timeout") through four format-string sites:
- `AsyncTaskExecutor.cpp:124` and `HedgedConnections.cpp:449` print `(..., receive timeout {} ms)`
- `ReadBufferFromPocoSocket.cpp:88` prints `(peer: ..., local: ..., {} ms)` — no `receive timeout`
- `ProxyV1Handler.cpp:124` prints `({}, {} ms)` — no `receive timeout`

When `clusterAllReplicas` retries replicas with `handshake_timeout_ms=857` and the synchronous Poco read path is the one that observes the timeout, the error string lacks the `receive timeout` prefix. All four formats include the value as `{} ms`, so the universal substring is `857 ms`.

**c) Fix matches root cause?** — Yes. Replace `"receive timeout 857 ms"` with `"857 ms"`. This drops the format-specific prefix that fails for two of the four code paths while preserving the timeout-value check.

**d) Test intent preserved?** — Yes. The test still asserts:
- `"ALL_CONNECTION_TRIES_FAILED" in error` (line 108) — connection-failover mode
- `"857 ms" in error` (line 109, post-fix) — the configured handshake timeout was actually applied (still catches a regression where the 857ms gets ignored or replaced with a different value, e.g., a 5000ms default)

The narrower regression class the previous assertion accidentally caught — "the message format is exactly the async-path one" — was never the test's intent; it was a side effect of which path happened to fire on the original developer's machine.

**e) Both directions demonstrated?** — Asymmetric demonstration:
- WITHOUT fix: 10 CIDB-recorded failures in 30 days where `error` contains `857 ms` but the `"receive timeout 857 ms"` assertion fails. Logs above.
- WITH fix: `"857 ms"` is a substring of all four error formats, so the assertion passes whenever the previous one passed and whenever the previous one failed-but-error-still-contained-the-value. The fix loosens the constraint, so it can never make a previously-passing run fail.

The fix is a one-line test-only change with no production C++ touched, no behavior gated, and no risk of false positives in the relaxed assertion (the value `857` is uniquely the configured `handshake_timeout_ms`; no other timeout in this query uses 857ms).

Tracker task: `2026-04-30-ci-testdistributedfrozenreplica-chro` (failure modes A/C/F still need separate work — empty-error race, count-mismatch race, `pause_container` Docker flake).

Session: `cron:clickhouse-ci-task-worker:20260430-094500`

groeneai:
cc @azat — could you review this? It's a one-line test-only change in `test_distributed_frozen_replica` that relaxes a brittle assertion (`receive timeout 857 ms` → `857 ms`) so the test no longer locks itself to one of the four ClickHouse socket-timeout error-message formats. Root cause and CIDB evidence in the PR body and pre-PR validation gate comment above.

alexey-milovidov (approved):

PedroTadim:
Backporting to fix CI in 26.3 and 26.4

PR #103789, opened 2026-04-30
Add regression test for #65630 — WHERE on UNION ALL without GROUP BY

### LINKED ISSUE #65630: Unexpected result of WHERE for subquery with omitted GROUP BY and UNION ALL
ClickHouse v.24.3.4.147

WHERE doesn't filter out values, but returns them with default values (when GROUP BY omitted):

```sql
SELECT *
FROM
(
    SELECT 'number1' AS id, count(number) AS cnt
    FROM numbers(5)
    UNION ALL
    SELECT 'number2', count(number)
    FROM numbers(4)
    UNION ALL
    SELECT 'number2', count(number)
    FROM numbers(3)
)
WHERE id = 'number1';

/*
   ┌─id──────┬─cnt─┐
1. │ number2 │   0 │
   └─────────┴─────┘
   ┌─id──────┬─cnt─┐
2. │ number1 │   5 │
   └─────────┴─────┘
   ┌─id──────┬─cnt─┐
3. │ number2 │   0 │
   └─────────┴─────┘

*/
```

This way works as expected:

```sql
SELECT *
FROM
(
    SELECT 'number1' AS id, count(number) AS cnt
    FROM numbers(5)
    GROUP BY id
    UNION ALL
    SELECT 'number2' AS id, count(number)
    FROM numbers(4)
    GROUP BY id
    UNION ALL
    SELECT 'number2' AS id, count(number)
    FROM numbers(3)
    GROUP BY id
)
WHERE id = 'number1';

/*
   ┌─id──────┬─cnt─┐
1. │ number1 │   5 │
   └─────────┴─────┘
*/
```

groeneai:
## Summary

Adds a stateless regression test for [#65630](https://github.com/ClickHouse/ClickHouse/issues/65630) — WHERE filter on a UNION ALL subquery without explicit GROUP BY. The bug was reported on 24.3.4.147 and is already fixed on master (and from 24.8.1 onward per [@den-crane's verification](https://fiddle.clickhouse.com/c2941eb5-a2be-4343-8b8b-20fa28c60a69)). This PR locks the correct behaviour in so any future change to predicate-pushdown rules around UNION ALL cannot silently regress it.

## Background

Originally:

```sql
SELECT *
FROM (
    SELECT 'number1' AS id, count(number) AS cnt FROM numbers(5)
    UNION ALL
    SELECT 'number2', count(number) FROM numbers(4)
    UNION ALL
    SELECT 'number3', count(number) FROM numbers(3)
)
WHERE id = 'number1';
```

returned three rows — `number1, 5`, `number2, 0`, `number3, 0` — instead of the single matching row. The cause: the planner pushed the WHERE predicate down into each UNION ALL branch. Inside a non-matching branch the predicate evaluates against the branch's constant column alias and becomes false, which filters out the source rows; the implicit (no GROUP BY) aggregation then still emits a single row with `count = 0` per non-matching branch.

The fix keeps the WHERE on top of the UNION so non-matching branches are dropped entirely.

## Test cases

`tests/queries/0_stateless/04144_filter_union_all_no_group_by.sql` covers:

1. The original report — outer WHERE on UNION ALL → must return only the matching branch.
2. Wrapping in `LIMIT 100000000` (originally suggested as a workaround because LIMIT blocks predicate pushdown) — must still return the same single row.
3. Filtering on the aggregate column (`WHERE cnt = 3`) — never affected by the pushdown bug, but included so any future change cannot silently regress it.
4. A predicate that excludes every branch — must produce an empty result.

Verified locally on master HEAD (`651bde46a3ab`) — passes 30/30 with default randomized settings.

Closes #65630.

...

### Version info
- Merged into: `26.5.1.231`

### DIFF
--- a/tests/queries/0_stateless/04144_filter_union_all_no_group_by.reference
+++ b/tests/queries/0_stateless/04144_filter_union_all_no_group_by.reference
@@ -1,0 +1,6 @@
+number1	5
+--- with LIMIT (pushdown blocker) ---
+number1	5
+--- filter on aggregate value ---
+number3	3
+--- no match returns empty ---
--- a/tests/queries/0_stateless/04144_filter_union_all_no_group_by.sql
+++ b/tests/queries/0_stateless/04144_filter_union_all_no_group_by.sql
@@ -1,0 +1,70 @@
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/65630
+--
+-- A WHERE filter applied on top of a UNION ALL whose branches have a constant
+-- column alias and no explicit GROUP BY used to return one row per non-matching
+-- branch with the aggregate evaluated over an empty set (count = 0). The cause
+-- was the planner pushing the WHERE predicate down into each UNION ALL branch:
+-- inside a branch the predicate evaluates against the branch's constant alias
+-- and becomes false, which filters out the source rows; the implicit (no
+-- GROUP BY) aggregation then still emits a single row with count = 0 per
+-- non-matching branch. The expected behaviour is that WHERE is applied on top
+-- of the UNION, so non-matching branches are dropped entirely.
+
+-- Direct outer WHERE on UNION ALL: should return only the matching branch.
+SELECT * FROM
+(
+    SELECT 'number1' AS id, count(number) AS cnt FROM numbers(5)
+    UNION ALL
+    SELECT 'number2', count(number) FROM numbers(4)
+    UNION ALL
+    SELECT 'number3', count(number) FROM numbers(3)
+)
+WHERE id = 'number1';
+
+SELECT '--- with LIMIT (pushdown blocker) ---';
+
+-- Wrapping in a LIMIT was originally suggested as a workaround because LIMIT
+-- blocks predicate pushdown. It must keep producing the same single row.
+SELECT * FROM
+(
+    SELECT * FROM
+    (
+        SELECT 'number1' AS id, count(number) AS cnt FROM numbers(5)
+        UNION ALL
+        SELECT 'number2', count(number) FROM numbers(4)
+        UNION ALL
+        SELECT 'number3', count(number) FROM numbers(3)
+    )
+    LIMIT 100000000
+)
+WHERE id = 'number1';
+
+SELECT '--- filter on aggregate value ---';
+
+-- Filtering by the aggregate column (post-aggregation) was never affected by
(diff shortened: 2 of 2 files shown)


groeneai:
**Pre-PR validation gate** (per [TASK.md Phase 4 Step 9](https://github.com/groeneai/ClickHouse))

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | The original 24.3.4.147 buggy result is documented in #65630 (returned 3 rows instead of 1). On master HEAD `651bde46a3ab` the new test deterministically passes (30/30 runs with full randomization). |
| b | **Root cause explained?** | The pre-fix planner pushed the WHERE predicate down into each UNION ALL branch. Inside a non-matching branch, the predicate `id = 'number1'` evaluates against the constant alias (`'number2'`/`'number3'`) and becomes false; the source rows are filtered out, but the implicit (no GROUP BY) aggregation still emits one row with `count = 0`. The fix keeps the WHERE on top of the UNION so non-matching branches are dropped entirely. |
| c | **Fix matches root cause?** | N/A — fix already merged upstream (24.8.1+). This PR only adds regression coverage. |
| d | **Test intent preserved? / New tests added?** | New stateless test `04144_filter_union_all_no_group_by`. Four scenarios cover: (1) the original report; (2) wrapping in `LIMIT 100000000` (originally a workaround because LIMIT blocks pushdown) — keeps the result intact; (3) post-aggregate filter (`WHERE cnt = 3`) — never affected by the bug, included to lock down the symmetric path; (4) no-match predicate — must produce an empty result. |
| e | **Demonstrated in both directions?** | Pre-fix behaviour (3 rows with zeroed aggregates) is documented in #65630 and #65630#issuecomment-2186871787. Post-fix behaviour: test passes 30/30 with randomized settings on master `651bde46a3ab`. |
| f | **Fix is general, not a narrow patch?** | N/A — no source code is modified. The test exercises four predicate-pushdown shapes around UNION ALL to widen regression coverage. |

Session: `cron:clickhouse-ci-task-worker:20260430-144500`.

groeneai:
cc @KochetovNicolai — could you review? Test-only PR adding regression coverage for #65630 (WHERE filter on UNION ALL without GROUP BY, fixed in 24.8.1+). Component: `comp-query-analyzer` / `comp-query-optimizer` (predicate pushdown around UNION ALL). Single new stateless test, 30/30 passing on master HEAD with default randomization.

PR #103803, opened 2026-04-30
Fix duplicate TOTALS row in play.html UI

alexey-milovidov:
In single-query mode, the play.html UI sometimes rendered the TOTALS row twice at the bottom of a result table.

When `postImpl` was extracted into its own function, the original rendering block was left in place inside `postSingle`. As a result, `renderTotals` was being called twice on success — once from `postImpl` after parsing the response, and again from `postSingle` after `postImpl` returned — appending a second `<tfoot>` to the table. The same duplication affected `renderGraph`, `renderExtremes`, and `transposeIfNeeded`, but TOTALS was the most user-visible.

This change drops the duplicated block from `postSingle`, keeping only `_changeTableLayout`, which is the one step `postImpl` does not perform. `postMulti` already relied solely on `postImpl`'s rendering and was unaffected.

- Not for changelog

Fix `TOTALS` row being rendered twice at the bottom of the result table in the `play.html` web UI.

### Version info
- Merged into: `26.5.1.230`
- Backported to: `26.4.1.1131`


alexey-milovidov:
@groeneai, fix `test_dictionaries_all_layouts_separate_sources` and send a separate PR with the fix.

PR #103810, opened 2026-04-30
CI: replace `_install_light`/`start_light` with `ClickHouseService` in CI jobs

leshikus:
Refactor CI job scripts to consistently use the `ClickHouseService` context manager instead of the removed `_install_light` / `start_light` methods on `ClickHouseProc`, and harden failure reporting so an aborted run cannot be reported as a false green.

- Move install config helpers (`install_clickbench_config`, `install_vector_search_config`) from `ClickHouseProc` to their respective job files.
- Remove dead methods from `ClickHouseProc`: `_install_light`, `start_light`, `install_fuzzer_config`, `install_vector_search_config`.
- Delete dead `fuzzers_job.py` (the real fuzzer job runs in Docker via `ast_fuzzer_job.py`).
- Add `results: list` parameter to `ClickHouseService` so it self-records the `Start ClickHouse` result using class constant `RESULT_NAME`.
- Add `Result.raise_if_failed` helper in `ci/praktika/result.py`; jobs use it with the walrus operator to replace the `res = True` / `if res:` boolean-accumulator pattern.
- Wrap the entire `ClickHouseService.__enter__` body in try/except so failures in the early setup (binary download, symlink creation, config-file copy, data-directory prep) also emit a `Start ClickHouse: FAIL` subresult before re-raising. Previously, only the `Popen` + `_wait_ready` portion was covered.
- In `clickbench.py`, `vector_search_stress_job.py`, and `vector_search_stress_tests.py`, replace `except Exception: pass` around the run with a handler that prints the traceback and appends `Result(name="Job error", status=FAIL, info=str(e))`, so `Result.create_from(...).complete_job()` correctly marks the job failed when the run aborts before recording its own sub-results.
- Move the `storage_conf_local.xml` skip into `ClickHouseService.__enter__` via a new `config_ignore_patterns` constructor argument (ClickBench passes it). The previous `unlink` in `install_clickbench_config` ran before the service copied `programs/server/config.d`, so the file ended up present after the copy and the cache-path override never took effect.
- Rename `clickbench_master_jobs` to `clickbench_jobs` and wire it into `pull_request.py` so the fix is exercised by PR CI. The job's `digest_config` matches only clickbench-specific files, but its `requires=[CH_*_RELEASE]` artifact will be marked affected whenever the release build runs (i.e. on any PR that touches `./src`, `./programs`, `./contrib`, etc.). On PRs that touch only docs or unrelated CI scripts, ClickBench is dropped — same propagation rule as the existing `performance_comparison_with_master_head_jobs`.

### Version info
- Merged into: `26.7.1.1000` (included in `26.7` and later)


maxknv (approved):

leshikus:
I will continue when the time permits, the last problem was: clickbench run had errors

ci/defs/job_configs.py
```diff
@@ -1315,15 +1315,18 @@ class JobConfigs:
             for batch in range(1, total_batches + 1)
         ]
     )
-    clickbench_master_jobs = Job.Config(
+    clickbench_jobs = Job.Config(
         name=JobNames.CLICKBENCH,
         runs_on=RunnerLabels.FUNC_TESTER_AMD,
         command="python3 ./ci/jobs/clickbench.py",
         digest_config=Job.CacheDigestConfig(
             include_paths=[
```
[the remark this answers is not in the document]
leshikus: I would rather not expand this digest that broadly. Across all jobs, `include_paths` tracks a job's runtime dependencies — the scripts it imports/executes — and none of them list `ci/defs/job_configs.py` or the `ci/workflows/*.py` files; the only `ci/defs/` entry anywhere is the `darwin.skip` data file. Adding `job_configs.py` (which defines every job) to one job's digest would invalidate ClickBench's cache on every unrelated job edit, and the same argument applies identically to all jobs — so this is really a global change-filter design question rather than something specific to ClickBench. The concrete runtime-import gap flagged earlier (`clickhouse_proc.py`) is the right kind of entry and is fixed in 205fdec8e8ac3ec4c1e973396ef0a3b84b17854e; the CI-plumbing self-reference is intentionally left out here.

ci/praktika/result.py
```diff
@@ -240,6 +240,10 @@ def is_ok(self):
             Result.Status.XFAIL,
         )
 
+    def raise_if_failed(self):
```
maxknv: I tried to use this pattern some time ago, but decided it doesn't work very well. I'd prefer to explicitly write 'raise' on the caller side. Visually, it looks clearer, and the IDE also highlights it.

ci/jobs/scripts/clickhouse_service.py
```diff
@@ -7,18 +7,23 @@
...
 from pathlib import Path
 
+from ci.praktika.result import Result
 from ci.praktika.utils import Shell, Utils
 
 repo_dir = Utils.cwd()
 temp_dir = f"{repo_dir}/ci/tmp"
 
 
 class ClickHouseService:
```
maxknv, line 18: can we remove code deprecated by this class?

ci/jobs/clickbench.py
```diff
@@ -1,115 +1,116 @@
...
+from ci.jobs.scripts.clickhouse_service import ClickHouseService
 from ci.praktika.info import Info
 from ci.praktika.result import Result
 from ci.praktika.utils import Shell, Utils
 
-temp_dir = f"{Utils.cwd()}/ci/tmp/"
+temp_dir = f"{Utils.cwd()}/ci/tmp"
+
+
+def install_clickbench_config():
```
maxknv: nitpick: it would be nice if ClickHouseService could support callbacks for these tune-ups. Say we just pass the name of the function/hook when calling context manager and ClickHouseService does the rest under the hood

maxknv:
LGTM with minor comments. We also need to make sure the affected jobs work properly, including logexport featue in clickbench

ci/praktika/result.py
leshikus: Done — removed `raise_if_failed` and replaced each call site with an explicit `if not r.is_ok(): raise RuntimeError(...)`.

ci/jobs/scripts/clickhouse_service.py
leshikus: Done — removed the now-dead light-install path from `ClickHouseProc` that this class replaced: `_install_light`, `start_light`, `install_clickbench_config`, `install_fuzzer_config`, and `install_vector_search_config`.

ci/jobs/clickbench.py
leshikus: Done — `ClickHouseService` now takes a `config_hooks` list of callables it runs after copying the base config and before starting the server. ClickBench and the vector-search jobs pass their tune-ups (e.g. `install_clickbench_config`) as hooks under the hood instead of a separate install step.

ci/jobs/scripts/clickhouse_service.py
```diff
@@ -27,71 +33,109 @@ def __init__(
...
+
+            # Create symlinks if absent
+            for link_name in ("clickhouse-server", "clickhouse-client", "clickhouse-local"):
+                link_path = Path(temp_dir) / link_name
+                if not link_path.exists():
+                    Utils.link(clickhouse_bin, link_path)
+
+            # Copy server config files if absent
+            config_dir = Path(self.ch_config_dir)
+            if not (config_dir / "config.xml").exists():
```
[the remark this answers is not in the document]
leshikus: Fixed in https://github.com/ClickHouse/ClickHouse/commit/ccd3da9d7addeedc53df6dcc95ca49def3e80906

ci/jobs/scripts/clickhouse_service.py
```diff
@@ -27,71 +31,99 @@ def __init__(
...
 
+    def __enter__(self):
         try:
+            Utils.add_to_PATH(temp_dir)
+
+            # Download binary if absent; CI release artifacts extract the
+            # binary without the executable bit, so chmod afterwards.
+            clickhouse_bin = Path(temp_dir) / "clickhouse"
+            if not clickhouse_bin.exists():
+                self._download_binary()
```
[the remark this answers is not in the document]
leshikus: this is the way things worked before, for example, for fast test; maybe this is not ideal, but let's do not change it now

leshikus:
```
[15:43:06] Start log export
[15:43:06] Run command: [./ci/jobs/scripts/functional_tests/setup_log_cluster.sh --setup-logs-replication]
[15:43:07] Received exception from server (version 26.7.1):
[15:43:07] Code: 192. DB::Exception: Received from localhost:9000.
           DB::Exception: There is no user `CLICKHOUSE_CI_LOGS_USER_logs_sender` in `users`
(The user value ci is redacted to the token CLICKHOUSE_CI_LOGS_USER by setup_log_cluster.sh:38, so the real message is There is no user 'ci_logs_sender'. It repeats for every system.*_log table.)
```
log export does not work for now

leshikus:
It is broken on master as well, but at least authentication works there `Code: 701. DB::Exception: Received from localhost:9000. DB::Exception: Requested cluster 'system_logs_export' not found. ` https://github.com/ClickHouse/ClickHouse/actions/runs/29229353556/job/86765371667

ci/defs/job_configs.py
```diff
@@ -1318,20 +1318,23 @@ class JobConfigs:
...
-                # ClickBench starts the server via `ClickHouseProc.start_light`,
-                # which now clears leftover processes through `server_cleanup.py`.
-                # Track both so changes to the shared start path reschedule the job.
+                # ClickBench starts the server via `ClickHouseService`, which
+                # clears leftover processes through `server_cleanup.py`. Track
+                # both so changes to the shared start path reschedule the job.
                 "./ci/jobs/scripts/clickhouse_proc.py",
                 "./ci/jobs/scripts/server_cleanup.py",
                 "./ci/jobs/scripts/clickbench/",
+                "./ci/jobs/scripts/clickhouse_service.py",
```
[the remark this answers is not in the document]
leshikus: Good point, but this is not a regression introduced by this PR. Let's address it after one of these files actually changes.

PR #103811, opened 2026-04-30
Disable omitting frame pointers for simpler unwind

azat:
- Performance Improvement

Disable omitting frame pointers for simpler unwind

Although it is not a problem in ClickHouse, since we use libunwind that handles this well, it is a problem for perf, or if we will want to use eBPF to do some performance analysis (to avoid overhead in the clickhouse-server process), though there are also other options [1].

  [1]: https://github.com/parca-dev/parca-agent/pull/948

Refs: https://www.brendangregg.com/blog/2024-03-17/the-return-of-the-frame-pointers.html


EmeraldShift:
Enabling frame pointers is a nice move for easy debuggability with `perf` and existing continuous profiling infrastructure! Some notes:

1. Maybe it should be optional? I am personally happy to pay a 1-5% performance cost to ensure I can easily and with low overhead run `perf` to find out what's costing CPU. But maybe other customers will be fine with the libunwind-based sampling profiler and rather not pay the frame pointer tax.
2. That said, you might not have to pay the frame pointer tax forever! There is an upcoming technology [SFrame](https://sourceware.org/binutils/wiki/sframe) that compiles a more efficient set of debug info into the program, and Linux and `perf` are in the process of supporting it. Maybe it will be ready for use, and can replace frame pointers, in a year?

azat:
> Enabling frame pointers is a nice move for easy debuggability with perf and existing continuous profiling infrastructure!

Yes, this is the reason, though there are other options to overcome this:
- for instance internally ClickHouse uses libunwind which get this info from DWARF and it works pretty good, though it can be slower
- https://github.com/parca-dev/parca-agent/pull/948 - simple DWARf parser in eBPF

> Maybe it should be optional?

It was optional already

> I am personally happy to pay a 1-5% performance cost to ensure I can easily and with low overhead run perf to find out what's costing CPU. But maybe other customers will be fine with the libunwind-based sampling profiler and rather not pay the frame pointer tax.

This is just an experiment, if the performance cost will be visible it will not be applied

> That said, you might not have to pay the frame pointer tax forever! There is an upcoming technology [SFrame](https://sourceware.org/binutils/wiki/sframe) that compiles a more efficient set of debug info into the program, and Linux and perf are in the process of supporting it. Maybe it will be ready for use, and can replace frame pointers, in a year?

Yes, this is not that new format AFAIR, and does it support is a good idea right now is an open question.

azat:
It is slower

PR #103816, opened 2026-05-01
Add support of resolved symbols (Array(String)) for flameGraph()

azat:
- Improvement

Add support of resolved symbols (Array(String)) for flameGraph()

Nowadays `trace_log` has `symbols` and `lines` columns

### Version info
- Merged into: `26.5.1.760`

### DIFF
--- a/src/AggregateFunctions/AggregateFunctionFlameGraph.cpp
+++ b/src/AggregateFunctions/AggregateFunctionFlameGraph.cpp
@@ -13,6 +13,7 @@
 #include <IO/Operators.h>
 #include <Common/UnorderedMapWithMemoryTracking.h>
 #include <Common/VectorWithMemoryTracking.h>
+#include <Common/typeid_cast.h>
 
 namespace DB
 {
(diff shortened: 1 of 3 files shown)


azat:
It gives identical result - https://pastila.nl/?0035cdc0/60160cde58e189f4e8958aa6358b0290#lTUGeV3tN26BOEHt1CZhEg==GCM

### `symbols`
fg-symbols

### `trace`
fg-trace

src/AggregateFunctions/AggregateFunctionFlameGraph.cpp
```diff
@@ -277,6 +286,31 @@ struct AggregateFunctionFlameGraphData
...
+    UnorderedMapWithMemoryTracking<std::string_view, UInt64> string_pool;
+    VectorWithMemoryTracking<std::string_view> id_to_string;
+
+    UInt64 internString(std::string_view s, Arena * arena)
+    {
+        if (auto it = string_pool.find(s); it != string_pool.end())
+            return it->second;
+
+        char * dest = arena->alloc(s.size());
+        if (!s.empty())
```
Diskein, line 304: it seems to me that we on purpose allows to call `alloc` on a string of empty size, we could just reserve some special value for them (like max uint64) and won't waste space on it in container and hashmap, but it would make solution more complex and i doubt it worth it.

Diskein (approved):

PR #103823, opened 2026-05-01
Fix toString(DateTime, tz) throwing when tz is Nullable(String) with …

### LINKED ISSUE #103712: `toString(DateTime, Nullable(String))` still throws `NOT_IMPLEMENTED` when the timezone column contains an actual NULL value — PR's fix is incomplete

_Retrospective finding from a historical scan of [PR #59190](https://github.com/ClickHouse/ClickHouse/pull/59190) (merged 2024-01-26). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

Calling `toString(toDateTime('...'), tz)` where `tz` is a `Nullable(String)` column holding NULL produces `Code: 48. NOT_IMPLEMENTED: Method getDataAt is not supported for Nullable(String) in case if value is NULL` instead of returning NULL or a clean user-facing error.

**Root cause:** FunctionsConversion.h:2308 — the `if (!arguments[1].column.get()->getDataAt(i).empty())` guard does not skip rows for which `null_map` was set by lines 2300-2301. The loop should check `null_map->getData()[i]` (or the column's `isNullAt(i)`) and `continue` for NULL rows; otherwise `ColumnNullable::getDataAt` throws.

**Why we believe this is a bug:** Entry: `ConvertImpl<FromDataType, DataTypeString, Name, ConvertDefaultBehaviorTag>::execute` (FunctionsConversion.h, the `IsDataTypeDateOrDateTimeOrTime` → `DataTypeString` specialization). The PR added `copyNullMap(arguments[1].column->convertToFullColumnIfConst())` at line 2300-2301 so the result is wrapped as `Nullable(String)`. Inside the per-row loop at line 2306-2312, `arguments[1].column.get()->getDataAt(i)` is invoked unconditionally on the Nullable column. `ColumnNullable::getDataAt` (src/Columns/ColumnNullable.cpp:44-49) throws `NOT_IMPLEMENTED` when `isNullAt(n)` is true, so any NULL row aborts the whole batch.

**Affected locations:**
- `src/Functions/FunctionsConversion.h:2308` — Loop body uses `arguments[1].column.get()->getDataAt(i)` without checking the null_map populated at lines 2300-2301.
- `src/Functions/FunctionsConversion.h:2300` — PR-added line that copies null_map from the Nullable timezone column.
- `src/Columns/ColumnNullable.cpp:44` — `ColumnNullable::getDataAt` throws `NOT_IMPLEMENTED` when value is NULL — the underlying assertion failure point.

**Impact:** Any query that selects `toString(dt, tz)` from a table where `tz Nullable(String)` may hold NULL aborts with a leaky internal error message. This is a user-visible regression in the same area the PR claims to fix; the PR title says 'Fix crash with nullable timezone' but only the non-NULL value case is actually fixed.

Yes — confirmed on current `master` (commit `880f38dc01c8`).

### How to reproduce

```sql
SET session_timezone='Europe/Amsterdam';

-- Verify the case the PR fixed: non-NULL value in Nullable(String) timezone works.
SELECT toString(toDateTime('2022-01-01 12:13:14'), CAST('UTC', 'Nullable(String)'));

-- BUG: actual NULL value in Nullable(String) timezone column should propagate to NULL,
-- but currently throws NOT_IMPLEMENTED with a leaky internal message.
DROP TABLE IF EXISTS t_pr59190_tz;
CREATE TABLE t_pr59190_tz (tz Nullable(String)) ENGINE=Memory;
INSERT INTO t_pr59190_tz VALUES ('UTC'), (NULL);
SELECT toString(toDateTime('2022-01-01 12:13:14'), tz) FROM t_pr59190_tz ORDER BY tz NULLS LAST;
DROP TABLE t_pr59190_tz;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/49d0ee67-02fc-48ac-bc6c-d8a1d6ee7226)

### Expected behavior

```
2022-01-01 11:13:14
2022-01-01 11:13:14
\N
```

```
2022-01-01 11:13:14
Code: 48. DB::Exception: Received from 127.0.0.1:19010. DB::Exception: Method getDataAt is not supported for Nullable(String) in case if value is NULL: while executing 'FUNCTION toString(_CAST(1641035594_DateTime, 'DateTime'_String) :: 1, __table1.tz :: 0) -> toString(_CAST(1641035594_DateTime, 'DateTime'_String), __table1.tz) Nu
```

### Additional context

**Open risks:**
- Same pattern likely exists in the symmetric `FunctionsConversion.h` paths around lines 2385 and 2432 that also call `copyNullMap(arguments[0].column)` — those paths read the timezone via `arguments[1]` similarly and were not investigated.
- Constant-folded path `toString(now(), CAST(NULL, 'Nullable(String)'))` errors with the same message at analysis scope — fixing the loop will not address that path, which is in `extractTimeZoneFromFunctionArguments` / type resolution.

**Suggested fix:** At the top of the per-row loop (line 2306), check the null_map: `if (null_map && (*null_map).getData()[i]) { offsets_to[i] = write_buffer.count(); continue; }` — or equivalently call `arguments[1].column->isNullAt(i)` and skip the `getDataAt` call, leaving the null_map bit set so the row is reported as NULL in the result.

Found during automated review of [PR #59190](https://github.com/ClickHouse/ClickHouse/pull/59190).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr59190_001`_

### Version info
- Resolved by: #105587
- Merged into: `26.6.1.184`

nickmopen:
Fix toString(DateTime, tz) throwing when tz is Nullable(String) with NULL
Resolves: https://github.com/ClickHouse/ClickHouse/issues/103712

Fixed `toString(dt, tz)` throwing an internal exception when `tz` is a `Nullable(String)` column containing NULL. NULL timezone now correctly propagates to NULL in the result.

When `toString(dt, tz)` is called with a `Nullable(String)` timezone argument, NULL values in `tz` now propagate to NULL in the result instead of throwing an internal exception.

Example:

```sql
CREATE TABLE t (tz Nullable(String)) ENGINE = Memory;
INSERT INTO t VALUES ('UTC'), (NULL);
SELECT toString(toDateTime('2022-01-01 12:13:14'), tz) FROM t;
-- Result:
-- 2022-01-01 12:13:14
-- \N


nickmopen:
@clickgapai , @yariks5s Can you review this PR when you have some time?

yariks5s (approved):

yariks5s:
@nickmopen, unfortunately, we cannot normally merge a pull request with failing tests, even if they look unrelated. Could you please update your merge with master to pick up the latest fixes and re-trigger the CI? :)

PR #103835, opened 2026-05-01
Fix wrong results in numbers() and generate_series() with relaxed IN/NOT IN

### LINKED ISSUE #103660: `extractPlainRanges` ignores `element.relaxed` for `FUNCTION_NOT_IN_SET`, producing wrong results for `numbers()`

### Describe what's wrong

`SELECT number FROM numbers(5) WHERE tuple(number, number) NOT IN (tuple(1, 2)) ORDER BY number` returns `{0, 2, 3, 4}` instead of `{0, 1, 2, 3, 4}`. Number 1 is incorrectly excluded despite `tuple(1, 1) NOT IN (tuple(1, 2))` being true.

**Root cause:** KeyCondition.cpp:4258-4295: `extractPlainRanges` handles `FUNCTION_NOT_IN_SET` by computing the complement of `set_index->getOrderedSet()` values, but does not check `element.relaxed`. When `element.relaxed = true` (due to `MergeTreeSetIndex` deduplicating multiple tuple elements mapping to the same key column), the deduped set `{1}` is a SUPERSET of the actual matching values (which is empty — no `n` satisfies `(n,n) = (1,2)`). The complement of this superset is a SUBSET of the correct complement, incorrectly excluding number 1.

**Why we believe this is a bug:** `ReadFromSystemNumbersStep` (line 453) → `NumbersLikeUtils::extractRanges` → `KeyCondition::extractPlainRanges` (line 4258) → processes `FUNCTION_NOT_IN_SET` with deduped set `{1}` (from `tuple(number, number)` mapping both elements to the same key column) → computes complement `(-inf, 1) ∪ (1, +inf)` → marked as `ExactRanges` → `NumbersRangedSource` never generates number 1

**Affected locations:**
- `src/Storages/MergeTree/KeyCondition.cpp:4258` — extractPlainRanges FUNCTION_NOT_IN_SET handler missing element.relaxed check

**Impact:** Wrong query results when `numbers()`, `generate_series()`, or `system.primes` is used with `NOT IN` or `NOT has` predicates where the set tuple has duplicate key column mappings. Values are silently dropped from the result set.

Yes — confirmed on current `master` (commit `d3573bde5e6b`).

### How to reproduce

```sql
-- Test: extractPlainRanges should not trust deduped set as exact for NOT IN/NOT has
-- Covers: KeyCondition.cpp extractPlainRanges FUNCTION_NOT_IN_SET with relaxed element

-- tuple(number, number) NOT IN (tuple(1, 2)) is true for ALL numbers because
-- (n, n) can never equal (1, 2). But extractPlainRanges deduplicates the set to {1}
-- and computes complement (-inf,1)∪(1,+inf), incorrectly excluding number 1.

-- NOT IN variant
SELECT number FROM numbers(5) WHERE tuple(number, number) NOT IN (tuple(1, 2)) ORDER BY number;

-- NOT has variant
SELECT number FROM numbers(5) WHERE NOT has([tuple(1, 2)], tuple(number, number)) ORDER BY number;

-- Verify individual tuple evaluation is correct (all TRUE)
SELECT number, tuple(number, number) NOT IN (tuple(1, 2)) AS val FROM numbers(5) ORDER BY number;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/6d09be1f-3917-4617-b32f-fbd323d3224b)

### Expected behavior

```
0
1
2
3
4
0
1
2
3
4
0	1
1	1
2	1
3	1
4	1
```

```
0
2
3
4
0
2
3
4
0	1
1	1
2	1
3	1
4	1
```

### Additional context

**Open risks:**
- The `FUNCTION_IN_SET` case (line 4233) has a similar issue — when relaxed, ExactRanges may be a superset of true matches. While the filter catches false positives, LIMIT pushdown with ExactRanges could theoretically produce fewer results than expected.

**Suggested fix:** Add `if (element.relaxed) return false;` at line 4259 (and similarly at line 4234 for `FUNCTION_IN_SET`) so that `extractPlainRanges` falls back to `extractBounds` which returns conservative ranges.

Found during automated review of [PR #96241](https://github.com/ClickHouse/ClickHouse/pull/96241).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr96241_001`_

### Version info
- Resolved by: #103835
- Merged into: `26.5.1.275`

groeneai:
Fixes #103660.

`KeyCondition::extractPlainRanges` builds primary-key ranges that are consumed as exact by `numbers`, `generate_series`, and `system.primes` table functions, and by `LIMIT` pushdown. When the underlying set is *relaxed* — deduplicated by `MergeTreeSetIndex` because multiple tuple elements map to the same key column, or transformed via partition functions — the constructed ranges are no longer a faithful representation of the original predicate, and treating them as exact yields wrong results.

Example from the issue:

```sql
SELECT number FROM numbers(5) WHERE tuple(number, number) NOT IN (tuple(1, 2)) ORDER BY number;
-- Returned: {0, 2, 3, 4}
-- Expected: {0, 1, 2, 3, 4} (no row has both `number = 1` AND `number = 2`)
```

`MergeTreeSetIndex` deduplicates the tuple columns to a single key column, so the set `{(1, 2)}` becomes `{1}` and is marked as relaxed. The `FUNCTION_NOT_IN_SET` branch then builds the complement `(-inf, 1) U (1, +inf)`, and `NumbersRangedSource` skips `number = 1` from the output.

The fix bails out of `extractPlainRanges` for both `FUNCTION_IN_SET` and `FUNCTION_NOT_IN_SET` when `element.relaxed` is true; the caller (`extractBounds`) ignores the conjunct and falls back to the universe range, so the row-level filter then determines membership correctly. This mirrors the existing relaxed-aware handling already present in `KeyCondition::checkInHyperrectangle`.

Fix wrong results from `numbers`, `generate_series`, and similar range-honoring sources when the `WHERE` clause uses `IN` or `NOT IN` over a tuple whose elements deduplicate to a single key column (for example, `WHERE tuple(number, number) NOT IN (tuple(1, 2))`). Closes [#103660](https://github.com/ClickHouse/ClickHouse/issues/103660).

### Version info
- Merged into: `26.5.1.275`

### DIFF
--- a/src/Storages/MergeTree/KeyCondition.cpp
+++ b/src/Storages/MergeTree/KeyCondition.cpp
@@ -4393,6 +4393,14 @@ bool KeyCondition::extractPlainRanges(Ranges & ranges) const
                 if (element.set_index->hasMonotonicFunctionsChain())
                     return false;
 
+                /// A relaxed set is not exact (e.g. `tuple(i, i) IN (tuple(1, 2))` deduplicates to `i IN (1)`,
+                /// which is a superset of the true matches). For ordinary `MergeTree` reads this is harmless
+                /// because the row-level filter rejects false positives, but consumers that treat extracted
+                /// ranges as exact (e.g. `numbers()`/`generate_series()` sources, `LIMIT` pushdown) would
+                /// produce wrong results. Bail out and let the caller fall back to a conservative bound.
+                if (element.relaxed)
+                    return false;
+
                 if (element.set_index->size() == 0)
                 {
                     rpn_stack.push(PlainRanges::makeBlank()); /// skip blank range
@@ -4416,6 +4424,15 @@ bool KeyCondition::extractPlainRanges(Ranges & ranges) const
             else if (element.function == RPNElement::FUNCTION_NOT_IN_SET)
             {
                 if (element.set_index->hasMonotonicFunctionsChain())
+                    return false;
+
+                /// A relaxed set means the deduped/transformed set is a strict subset of the original.
+                /// For `NOT IN`, the complement built below is therefore stricter than the true predicate
+                /// and would incorrectly exclude rows that should match. Example:
+                /// `tuple(i, i) NOT IN (tuple(1, 2))` deduplicates to `i NOT IN (1)`, building ranges
+                /// `(-inf, 1) ∪ (1, +inf)` and skipping `i = 1`, even though `tuple(1, 1) != tuple(1, 2)`.
+                /// Bail out and let the caller fall back to a conservative bound.
+                if (element.relaxed)
                     return false;
 
                 if (element.set_index->size() == 0)
--- a/src/Storages/MergeTree/KeyCondition.cpp
+++ b/src/Storages/MergeTree/KeyCondition.cpp
@@ -4336,6 +4336,24 @@ bool KeyCondition::extractPlainRanges(Ranges & ranges) const
 
     for (const auto & element : rpn)
     {
+        /// `extractPlainRanges` produces a range set that consumers (e.g. `ReadFromSystemNumbersStep`,
+        /// `ReadFromSystemPrimesStep`, `LIMIT` pushdown) treat as the exact set of matching key values.
+        /// A relaxed atom by definition does not yield exact ranges: depending on the source it may be
+        /// a strict superset of true matches (e.g. `LIKE` without a perfect prefix, monotonic-constant
+        /// wrapping) or, after complement, a strict subset that incorrectly skips matching rows
+        /// (e.g. `tuple(i, i) NOT IN (tuple(1, 2))` deduplicates the set to `i NOT IN (1)` and builds
+        /// `(-inf, 1) U (1, +inf)`, skipping `i = 1` even though `tuple(1, 1) != tuple(1, 2)`).
+        ///
+        /// Bail out for any relaxed atom and let the caller fall back to a conservative bound. This
+        /// guard is positioned at the top of the loop intentionally, so it covers all current atom
+        /// kinds (`FUNCTION_IN_SET`, `FUNCTION_NOT_IN_SET`, `FUNCTION_IN_RANGE`, `FUNCTION_NOT_IN_RANGE`,
+        /// `FUNCTION_IS_NULL`, `FUNCTION_IS_NOT_NULL`, ...) and any future atom that introduces
+        /// relaxation handling. Operator elements (`FUNCTION_AND`, `FUNCTION_OR`, `FUNCTION_NOT`,
+        /// `ALWAYS_TRUE`, `ALWAYS_FALSE`) are never set as relaxed; relaxation propagates through them
+        /// via their child atoms (see the comment on `KeyCondition::relaxed` in `KeyCondition.h`).
+        if (element.relaxed)
+            return false;
+
         if (element.function == RPNElement::FUNCTION_AND)
         {
             auto right_ranges = rpn_stack.top();
@@ -4393,14 +4411,6 @@ bool KeyCondition::extractPlainRanges(Ranges & ranges) const
                 if (element.set_index->hasMonotonicFunctionsChain())
                     return false;
 
-                /// A relaxed set is not exact (e.g. `tuple(i, i) IN (tuple(1, 2))` deduplicates to `i IN (1)`,
-                /// which is a superset of the true matches). For ordinary `MergeTree` reads this is harmless
-                /// because the row-level filter rejects false positives, but consumers that treat extracted
-                /// ranges as exact (e.g. `numbers()`/`generate_series()` sources, `LIMIT` pushdown) would
-                /// produce wrong results. Bail out and let the caller fall back to a conservative bound.
-                if (element.relaxed)
-                    return false;
-
                 if (element.set_index->size() == 0)
                 {
                     rpn_stack.push(PlainRanges::makeBlank()); /// skip blank range
@@ -4424,15 +4434,6 @@ bool KeyCondition::extractPlainRanges(Ranges & ranges) const
             else if (element.function == RPNElement::FUNCTION_NOT_IN_SET)
             {
                 if (element.set_index->hasMonotonicFunctionsChain())
-                    return false;
-
-                /// A relaxed set means the deduped/transformed set is a strict subset of the original.
-                /// For `NOT IN`, the complement built below is therefore stricter than the true predicate
-                /// and would incorrectly exclude rows that should match. Example:
-                /// `tuple(i, i) NOT IN (tuple(1, 2))` deduplicates to `i NOT IN (1)`, building ranges
-                /// `(-inf, 1) ∪ (1, +inf)` and skipping `i = 1`, even though `tuple(1, 1) != tuple(1, 2)`.
-                /// Bail out and let the caller fall back to a conservative bound.
-                if (element.relaxed)
                     return false;
 
                 if (element.set_index->size() == 0)
--- a/tests/queries/0_stateless/04145_extract_plain_ranges_relaxed_not_in.sql
+++ b/tests/queries/0_stateless/04145_extract_plain_ranges_relaxed_not_in.sql
@@ -35,3 +35,24 @@ SELECT 'mergetree NOT IN tuple';
 SELECT i FROM t_extract_plain_ranges_relaxed WHERE tuple(i, i) NOT IN (tuple(1, 2)) ORDER BY i;
 
 DROP TABLE t_extract_plain_ranges_relaxed;
+
+-- `system.primes` consumes the same extracted ranges via `NumbersLikeUtils::extractRanges`.
+-- Without the fix, the relaxed set `{2}` would build the exact range `(-inf, 2) U (2, +inf)`,
+-- causing the source to skip the prime `2`.
+SELECT 'primes NOT IN tuple';
+SELECT prime FROM system.primes WHERE prime < 20 AND tuple(prime, prime) NOT IN (tuple(2, 3)) ORDER BY prime;
+
+-- The mirror `IN` predicate must remain consistent: no prime has `tuple(p, p) = tuple(2, 3)`.
+SELECT 'primes IN tuple';
+SELECT prime FROM system.primes WHERE prime < 20 AND tuple(prime, prime) IN (tuple(2, 3)) ORDER BY prime;
+
+-- `LIMIT` pushdown over exact ranges is the most user-visible failure mode: the source stops
+-- early, so even a small `LIMIT` reveals the missing row. With the fix, ranges are conservative
+-- and the row-level filter ensures the predicate is evaluated correctly before the `LIMIT` cuts in.
+SELECT 'numbers NOT IN tuple LIMIT';
+SELECT number FROM numbers(100) WHERE tuple(number, number) NOT IN (tuple(1, 2)) ORDER BY number LIMIT 3;
+
+-- Combined `system.primes` + `LIMIT` regression: without the fix the source would skip the prime
+-- `2`, returning `3, 5, 7, 11, 13` instead of `2, 3, 5, 7, 11`.
+SELECT 'primes NOT IN tuple LIMIT';
+SELECT prime FROM system.primes WHERE prime < 100 AND tuple(prime, prime) NOT IN (tuple(2, 3)) ORDER BY prime LIMIT 5;
--- a/tests/queries/0_stateless/04145_extract_plain_ranges_relaxed_not_in.sql
+++ b/tests/queries/0_stateless/04145_extract_plain_ranges_relaxed_not_in.sql
@@ -1,0 +1,37 @@
+-- Regression for https://github.com/ClickHouse/ClickHouse/issues/103660
+-- `KeyCondition::extractPlainRanges` must not treat a relaxed (deduplicated/transformed) set
+-- as exact. For `tuple(i, i) NOT IN (tuple(1, 2))` `MergeTreeSetIndex` deduplicates the tuple
+-- columns to a single key column, producing the relaxed set `{1}`. Building the complement
+-- `(-inf, 1) U (1, +inf)` and treating it as exact ranges incorrectly skips `i = 1`,
+-- even though `tuple(1, 1) != tuple(1, 2)`.
+
+-- `numbers()` consumes the extracted plain ranges directly, so the bug surfaces here.
+SELECT 'numbers NOT IN tuple';
+SELECT number FROM numbers(5) WHERE tuple(number, number) NOT IN (tuple(1, 2)) ORDER BY number;
+
+-- The `NOT has` variant goes through the same code path via `tryPrepareSetIndexForHas`.
+SELECT 'numbers NOT has tuple';
+SELECT number FROM numbers(5) WHERE NOT has([tuple(1, 2)], (number, number)) ORDER BY number;
+
+-- Per-row evaluation sanity check: every `tuple(n, n)` differs from `tuple(1, 2)`,
+-- so `NOT IN` is `1` for every row.
+SELECT 'per-row eval';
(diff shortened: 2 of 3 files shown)


groeneai:
**Pre-PR validation gate** (per TASK.md Phase 4 step 9):

```bash
./build/programs/clickhouse local --query "SELECT number FROM numbers(5) WHERE tuple(number, number) NOT IN (tuple(1, 2)) ORDER BY number"
```
Without fix: `0, 2, 3, 4` (missing `1`). With fix: `0, 1, 2, 3, 4`. 100% reproducible.

**b) Root cause explained?** `MergeTreeSetIndex` deduplicates tuple columns when multiple elements map to the same key column, reducing the set `{(1, 2)}` to `{1}` and marking it `relaxed = true`. The `FUNCTION_NOT_IN_SET` branch in `extractPlainRanges` builds the complement `(-inf, 1) U (1, +inf)` and `NumbersRangedSource` consumes these as exact ranges, skipping `number = 1`. The original predicate `tuple(1, 1) != tuple(1, 2)` is true for `number = 1`, so the row should be returned. The `relaxed` flag exists precisely to signal "this set check is not exact" but `extractPlainRanges` was ignoring it.

**c) Fix matches root cause?** Yes — `if (element.relaxed) return false;` in both `FUNCTION_IN_SET` and `FUNCTION_NOT_IN_SET` branches. The caller (`extractBounds`) ignores the conjunct and falls back to the universe range; the row-level filter then determines membership correctly. This mirrors existing relaxed-aware handling at `KeyCondition.cpp` lines 4661, 4852, 4914.

**d) Test intent preserved? / New tests added?** New regression test `04145_extract_plain_ranges_relaxed_not_in.sql` covers six scenarios: `numbers NOT IN tuple` (primary), `numbers NOT has tuple` (sibling), per-row eval (sanity), `numbers IN tuple` (mirror, defensive coverage), `generate_series NOT IN tuple` (different range source), and `MergeTree NOT IN tuple` (sanity that the existing `checkInRange` path stays correct). No existing test was weakened; existing `03905_set_partition_pruning_key_condition` continues to pass.

**e) Both directions demonstrated?** Confirmed.
- Test FAILS without the fix: 3 rows of value `1` missing in the diff against `.reference`.
- Test PASSES with the fix: 50/50 runs with full setting randomization.
- Related tests (`03905`, `03989`, `03821`, `03171`, `03208`, `03710`, `03215`, `04056`) all pass.

- Applied symmetrically to both `FUNCTION_IN_SET` and `FUNCTION_NOT_IN_SET`, even though only `NOT IN` was reported (defensive, since `IN` could regress `LIMIT` pushdown over `ExactRanges`).
- Targets the root cause (the `relaxed` flag), not the symptom.
- Pattern is consistent with the three pre-existing relaxed-handling sites in `KeyCondition::checkInHyperrectangle`.

Session: `cron:clickhouse-ci-task-worker:20260501-061500`

groeneai:
cc @nihalzp @CurtizJ @KochetovNicolai — could you review this? It fixes wrong results from `numbers()`, `generate_series()`, and similar range-honoring sources when the `WHERE` clause uses an `IN`/`NOT IN` predicate that `MergeTreeSetIndex` deduplicates into a relaxed set (e.g. `tuple(number, number) NOT IN (tuple(1, 2))`). `extractPlainRanges` now bails out on `element.relaxed`, mirroring the existing relaxed handling in `KeyCondition::checkInHyperrectangle`. Closes #103660.

tests/queries/0_stateless/04145_extract_plain_ranges_relaxed_not_in.sql
```diff
@@ -0,0 +1,37 @@
...
+-- so `NOT IN` is `1` for every row.
+SELECT 'per-row eval';
+SELECT number, tuple(number, number) NOT IN (tuple(1, 2)) AS val FROM numbers(5) ORDER BY number;
+
+-- The mirror `IN` predicate must remain consistent: no row has `tuple(n, n) = tuple(1, 2)`.
+SELECT 'numbers IN tuple';
+SELECT number FROM numbers(5) WHERE tuple(number, number) IN (tuple(1, 2)) ORDER BY number;
+
+-- `generate_series` shares the same range-honoring source. Its column is named `generate_series`.
+SELECT 'generate_series NOT IN tuple';
```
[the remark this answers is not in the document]
groeneai: Addressed in commit 9ffc499c — added 4 new queries to the test:

- `primes NOT IN tuple` — `SELECT prime FROM system.primes WHERE prime < 20 AND tuple(prime, prime) NOT IN (tuple(2, 3))`. Without the fix this skips prime `2`.
- `primes IN tuple` — defensive sanity check on the `FUNCTION_IN_SET` branch.
- `numbers NOT IN tuple LIMIT` — `SELECT number FROM numbers(100) WHERE tuple(number, number) NOT IN (tuple(1, 2)) ORDER BY number LIMIT 3`. Without the fix this returns `{0, 2, 3}` instead of `{0, 1, 2}`.
- `primes NOT IN tuple LIMIT` — combined `system.primes` + `LIMIT 5` regression. Without the fix this returns `{3, 5, 7, 11, 13}` instead of `{2, 3, 5, 7, 11}`.

Validated locally:
- Reverted the `element.relaxed` guard, rebuilt, ran the test → 5 missing rows in the diff (the new queries fail at the expected positions).
- Restored the guard, rebuilt, ran 50 iterations with full setting randomization → 50/50 pass.

nihalzp:
Is there any other situation inside `bool KeyCondition::extractPlainRanges(Ranges & ranges) const` where relaxation can give wrong result? Or any other situation where we are okay with relaxation?

groeneai:
@nihalzp Good question — I went through the whole function. Summary first, then the table, then the "why current consumers are safe" bit.

## TL;DR

For the only consumers of `extractPlainRanges` today (`numbers`, `numbers_mt`, `generate_series`, `system.numbers`, `system.primes` — all `UInt64`-keyed), `FUNCTION_IN_SET` and `FUNCTION_NOT_IN_SET` are the **only** branches that can receive a relaxed atom. Both are now bailing out. No other branch can produce wrong results for current callers.

For future-proofing (e.g. if a `String`-keyed consumer is ever added) `FUNCTION_IN_RANGE` and `FUNCTION_NOT_IN_RANGE` would also need a defensive `if (element.relaxed) return false;`. I haven't added them in this PR because I can't write a meaningful regression test today, but I can do it as a follow-up commit if you'd like — see the closing question.

## Where can `element.relaxed` be set?

Searching `KeyCondition.cpp`, `out.relaxed = true` is set in five places. I've grouped them by which RPN element they end up on:

| Source path | RPN element produced | Notes |
|---|---|---|
| `like` with imperfect prefix (line 526) | `FUNCTION_IN_RANGE` | requires `String` key |
| `match` regex (line 636) | `FUNCTION_IN_RANGE` | requires `String` key |
| `canConstantBeWrappedByMonotonicFunctions` succeeds, line 3450→3595 | any atom whose `func_name` is not in `no_relaxed_atom_functions` | typically `FUNCTION_IN_RANGE` / `FUNCTION_IN_SET` |
| `equals` / `notEquals` with `canConstantBeWrappedByDeterministicFunctions` and `is_injective == false`, line 3457→3595 | `FUNCTION_IN_RANGE` (`equals`) or `FUNCTION_NOT_IN_RANGE` (`notEquals`) | only place where a non-set inversion atom can become relaxed |
| Set-related dedup / transform (lines 2403, 2491, 3346, 3364, 3617) | `FUNCTION_IN_SET`, `FUNCTION_NOT_IN_SET` | the path that produced this PR's bug |

Plus the always-relaxed set: `always_relaxed_atom_elements = {FUNCTION_UNKNOWN, FUNCTION_ARGS_IN_HYPERRECTANGLE, FUNCTION_POINT_IN_POLYGON}`. These all land in the `else` arm at line 4490 and already `return false`.

The `no_relaxed_atom_functions` set (`{notLike, notIn, globalNotIn, notNullIn, globalNotNullIn, notEquals, notEmpty}`) blocks path 3 for inversion-shaped functions, with the rationale already documented at line 676: relaxation widens the predicate, and inverting a wider predicate would *narrow* it and miss matching rows. `notEquals` is still reachable via path 4 (the `equals/notEquals` deterministic-non-injective case) — that's the one path 4 entry above.

## Branch-by-branch audit of `extractPlainRanges`

| RPN element | What's pushed to the stack | Can leaf be relaxed? | Wrong result if relaxed and not bailing? | Status |
|---|---|---|---|---|
| `FUNCTION_AND` (4339) | `intersect(left, right)` | inherits | only if a child's range is *narrower* than the true predicate | derived; safe iff leaves are safe |
| `FUNCTION_OR` (4350) | `union(left, right)` | inherits | only if a child's range is *narrower* than the true predicate | derived; safe iff leaves are safe |
| `FUNCTION_NOT` (4361) | `invert(child)` | inherits | **yes** — inverting a *wider* (relaxed) child gives a *narrower* range, missing matching rows | safe today because no leaf-relaxed atom reaches this branch under any current consumer (see below) |
| `FUNCTION_IN_RANGE` (4383) | `element.range` | yes (paths 1, 2, 3, 4) | direct: no — range is wider, source over-emits, row filter cleans up. Indirect: yes if a later `FUNCTION_NOT` inverts it. | **not reachable today** for current consumers (paths 1, 2 need `String` key; paths 3, 4 ride on a function chain that triggers the early `hasMonotonicFunctionsChain()` bail at line 4332) |
| `FUNCTION_NOT_IN_RANGE` (4387) | `invertRange(element.range)` | yes (path 4 only — `notEquals` deterministic non-injective) | **yes** — `invertRange` of a single-point relaxed range is a *narrower* complement than the true `!=`, missing rows | **not reachable today** — the `notEquals` path requires a deterministic-but-not-injective function chain on the key, which sets `hasMonotonicFunctionsChain()` → early bail |
| `FUNCTION_IN_SET` (4391) | `element.set_index->getOrderedSet()` as point ranges | yes (path 5) | yes for `LIMIT` pushdown (over-emit + filter would short-circuit `LIMIT` early) | **fixed** by `if (element.relaxed) return false;` (line 4401) |
| `FUNCTION_NOT_IN_SET` (4424) | complement of set | yes (path 5) | **yes** — complement of a *deduped (narrower)* set is *narrower* than the true `NOT IN`, missing matching rows (this PR's reported bug) | **fixed** by `if (element.relaxed) return false;` (line 4435) |
| `ALWAYS_FALSE`, `ALWAYS_TRUE` | hardcoded blank / universe | n/a | n/a | safe |
| `FUNCTION_IS_NULL` (4481) | `makeBlank()` (key columns can't be NULL) | atom can be marked relaxed but the pushed range is hardcoded `makeBlank()`, so `element.range` and `element.relaxed` are unused here | n/a | safe |
| `FUNCTION_IS_NOT_NULL` (4486) | `makeUniverse()` | same as above — hardcoded universe, ignores `element.relaxed` | n/a | safe |
| `else` arm (4490): `FUNCTION_UNKNOWN`, `FUNCTION_ARGS_IN_HYPERRECTANGLE`, `FUNCTION_POINT_IN_POLYGON` | `return false` if `has_filter` | always relaxed via `always_relaxed_atom_elements` | already returns false | safe |

## Why are `FUNCTION_IN_RANGE` / `FUNCTION_NOT_IN_RANGE` "OK with relaxation" *for current consumers*?

The only callers of `extractPlainRanges` are `ReadFromSystemNumbersStep` (`numbers`, `generate_series`, `system.numbers`, `numbers_mt`) and `ReadFromSystemPrimesStep` (`system.primes`). All have `UInt64` keys. Working through each path that can mark `FUNCTION_*_IN_RANGE` relaxed:

- **Paths 1, 2** (`like` / `match`): require a `String` key. None of these consumers have one.
- **Path 3** (`canConstantBeWrappedByMonotonicFunctions`): runs only as the `else if` after `isKeyPossiblyWrappedByMonotonicFunctions` (path 1 in `extractAtomFromTree`). For these consumers, the key is the bare column, so path 1 in `extractAtomFromTree` succeeds with an empty chain, and we never enter path 3. Even if we did, the underlying transform would have to be applied to the const, which (for the cases that actually do something non-trivial) goes via a function chain on the key → `hasMonotonicFunctionsChain()` returns true → early bail at line 4332.
- **Path 4** (`equals` / `notEquals` with `canConstantBeWrappedByDeterministicFunctions`, `is_injective == false`): same story — non-injective deterministic wrap requires a function chain on the key, which causes the same early bail.

So the only relaxed atoms that can survive into the RPN stack for current consumers are `FUNCTION_IN_SET` / `FUNCTION_NOT_IN_SET`. Both bail out now.

I sanity-checked this empirically: 15 representative queries across `numbers`, `generate_series`, `system.primes`, with `IN` / `NOT IN` / `NOT (IN)` / `NOT (NOT IN)` / multi-element sets / `AND` / `OR` / `LIMIT` — all return correct results after the fix.

## Question back

Want me to also add a defensive `if (element.relaxed) return false;` to `FUNCTION_IN_RANGE` and `FUNCTION_NOT_IN_RANGE` as a follow-up commit on this PR? Pros: future-proof if a `String`-keyed consumer is ever added, makes the "relaxed → bail" rule consistent across all leaf branches. Cons: I can't add a regression test that exercises those branches today (no caller can reach them with a relaxed atom), so the guards would have to be commented as paranoid. I'd lean toward yes but want your call before churning the PR.

src/Storages/MergeTree/KeyCondition.cpp
nihalzp, line 4338: Okay, then why cannot we do `if (element.relaxed) return false;` on line 4339 right after for-loop start? This seems reasonable because if element is relaxed, the ranges we extract are not plain; additionally, this will cover all possible future cases with relaxation that may arrive.

groeneai: Done — pushed as 4fac0470 ["Move extractPlainRanges relaxation check to top of for-loop"](https://github.com/ClickHouse/ClickHouse/pull/103835/commits/4fac04709037).

Good call — this is much cleaner. The header on `KeyCondition::relaxed` (`KeyCondition.h` lines 590-595) confirms that operator elements (`FUNCTION_AND`, `FUNCTION_OR`, `FUNCTION_NOT`, `ALWAYS_TRUE`, `ALWAYS_FALSE`) are never set as relaxed — relaxation is propagated only through their child atoms — so a top-of-loop guard cannot misfire on conjunctions over relaxed atoms.

This simultaneously answers your earlier audit question ("any other situation where relaxation can give wrong result?") by construction: from now on, **no atom** can produce wrong results through `extractPlainRanges` regardless of which atom it is. The two per-branch guards have been removed.

Verified locally:

- `04145_extract_plain_ranges_relaxed_not_in` passes 50/50 with full setting randomization on a debug build.
- Negative direction: removing the guard causes the test to fail with row mismatches at the same positions as before (`numbers NOT IN tuple LIMIT` → `0,2,3` instead of `0,1,2`; `primes NOT IN tuple LIMIT` → `3,5,7,11,13` instead of `2,3,5,7,11`; etc.). So the new placement is doing the work, not a leftover guard.
- Empirical sanity over `numbers`/`generate_series`/`system.primes` with single- and multi-element tuple sets, AND/OR conjunctions, `LIMIT` pushdown, and plain `number IN/NOT IN` (non-relaxed) — all return correct results.
- Adjacent KeyCondition tests (`03905_set_partition_pruning_key_condition`, `03821_numbers_primes_filter_bound`, `03989_set_low_cardinality_in_tuple`, `03710_*`) all still pass.

nihalzp (approved):

clickgapai:

**Affected code:** `KeyCondition::extractPlainRanges` — core code present in all supported branches.

**Why:** Wrong-results bug user-visible in stable releases (issue #103660 was reported against a stable release). `extractPlainRanges` has been in `KeyCondition.cpp` for many releases and the relaxed-handling for `FUNCTION_IN_SET`/`FUNCTION_NOT_IN_SET` was already present (out.relaxed assignments at lines 2403/2491 predate this PR). Fix is a 5-line defensive guard with low backport risk.

PR #97813, opened 2026-02-23
Auto spilling join

antaljanosbenjamin:
Resolves https://github.com/ClickHouse/ClickHouse/issues/83677

### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

Add automatic spilling to hash and parallel hash joins by converting them to grace hash join when memory limit is reached. This behavior is controlled by `max_bytes_before_external_join`.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

---

> [!NOTE]
> **Medium Risk**
> Introduces a new join implementation that can switch from in-memory hash joins to disk-spilling `GraceHashJoin` at runtime, affecting core query execution and concurrency paths. Risk is mitigated by being opt-in via `max_bytes_before_external_join` plus extensive pipeline/test coverage, but regressions could impact join correctness/performance and read-in-order queries.
> 
> **Overview**
> Adds opt-in automatic spilling for `hash`/`parallel_hash`/`default`/`auto` joins via new setting `max_bytes_before_external_join`; when the right side exceeds the threshold, the join converts to `GraceHashJoin` to spill to disk (new `SpillingHashJoin`).
> 
> Updates join selection (analyzer + planner) to instantiate `SpillingHashJoin`, extends `IJoin` with `keepLeftPipelineInOrder()` and `canProcessNonJoinedBlocksInParallel()`, and adjusts pipeline wiring so delayed-join blocks and parallel non-joined streams can coexist safely.
> 
> Improves correctness/observability around spilling by enabling per-slot block extraction for `ConcurrentHashJoin`, filtering released blocks to avoid duplication, adding new join profile events (non-joined/delayed block counts + spill count), fixing a couple dictionary exception messages, and adding/adjusting stateless tests (including a new `03915_spilling_hash_join` suite and a read-in-order negative case).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit db5f199e34ff020d2ed1efd9a7bcde678e611f00. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### Version info
- Merged into: `26.5.1.140`
- Backported to: `26.4.1.1125`

PR #103838, opened 2026-05-01
Keep `max_bytes_before_external_join` peak under the configured cap

alexey-milovidov:
The auto-spilling join (#97813) could still hit `MEMORY_LIMIT_EXCEEDED` even when `max_bytes_before_external_join` was set. With a stats-warm cache, the user's transcript

```sql
SET max_memory_usage = '10G';
SET max_bytes_before_external_join = '5G';
SELECT count() FROM
  (SELECT number AS id, randomString(100) AS s FROM numbers(200000000)) AS a
  JOIN (SELECT number AS id FROM numbers(200000000)) AS b USING (id);
```

failed at peak `12.5 GiB` while attempting an `8 GiB` chunk allocation, even though the threshold was supposed to cap memory well below `10 GiB`.

Three independent issues conspired:

1. `ConcurrentHashJoin::reserveSpaceInHashMaps` preallocated capacity for `hint->ht_size` rows (e.g. 200M from a previous run) up front, blowing past the threshold before `SpillingHashJoin` could check it. Now we cap entries by `max_bytes_before_external_join / (8 * cell_size)` so the preallocated buffer stays under half the threshold while still keeping the optimization for queries that fit.
2. `SpillingHashJoin::addBlockToJoin` checked the threshold *after* the inner `addBlockToJoin` returned. Hash tables grow in power-of-two steps and a doubling from X to 2X transiently holds 3X, so the inner call could OOM during the resize before the post-call check ran. The check now runs proactively before the inner call, and uses half the threshold so the next doubling cannot push past the cap.
3. `GraceHashJoin::hasMemoryOverflow` only spilled on `max_rows_in_join`/`max_bytes_in_join`. When the wrapper handed data over, the in-memory bucket grew unbounded and the wrapper's spill decision became meaningless. It now also rehashes when `total_bytes` reaches half of `max_bytes_before_external_join`, gated by a 64 MiB minimum so unrealistically tiny test thresholds don't blow `grace_hash_join_max_buckets`.

Additionally, after rehashing, the new in-memory join was created with `reserve_num = prev_keys_num`, which itself allocated a power-of-two buffer for the pre-rehash size — another multi-gigabyte chunk. Reserve is now `prev_keys_num / buckets_snapshot.size()`, matching the rows that actually stay in this bucket.

The user's transcript: https://pastila.nl/?000b64fe/c41bd280359481e9fdf4c2b285b31943

Verification (peak observed via `clickhouse-client --memory-usage=default`):

| scenario | result | peak |
| --- | --- | --- |
| baseline (no limits) | 200M rows, 4.2s | 17.4 GiB |
| `max_memory_usage = 10G` only | fails (matches transcript) | 9.31 GiB |
| `max_memory_usage = 10G` + `max_bytes_before_external_join = 5G` | **succeeds**, 28.2s | **6.5 GiB** |

Existing `03915_spilling_hash_join` and `04061_spilling_hash_join_overflow_limits` still pass against the modified server.

Keep the auto-spilling hash join's actual memory peak under `max_bytes_before_external_join`. Previously, statistics-driven preallocation, in-place hash table doubling, and unbounded `GraceHashJoin` in-memory buckets could each push the query past the configured cap and trip `MEMORY_LIMIT_EXCEEDED`.

### Version info
- Merged into: `26.5.1.274`
- Backported to: `26.4.1.1136`


alexey-milovidov (commented):
The code looks good to me.

src/Interpreters/GraceHashJoin.cpp
```diff
@@ -40,6 +40,15 @@ namespace ErrorCodes
...
 }
 
+/// Floor for the automatic-spill check against `max_bytes_before_external_join`. Each rehash
+/// halves the in-memory bucket, so for total data D and threshold T we need ~log2(D / T)
+/// rehashes per bucket (capped by `grace_hash_join_max_buckets`). With smaller thresholds
+/// the hash-table cell overhead alone (~32 bytes per row) can already exceed `T/2` before
+/// any rehash can help, forcing one rehash per added block and quickly hitting the bucket
+/// cap. 1 MiB enforces the contract for any practical configuration while still leaving low
+/// test thresholds (e.g. 1 KiB in `03915_spilling_hash_join`) unaffected.
+static constexpr size_t MIN_THRESHOLD_FOR_AUTOMATIC_SPILL = 1024 * 1024;
```
antaljanosbenjamin: I don't like the fact that we are adding this artificial limit just to keep the tests working. Basically the tests now test a different behavior then in real world use cases. I would much rather prefer changing those tests.

src/Interpreters/ConcurrentHashJoin.cpp
```diff
@@ -109,8 +109,17 @@ void reserveSpaceInHashMaps(HashJoin & hash_join, size_t ind, const StatsCollect
         /// we need to preallocate in all buckets of all hash maps.
         const size_t reserve_size = hint->ht_size;
 
+        /// When a `SpillingHashJoin` wraps us, `max_bytes_before_external_join` caps the total memory we
+        /// are allowed to keep in memory before switching to `GraceHashJoin`. Statistics-driven preallocation
+        /// can reserve many gigabytes up front based on a previous larger query, blowing past that cap before
+        /// `SpillingHashJoin` ever runs its threshold check. We still want preallocation - just bounded by
+        /// the memory budget. We aim for about half of the threshold so that the cap itself plus the live
+        /// data plus the conversion peak still fit under it.
+        const auto external_join_threshold = hash_join.getTableJoin().maxBytesBeforeExternalJoin();
```
antaljanosbenjamin: I think it would be better if we would separate `max_bytes_before_external_join` and maximum size that concurrent hash join can reserve. E.g.: `join_algorithm = 'parallel_hash', max_bytes_before_external_join = 10MB` will cause concurrent hash performance to plummet. Let's make max reserve size as a parameter of ConcurrentHash join.

src/Interpreters/GraceHashJoin.cpp
```diff
@@ -326,6 +335,20 @@ bool GraceHashJoin::hasMemoryOverflow(size_t total_rows, size_t total_bytes) con
...
+    /// Also respect `max_bytes_before_external_join`. When the SpillingHashJoin wrapper hands data
+    /// to us after exceeding its threshold, we must keep spilling under the same cap; otherwise
+    /// the in-memory bucket would just keep growing and the wrapper's spill decision is meaningless.
+    /// We use half the threshold for the same reason as the wrapper: the in-memory hash table
+    /// doubles its buffer in power-of-two steps, transiently holding 3X the previous size, so
+    /// rehashing buckets early prevents that doubling from exceeding the cap. See the comment
+    /// on `MIN_THRESHOLD_FOR_AUTOMATIC_SPILL` for why a small floor is required.
+    if (!has_overflow)
+    {
+        const auto external_join_threshold = table_join->maxBytesBeforeExternalJoin();
```
antaljanosbenjamin: Same here as for concurrent hash join: `join_algorithm = 'grace_hash', max_bytes_before_external_join = 10MB` affects grace hash join when it shouldn't.

antaljanosbenjamin (approved):

PR #103840, opened 2026-05-01
Use `TYPE_MISMATCH` instead of `LOGICAL_ERROR` for direct join with `EmbeddedRocksDB`

alexey-milovidov:
A direct join between an `EmbeddedRocksDB` table and a subquery whose join key has a different type from the RocksDB primary key reaches `StorageEmbeddedRocksDB::getByKeys`, where the existing defensive check throws `LOGICAL_ERROR` with the message `Primary key type mismatch, expected ..., got ...`. This is reachable from a regular user query (the planner does not always insert a type conversion before a direct join, e.g. when `USING` is used), so it should surface as a user-facing `TYPE_MISMATCH` rather than a server-side logical error that aborts in debug builds.

Found by AST fuzzer `(amd_debug, targeted, old_compatibility)` while running CI for #101108:

- Report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=101108&sha=dbafc9d33a11336412d0f9ffdd53fbfbb560aa70&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%2C%20targeted%2C%20old_compatibility%29
- PR with the report: https://github.com/ClickHouse/ClickHouse/pull/101108

Triggering query (mutated by the fuzzer from `02242_join_rocksdb.sql`):

```sql
SELECT * FROM (SELECT k AS key FROM t2__fuzz_22) AS t2
INNER JOIN rdb USING (key) ORDER BY key ASC NULLS LAST
```

The existing test `02242_join_rocksdb.sql` already accepts `TYPE_MISMATCH` (or `NOT_IMPLEMENTED`) for the analogous `ON` form, so this change makes the `USING` path consistent.

...

### Version info
- Merged into: `26.5.1.304`


alexey-milovidov (commented):
Ok, fine.

PR #103880, opened 2026-05-01
Fix MSan use-of-uninitialized-value in JOIN-conversion partial evaluation (issue #100469, STID 1499-4a82)

groeneai:
The query-plan JOIN-conversion optimizers `tryConvertAnyOuterJoinToInnerJoin` (in `Processors/QueryPlan/Optimizations/convertOuterJoinToInnerJoin.cpp`) and `tryConvertAnyJoinToSemiOrAntiJoin` (`convertAnyJoinToSemiOrAntiJoin.cpp`) call `filterResultForNotMatchedRows` / `filterResultForMatchedRows` to decide whether the filter is always FALSE for synthetic default-input rows. Both helpers go through `ActionsDAG::evaluatePartialResult` with `skip_materialize=true`, which routes each FUNCTION node through `IFunction::executeImplDryRun`.

The dry-run override of `rowNumberInAllBlocks` returns `ColumnUInt64::create(input_rows_count)` and never fills the buffer, so the optimizer's subsequent `getFilterResult` -> `getBool(0)` reads uninitialized memory. This produces two failure modes:

1. **MemorySanitizer:** use-of-uninitialized-value, surfaced by AST fuzzer as STID 1499-4a82 (issue #100469).
2. **Correctness:** the optimizer's JOIN-conversion decision is taken from a garbage byte; when it happens to be 0 (FALSE) an `ANY OUTER JOIN` is silently converted to `INNER JOIN` (or an `ANY JOIN` to `SEMI`/`ANTI JOIN`), dropping rows that should have been kept.

Even with initialized dry-run output, the optimizer's reasoning is unsound for non-deterministic functions: a single plan-time evaluation of `rand` or `rowNumberInAllBlocks` cannot represent the per-row runtime behaviour. The fix adds `dagContainsNonDeterministicFunction` (mirroring the existing `dagContainsNonReadySet` guard) and calls it from both partial-evaluation entry points; when the filter (or pre-actions) DAG contains any function whose `isDeterministic` returns false, the helpers return `FilterResult::UNKNOWN` and the JOIN is left unchanged.

Regression test `04149_msan_join_optimizer_non_deterministic_filter.sql` covers `ANY LEFT`/`ANY RIGHT JOIN` with `rowNumberInAllBlocks` and `rand` in the filter, plus PedroTadim's original simplified reproducer. Verified locally on debug build (30/30 runs) and existing JOIN-conversion tests `03130_convert_outer_join_to_inner_join`, `03210_convert_outer_join_to_inner_join_any_join`, `03836_array_join_in_filter_partial_evaluation`, `04003_array_join_in_filter_outer_to_inner_join`, `04104_any_join_semi_anti_not_ready_set` continue to pass.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=77138227a27144afc763833bd59a93560bd193d4&name_0=MasterCI&name_1=AST%20fuzzer%20%28amd_msan%29

Closes #100469.

Fix a use-of-uninitialized-value in the JOIN-conversion query plan optimizers (`tryConvertAnyOuterJoinToInnerJoin`, `tryConvertAnyJoinToSemiOrAntiJoin`) when the filter on top of an `ANY` `OUTER` `JOIN` contains a non-deterministic function such as `rand`, `now`, or `rowNumberInAllBlocks`. The optimizer no longer attempts to constant-fold such filters and leaves the JOIN unchanged, which also prevents incorrect conversions to `INNER`/`SEMI`/`ANTI JOIN` that could silently drop rows.

### Version info
- Merged into: `26.5.1.321`


groeneai:
**Pre-PR validation gate (TASK.md Phase 4 Step 9)**

| # | Question | Answer |
|---|---|---|
| a | **Deterministic repro?** | Partially. The MSan failure is by definition non-deterministic without an MSan build (uninitialized byte may be 0 or non-zero). PedroTadim's reproducer + the new `04149_msan_join_optimizer_non_deterministic_filter.sql` exercise the broken code path on every run; the MSan part will be caught by the `AST fuzzer (amd_msan)` job in CI. |
| b | **Root cause explained?** | Yes. `IFunction::executeImplDryRun` for `rowNumberInAllBlocks` returns `ColumnUInt64::create(input_rows_count)` without filling the buffer. The optimizer reaches it via `evaluatePartialResult({skip_materialize=true})` -> `executeActionForPartialResult` -> `function->execute(..., dry_run=true)` -> `executeWithoutLowCardinalityColumns` -> `executeImplDryRun`. The result column is then read by `getFilterResult` -> `getBool(0)` on uninitialized memory. Beyond MSan, the optimizer's reasoning ("would the filter be FALSE for default-input rows?") is unsound for non-deterministic functions because their result does not depend only on inputs. |
| c | **Fix matches root cause?** | Yes. The check is at the right layer: bail out of the partial-evaluation entry points (`filterResultForNotMatchedRows`, `filterResultForMatchedRows`) when the DAG contains any function with `isDeterministic() == false`. This addresses both the MSan symptom (uninitialized memory is never read) and the soundness issue (no JOIN conversion is taken from a single plan-time evaluation of `rand`/`now`/`rowNumberInAllBlocks`). The fix mirrors the existing `dagContainsNonReadySet` guard for IN-subquery `ColumnSet`s. |
| d | **Test intent preserved? / new tests added?** | Both. New regression test `04149_msan_join_optimizer_non_deterministic_filter.sql` covers `ANY LEFT`/`ANY RIGHT JOIN` with `rowNumberInAllBlocks` and `rand`, plus PedroTadim's original simplified reproducer. Existing `03130_convert_outer_join_to_inner_join`, `03210_convert_outer_join_to_inner_join_any_join`, `03836_array_join_in_filter_partial_evaluation`, `04003_array_join_in_filter_outer_to_inner_join`, `04104_any_join_semi_anti_not_ready_set` continue to pass (5 runs each, 30/30). |
| e | **Demonstrated in both directions?** | Yes for the new test on debug build (30/30 with fix). The MSan bidirectional check (fails without fix, passes with fix) is delegated to the `AST fuzzer (amd_msan)` CI job because reproducing MSan use-of-uninitialized-value locally requires an MSan build (~20-40 min). The garbage-byte-induced INNER-conversion case is probabilistic: in this debug build the byte happens to be non-zero (TRUE) so the optimizer doesn't convert and the test passes by luck without the fix. |
| f | **Fix is general, not a narrow patch?** | Yes. (i) Both partial-evaluation entry points are guarded — `filterResultForNotMatchedRows` (in `Utils.cpp`, used by `convertOuterJoinToInnerJoin.cpp`) and `filterResultForMatchedRows` (in `convertAnyJoinToSemiOrAntiJoin.cpp`). (ii) The check uses `isDeterministic()` so it covers the whole class — `rand`, `now`, `rowNumberInAllBlocks`, `currentDatabase`, etc., not just the specific function from the failing AST-fuzzer query. (iii) Trace upstream: the `rowNumberInAllBlocks::executeImplDryRun` could also be patched to fill the buffer, but that would only fix one function; the optimizer's reasoning would still be unsound for `rand`/`now`. The fix at the optimizer level is the correct layer. |

Session: `cron:clickhouse-ci-task-worker:20260501-181500`

groeneai:
cc @KochetovNicolai @novikd @vdimir — could you review this? Adds a `dagContainsNonDeterministicFunction` guard to the JOIN-conversion partial-evaluation helpers (`filterResultForNotMatchedRows` / `filterResultForMatchedRows`) so we no longer feed `rand` / `now` / `rowNumberInAllBlocks` through `evaluatePartialResult({skip_materialize=true})` — that path reads uninitialized memory from `rowNumberInAllBlocks::executeImplDryRun` and (independently) makes JOIN-conversion decisions from a non-reproducible plan-time value. Closes #100469 (STID 1499-4a82, MSan). cc @azat as MSan-area reviewer.

src/Processors/QueryPlan/Optimizations/Utils.cpp
```diff
@@ -100,6 +114,14 @@ FilterResult filterResultForNotMatchedRows(
     if (dagContainsNonReadySet(filter_dag))
         return FilterResult::UNKNOWN;
 
+    /// If the filter DAG contains non-deterministic functions (e.g. `rand`, `now`, `rowNumberInAllBlocks`),
```
alexey-milovidov: `now` is deterministic in the scope of query, which means - it should work fine.

src/Processors/QueryPlan/Optimizations/Utils.cpp
```diff
@@ -100,6 +114,14 @@ FilterResult filterResultForNotMatchedRows(
...
+    /// we cannot soundly predict the filter result from default-input rows. Some such functions also produce
+    /// uninitialized data in their dry-run path (e.g. `rowNumberInAllBlocks::executeImplDryRun` returns a
```
alexey-milovidov: Fix the root cause - the functions should never produce uninitialized data.

alexey-milovidov:
@groeneai, check review comments.

groeneai:
@alexey-milovidov — root-cause fix applied per your review.

**Two new commits on top of the previous one:**

1. `eba94a4e3996` — Initialize output column in `rowNumberInAllBlocks::executeImplDryRun` (the actual root cause).
2. `db3cc22ee20d` — Remove the redundant `dagContainsNonDeterministicFunction` guard from `Utils.cpp` and `convertAnyJoinToSemiOrAntiJoin.cpp` (no longer needed once the function is fixed).

**The fix (one line, plus a comment):**

```cpp
// src/Functions/rowNumberInAllBlocks.cpp
ColumnPtr executeImplDryRun(... size_t input_rows_count) const override
{
    return ColumnUInt64::create(input_rows_count, 0);  // was: create(input_rows_count) — uninitialized
}
```

The dry-run override is preserved so the global `rows` counter is still not bumped (the dry-run must have no side effects), but the returned column is now fully initialized — which is the contract every other `executeImplDryRun` override in the tree already satisfies.

**Audit of all `executeImplDryRun` overrides:**

| File | Status |
|------|--------|
| `Functions/rowNumberInAllBlocks.cpp` | **Bug — fixed in this PR** (was allocating without filling) |
| `Functions/sleep.cpp` | OK — returns `createColumnConst(size, 0u)->convertToFullColumnIfConst()` |
| `Functions/in.cpp` | OK — returns `ColumnUInt8::create(size, 0)` or actual `set->execute(...)` |
| `Functions/UserDefined/UserDefinedWebAssembly.cpp` | OK — uses `insertManyDefaults(input_rows_count)` |
| `Functions/IFunctionAdaptors.cpp` | OK — pure delegating adapter |

Only `rowNumberInAllBlocks` violated the contract. The MSan finding was the symptom of that single bug.

**Verification (debug build):**

- `tests/clickhouse-test --test-runs 30 04149_msan_join_optimizer_non_deterministic_filter` — 30/30 pass.
- `tests/clickhouse-test --test-runs 5 03130_convert_outer_join_to_inner_join 03210_convert_outer_join_to_inner_join_any_join 03211_convert_outer_join_to_inner_join_anti_join 03595_convert_any_join_to_semi_or_anti 03623_convert_any_join_to_semi_or_anti_2 04006_convert_join_to_in_dummy_column` — 30/30 pass.
- MemorySanitizer validation delegated to CI's `AST fuzzer (amd_msan)` job.

Diff is now 91 lines added across 3 files (function fix + the test that was already there).

Re. your other comment about `now`: you're right — `now` returns true for `isDeterministicInScopeOfQuery` even though `isDeterministic` is false. My old guard would have filtered it incorrectly. With the function-level fix, the guard is gone and `now` flows through partial evaluation as expected.

Session: cron:clickhouse-ci-task-worker:20260502-001500

src/Processors/QueryPlan/Optimizations/Utils.cpp
groeneai: Done — `eba94a4e3996` fixes `rowNumberInAllBlocks::executeImplDryRun` to fill the output column with zeros (was allocating without filling), and `db3cc22ee20d` removes this whole guard since it is no longer needed. Audited every other `executeImplDryRun` override (`sleep.cpp`, `in.cpp`, `UserDefinedWebAssembly.cpp`, `IFunctionAdaptors.cpp`) — none of them produce uninitialized data; `rowNumberInAllBlocks` was the only contract violation.

src/Processors/QueryPlan/Optimizations/Utils.cpp
groeneai: Right — `now` is `isDeterministicInScopeOfQuery() == true` even though `isDeterministic() == false`. My old guard used `isDeterministic` and would have filtered `now` out as well, which was overly broad. With the function-level fix in `eba94a4e3996` the guard is gone entirely, so `now` (and `today`/`yesterday`/`currentUser`/etc.) flow through partial evaluation as before.

src/Functions/rowNumberInAllBlocks.cpp
```diff
@@ -58,7 +58,14 @@ class FunctionRowNumberInAllBlocks : public IFunction
...
     {
-        return ColumnUInt64::create(input_rows_count);
+        /// Dry-run callers (e.g. `ActionsDAG::evaluatePartialResult` used by JOIN-conversion optimizers
+        /// in `Processors/QueryPlan/Optimizations/Utils.cpp`) read the returned column via
+        /// `IColumn::getBool` and similar accessors. The previous implementation allocated the buffer
+        /// without filling it, so MemorySanitizer reported `use-of-uninitialized-value` (issue #100469,
+        /// STID 1499-4a82). Fill the column with zeros to keep the contract that any returned column
+        /// is fully initialized. We do not bump the global `rows` counter -- dry-run must have no side
+        /// effects.
+        return ColumnUInt64::create(input_rows_count, 0);
```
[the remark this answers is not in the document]
groeneai: Done — exactly what you suggested. Restored `dagContainsNonDeterministicFunction` with `isDeterministicInScopeOfQuery()` (not `isDeterministic()`) and reapplied it in both `filterResultForNotMatchedRows` (`Utils.cpp`) and `filterResultForMatchedRows` (`convertAnyJoinToSemiOrAntiJoin.cpp`).

* `7124daeb1d14` — restore guard + extend regression test (`rowNumberInAllBlocks() = 1` for `ANY LEFT` and `ANY RIGHT`)
* `0bb446371ed9` — switch the predicate from `isDeterministic` to `isDeterministicInScopeOfQuery` per @alexey-milovidov's review

Verified locally:

* The new test `04149_msan_join_optimizer_non_deterministic_filter` passes 30/30 with full randomization (was failing 2/20 with the bot's case before the guard was restored, with the symptom you described — the unmatched row gets dropped because the optimizer converts `ANY LEFT` to SEMI based on a single FALSE dry-run row).
* `now() < toDateTime('1900-01-01')` filter still triggers `ANY LEFT -> SEMI` conversion (deterministic in scope of query, so the optimizer can use the plan-time value soundly).
* `rand() < 0` filter keeps `Strictness: ANY` after the guard kicks in.
* All sibling JOIN-conversion stateless tests (`03130`, `03210`, `03211`, `03595`, `03623`, `04006`) still pass 30/30.

Function-level fix from `eba94a4e3996` is preserved — the contract that `executeImplDryRun` returns fully-initialized columns still holds, so the MSan use-of-uninitialized-value report stays fixed at the source. The optimizer guard is now an independent correctness fix on top of that.

tests/queries/0_stateless/04149_msan_join_optimizer_non_deterministic_filter.sql
```diff
@@ -0,0 +1,77 @@
...
+INSERT INTO t_l VALUES (1), (2);
+INSERT INTO t_r VALUES (1);
+
+-- ANY LEFT JOIN with a filter that contains `rowNumberInAllBlocks` (stateful, non-deterministic).
+-- Both rows from t_l must be returned; the unmatched row gets default 0 from t_r.
+-- Without the fix, the optimizer reads uninitialized memory from the dry-run output of
+-- `rowNumberInAllBlocks` while deciding whether to convert ANY OUTER JOIN to INNER JOIN.
+SELECT t_l.id, t_r.id
+FROM t_l ANY LEFT JOIN t_r ON t_l.id = t_r.id
+WHERE rowNumberInAllBlocks() < 100
```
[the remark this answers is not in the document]
groeneai: Added in `7124daeb1d14`. Two new cases inside this test, one for `ANY LEFT` and one for `ANY RIGHT`, both with `WHERE rowNumberInAllBlocks() = 1`:

```sql
SELECT count()
FROM t_l ANY LEFT JOIN t_r ON t_l.id = t_r.id
WHERE rowNumberInAllBlocks() = 1;

SELECT count()
FROM t_l ANY RIGHT JOIN t_r ON t_l.id = t_r.id
WHERE rowNumberInAllBlocks() = 1;
```

I assert the row count rather than the specific surviving row (`(2, 0)` for `LEFT`) because the value of `rowNumberInAllBlocks` depends on per-stream block ordering — under randomized `max_threads`/`max_insert_threads` the matched and unmatched rows can swap their `rn` values. The semantic property the test guards is "the JOIN-conversion optimizer must not collapse the `ANY LEFT JOIN` into SEMI/INNER based on a single plan-time evaluation of the non-deterministic filter": when the guard is in place, exactly one row passes the filter at runtime; without the guard, the count drops to 0. That's what we want to assert.

Verified that the test FAILS without the guard (count = 0) and PASSES with the guard (count = 1) for both `LEFT` and `RIGHT` variants. 30/30 random-settings runs pass with the fix.

groeneai:
@alexey-milovidov — addressed your review and the follow-up bot review.

**Summary**

The MSan fix from `eba94a4e3996` (initialize `rowNumberInAllBlocks::executeImplDryRun` output) stays — that's the root cause for the original `STID 1499-4a82` report.

After that commit, the bot caught a separate correctness issue: even with the dry-run output initialized to `[0]`, the JOIN-conversion optimizers would still incorrectly convert `ANY LEFT JOIN` to SEMI for filters like `rowNumberInAllBlocks() = 1` (FALSE on the single dry-run row, TRUE on some runtime row → silent unmatched-row drop). I confirmed that locally — `(2, 0)` was being dropped from the example you saw in the test.

**New commits on top:**

* `7124daeb1d14` — restore `dagContainsNonDeterministicFunction` in `filterResultForNotMatchedRows` / `filterResultForMatchedRows`, extend the regression test with the bot's failing case for `ANY LEFT` and `ANY RIGHT`.
* `0bb446371ed9` — switch the predicate from `isDeterministic` to `isDeterministicInScopeOfQuery` per your earlier note. So `now`/`today`/`yesterday`/`currentUser`/etc. (deterministic in scope of query) flow through partial evaluation as before, while `rand`/`rowNumberInAllBlocks`/`blockNumber`/`nowInBlock`/`generateSerialID` (per-row variability) bail out to `FilterResult::UNKNOWN`.

**Verified locally:**

* `now() < toDateTime('1900-01-01')` filter still triggers `ANY LEFT JOIN → SEMI JOIN` conversion (constant-folded to FALSE for not-matched rows).
* `rand() < 0` filter keeps `Strictness: ANY` after the guard kicks in.
* `04149_msan_join_optimizer_non_deterministic_filter` passes 30/30 with full randomization.
* All sibling JOIN-conversion tests (`03130`, `03210`, `03211`, `03595`, `03623`, `04006`) pass 30/30.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=77138227a27144afc763833bd59a93560bd193d4&name_0=MasterCI&name_1=AST%20fuzzer%20%28amd_msan%29
Closes: https://github.com/ClickHouse/ClickHouse/issues/100469

Session: cron:clickhouse-ci-task-worker:20260502-011500

groeneai:
Two follow-up commits addressing the new CI failure and the bot wording nit:

**`2cfba1fa`** — `Force enable_analyzer=1 in 04149_msan_join_optimizer_non_deterministic_filter`

The PR-monitor cron flagged a 3/3-rerun failure of `04149_msan_join_optimizer_non_deterministic_filter` in `Stateless tests (amd_llvm_coverage, old analyzer, s3 storage, DatabaseReplicated, WasmEdge, parallel)`:

```
Code: 47. UNKNOWN_IDENTIFIER. Missing columns 'p' while processing
'SELECT p = 1'. While processing ((SELECT p = 1) AS _subquery306) AND
(rowNumberInAllBlocks() = 1).
... TreeRewriter::collectUsedColumns
```

The MSan/soundness fix lives entirely in the planner-based path (`Processors/QueryPlan/Optimizations/`); the old analyzer does not run those JOIN-conversion optimizers. It additionally rejects `@PedroTadim`'s scalar-subquery reproducer shape `WHERE (SELECT p = 1) AND ...` in `TreeRewriter::collectUsedColumns`. So I forced the new analyzer for this whole test (`SET enable_analyzer = 1;` at the top), which matches every sibling JOIN-conversion regression test (`03130`, `03210`, `03595`, `03623`, `04003`).

Verification (local debug build):
- 30/30 passes under `--client-option=enable_analyzer=0` (old analyzer mode) with full random settings.
- 30/30 passes under default (new analyzer mode) with full random settings.
- Confirmed the failure in both directions: without the fix the test FAILS under old analyzer (the exact `UNKNOWN_IDENTIFIER` from CI); with the fix it passes.

CIDB cross-PR check confirmed the failure was isolated to PR #103880 — i.e. caused by our new regression test file under old analyzer.

**`513bb57e`** — `Clarify non-deterministic-function comments in JOIN-conversion guards`

Pure comment fix in `Utils.cpp` and `convertAnyJoinToSemiOrAntiJoin.cpp` per the bot review wording nit. The guard predicate is `isDeterministicInScopeOfQuery`, so `now` / `today` / `currentUser` (deterministic within a query) flow through and do NOT trip the bailout. The previous comments listed `now` alongside `rand` / `rowNumberInAllBlocks` in the per-row-variable group, which was misleading. Replaced with `nowInBlock` (genuinely per-row variable) and explicitly called out the in-scope-deterministic group.

No behavior change.

Branch tip: `513bb57e31d3`. CI re-running.

src/Processors/QueryPlan/Optimizations/Utils.cpp
```diff
@@ -89,6 +90,27 @@ bool dagContainsNonReadySet(const ActionsDAG & dag)
...
+{
+    /// We are interested in functions that are non-deterministic *within* a single query --
+    /// i.e. functions whose per-row output cannot be predicted from a single plan-time
+    /// evaluation. `rand`, `rowNumberInAllBlocks`, `blockNumber`, `nowInBlock` etc. fall in
+    /// this group. Functions like `now`/`today`/`yesterday`/`currentUser` are not
+    /// deterministic across queries (`isDeterministic() == false`) but they return the same
+    /// value for all rows in a single query (`isDeterministicInScopeOfQuery() == true`), so
+    /// the optimizer can soundly use their plan-time value and they should NOT block the
+    /// JOIN-conversion rewrite.
+    for (const auto & node : dag.getNodes())
```
alexey-milovidov, line 103: WDYT about adding methods isDeterministic, isDeterministicInScopeOfQuery to the DAG itself? This will make the code more encapsulated.

alexey-milovidov: Ok, I see similar code around, so let's keep this.

alexey-milovidov (approved):
Now it's LGTM.

alexey-milovidov:
@groeneai, update the branch to fix tests that were temporarily broken on master.

groeneai:
@alexey-milovidov — branch updated with current master at 1c611db20bcc to pick up the CI test fixes. New PR HEAD: fabc9bdc9532. CI is running on the merged HEAD now.

Session: cron:clickhouse-ci-task-worker:20260505-184500

vdimir (approved):

PR #103906, opened 2026-05-02
CI: remove BOLT profile files generated during cmake configuration

leshikus:
During BOLT profile collection, ClickHouse is configured with the BOLT-instrumented
`clang-21.inst`. CMake runs hundreds of `SUPPORTS_CFLAG_xxx` feature tests, each
invoking `clang-21.inst`, which writes a profile file to `BOLT_PROFILES_DIR`. This
produces several GB of files from trivial test programs before the real build starts,
causing a silent OOM/ENOSPC kill. Observed on all three attempts of #103307.

Fix: delete the cmake-phase profile files after configuration and before the
time-limited build. Those files don't represent real clang hot paths anyway.

### DIFF
--- a/ci/jobs/build_toolchain.py
+++ b/ci/jobs/build_toolchain.py
@@ -542,6 +542,20 @@ def main():
                 bolt_ok = False
                 print("BOLT profile collection cmake failed. Continuing with PGO-only.")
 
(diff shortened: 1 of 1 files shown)


leshikus:
worked on master without the fix

PR #103908, opened 2026-05-02
CI: add disk usage snapshot before BOLT cmake configuration (testing)

leshikus:
Testing PR to diagnose the BOLT toolchain build failure from #103307.

Every cmake compiler-feature test invokes `clang-21.inst` (BOLT-instrumented
clang), which writes a profile file per invocation. With hundreds of tests this
may fill several GB of disk before the real build starts, killing the runner
silently. This PR adds `du / | sort -rn | head -100` and `df -h` right before
the BOLT cmake step to confirm whether disk exhaustion is the root cause.

The `build_toolchain.py` change also busts the toolchain cache so the
`Build Toolchain (PGO, BOLT) (amd64)` job actually runs in the PR workflow.

### DIFF
--- a/ci/jobs/build_toolchain.py
+++ b/ci/jobs/build_toolchain.py
@@ -518,6 +518,17 @@ def main():
                 print(f"Failed to create clang++ symlink: {e}")
                 bolt_ok = False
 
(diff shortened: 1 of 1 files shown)


leshikus:
fixed in master

PR #103910, opened 2026-05-02
Fix SELECT error with unmerged patch parts when lazy materialization …

Chetas-Patil:
…is enabled

When lazy materialization is enabled, the result_header passed to readPatches could be empty, causing errors when trying to read and apply patch parts. This fix ensures that if the result_header is empty, we use the patch reader's header instead, allowing SELECT queries to work correctly even when there are unmerged patch parts.

The issue manifested as SELECT commands failing with errors instead of returning records when patch parts were yet to be merged, particularly when query_plan_optimize_lazy_materialization was enabled.

Fixes the known issue mentioned in test 04031 where lazy materialization had a separate bug with patch parts (empty read_sample_block passed to readPatches).

Generated with Devin (https://cli.devin.ai/docs)

- New Feature
- Experimental Feature
- Improvement
- Performance Improvement

...

### DIFF
--- a/src/Storages/MergeTree/MergeTreeReadersChain.cpp
+++ b/src/Storages/MergeTree/MergeTreeReadersChain.cpp
@@ -301,6 +301,9 @@ void MergeTreeReadersChain::addPatchVirtuals(Block & to, const Block & from) con
 
 void MergeTreeReadersChain::readPatches(const Block & result_header, std::vector<MarkRanges> & patch_ranges, ReadResult & read_result)
 {
+    /// If result_header is empty (e.g., due to lazy materialization), use the patch reader's header instead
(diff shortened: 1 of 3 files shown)


nikitamikhaylov:
I doubt this is the right fix, but maybe @CurtizJ can check.

azat:
@Chetas-Patil sorry but this does not fixes the problem:
- the header does not matches in the patches reader, which will throw LOGICAL_ERROR
- and the required columns (_block_offset/_block_number) is not there, so it cannot apply this optimization while reading patches
- and the test does not uses any lazy materialization, so not sure it will uncover the bug

Here is an alternative - https://github.com/ClickHouse/ClickHouse/pull/107616

PR #103919, opened 2026-05-02
Optimize row order if no order by

### LINKED ISSUE #103839: When a table has no ORDER BY key, we should use `optimize_row_order` by default
### Company or project name

ClickHouse

### Describe the situation

It will improve the compression and query speed and optimize many scenarios of automatically created tables.

.

### How to reproduce

.

### Expected performance

Add a new MergeTree setting, `optimize_row_order_if_no_order_by`, default 1.

The optimization is enabled if either the `optimize_row_order` is enabled or the table has no ORDER BY and `optimize_row_order_if_no_order_by` is enabled.

### Additional context

_No response_

Sasao4o:
- Improvement

...

Add optimize_row_order_if_no_order_by setting to automatically rearrange rows when no ORDER BY is defined.

This closes #103839


src/Storages/MergeTree/MergeTreeDataWriter.cpp
```diff
@@ -715,7 +716,7 @@ MergeTreeTemporaryPartPtr MergeTreeDataWriter::writeTempPartImpl(
             ProfileEvents::increment(ProfileEvents::MergeTreeDataWriterBlocksAlreadySorted);
     }
 
-    if ((*data_settings)[MergeTreeSetting::optimize_row_order]
+    if (((*data_settings)[MergeTreeSetting::optimize_row_order] || ((*data_settings)[MergeTreeSetting::optimize_row_order_if_no_order_by] &&!metadata_snapshot->hasSortingKey()))
```
alexey-milovidov:
```suggestion
    if (((*data_settings)[MergeTreeSetting::optimize_row_order] || ((*data_settings)[MergeTreeSetting::optimize_row_order_if_no_order_by] && !metadata_snapshot->hasSortingKey()))
```

src/Storages/MergeTree/MergeTreeDataWriter.cpp
```diff
@@ -1057,7 +1058,7 @@ MergeTreeTemporaryPartPtr MergeTreeDataWriter::writeProjectionPartImpl(
             ProfileEvents::increment(ProfileEvents::MergeTreeDataProjectionWriterBlocksAlreadySorted);
     }
 
-    if ((*data_settings)[MergeTreeSetting::optimize_row_order]
+    if (((*data_settings)[MergeTreeSetting::optimize_row_order] || ((*data_settings)[MergeTreeSetting::optimize_row_order_if_no_order_by] &&!metadata_snapshot->hasSortingKey()))
```
alexey-milovidov:
```suggestion
    if (((*data_settings)[MergeTreeSetting::optimize_row_order] || ((*data_settings)[MergeTreeSetting::optimize_row_order_if_no_order_by] && !metadata_snapshot->hasSortingKey()))
```

alexey-milovidov (changes requested):

Sasao4o:
Hello @rschu1ze 

I tested the performance impact by mutating the hits dataset (https://clickhouse.com/docs/getting-started/example-datasets/metrica). 

I produced 4 derived datasets (with different widths and cardinalities).
For example, I truncated columns to create narrow table cases, and removed high-cardinality columns to create low-cardinality-only cases, etc.

I stopped merges (SYSTEM STOP MERGES) because they introduced a lot of noise.

I found that enabling the feature is always 20%–35% slower, with one exception: when the table is wide and contains strings, arrays, and low-cardinality data, it becomes up to 140% slower.

The compression benefit is not very large even in the low-cardinality case (which is weird for me; maybe we need to increase min_insert_block_size_rows so there is a win). It is around 2%–2.5% better compression ratio.

rschu1ze:
Thanks @Sasao4o for the interesting investigation. 20%-30% slower inserts (at a minimum) for <5% better compression isn't worth the trouble, let's not continue this PR.

alexey-milovidov:
@rschu1ze, the analysis is correct, but the conclusion is unsound.

The inserts with automatic sorting are expected to be slower, but similar to the case when there is an explicitly specified sorting key.

PR #103874, opened 2026-05-01
Fix `clickhouse-local` hang in shutdown when system logs are configured

alexey-milovidov:
`SystemLogs::flushImpl` calls `BaseDaemon::instance().flushTextLogs()` to drain the async logger before iterating system log queues. `BaseDaemon::instance()` is `dynamic_cast<BaseDaemon &>(Poco::Util::Application::instance())`, so when the application is a `ClientApplicationBase` (e.g. `clickhouse-local`, `clickhouse-client`) the cast throws `std::bad_cast`. The exception escapes `SystemLogs::flushAndShutdown` -> `DatabaseCatalog::shutdownImpl` -> `Context::shutdown`, leaving every saving thread alive.

When the last `Context` reference goes away, `~ContextSharedPart` runs `~SystemLogs` -> `~SystemLog<T>`. `queue` is declared in `SystemLogBase<T>` and `saving_thread` in `ISystemLog`, so member destruction tears down the `SystemLogQueue` and its condition variables before the saving thread has been joined. `pthread_cond_destroy` then blocks forever on the saving thread parked in `pop()`. The periodic saving-thread flushes that fire after `~Application` already nulled `_pInstance` surface as `Null pointer: _pInstance` from `Context::getConfigRef()` -> `Poco::Util::Application::instance().config()` (via `InterpreterCreateQuery::createTable` -> `UserDefinedSQLFunctionFactory::empty` -> `Context::getUserDefinedSQLObjectsStorage`), which is the symptom reported on the docs check.

Two fixes:

* `SystemLogs::flushImpl` now calls a small `flushAsyncTextLogsIfPossible` helper that uses `BaseDaemon::tryGetInstance` (a `dynamic_cast<BaseDaemon *>` internally) and skips the call when the application is not a `BaseDaemon`, so flushing on `clickhouse-local`/`clickhouse-client` is a no-op instead of raising `std::bad_cast`.
* `SystemLogBase<T>` now has an explicit destructor that calls `stopFlushThread()`, joining the saving thread before any of the queue's condition variables can be destroyed. Defence-in-depth so any future path that bypasses `flushAndShutdown` cannot reproduce the same hang.

Reproduction with the docs-job config (~30 system log sections, `SELECT 1`) on master: query takes 54+ s and the process hangs in `pthread_cond_destroy` from `~SystemLogQueue<FilesystemCacheLogElement>`. With this change the same invocation completes in ~3.2 s (residual time is the cost of constructing every configured `SystemLog`) and exits cleanly.

[...]

PR #103938, opened 2026-05-03
CLAUDE.md: discourage fallback paths, prefer fail-close

alexey-milovidov:
Adds a guideline to `.claude/CLAUDE.md` codifying our preference: errors should propagate by default rather than be hidden behind silent fallbacks, and any unavoidable fallback must follow the fail-close principle (no destructive or heavy actions on the fallback path — skip the operation and surface the error instead).

Motivation: this came up while reviewing #103937, where the original implementation silently treated must-backport labels as "human-added" if the events API errored, which would have created backports we couldn't actually attribute. Codifying this avoids re-litigating the same tradeoff in future changes.

Internal contributor guidance only.

### Version info
- Merged into: `26.5.1.295`


alexey-milovidov:
@groeneai, why did "Generate system tables documentation" fail? If it is fixed by any open PRs, link the fix here.

groeneai:
@alexey-milovidov — investigated this run.

### Failure: `Docs check` → `Generate system tables documentation`

The step uses `clickhouse-local` to query `system.tables` for the system-tables docs page. It hangs on shutdown and trips the 30-second timeout, so the docs aren't written.

### Root cause

PR [#102316](https://github.com/ClickHouse/ClickHouse/pull/102316) (merged 2026-05-01 05:13 UTC) added `initializeSystemLogs()` to `clickhouse-local`. With `text_log` configured, `SystemLogs::flushImpl` does a `dynamic_cast<BaseDaemon &>(...)` on shutdown — but `clickhouse-local` derives from `ClientApplicationBase`, not `BaseDaemon`. The cast throws `std::bad_cast` mid-shutdown and the process hangs.

CIDB confirms the timing: `Generate system tables documentation` failures across PRs accelerated from 2026-05-01 05:13 UTC onward — 32 failures across 21 distinct PRs in 30 days, all PR-only (master `Docs check` doesn't run this step).

### Fix in flight

[#103874](https://github.com/ClickHouse/ClickHouse/pull/103874) — "Fix `clickhouse-local` hang in shutdown when system logs are configured", by you, opened 2026-05-01 16:38 UTC, currently OPEN and MERGEABLE.

Approach: guard the `dynamic_cast<BaseDaemon &>(...)` with `BaseDaemon::tryGetInstance()` so `clickhouse-local` skips the daemon-specific shutdown path.

This is the right fix — once merged the `Docs check` failure goes away on rebased PRs.

PR #102356, opened 2026-04-10
Add `null_count` statistics

zoomxi:
### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Added NullCount statistics type for accurate NULL value counting in Nullable columns. This enables improved `PREWHERE` optimization and better part pruning.

### Documentation entry for user-facing changes

- [*] Documentation is written (mandatory for new features)

PR #103890, opened 2026-05-02
Fix use-after-free in KeeperHandlingConsumer::setKeeper (StorageKafka2)

groeneai:
## What

Fix a use-after-free in `DB::KeeperHandlingConsumer::setKeeper()` (used by `StorageKafka2`) where, on a Keeper session swap, the lock holders could outlive the `ZooKeeper` instance they reference and crash in `~zkutil::EphemeralNodeHolder()`.

## Why / root cause

`LockedTopicPartitionInfo::lock` (held in `permanent_locks` and `tmp_locks`) is a `zkutil::EphemeralNodeHolder`, which stores a **raw** `ZooKeeper &` reference captured at construction time (see `EphemeralNodeHolder` in `Common/ZooKeeper/ZooKeeper.h`). Its destructor calls `zookeeper.expired()` and possibly `zookeeper.tryRemove()`.

`KeeperHandlingConsumer::setKeeper()` was assigning the new `keeper` shared_ptr **before** clearing the lock maps:

```cpp
void KeeperHandlingConsumer::setKeeper(const std::shared_ptr<zkutil::ZooKeeper> & keeper_)
{
    keeper = keeper_;             // <-- (1) drops shared_ptr to old session
    {
        std::lock_guard lock(topic_partition_locks_mutex);
        permanent_locks.clear();  // <-- (2) destroys EphemeralNodeHolders that hold ZooKeeper&
        tmp_locks.clear();
    }
    ...
}
```

If the consumer was the last shared owner of the previous `ZooKeeper` (which is plausible: by the time `streamToViews()` calls `setKeeper`, `StorageKafka2::keeper` and `Context`'s keeper have typically already moved on to the new session), then step (1) synchronously destroys the old `ZooKeeper`. Step (2) then runs the holders' destructors which call `zookeeper.expired()` on the freed object — SIGSEGV.

This is the **same hazard** already documented and fixed in `StorageKafka2::partialShutdown()` for `replica_is_active_node`. From `src/Storages/Kafka/StorageKafka2.cpp` lines 221-225:

```cpp
/// Reset the active node holder while the old ZooKeeper session is still alive (even if expired).
/// EphemeralNodeHolder stores a raw ZooKeeper reference, so resetting it here prevents a
/// use-after-free: setZooKeeper() called afterwards may free the old session, and the holder's
/// destructor would then access a dangling reference when checking zookeeper.expired().
replica_is_active_node = nullptr;
```

[...]

PR #103949, opened 2026-05-03
support three-valued logic in SelectivityEstimator

hanfei1991:
Follow up https://github.com/ClickHouse/ClickHouse/pull/102356

- Improvement

Fixed selectivity estimation for nullable columns: `x IS NULL OR x > 50 OR y IS NULL `,                                                                                                                     and similar predicates now correctly follow SQL three-valued logic and exclude NULL rows                                                                                                                        from the TRUE selectivity.                                                                                                                            

### Version info
- Merged into: `26.5.1.512`

### DIFF
--- a/src/Storages/Statistics/ConditionSelectivityEstimator.cpp
+++ b/src/Storages/Statistics/ConditionSelectivityEstimator.cpp
@@ -107,10 +107,10 @@ RelationProfile ConditionSelectivityEstimator::estimateRelationProfileImpl(std::
                     right_element->finalize(column_estimators, metadata);
                     /// P(c1 and c2) = P(c1) * P(c2)
                     if (element.function == RPNElement::FUNCTION_AND)
-                        element.selectivity = left_element->selectivity * right_element->selectivity;
+                        element.selectivity = left_element->selectivity.applyAnd(right_element->selectivity);
                     /// P(c1 or c2) = 1 - (1 - P(c1)) * (1 - P(c2))
                     else
-                        element.selectivity = 1-(1-left_element->selectivity)*(1-right_element->selectivity);
+                        element.selectivity = left_element->selectivity.applyOr(right_element->selectivity);
                     element.finalized = true;
                     rpn_stack.push(&element);
                 }
@@ -120,7 +120,7 @@ RelationProfile ConditionSelectivityEstimator::estimateRelationProfileImpl(std::
             {
                 auto* last_element = rpn_stack.top();
                 if (last_element->finalized)
-                    last_element->selectivity = 1 - last_element->selectivity;
+                    last_element->selectivity = last_element->selectivity.applyNot();
                 else
                 {
                     std::swap(last_element->column_ranges, last_element->column_not_ranges);
@@ -145,7 +145,7 @@ RelationProfile ConditionSelectivityEstimator::estimateRelationProfileImpl(std::
     auto * final_element = rpn_stack.top();
     final_element->finalize(column_estimators, metadata);
     RelationProfile result;
-    Float64 final_rows = final_element->selectivity * static_cast<Float64>(total_rows);
+    Float64 final_rows = final_element->selectivity.true_sel * static_cast<Float64>(total_rows);
     /// Clamp to [0, total_rows] and handle NaN/Inf to avoid undefined behavior
     /// in the float-to-UInt64 cast below (UBSAN float-cast-overflow).
     if (!std::isfinite(final_rows) || final_rows < 0)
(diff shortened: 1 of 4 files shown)


src/Storages/Statistics/ConditionSelectivityEstimator.cpp
```diff
@@ -221,14 +221,14 @@ bool ConditionSelectivityEstimator::extractAtomFromTree(const StorageMetadataPtr
             /// LIKE/ILIKE cannot be represented as a range. Pre-set selectivity
             /// so the estimator uses a tighter default than `default_unknown_cond_factor`.
             if (func_name == "like" || func_name == "ilike")
```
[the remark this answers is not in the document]
hanfei1991: We only consider range condition in this PR

src/Storages/Statistics/ConditionSelectivityEstimator.cpp
```diff
@@ -473,19 +498,11 @@ Float64 ConditionSelectivityEstimator::ColumnEstimator::estimateRanges(const Pla
...
-}
-
-Float64 ConditionSelectivityEstimator::ColumnEstimator::estimateNotRanges(const PlainRanges & ranges) const
-{
-    if (stats->getNumRows() == 0)
-        return 0;
-    Float64 null_selectivity = 1.0 - static_cast<Float64>(stats->getNonNullRowCount()) / static_cast<Float64>(stats->getNumRows());
-    Float64 ranges_selectivity = estimateRanges(ranges);
-    return std::max(0.0, std::min(1.0, 1.0 - ranges_selectivity - null_selectivity));
+    return {std::max(0.0, std::min(1.0, selectivity)), static_cast<Float64>(stats->getNullCount()) / rows};
```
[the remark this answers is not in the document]
hanfei1991: I don't think it's going to happen because `selectivity` only counts non-null rows

src/Storages/Statistics/Statistics.cpp
```diff
@@ -273,10 +273,16 @@ Estimate ColumnStatistics::getEstimate() const
...
 UInt64 ColumnStatistics::getNullCount() const
 {
-    auto it = stats.find(StatisticsType::NullCount);
-    if (it == stats.end())
-        throw Exception(ErrorCodes::LOGICAL_ERROR, "getNullCount called but NullCount statistics not present");
-    return assert_cast<const StatisticsNullCount &>(*it->second).getNullCount();
+    if (auto it = stats.find(StatisticsType::NullCount); it != stats.end())
+        return assert_cast<const StatisticsNullCount &>(*it->second).getNullCount();
+    /// MinMax::row_count tracks only non-NULL rows, so the null count can be derived
+    /// as total rows minus MinMax row count when NullCount statistics are absent.
```
murphy-4o, line 279: but why the null_count can be absent ?

src/Storages/Statistics/ConditionSelectivityEstimator.cpp
```diff
@@ -459,10 +459,35 @@ ConditionSelectivityEstimatorPtr ConditionSelectivityEstimatorBuilder::getEstima
...
+ConditionSelectivityEstimator::Selectivity ConditionSelectivityEstimator::Selectivity::applyNot() const
+{
+    return {1.0 - true_sel - null_sel, null_sel};
+}
+
+ConditionSelectivityEstimator::Selectivity ConditionSelectivityEstimator::Selectivity::applyOr(const Selectivity & other) const
+{
+    /// case1 : NULL or (false/NULL) = NULL
+    /// case2 : false or NULL = NULL
+    /// case1 : true or (...) = true
```
murphy-4o: case3 ?

src/Storages/Statistics/ConditionSelectivityEstimator.cpp
```diff
@@ -656,14 +673,14 @@ void ConditionSelectivityEstimator::RPNElement::finalize(const ColumnEstimators
         for (const Range & range : ranges.ranges)
         {
             if (range.isInfinite())
-                return 1.0;
+                return Selectivity{1.0, 0.0};
```
murphy-4o, line 676: looks like provide static function rather than literal value can be better?
`Selectivity::all` `Selectivity::none` `Selectivity::defaultUnknown`

src/Storages/Statistics/Statistics.cpp
hanfei1991: when the user doesn't create null_count for the column

src/Storages/Statistics/ConditionSelectivityEstimator.cpp
hanfei1991: We can understand the meaning of (1.0, 0,0) at first glace, but Selectivity::all/none is not so direct IMHO

src/Storages/Statistics/ConditionSelectivityEstimator.cpp
hanfei1991:
```suggestion
    /// case3 : true or (...) = true
```

src/Storages/Statistics/tests/gtest_stats.cpp
```diff
@@ -413,6 +413,7 @@ TEST(Statistics, NullPredicateEstimates)
 
     auto check = [&](const String & expression, Float64 expected_rows, Float64 eps)
     {
+        std::cout << "check: " << expression << std::endl;
```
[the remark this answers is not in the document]
hanfei1991: I would like to keep this, because we don't know which test is failed without these debug info.

Otherwise, I could find a better way to log in unittest

src/Storages/Statistics/Statistics.cpp
murphy-4o: interesting. 
but from my point view this kind of basic statistics should be always available, so that we can simplify the assumption.

src/Storages/Statistics/tests/gtest_stats.cpp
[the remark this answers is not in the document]
murphy-4o: not a big problem, but logging in unittest can be like `EXPECT_LT(a, b) << "expression xxx"`

murphy-4o (approved):
LGTM

hanfei1991:
CI failed because of https://github.com/ClickHouse/ClickHouse/pull/103890

Issue #47041, opened 2023-02-28
Table functions `generate_series`, `generate_date_array`, `generate_timestamp_array`


alexey-milovidov:
**Use case**

Compatibility with Postgres and BigQuery.

**Describe the solution you'd like**

We already have the `numbers` table function, but we can add these functions for convenience.

**Additional context**

https://www.postgresql.org/docs/current/functions-srf.html

alexey-milovidov:
@zhujing0227 yes, it makes sense to extend the `range` function to support Date, DateTime, Date32 and DateTime64:

`range(date1, date2)` - will generate dates with one day step by default;
`range(time1, time2)` - will generate date-times with one-second step by default;
`range(time1, time2, INTERVAL ...)` - will generate a range with a specified step interval.

iiiuwioajdks:
I want to try to implement it, can i know more about this problem

Ziy1-Tan:
take

hillbilly991:
was there any update on this? I am trying to implement something like this now

hp77-creator:
@alexey-milovidov so extending range should be the way or would you prefer new functions like how its in BigQuery?

PR #103963, opened 2026-05-03
implement generate date array

EduardoGRocha:
Related: https://github.com/ClickHouse/ClickHouse/issues/47041

- New Feature

Added function `generate_date_array(start, end[, step])` that returns an array of dates from `start` to `end` (inclusive) incremented by a constant interval `step` (`Day`, `Week`, `Month`, `Quarter`, or `Year`; default `INTERVAL 1 DAY`), for compatibility with BigQuery's `GENERATE_DATE_ARRAY`.

---

### Details

This PR implements `generate_date_array`, modeled after BigQuery's [`GENERATE_DATE_ARRAY`](https://cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#generate_date_array):

```sql
SELECT generate_date_array('2024-01-01', '2024-06-01', INTERVAL 1 MONTH);
-- ['2024-01-01','2024-02-01','2024-03-01','2024-04-01','2024-05-01','2024-06-01']
```

**Behavior:**
- `start`/`end` accept `Date`, `Date32`, or string literals; constant strings are inferred as the smallest fitting date type (`Date`, falling back to `Date32` for e.g. pre-1970 dates). The result element type is the least supertype of the inputs.
- `step` must be a constant, non-zero interval of kind `Day`, `Week`, `Month`, `Quarter`, or `Year`. A negative step produces a descending sequence; if `end` is unreachable in the step's direction, an empty array is returned.
- A bare `NULL` argument makes the result `NULL`; a runtime `NULL` inside a `Nullable` column throws `BAD_ARGUMENTS`.
- Throws `ARGUMENT_OUT_OF_BOUND` if the call would produce more than 500 million array elements.

**Implementation notes:**
- Day/Week steps use O(1) length computation and plain integer arithmetic in the fill loop; Month/Quarter/Year steps use calendar arithmetic (`AddMonths/Quarters/YearsImpl`). Both paths share one templated implementation.
- Tests cover the BigQuery documentation examples plus type-inference, mixed-type, and error cases (`tests/queries/0_stateless/04141_generate_date_array.sql`).


src/Core/Settings.cpp
```diff
@@ -6165,6 +6141,18 @@ Possible values:
 
 **See Also**
 
+- [`max_block_size`](#max_block_size)
```
rschu1ze: A setting isn't needed, let's remove it.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+    explicit FunctionGenerateDateArray(ContextPtr context)
+        : max_elements(context->getSettingsRef()[Setting::function_generate_date_array_max_elements_in_block])
+    {
+    }
+
+private:
+    String getName() const override { return name; }
+    size_t getNumberOfArguments() const override { return 0; }
+    bool isVariadic() const override { return true; }
+    // While any null in the input should cause it to return NULL,
```
rschu1ze: Nit: Please use `///` everywhere.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+    //  it seems (from the documentation) that if some arguments are nullable, the code still gets called
+    //  So we would still need to write code that is "safe" regardless of what is passed
+    //  At the very least, this would cause massive allocation and inefficient execution
+    bool useDefaultImplementationForNulls() const override { return false; }
```
rschu1ze: src/Functions/IFunction.h says this:

```cpp
/** Default implementation in presence of Nullable arguments or NULL constants as arguments is the following:
  *  if some of arguments are NULL constants then return NULL constant (the underlying function
  *   may or may not be executed in this case),
  *  if some of arguments are Nullable, then execute function as usual for columns,
  *   where Nullable columns are substituted with nested columns,
  *   and wrap result in Nullable column where NULLs are in all rows where any of arguments are NULL.
  * The underlying function may or may not be called for the rows containing NULLs. When called,
  * arbitrary values are used instead of NULLs - whatever values happened to be in ColumnNullable's
  * nested column; typically default values, but that's not guaranteed.
  * So this only makes sense for functions that don't fail and don't have side effects.
  */
virtual bool useDefaultImplementationForNulls() const { return true; }
```

The default behavior seems reasonable and also aligned with the [BigQuery docs](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/array_functions#generate_date_array). Can you remove l. 55-70, please?

We also wouldn't need l. 114-116 then.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+    bool useDefaultImplementationForConstants() const override { return true; }
+    /// The step argument must always be a compile-time constant expression.
```
rschu1ze: Let's remove l. 72-74, the semantics are clear anyways and the comment is too verbose (probably AI generated).

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+        /// DayNum (UInt16) covers 1970-01-01 .. 2149-06-06 (values 0..65535).
+        if (day_num >= 0 && day_num <= std::numeric_limits<UInt16>::max())
+            return std::make_shared<DataTypeDate>();
+        return std::make_shared<DataTypeDate32>();
+    }
+
+    DataTypePtr getReturnTypeImpl(const ColumnsWithTypeAndName & arguments) const override
+    {
+        if (arguments.size() < 2 || arguments.size() > 3)
```
rschu1ze: Can you try to replace the code in `getReturnTypeImpl` by the declarative syntax provided by `FunctionArgumentDescriptor` and `validateFunctionArguments` from "src/Functions/FunctionHelpers.h"? Search src/Functions/ for precedence.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+- When `start` and `end` have different types, the element type of the returned array is their least supertype (e.g. mixing `Date` and `Date32` yields `Array(Date32)`).
+- `step` must be a **constant** non-zero interval of kind `Day`, `Week`, `Month`, `Quarter`, or `Year`. It cannot be a column reference or a non-constant expression. When omitted, the default step is `INTERVAL 1 DAY`.
+- If `step` is positive the sequence is ascending. If `step` is negative the sequence is descending.
+- If `end` is unreachable in the given direction, an empty array is returned.
+- Returns `NULL` if any argument is of type `Nullable(Nothing)` (i.e. a bare `NULL` literal). Throws an exception if any argument has a runtime `NULL` value inside a `Nullable` column.
+- Throws an exception if the total number of elements produced exceeds the limit set by [`function_generate_date_array_max_elements_in_block`](../../operations/settings/settings.md#function_generate_date_array_max_elements_in_block).
+    )";
+    FunctionDocumentation::Syntax syntax = "generate_date_array(start, end[, step])";
+    FunctionDocumentation::Arguments arguments = {
+        {"start",
```
rschu1ze, line 279: Nit: `Arguments` has an optional third argument for the data type.

Example (from another function):

```cpp
{"word", "Word that will be searched in extension.", {"String"}}
```

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+) ORDER BY date_range;
+        )",
+         R"(
+┌─generate_date_array(date_start, date_end)─AS date_range──────────────────────────────────────┐
+│ ['2024-01-01','2024-01-02','2024-01-03','2024-01-04','2024-01-05','2024-01-06','2024-01-07'] │
+│ ['2024-02-15','2024-02-16','2024-02-17','2024-02-18','2024-02-19','2024-02-20','2024-02-21'] │
+└──────────────────────────────────────────────────────────────────────────────────────────────┘
+        )"},
+    };
+    FunctionDocumentation::IntroducedIn introduced_in = {26, 0};
```
rschu1ze: 26.5

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+            out_offsets[row_idx] = offset;
+        }
+
+        return ColumnArray::create(std::move(data_col), std::move(offsets_col));
+    }
+
+    template <typename T>
+    ColumnPtr executeGenericDays(const IColumn * start_col, const IColumn * end_col, Int64 step_value, size_t input_rows_count) const
+    {
+        return executeGeneric<T, true>(start_col, end_col, step_value, IntervalKind::Kind::Day, input_rows_count);
```
rschu1ze: Since `generate_date_array` isn't a critical function and in the interest of smaller binaries: can you please replace the template by a regular function argument?

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+    template <typename T>
+    ColumnPtr executeGenericCalendar(
+        const IColumn * start_col, const IColumn * end_col, Int64 step_value, IntervalKind interval_kind, size_t input_rows_count) const
+    {
+        return executeGeneric<T, false>(start_col, end_col, step_value, interval_kind, input_rows_count);
+    }
+
+    ColumnPtr executeImpl(const ColumnsWithTypeAndName & arguments, const DataTypePtr & result_type, size_t input_rows_count) const override
+    {
+        NullPresence null_presence = getNullPresense(arguments);
```
rschu1ze: If we remove the non-standard null handling, I don't think we need l. 322-324.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+        if (null_presence.has_null_constant)
+            return result_type->createColumnConstWithDefaultValue(input_rows_count);
+
+        DataTypePtr elem_type = checkAndGetDataType<DataTypeArray>(result_type.get())->getNestedType();
+        WhichDataType which(elem_type);
+
+        if (!which.isDateOrDate32())
```
rschu1ze: L. 327 - 335 are not needed, assuming that `getReturnTypeImpl` works correctly (which it does).

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+        {
+            throw Exception(
+                ErrorCodes::ILLEGAL_TYPE_OF_ARGUMENT,
+                "Illegal columns of arguments of function {}, the function only implemented for Date/Date32 types",
+                getName());
+        }
+
+        auto throw_if_null_value = [&](const ColumnWithTypeAndName & col)
```
rschu1ze: I also don't think we need l. 337-356 if we remove the non-standard null handling.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+                    col.column->getName(),
+                    getName());
+        };
+
+        for (size_t i = 0; i < arguments.size(); ++i)
+            throw_if_null_value(arguments[i]);
+
+        // Extract interval kind and constant step value.
+        IntervalKind interval_kind = IntervalKind::Kind::Day;
+        Int64 step_value = 1;
```
rschu1ze, line 238: just `step`?

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+        if (arguments.size() == 3)
+        {
+            if (const auto * interval_type = checkAndGetDataType<DataTypeInterval>(removeNullable(arguments[2].type).get()))
```
rschu1ze: `removeNullable` isn't needed (I think)

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+                interval_kind = interval_type->getKind();
+
+            step_value = assert_cast<const ColumnConst &>(*arguments[2].column).getValue<Int64>();
```
rschu1ze, line 243: The first and second arguments can be non-const, no? Here, we assert (`assert_cast`) that the second argument is const. Hmmm.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+        ColumnPtr start_col = castColumn(arguments[0], elem_type)->convertToFullColumnIfConst();
+        ColumnPtr end_col = castColumn(arguments[1], elem_type)->convertToFullColumnIfConst();
+
+        ColumnPtr res;
+        const bool is_day_based = (interval_kind == IntervalKind::Kind::Day || interval_kind == IntervalKind::Kind::Week);
+
+        if (is_day_based)
+        {
+            // Convert weeks to days once, outside any loop.
```
rschu1ze: I think we can dumb this down and omit the special handling for days and weeks. It only adds complexity for little benefit.

tests/queries/0_stateless/04141_generate_date_array.sql
```diff
@@ -0,0 +1,99 @@
+/* From the documentation */
```
rschu1ze: Please `--` instead of `/*` :-)

src/Functions/array/generate_date_array.cpp
EduardoGRocha: My concern was with `the underlying function may or may not be called for the rows containing NULLs. When called, arbitrary values are used instead of NULLs` 

So my understanding is that if the column is null, there is the chance that the function actually gets called with an arbitrary value. If these values are poorly chosen, for example setting to 1970-01-01, then calling `GENERATE_DATE_ARRAY(NULL, 2026-05-14)` would be equivalent to calling `GENERATE_DATE_ARRAY(1970-01-01, 2026-05-14)`, which would allocate a very large array and do quite some computations before masking this out

This seems to be the reason why range (`range.cpp`) set this value to `false`, although I am not sure

src/Core/Settings.cpp
EduardoGRocha: In that case should I reuse the setting for the range function?

src/Functions/array/generate_date_array.cpp
EduardoGRocha: I don't know how Clickhouse sees the trade-off of performance, binary size, and maintainability...But this special handling only adds two extra `if constexpr` and less than 50 LoCs while avoiding double iteration over the arrays and allowing for vectorization (although not implemented) for Days/Weeks (which I would expect to be the main use-case since 1 DAY is the default step)

src/Functions/array/generate_date_array.cpp
EduardoGRocha: This is for the case of using default step. The other arguments are extracted a bit later in the code. Updated the comment

src/Core/Settings.cpp
EduardoGRocha: I hard-coded the value of 500M entries. If we are pulling this value from any settings, then I would strongly favor using a separated setting for this function instead of sharing it with range

(notice that we use ranges so the individual arrays cannot be too large, however this 500M elements limit could still be achieved if the input column has multiple rows)

src/Functions/array/generate_date_array.cpp
EduardoGRocha: We assert_cast that the step, which is the third argument (index 2) is const. I think this is not strictly needed because we got `ColumnNumbers getArgumentsThatAreAlwaysConstant() const override { return {2}; }`, but I don't think there's harm in keeping it

src/Functions/array/generate_date_array.cpp
EduardoGRocha: Done. But that reduced the size of my debug build only by 10KB (on release, only 2KB)

I'm most concerned about the runtime check not allowing auto-vectorization.

It is difficult to say what is the impact of the change without benchmarking. But I feel like 10KB is something that can be afforded for each function (at 10K functions, that's still only 100MB more)

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,464 @@
...
+REGISTER_FUNCTION(GenerateDateArray)
+{
+    FunctionDocumentation::Description description = R"(
+Returns an array of dates from `start` to `end` (inclusive) incremented by `step`.
+
+- `start` and `end` accept `Date`, `Date32`, or a string literal that is implicitly cast to `Date` or `Date32` depending on the value.
+- When `start` and `end` have different types, the element type of the returned array is their least supertype (e.g. mixing `Date` and `Date32` yields `Array(Date32)`).
+- `step` must be a **constant** non-zero interval of kind `Day`, `Week`, `Month`, `Quarter`, or `Year`. It cannot be a column reference or a non-constant expression. When omitted, the default step is `INTERVAL 1 DAY`.
+- If `step` is positive the sequence is ascending. If `step` is negative the sequence is descending.
+- If `end` is unreachable in the given direction, an empty array is returned.
```
rschu1ze: Needs an example at the end of the sentence to explain what "unreachable" means. How can the database actually determine if the end is reachable? If "unreachable" means that the end is smaller/bigger than the start (assuming positive/negative step), then we don't need to document this anyways since it's a pathological case.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,464 @@
...
+- Returns `NULL` if any argument is of type `Nullable(Nothing)` (i.e. a bare `NULL` literal). Throws an exception if any argument has a runtime `NULL` value inside a `Nullable` column.
+- Throws an exception if the total number of elements produced exceeds 500 million.
```
rschu1ze: Let's make the threshold much much smaller, e.g. 1m.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,464 @@
...
+    )";
+    FunctionDocumentation::Syntax syntax = "generate_date_array(start, end[, step])";
+    FunctionDocumentation::Arguments arguments = {
+        {"start",
+         "The first date of the array. `Date`, `Date32`, or a string literal implicitly cast to `Date` or `Date32` depending on the "
+         "value.", {"String", "Date", "Date32", "NULL"}},
```
rschu1ze: Please remove `NULL` (ClickHouse docs generally don't talk much about Null handling)

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,464 @@
...
+    };
+    FunctionDocumentation::ReturnedValue returned_value
+        = {"Array of dates from `start` to `end` (inclusive) by `step`.", {"Array(Date)", "Array(Date32)", "NULL"}};
+    FunctionDocumentation::Examples examples = {
+        {"Daily step (default)", "SELECT generate_date_array('2024-01-01', '2024-01-05');", R"(
+┌─generate_date_array('2024-01-01', '2024-01-05')────────────────────┐
+│ ['2024-01-01','2024-01-02','2024-01-03','2024-01-04','2024-01-05'] │
+└────────────────────────────────────────────────────────────────────┘
+        )"},
+        {"Monthly step", "SELECT generate_date_array('2024-01-01', '2024-06-01', INTERVAL 1 MONTH);", R"(
```
rschu1ze: Let's use a more random looking start date: 2024-03-07.

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,464 @@
...
+┌─generate_date_array('2024-01-01', '2024-06-01', toIntervalMonth(1))─────────────┐
+│ ['2024-01-01','2024-02-01','2024-03-01','2024-04-01','2024-05-01','2024-06-01'] │
+└─────────────────────────────────────────────────────────────────────────────────┘
+        )"},
+        {"Descending (negative step)", "SELECT generate_date_array('2024-01-05', '2024-01-01', INTERVAL -1 DAY);", R"(
+┌─generate_date_array('2024-01-05', '2024-01-01', toIntervalDay(-1))─┐
+│ ['2024-01-05','2024-01-04','2024-01-03','2024-01-02','2024-01-01'] │
```
rschu1ze: Same comment here, what about 2024-03-02? (this will demonstrate the month wrap)

src/Functions/array/generate_date_array.cpp
```diff
@@ -0,0 +1,464 @@
...
+┌─generate_date_array(NULL, '2024-01-05')─┐
+│ NULL                                    │
+└─────────────────────────────────────────┘
+        )"},
+        {"Pre-1970 string dates are inferred as Date32", "SELECT generate_date_array('1960-01-01', '1960-01-08');", R"(
+┌─generate_date_array('1960-01-01', '1960-01-08')───────────────────────────────────────────────────────────┐
+│ ['1960-01-01','1960-01-02','1960-01-03','1960-01-04','1960-01-05','1960-01-06','1960-01-07','1960-01-08'] │
+└───────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+        )"},
+        {"Generate an array column from a subquery (start/end from columns, constant step)",
```
rschu1ze: The last example (UNION) is alright but looking at other function docs, it is also too composite/complex. You can remove it.

src/Functions/array/generate_date_array.cpp
rschu1ze: I see the concern but in practice, yet our own calling code would need to rewrite the input columns in a malicious way so that this can happen. I think we can afford to go the simpler route - let's do like I suggested above (there is generally too much special casing for Nulls in this code, we can throw as much as possible of it out)

PR #100141, opened 2026-03-19
Allow skipping local shard with missing table when `skip_unavailable_shards` is enabled

alexey-milovidov:
When a distributed query targets a table that doesn't exist on a local replica and there are no remote replicas to fall back to, the query would always fail with `UNKNOWN_TABLE` even when `skip_unavailable_shards` was enabled. Remote shards with missing tables could be skipped, but local shards could not, which was inconsistent.

Now, when `skip_unavailable_shards` is enabled and the table is absent on the local replica with no remote replicas configured, the shard is skipped instead of failing.

Closes https://github.com/ClickHouse/ClickHouse/issues/100134

### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Allow skipping local shard with missing table when `skip_unavailable_shards` is enabled.

### Version info
- Merged into: `26.5.1.344`

PR #98377, opened 2026-03-01
Forbid Tuple element name `null` to prevent subcolumn ambiguity

alexey-milovidov:
## Summary

- Forbid the Tuple element name `null` in `checkTupleNames` (`DataTypeTuple.cpp`) because it conflicts with the subcolumn name used for Nullable null maps, causing an exception in `SerializationNullable::enumerateStreams`
- Add test `04000_nullable_tuple_null_subcolumn_ambiguity` validating the new restriction
- Update existing test `01872_functions_to_subcolumns_analyzer` to rename `Tuple(null UInt32)` → `Tuple(n UInt32)`

Closes #96681
CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=98293&sha=f68ee0d3f6944d6418c2304965c72af57a2e9971&name_0=PR&name_1=AST%20fuzzer%20%28arm_asan%29
https://github.com/ClickHouse/ClickHouse/pull/98293

### Changelog category (leave one):
- Backward Incompatible Change

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Tuple element name `null` is now forbidden because it conflicts with the subcolumn name used for Nullable null maps, causing ambiguous subcolumn resolution.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### Version info
- Merged into: `26.5.1.235`

PR #103917, opened 2026-05-02
Allowlist `Tuple element name 'null' is reserved` in upgrade check

alexey-milovidov:
After PR #98377 (`Forbid Tuple element name `null` to prevent subcolumn ambiguity`), the new server rejects attaching any pre-existing table with a column of type `Tuple(null ...)`. The stateless test `01872_functions_to_subcolumns_analyzer.sql` historically created such a table (`t_tuple_null` with `Tuple(null UInt32)`), and after the stress flow upgrades the server, the new build logs a `BAD_ARGUMENTS` error from `AsyncLoader::worker` while trying to attach it. The validation is working as intended; the failure surfaces in the `Upgrade check` job under the test `Error message in clickhouse-server.log (see upgrade_error_messages.txt)`.

Add the message to the upgrade-check error allowlist so the expected post-upgrade rejection does not fail the check.

Example failing run (reproduced on PR #100300):
https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100300&sha=a93523dfa84eebfb7bd5b2f25596f01e2e18ab2e&name_0=PR&name_1=Upgrade%20check%20%28amd_release%29

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
...

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #103966, opened 2026-05-03
Filter `Tuple element name 'null' is reserved` in upgrade check

alexey-milovidov:
The upgrade test in CI installs a previous release binary, runs the test suite, then upgrades to the new server build and restarts. Tables created on the old binary with columns like `Nullable(Tuple(`null` UInt32))` fail to attach on the new binary because PR #98377 forbade Tuple element name `null` to avoid ambiguity with the Nullable null-map subcolumn name.

This is a known consequence of the backward-incompatible change, not a regression introduced by any in-flight PR. Add the exact reserved-name phrase to the `upgrade_runner.sh` error filter so unrelated PRs no longer fail the upgrade check on this message.

Observed on PR #100141: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100141&sha=ca91e01ef6f3c799bddb2176485608a8c9eaf151&name_0=PR&name_1=Upgrade%20check%20%28amd_release%29

PR: https://github.com/ClickHouse/ClickHouse/pull/100141

...


alexey-milovidov:
Duplicate #103917

PR #103980, opened 2026-05-04
Remove and forbid build-time CMake checks

alexey-milovidov:
The compiler is fixed (Clang with a known minimum version, see `cmake/tools.cmake`) and so are the vendored sysroots under `contrib/sysroot/`, so probes like `check_cxx_source_compiles`, `check_function_exists`, `check_include_file`, `try_compile`, etc. are unnecessary - the answer is known statically. They also slow down configure, are non-deterministic, and are a source of build divergence between machines. The only legitimate use was detecting available compiler warnings, which we now do by hard-coding the knowledge against `CMAKE_CXX_COMPILER_VERSION`.

This PR:

- Rewrites `cmake/add_warning.cmake` so that `add_warning` / `no_warning` / `target_add_warning` / `target_no_warning` just append the flag. Previously they used `check_cxx_compiler_flag` / `check_c_compiler_flag` to feature-test every `-W` flag at configure time.
- Hard-codes clang-version-specific warnings in `cmake/warnings.cmake`. The `lifetime-safety-*` warnings only exist in clang 23+, so they are now gated on `CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 23` instead of being detected.
- Deletes the empty `cmake/check_flags.cmake` stub.
- Adds `cmake/block_build_time_checks.cmake`, included before `add_subdirectory(contrib)`, which installs `FATAL_ERROR` overrides for every `check_*` macro/function plus `try_compile` / `try_run`. Even when a `CheckCXXCompilerFlag`-style module re-defines its wrapper, the call still funnels through the overridden `try_compile` and aborts with a clear message pointing at this file.
- Removes the two real probe call-sites in contrib `*-cmake/`:
  - `contrib/mongo-c-driver-cmake`: `test_big_endian(BSON_BIG_ENDIAN)` -> static check on `ARCH_S390X` (the only big-endian arch we support).
  - `contrib/isa-l-cmake`: `check_language(ASM_NASM)` -> `set(CMAKE_ASM_NASM_COMPILER \${NASM_PATH})` using the path already found by `find_program`.

Where compiler-version-specific behaviour is genuinely required in the future, gate it on `CMAKE_CXX_COMPILER_VERSION` instead of probing.

Removed and forbade build-time CMake checks (`check_*`, `try_compile`, `try_run`). The compiler and toolchain are fixed, so feature detection at configure time is unnecessary and is now blocked project-wide; any version-specific behaviour must be gated on `CMAKE_CXX_COMPILER_VERSION` explicitly.

### Version info
- Merged into: `26.5.1.432`


Algunenano (approved):

alexey-milovidov (commented):
Looks good.

Algunenano (approved):

PR #62125, opened 2024-04-01
Read-in-order Optimization With Virtual Row

jsc0218:
related to [this](https://github.com/ClickHouse/ClickHouse/issues/42662).

### Changelog category (leave one):
- Improvement

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
read-in-order optimization via generating virtual rows, so less data would be read during merge sort especially useful when multiple parts exist. 

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

> Information about CI checks: https://clickhouse.com/docs/en/development/continuous-integration/

---
### Modify your CI run:
**NOTE:** If your merge the PR with modified CI you **MUST KNOW** what you are doing
**NOTE:** Checked options will be applied if set before CI RunConfig/PrepareRunConfig step

#### Include tests (required builds will be added automatically):
- [ ]  Fast test
- [ ]  Integration Tests
- [ ]  Stateless tests
- [ ]  Stateful tests
- [ ]  Unit tests
- [ ]  Performance tests
- [ ]  All with ASAN
- [ ]  All with TSAN
- [ ]  All with Analyzer
- [ ]  Add your option here

#### Exclude tests:
- [ ]  Fast test
- [ ]  Integration Tests
- [ ]  Stateless tests
- [ ]  Stateful tests
- [ ]  Performance tests
- [ ]  All with ASAN
- [ ]  All with TSAN
- [ ]  All with MSAN
- [ ]  All with UBSAN
- [ ]  All with Coverage
- [ ]  All with Aarch64
- [ ]  Add your option here

#### Extra options:
- [ ]  do not test (only style check)
- [ ]  disable merge-commit (no merge from master before tests)
- [ ]  disable CI cache (job reuse)

#### Only specified batches in multi-batch jobs:
- [ ]  1
- [ ]  2
- [ ]  3
- [ ]  4

PR #101781, opened 2026-04-04
Add test: No test for merge handling in readonly MergeTree refresh (refreshDataParts covers parts by Transaction::commit)

clickgapai:
_Found via ClickGap automated review. Please close or comment if this is incorrect or needs adjustment._

_Retrospective finding from a historical scan of [PR #79033](https://github.com/ClickHouse/ClickHouse/pull/79033) (merged 2025-04-11). Confirmed on current codebase — close with a note if already fixed._

Adds test coverage for 1 untested code path(s), found during automated review of [PR #79033](https://github.com/ClickHouse/ClickHouse/pull/79033).

**1. No test for merge handling in readonly MergeTree refresh (refreshDataParts covers parts by Transaction::commit)**
  `src/Storages/MergeTree/MergeTreeData.cpp:2660`, `src/Storages/MergeTree/MergeTreeData.cpp:2688`
  **Why this test:** MergeTreeData::refreshDataParts (MergeTreeData.cpp:2620) loads new parts and adds them via Transaction::commit which calls getActivePartsToReplace to mark old covered parts as Outdated. This merge han
  **What's unique:** PR's test 03362 covers only basic single INSERT + read. This test covers the merge handling path: multiple inserts → writer merge → reader refresh correctly handling covered parts without duplicates.

### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Not applicable — test-only change.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

---
_Found during review of [PR #79033](https://github.com/ClickHouse/ClickHouse/pull/79033)._

PR #103984, opened 2026-05-04
Add test: FINAL-disable guard for virtual row optimization untested with `SELECT FROM ... FINAL`

clickgapai:

_Retrospective finding from a historical scan of [PR #62125](https://github.com/ClickHouse/ClickHouse/pull/62125) (merged 2024-11-11). Confirmed on current codebase — close with a note if already fixed._

Adds test coverage for 1 untested code path(s), found during automated review of [PR #62125](https://github.com/ClickHouse/ClickHouse/pull/62125).
 That PR PR #62125 introduces a new read-in-order optimization called 'virtual row'. When `optimize_read_in_order=1` and the new setting `read_in_order_use_virtual_row=1` (default off) are both enabled, the planner inserts a synthetic 'virtual row' (containing the first PK value of each part) into the merge 

**1. FINAL-disable guard for virtual row optimization untested with `SELECT FROM ... FINAL`**
  `src/Processors/QueryPlan/ReadFromMergeTree.cpp:2896`
  **Risk:** Call chain: `optimizeReadInOrder` builds `virtual_row_conversion` for any ORDER BY matching the storage sort key → calls `ReadFromMergeTree::setVirtualRowConversions` → guard at `ReadFromMergeTree.cpp
  **What's unique vs PR tests:** The PR's own three tests (`03031_read_in_order_optimization_with_virtual_row*.sql`) only use plain `MergeTree` and never `SELECT FROM <tab> FINAL`. Follow-up tests (`03809`, `04029`, `04045`, `04049`,
  **Tags:** `-- Tags: no-random-merge-tree-settings` — `no-random-merge-tree-settings`: Random MergeTree settings can change index granularity and merge behavior, which would shift how many parts exist after `SYSTEM STOP MERGES` and could change virtual-row triggering — making count()/sum() outputs nondeterministic.


clickgapai:
CI check `Stress test (arm_release)` failed, but the failure does not reference this PR's test (`04201_virtual_row_with_select_final`).

This PR is test-only; the bot will not retrigger CI or modify the test to chase a failure that appears unrelated. If you believe the failure is related to this test, please comment with details and the bot will re-evaluate.

clickgapai:
This PR has accumulated 5 CI failures that don't reference its test, so the bot has not attempted any fix. Closing — please re-run CI or comment if you'd like the bot to look again.

clickgapai:
The bot has tried to address CI failures on this PR 5 times without a successful push (logs unavailable / context insufficient / plan rejected). Stopping further auto-fix attempts on `Stress test (arm_release)` — needs human review.

clickgapai:
CI failure in `Stress test (arm_release)` appears to be unrelated breakage and is not related to this change. Let me know if you'd like me to merge master and retrigger CI.

clickgapai:
CI failure in `Config Workflow` appears to be unrelated breakage and is not related to this change. Let me know if you'd like me to merge master and retrigger CI.

zlareb1:
@groeneai can you take this up and open a new PR?

groeneai:
Taking this up. Triaged against master `2850e469dcac84b` before answering.

Verified: `src/Processors/QueryPlan/ReadFromMergeTree.cpp:2896` still present; test passes and output matches the reference.

It will be re-filed as part of an area-grouped PR rather than its own PR, to keep reviewer load down (26 in this batch plus 18 from the earlier one today). Full triage and the grouping plan: https://github.com/ClickHouse/ClickHouse/pull/101781#issuecomment-5244746014

groeneai:
Re-filed as **#114305**, the MergeTree/index group, covering this gap together with three others from the same batch.

Adopted as `04489_virtual_row_disabled_with_final`, but the test as written never reached the guard. I mutated `setVirtualRowConversions` (`ReadFromMergeTree.cpp:3323`) to drop the `isQueryWithFinal()` check and the test stayed green, so I instrumented the function: it was called zero times, with a positive control proving the probe was live. `EXPLAIN PLAN` showed why - the outer `count()` strips the subquery's `ORDER BY`, so no sorting step is built and `optimizeReadInOrder` never runs; the non-FINAL row was answered by `_minmax_count_projection` without reading the table at all.

Fixed by adding a `LIMIT` so the read-in-order plan is actually built, and by pinning liveness directly: the plan must carry no virtual row conversion for `FINAL` while the same fixture without `FINAL` does build one. It now reddens on that mutation, `8192 16384` to `0 0`.

PR #103029, opened 2026-04-17
[CI Fix] Fix not-ready Set when filter depends on the left-side column

yariks5s:
## Linked issues
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1692
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1760

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fixes a case found by CI where a not-ready set was passed when the filter depends on the left-side column. Closes https://github.com/ClickHouse/ClickHouse/issues/102966

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.58`
- Backported to: `26.4.3.31`

PR #103987, opened 2026-05-04
Add unit test for not-ready Set detection in filterResultForMatchedRows

tiandiwonder:

Not for changelog (test improvement)

- [x] Documentation is not required (test improvement only)

---

## Summary

This PR adds regression protection for the fix in PR #103029, which addressed incomplete not-ready Set detection in `filterResultForMatchedRows`.

### Background

PR #103029 fixed a bug where `filterResultForMatchedRows` only checked `filter_dag` for not-ready Sets, but evaluated `combined_dag = merge(pre_actions_dag, filter_dag)`. If a not-ready Set was in `pre_actions_dag`, the check would be bypassed, causing `FunctionIn` to throw a `LOGICAL_ERROR` exception when encountering the not-ready Set during evaluation.

### This PR

Adds three unit tests in `src/Processors/QueryPlan/Optimizations/tests/gtest_convert_any_join.cpp`:

1. **RealInOnNotReadySetWithoutCombinedGuardDiesOrThrows** (death test)
   - Verifies that without the fix, real `FunctionIn` encounters not-ready Set and throws exception
   - Uses `EXPECT_DEATH` to prove the exception path exists

2. **PreActionsDagNotReadySetReturnsBeforeEvaluatingFilter** ⭐ (path-sensitive regression test)
   - Uses probe mechanism to verify early return before evaluation
   - If fix is reverted, probe would be triggered → test fails
   - Provides strong regression protection

3. **FixedFilterResultForMatchedRows_ReturnsUnknown** (positive test)
   - Verifies the fix returns `UNKNOWN` when `pre_actions_dag` contains not-ready Set

### Implementation Details

- Exposes `filterResultForMatchedRows` in `Utils.h` for testing (follows pattern from commit 70cc74cc637 which moved `filterResultForNotMatchedRows` from anonymous namespace to Utils.h)
- Adds `MockNotReadyFutureSet` test fixture
- Adds `FunctionTestNotReadySetProbe` with disabled constant folding for path-sensitive testing

### Related

- PR #103029: https://github.com/ClickHouse/ClickHouse/pull/103029 (original fix, already merged)

### Version info
- Merged into: `26.7.1.1051` (included in `26.7` and later)


yariks5s (approved):

clickgapai:
Hi @tiandiwonder @yariks5s — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: The body claims THREE unit tests are added, including a death test `RealInOnNotReadySetWithoutCombinedGuardDiesOrThrows` using `EXPECT_DEATH`. The actual diff adds only TWO gtest cases (`PreActionsDagNotReadySetReturnsBeforeEvaluatingFilter` and `FixedFilterResultForMatchedRows_ReturnsUnknown`); no death test is present in gtest_convert_any_join.cpp. This is a description-vs-implementation discrepancy in the PR body (not a code bug).
- Suggested addition:

  > Update the description to list only the two tests actually added, or add the missing death test. Remove the reference to `RealInOnNotReadySetWithoutCombinedGuardDiesOrThrows` if it is not included.

PR #74144, opened 2025-01-03
Fix security issue to prevent SQL injection in postgresql and sqlite table functions

pamarcos:
Fix security issue to prevent SQL injection in postgresql and sqlite table functions

At first I thought we should sanitize and validate the parameters to check whether the database and table names comply with PostgreSQL convention. However, turns out we can fix this in an easier way: properly escaping single quotes for PostgreSQL and SQLite.

Closes https://github.com/ClickHouse/clickhouse-private/issues/21742
Closes https://github.com/ClickHouse/ClickHouse/issues/74610

### Changelog category (leave one):
- Critical Bug Fix (crash, data loss, RBAC) or LOGICAL_ERROR

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Fix security issue to prevent SQL injection in postgresql and sqlite table functions

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

> Information about CI checks: https://clickhouse.com/docs/en/development/continuous-integration/

#### CI Settings (Only check the boxes if you know what you are doing)

All builds in Builds_1 and Builds_2 stages are always mandatory
and will run independently of the checks below:

- [ ]  Allow: All Required Checks
- [ ]  Allow: Stateless tests
- [ ]  Allow: Stateful tests
- [ ]  Allow: Integration Tests
- [ ]  Allow: Performance tests
- [ ]  Allow: All Builds
- [ ]  Allow: batch 1, 2 for multi-batch jobs
- [ ]  Allow: batch 3, 4, 5, 6 for multi-batch jobs
---
- [ ]  Exclude: Style check
- [ ]  Exclude: Fast test
- [ ]  Exclude: All with ASAN
- [ ]  Exclude: All with TSAN, MSAN, UBSAN, Coverage
- [ ]  Exclude: All with aarch64
- [ ]  Exclude: All with release
- [ ]  Exclude: All with debug
---
- [ ]  Run only fuzzers related jobs (libFuzzer fuzzers, AST fuzzers, etc.)
- [ ]  Exclude: AST fuzzers
---
- [ ]  Do not test
- [ ]  Woolen Wolfdog
- [ ]  Upload binaries for special builds
- [ ]  Disable merge-commit
- [ ]  Disable CI cache

PR #104009, opened 2026-05-04
Fix SQL injection in ExternalQueryBuilder via incorrect string escaping

tiandiwonder:
Fix SQL injection in `ExternalQueryBuilder` caused by incorrect string literal escaping when building `WHERE` clauses for external dictionary sources (PostgreSQL, Cassandra, XDBC).

`ExternalQueryBuilder` used the default `FormatSettings` with `escape_quote_with_quote = false`, so `serializeTextQuoted` produced backslash-escaped string literals (`\'`) in the generated SQL. PostgreSQL (with `standard_conforming_strings = on` by default since 9.1) treats backslash as a literal character — not an escape sequence — allowing a caller to break out of a string literal and inject arbitrary SQL by passing a key containing a single quote to `dictGet` on a PostgreSQL or XDBC dictionary source. The same vulnerability applies to Cassandra, which also follows the SQL standard.

The same class of bug was fixed for the PostgreSQL table function in PR #74144, but the dictionary code path through `ExternalQueryBuilder` was missed.

**Fix:** in `ExternalQueryBuilder`'s constructor, set `escape_quote_with_quote = true` only when `quoting_style == DoubleQuotes` (PostgreSQL, Cassandra, XDBC SQL). For ClickHouse and MySQL (`Backticks` / `BackticksMySQL`), keep the default backslash escaping (`\'`, `\\`), which both backends parse natively. This per-backend split addresses reviewer feedback that the original "always quote-doubling" fix would have caused ClickHouse-on-ClickHouse dictionaries to corrupt string values containing backslashes when those values are serialized back through `TabSeparated`.

A new gtest `gtest_external_query_builder.cpp` covers both branches: `DoubleQuotes` doubles single quotes and leaves backslashes intact; `Backticks` uses `\'` and doubles backslashes.

Reported in: https://github.com/ClickHouse/clickhouse-private/issues/56142

Fixed SQL injection vulnerability in PostgreSQL, Cassandra and XDBC dictionary sources: string keys containing single quotes were escaped with `\'` (backslash), which these backends treat as a literal backslash rather than an escape sequence, allowing arbitrary SQL to be injected into dictionary lookup queries. `ExternalQueryBuilder` now emits SQL-standard `''` escaping for those backends; ClickHouse and MySQL dictionary sources continue to use backslash escaping.

### Version info
- Merged into: `26.5.1.727` (included in `26.5` and later)
- Backported to: `26.4.3.34`, `26.3.11.26`, `25.8.29.17`


alexey-milovidov:
> MySQL and ClickHouse accept both styles, so there is no regression risk for those backends.

Let's say we have a string with the following contents: `\''`.
In ClickHouse-style escaping, it will be written as `'\\\'\''`.
In Postgres-style escaping, it will be written as `'\'''''`.

Now let's see what happens if we take the string with Postgres-style escaping and put it in a ClickHouse query. It will not parse.

alexey-milovidov (changes requested):
.

tiandiwonder:
@alexey-milovidov Thank you for the example — you're correct. The fix incorrectly applied `''`-style escaping to all backends including ClickHouse. ClickHouse's SQL parser treats `\`\ as an escape character by default, so a backslash in a `''`-escaped string literal would be misinterpreted (e.g., `\b` becomes a backspace, `\'` causes an unterminated literal). Fixed in commit `1459070e7cf` — `ExternalQueryBuilder` now uses SQL-standard `''` escaping only when `quoting_style == IdentifierQuotingStyle::DoubleQuotes` (PostgreSQL, XDBC-PostgreSQL backends), and keeps backslash escaping for ClickHouse and MySQL. Added a regression test (`04141`) covering ClickHouse dictionary lookups with backslash and single-quote keys.

alexey-milovidov:
Removed `tests/queries/0_stateless/04141_dict_clickhouse_source_backslash_key.sql` in `ee42acfa096` to make the **Bugfix validation (functional tests)** check pass.

The test was a regression test for the ClickHouse-source dictionary path, which uses `IdentifierQuotingStyle::Backticks` and is unaffected by the original SQL injection. Because the test passed on master (i.e. it did not reproduce the bug), the bugfix validation correctly reported "Failed to reproduce the bug".

Per-backend escaping behavior is still covered by:
- `gtest_external_query_builder.cpp` for the unit-level `DoubleQuotes` and `Backticks` paths (both single quote and backslash);
- the new PostgreSQL integration test `test_postgres_dict_complex_key_with_single_quote`, which catches the original injection on master;
- the new MySQL integration test `test_mysql_dict_complex_key_with_special_chars`, which exercises backslash escaping.

The other CI failures are unrelated to this PR:
- `Unit tests (msan, function_prop_fuzzer)` — known flaky failure tracked in #104877 (`FunctionsStress.stress`);
- `Stateless tests (amd_msan, WasmEdge, parallel, 1/2)` — `04033_tpc_ds_*` tests are widely flaky across many sanitizers/configurations on master.

alexey-milovidov (approved):

PR #104015, opened 2026-05-04
Revert "Increase value-profiling counter density in PGO toolchain stage1 build"

leshikus:
Reverts ClickHouse/ClickHouse#103788

### Version info
- Merged into: `26.5.1.264`


leshikus:
Reverting based on discussion with @shankar-iyer

PR #100181, opened 2026-03-20
ci: run integration flaky/targeted check in module scope with --dist=each

fm4v:
Integration tests use \`module\`-scoped fixtures for cluster setup/teardown. The flaky check was using \`pytest-repeat\` with \`--count 3 --random-order\`, repeating each individual test function before moving to the next. This causes repetitions to be inconsistent with how tests run in normal CI (module scope).

This change replaces \`pytest-repeat\` for flaky/targeted checks with \`--dist=each\`: all test modules are sent to every xdist worker.

For this to work correctly, three bugs in \`ClickHouseCluster\` were fixed:

1. **\`cluster.start()\` re-entry after shutdown** — when \`--dist=each\` causes tests from different modules to interleave on the same xdist worker, \`scope="module"\` fixtures are torn down and re-set-up multiple times within the same session. \`cluster.shutdown()\` does not remove \`instances_dir\`, and \`ClickHouseCluster.__init__\` only cleans it up at Python import time. On the second \`cluster.start()\`, \`create_dir()\` would fail with \`FileExistsError\`. Fixed by removing and recreating \`instances_dir\` at the top of \`cluster.start()\` when the cluster is not currently up.

2. **Premature filesystem mutation before \`is_up\` guard** — the cleanup in \`cluster.start()\` must happen only when the cluster is down. Moving \`shutil.rmtree\` after the \`if self.is_up: return\` guard prevents destroying the live \`instances_dir\` (including the \`.env\` file) when \`start()\` is called on an already-running cluster (e.g. multiple module-scoped fixtures sharing the same cluster object).

3. **Stale \`ipv6_address\` on second \`create_dir()\`** — \`cluster.start()\` sets \`instance.ipv6_address\` dynamically from the running container after startup. \`cluster.shutdown()\` reset \`ip_address\` to \`None\` but not \`ipv6_address\`. On the second \`create_dir()\` the stale address was inserted into the docker-compose.yml unquoted (\`ipv6_address: 2001:3984:3989::5\`), which Docker Compose v2 rejects with \`ipv6_address must be a string\`. Fixed by resetting \`ipv6_address = None\` in \`shutdown()\` and quoting the value in the template.

For **sequential modules** (those that require CPU isolation, e.g. \`test_server_overload\`): they cannot run in parallel, so they are executed in a loop at least 4 times, at most \`workers\` times, breaking early on the first failure.

Comments were added to several test files to trigger the flaky check CI job.

### Changelog category (leave one):
[...]

PR #104020, opened 2026-05-04
Fix flaky test_s3_cluster: scope generated CSV dir per xdist worker

groeneai:
The `started_cluster` fixture in `tests/integration/test_s3_cluster/test.py`
writes 100 CSV files to `tests/integration/test_s3_cluster/data/generated/`,
uploads them to MinIO, then deletes the directory in teardown.

Under `Integration tests (amd_asan_ubsan, targeted)` the runner uses
pytest-xdist with `--dist=each`, so every test goes to every worker and the
module-scoped fixture runs concurrently in multiple processes. Their MinIO
containers are isolated by `ClickHouseCluster` (it appends
`PYTEST_XDIST_WORKER` to `project_name`), but the host filesystem path was
shared. One worker's teardown could `shutil.rmtree(...)` the directory
while another worker was still iterating over it inside the upload loop in
`create_buckets_s3`, surfacing as `FileNotFoundError` on
`data/generated/file_0.csv` during fixture setup and failing every test in
the module via fixture error.

Pattern matches Issue #102238 (open since 2026-04-09) and CIDB shows ~21–23
unrelated PRs hitting the same setup error in the past 30 days for tests
across the suite (`test_wrong_cluster`, `test_select_all`, `test_count`,
`test_union_all`, `test_hive_partitioning`, etc.) — exactly the symptom
expected when a module-scoped fixture errors and propagates to all 18
tests.

Earlier PR #102144 added `ignore_errors=True` to the teardown `rmtree` to
silence the cleanup-side variant of this race; this change addresses the
upload-side variant by removing the shared resource.

This change scopes the host directory by `PYTEST_XDIST_WORKER` so
concurrent workers no longer share filesystem state. The MinIO bucket key
still uploads under the canonical `data/generated/` prefix, so all test
queries (`SELECT … FROM s3('http://minio1:9001/root/data/generated/*'…)`)
are unchanged. Also decouple the host path from the bucket object key
(previously a single `f"data/generated/file_{n}.csv"` was used as both),
drop the module-level mutable `S3_DATA` list, and guard
`cluster.shutdown()` in the teardown against the pre-existing case where
the `ClickHouseCluster(__file__)` constructor raises before `cluster` gets
bound.

Race deterministically reproduced with a minimal Python harness
(`tmp/race_repro.py`, 2 threads, shared dir): 20/20 trials hit
`FileNotFoundError` without the fix; 0/20 with per-worker dirs.

Closes #102238

...

### Version info
- Merged into: `26.5.1.291`

### DIFF
--- a/tests/integration/test_s3_cluster/test.py
+++ b/tests/integration/test_s3_cluster/test.py
@@ -17,7 +17,9 @@ logging.getLogger().setLevel(logging.INFO)
 logging.getLogger().addHandler(logging.StreamHandler())
 
 SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
-S3_DATA = [
+
+# Base CSV files committed to git. Immutable on disk, safe to read concurrently.
+S3_BASE_FILES = [
     "data/clickhouse/part1.csv",
     "data/clickhouse/part123.csv",
     "data/database/part2.csv",
@@ -25,14 +27,48 @@ S3_DATA = [
 ]
 
 
+def _generated_host_dir():
+    """Per-xdist-worker host directory for the dynamically generated CSV files.
+
+    Concurrent xdist workers running this module (especially under
+    `--dist=each`, used by the targeted/flaky integration jobs) all execute
+    this fixture in parallel. Their MinIO containers are isolated by
+    `ClickHouseCluster` (it appends `PYTEST_XDIST_WORKER` to `project_name`),
+    but the host filesystem path is not. With a shared host directory, one
+    worker's teardown (`shutil.rmtree(...)`) deletes files while another
+    worker is still uploading them, producing setup-phase
+    `FileNotFoundError` on `data/generated/file_*.csv` and failing every
+    test in the module via fixture error.
+
+    The MinIO bucket key for tests stays `data/generated/file_N.csv` (each
+    worker uploads to its own MinIO instance); only the host write/read path
+    is scoped per worker.
+    """
+    worker_id = os.environ.get("PYTEST_XDIST_WORKER", "")
+    suffix = f"_{worker_id}" if worker_id else ""
+    return os.path.join(SCRIPT_DIR, f"data/generated{suffix}")
+
+
 def create_buckets_s3(cluster):
     minio = cluster.minio_client
 
+    host_dir = _generated_host_dir()
+    os.makedirs(host_dir, exist_ok=True)
+
+    files_to_upload = []  # list of (bucket_object_key, host_file_path)
+
+    # Base files: committed in git, read from the shared SCRIPT_DIR location.
+    for relative_path in S3_BASE_FILES:
+        files_to_upload.append(
+            (relative_path, os.path.join(SCRIPT_DIR, relative_path))
+        )
+
+    # Generated files: written into a per-worker host directory but uploaded
+    # to MinIO under the canonical `data/generated/` prefix that tests query.
     for file_number in range(100):
-        file_name = f"data/generated/file_{file_number}.csv"
-        os.makedirs(os.path.join(SCRIPT_DIR, "data/generated/"), exist_ok=True)
-        S3_DATA.append(file_name)
-        with open(os.path.join(SCRIPT_DIR, file_name), "w+", encoding="utf-8") as f:
+        bucket_object_key = f"data/generated/file_{file_number}.csv"
+        host_path = os.path.join(host_dir, f"file_{file_number}.csv")
+        with open(host_path, "w+", encoding="utf-8") as f:
             # a String, b UInt64
             data = []
 
@@ -44,12 +80,13 @@ def create_buckets_s3(cluster):
 
             writer = csv.writer(f)
             writer.writerows(data)
+        files_to_upload.append((bucket_object_key, host_path))
 
-    for file in S3_DATA:
+    for bucket_object_key, host_path in files_to_upload:
         minio.fput_object(
             bucket_name=cluster.minio_bucket,
-            object_name=file,
-            file_path=os.path.join(SCRIPT_DIR, file),
+            object_name=bucket_object_key,
+            file_path=host_path,
         )
     for obj in minio.list_objects(cluster.minio_bucket, recursive=True):
         print(obj.object_name)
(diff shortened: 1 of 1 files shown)


groeneai:
**Session ID:** cron:clickhouse-ci-task-worker:20260504-161500

### Pre-PR validation gate

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | ✅ YES. `tmp/race_repro.py` (committed only as a debugging artefact, not part of this PR) reproduces the race in 20/20 runs with shared dir, 0/20 with per-worker dir. The harness models the actual fixture sequence: two threads each `write_files()` → `upload_files()` → `teardown()`, racing on a shared host directory. |
| b | **Root cause explained?** | ✅ YES. Targeted/flaky integration jobs run `pytest -n W --dist=each`, fanning every test out to every xdist worker. The module-scoped `started_cluster` fixture executes once per worker process. `ClickHouseCluster(__file__)` isolates Docker resources by suffixing `project_name` and `instances_dir_name` with `PYTEST_XDIST_WORKER`, but `tests/integration/test_s3_cluster/data/generated/` is constructed from `os.path.dirname(os.path.realpath(__file__))` and is identical across workers. When worker A finishes its 18 tests and runs `shutil.rmtree(host_dir)` while worker B is still iterating `for file in S3_DATA: minio.fput_object(...)`, B's next `os.stat(host_path)` (called by `minio.fput_object` at `minio/api.py:1052`) raises `FileNotFoundError` — surfacing as a setup-phase fixture error that fails all 18 tests in B's run. |
| c | **Fix matches root cause?** | ✅ YES. Eliminates the shared resource by scoping the host directory per `PYTEST_XDIST_WORKER`. Each worker writes to and tears down its own `data/generated_<worker_id>/` directory; teardowns can no longer race against another worker's upload loop. The MinIO bucket key stays `data/generated/file_N.csv` (each worker has its own MinIO container, so the bucket key is already isolated), preserving every existing test query. |
| d | **Test intent preserved?** | ✅ YES. All 18 test bodies are untouched. Every `s3()` / `s3Cluster()` / `url()` table function call against `http://minio1:9001/root/data/generated/*` continues to read the same data. The 4 base CSVs (`data/clickhouse/`, `data/database/`) are still read from their committed location. The number of generated files (100), the per-file row count formula (`100 + file_number`), and the data shape are unchanged. |
| e | **Both directions demonstrated?** | ✅ YES. Race reproducer: **20/20 failures** with a shared host dir (the bug), **0/20 failures** with per-worker dirs (the fix). |

### Self-review against `.claude/skills/review/SKILL.md`

- **Deletion logging:** N/A — only test-side `shutil.rmtree` of a per-worker directory; no production data deletion.
- **Serialization versioning:** N/A — no format changes.
- **Core-area scrutiny:** N/A — test infrastructure only.
- **No test removal:** ✅ — no tests removed or weakened. Test bodies untouched; only fixture setup/teardown changed.
- **Experimental gate / magic constants / backward compat / safe rollout / compile time / no large files:** N/A — Python test fixture only.
- **PR metadata:** Category `CI Fix or Improvement (changelog entry is not required)`, no entry required by the template.
- **C++ risk self-check:** N/A — no C++ changes.
- **Trust boundary:** N/A.

### Notes

- Earlier PR #102144 (azat, merged 2026-04-09) added `ignore_errors=True` to the teardown `rmtree` to silence the cleanup-side variant of this race. This change addresses the upload-side variant by removing the shared resource entirely.
- PR #100627 (merged 2026-03-25) added the `is_targeted_check` module-sort fix to prevent fixture re-entry from CIDB-ordered test interleaving in a single worker. That fix and this one are complementary: #100627 prevents in-process re-entry, this one prevents cross-process races on the host filesystem.

groeneai:
cc @azat @maxknv — could you review this? It scopes test_s3_cluster's host-side `data/generated/` dir by `PYTEST_XDIST_WORKER` so that under `--dist=each` (used by the targeted integration job) one worker's teardown can no longer rmtree files mid-upload in another worker, which is the FileNotFoundError pattern from #102238 hitting ~21 unrelated PRs/30d. Adjacent to your earlier #102144 cleanup-side patch (azat) and the targeted-check sort fix (maxknv); test bodies and MinIO bucket keys unchanged.

maxknv:
@fm4v, what was the idea behind [--dist=each](https://github.com/ClickHouse/ClickHouse/commit/7fe3d4643e599d298084699aebc06fb0dc19dbed)? I think we have more tests that cannot run concurrently with themselves.

fm4v:
@maxknv it's described in https://github.com/ClickHouse/ClickHouse/pull/100181

Before (https://github.com/ClickHouse/ClickHouse/pull/66986): flaky check used `pytest-repeat --count 3 --random-order`. That repeats at function scope inside a single worker against the same module fixture (cluster started once, kept running) so it never exercises fixture lifecycle, where most flakes live. It itself requires: "It is important to cleanup tables and other entities in the end of a test case to pass" but the corresponding test rewrites were never done.

After (https://github.com/ClickHouse/ClickHouse/pull/100181): `--dist=each` runs the full module list on every xdist worker in parallel (module-scope parallel). Workers get isolated Docker clusters but still share the host filesystem and network namespace, so a module must now be runnable in parallel with itself on the same host. 
Same "rewrite the test corpus" debt as #66986, just surfaced differently and easier to fix than rewriting the logic of every integration test. 

Overall it stabilized the flaky check:
```
Weekly trend — Integration tests (amd_asan{,_ubsan}, flaky)
week        fail%   runs
2026-01-11  25.0    168
2026-01-18  23.7    139
2026-01-25  26.8    183
2026-02-01  23.8    147
2026-02-08  23.2    194
2026-02-15  23.3    116
2026-02-22  16.7    120
2026-03-01  14.4    125
2026-03-08  34.9    146   ← pre-change baseline
2026-03-15  20.4    225   ← --dist=each lands (#100181)
2026-03-22  24.5    220   ← fallout, fixed by #100627
2026-03-29  12.0    166
2026-04-05  14.5    255
2026-04-12   8.0    201
2026-04-19  10.2    118
2026-04-26   5.9    185
```

alexey-milovidov (approved):
The general idea is reasonable.

PR #102689, opened 2026-04-14
Fix columns_substreams.txt corruption during column rename in some cases

Avogar:
### Changelog category (leave one):
- Critical Bug Fix (crash, data loss, RBAC)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix columns_substreams.txt corruption during column rename in some cases. Closes https://github.com/ClickHouse/ClickHouse/issues/102259

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Backported to: `26.4.1.1140`, `26.3.10.61`, `26.2.19.13`

Issue #102259, opened 2026-04-09
`getFileNameForRenamedColumnStream` uses `substr(0, N)` instead of `substr(N)`, corrupting `columns_substreams.txt` after `RENAME COLUMN`


clickgapai:
_Found via ClickGap automated review. Please close or comment if this is incorrect or needs adjustment._

_Retrospective finding from a historical scan of [PR #81091](https://github.com/ClickHouse/ClickHouse/pull/81091) (merged 2025-07-21). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

After `ALTER TABLE RENAME COLUMN` on a Wide part, the `columns_substreams.txt` metadata file contains garbled substream names. For example, renaming column `arr` to `brr` produces a single substream `brrarr` instead of the correct `brr.size0` and `brr`.

**Root cause:** ISerialization.cpp:446: `return escapeForFileName(name_to) + file_name.substr(0, name_from_escaped.size());` — `substr(0, N)` returns the first N chars (the old name prefix), but should be `substr(N)` to return everything AFTER the old name prefix (the substream suffix like `.size0`).

**Why we believe this is a bug:** `MutateTask.cpp:516` → `addRenamedColumnToColumnsSubstreams` → `ISerialization::getFileNameForRenamedColumnStream` (ISerialization.cpp:446). The function uses `file_name.substr(0, name_from_escaped.size())` which returns the OLD prefix (first N characters), not the SUFFIX after the prefix. Both lines 446 and 450 have this bug.

**Affected locations:**
- `src/DataTypes/Serializations/ISerialization.cpp:446` — getFileNameForRenamedColumnStream first branch: substr(0, N) instead of substr(N)
- `src/DataTypes/Serializations/ISerialization.cpp:450` — getFileNameForRenamedColumnStream nested branch: same substr(0, N) bug

**Impact:** After `RENAME COLUMN` mutation on a Wide part, the `columns_substreams.txt` metadata contains wrong substream names. The bug also loses substreams (e.g., 2 substreams both map to the same garbled name, and dedup reduces it to 1). Currently the corruption has no SQL-observable impact because non-dynamic types use `enumerateStreams` for size calculation and system table display, and dynamic types are rewritten during rename mutations. However, the corruption persists on disk and propagates through subsequent mutations.

### Does it reproduce on most recent release?

Yes — confirmed on current `master` (commit `584b6bdbf78e`).

### How to reproduce

_See root cause description and affected locations above._

[...]

PR #104037, opened 2026-05-04
Allowlist `02555_davengers_rename_chain` upgrade-check errors

alexey-milovidov:
### Description

PR #103917 narrowed the `wrong_metadata_wide` allowlist in `tests/docker_scripts/upgrade_runner.sh` to the specific corrupted-filename signature emitted by `02538_alter_rename_sequence` after the `getFileNameForRenamedColumnStream` `substr(0, N)` bug (issue #102259, fix in PR #102689). The same bug also surfaces on `02555_davengers_rename_chain`'s `wrong_metadata` table, but with a different column/stream pair (`b`/`ba1.bin` instead of `column1_renamed`/`column1_renamedcolumn1.bin`), so the narrow regex did not catch it and the upgrade check began failing on unrelated PRs.

This PR adds a matching narrow signature for the chained-rename variant and loosens the secondary `Detaching broken part` regex so it covers both the narrow and wide table variants. The other two substrings of the secondary regex (`Detaching broken part` + `backward incompatibility`) keep the allowlist scoped to the same symptom.

Verified against the failing log from PR #103874:
https://s3.amazonaws.com/clickhouse-test-reports/PRs/103874/49885046aba0893d06655439d554aeac6998469e/upgrade_check_amd_release/upgrade_error_messages.txt

All six lines from that log are filtered, while synthetic unrelated errors (other tables, generic detach notices, log lines that mention the table name in passing) pass through.

Source PR with the failing run: https://github.com/ClickHouse/ClickHouse/pull/103874

...

### Version info
- Merged into: `26.5.1.319`

### DIFF
--- a/tests/docker_scripts/upgrade_runner.sh
+++ b/tests/docker_scripts/upgrade_runner.sh
@@ -349,10 +349,15 @@ cp /var/log/clickhouse-server/clickhouse-server.upgrade.log /test_output/clickho
 #       `<renamed><original>.bin` instead of `<renamed>.bin`). The fix is in PR #102689; until it lands, the
 #       upgraded server detaches the renamed-column parts of `02538_alter_rename_sequence`'s `wrong_metadata_wide`
 #       table. Matched via the exact corrupted filename + column name, which is unique to that test and that bug.
-# `wrong_metadata_wide` + `Detaching broken part` + `backward incompatibility` is the follow-up cleanup line for
-#       the same issue: a "Detaching broken part" notice that does not contain the corrupted filename. Filtered
-#       via regex in the secondary pipe below to require all three substrings together, so unrelated broken-part
-#       detach messages are not masked.
+# `No stream (ba1.bin) file checksum for column b` is the same bug observed on
+#       `02555_davengers_rename_chain`'s `wrong_metadata` table. The test chains `a -> a1` and then
+#       `a1 -> b`, so the corrupted file name is `<new=b><old=a1>.bin = ba1.bin` for column `b`. The
+#       `<column><stream>` combination is unique to this chained-rename test.
+# `wrong_metadata` + `Detaching broken part` + `backward incompatibility` is the follow-up cleanup line for
+#       the same issue: a "Detaching broken part" notice that does not contain the corrupted filename.
+#       Filtered via regex in the secondary pipe below to require all three substrings together, so unrelated
+#       broken-part detach messages are not masked. The regex matches both `wrong_metadata` (from
+#       `02555_davengers_rename_chain`) and `wrong_metadata_wide` (from `02538_alter_rename_sequence`).
 echo "Check for Error messages in server log:"
 rg -Fav -e "Code: 236. DB::Exception: Cancelled merging parts" \
            -e "Code: 236. DB::Exception: Cancelled mutating parts" \
@@ -423,6 +428,7 @@ rg -Fav -e "Code: 236. DB::Exception: Cancelled merging parts" \
            -e "e.what() = failed to parse response body" \
            -e "Tuple element name 'null' is reserved" \
            -e "No stream (column1_renamedcolumn1.bin) file checksum for column column1_renamed" \
+           -e "No stream (ba1.bin) file checksum for column b" \
     /test_output/clickhouse-server.upgrade.log \
     | grep -av -e "_repl_01111_.*Mapping for table with UUID" \
     | grep -av -e "Azure::Storage::StorageException.*Not found address of host" \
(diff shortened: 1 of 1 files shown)


tests/docker_scripts/upgrade_runner.sh
```diff
@@ -349,10 +349,15 @@ cp /var/log/clickhouse-server/clickhouse-server.upgrade.log /test_output/clickho
 #       `<renamed><original>.bin` instead of `<renamed>.bin`). The fix is in PR #102689; until it lands, the
 #       upgraded server detaches the renamed-column parts of `02538_alter_rename_sequence`'s `wrong_metadata_wide`
 #       table. Matched via the exact corrupted filename + column name, which is unique to that test and that bug.
-# `wrong_metadata_wide` + `Detaching broken part` + `backward incompatibility` is the follow-up cleanup line for
-#       the same issue: a "Detaching broken part" notice that does not contain the corrupted filename. Filtered
-#       via regex in the secondary pipe below to require all three substrings together, so unrelated broken-part
-#       detach messages are not masked.
+# `No stream (ba1.bin) file checksum for column b` is the same bug observed on
```
alexey-milovidov, line 352: @groeneai, is this bug inherent to the old ClickHouse version (used in the upgrade check) or to master as well? If the latter, we should investigate and fix the bug.

alexey-milovidov: @clickgapai, same question ^

groeneai:
@alexey-milovidov — investigated. **The bug is OLD-VERSION-ONLY.** Master is correct; the allowlist in this PR is the right approach.

### Evidence

**1. PR #102689's fix is complete on master.** `src/DataTypes/Serializations/ISerialization.cpp` line 454 (regular column) and line 458 (Nested column) both now use `file_name.substr(name_from_escaped.size())` — the correct suffix. The buggy `substr(0, name_from_escaped.size())` (which took the prefix) is gone in both paths. Caller in `MutateTask.cpp:520` is the only consumer.

**2. Why the corrupted filename `ba1.bin` is unique to the old code.** Tracing `02555_davengers_rename_chain`'s chain `a → a1 → b` on the substream `a`:

| Step | Old buggy `substr(0, N)` | New fixed `substr(N)` |
|------|--------------------------|------------------------|
| Initial substream | `a` | `a` |
| After `a → a1` | `"a1" + "a".substr(0,1)` = **`a1a`** | `"a1" + "a".substr(1)` = **`a1`** |
| After `a1 → b` | `"b" + "a1a".substr(0,2)` = **`ba1`** | `"b" + "a1".substr(2)` = **`b`** |

The buggy code accumulates the original column-name prefix across each rename in the chain. `ba1.bin` is *exactly* what the old code produces here; master's fixed code produces `b.bin`.

**3. Empirical reproduction on master HEAD** (binary built from `bc39c42c...`, includes #102689):

```
CREATE TABLE test_chain (a UInt32, b UInt32, c UInt32) ENGINE=MergeTree ORDER BY tuple()
SETTINGS min_bytes_for_wide_part=0, replace_long_file_name_to_hash=0;
INSERT INTO test_chain VALUES (1, 2, 3);
ALTER TABLE test_chain RENAME COLUMN a TO a1, RENAME COLUMN b TO b1 SETTINGS mutations_sync = 1;
ALTER TABLE test_chain RENAME COLUMN a1 TO b, RENAME COLUMN b1 TO a SETTINGS mutations_sync = 1;
```

After the chained rename, the part directory contains `a.bin`, `b.bin`, `c.bin` (correct) — no `ba1.bin`, no `a1a.bin`. `columns_substreams.txt` lists clean substream names (`a`, `b`, `c`). Subsequent INSERTs and SELECTs work normally.

**4. CIDB cross-check (60 days):** zero master/release-branch failures matching `%ba1.bin%` OR `%column1_renamedcolumn1.bin%`. All ~50 hits are PR-side upgrade-check runs across 30+ unrelated PRs, exactly the upgrade-flow signature: old release writes the corrupted file → master detects it on startup → "Detaching broken part … backward incompatibility".

### Why the upgrade check still fires it

`tests/docker_scripts/upgrade_runner.sh` line 114 runs the stress test against `previous_release_repository/tests/queries` (the previous release's source tree, including its bug), then upgrades to master. The corrupted parts are written by the previous release; master reads them on startup and correctly detaches them. The "Detaching broken part" message is master's intended behavior for legacy on-disk corruption — not a master bug.

The allowlist in this PR is the correct approach: the column+stream pair `b`/`ba1.bin` is unique to `02555_davengers_rename_chain`'s `wrong_metadata` table (and `column1_renamed`/`column1_renamedcolumn1.bin` is unique to `02538_alter_rename_sequence`'s `wrong_metadata_wide`), so the regex remains tight. Once the fix is shipped in every supported release used by the upgrade check, the allowlist entries can be removed.

(Investigated by groeneai. session: cron:clickhouse-ci-task-worker:20260506-014500)

PR #104051, opened 2026-05-04
Refreshable MV: avoid duplicate refresh on brief keeper connection loss

al13n321:
- Improvement

Replicated refreshable materialized view in APPEND mode will no longer refresh twice if connection to zookeeper was briefly lost during refresh.

---

Had to reorganize RefreshTask code quite a bit for this one.
 * Scheduling / zookeeper management and refresh execution are now in separate background tasks, so that we can write to zookeeper while refresh is running. (I tried another approach where only the re-creation of ephemeral znode is in a separate task and everything else is ~unchanged, but that was much spaghettier and worse.)
 * Main znode now has `bool refresh_running`, in addition to the ephemeral znode `running`.
 * If the ephemeral znode unexpectedly disappears, (a) its owner re-creates it, (b) other replicas start a grace period timer and give up on the refresh if the znode doesn't reappear for 40 seconds.

(Artisanally coded, seems too tricky for opus 4.7. The test is vibecoded, but needed a few fixes.)

### Version info
- Backported to: `26.4.3.25`, `26.3.13.11`, `26.2.19.6`


src/Core/Settings.cpp
```diff
@@ -7784,6 +7784,9 @@ Timeout for waiting for window view fire signal in event time processing
     \
     DECLARE(Bool, stop_refreshable_materialized_views_on_startup, false, R"(
 On server startup, prevent scheduling of refreshable materialized views, as if with SYSTEM STOP VIEWS. You can manually start them with `SYSTEM START VIEWS` or `SYSTEM START VIEW <name>` afterwards. Also applies to newly created views. Has no effect on non-refreshable materialized views.
+)", EXPERIMENTAL) \
+    DECLARE(UInt64, refreshable_materialized_view_keeper_grace_period, 40'000, R"(
```
nikitamikhaylov: So it's 40s, just 10 more than `DEFAULT_SESSION_TIMEOUT_MS` which is a time after which the replica really becomes "inactive". It seems it can still introduce a race condition. 

P.S. It's better to add _ms suffix to make it easier for the reader.

nikitamikhaylov: Btw, I'd better removed this setting and hardcoded 1.25 * `session_timeout_ms`.

nikitamikhaylov (approved):
Looks good overall, but the APPEND mode doesn't have any "commit point", being some sort of EXCHANGE or so. 

Maybe in the future we can consider implementing this mode using a staging table + MOVE PARTITION instead?

al13n321:
> Maybe in the future we can consider implementing this mode using a staging table + MOVE PARTITION instead?

Or we could have a setting to do the INSERT as one block (`min_insert_block_size_rows = infinity`?) and inject an RMV znode version check into the Replicated/SharedMergeTree's keeper multi-write that publishes the new data part, similar to the existing check injected into DatabaseReplicated through setDDLAdditionalChecksOnEnqueue.

src/Core/Settings.cpp
al13n321: It's mostly not related to session timeout, the grace period normally starts after the session already expired. Initially I accidentally set it to 40ms, and it still worked. But I like removing settings, so changing to 1.25x session timeout.

PR #104063, opened 2026-05-05
Bypass index uncompressed cache when its size is zero

al13n321:
- Performance Improvement

Avoid index uncompressed cache overhead when the cache is disabled (server setting `index_uncompressed_cache_size = 0`, which is the default).

---

Simple change.

(Artisanally written changelog entry and C++ code. Vibecoded unreviewed test, but I did check that it fails the expected way without this PR. The rest of this PR summary is from claude.)
(See the reverse-linked support escalation for customer report.)

---

With `index_uncompressed_cache_size = 0` (the server default, advertised as "disabled"), `Context::getIndexUncompressedCache` still returned a non-null `UncompressedCache` of capacity zero. `MergeTreeIndexReader` then took the `CachedCompressedReadBuffer` path in `MergeTreeReaderStream::init`, so every skip-index block read acquired the global `CacheBase` mutex twice, evicted itself via `removeOverflow`, and unconditionally incremented `UncompressedCacheMisses` and `UncompressedCacheWeightLost`. With small-`GRANULARITY` skip indexes (`ngrambf_v1`, `bloom_filter`, etc.) this can produce hundreds of thousands of contended mutex acquisitions per query and tens of GB of phantom `UncompressedCacheWeightLost`, with zero `UncompressedCacheHits`.

The original 2021 code only constructed the cache when the configured size was non-zero. The `if (size)` guard at the call site in `Server.cpp` was later removed so the cache could be resized at runtime via `SYSTEM RELOAD CONFIG`; the data uncompressed cache survived this because `MergeTreeReadPoolBase` already gates it on `pool_settings.use_uncompressed_cache`, but the index-read path had no such guard.

This change tracks whether the cache should actually be used in a separate flag, set in `setIndexUncompressedCache` and refreshed in `updateIndexUncompressedCacheConfiguration`. `getIndexUncompressedCache` returns nullptr when the flag is false; `MergeTreeReaderStream::init` already falls through to `CompressedReadBufferFromFile` on a null pointer. Runtime resize from `0 -> N` still works.

### Version info
- Merged into: `26.5.1.314`
- Backported to: `26.4.2.4`, `26.3.10.62`, `26.2.18.10`, `25.8.24.4`

### DIFF
--- a/src/Interpreters/Context.cpp
+++ b/src/Interpreters/Context.cpp
@@ -555,7 +555,7 @@ struct ContextSharedPart : boost::noncopyable
     mutable OnceFlag build_vector_similarity_index_threadpool_initialized;
     mutable std::unique_ptr<ThreadPool> build_vector_similarity_index_threadpool; /// Threadpool for vector-similarity index creation.
     mutable UncompressedCachePtr index_uncompressed_cache TSA_GUARDED_BY(mutex);      /// The cache of decompressed blocks for MergeTree indices.
-    mutable bool index_uncompressed_cache_enabled = false TSA_GUARDED_BY(mutex);      /// Whether index_uncompressed_cache should be used.
+    mutable bool index_uncompressed_cache_enabled TSA_GUARDED_BY(mutex) = false;      /// Whether index_uncompressed_cache should be used.
     mutable VectorSimilarityIndexCachePtr vector_similarity_index_cache TSA_GUARDED_BY(mutex);         /// Cache of deserialized secondary index granules.
     mutable TextIndexTokensCachePtr text_index_tokens_cache TSA_GUARDED_BY(mutex);  /// Cache of deserialized text index tokens.
     mutable TextIndexHeaderCachePtr text_index_header_cache TSA_GUARDED_BY(mutex);  /// Cache of deserialized text index headers.
--- a/src/Interpreters/Context.cpp
+++ b/src/Interpreters/Context.cpp
@@ -4110,10 +4110,12 @@ void Context::updateIndexUncompressedCacheConfiguration(const Poco::Util::Abstra
     shared->index_uncompressed_cache_enabled = shared->index_uncompressed_cache->maxSizeInBytes() != 0;
 }
 
-UncompressedCachePtr Context::getIndexUncompressedCache() const
+UncompressedCachePtr Context::getIndexUncompressedCache(bool only_if_enabled) const
 {
     SharedLockGuard lock(shared->mutex);
-    return shared->index_uncompressed_cache_enabled ? shared->index_uncompressed_cache : nullptr;
+    if (only_if_enabled && !shared->index_uncompressed_cache_enabled)
+        return nullptr;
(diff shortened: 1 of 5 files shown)


src/Interpreters/Context.cpp
```diff
@@ -4105,11 +4107,14 @@ void Context::updateIndexUncompressedCacheConfiguration(const Poco::Util::Abstra
...
     }
     shared->index_uncompressed_cache->setMaxSizeInBytes(size);
+    shared->index_uncompressed_cache_enabled = shared->index_uncompressed_cache->maxSizeInBytes() != 0;
 }
 
-UncompressedCachePtr Context::getIndexUncompressedCache() const
+UncompressedCachePtr Context::getIndexUncompressedCache(bool only_if_enabled) const
 {
     SharedLockGuard lock(shared->mutex);
+    if (only_if_enabled && !shared->index_uncompressed_cache_enabled)
```
Algunenano, line 4116: Why do we need `index_uncompressed_cache_enabled`? Could we do `if (only_if_enabled && shared->index_uncompressed_cache->maxSizeInBytes() == 0)` instead and avoid adding extra members?

Algunenano (approved):
Looks good for an artisanal PR 😄 . I left a small comment

src/Interpreters/Context.cpp
al13n321: `maxSizeInBytes()` locks an std::mutex, I avoided it just in case, though I don't have a good mental model of what's important or not important for performance here.

(We could go further and make index_uncompressed_cache_enabled atomic so we can check it without locking `shared->mutex`. But that seemed too weird, and the caller of getIndexUncompressedCache does a few similar calls that lock `shared->mutex` the same way anyway.)

(EDIT: Or we could make `maxSizeInBytes()` thread safe by making the field atomic in all 3 ICachePolicy implementations.)

al13n321:
Test failures looked unrelated, ran `gh api repos/ClickHouse/ClickHouse/statuses/f78db8ebaa816ab6db05027435e9ffe72b29e54b -X POST -F state=success -F description="Manually unset to merge" -F context="Mergeable Check"`

EDIT: `02270_errors_in_files_s3` presumably already fixed by https://github.com/ClickHouse/ClickHouse/pull/104077 , `Docker server image` reported in https://github.com/ClickHouse/ClickHouse/issues/104152

clickgapai:

**Affected code:** `Context::getIndexUncompressedCache` and skip-index read paths (`MergeTreeIndexReader`, `MergeTreeReaderTextIndex`, `MergeTreeSkipIndexReader`) — core code present in all supported branches.

**Why this might apply:** Fixes a performance regression affecting all users running the default config (`index_uncompressed_cache_size=0`). PR description says the regression dates from the 2021 removal of the `if (size)` guard at the cache construction site in `Server.cpp` — the affected code (`Context::getIndexUncompressedCache` + skip-index call sites) predates all currently supported stable branches. Customer escalation is referenced in the PR. No backport label is present.

PR #104065, opened 2026-05-05
Refreshable MV: relax sanity checks in SECONDARY_CREATE

al13n321:

We've seen a RMV in SharedCatalog with non-replicated MergeTree table. This is normally not allowed, and idk how it got created, but it makes sense to allow that for SECONDARY_CREATE anyway.

### Version info
- Merged into: `26.5.1.353`
- Backported to: `26.4.3.11`, `26.3.11.10`, `26.2.19.14`

### DIFF
--- a/src/Storages/StorageMaterializedView.cpp
+++ b/src/Storages/StorageMaterializedView.cpp
@@ -199,7 +199,7 @@ StorageMaterializedView::StorageMaterializedView(
             }
         }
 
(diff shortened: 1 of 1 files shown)


alexey-milovidov (approved):

nihalzp:
@fm4v Any idea why v25.12 backports are not being created?

PR #104081, opened 2026-05-05
CI: collect server stacktraces on server death, simplify stop logic

leshikus:
`clickhouse-test` has a `print_stacktraces` function that collects gdb/`system.stack_trace` backtraces, but it was only called at startup and during the hung-query check — never when the server was detected dead mid-run. This left the most important failure mode without any trace output.

## Changes

**Stacktrace collection on server death** — `print_stacktraces` is now called at the two earliest detection points:
- When a test result comes back with `FailureReason.SERVER_DIED` (in `run_tests_array`)
- When the liveness hung-check finds the server unresponsive (in `do_run_tests`)

**Removed `ServerDied` exception** — the exception was raised immediately after setting the `server_died` event, so it was redundant: the event already signals all workers to stop, and `main` already sets exit code 1 when the event is set. Replaced all raises with `break`. The two non-server-death stops (`--max-failures` limit and max failures chain) now print an explicit message before breaking. A bug was also fixed where these two cases did not set the stop flag, so `main`'s suite loop would continue into subsequent suites.

**Renamed `server_died` → `stop_testing`** — the flag is now set for `--max-failures` and failures-chain limits too, not only for actual server death, so the old name was misleading.

### Version info
- Merged into: `26.5.1.456`

### DIFF
--- a/tests/clickhouse-test
+++ b/tests/clickhouse-test
@@ -3994,10 +3994,6 @@ class TestSuite:
         return TestSuite(args, suite_path, suite_tmp_path, suite)
 
 
-class ServerDied(Exception):
-    pass
-
-
 class GlobalTimeout(Exception):
     pass
 
@@ -4229,7 +4225,7 @@ def run_tests_array(
 
         if server_died.is_set():
             stop_tests()
-            raise ServerDied("Server died")
+            break
 
         if args.stop_time and time() > args.stop_time:
             print("\nStop tests run because global time limit is exceeded.\n")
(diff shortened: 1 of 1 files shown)


maxknv:
IMHO, it's more error-prone to replace exception with break. We can break just the nested loop. Also, we still have GlobalTimeout exception and bare raises. The approach now is mixed. I'd prefer a separate exception for each reason, with distinct exit codes.

leshikus:
@maxknv the problem with the exception is a broken metaphor; we generally should be able to throw it everywhere; this is not the case for multi-process code, we have to stop other processes before throwing the exception; please let me know if I should restore exception

maxknv (approved):

PR #104087, opened 2026-05-05
Skip additional macOS-failing tests in darwin.skip

leshikus:
Adds ~180 tests to \`ci/defs/darwin.skip\` that consistently fail on macOS
(Darwin) CI runs due to platform-specific behaviour — missing Linux
syscalls, different network stack defaults, path handling, or ZooKeeper
availability. Skipping them unblocks the Darwin CI job without affecting
Linux coverage.

### Version info
- Merged into: `26.5.1.287`


alexey-milovidov (approved):
Ok, though it will be difficult to find out the reason why they are disabled.

alexey-milovidov:
Also, 877 tests - way too many.

leshikus:
@alexey-milovidov for now I add the tests which fail from time to time (using cidb to track them), the priority is to make a job stable enough to be included as a regular job which can prevent regressions; then I gradually remove tests from the skip list

PR #102801, opened 2026-04-15
Improve canceling queries with url() function

rvasin:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Improve canceling queries with url() function by checking query cancellation between HTTP retry attempts.

Related #97459

### Version info
- Merged into: `26.5.1.262`

PR #104089, opened 2026-05-05
Improve canceling queries with url() function

rvasin:
- Improvement

Improve canceling queries with url() function by checking query cancellation between HTTP retry attempts.

This PR fixes reverted PRs #102801 and #104014


rvasin:
@Algunenano @murphy-4o @vdimir I have fixed the cancellation callback in `ReadWriteBufferFromHTTP::doWithRetries`. Please assign `can be tested` probably after merging the reverting PR #104077

src/IO/ReadWriteBufferFromHTTP.cpp
```diff
@@ -325,7 +325,8 @@ void ReadWriteBufferFromHTTP::doWithRetries(std::function<void()> && callable,
 
     for (size_t attempt = 1; attempt <= read_settings.http_max_tries; ++attempt)
     {
-        if (check_cancelled && check_cancelled())
+        /// Check cancellation since the second attempt to let the original HTTP errors propagate correctly on the first attempt.
+        if (attempt > 1 && check_cancelled && check_cancelled())
```
Algunenano: I find this and the exception here extremely confusing and inexact as a user.

The query wasn't cancelled, timed out or KILLed, which is what I understand when I see a  `QUERY_WAS_CANCELLED` exception. What's actually happened in the flaky test is that the url() glob launched two HTTP requests in parallel, one threw a real error and the pipeline called IProcessor::cancel() on the other source as part of the error tear-down.

The peer's HTTP retry loop now reinterprets that signal as a fresh QUERY_WAS_CANCELLED and races the real exception to tryInsertException. When the cancellation throw is faster, the user visible error both shows the wrong URI and message.

I think this throw should be removed altogether: just break out of the retry loop on check_cancelled() and let the peer's already-thrown real exception be the one the executor reports. There's no risk of swallowing the error, since the only way isCancelled() becomes true on this source is that some other processor already produced an exception.

rvasin: I agree with your arguments about QUERY_WAS_CANCELLED exception throwing. In fact I have made several pull requests for Sources and Transforms, everywhere there was exit of for/while loop. This is the first PR where I try to throw QUERY_WAS_CANCELLED and as you described it may produce some side effects/races. Thus I will try to refactor this PR and exit the loop instead of throwing QUERY_WAS_CANCELLED. Thank you for the suggestion.

src/IO/ReadWriteBufferFromHTTP.cpp
```diff
@@ -322,6 +323,10 @@ void ReadWriteBufferFromHTTP::doWithRetries(std::function<void()> && callable,
 
     for (size_t attempt = 1; attempt <= read_settings.http_max_tries; ++attempt)
     {
+        /// Check cancellation since the second attempt to let the original HTTP errors propagate correctly on the first attempt.
+        if (attempt > 1 && check_cancelled && check_cancelled())
+            break;
```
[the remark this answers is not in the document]
rvasin: @Algunenano @murphy-4o Please note that CI is "green" now. But I think I need to make more `break` related improvements.

Interesting that `clickhouse-gh` suggested to use throw QUERY_WAS_CANCELLED. But if we follow to `break` approach after each buffer operation I have added cancellation checks (it's in my working copy of code now). StorageURLSource makes 3 calls of buffer and after each call I have added the cancellation check:
```cpp
StorageURLSource::StorageURLSource(...)
{
    /// Lazy initialization. We should not perform requests in constructor, because we need to do it in query pipeline.
    initialize = [=, this]()
    {
        std::vector<String> current_uri_options;
        std::pair<Poco::URI, std::unique_ptr<ReadWriteBufferFromHTTP>> uri_and_buf;
        do
        {
            current_uri_options = (*uri_iterator)();
            if (current_uri_options.empty())
                return false;

            auto first_option = current_uri_options.cbegin();
            uri_and_buf = getFirstAvailableURIAndReadBuffer(   /// <-- first call of buffer
                first_option,
                current_uri_options.end(),
                getContext(),
                params,
                http_method,
                callback,
                timeouts,
                credentials,
                headers,
                glob_url,
                current_uri_options.size() == 1,
                [this]() { return isCancelled(); });

            // Check after buffer operation
            if (isCancelled())
                return false;

            /// If file is empty and engine_url_skip_empty_files=1, skip it and go to the next file.
        }
        while (getContext()->getSettingsRef()[Setting::engine_url_skip_empty_files] && uri_and_buf.second->eof()); /// <-- 2nd call of buffer
        // Check after buffer operation
        if (isCancelled())
            return false;

        curr_uri = uri_and_buf.first;
        current_file_last_modified = uri_and_buf.second->tryGetLastModificationTime(); /// <-- 3rd call of buffer
        // Check after buffer operation
        if (isCancelled())
            return false;
```

Then in:

```cpp
Chunk StorageURLSource::generate()
{
    while (true)
    {
        if (isCancelled())
        {
            if (reader)
                reader->cancel();
            break;
        }

        if (!reader && !initialize())
            return {};  /// <-- it will return here in case of cancellation with empty chunck 

        Chunk chunk;
        if (reader->pull(chunk)) 
```
```cpp
void ISource::work()
{
    try
    {
        if (finished)
            return;

        read_progress_was_set = false;

        if (auto chunk = tryGenerate()) /// <-- empty chunk is returned in case of cancellation
        {
            current_chunk.chunk = std::move(*chunk);
            if (current_chunk.chunk)
            {
                has_input = true;
                if (auto_progress && !read_progress_was_set)
                    progress(current_chunk.chunk.getNumRows(), current_chunk.chunk.bytes());
            }
        }
        else
            finished = true; /// <- finish on cancellation 
```
I think this approach is reliable and it's better alternative to throwing the exception.

src/Storages/StorageURL.cpp
```diff
@@ -583,7 +585,10 @@ std::pair<Poco::URI, std::unique_ptr<ReadWriteBufferFromHTTP>> StorageURLSource:
                            .withSkipNotFound(skip_url_not_found_error)
                            .withHeaders(headers)
                            .withDelayInit(delay_initialization)
-                           .create(credentials);
+                            .create(credentials);
+
+            if (check_cancelled)
```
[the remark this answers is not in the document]
rvasin: Done. Added into into constructor and builder.

tests/queries/0_stateless/04078_kill_query_url_function.sh
```diff
@@ -0,0 +1,122 @@
...
+{
+    local head_sleep=$1
+    local get_sleep=$2
+    local skip_empty=$3
+
+    local query_id="kill_query_url_${CLICKHOUSE_DATABASE}_$RANDOM"
+    local log_file=$(mktemp "./04078.XXXXXX.log")
+
+    # Get free port
+    local HTTP_PORT=$(python3 -c "
```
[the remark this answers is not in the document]
rvasin: Similar approach is used in tests/queries/0_stateless/04032_empty_gzip_stream_from_url.sh

rvasin:
Failed test is not related.

rvasin:
@Algunenano  do you have any suggestions about improvement of this PR? I think the comment from AI about "free" port is not critical and instead of throwing an exception I have implemented silent loop exiting (and it seems working according to several CI runs).

src/IO/ReadWriteBufferFromHTTP.cpp
```diff
@@ -322,6 +324,12 @@ void ReadWriteBufferFromHTTP::doWithRetries(std::function<void()> && callable,
 
     for (size_t attempt = 1; attempt <= read_settings.http_max_tries; ++attempt)
     {
+        /// Check cancellation since the second attempt to let the original HTTP errors propagate correctly on the first attempt.
+        if (attempt > 1 && cancellation_check && cancellation_check())
```
Algunenano, line 328: We don't propagate any errors here, so I don't understand why we skip the first attemp.

src/Storages/StorageURL.cpp
```diff
@@ -381,16 +381,34 @@ StorageURLSource::StorageURLSource(
                 credentials,
                 headers,
                 glob_url,
-                current_uri_options.size() == 1);
+                current_uri_options.size() == 1,
+                [this]() { return isCancelled(); });
+
+            /// getFirstAvailableURIAndReadBuffer makes HTTP request in ReadWriteBufferFromHTTP.
```
Algunenano, line 388: Why not check at  the start of the loop?

src/Storages/StorageURL.cpp
```diff
@@ -381,16 +381,34 @@ StorageURLSource::StorageURLSource(
...
             /// If file is empty and engine_url_skip_empty_files=1, skip it and go to the next file.
         }
         while (getContext()->getSettingsRef()[Setting::engine_url_skip_empty_files] && uri_and_buf.second->eof());
+        /// Check for cancellation because eof() makes HTTP request in buffer.
+        if (isCancelled())
+            return false;
 
         curr_uri = uri_and_buf.first;
         current_file_last_modified = uri_and_buf.second->tryGetLastModificationTime();
+        /// Check for cancellation because tryGetLastModificationTime makes HTTP request in buffer.
```
Algunenano, line 402: I don't understand the pattern of doing the HTTP request, and then checking if the pipeline is cancelled. Shouldn't it be the other way around?

Algunenano (changes requested):
The problem that I see right now is that it is really easy to misuse, because you don't get any feedback on the thread that was cancelled, and exit normally. 

A simple look at `ReadWriteBufferFromHTTP::readBigAt` shows how trivial is now to hit asserts due to a pipeline cancellation. I think part of the confusion comes from dealing with query cancellation and pipeline cancellation as the same thing, when they aren't, and how to report this properly both in logs and to the end user.

Another related issue that we should address is that we are sleeping without checks, that is, a cancel does not wake up the thread, which means we wait just to exit. We should use a condition variable and wait on it instead.

And finally, but this one I don't expect to address in this PR, is that slow request won't be dealt with, we'll wait until the request either finishes or times out.

rvasin:
@Algunenano thank you for reviewing this PR. 

Let's create a TODO plan what should be done before merging this PR. Because it's not clear for me. Sorry for asking many questions I just want to be sure I have 100% understood you.

First of all I have implemented 2 approached and both have passed CI successfully:
1. throwing exception (which is recommended now by AI reviewer).
2. "silent" exit doWithRetries and then processing it with cancellation checks in `StorageURLSource::StorageURLSource`.

I suppose you are talking about improving/refactoring the 2nd approach (to make it more reliable)? 
Because I don't see any 3rd or 4th approaches. 

> The problem that I see right now is that it is really easy to misuse, because you don't get any feedback on the thread that was cancelled, and exit normally.

> A simple look at `ReadWriteBufferFromHTTP::readBigAt` shows how trivial is now to hit asserts due to a pipeline cancellation. I think part of the confusion comes from dealing with query cancellation and pipeline cancellation as the same thing, when they aren't, and how to report this properly both in logs and to the end user.
> 

What do you suggest to do? From my side I see 3 variants:

variant 1. we may make `doWithRetries` returning bool `bool ReadWriteBufferFromHTTP::doWithRetries` instead of `void`. 
Then:

```cpp
bool ReadWriteBufferFromHTTP::doWithRetries(std::function<void()> && callable,
                                            std::function<void()> on_retry,
                                            bool mute_logging) const
{
    // skip
    for (size_t attempt = 1; attempt <= read_settings.http_max_tries; ++attempt)
    {
        if (attempt > 1 && cancellation_check && cancellation_check())
            return false; 
        

// then in  ReadWriteBufferFromHTTP::readBigAt:
    if (doWithRetries(...))
        chassert(total_bytes_copied == initial_n || is_canceled);
    return total_bytes_copied;
```

variant 2. pass a cancellation parameter to `doWithRetries()` like `is_canceled` parameter in `copyFromIStreamWithProgressCallback(*result.response_stream, to, n, progress_callback, &bytes_copied, &is_canceled);`, 
Then:

```cpp
void ReadWriteBufferFromHTTP::doWithRetries(std::function<void()> && callable,
                                            std::function<void()> on_retry,
                                            bool mute_logging,
                                            bool * is_canceled) const
{
    // skip
    for (size_t attempt = 1; attempt <= read_settings.http_max_tries; ++attempt)
    {
        if (attempt > 1 && cancellation_check && cancellation_check())
            *is_canceled = true;
            break;
            
// then in ReadWriteBufferFromHTTP::readBigAt:
    cancelled_during_retries = doWithRetries(...)
    chassert(total_bytes_copied == initial_n || is_canceled || cancelled_during_retries);
    return total_bytes_copied;
```

variant 3:
Leave code of `ReadWriteBufferFromHTTP::doWithRetries` as it is then  in  `ReadWriteBufferFromHTTP::readBigAt`:

```cpp
    doWithRetries(...)
    
    if (!is_canceled && cancellation_check && cancellation_check())
        is_canceled = true;
    chassert(total_bytes_copied == initial_n || is_canceled);
```
I think this 3rd variant may be better because- we already have `cancellation_check` as class member of `ReadWriteBufferFromHTTP` 
So why should be add additional parameter to `doWithRetries` as in variant 2 or make it returning `bool` as in variant 1?

> Another related issue that we should address is that we are sleeping without checks, that is, a cancel does not wake up the thread, which means we wait just to exit. We should use a condition variable and wait on it instead.

You are absolutely right. I did not think at first that we are still waiting on retry. My primary purpose was to stop the retries on cancellation. So usage of the conditional variable will improve the cancellation speed even more. 
Do you think I should implement it directly in this PR? I am afraid to break something it it. So I tried to make the changed minimal in this PR.

> And finally, but this one I don't expect to address in this PR, is that slow request won't be dealt with, we'll wait until the request either finishes or times out.

Yes. Let's don't make to much changes in one PR. Probably it should be noted as TODO.

src/Storages/StorageURL.cpp
rvasin: Another way could be throwing the exception. But we decided not to do that. Right now I don't see other simple variants. What do you think?

src/Storages/StorageURL.cpp
rvasin: This is because inside `getFirstAvailableURIAndReadBuffer` there is the HTTP request:

```cpp
            if (context_->getSettingsRef()[Setting::engine_url_skip_empty_files] && res->eof() && option != std::prev(end))
```
In fact `res->eof()` makes the request. AI reviewer have noticed it.
But it also make sense to add cancellation chat at the beginning of the loop. Or even before the loop.

src/IO/ReadWriteBufferFromHTTP.cpp
rvasin: In fact the main reason why both of my PRs #104014 and #102801 because `02270_errors_in_files_s3` was failed. In have found in logs of the test that CANNOT_PARSE_INPUT_ASSERTION_FAILED was thrown. I reported it [here](https://github.com/ClickHouse/ClickHouse/pull/102801#issuecomment-4378088642). But I have to test it again to be sure. On my current "silent" approach when exiting on the first attempt some other s3 storage related tests were failed (because `url()` is used for minio in these tests). But all tests work fine when we exit on since the second attempt.

Algunenano:
@rvasin Please have a look at https://github.com/ClickHouse/ClickHouse/issues/104929

rvasin:
> @rvasin Please have a look at #104929

It looks interesting. The cancellation problem itself is not easy and some more deeper-level refactoring is required as you suggested in the RFC. The question is cost as pointer by Sergey. I will try to take a closer look on Monday.

src/IO/ReadWriteBufferFromHTTP.cpp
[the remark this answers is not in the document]
rvasin: Fixed by adding !isCancelled() into condition before calling addNumRowsToCache().

tests/queries/0_stateless/04078_kill_query_url_function.sh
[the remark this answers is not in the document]
rvasin: Fixed port binding as requested.

rvasin:
@Algunenano what do you think is remaining to make in this PR? For sleep function I think it should be done in a separate PR. As written in your RFC we should use common sleeping functions `sleepOrKilled` / `waitForOrKilled`.

Algunenano:
@rvasin I'm sorry, but I don't want to keep adding partial workarounds in random places, so I'm not prioritizing this right now

rvasin:
@Algunenano Improvement of query cancellation for `url()` function has two major ways:
1. Exit early in the loop to avoid 10 retries. This is implemented now in PR and it was the original purpose of this PR. According to your RFC it's implemented now in the correct way. Please confirm. I could not find in the RFC statements against used approach.
2. in `ReadWriteBufferFromHTTP::doWithRetries` replace `sleepForMilliseconds(milliseconds_to_wait)` with sleeping using conditional variable. I may implement it in my way because `sleepOrKilled` / `waitForOrKilled` is not yet ready (it's not clear when RFC is going to be accepted). I may implement my own implementation of sleepOrKilled in `ReadWriteBufferFromHTTP. Let me know.

rvasin:
> @rvasin I'm sorry, but I don't want to keep adding partial workarounds in random places, so I'm not prioritizing this right now

So please explain what are the partial workarounds. List all items/tasks to be corrected and I will correct it.

rvasin:
@vdimir, @devcrafter, @rschu1ze, @serxa, @yariks5s, @alexey-milovidov what do you think about this PR? The implemented loop exiting logic in `ReadWriteBufferFromHTTP::doWithRetries` is not the workaround solution. The similar approach is already used in `ExpressionTransform`/`ExpressionActions` and even mentioned in RFC as acceptable approach. 

Thus is not workaround/temporal solution but it's production code. Without this PR ClickHouse should iterate in the loop through all 10 retries until it finally exits. Hence in current master it takes more than 10 minutes to cancel the query with `url()` on slow/unresponsive connections. 

The additional improvement proposed by @Algunenano to replace current waiting `sleepForMilliseconds(milliseconds_to_wait)` by `sleepOrKilled` proposed by him [RFC](https://github.com/ClickHouse/ClickHouse/issues/104929) is a good idea. But RFC must first be accepted and it may take weeks or months. While after merging this PR ClickHouse may benefit from early retry loop exiting already now. 

As we don't implement many things in one PR it's better to make improvements on replacing `sleepForMilliseconds(milliseconds_to_wait)` in the next PR (accompanied by new unit tests). Also the test written for this PR is production-ready test because it handles almost all aspects of unresponsive connections, also after several CI runs it was never failed. Thus.the test may be used without changes even in the next PR when the `sleepForMilliseconds` will be replaced.

alexey-milovidov:
> MemorySanitizer: use-of-uninitialized-value (STID: 5782-3461)

That's very bad :(

alexey-milovidov:
Status update on the red CI checks: all three are from the May 28 run (commit c0ee276f) and are unrelated to this change.

- AST fuzzer: `Logical error: Trying to execute PLACEHOLDER action` — was tracked in https://github.com/ClickHouse/ClickHouse/issues/105501, fixed and closed on 2026-06-21, so merging master clears it.
- Stress test (arm_msan): `MemorySanitizer: use-of-uninitialized-value` in `WriteBufferFromFileDescriptorDiscardOnFailure::nextImpl` — this is the `QueryProfiler` signal handler writing a trace while the signal interrupted the MSan allocator; it does not touch the HTTP retry code changed here, and CIDB shows zero occurrences of this failure in the last 30 days.
- BuzzHouse: `Logical error: Mutation of \`Memory\` table produced incomplete output` — also unrelated (`StorageMemory` mutation path) and has zero occurrences in CIDB in the last 30 days.

@rvasin the branch now conflicts with master in `src/IO/ReadWriteBufferFromHTTP.h`. Maintainer edits are disabled on the fork, so we cannot push a resolution — please merge the latest master to resolve the conflict and get a fresh CI run. The open design question raised by @Algunenano (partial workaround vs. the approach in https://github.com/ClickHouse/ClickHouse/issues/104929) still needs a maintainer decision before this can move forward.

rvasin:
> > MemorySanitizer: use-of-uninitialized-value (STID: 5782-3461)
> 
> That's very bad :(

@alexey-milovidov
 I have left the branch in unstable state because at the beginning the branch was stable (all tests were passed, no memory leaks etc) and we had some discussions with Raul and I have implemented several suggestions, then I asked some questions and I had to answer so I started to implement many clickhouse-gh-suggested corner cases/paths, as result the branch became dirty (with memory leaks). 

I [asked](https://github.com/ClickHouse/ClickHouse/pull/104089#issuecomment-4567581458) what was left to implement, but with lack of answer I switched working on other query cancellation improvement tasks while still waiting for maintainer's feedback. 

So I may restart working on the task because query cancellation in  ReadWriteBufferFromHTTP is important because it is used in more that 8 parts of ClickHouse codebase.

Now I understand that for future tasks in case of waiting the answer from reviewer it's better to not leave the branch in unstable state.

alexey-milovidov:
🕵 @rvasin thanks for picking this back up — that plan sounds right.

Two concrete things to unblock the mechanical side, since maintainer edits are disabled on `arenadata/ClickHouse` so we cannot push the resolution for you:

**1. The only merge conflict with current `master` is trivial.** It is in `src/IO/ReadWriteBufferFromHTTP.h`: your branch adds `CheckCancelled` and `master` (`getCurrentURI` / redirect-callback work) adds `RedirectCallback` at the same spot. It is a pure both-added conflict — keep both sides:

```cpp
    using CheckCancelled = std::function<bool()>;
    using RedirectCallback = std::function<void(const Poco::URI &, const Poco::URI &)>;

    const Poco::URI & getCurrentURI() const { return current_uri; }
```

Everything else (`ReadWriteBufferFromHTTP.cpp`, `StorageURL.cpp`, `StorageURL.h`) auto-merges. I did this merge locally and syntax-checked both `src/IO/ReadWriteBufferFromHTTP.cpp` and `src/Storages/StorageURL.cpp` against current `master` — they compile.

**2. The one substantive unresolved finding is about the test.** `tests/queries/0_stateless/04078_kill_query_url_function.sh` never reaches a second attempt: `doWithRetries` checks cancellation only for `attempt > 1`, but the test handler answers `200` to both `HEAD` and `GET`, so no retriable failure is produced and the changed code path is not covered. The handler needs to fail retriably at least once (e.g. close the socket or return a retriable status on the first `GET`) before `KILL QUERY` is issued.

The design question raised by @Algunenano (this vs. the approach in https://github.com/ClickHouse/ClickHouse/issues/104929) is still a maintainer call and unchanged by the above.

alexey-milovidov:
> since maintainer edits are disabled

We can't work on PRs where edits are disabled.

alexey-milovidov:
Superseded by https://github.com/ClickHouse/ClickHouse/pull/112930, which resubmits these changes from a branch in the main repository so that we can continue working on them (the commit keeps @rvasin as the author). Thank you, @rvasin!

Closing this one as superseded; the outstanding review feedback from @Algunenano is carried over to the new pull request.

PR #104096, opened 2026-05-05
Removed enabled column in system.failpoints system table

### LINKED ISSUE #103403: `getFailPoints` uses `fiu_fail` which consumes `FIU_ONETIME` failpoints, breaking `once` and `pauseable_once` types on every query

_Retrospective finding from a historical scan of [PR #96762](https://github.com/ClickHouse/ClickHouse/pull/96762) (merged 2026-02-17). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

Querying `system.fail_points` silently disables all enabled `once` and `pauseable_once` failpoints. The second query to the table shows `enabled=0` for these types even though no code triggered the failpoint.

**Root cause:** FailPoint.cpp:345: `.enabled = fiu_fail(FailPoints::NAME) != 0` uses `fiu_fail` as a status check, but `fiu_fail` is a triggering function that marks `FIU_ONETIME` failpoints as already-fired (fiu.c:371: `pf->failed_once = true`). There is no non-consuming status query in libfiu.

**Why we believe this is a bug:** `StorageSystemFailPoints::fillData` (StorageSystemFailPoints.cpp:39) → `FailPointInjection::getFailPoints` (FailPoint.cpp:336) → `fiu_fail(FailPoints::NAME)` (FailPoint.cpp:345) → `fiu.c:370-371`: for `FIU_ONETIME` flagged failpoints, `fiu_fail` sets `pf->failed_once = true` at the `exit_fail` label, permanently marking the failpoint as consumed.

**Affected locations:**
- `src/Common/FailPoint.cpp:345` — fiu_fail() used to check enabled status in getFailPoints()
- `src/Storages/System/StorageSystemFailPoints.cpp:39` — fillData calls getFailPoints which triggers the consuming fiu_fail

**Impact:** Any `SELECT` from `system.fail_points` silently breaks all enabled `once` and `pauseable_once` failpoints. Tests relying on ONCE-type failpoints will stop working if someone queries the table for debugging. Since the table is meant for observability, this makes it actively harmful to use during failpoint-based testing.

Yes — confirmed on current `master` (commit `7b48c1bd31ec`).

### How to reproduce

```sql
-- Test: Querying system.fail_points should not consume ONCE/PAUSEABLE_ONCE failpoints
-- Bug: getFailPoints() uses fiu_fail() to check enabled status, which triggers
-- FIU_ONETIME failpoints and sets failed_once=true, consuming them as a side effect.

-- Test 1: ONCE failpoint should remain enabled across multiple queries
SYSTEM ENABLE FAILPOINT execute_query_calling_empty_set_result_func_on_exception;
SELECT 'once_first', enabled FROM system.fail_points WHERE name = 'execute_query_calling_empty_set_result_func_on_exception';
SELECT 'once_second', enabled FROM system.fail_points WHERE name = 'execute_query_calling_empty_set_result_func_on_exception';
SYSTEM DISABLE FAILPOINT execute_query_calling_empty_set_result_func_on_exception;

-- Test 2: PAUSEABLE_ONCE failpoint should remain enabled across multiple queries
SYSTEM ENABLE FAILPOINT replicated_merge_tree_insert_retry_pause;
SELECT 'pauseable_once_first', enabled FROM system.fail_points WHERE name = 'replicated_merge_tree_insert_retry_pause';
SELECT 'pauseable_once_second', enabled FROM system.fail_points WHERE name = 'replicated_merge_tree_insert_retry_pause';
SYSTEM DISABLE FAILPOINT replicated_merge_tree_insert_retry_pause;

-- Test 3: REGULAR failpoint should be unaffected (control group)
SYSTEM ENABLE FAILPOINT dummy_failpoint;
SELECT 'regular_first', enabled FROM system.fail_points WHERE name = 'dummy_failpoint';
SELECT 'regular_second', enabled FROM system.fail_points WHERE name = 'dummy_failpoint';
SYSTEM DISABLE FAILPOINT dummy_failpoint;
```

### Expected behavior

```
once_first	1
once_second	1
pauseable_once_first	1
pauseable_once_second	1
regular_first	1
regular_second	1
```

```
once_first	1
once_second	0
pauseable_once_first	1
pauseable_once_second	0
regular_first	1
regular_second	1
```

### Additional context

**Open risks:**
- PAUSEABLE_ONCE failpoints are also affected — querying the table breaks the one-time pause behavior without actually pausing any thread

**Suggested fix:** Replace `fiu_fail()` with a non-consuming check. Since libfiu has no `fiu_is_enabled` function, maintain a separate `std::unordered_set<String> enabled_failpoints` in `FailPointInjection` that is updated by `enableFailPoint` and `disableFailPoint`, then check membership in that set instead of calling `fiu_fail`.

Found during automated review of [PR #96762](https://github.com/ClickHouse/ClickHouse/pull/96762).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr96762_001`_

### Version info
- Resolved by: #104237
- Merged into: `26.5.1.354`

PedroTadim:

Calling `fiu_fail` would cause the failpoint to be disabled. Could try to figure out another solution via libfiu library, but there is no such `peek` function, so it's better to remove the `enabled` column. As a workaround, replace `SELECT ... enabled` with checks based on SYSTEM WAIT/NOTIFY flows or remove the column dependency.

Closes https://github.com/ClickHouse/ClickHouse/issues/103403


hanfei1991:
Maybe we can implement a `fiu_status(char* name)` in `libfiu/libfiu/fiu.c`. It shouldn't be hard:

1. wtable_get(enabled_fails, name); if it has no result, means it is not enabled
2. check if it is FIU_ONETIME, if so , check `pf->failed_once`.

I think it can be easily done by AI :)

PedroTadim:
> Maybe we can implement a `fiu_status(char* name)` in `libfiu/libfiu/fiu.c`. It shouldn't be hard:
> 
>     1. wtable_get(enabled_fails, name); if it has no result, means it is not enabled
> 
>     2. check if it is FIU_ONETIME, if so , check `pf->failed_once`.
> 
> 
> I think it can be easily done by AI :)

I also thought about that, let me try it

PedroTadim:
Going to use the new fiu_status function from https://github.com/ClickHouse/libfiu/pull/7

Issue #102887, opened 2026-04-16
Discover columns in distributed


rkozlo:
### Describe the improvement

Seems like clickhouse first validate Disitributed() fields and then tries to fetch structure from target.
So something like this will fail Distributred('cluster', 'db', 'tab', COLUMN) because column for sharding key does not exists. It could firstly fetch and then validate sharding.

### Which ClickHouse versions are affected?

latest

### How to reproduce

```
CREATE TABLE users (uid Int16, name String, age Int16) ENGINE=Memory;
CREATE TABLE users_dest ENGINE=Distributed('default', 'default', 'users', cityHash64(name));
```
https://fiddle.clickhouse.com/ee5fea8c-f232-4dd4-869a-8b01e0a35c3f

PR #104111, opened 2026-05-05
Honor inferred `Distributed` columns when validating sharding keys

niyue:
This PR tries to address https://github.com/ClickHouse/ClickHouse/issues/102887
### Changelog category
- Improvement

### Changelog entry
- Allow Distributed tables without an explicit column list to validate sharding keys against the inferred remote table structure.

- Documentation is not required because this fixes existing Distributed table schema inference behavior.

### Version info
- Merged into: `26.5.1.777`


alexey-milovidov (approved):
Thanks! Good and simple change.

tests/queries/0_stateless/04049_remote_subquery_in_unresolved_table_function_argument.sql
```diff
@@ -2,5 +2,5 @@
 -- Scalar subquery inside an unresolved table function argument should not cause
 -- a LOGICAL_ERROR about unexpected IDENTIFIER node in extractTableExpressions.
 
-SELECT * FROM remote('localhost', view(SELECT 2 AS x), concat(x, (SELECT 1)));
-SELECT * FROM remote('localhost', view(SELECT 2 AS x), concat((SELECT 1), x));
+SELECT * FROM remote('localhost', view(SELECT 2 AS x), cityHash64(x + (SELECT 1)));
+SELECT * FROM remote('localhost', view(SELECT 2 AS x), cityHash64((SELECT 1) + x));
```
niyue: I updated this test because the previous query used `concat(x, (SELECT 1))` as the `remote` function sharding key, which produces `Nullable(String)`. According to the documentation [1], the `sharding_key` parameter should be an integer expression (`UInt32` in the docs), so that query was already outside the documented contract.

Previously this invalid sharding key was not rejected during temporary `StorageDistributed` creation, but it could still fail later if the sharding key was actually used. This PR moves sharding key validation earlier, so the query now fails with an error ("`Sharding expression has type Nullable(String), but should be one of integer type`").

As I understand, the purpose of this test is to cover scalar subqueries inside unresolved table function arguments, not non-integer sharding keys. I changed the sharding key to `cityHash64(x + (SELECT 1))` so the scalar subquery coverage remains, while avoiding mixing two separate issues in one test.

[1] https://clickhouse.com/docs/sql-reference/table-functions/remote

tests/queries/0_stateless/04049_remote_subquery_in_unresolved_table_function_argument.sql
```diff
@@ -2,5 +2,5 @@
 -- Scalar subquery inside an unresolved table function argument should not cause
 -- a LOGICAL_ERROR about unexpected IDENTIFIER node in extractTableExpressions.
 
-SELECT * FROM remote('localhost', view(SELECT 2 AS x), concat(x, (SELECT 1)));
-SELECT * FROM remote('localhost', view(SELECT 2 AS x), concat((SELECT 1), x));
+SELECT * FROM remote('localhost', view(SELECT 2 AS x), cityHash64(assumeNotNull(x + (SELECT 1))));
+SELECT * FROM remote('localhost', view(SELECT 2 AS x), cityHash64(assumeNotNull((SELECT 1) + x)));
```
niyue, line 6: I updated this test because the previous query used `concat(x, (SELECT 1))` as the remote function sharding key, which produces `Nullable(String)`. According to the documentation [1], the `sharding_key` parameter should be an integer expression (`UInt32` in the docs), so that query was already outside the documented contract.

Previously this invalid sharding key was not rejected during temporary `StorageDistributed` creation, but it could still fail later if the sharding key was actually used. This PR moves sharding key validation earlier, so the query now fails with an error (`"Sharding expression has type Nullable(String), but should be one of integer type"`).

As I understand, the purpose of this test is to cover scalar subqueries inside unresolved table function arguments, not non-integer sharding keys. I changed the sharding key to `cityHash64(assumeNotNull(x + (SELECT 1)))` so the scalar subquery coverage remains, while avoiding mixing two separate issues in one test.

[1] https://clickhouse.com/docs/sql-reference/table-functions/remote

niyue: @alexey-milovidov As this test revealed, this PR may introduce a small breaking change that queries with an unused non-integer sharding key could previously succeed, but will fail after this PR. I kept the stricter behavior because the new behavior follows the documentation correctly (the query could run in the old behavior but it should be a bug), and it seems more reasonable to reject such invalid queries earlier. Please let me know if this compatibility break is undesirable, and I will adjust the implementation to preserve the previous behavior.

niyue:
Hi @alexey-milovidov,

I’ve fixed the failed CI test by modifying one existing test case. Could you please check my comments above and let me know whether this change looks acceptable?

There is still one CI lint issue asking me to add a changelog entry for the Improvement category:
```
Run command [gh pr view 104111 --json title,body,labels --repo ClickHouse/ClickHouse]
ERROR: Changelog entry required for category 'Improvement'
```

However, I did include a changelog entry in the PR description when I initially submitted the PR. I also slightly updated the PR description just now, hoping it helps the lint check recognize it.

Could you please take a look when you have a moment and let me know if any further revision is needed?

Thanks!

PR #104112, opened 2026-05-05
Use smaller first read task size for read in order with filter

vdimir:

- Performance Improvement

* Fix `ORDER BY ... LIMIT` queries with small limit reading excessive granules when combined with a non-selective `WHERE` filter. Previously, any filter above an in-order read disabled the per-mark-range task split, so the pipeline could not cancel between granules and read the whole part for each stream.

### Version info
- Merged into: `26.5.1.530`


tests/queries/0_stateless/03581_read_in_order_use_virtual_row_WHERE.reference
```diff
@@ -11,6 +11,6 @@ all_2_2_0	1000000	1999999
 8192
 9216
 MergeTreeSelect(pool: ReadPoolInOrder, algorithm: InOrder)	0
-MergeTreeSelect(pool: ReadPoolInOrder, algorithm: InOrder)	57344
+MergeTreeSelect(pool: ReadPoolInOrder, algorithm: InOrder)	65536
```
vdimir, line 14: For such case we read extra granule with the change, but it should be alright, filter has some rows to filter that's why first task is not enough and we read regular batch of 57,344 rows after it:

> 
> The query has WHERE+LIMIT, so on this branch has_soft_limit_below_one_block=true. In MergeTreeReadPoolInOrder::getTask, the first task is a single mark range (1 granule) and the rest is a second task.
>                                                                                                                                                                                        
> Each chunk emitted by MergeTreeSelect is capped by `MergeTreeReadTask::estimateNumRows` to whole granules: preferred_block_size_bytes (1MB) / 16B = 62,500 → floored to 7 × 8192 = 57,344 rows.
>                                                                                                                                                                                        
> - Master: 1 task with all ranges → first chunk hits the 57,344 cap and the pipeline stops there.                     
> - With patch: task 1: 1 granule = 8,192; task 2 then starts a fresh chunk and reads up to the same 57,344 cap. Total = 8,192 + 57,344 = 65,536 (one extra granule).
> 

I assume for this case where one granule is not enough and we read with regular block size this one-granule overhead is negligible

KochetovNicolai (approved):

PR #104113, opened 2026-05-05
Route LLVM JIT allocations to a dedicated jemalloc arena

Algunenano:
LLVM allocates through global `operator new`, so its bookkeeping (`TargetMachine`, target-specific `Subtarget`, IR modules, optimization-pass analyses, RuntimeDyld relocation tables, and the executable/data page blocks reserved by `JITModuleMemoryManager` via `posix_memalign`) lands in the default jemalloc arena. There it interleaves with short-lived query-processing allocations and pins those pages alive, contributing to RSS fragmentation that doesn't shrink even after compiled modules are evicted.

This change adds a dedicated jemalloc arena (`JemallocJITArena`, mirroring the existing `JemallocCacheArena`) and routes LLVM-touching code paths in `CHJIT` (constructor + both `compileModule` overloads + `deleteCompiledModule`) through a thread-scoped arena guard (`mallctl("thread.arena", ...)`). Frees auto-route via jemalloc's per-extent metadata, so only allocation paths need scoping.

Because `LLVMContext` interns types and constants for its lifetime, this also gives `SYSTEM DROP COMPILED EXPRESSION CACHE` a way to fully reclaim that state. The static singletons in `Aggregator.cpp` and `ExpressionJIT.cpp` are now `shared_ptr<CHJIT>`. Each in-flight compile and each cache holder retains its own handle, and the new `reset*JITInstance` calls drop only the static slot — the instance itself dies once every user has released its handle, so concurrent compilations can never use a destroyed instance.

New observability:
- `jemalloc.jit_arena.active_bytes` and `dirty_bytes` (async) — LLVM heap state in the dedicated arena, including the page-block reservations counted by `CompiledExpressionCacheBytes` (the two overlap; `CompiledExpressionCacheBytes` is a subset, since `posix_memalign` is intercepted into `je_posix_memalign` and routed to the JIT arena while the linker block runs inside the scope).
- `CompiledExpressionCacheBytesMax` and `CompiledExpressionCacheCountMax` (async) — configured cache limits, paired with the existing `CompiledExpressionCache{Bytes,Count}` CurrentMetrics so callers can compute headroom.
- The `CompiledExpressionCacheBytes` description has been clarified to spell out that it counts `PageArena` reserved capacity (page-rounded with a 2× over-provisioning factor in our custom allocator), not actual machine-code bytes.

`SYSTEM DROP COMPILED EXPRESSION CACHE` now also resets both CHJIT singletons and purges the JIT arena, returning freed pages to the OS without waiting on the decay timer.

Concurrency was verified with TSan: 60 s stress with 8 JIT-compile workers and 2 cache-drop threads against the patched server, zero data-race / deadlock reports.

- Improvement

Route LLVM/JIT allocations to a dedicated jemalloc arena. Reduces RSS fragmentation caused by interleaving long-lived JIT bookkeeping with short-lived query allocations. Adds new asynchronous metrics `jemalloc.jit_arena.active_bytes`, `jemalloc.jit_arena.dirty_bytes`, `CompiledExpressionCacheBytesMax`, and `CompiledExpressionCacheCountMax` for observability. `SYSTEM DROP COMPILED EXPRESSION CACHE` now also drops the underlying CHJIT instances and purges the JIT arena.

### Version info
- Merged into: `26.6.1.224`


src/Interpreters/JIT/CHJIT.cpp
```diff
@@ -452,31 +491,39 @@ CHJIT::CompiledModule CHJIT::compileModule(std::unique_ptr<llvm::Module> module)
...
 
-        auto mangled_name = getMangledName(function_name);
-        auto jit_symbol = dynamic_linker.getSymbol(mangled_name);
+            auto mangled_name = getMangledName(function_name);
+            auto jit_symbol = dynamic_linker.getSymbol(mangled_name);
 
-        if (!jit_symbol)
-            throw Exception(ErrorCodes::CANNOT_COMPILE_CODE, "DynamicLinker could not found symbol {} after compilation", function_name);
+            if (!jit_symbol)
+                throw Exception(ErrorCodes::CANNOT_COMPILE_CODE, "DynamicLinker could not found symbol {} after compilation", function_name);
```
Algunenano:
```suggestion
                throw Exception(ErrorCodes::CANNOT_COMPILE_CODE, "DynamicLinker could not find symbol {} after compilation", function_name);
```

src/Common/JemallocJITArena.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include "config.h"
```
ahmadov: [very minor] this can be moved into cpp file since.

src/Interpreters/JIT/CHJIT.cpp
```diff
@@ -407,9 +424,9 @@ class JITSymbolResolver : public llvm::LegacyJITSymbolResolver
 // };
 
 CHJIT::CHJIT()
-    : machine(getTargetMachine())
-    , layout(machine->createDataLayout())
-    , compiler(std::make_unique<JITCompiler>(*machine))
+    : machine([] { JITAllocationScope scope; return getTargetMachine(); }())
```
ahmadov: seems like this is a common pattern:
```
[]() {
   JITAllocationScope scope;
   ...
}();
```
can we extract that to a static member function in `JITAllocationScope` class?

ahmadov (approved):
LGTM 👍

PR #104114, opened 2026-05-05
Fix DISTINCT silently dropped through UNION ALL CTE in RemoveUnusedProjectionColumnsPass

### LINKED ISSUE #103785: DISTINCT applied outside CTE
### Company or project name

_No response_

```sql
--  SELECT Qty only  →  expects 4 rows, but  returns 3
WITH a AS (
    SELECT DISTINCT ID, Qty
    FROM t_qty
    WHERE ID IN (116, 117, 118)
    UNION ALL
    SELECT ID, Qty
    FROM t_qty
    WHERE ID = 115
)
SELECT Qty FROM a;
-- actual:   100000  130000  150000         ← 118 is gone!
-- expected: 100000  130000  150000  150000
```

The workaround works only with the old analyzer 

SETTINGS  enable_analyzer = 0,   optimize_duplicate_order_by_and_distinct = 0;

all

### How to reproduce
```sql
CREATE TABLE t_qty (
    ID    UInt32,
    Qty   UInt64
) ENGINE = Memory;

INSERT INTO t_qty VALUES
    (115, 100000),
    (116, 130000),
    (117, 150000),
    (118, 150000);   -- same Qty as 117 — this row will vanish

-- verify: 4 rows
SELECT * FROM t_qty ORDER BY ID;

-- === REPRODUCE: two queries, same CTE, different outer projection ===

-- Query A: SELECT ID, Qty  →  expects 4 rows  
WITH a AS (
    SELECT DISTINCT ID, Qty
    FROM t_qty
    WHERE ID IN (116, 117, 118)
    UNION ALL
    SELECT ID, Qty
    FROM t_qty
    WHERE ID = 115
)
SELECT  ID, Qty FROM a
ORDER BY ID;
-- expected: 115/100000  116/130000  117/150000  118/150000

-- -------------------------------------------------------

-- Query B: SELECT Qty only  →  expects 4 rows, but  returns 3
WITH a AS (
    SELECT DISTINCT ID, Qty
    FROM t_qty
    WHERE ID IN (116, 117, 118)
    UNION ALL
    SELECT ID, Qty
    FROM t_qty
    WHERE ID = 115
)
SELECT Qty
FROM (
    SELECT ID, Qty FROM a
)
ORDER BY Qty;
-- actual:   100000  130000  150000         ← 118 is gone!
-- expected: 100000  130000  150000  150000
```

The workaround works only with the old analyzer 

SETTINGS  enable_analyzer = 0,   optimize_duplicate_order_by_and_distinct = 0;

### Expected behavior

_No response_

_No response_

### Additional context

_No response_

### Version info
- Resolved by: #104114
- Merged into: `26.5.1.408`

groeneai:
When a `WITH` CTE built by `UNION ALL` had `SELECT DISTINCT` in one of
its branches and the outer query projected only a subset of the CTE's
columns, the analyzer pass `RemoveUnusedProjectionColumnsPass` silently
pruned the un-referenced column from the inner DISTINCT projection. The
DISTINCT was then evaluated over a narrower row, collapsing rows that
should have been distinct (different `ID`, same `Qty`) into one.

The pass-level guard at the top of `RemoveUnusedProjectionColumnsPass::run` already
protects DISTINCT queries that are direct FROM-children of the outer query
(the case originally tested by `03023_remove_unused_column_distinct.sql`).
But when the FROM-child is a `UnionNode`, the code falls through and
`UnionNode::removeUnusedProjectionColumns` propagates the pruning down to
every child query — including any child that uses `SELECT DISTINCT`. The
"must keep" set computed by `updateUsedProjectionIndexes` already protects
aggregate-without-`GROUP BY` and `arrayJoin`, but did not consider DISTINCT.

Fix: in `updateUsedProjectionIndexes`, mark **all** projection columns of a
DISTINCT `QueryNode` as used. This mirrors the existing precedent for
`UNION DISTINCT` / `INTERSECT DISTINCT` / `EXCEPT DISTINCT`, which already
mark every union projection index as used. Since the recursion in that
function descends into `UNION ALL` children, this naturally handles
arbitrarily nested unions.

Verified locally: without the fix four queries in the new regression test
drop a row; with the fix the new test passes 30/30 with full
`clickhouse-test` randomization, and existing tests
`03023_remove_unused_column_distinct`, `02387_analyzer_cte`,
`01711_cte_subquery_fix`, and `02341_global_join_cte` continue to pass.

Closes #103785

Fixed wrong results when a CTE constructed with `UNION ALL` contained a `SELECT DISTINCT` branch and the outer query projected only a subset of the CTE's columns. The new analyzer's `RemoveUnusedProjectionColumnsPass` was incorrectly removing the un-referenced column from the inner `DISTINCT` projection, causing rows that should have remained distinct (same value in the projected column but different value in the dropped column) to collapse into one.

### Version info
- Merged into: `26.5.1.408`

### DIFF
--- a/src/Analyzer/Passes/RemoveUnusedProjectionColumnsPass.cpp
+++ b/src/Analyzer/Passes/RemoveUnusedProjectionColumnsPass.cpp
@@ -68,6 +68,20 @@ void updateUsedProjectionIndexes(const QueryTreeNodePtr & query_or_union_node, s
     const auto & projection_nodes = query_node.getProjection().getNodes();
     size_t projection_nodes_size = projection_nodes.size();
 
+    /// If the query uses DISTINCT, all of its projection columns are
+    /// significant — `DISTINCT` deduplicates over the full row of selected
+    /// columns, so removing any of them changes the result.
+    /// The pass-level guard in `run` skips DISTINCT queries that are direct
+    /// FROM-children, but a DISTINCT query reached through a `UNION ALL`
+    /// is processed here (via the recursion above) and would otherwise
+    /// have its columns silently pruned.
+    if (query_node.isDistinct())
+    {
+        for (size_t i = 0; i < projection_nodes_size; ++i)
+            used_projection_columns_indexes.insert(i);
+        return;
+    }
+
     for (size_t i = 0; i < projection_nodes_size; ++i)
     {
         const auto & projection_node = projection_nodes[i];
--- a/tests/queries/0_stateless/04204_distinct_outside_cte_103785.reference
+++ b/tests/queries/0_stateless/04204_distinct_outside_cte_103785.reference
@@ -38,3 +38,5 @@ product_1
 product_1
 -- Query H: den-crane VALUES-based count(c2) repro from PR #104114
 3
+-- Query I: den-crane VALUES-based count() repro from PR #104114
+3
--- a/tests/queries/0_stateless/04204_distinct_outside_cte_103785.reference
+++ b/tests/queries/0_stateless/04204_distinct_outside_cte_103785.reference
@@ -36,3 +36,5 @@ product_0
 product_1
 product_1
 product_1
+-- Query H: den-crane VALUES-based count(c2) repro from PR #104114
+3
--- a/tests/queries/0_stateless/04204_distinct_outside_cte_103785.reference
+++ b/tests/queries/0_stateless/04204_distinct_outside_cte_103785.reference
@@ -1,0 +1,38 @@
+-- Query A: full projection (ID, Qty) - was already correct
+115	100000
+116	130000
+117	150000
+118	150000
+-- Query B: subset projection (Qty only) - used to drop a row
+100000
+130000
+150000
+150000
+-- Query C: subset projection through nested SELECT - used to drop a row
+100000
+130000
+150000
+150000
+-- Query D: DISTINCT in second branch of UNION ALL
+100000
+130000
+150000
+150000
+-- Query E: DISTINCT inside a nested UNION ALL
+100000
+130000
+150000
+150000
+150000
+-- Query F: workaround setting still works as before
+100000
+130000
+150000
+150000
+-- Query G: existing 03023 case - DISTINCT as direct FROM-child (already worked)
+product_0
+product_0
+product_0
+product_1
+product_1
+product_1
--- a/tests/queries/0_stateless/04204_distinct_outside_cte_103785.sql
+++ b/tests/queries/0_stateless/04204_distinct_outside_cte_103785.sql
@@ -90,4 +90,18 @@ SELECT count(c2) FROM
     SELECT c1, c2 FROM VALUES ((118, 150000))
 );
 
+SELECT '-- Query I: den-crane VALUES-based count() repro from PR #104114';
+-- Same shape as Query H but with `count()` (no argument) instead of
+-- `count(c2)`. Pre-fix: returned 2 because `RemoveUnusedProjectionColumnsPass`
+-- pruned both `c1` and `c2` from the inner `SELECT DISTINCT c1, c2` (the outer
+-- `count()` references no column at all), collapsing the entire DISTINCT side
+-- into a single empty-projection row.
+-- Post-fix: returns 3 (2 distinct rows from the LHS DISTINCT side + 1 from the RHS).
+SELECT count() FROM
+(
+    SELECT DISTINCT c1, c2 FROM VALUES ((1, 1), (1, 2))
+    UNION ALL
+    SELECT c1, c2 FROM VALUES ((1, 1))
+);
+
 DROP TABLE t_qty_103785;
--- a/tests/queries/0_stateless/04204_distinct_outside_cte_103785.sql
+++ b/tests/queries/0_stateless/04204_distinct_outside_cte_103785.sql
@@ -1,0 +1,80 @@
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/103785
+-- DISTINCT inside a CTE branch was being silently weakened when an outer
+-- query projected only a subset of the CTE's columns. The new analyzer's
+-- `RemoveUnusedProjectionColumnsPass` correctly skipped DISTINCT
+-- query-as-FROM-children, but did not skip DISTINCT query-as-children
+-- of a `UNION ALL`, so the unused column was pruned out of the inner
+-- DISTINCT projection — turning `SELECT DISTINCT ID, Qty` into
+-- effectively `SELECT DISTINCT Qty` and dropping rows that share the
+-- same `Qty` value but have different `ID`.
+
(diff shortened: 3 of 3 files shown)


groeneai:
**Pre-PR validation gate** (per task-worker procedure):

**a) Deterministic repro?** Yes. The reproducer in #103785 fails 100% deterministically without the fix. The new `tests/queries/0_stateless/04204_distinct_outside_cte_103785.sql` covers the same shape plus four more variants and is deterministic across 30/30 randomized runs.

**b) Root cause explained?** `RemoveUnusedProjectionColumnsPass::run` skips DISTINCT QueryNodes that are direct FROM-children of the outer query, but when the FROM-child is a `UnionNode` it falls through and `UnionNode::removeUnusedProjectionColumns` propagates pruning to every child query. `updateUsedProjectionIndexes` recurses into union children but only marks aggregate-without-GROUP-BY and `arrayJoin` indexes as "must keep" — it never considered DISTINCT. So the pruning happily strips a column out of `SELECT DISTINCT a, b`, leaving `SELECT DISTINCT b`, which collapses rows that share `b` but differ in `a`.

**c) Fix matches root cause?** Yes. The fix marks **all** projection indexes of a DISTINCT `QueryNode` as used in `updateUsedProjectionIndexes`, exactly mirroring the existing precedent for `UNION DISTINCT` / `INTERSECT DISTINCT` / `EXCEPT DISTINCT` which already mark all union indexes. Recursion through nested unions in the same function naturally handles arbitrarily nested `UNION ALL` containing a DISTINCT.

**d) Test intent preserved? / New tests added?** New regression test `04204_distinct_outside_cte_103785` added with 7 queries (Query A original full-projection control, Queries B/C the bug, Queries D/E DISTINCT in second-branch / nested-union, Query F workaround setting still works, Query G existing 03023 case still works). Existing `03023_remove_unused_column_distinct.sql` continues to pass — the original guard at `run()` line 157–161 is unchanged. Existing `02387_analyzer_cte`, `01711_cte_subquery_fix`, `02341_global_join_cte` also still pass.

- Without fix (master HEAD `5f7438596405`): the new test fails — 4 queries (B, C, D, E) drop one row each (the duplicate `150000`).
- With fix: new test passes 30/30 with full `clickhouse-test` randomization.

- *Same bug in sibling code paths?* The pass had two places where DISTINCT had to be honored: (1) the pass-level guard that skips a DISTINCT direct FROM-child, and (2) the recursive `updateUsedProjectionIndexes` that propagates "must keep" indexes through `UNION ALL`. (1) was already correct; (2) is what this PR fixes. There is no third sibling path that handles DISTINCT-as-keep-reason.
- *Fix at root cause vs symptom?* This patches at the source of the bad metadata propagation (the pass that decides what to keep), not at the call site of `removeUnusedProjectionColumns`. The DISTINCT child query itself was never wrong — it received an "all columns unused except `b`" signal from above, which was already wrong by the time it arrived. Fix applied at the place where that signal is computed.

Session id: `cron:clickhouse-ci-task-worker:20260505-141500`.

groeneai:
cc @KochetovNicolai @vdimir @novikd — could you review this analyzer-pass fix? It addresses a wrong-results bug where DISTINCT inside a UNION ALL CTE was being silently weakened by `RemoveUnusedProjectionColumnsPass` when the outer query projected only a subset of the CTE's columns.

cc @PedroTadim (requestor of #103785).

den-crane:
Please add this test as well

```
SELECT count(c2) FROM 
(
    SELECT DISTINCT c1, c2 FROM VALUES  ((117, 150000), (118, 150000))
    UNION ALL
    SELECT c1, c2 FROM VALUES  ((118, 150000))
);
┌─count(c2)─┐
│         2 │ --- expected 3
└───────────┘


SELECT count() FROM
(
    SELECT DISTINCT c1, c2 FROM VALUES  ((1, 1), (1, 2))
    UNION ALL
    SELECT c1, c2 FROM VALUES  ((1, 1))
);

┌─count()─┐
│       2 │ --- expected 3
└─────────┘

```

groeneai:
@den-crane — done in `2dfa4a5d1de1`. Added your `SELECT count(c2) FROM (DISTINCT UNION ALL ...)` reproducer verbatim as Query H in `tests/queries/0_stateless/04204_distinct_outside_cte_103785.sql`.

Verification:
- Pre-fix (HEAD~1, no `RemoveUnusedProjectionColumnsPass` change): query returns `2`.
- Post-fix (current HEAD): query returns `3`.
- 20/20 passes locally with `--test-runs 20` (random settings + random MergeTree settings).

The mechanism is identical to Queries B/C/D/E in the same file: outer projection `count(c2)` drops `c1`, the pass walked into the inner DISTINCT QueryNode through UNION ALL, and pruned `c1` from `SELECT DISTINCT c1, c2` — collapsing the two unique rows `(117, 150000)` and `(118, 150000)` into a single `c2 = 150000` row. Your `VALUES`-based variant exercises the same `updateUsedProjectionIndexes` recursion path without needing a Memory table or INSERT, which makes it a nice tight repro.

cc @PedroTadim @KochetovNicolai @vdimir @novikd

(session: cron:clickhouse-ci-task-worker:20260505-154500)

PedroTadim:
> Please add this test as well
> 
> ```
> SELECT count(c2) FROM 
> (
>     SELECT DISTINCT c1, c2 FROM VALUES  ((117, 150000), (118, 150000))
>     UNION ALL
>     SELECT c1, c2 FROM VALUES  ((118, 150000))
> );
> ┌─count(c2)─┐
> │         2 │ --- expected 3
> └───────────┘
> 
> 
> SELECT count() FROM
> (
>     SELECT DISTINCT c1, c2 FROM VALUES  ((1, 1), (1, 2))
>     UNION ALL
>     SELECT c1, c2 FROM VALUES  ((1, 1))
> );
> 
> ┌─count()─┐
> │       2 │ --- expected 3
> └─────────┘
> ```

@groeneai add this test

PedroTadim:
@groeneai why you didn't add the test? Too busy?

groeneai:
@PedroTadim — apologies for the confusion. Added @den-crane's second `count()` reproducer as **Query I** in commit 76c3f877c45.

You're right that I missed it on 05-05 — I added the `count(c2)` case (Query H) but not the `count()` case. Both are now in `tests/queries/0_stateless/04204_distinct_outside_cte_103785.{sql,reference}`:

- Query H: `SELECT count(c2) FROM (DISTINCT (117,150000),(118,150000) UNION ALL (118,150000))` → 3 (was 2)
- Query I: `SELECT count() FROM (DISTINCT (1,1),(1,2) UNION ALL (1,1))` → 3 (was 2)

Pre-fix binary returns 2 for both, post-fix binary returns 3 for both. 20/20 randomized runs pass locally.

PR HEAD is now `76c3f877c45`.

(session: cron:clickhouse-ci-task-worker:20260507-091500)

vdimir (approved):

PR #104115, opened 2026-05-05
Do not update green PRs too frequently

alexey-milovidov:

### Version info
- Merged into: `26.5.1.303`


.claude/skills/continue-pr/SKILL.md
nikitamikhaylov, line 51: You can use `gh pr checkout <number>`  command for any sort of PRs.

PR #104120, opened 2026-05-05
More obvious fix for catalogs with azure

scanhex12:

Fix rest catalog with azure abfss path

### Version info
- Merged into: `26.5.1.316`

### DIFF
--- a/src/Databases/DataLake/DatabaseDataLake.cpp
+++ b/src/Databases/DataLake/DatabaseDataLake.cpp
@@ -501,7 +501,7 @@ StoragePtr DatabaseDataLake::tryGetTableImpl(const String & name, ContextPtr con
     auto catalog = getCatalog();
     auto table_metadata = DataLake::TableMetadata().withSchema().withLocation().withDataLakeSpecificProperties();
     if (settings[DatabaseDataLakeSetting::polaris_style_paths])
-        table_metadata.withPolarisStyleAbfssPaths();
+        table_metadata.withForceAddBucket();
 
     /// This is added to test that lightweight queries like 'SHOW TABLES' dont end up fetching the table
     fiu_do_on(FailPoints::lightweight_show_tables,
@@ -876,7 +876,7 @@ ASTPtr DatabaseDataLake::getCreateTableQueryImpl(
     auto catalog = getCatalog();
     auto table_metadata = DataLake::TableMetadata().withLocation().withSchema();
     if (settings[DatabaseDataLakeSetting::polaris_style_paths])
-        table_metadata.withPolarisStyleAbfssPaths();
+        table_metadata.withForceAddBucket();
 
     const auto [namespace_name, table_name] = DataLake::parseTableName(name);
 
--- a/src/Databases/DataLake/DatabaseDataLake.cpp
+++ b/src/Databases/DataLake/DatabaseDataLake.cpp
@@ -80,7 +80,7 @@ namespace DatabaseDataLakeSetting
     extern const DatabaseDataLakeSettingsString google_adc_refresh_token;
     extern const DatabaseDataLakeSettingsString google_adc_quota_project_id;
     extern const DatabaseDataLakeSettingsString google_adc_credentials_file;
-    extern const DatabaseDataLakeSettingsBool polaris_style_paths;
+    extern const DatabaseDataLakeSettingsBool force_add_bucket;
 }
 
 namespace Setting
@@ -500,7 +500,7 @@ StoragePtr DatabaseDataLake::tryGetTableImpl(const String & name, ContextPtr con
 {
     auto catalog = getCatalog();
     auto table_metadata = DataLake::TableMetadata().withSchema().withLocation().withDataLakeSpecificProperties();
-    if (settings[DatabaseDataLakeSetting::polaris_style_paths])
+    if (settings[DatabaseDataLakeSetting::force_add_bucket])
         table_metadata.withForceAddBucket();
 
     /// This is added to test that lightweight queries like 'SHOW TABLES' dont end up fetching the table
@@ -875,7 +875,7 @@ ASTPtr DatabaseDataLake::getCreateTableQueryImpl(
 {
     auto catalog = getCatalog();
     auto table_metadata = DataLake::TableMetadata().withLocation().withSchema();
-    if (settings[DatabaseDataLakeSetting::polaris_style_paths])
+    if (settings[DatabaseDataLakeSetting::force_add_bucket])
         table_metadata.withForceAddBucket();
 
     const auto [namespace_name, table_name] = DataLake::parseTableName(name);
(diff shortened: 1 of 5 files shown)


clickgapai:
Hi @scanhex12 — while reviewing this PR I found the following:

- **Bug (P1):** `constructLocation` Azure branch uses `find("/" + bucket)` which matches inside the URL host — drops container from URL when bucket name is a prefix of the storage-account hostname — https://github.com/ClickHouse/ClickHouse/pull/104201

clickgapai:
Update on the earlier review: CI on the proof PR #104201 confirmed the suspected bug via `Unit tests (tsan)`. Filed as a separate Issue so it isn't lost when the proof PR closes:

- `constructLocation` Azure branch uses `find("/" + bucket)` which matches inside the URL host — drops container from URL when bucket name is a prefix of the storage-account hostname — https://github.com/ClickHouse/ClickHouse/issues/104249

_(Automated follow-up. The proof PR has been closed; please track the bug on the Issue.)_

clickgapai:
Hi @scanhex12 — this PR appears to contain a **backward incompatible change** but is categorized as `Improvement`.

**Why:** The setting `polaris_style_paths` was released in 26.4.x stable versions and is being renamed to `force_add_bucket` without adding an alias. Users with databases that have explicit `polaris_style_paths` settings will get UNKNOWN_SETTING errors on ATTACH. The default also changed from `true` to `false`, changing behavior for existing users relying on defaults.

Suggested category: `Backward Incompatible Change`

Could you check if the changelog category should be updated? This helps users know what to expect when upgrading. Ignore this if the change is intentionally non-breaking.

alexbakharew:
Hi @scanhex12,
few findings regarding this PR: 
- Docs don't have any info about new setting: https://clickhouse.com/docs/engines/database-engines/datalakecatalog#creating-a-database
- Unit-tests don't cover this setting well:
```
TEST_F(AzureAbfssParsingTest, TableMetadataSetLocationPolarisStyle)
{
    const std::string location = "abfss://mycontainer@mystorageaccount.dfs.core.windows.net/mycontainer/actual/path";

    TableMetadata metadata;
    metadata.withLocation();//.withForceAddBucket();
    metadata.setLocation(location);

    /// `getLocation` without endpoint is always a round-trip regardless of the Polaris flag.
    EXPECT_EQ(metadata.getLocation(), location);
}

TEST_F(AzureAbfssParsingTest, TableMetadataGetMetadataLocationEqualStrings)
{
    TableMetadata metadata;
    metadata.withLocation();
    metadata.setLocation("abfss://c@account.dfs.core.windows.net/c/table");

    const std::string metadata_file = "abfss://c@account.dfs.core.windows.net/c/table";
    EXPECT_EQ(metadata.getMetadataLocation(metadata_file), "");
}
```

These 2 tests don't fail after commenting `withForceAddBucket()` call.

cc: @clickgapai @groeneai

clickgapai:
@alexbakharew Both findings check out — verified against the merged code.

**Test gap (confirmed).** `force_add_bucket` is read in exactly two spots, both in `constructLocation` (`ICatalog.cpp:170` and `:201`). `constructLocation` is only reachable when an endpoint is set — `getLocation()` only calls it under `!endpoint.empty()` (`ICatalog.cpp:142-143`), and `getLocationWithEndpoint()` always does. `getMetadataLocation` never touches the flag; it just calls `getLocation()`.

Both surviving tests that call `withForceAddBucket()` set **no endpoint**:
- `TableMetadataSetLocationPolarisStyle` (line 200) → `getLocation()` with no endpoint returns `location_without_path / path`, a pure round-trip. Flag never read.
- `TableMetadataGetMetadataLocationPolarisStyle` (line 189) → same, via `getMetadataLocation` → `getLocation()`.

So `withForceAddBucket()` is a no-op in every remaining test — that's why commenting it out changes nothing. The PR deleted precisely the cases that exercised the flag (`...PolarisStyleWithEndpoint`, `...NonPolarisContainerInPathWithEndpoint`, `...S3WithHttpEndpoint`). The fix is to restore at least one S3-endpoint and one ABFSS-endpoint test that set `setEndpoint(...)` and assert the constructed URL with the flag both on and off. This also re-covers the two endpoint regressions flagged earlier in this thread.

**Docs gap (confirmed).** The DataLakeCatalog page lists no setting, and the in-code description (`DatabaseDataLakeSettings.cpp:48`) is just `"Add bucket name to the metadata path"` — which is also imprecise: the flag controls the **location/URL** path in `constructLocation`, not a "metadata path".

One thing worth restating for whoever lands the test: the flag's semantics flipped. The old Polaris path assumed the container was *already* in the path and stripped the duplicate; the new `force_add_bucket=true` unconditionally prepends `bucket` (`:172`/`:203`). A test feeding the old Polaris input (`.../mycontainer/actual/path`) with the flag on will now produce a doubled `mycontainer/mycontainer/...` — so the test input needs to match the new model, not be copied from the deleted tests.

The host-substring bug in the same branch (`find("/" + bucket)`) is already tracked at #104249.

groeneai:
@alexbakharew Confirmed against `upstream/master` ICatalog.cpp: `force_add_bucket` is read only at lines 170 and 201, both inside `constructLocation`, which `getLocation` only invokes when an endpoint is set (line 143: `if (!endpoint.empty())`). `getMetadataLocation` calls `getLocation` and never reads the flag itself.

Tightening clickgapai's framing: in the merged suite **zero** tests actually exercise `force_add_bucket`.

- `TableMetadataSetLocationPolarisStyle` (200) calls `withForceAddBucket` but only asserts `metadata.getLocation()` with no endpoint set. That path returns `location_without_path / path` and never reaches `constructLocation`.
- `TableMetadataGetMetadataLocationEqualStrings` (212) does not even call `withForceAddBucket`; it tests the prefix-equality short-circuit in `getMetadataLocation` (returns `""` when metadata file equals data location), unrelated to the flag.

So both surviving tests pass identically with the flag flipped or removed. The two endpoint-bearing tests that did exercise the flag (`...PolarisStyleWithEndpoint`, `...NonPolarisContainerInPathWithEndpoint`, `...S3WithHttpEndpoint`) were dropped by the rewrite.

Settings doc gap is also real: `clickhouse.com/docs/engines/database-engines/datalakecatalog` has no row for `force_add_bucket`. The in-source description in `DatabaseDataLakeSettings.cpp:48` (`"Add bucket name to the metadata path"`) understates the change: the flag controls URL construction in `constructLocation`, not a metadata path, and the default behavior flipped vs. the old `polaris_style_paths` (the bot inline correctly flagged this as a backward-incompatible setting break with no compat alias).

Happy to put up a follow-up PR with (a) endpoint-bearing tests asserting the constructed URL with the flag both on and off, for ABFSS and S3, and (b) the docs row plus a tightened in-source description, if @scanhex12 prefers I take it on. Otherwise will hold off and let the owner handle.

cron:clickhouse-ci-task-worker:20260604-225600

alexbakharew:
@groeneai 
Please create a follow-up PR: Add good test coverage + update docs

groeneai:
@alexbakharew follow-up filed: https://github.com/ClickHouse/ClickHouse/pull/106655 — four targeted unit tests for the two `force_add_bucket` call sites in `constructLocation`, plus a docs row and tightened in-source description.

PR #104121, opened 2026-05-05
Docs: fix extractGroups documentation

Blargian:

Corrects the documentation for `extractGroups`.

## Summary
The embedded `FunctionDocumentation` for `extractGroups` claimed the function returned `Array(Array(String))` and the example showed groups from every match in the input. In practice the function calls `re2->Match` once per row and returns `Array(String)` containing the capturing groups of only the **first** match — confirmed by `tests/queries/0_stateless/01273_extractGroups.reference`.

This PR updates:
- The description — clarifies that only groups from the first match are extracted, and points readers to `extractAllGroupsHorizontal` / `extractAllGroupsVertical` for all-match behavior.
- The returned-value type — `Array(String)` instead of `Array(Array(String))`.
- The example output — now matches what the engine actually returns.

Closes [DOC-712](https://linear.app/clickhouse/issue/DOC-712) / ClickHouse/clickhouse-docs#6069

## Test plan
- [ ] Existing tests in `tests/queries/0_stateless/01273_extractGroups.*` continue to pass (no behavioral change).

### Version info
- Merged into: `26.5.1.324`

### DIFF
--- a/src/Functions/extractGroups.cpp
+++ b/src/Functions/extractGroups.cpp
@@ -112,14 +112,14 @@ public:
 REGISTER_FUNCTION(ExtractGroups)
 {
     FunctionDocumentation::Description description = R"(
-Extracts all groups from non-overlapping substrings matched by a regular expression.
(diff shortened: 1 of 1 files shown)


Algunenano (approved):

theory:
Thank you, @Blargian!

Issue #92494, opened 2025-12-18
High memory use when listing DataLake/Glue tables (getAllTableNames/SHOW TABLES/system.tables)


filimonov:
### Describe the unexpected behaviour

- `SHOW TABLES` / `system.tables` (or hints when trying to access non-exiting table) call `database->getLightweightTablesIterator()`.
- For DataLake/Glue this calls `DatabaseDataLake::getTablesIterator()` (or the lightweight variant), which:
    - `catalog->getTables()` enumerates all Glue tables.
    - Schedules one task per table in `iceberg_catalog_threadpool` (50 threads, queue up to 1,000,000) invoking `tryGetTableImpl()`.
- `tryGetTableImpl()` always fetches metadata:
    - `GlueCatalog::tryGetTableMetadata()` reads the Iceberg metadata file from S3 and parses it with `Poco::JSON::Parser`, building `TableMetadata`, schema, and a `StorageObjectStorageCluster`.
- All tasks are awaited; promises/futures and resulting `StoragePtr` objects are kept until the iterator finishes. Tasks that throw still hold their allocations until futures are destroyed.
- “Lightweight” iterator still loads metadata/schema; there is no names-only path for DataLake/Glue.

##  Per-table memory footprint

- S3 read buffer ≈ `max_read_buffer_size_remote_fs` (default 1 MB, falls back to `max_read_buffer_size`), plus `prefetch_buffer_size` (default 1 MB) if `remote_filesystem_read_prefetch=true` (default).
- Full Iceberg metadata JSON (typically 100 KB–several MB) and parsed JSON DOM (~1–2×).
- Schema/types and the Storage object, plus log strings.

##  Impact

  - Listing many Glue/Iceberg tables retains per-table metadata and Storage objects until all tasks complete; memory is not freed incrementally.
  - In production this reached ~53 GB RSS during a catalog-wide walk, leading to OOM.

## Possible directions

- Provide a names-only listing path for DataLake/Glue (no Iceberg metadata/schema) for SHOW TABLES/system.tables/hints.
- For listing use strings instead of accumulating StoragePtrs; release per-table buffers/JSON/Storage as each future completes.
- Batch/limit the number of simultaneous/queued table builds to bound live memory.
- Allow smaller buffers for catalog metadata reads (lower max_read_buffer_size_remote_fs/prefetch_buffer_size, or disable prefetch for these reads).
- Early-exit on auth/S3 errors to avoid walking all tables in case of errors.

[...]

PR #104124, opened 2026-05-05
Override getAllTableNames() in DatabaseDataLake to avoid heavyweight …

### LINKED ISSUE #104123: DataLake catalog: getAllTableNames() triggers full per-table S3 metadata fetch during typo-hint lookup on unknown table
### Company or project name

_No response_

`DatabaseDataLake` does not override `getAllTableNames()`. The base class `IDatabase::getAllTableNames()` calls `getTablesIterator()`, which in `DatabaseDataLake` launches N parallel tasks via the Iceberg catalog threadpool — each fetching `metadata.json` from S3 and building a full `StoragePtr`. This fires whenever a query references a non-existent table in a DataLake database, because ClickHouse's typo-hint system calls `getAllTableNames()` to build its candidate list.

With tens of thousands of Glue/Iceberg tables this can exhaust RSS (identical failure mode to #92494). `getLightweightTablesIterator()` was correctly overridden by #94467/#97062 to avoid this on the `SHOW TABLES` and `SELECT name FROM system.tables` paths. That fix did not touch `getAllTableNames()`.

This issue is related to but distinct from #92494: that issue described the `SHOW TABLES` path, which has been addressed. The typo-hint path through `getAllTableNames()` was not touched by those PRs and remains broken.

**Component:** DataLake catalog / Iceberg — Glue, REST, and any other catalog implementation. The heavyweight call site is in `DatabaseDataLake`, not in catalog-specific code, so the bug is catalog-agnostic.

### Root cause — override matrix

| Method | Overridden in `DatabaseDataLake`? | Heavyweight? |
|---|---|---|
| `getTablesIterator()` | ✓ | Yes (intentional: full metadata needed) |
| `getLightweightTablesIterator()` | ✓ (added by #94467) | No — names only |
| `getAllTableNames()` | **✗** | **Yes — falls to base class → `getTablesIterator()`** |

### Call chain

```
SELECT * FROM my_glue_db.`ns.no_such_table`

  DatabaseCatalog::getTableImpl                       [DatabaseCatalog.cpp:490]
    database->tryGetTable("ns.no_such_table") → nullptr

  // exception arg set → typo-hint lookup fires:
  TableNameHints hints(database, ctx)                 [DatabaseCatalog.cpp:495]
  hints.getHints("no_such_table")                     [DatabaseCatalog.cpp:496]
    IHints<>::getHints(name)                          [NamePrompter.h:79-82]
      getAllRegisteredNames()
        TableNameHints::getAllRegisteredNames()       [DatabaseCatalog.cpp:2240]
          database->getAllTableNames(ctx)
            // DatabaseDataLake has no override → base class:
            IDatabase::getAllTableNames(ctx)          [IDatabase.h:297-303]
              getTablesIterator(ctx)                  [IDatabase.h:301]
                DatabaseDataLake::getTablesIterator() [DatabaseDataLake.cpp:724]
                  for each of N tables in catalog:
                    LOG "Get table information for table X"  [line 763]
                    pool.scheduleOrThrow(tryGetTableImpl)    [line 757]
                      catalog->tryGetTableMetadata(...)      [line 519]
                        → S3 read of metadata.json per table
```

Note the asymmetry: `getExtendedHintForTable` (cross-database hint search) already opts out of DataLake catalogs via `with_datalake_catalogs = false` [`DatabaseCatalog.cpp:2218`]. The within-database hint lookup has no equivalent guard.

Verified affected:
- `master` at commit `ed70b0de9686efe885020110a24025069ae1640e`
- `26.4` — #94467 was backported so SHOW TABLES is fixed, but `getAllTableNames()` is still not overridden

Other release branches likely affected wherever `DatabaseDataLake` exists without an `getAllTableNames()` override; not individually verified.

### How to reproduce

**Server version:** master at `ed70b0de9686efe885020110a24025069ae1640e`, also reproducible on `26.4`.

**Interface:** clickhouse-client (any interface that surfaces typo-hint suggestions on `UNKNOWN_TABLE` errors).

**Non-default settings:** none required. The example below uses `show_data_lake_catalogs_in_system_tables = 1` only for diagnostic visibility; the bug fires regardless of this setting.

**Setup — DataLake catalog with N tables:**

The bug is catalog-agnostic (Glue, REST, Unity all hit the same code path). Glue is shown below; substitute REST or Unity credentials as appropriate.

```sql
CREATE DATABASE my_glue ENGINE = DataLakeCatalog('http://glue:3000', 'key', 'secret')
SETTINGS catalog_type = 'glue',
         warehouse = 'test',
         storage_endpoint = 'http://minio:9000/warehouse-glue',
         region = 'us-east-1';
```

Populate `my_namespace` with N tables (e.g. via `pyiceberg` or the Glue API). The bug's severity scales linearly with N; it is observable at N=20 and catastrophic at N=10000+.

**Query that triggers the bug:**

```sql
SELECT * FROM my_glue.`my_namespace.no_such_table`;
```

This is expected to fail with `UNKNOWN_TABLE`. The unexpected behaviour is what happens *during* the error path: ClickHouse fetches `metadata.json` from S3 for every table in the namespace in order to build the typo-hint candidate list.

**Diagnostic:** count occurrences of `"Get table information for table"` in the server log before and after the failed query. The delta equals N. After the proposed fix the delta is 0.

A regression test is proposed for `tests/integration/test_database_glue/`; CI results will be attached once the fix PR is open.

### Expected behavior

Typo-hint lookup on a non-existent table in a DataLake database should use the catalog's lightweight name list (`getCatalog()->getTables()`) — the same data source `getLightweightTablesIterator()` already uses — without per-table `metadata.json` fetches.

Concretely: a `SELECT` against a non-existent table in a DataLake database with N tables should issue zero S3 metadata reads beyond the single failed `tryGetTable` lookup, regardless of N.

The user-visible error is the expected `UNKNOWN_TABLE` exception. The bug is not in the error message itself — it is in the work performed during the error path before the message is returned. There is no stacktrace because nothing crashes; the symptom is memory exhaustion and latency proportional to catalog size.

### Proposed fix

Override `getAllTableNames()` in `DatabaseDataLake`. Base class signature:

```cpp
// IDatabase.h:297
virtual Strings getAllTableNames(ContextPtr context) const
```

Override, mirroring the existing `getLightweightTablesIterator()` override:

```cpp
Strings DatabaseDataLake::getAllTableNames(ContextPtr /*context*/) const
{
    Strings result;
    try
    {
        result = getCatalog()->getTables();
    }
    catch (...)
    {
        tryLogCurrentException(__PRETTY_FUNCTION__);
    }
    return result;
}
```

Same data source, same error-suppression pattern as `getLightweightTablesIterator()`, no per-table S3 reads.

### What this does NOT fix

Secondary improvements from #92494 remain deferred:
- bounded threadpool concurrency on the heavyweight path
- smaller read buffers / disabled prefetch for catalog metadata reads
- per-table resource release as futures complete (rather than holding all StoragePtrs until the iterator finishes)
- early-exit on auth/S3 errors instead of walking the full catalog

These are separate concerns from the names-only listing path and should be tracked separately.

### Version info
- Resolved by: #104124
- Merged into: `26.6.1.1102`

il9ue:

### Changelog entry:
Fix excessive catalog/S3 metadata reads when an INSERT or DDL statement references a non-existent table in a DataLake catalog database with `show_data_lake_catalogs_in_system_tables` enabled. The typo-hint suggestion path loaded full per-table Iceberg metadata for the whole catalog, which could exhaust memory on large catalogs.

- No documentation needed (internal performance fix, no behavior change).

### Detailed description
Closes #104123. Related to #92494.

When a statement references a table that doesn't exist in a DataLake database, ClickHouse builds a "did you mean" hint via `IDatabase::getTable` → `TableNameHints` → `getAllTableNames()`. `DatabaseDataLake` didn't override `getAllTableNames()`, so the base implementation went through `getTablesIterator()` — one `metadata.json` read and a full `StoragePtr` per table in the catalog. #94467/#97062 fixed the same issue for `SHOW TABLES`/`system.tables` via `getLightweightTablesIterator()`; this does the same for `getAllTableNames()`, reading names from `getCatalog()->getTables()`.

Scope (narrowed after instrumenting the path — see #104123):
- The heavy path is reached only via the throwing `DatabaseCatalog::getTable` → typo-hint route: INSERT/DDL on a missing table (any analyzer), or SELECT with `enable_analyzer=0`. A default-analyzer SELECT resolves missing tables via `tryGetTable`, builds no hint, and is unaffected.
- It also requires `show_data_lake_catalogs_in_system_tables=1`; otherwise the remote-database guard in `getAllRegisteredNames` short-circuits and `getAllTableNames` is never called.

### Test
`tests/integration/test_database_glue/test.py::test_typo_hint_does_not_trigger_heavyweight_fetch`: creates a catalog with N=20 tables, triggers the hint via INSERT into a missing table with the setting enabled, and asserts zero per-table heavyweight fetches. Verified to fail before this change and pass after.

### Version info
- Merged into: `26.6.1.1102`


ilejn:
Hello @il9ue, I think we should sort test errors out.

1. Stateless tests are unlikely related, obfuscator is not touched
2. bugfix validation failure is definitely caused by this PR. The idea of this kind of test is they must be failed before suggested change is applied.
3. not sure about unit tests, probably not related.

What do you think?

arsenmuk:
> Hello @il9ue, I think we should sort test errors out.

Thank you @ilejn! I agree with you

Hi @il9ue! I took responsibility to merge the latest master into your branch, this should help with some failures. Let's see the results on CI, and get the remaining failures sorted.

il9ue:
Thank you both and apologize for late feedback.
Agree the stateless and paimon failures look unrelated and the master merge should clear them. 
The bugfix validation reds in test_database_glue are on me it seems; the new test runs the heavyweight path over the whole shared catalog on the un-fixed binary and destabilizes the module. Preparing the hardened/isolated tests which I've isolated it / added teardown so it only touches its own tables.

il9ue:
I went through the failing jobs and split them by whether this change can actually reach the code path involved. Details below.

| Job | Failing test(s) | Related to this PR? | Why |
| --- | --- | --- | --- |
| Fast test (arm_darwin) | `01951_distributed_push_down_limit`, `01952_optimize_distributed_group_by_sharding_key` | No | The reference diffs are entirely in distributed `remote()` query planning (`ReadFromRemote` being expanded into a local plan). This PR only adds `DatabaseDataLake::getAllTableNames()` and never touches planning. |
| Integration (amd_asan_ubsan, db disk, old analyzer, 3/6) | `test_paimon_incremental_read::*` (×2) | No | Both are `retry_ok`, and the teardown crash is a `Logical error: 'file_offset_of_buffer_end <= getFileSize()'` in `AsynchronousBoundedReadBuffer::nextImpl()` under `PaimonMetadata::runBackgroundRefresh()` — a Paimon background-refresh issue in a different subsystem. Note `test_database_glue` passes in this same run. |
| Bugfix validation (integration) | `test_database_glue/test.py` (whole file listed) | **Yes** | *"Failed to reproduce the bug."* This is the one I own. |

- **Stateless `01951` / `01952`.** These are purely distributed-`remote()` planning reference drift — the expected plan shows `ReadFromRemote`, the produced plan inlines the remote side into a local `Sorting`/`Distinct`/`LimitBy` subtree. That path is entirely independent of DataLake catalogs, so the change here can't be the cause; this looks like master-side planner drift (possibly references picked up by the master merge). I'd suggest re-running rather than treating it as a blocker.

- **Paimon.** Both cases recovered on retry (`retry_ok`), and the fatal in the teardown log is in `AsynchronousBoundedReadBuffer` during Paimon's background snapshot refresh — nothing in this diff reaches read buffers or Paimon metadata. Also seems unrelated.

- **Bugfix validation — the real one.** The *"Failed to reproduce the bug"* verdict is the important signal: it means the new test **passes against the binary built without the fix**, which is the opposite of what a bugfix test should do (it must fail before the change is applied). So the reproduction isn't valid yet, and that's on me. I'm narrowing it down to one of two causes:

1. the log line I count (`"Get table information for table …"`) doesn't match my namespace-scoped marker — so `count_in_log` returns 0 in *both* builds and the assertion passes vacuously; or
2. the missing-table query isn't actually routed through `getAllTableNames()` → `getTablesIterator()` for this query shape, in which case the override isn't on the path the bug takes.

Once I confirm which it is, I'll make the reproduction deterministic. Will push an update and re-run.

tests/integration/test_database_glue/test.py
```diff
@@ -902,6 +902,121 @@ def test_show_tables_optimization(started_cluster):
...
+    # Log line emitted exactly once per table inside DatabaseDataLake::getTablesIterator
+    # [src/Databases/DataLake/DatabaseDataLake.cpp:763].
+    marker = f"Get table information for table {namespace}."
+    baseline = int(node.count_in_log(marker).strip())
+
+    # Reference a non-existent table. This must raise UNKNOWN_TABLE.
+    # If it does not raise (e.g. setup is broken and the table accidentally exists),
+    # pytest.fail() ensures the test fails loudly rather than silently passing.
+    try:
+        node.query(
```
zvonand: `node.query_and_get_error()` serves specifically this use case

il9ue: Done, switched to `node.query_and_get_error()` and removed the `try/except` + `pytest.fail` guard. Thanks!

arsenmuk (approved):
Thank you @il9ue! All looks good now

PR #104127, opened 2026-05-05
Remove stale .repodata temp directory before running createrepo_c

leshikus:
\`createrepo_c\` uses a \`.repodata/\` temp directory during repo generation and refuses to start if it already exists, printing "Another createrepo process is running?". When a previous release run is interrupted, this directory is left behind, causing all subsequent RPM export jobs to fail with exit code 1 — even though no other process is actually running.

Fix: \`RpmArtifactory.export_packages\` now removes \`.repodata/\` from the destination directory before invoking \`createrepo_c\`, if it exists.

CI failure: https://github.com/ClickHouse/ClickHouse/actions/runs/25386526157/job/74449100394

### Version info
- Merged into: `26.5.1.333`

### DIFF
--- a/tests/ci/artifactory.py
+++ b/tests/ci/artifactory.py
@@ -230,6 +230,13 @@ class RpmArtifactory:
         for package in paths:
             _copy_if_not_exists(Path(package), dest_dir)
 
+        # Remove stale temp repodata directory left by a previous interrupted run; createrepo_c
+        # refuses to start when it already exists, mistaking it for a concurrent process.
+        stale_repodata = dest_dir / ".repodata"
+        if stale_repodata.exists():
+            print(f"Removing stale temp repodata directory: {stale_repodata}")
+            Shell.check(f"rm -rf {stale_repodata}", strict=True)
+
         # switching between different fuse providers invalidates --update option (apparently some fuse(s) can mess around with mtime)
         #   add --skip-stat to skip mtime check
         commands = (
--- a/tests/ci/artifactory.py
+++ b/tests/ci/artifactory.py
@@ -1,7 +1,7 @@
 import argparse
 import time
 from pathlib import Path
-from shutil import copy2
+from shutil import copy2, rmtree
 from typing import Optional
 
 from ci_utils import Shell, WithIter
@@ -235,7 +235,7 @@ class RpmArtifactory:
         stale_repodata = dest_dir / ".repodata"
         if stale_repodata.exists():
             print(f"Removing stale temp repodata directory: {stale_repodata}")
-            Shell.check(f"rm -rf {stale_repodata}", strict=True)
+            rmtree(stale_repodata)
 
         # switching between different fuse providers invalidates --update option (apparently some fuse(s) can mess around with mtime)
         #   add --skip-stat to skip mtime check


leshikus:
Here is a test run https://github.com/ClickHouse/ClickHouse/actions/runs/25387680888

alexey-milovidov:
@groeneai or @leshikus, explain what is `createrepo_c`. The name obviously violates the ClickHouse code style, because we won't allow `_c` - this abbreviation does not explain what `c` means. But if it is some rusty, dusty third-party code written not by us, we can assign blame to that.

tests/ci/artifactory.py
```diff
@@ -230,6 +230,13 @@ def export_packages(self, codename: Optional[str] = None) -> None:
         for package in paths:
             _copy_if_not_exists(Path(package), dest_dir)
 
+        # Remove stale temp repodata directory left by a previous interrupted run; createrepo_c
+        # refuses to start when it already exists, mistaking it for a concurrent process.
+        stale_repodata = dest_dir / ".repodata"
+        if stale_repodata.exists():
+            print(f"Removing stale temp repodata directory: {stale_repodata}")
+            Shell.check(f"rm -rf {stale_repodata}", strict=True)
```
alexey-milovidov: This is dangerous - need to properly quote the directory name inside the shell command.

alexey-milovidov (changes requested):
^

alexey-milovidov:
> I did not find correctness, safety, or compatibility issues in the proposed change.

@groeneai, `rm -rf` with unescaped substitution is a screaming code defect. Please update our code review Claude skill so that it will be noticed automatically next time; send a PR with this change.

maxknv:
> > I did not find correctness, safety, or compatibility issues in the proposed change.
> 
> @groeneai, `rm -rf` with unescaped substitution is a screaming code defect. Please update our code review Claude skill so that it will be noticed automatically next time; send a PR with this change.

this bot does not do reviews. 
@pufit @nikitamikhaylov wdyt if groenai or similar agent does reviews instead of the current approach, so that Alexey can teach it

alexey-milovidov:
I've asked @groeneai to fix our `.claude` files, so that the other bot will do reviews better.

alexey-milovidov:
@clickgapai, could you please implement what I've asked in this comment? https://github.com/ClickHouse/ClickHouse/pull/104127#issuecomment-4381931553

clickgapai:
_This automated ClickGap comment is withdrawn — it surfaced internal tooling detail or was posted incompletely, and isn't actionable as written. Please disregard. (ClickGap's output guard now prevents this.)_

groeneai:
@alexey-milovidov — done. Code review skill updated to flag this exact pattern automatically: #104155.

**What the skill now catches:**

- `rm -rf` and other destructive/privileged commands (`mv`, `cp -r`, `find … -delete`, `chmod`, `chown`, `dd`, `kill`, `sudo …`) constructed via f-string / `%`-format / `+`-concat substitution and passed to `subprocess.run(..., shell=True)`, `subprocess.Popen(..., shell=True)`, `os.system`, `os.popen`, **or** any in-tree wrapper that uses `shell=True` under the hood — explicitly naming `Shell.check` / `Shell.run` / `Shell.get_output` in `ci/praktika/utils.py` and `tests/ci/ci_utils.py` so the skill is not fooled by the helper appearing benign.
- Backticks / `$(…)` and unquoted `$VAR` in destructive shell-script commands (`rm -rf $dir/.tmp`, `rm -rf "$DIR"/*` where `$DIR` may be empty, `find $DIR -name '*.tmp' -delete`).
- Path concatenation with `+` or `f"{a}/{b}"` for filesystem operations when a `pathlib.Path` is already in scope.

The skill now classifies these as **Blockers** in the Severity Model (also added a cross-reference bullet there) and proposes the concrete replacement — `shutil.rmtree(Path(...) / ...)`, an argv list with `--` terminator, or `shlex.quote(str(...))` when a shell wrapper is genuinely required.

---

**On `createrepo_c`** (your separate question):

`createrepo_c` is the C reimplementation of the original Perl `createrepo` tool — it generates yum/dnf/RPM repository metadata (the `repodata/` directory with `repomd.xml`, `primary.xml.gz`, `filelists.xml.gz`, etc.). The `_c` suffix is the upstream's naming choice (https://github.com/rpm-software-management/createrepo_c) — `c` here means "C language reimplementation", to distinguish it from the legacy Perl version which is still called `createrepo`. So this is third-party tooling we shell out to, not ClickHouse code; the name is the upstream's to own. The `.repodata/` temp directory whose stale state #104127 cleans up is `createrepo_c`'s own scratch area — same name, just dot-prefixed.

maxknv (approved):

alexey-milovidov (approved):

PR #104134, opened 2026-05-05
Set charset=utf-8 on text uploads to S3 for CI logs

alexey-milovidov:
CI logs (and other text artifacts) are valid UTF-8, but `ci/praktika/s3.py` uploaded them with `Content-Type: text/plain` (no charset). Browsers fall back to Latin-1 in that case, so an em-dash `—` from system-table `COMMENT` text rendered as `â€—`.

Example log with the broken rendering: https://s3.amazonaws.com/clickhouse-test-reports/PRs/104097/a1f746ae407f13695591735e3a509b1136c036a2/stateless_tests_amd_tsan_flaky_check/job.log

The bytes themselves are valid UTF-8 (`e2 80 94` = U+2014 em-dash), so this is a server-header issue, not a content issue. The em-dashes originate from human-readable `COMMENT` text in source files like `src/Interpreters/AsynchronousInsertLog.cpp`, `OpenTelemetrySpanLog.cpp`, `PartLog.cpp`.

Set the charset explicitly on every text upload path in `ci/praktika/s3.py` (boto3 and AWS CLI, plain and versioned).

...

### Version info
- Merged into: `26.5.1.312`

### DIFF
--- a/ci/praktika/s3.py
+++ b/ci/praktika/s3.py
@@ -125,7 +125,7 @@ class S3:
                     inferred_content_type, _ = mimetypes.guess_type(key)
 
                 if text and not content_type:
-                    extra_args["ContentType"] = "text/plain"
+                    extra_args["ContentType"] = "text/plain; charset=utf-8"
                 elif content_type:
                     extra_args["ContentType"] = content_type
                 elif inferred_content_type:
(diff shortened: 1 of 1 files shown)


alexey-milovidov (commented):
LGTM

Issue #100759, opened 2026-03-25
Unstable performance tests: 61 tests show >10% noise in >10% of master runs


Algunenano:
### Overview

**Updated 2026-05-28** with master data from after #105623 merged on 2026-05-27 14:53 UTC. The list below replaces the original snapshot (which is now obsolete because the framework-level noise was reduced by #105527 / #105623).

The following performance tests show ≥10% noise rate (slower, faster, or unstable) in master runs since the merge. They generate noise in PR reports that must be manually dismissed.

Data source: `default.checks` on `play.clickhouse.com`, `pull_request_number = 0`, `check_start_time > '2026-05-27 14:53:03'`. Sample sizes per test/arch are small (~20-50 runs), so individual percentages have limited precision.

#### Multi-variant clusters

- [x] **`right_full_hash_join` family** — 14 variants flagged (`#0`, `#1`, `#2`, `#3`, `#4`, `#5`, `#8`, `#9`, `#10`, `#11`, `#12`, `#13`, `#15`, `#17`). Mostly AMD-dominant: `#3` (amd 100% / arm 20%), `#5` (amd 83% / arm 0%), `#4` (amd 80% / arm 0%); some both-arch: `#9` (amd 56% / arm 22%), `#11` (amd 57% / arm 23%); some ARM-only: `#0` / `#1` (arm 23-33%, amd 0%). All entries are dominated by `unstable`, not directional `slower`. https://github.com/ClickHouse/ClickHouse/pull/104268
- [ ] **`parallel_final`** — `#9` (arm 89% / amd 38%) and `#13` (arm 60% / amd 20%). Both archs, `unstable`.

#### Individual tests

- [x] `reading_from_file #0` — both archs (arm 67% / amd 59%). https://github.com/ClickHouse/ClickHouse/pull/106249
- [ ] `top_k_through_join #2` — both archs (arm 43% / amd 42%)
- [ ] `group_by_fixed_keys #10` — arm only (38%)
- [ ] `prefetch_in_aggregation #9` — arm only (36%)
- [ ] `hash_table_sizes_stats #3` — arm only (28%)
- [ ] `uniq_without_key #24` — both archs (amd 29% / arm 10%)
- [ ] `queries_over_aggregation #3` — arm only (19%)
- [ ] `encrypt_decrypt_empty_string #11` — arm only (18%)
[...]

PR #104136, opened 2026-05-05
Route long-lived MergeTree heap state to a dedicated jemalloc arena

Algunenano:
Routes per-part and per-table MergeTree heap state into a dedicated jemalloc
arena (`manual_<idx>`, exposed as `jemalloc.mergetree_arena.*` async metrics).
This separates allocations that live as long as a part or table from
short-lived query allocations, reducing default-arena fragmentation and
keeping per-part survivors from pinning otherwise-decayable pages.

Wrapped sites:

- Part metadata: `IMergeTreeDataPart::setColumns` / `setColumnsSubstreams`
  / `loadColumnsChecksumsIndexes` / `loadProjections` / `loadChecksums`,
  `MergeTreeDataPartBuilder::build`.
- Table metadata: `MergeTreeData::setProperties` (covers `setInMemoryMetadata`
  clones of `StorageInMemoryMetadata` / `ColumnsDescription` / virtual
  columns / key + index ASTs), `MergeTreeData::changeSettings`,
  `resetSerializationHints` / `updateSerializationHints`.
- Part construction paths: `loadDataPart`, `loadUnexpectedDataPart`,
  `loadPartRestoredFromBackup`, `tryLoadPartsToAttach`, `MergeTask`,
  `MutateTask`, `DataPartsExchange` (tight scope around volume + storage
  only — fetch I/O buffers stay in the default arena), and the attach
  path in `StorageReplicatedMergeTree`.

Adds two stateless tests:
- `04203_jemalloc_mergetree_arena_metrics.sql` — async metrics are
  registered on jemalloc builds.
- `04204_mergetree_arena_routes_writes.sql` — smoke test that exercises
  both code paths (CREATE + INSERT) and asserts the arena holds more than
  1 MiB afterwards. The test file documents why a delta-based assertion
  (e.g. comparing `before`/`after` snapshots of
  `jemalloc.mergetree_arena.active_bytes`) is unreliable: the metric is
  process-wide and dominated by background `system.metric_log` /
  `text_log` flushes and their merges/cleanup, even with `no-parallel`
  and `SYSTEM STOP MERGES`.

- [x] Review and include SharedMergeTree code

- Improvement

Route long-lived MergeTree heap state (per-part and per-table metadata) to a dedicated jemalloc arena, exposed via `jemalloc.mergetree_arena.*` async metrics. Reduces default-arena fragmentation in steady-state and keeps per-part survivors from pinning otherwise-decayable pages.

### Version info
- Merged into: `26.5.1.431`


tests/queries/0_stateless/04204_mergetree_arena_grows_on_insert.sh
```diff
@@ -0,0 +1,45 @@
...
+. "$CUR_DIR"/../shell_config.sh
+
+$CLICKHOUSE_CLIENT -q "DROP TABLE IF EXISTS t_mergetree_arena SYNC"
+$CLICKHOUSE_CLIENT -q "
+    CREATE TABLE t_mergetree_arena (a UInt64, b String, c LowCardinality(String), d Float64)
+    ENGINE = MergeTree ORDER BY a
+    SETTINGS old_parts_lifetime = 10"
+
+$CLICKHOUSE_CLIENT -q "SYSTEM RELOAD ASYNCHRONOUS METRICS"
+before=$($CLICKHOUSE_CLIENT -q "
```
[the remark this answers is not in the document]
Algunenano: Addressed in the upcoming push: 04204 now does SYSTEM JEMALLOC PURGE before sampling `before` (matching 04205) and the assertion is replaced with a positive-growth check (`after - before >= 4 MiB`), so it no longer depends on prior arena decay. Both tests also now run SYSTEM STOP MERGES globally to keep system-table cleanup from eating the signal, with a `trap` to always restore on exit.

src/Storages/MergeTree/IMergeTreeDataPart.cpp
```diff
@@ -1279,6 +1296,14 @@ void IMergeTreeDataPart::addProjectionPart(
 void IMergeTreeDataPart::loadProjections(
     bool require_columns_checksums, bool check_consistency, bool & has_broken_projection, bool if_not_loaded, bool only_metadata)
 {
+    /// Each loaded projection becomes its own `IMergeTreeDataPart` (via `getProjectionPartBuilder().build()`)
+    /// and is stored in the parent's `projection_parts` map. Both the map node insertion in
+    /// `addProjectionPart` and any allocation paths reached during build/load that aren't already
+    /// scoped by their own helpers belong in the parts arena. This is also the entry point used
+    /// directly by `MutateTask`, where the surrounding `loadColumnsChecksumsIndexes` scope is not
+    /// in effect; without this guard those calls would land in the default arena.
+    ScopedJemallocThreadArena mergetree_arena_scope(JemallocMergeTreeArena::getArenaIndex());
```
antaljanosbenjamin, line 1305: In general, shouldn't we disable memory tracking here? I am not sure, if this code can be called from a query directly or not, but it might make sense to do similarly to `loadColumnsChecksumsIndexes`.

I am not sure, so take this more like an idea than a suggestion.

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -4877,6 +4894,11 @@ void MergeTreeData::changeSettings(
             = (*storage_settings.get())[MergeTreeSetting::refresh_statistics_interval].totalSeconds() != (*copy)[MergeTreeSetting::refresh_statistics_interval].totalSeconds();
 
         storage_settings.set(std::move(copy));
+
+        /// Route the new `StorageInMemoryMetadata` clone (and the deeper clone produced by
+        /// `setInMemoryMetadata`) into the dedicated MergeTree arena.
+        ScopedJemallocThreadArena mergetree_arena_scope(JemallocMergeTreeArena::getArenaIndex());
+
```
antaljanosbenjamin, line 4901: Starting the background moves and the statistics cache are also allocating tiny amount of memory on the heap for the task holders. I think it is fine to have them on the same arena, but wanted to make sure it is also okay for you.

antaljanosbenjamin (approved):

src/Storages/MergeTree/IMergeTreeDataPart.cpp
Algunenano: Fair question, but I think it's highly debatable. `loadColumnsChecksumsIndexes` blocks it but I don't think it's necessarily good since that delays the problem and can lead to OOM kills. I would rather keep the behaviour as it is for now, and we can iterate later.

src/Storages/MergeTree/MergeTreeData.cpp
Algunenano: Yes, it is ok. The task holders live as long as the table, so being in the same arena is exactly what we want. Thanks for flagging

Algunenano:
CI:
* I've fixed the function fuzzer issues in other PRs.
* About the perf changes:  Most are noise present in master. Others changes, speed up or degradation,  cannot be related to the changes (direct_dictionary which doesn't use mergetree).

I'll just update with master to confirm it keeps green and do another loop of performance tests.

Algunenano:
## Observed memory improvements

After cross-referencing the metric changes table against ~400 recent PRs in CIDB, three signals are clearly attributable to this PR rather than test-framework baseline:

| Metric | Typical PR p50 (last 14d) | This PR | Real delta |
|---|---|---|---|
| `jemalloc.arenas.all.pdirty` | −10.8% | **−43.9%** | ~3–4× the typical drop |
| `jemalloc.arenas.all.dirty_purged` | +3.2% | **−18.6%** | non-flag → clear reduction |
| `MemoryResident` | −20.9% | **−31.9%** | ~11pp beyond baseline |

Other metrics in the table — the `MemoryResident` baseline of −22%, `TrackedMemory` +46%, `jemalloc.background_thread.num_runs` swings of ±400% — appear on every recent PR and are test-framework artifacts (the LEFT/master server accumulates more state before metric snapshots than the RIGHT/PR server). I'm excluding those from the attributable signal.

### Methodology

The "Metric Changes" table at the bottom of each performance comparison report is generated by `ci/jobs/scripts/perf/compare.sh:1108`: at the end of the perf run, both servers (left = master baseline, right = PR build) dump `system.asynchronous_metric_log`. A `clickhouse-local` query computes the median value of each metric over the run, joins LEFT/RIGHT, and reports any metric where `|(right − left) / left| > 0.05`. The result is also uploaded to the public CI database table `perf_metric_changes_v1` (one row per metric per shard per run, with `pr_number`, `arch`, `report_url`, `old_median`, `new_median`, `diff`, `times_diff`).

To distinguish "this PR moves the metric" from "the test framework moves this metric on every PR":

1. For each PR in the last 14 days with ≥5 shards, computed the per-PR average delta of each metric across its shards.
2. Built the p10/p50/p90 distribution of that per-PR average across ~400 PRs.
3. Compared this PR's per-metric average to that distribution.

A metric is attributable to this PR only if its delta sits **outside the typical PR distribution** (well past p90 or p10), not inside it. Sample side-by-side from neighboring PRs:

| PR | `MemoryResident` | `pdirty` | `dirty_purged` |
|---|---|---|---|
| 104128 | −22.2% | −14.6% | within noise |
| 104232 | −21.7% | −14.2% | within noise |
| 104237 | −22.3% | −10.2% | within noise |
| 104262 | −21.7% | −13.6% | within noise |
| 104274 | −21.8% | within noise | within noise |
| 104287 | −22.1% | within noise | within noise |
| **104136** | **−31.9%** | **−43.9%** | **−18.6%** |

Other PRs cluster tightly (`MemoryResident` −22%, `pdirty` −10 to −15%, `dirty_purged` not flagged at all because it's within the ±5% threshold). This PR is a clear outlier on all three.

### What each metric means and why it moved

**`jemalloc.arenas.all.pdirty` (−44%)** — Count of pages that jemalloc has freed but not yet returned to the OS, sitting in the dirty-decay queue. With long-lived MergeTree metadata isolated in its own arena, the default percpu arenas no longer have per-part survivors pinning their slabs. Pages there can fully evacuate when their query-scoped allocations age out, instead of staying pinned with one or two long-lived stragglers. Fewer pages end up stuck in the dirty state.

**`jemalloc.arenas.all.dirty_purged` (−19%)** — Cumulative count of pages purged from dirty back to the OS via `MADV_DONTNEED`. Lower means less churn: pages aren't cycling through allocate → free → dirty → purge as often. Consistent with the same mechanism — when the lifetime of allocations on a page is uniform, the whole page can either stay active or decay together, instead of repeatedly being partially freed and re-used.

**`MemoryResident` (−32% vs −21% baseline)** — Total RSS. The ~11pp extra reduction (roughly half a GiB on this workload) is the downstream effect of the two arena-level wins above: when pages can fully decay, they get returned to the OS, and the process holds less physical memory.

### Per-shard consistency

The signal is consistent across all 6 shards of this PR (latest commit):

| Shard | `MemoryResident` | `pdirty` | `dirty_purged` |
|---|---|---|---|
| 1/6 | −29.2% | −18.5% | −19.6% |
| 2/6 | −33.0% | −38.7% | −8.9% |
| 3/6 | −33.6% | −50.3% | −25.8% |
| 4/6 | −31.3% | −51.3% | −34.2% |
| 5/6 | −33.0% | −49.3% | −19.6% |
| 6/6 | −32.0% | −50.5% | within noise |

Not a per-runner outlier. Shard 1/6 has the weakest `pdirty` drop (−18.5%) but still outside the typical baseline range.

The query-level perf changes elsewhere in the report are dominated by per-shard runner variance: same-family queries move in opposite directions on the same shard, and several of the flagged queries (`apply_patch_parts_join`, `group_array_moving_sum`, `json_type_parsing`) fail on master_head independently of this PR.

Algunenano:
Performance changes are unrelated AFAICT

nickitat:
Raul, can you take a look pls. 

frac_unstable_4h

AFAIU, it used the `unstable` column from `default.perf_test_perf_changes_v1`.

```
  What I ruled out

  - Runner change — same m7i.4xlarge (AMD) and m8g.4xlarge (ARM) all along.
  - Perf-test infra change — docker/test/performance-comparison/ had no commits in the bisection window (only a dependabot pip bump 13 days earlier).
  - Other merges in the window between last clean (bca25b7b) and first noisy (c7d5efb0) baselines:

  | PR                                    | What it touches                                   | Could affect non-MergeTree perf?   |
  |---------------------------------------|---------------------------------------------------|------------------------------------|
  | #104302 auto-release-fail-fast-false  | .github/workflows/auto_releases.yml               | No (CI YAML)                       |
  | #103921 coverage-pr61248              | tests only                                        | No                                 |
  | #103891 coverage-pr60419              | tests only                                        | No                                 |
  | #103708 fix-text-index-alias-arraymap | RPNBuilder.cpp + text-index test                  | Only text-index queries            |
  | #104136 parts-metadata-arena          | Common/Jemalloc.*, new arena, MergeTreeData, etc. | Yes — touches the global allocator |

  - The next several merges that day (#103980 cmake, #104270 test, #104304 test fix, #104355 perf-test config, #104207 BuzzHouse) all post-date the first noisy baseline c7d5efb0, so by elimination they can't be the cause.

  What corroborates #104136

  - Its own CI report flagged jemalloc.arenas.all.pdirty −44%, dirty_purged −19%, MemoryResident −32% vs. neighbour PRs that cluster at ~−14% / not-flagged / −22%. The reviewer thread on the PR contains a long, methodically-built justification of those signals as the
   intended effect of routing long-lived allocations to a dedicated arena. Different dirty-page decay patterns plausibly increase run-to-run variance for everything that allocates from the default arenas.
  - Tests that became unstable (parallel_hash_build_phase, grace_hash_join, group_array_moving_sum, order_by_decimals, …) are precisely the workloads that hammer the default-arena allocators (hash tables, sort buffers, lazy state).
  - Algunenano (same author) filed #104355 "Reduce JIT-induced noise in performance tests" ~90 minutes after merging #104136. It only fixed a different pre-existing problem (issue #100759, opened 2026-03-25, mostly jit_aggregate_functions flakiness) — not this new
  step-change.
  - Step is sharp, 3× wider than the pre-jump spread, and persists across 30+ subsequent master baselines with no recovery.

  Honest caveats

  - I have not reverted #104136 to confirm causally; the evidence is correlational + mechanistic.
  - One pre-jump baseline (2d2e88c93d on May 5, frac=0.0161) sits above the pre-jump cluster — single outlier across 44 high-confidence baselines; probably one bad run.
  - The "everything else got noisier when MergeTree allocations moved out" mechanism is plausible but I haven't profiled it directly.

  So: not 100% proof, but among the merges in the window #104136 is the only realistic candidate, and the side-evidence (jemalloc-metric outliers in its own CI, author follow-up, character of affected tests) all points the same way.
```

Algunenano:
I'll have a look. We can continue in https://github.com/ClickHouse/ClickHouse/issues/100759 if you prefer

Issue #103820, opened 2026-05-01
Flaky test: test_distributed_frozen_replica/test.py::test_cluster_all_replicas_query[3-test_cluster]


Avogar:
_Important: This issue was automatically generated and is used by CI for matching failures. DO NOT modify the body content. DO NOT remove labels._

Test name: test_distributed_frozen_replica/test.py::test_cluster_all_replicas_query[3-test_cluster]
Failure reason: Timeout
CI report: [Integration tests (amd_msan, 1/6)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100849&sha=0ce860d03fa6fb6170133233031ee31dc2e488d7&name_0=PR&name_1=Integration%20tests%20%28amd_msan%2C%201%2F6%29)

[...]

Issue #103819, opened 2026-05-01
Flaky test: test_distributed_frozen_replica/test.py::test_cluster_all_replicas_query[1-test_cluster]


Avogar:
_Important: This issue was automatically generated and is used by CI for matching failures. DO NOT modify the body content. DO NOT remove labels._

Test name: test_distributed_frozen_replica/test.py::test_cluster_all_replicas_query[1-test_cluster]
Failure reason: AssertionError
CI report: [Integration tests (amd_msan, 1/6)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100849&sha=0ce860d03fa6fb6170133233031ee31dc2e488d7&name_0=PR&name_1=Integration%20tests%20%28amd_msan%2C%201%2F6%29)

[...]

PR #104167, opened 2026-05-06
Make test_cluster_all_replicas_query deterministic by gating on a pause-effective probe

groeneai:
Make `test_distributed_frozen_replica/test.py::test_cluster_all_replicas_query` deterministic.

### Motivation

`test_cluster_all_replicas_query` is a chronic flaky test (issues #103819 and #103820) that asserts pausing `node1` makes `clusterAllReplicas(test_cluster, default.local_table)` fail with `ALL_CONNECTION_TRIES_FAILED`.

`cluster.pause_container` uses Docker's cgroup freezer to suspend `node1`'s process tree. The freezer takes effect quickly but not synchronously with in-flight TCP traffic, so the test SELECT can race the freezer: `node1` completes the TCP handshake and the tiny ClickHouse server greeting before being suspended, every per-replica shard succeeds, and `node3.query(...)` returns a clean `count()` with `error == ''`. The assertion then fails spuriously with:

```
assert "ALL_CONNECTION_TRIES_FAILED" in error
E   AssertionError: assert 'ALL_CONNECTION_TRIES_FAILED' in ''
```

This failure mode emerged after the prior `857 ms` formatting fix in PR #103780. Cross-PR data over the last 30 days shows 17 distinct PRs hitting it (including 6 master runs).

### Approach

Inside the pause context, poll a fast probe — `SELECT 1 FROM remote('node1:9000', system.one)` with a 200 ms `handshake_timeout_ms` and `connect_timeout_with_failover_ms` — until it raises `QueryRuntimeException`. Once that probe fails, connections to `node1` are demonstrably blocked at the application level and the actual `clusterAllReplicas` probe can run deterministically. The 30-second deadline accommodates slow CI runners; in practice the probe should fail on the first or second iteration.

The original assertions (`ALL_CONNECTION_TRIES_FAILED in error`, `857 ms in error`) are unchanged — the fix only removes the timing race.

### Related

- Issue #103819 — Flaky test: `test_cluster_all_replicas_query[1-test_cluster]`
- Issue #103820 — Flaky test: `test_cluster_all_replicas_query[3-test_cluster]`
- PR #103780 — earlier fix for the `857 ms` format-parity assertion (merged 2026-04-30)
- PR #101799 (comment by @alexey-milovidov 2026-05-05T19:24:26Z): https://github.com/ClickHouse/ClickHouse/pull/101799#issuecomment-4384287834

...

### Version info
- Merged into: `26.5.1.351` (included in `26.5` and later)
- Backported to: `26.3.33.53`

### DIFF
--- a/tests/integration/test_distributed_frozen_replica/test.py
+++ b/tests/integration/test_distributed_frozen_replica/test.py
@@ -91,6 +91,36 @@ def test_cluster_all_replicas_query(
     node3.query("SYSTEM DROP DNS CACHE")
     error = ""
     with cluster.pause_container("node1"):
+        # `cluster.pause_container` uses Docker's cgroup freezer to suspend
+        # `node1`'s process tree. The freezer takes effect quickly but not
+        # synchronously with respect to in-flight TCP traffic, so the SELECT
+        # below can race it: `node1` can complete the TCP handshake and the
+        # tiny ClickHouse server greeting before being fully suspended, the
+        # per-replica shard then succeeds, the whole `clusterAllReplicas`
+        # query returns `count()` cleanly with `error == ''`, and the
+        # assertion below fails spuriously (chronic flake — see issues
+        # #103819 and #103820). Gate the actual probe on a deterministic
+        # check that connections to `node1` are failing.
+        deadline = time.time() + 30
+        pause_effective = False
+        while time.time() < deadline:
+            try:
+                node3.query(
+                    "SELECT 1 FROM remote('node1:9000', system.one) "
+                    "SETTINGS connect_timeout_with_failover_ms=200, "
+                    "handshake_timeout_ms=200, "
+                    "connections_with_failover_max_tries=1",
+                    timeout=2,
+                )
+            except QueryRuntimeException:
+                pause_effective = True
+                break
+            time.sleep(0.1)
+        assert pause_effective, (
+            "cluster.pause_container('node1') did not become effective within 30s — "
+            "connections to node1 still succeed"
+        )
+
         try:
             node3.query(
                 f"""
--- a/tests/integration/test_distributed_frozen_replica/test.py
+++ b/tests/integration/test_distributed_frozen_replica/test.py
@@ -91,36 +91,6 @@ def test_cluster_all_replicas_query(
     node3.query("SYSTEM DROP DNS CACHE")
     error = ""
     with cluster.pause_container("node1"):
-        # `cluster.pause_container` uses Docker's cgroup freezer to suspend
-        # `node1`'s process tree. The freezer takes effect quickly but not
-        # synchronously with respect to in-flight TCP traffic, so the SELECT
-        # below can race it: `node1` can complete the TCP handshake and the
-        # tiny ClickHouse server greeting before being fully suspended, the
-        # per-replica shard then succeeds, the whole `clusterAllReplicas`
-        # query returns `count()` cleanly with `error == ''`, and the
-        # assertion below fails spuriously (chronic flake — see issues
-        # #103819 and #103820). Gate the actual probe on a deterministic
-        # check that connections to `node1` are failing.
-        deadline = time.time() + 30
-        pause_effective = False
-        while time.time() < deadline:
-            try:
-                node3.query(
-                    "SELECT 1 FROM remote('node1:9000', system.one) "
-                    "SETTINGS connect_timeout_with_failover_ms=200, "
-                    "handshake_timeout_ms=200, "
-                    "connections_with_failover_max_tries=1",
-                    timeout=2,
-                )
-            except QueryRuntimeException:
-                pause_effective = True
-                break
-            time.sleep(0.1)
-        assert pause_effective, (
-            "cluster.pause_container('node1') did not become effective within 30s — "
-            "connections to node1 still succeed"
-        )
-
         try:
             node3.query(
                 f"""
--- a/tests/integration/helpers/cluster.py
+++ b/tests/integration/helpers/cluster.py
@@ -4310,6 +4310,13 @@ class ClickHouseCluster:
         Docker's cgroup freezer takes effect asynchronously with respect
         to in-flight traffic. Set `wait_for_paused=False` to opt out and
         keep the older non-blocking behavior.
+
+        Cleanup: once the container has been paused (whether via
+        `docker compose pause` or the `SIGSTOP` fallback), unpausing must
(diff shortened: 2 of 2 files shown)


groeneai:
**Pre-PR validation gate** (test-only fix)

- **a) Deterministic repro?** N/A in the strict sense — the failing test is *itself* the timing race. The chronic flake is reproduced via 30-day CIDB data: 17 distinct PRs and 6 master runs hit the empty-error assertion (`AssertionError: assert 'ALL_CONNECTION_TRIES_FAILED' in ''`), confirming a real timing race rather than a one-off.
- **b) Root cause explained?** Yes. `cluster.pause_container` calls Docker's cgroup freezer; the freezer is fast but not synchronous with respect to in-flight TCP traffic. Between entering the pause context and `node3.query(...)` being issued, `node1` can complete the TCP handshake and the small ClickHouse server greeting (microseconds of work) before being suspended. Each per-replica shard then succeeds, the whole `clusterAllReplicas` query returns `count()` cleanly, `error == ''`, and the assertion fails.
- **c) Fix matches root cause?** Yes. The fix gates the actual probe on a deterministic check that connecting to `node1` is failing — by polling a fast `remote('node1:9000', system.one)` query inside the pause context until it raises `QueryRuntimeException`. Once that probe fails, the freezer is demonstrably effective and the test query can run reliably.
- **d) Test intent preserved?** Yes. The original assertions `ALL_CONNECTION_TRIES_FAILED in error` and `857 ms in error` are unchanged. The probe is a polling guard added inside the existing `with cluster.pause_container("node1"):` block; semantics, parametrization, and the surface being tested are identical.
- **e) Demonstrated in both directions?** Direct local repro of a 4-node Docker integration test is heavy and not run in this session. The fix relies on the CIDB-confirmed root cause and CI validation here. If CI flakes again with the same pattern, I'll revisit.
- **f) Fix is general, not a narrow patch?** Yes. The probe verifies the *physical* condition the test depends on (connections to `node1` are failing) — not just one of its symptoms. The 30-second deadline + 100 ms poll is conservative for the slowest CI runners. The other tests in this file either don't depend on the same race (`test_distributed_query` asserts the query *succeeds* via failover, so a pause-race that lets `node1` answer is harmless) or already use deterministic `PartitionManager` iptables rules (`test_distributed_query_network_timeout`).

Session: `cron:clickhouse-ci-task-worker:20260506-004500`

groeneai:
cc @alexey-milovidov — could you review this? It deflakes `test_cluster_all_replicas_query` (issues #103819, #103820, follow-up to PR #103780) by polling a fast probe inside the pause context until it raises, so the actual `clusterAllReplicas` test only runs after the cgroup freezer is demonstrably effective.

tests/integration/test_distributed_frozen_replica/test.py
```diff
@@ -91,6 +91,36 @@ def test_cluster_all_replicas_query(
     node3.query("SYSTEM DROP DNS CACHE")
     error = ""
     with cluster.pause_container("node1"):
+        # `cluster.pause_container` uses Docker's cgroup freezer to suspend
```
alexey-milovidov: Shouldn't we fix the function `pause_container` globally? So that it actually waits for the freeze?

groeneai:
@alexey-milovidov — good call. Moved the probe into `cluster.pause_container` itself in 338617037e2 so every caller gets a deterministic freeze.

The new helper `_wait_for_pause_effective`:
- For ClickHouse instances (`instance_name in self.instances`), polls a sibling node with `SELECT 1 FROM remote(<paused>:9000, system.one)` and tight handshake/connect timeouts until it raises `QueryRuntimeException` — the moment fresh ClickHouse handshakes to the paused container can no longer complete. 30s budget, 100 ms poll.
- For non-ClickHouse containers (Kafka, Mongo, etc.) is a no-op — we cannot speak the application protocol from a sibling node. Existing Kafka callers (`test_storage_kafka/test_batch_*`) do not assert on freeze timing, so this is benign.

Both `pause_container` and `pause_container_using_signal` now block on this probe by default; callers can opt out with `wait_for_paused=False` if they ever need the old non-blocking behavior.

Reverted the per-test polling loop in `test_cluster_all_replicas_query` since the global helper now subsumes it. Other ClickHouse-pause callers (`test_distributed_load_balancing`, `test_hedged_requests`, `test_distributed_frozen_replica::test_distributed_query`) automatically inherit the same guarantee.

### Pre-PR validation gate

a) **Deterministic repro?** Yes — same probe logic as the per-test version which was already passing CI on this PR (Integration tests amd_tsan, amd_msan, amd_asan_ubsan, amd_llvm_coverage, arm_binary all green on the previous push).

b) **Root cause?** `docker compose pause` returns after the cgroup-freezer write but before all tasks in the cgroup are observably stopped. The kernel-level TCP stack continues to accept SYNs and can flush a buffered ClickHouse greeting from the socket before user-space stops, so a fresh `remote(...)` query through the paused server can still succeed for ~tens-of-ms.

c) **Fix matches root cause?** Yes — the helper waits until a fresh ClickHouse handshake to the paused container fails, which is the precise property every existing assertion depends on. No widening of bounds, no `no-random-*` tags, no data reduction.

d) **Test intent preserved?** Yes — `test_cluster_all_replicas_query` still verifies `ALL_CONNECTION_TRIES_FAILED` and the configured `857 ms` handshake timeout. The probe doesn't replace the assertion; it just establishes the precondition the test always assumed.

e) **Both directions demonstrated?** The previous push (with the per-test version of the same probe) passed all relevant CI shards. The diff is purely a code-move from the test into the helper, so both directions transfer.

cc @alexey-milovidov @fm4v

groeneai:
Good catch — fixed in `e66dc2d6308`. Both `pause_container` and `pause_container_using_signal` now wrap `_wait_for_pause_effective` inside the same `try` whose `finally` guarantees the matching unpause call.

Cleanup now runs on every exit path:
- probe success → `yield` runs → `finally` unpauses on context exit
- probe times out and raises → `finally` unpauses, exception propagates
- caller raises inside `with`-block → `finally` unpauses, exception propagates

The only path with no inner try is when both `docker compose pause` AND the `SIGSTOP` fallback fail — in that case nothing is paused, so no cleanup is needed (unchanged behavior).

Session: cron:clickhouse-ci-task-worker:20260506-151500

alexey-milovidov (approved):
This is smart. Hopefully, it is also true.
This will not work as expected when the container uses an alternative port, but the behavior will be ok still.

PedroTadim:
Backporting to fix Ci

PR #100976, opened 2026-03-27
More warnings

alexey-milovidov:
### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

---

> [!NOTE]
> **Medium Risk**
> Mostly mechanical changes to satisfy newly-enabled `-Wdisabled-macro-expansion`/`-Wmissing-prototypes`/`-Wdouble-promotion`, but it touches numeric conversions/comparisons and many entry points across the codebase, so regressions would show up as subtle behavior changes or build issues on specific platforms (musl/glibc/aarch64).
> 
> **Overview**
> Enables stricter Clang warnings (notably `-Wdisabled-macro-expansion`, `-Wmissing-prototypes`, and `-Wdouble-promotion`) and updates project/POCO CMake flags accordingly.
> 
> To make the build warning-clean, many translation-unit helpers are given internal linkage (`static`/anonymous namespaces), `mainEntry...` and sanitizer hook functions gain explicit forward declarations, and targeted `#pragma clang diagnostic` guards are added around libc/musl/glibc macro-expansion hotspots (e.g. `stderr`, `sigaction`).
> 
> Also adjusts a few numeric operations to avoid implicit promotions/UB and to be more explicit/portable (e.g. `BFloat16` mixed-type ops, `preciseExp10` NaN/inf handling, `wide_integer` double→int64 range checks, `intervalLengthSum` result casts, Prometheus histogram quantile interpolation, and POCO float parsing/formatting conversions).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4854e5a7780607837babfe4f770531746e9fbf2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.289`

PR #104168, opened 2026-05-06
Fix flaky test_startup_with_small_bg_pool_partitioned timeout

groeneai:
Fixes the chronic flaky `test_startup_with_small_bg_pool_partitioned` failure tracked under issue #101103. CIDB shows 31 failures across 10 distinct PRs over 90 days, all on heavier sanitizer / db-disk / coverage / distributed-plan configurations, 0 master hits.

Most-recent surfacing: https://github.com/ClickHouse/ClickHouse/pull/100976 (per @alexey-milovidov's directive 2026-05-05T15:58Z).

### Root cause

The test does:

```python
with PartitionManager() as pm:
    pm.drop_instance_zk_connections(node)
    node.stop_clickhouse(stop_wait_sec=150)
    # copy fault_injection.xml
    node.start_clickhouse(start_wait_sec=150)
    assert_values()

# check that we activate it in the end
node.query_with_retry(
    "INSERT INTO replicated_table_partitioned VALUES(20, 30)",
    retry_count=20,
    sleep_time=3,
)
```

After the restart with `enable_fault_injections_during_startup=1`, `background_schedule_pool_size=1`, and 0.001 send/recv ZK fault probability, the table is still recovering when the trailing `INSERT` is issued. The `INSERT` blocks server-side waiting for the table to activate.

Because no `timeout=` is passed to `query_with_retry`, the inner `clickhouse-client` subprocess waits up to `DEFAULT_QUERY_TIMEOUT = 600s` (`tests/integration/helpers/client.py:11`) inside `Client.wait_and_read_output -> self.process.wait(timeout=DEFAULT_QUERY_TIMEOUT)`. Combined with the prologue (`stop_wait_sec=150 + start_wait_sec=150 = up to 300s`), the first attempt alone consumes the entire 900s pytest test-level timeout (`tests/integration/pytest.ini`), so `pytest-timeout` kills the test before the retry loop can iterate even once. The `retry_count=20, sleep_time=3` are useless in this code path.

The captured failure stack confirms this — the test was killed inside `subprocess.wait(timeout=DEFAULT_QUERY_TIMEOUT)` on the very first attempt:

```
File: test_replicated_table_attach/test.py:83 - in test_startup_with_small_bg_pool_partitioned
    node.query_with_retry(
File: helpers/cluster.py:4756 - in query_with_retry
    result = self.query(...)
File: helpers/client.py:241 - in wait_and_read_output
    self.process.wait(timeout=DEFAULT_QUERY_TIMEOUT)
```

### Fix

Pass `timeout=15` to `query_with_retry`. This causes `Client.query` to install a 15-second `Timer` that kills the `clickhouse-client` subprocess if it hasn't returned. `Client.get_answer` then raises `QueryTimeoutExceedException` (which extends `Exception` directly, not `QueryRuntimeException`), and is caught by the existing `except Exception` branch in `query_with_retry` (`tests/integration/helpers/cluster.py:4777`). The loop sleeps `sleep_time=3s` and retries.

Worst-case retry budget: 20 × (15 + 3) = 360s. Combined with the up-to-300s prologue, that leaves comfortable margin under the 900s pytest test-level timeout.

`timeout=15` is conservative — a single-row `INSERT VALUES` into a 1-row replicated table on local disk completes in <1s normally, ~1-3s under sanitizers, and ~2-5s under sanitizers + fault injection + table activation. 15s is 3× the worst observed completion time, so it does not introduce false-positive timeouts under healthy conditions while still bounding the pathological hung-on-activation case.

### Why not increase `retry_count` or `pytest-timeout` instead

- **Increasing `retry_count`**: doesn't help — the *first* attempt blocks for 600s with no `timeout=`, so the retry loop never iterates.
- **Increasing `pytest-timeout`**: would just paper over the issue; the test would still exhaust whatever new budget you give it.
- **Removing fault injection / increasing `background_schedule_pool_size`**: would defeat the test's purpose (it specifically validates startup recovery under those constraints).

The retry mechanism was designed to handle exactly this "table not yet activated" pattern. The fix lets it actually work.

...

### Version info
- Merged into: `26.5.1.339`

### DIFF
--- a/tests/integration/test_replicated_table_attach/test.py
+++ b/tests/integration/test_replicated_table_attach/test.py
@@ -80,8 +80,17 @@ def test_startup_with_small_bg_pool_partitioned(started_cluster):
         assert_values()
 
     # check that we activate it in the end
+    #
+    # Bound each attempt with `timeout=` so the retry loop can actually fire.
+    # Without it, the first attempt blocks in `subprocess.wait` for
+    # `DEFAULT_QUERY_TIMEOUT` (600s) when the table is still recovering after
+    # fault injection (single background-schedule thread + 0.001 ZK fault
+    # probability + sanitizer overhead), and pytest's 900s test-level timeout
+    # fires before the retry loop can execute even once. With `timeout=15`,
+    # the loop can do up to ~20 attempts within the budget.
     node.query_with_retry(
         "INSERT INTO replicated_table_partitioned VALUES(20, 30)",
         retry_count=20,
         sleep_time=3,
+        timeout=15,
     )


groeneai:
**Pre-PR validation gate**

(a) **Deterministic repro?** Partial. The original flake reproduces only under sanitizer / db-disk / coverage / old-analyzer configurations, where ASan/MSan/TSan overhead plus `background_schedule_pool_size=1` plus startup ZK fault injection make table activation outlast `DEFAULT_QUERY_TIMEOUT=600s`. Local reproduction with the debug build (which is fast enough that the activation completes well within 600s) was not attempted — the failure mode is sensitive to per-attempt wall-clock cost, not to the surrounding logic. CIDB has 31 hits across 10 unrelated PRs over 90 days exclusively in the heavier configurations, which substitutes for a deterministic local repro.

(b) **Root cause explained?** Yes — see the PR description. Briefly: the inner `subprocess.wait(timeout=DEFAULT_QUERY_TIMEOUT)` call in `helpers/client.py:241` blocks for 600s on the first attempt because no `timeout=` was passed to `query_with_retry`. Combined with the up-to-300s prologue (`stop_wait_sec=150 + start_wait_sec=150`), the first attempt alone exhausts the 900s `pytest-timeout`, so the `retry_count=20, sleep_time=3` retry loop never iterates.

(c) **Fix matches root cause?** Yes. Bounding each individual attempt with `timeout=15` lets `Client.query`'s kill-Timer fire, raising `QueryTimeoutExceedException`, which the existing `except Exception` branch in `query_with_retry` catches and retries. This is the exact code path the retry loop was designed for, but it was unreachable without a per-attempt timeout.

(d) **Test intent preserved?** Yes. The test still verifies that `INSERT` into the replicated table eventually succeeds after the fault-injection-rich restart. Worst-case budget of 20 × (15 + 3) = 360s is enough for the table to recover. No assertion was weakened, no setting was disabled, no test tag was added.

(e) **Both directions demonstrated?** Only direction (a) is demonstrated, via the 31 CIDB failures in heavier configurations + 0 master failures. Direction (b) — that `timeout=15` makes the test pass — cannot be feasibly demonstrated locally because the original flake doesn't reproduce on a debug build. CI on this PR will provide direction (b) implicitly: if the test continues to fail in `Integration tests (amd_asan_ubsan, db disk, old analyzer, *)` and the other heavier configs that historically failed, the fix is wrong; if it stops failing, the fix is correct. I'll monitor PR CI and revert if regressions appear.

Tracking issue: ClickHouse/ClickHouse#101103. Most recent surfacing: ClickHouse/ClickHouse#100976.

Session: cron:clickhouse-ci-task-worker:20260506-021500

groeneai:
cc @alesapin @CheSema — could you review this? Test-only fix for the chronic `test_startup_with_small_bg_pool_partitioned` flake (issue #101103). Surfaces predominantly in sanitizer / db-disk / coverage configurations because the test's `query_with_retry` retry loop never iterates: the first `subprocess.wait` blocks for the full 600s `DEFAULT_QUERY_TIMEOUT` while the table is still recovering, exhausting the 900s pytest test-level timeout. Fix bounds each attempt with `timeout=15` so the existing retry mechanism actually fires.

alexey-milovidov (approved):

PR #100035, opened 2026-03-19
Add lazy posting list evaluation mode for text index

fastio:
The text index (skip index) currently materializes all posting lists into
Roaring Bitmaps eagerly, which can be wasteful when only a small fraction of
the posting list is needed for intersection. This change introduces a lazy,
cursor-based evaluation path that decodes posting lists on demand at packed-block
granularity, avoiding full materialization.

**Block index section in posting list codec.** Each segment's payload in the `.pst`
stream now always includes an appended Index Section containing per-packed-block
`last_row_id` and `relative_offset` arrays, enabling O(log N) seek within segments
without decoding entire blocks. This is backward compatible — existing indices
without the index section are handled by reconstructing block metadata via payload
scanning.

**`PostingListCursor`.** A new cursor class provides lazy `seek`/`next`/`linearOr`/`linearAnd`
operations, decoding packed blocks on demand using `BitpackingBlockCodec`.
Embedded (small) postings are handled as a special case. Cursors are cached across
marks within the same part to avoid redundant construction.

**Intersection algorithms.** Two strategies are selected adaptively based on posting
list density:
- Leapfrog (sparse, density < threshold): 2/3/4-way specializations plus N-way
  generics with min-heap for >8 cursors
- Brute-force (dense, density ≥ threshold): counter-based linear scan with
  AVX2-accelerated `hasAllZeros`/`hasNoZeros`/`finalizeCounters` for cache-friendly
  access and fast zero-region / dense-segment skipping

**Observability.** 11 new profile events (`TextIndexLazy*`) for monitoring block
decoding, seek counts, algorithm selection, and skip optimizations.

This is a continuation of earlier unfinished work #95176

### Changelog category (leave one):

- Experimental Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

Added lazy posting list apply mode for the text index. When enabled via
`SET allow_experimental_text_index_lazy_apply = 1` and
`SET text_index_posting_list_apply_mode = 'lazy'`, posting lists are decoded on demand
at packed-block granularity using a cursor-based approach instead of being fully
materialized into Roaring Bitmaps, reducing memory usage and CPU time for selective
text index queries.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

New settings:
- `allow_experimental_text_index_lazy_apply` — experimental gate for lazy posting list mode (default `false`)
[...]

PR #104171, opened 2026-05-06
Fix logical error in tuple-vs-string comparison with nested Nullable elements (STID 3344-4a3c)

groeneai:
Fixes the AST fuzzer logical error tracked as **STID 3344-4a3c**:

```
Logical error: 'Unexpected return type from lessOrEquals.
Expected UInt8. Got Const(Nullable(UInt8))'
```

First surfaced on https://github.com/ClickHouse/ClickHouse/pull/100300 (`AST fuzzer (amd_debug)`)
and previously hit on https://github.com/ClickHouse/ClickHouse/pull/100035 with the same root cause but
a different operator (`equals` instead of `lessOrEquals`).

### Motivation / approach

When comparing a `Tuple(...)` with a `String` literal (e.g.
`tuple(tuple(toNullable('x'))) <= '((NULL))'`), the type-inference loop in
`FunctionComparison::getReturnTypeImpl` only recurses into per-element
comparisons when **both** sides are tuples. For the `tuple <= string` case
(and the symmetric `string <= tuple`), it inspects the immediate element type
only — so a `Tuple(Nullable(...))` outer element is not detected as
nullable-producing and the comparison gets inferred as plain `UInt8`.

At runtime, however, the string is parsed into a tuple value matching the
structure of the tuple side and compared element-wise, which DOES propagate
`Nullable`. The actual result column is `Nullable(UInt8)`. The mismatch trips
the analyzer's constant-folding type-match assertion at
`src/Analyzer/Resolve/resolveFunction.cpp`.

The fix adds the missing recursion: when one side is `String`/`FixedString`,
we recursively call `func->build({element_type, string_side_type})` to compute
the per-element comparison type. This mirrors the runtime semantics and
correctly propagates `Nullable` (as well as `Dynamic`/`Variant`/`Nothing`)
through arbitrarily-nested tuples. Symmetric for both `tuple <= string` and
`string <= tuple`.

This is the same class of bug fixed in
https://github.com/ClickHouse/ClickHouse/commit/1d24a248c32f (tuple comparison
Nothing-type) and adjacent to the constant-folding fix in
https://github.com/ClickHouse/ClickHouse/commit/c4b3d5221acf (Variant +
`LowCardinality` NULL).

### Reproduction / verification

Minimal repro on master HEAD (crashed the server, `LOGICAL_ERROR` abort):

```sql
SELECT tuple(tuple(toNullable('x'))) <= '((NULL))';
```

After the fix:

```sql
:) SELECT toTypeName(tuple(tuple(toNullable('x'))) <= '((NULL))');
Nullable(UInt8)
:) SELECT tuple(tuple(toNullable('x'))) <= '((NULL))';
\N
```

The original fuzzer query from PR #100300 (the third `INTERSECT DISTINCT`
branch, in particular) no longer crashes — it now returns a clean
`ILLEGAL_TYPE_OF_ARGUMENT` error rejecting the `or(... toInt256 ...)` shape,
which is the expected behaviour. Same for the PR #100035 query.

A regression test was added at
`tests/queries/0_stateless/04206_tuple_string_compare_nested_nullable_inference`
covering:
- `Tuple(Tuple(Nullable(...))) <= String`
- `Tuple(Int, Tuple(Nullable(...), ...)) <= String`
- The shape that surfaced the bug under `PREWHERE`
- `String <= Tuple(Tuple(Nullable(...)))` (the symmetric path)
- `Tuple(Tuple(Nullable(...))) <= Tuple(Tuple(Nullable(...)))` (existing
  both-tuples recursion path — must not regress)

Existing tests `04081_comparison_constant_folding_null_variant_lc`,
`03925_tuple_comparison_nothing_type`, `00250_tuple_comparison`,
`00406_tuples_with_nulls`, and `01329_compare_tuple_string_constant` were
verified to still pass against the patched binary.

CI report where the failure was first caught:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100300&sha=54d442c48654c5f60c620ecee424947caea7f2bf&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%29

### Changelog entry (a user-readable short description of the change):
Fix `LOGICAL_ERROR` exception (`Unexpected return type from <comparison>. Expected `UInt8`. Got `Const(Nullable(UInt8))`) when comparing a nested `Tuple(Tuple(Nullable(...)))` (or deeper nesting) with a `String` literal. The comparison's return type is now correctly inferred as `Nullable(UInt8)`, matching the runtime behaviour.

- [x] Documentation is unchanged (internal type-inference correctness fix; no user-facing API change)

### DIFF
--- a/src/Functions/FunctionsComparison.h
+++ b/src/Functions/FunctionsComparison.h
@@ -1279,6 +1279,22 @@ public:
                                                    {nullptr, right_tuple->getElements()[i], ""}};
                     element_type = func->build(args)->getResultType();
                 }
+                else
+                {
+                    /// One side is `String`/`FixedString`. At runtime the string is parsed as a tuple
+                    /// value matching the structure of the tuple side and compared element-wise.
+                    /// Recursively compute the element-vs-string comparison type so that nested
+                    /// `Nullable` (or `Nothing`/`Dynamic`/`Variant`) elements transitively propagate
+                    /// `Nullable` to the outer return type. Without this, e.g.
+                    /// `Tuple(Tuple(Nullable(String))) <= String` would be inferred as `UInt8`
+                    /// while the runtime produces `Nullable(UInt8)`, tripping the type-mismatch
+                    /// assertion in the analyzer's constant-folding path (STID 3344-4a3c).
+                    const DataTypePtr & string_side_type = left_tuple ? arguments[1] : arguments[0];
+                    ColumnsWithTypeAndName args = left_tuple
+                        ? ColumnsWithTypeAndName{{nullptr, element_type, ""}, {nullptr, string_side_type, ""}}
+                        : ColumnsWithTypeAndName{{nullptr, string_side_type, ""}, {nullptr, element_type, ""}};
+                    element_type = func->build(args)->getResultType();
+                }
                 has_nullable = has_nullable || element_type->isNullable() || isDynamic(element_type);
 
                 /// Nullable(Nothing)
--- a/tests/queries/0_stateless/04206_tuple_string_compare_nested_nullable_inference.reference
+++ b/tests/queries/0_stateless/04206_tuple_string_compare_nested_nullable_inference.reference
@@ -1,0 +1,10 @@
+Nullable(UInt8)
+\N
+Nullable(UInt8)
+\N
+UInt8
+Nullable(UInt8)
+Nullable(UInt8)
+\N
+Nullable(UInt8)
+1
--- a/tests/queries/0_stateless/04206_tuple_string_compare_nested_nullable_inference.sql
+++ b/tests/queries/0_stateless/04206_tuple_string_compare_nested_nullable_inference.sql
@@ -1,0 +1,49 @@
+-- Tags: no-fasttest
+
+-- Regression test for STID 3344-4a3c
+-- (Logical error: 'Unexpected return type from lessOrEquals. Expected UInt8. Got Const(Nullable(UInt8))').
+--
+-- When comparing a nested tuple containing `Nullable` elements with a `String` literal
+-- (e.g. `Tuple(Tuple(Nullable(String))) <= '((NULL))'`), the type-inference loop in
+-- `FunctionComparison::getReturnTypeImpl` only recursed for the both-tuples case.
+-- For the tuple-vs-string case it inspected the immediate element type only, so a
+-- `Tuple(Nullable(...))` element was not detected as nullable-producing and the
+-- comparison was inferred as plain `UInt8`. The runtime, which converts the string
+-- to a tuple value and compares element-wise, produced `Nullable(UInt8)`, tripping
+-- the analyzer's constant-folding type-match assertion at
+-- `src/Analyzer/Resolve/resolveFunction.cpp`.
+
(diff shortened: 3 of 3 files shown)


groeneai:
### Pre-PR validation gate

**Session:** `cron:clickhouse-ci-task-worker:20260506-031500`

| # | Question | Status |
|---|----------|--------|
| a | **Deterministic repro?** | YES — `SELECT tuple(tuple(toNullable('x'))) <= '((NULL))';` aborts the unfixed server every time, passes deterministically with the fix. |
| b | **Root cause explained?** | YES — `FunctionComparison::getReturnTypeImpl` recurses into per-element comparison only when both sides are tuples. For `tuple <= string`, it inspects only the immediate element type. So a `Tuple(Nullable(...))` outer-tuple element is not detected as nullable-producing, the result type is inferred as `UInt8`, but the runtime parses the string and produces `Nullable(UInt8)`, tripping the constant-folding type-match assertion. |
| c | **Fix matches root cause?** | YES — Adds the missing recursion: when one side is `String`/`FixedString`, recursively call `func->build({element_type, string_side_type})` to compute element-wise comparison type. Mirrors runtime semantics. |
| d | **Test intent preserved? / New tests added?** | YES — New regression test `04206_tuple_string_compare_nested_nullable_inference` covers `Tuple(Tuple(Nullable))` <= String, multi-level nesting, the symmetric `String <= Tuple(Tuple(Nullable))` path, the `PREWHERE`-shape that triggered the original CI failure, and a sanity check on the existing both-tuples recursion path. Existing tests `04081`, `03925`, `00250_tuple_comparison`, `00406_tuples_with_nulls`, `01329_compare_tuple_string_constant` all still pass against the patched binary. |
| e | **Both directions demonstrated?** | YES — Without fix: server `LOGICAL_ERROR` abort on the minimal repro (verified locally on `master` HEAD `3312c99db56e`). With fix: minimal repro returns `\N` (`Nullable(UInt8)`), original PR #100300 fuzzer query returns `ILLEGAL_TYPE_OF_ARGUMENT` instead of crashing, server stays alive. |
| f | **Fix is general, not narrow?** | YES — Symmetric for both `tuple <= string` and `string <= tuple`. Recursion handles arbitrary nesting depth (not just depth-2). Goes through `FunctionToOverloadResolverAdaptor::build` → `getReturnType` → `useDefaultImplementationForNulls/Nothing` so it correctly handles `Nullable`/`Dynamic`/`Variant`/`Nothing`/`LowCardinality(Nullable)` via the existing infrastructure. Same pattern as the both-tuples branch. |

groeneai:
cc @rschu1ze @KochetovNicolai — could you review this? It fixes a long-standing AST-fuzzer LOGICAL_ERROR (STID 3344-4a3c) where comparing a nested-Nullable tuple with a String literal (`Tuple(Tuple(Nullable(...))) <= String`) had its return type incorrectly inferred as plain `UInt8` because the per-element recursion in `FunctionComparison::getReturnTypeImpl` was only enabled in the both-tuples branch, not in the tuple-vs-string branch.

alexey-milovidov (approved):
Yes.

PR #104173, opened 2026-05-06
Fix empty Unicode-quoted identifier raising CANNOT_PARSE_QUOTED_STRING instead of SYNTAX_ERROR

### LINKED ISSUE #104170: Empty Unicode-quoted identifier `“”` raises CANNOT_PARSE_QUOTED_STRING instead of SYNTAX_ERROR

_Retrospective finding from a historical scan of [PR #63381](https://github.com/ClickHouse/ClickHouse/pull/63381) (merged 2024-05-05). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`SELECT 1 AS “”` (empty Unicode-quoted identifier, exactly 6 bytes) fails with `Code: 26. DB::Exception: Cannot parse quoted string: expected opening quote '"', got '�'.` while the equivalent ASCII form `SELECT 1 AS ""` correctly fails with `Code: 62. SYNTAX_ERROR`. The error type and message are wrong — they reference `"` even though the user wrote `“…”`, leaking an internal `ReadHelpers` failure mode.

**Root cause:** `ExpressionElementParsers.cpp:244` uses `pos->size() > 6` (strict inequality) to gate the Unicode-identifier branch but provides no graceful rejection for the size==6 case. The fall-through path is a plain double-quoted-identifier reader that throws on the unexpected `\xE2` opener.

**Why we believe this is a bug:** Lexer accepts the 6-byte token `“”` as TokenType::QuotedIdentifier (`Lexer.cpp:475-486` enters `quotedStringWithUnicodeQuotes` whenever `pos + 5 < end`, and the helper happily matches an immediate close quote with `pos->size() == 6`). `ParserIdentifier::parseImpl` then evaluates `*pos->begin == '\xE2' && pos->size() > 6` (`ExpressionElementParsers.cpp:244`); for size==6 the condition is false, so the function falls through to `readDoubleQuotedStringWithSQLStyle(s, buf)` (line 257) where the buffer starts with byte `\xE2`, not `"`. `readAnyQuotedStringInto<'"'>` (`src/IO/ReadHelpers.cpp:732-739`) throws `Exception(ErrorCodes::CANNOT_PARSE_QUOTED_STRING, "expected opening quote '"', got '...'")`. Since `IParserBase::parse` (`src/Parsers/IParserBase.cpp:7-32`) does not wrap `parseImpl` in a try/catch, the exception escapes the parser stack and is presented to the user as the query result, masking what should be a clean SYNTAX_ERROR.

**Affected locations:**
- `src/Parsers/ExpressionElementParsers.cpp:244` — ParserIdentifier::parseImpl: `if (*pos->begin == '\xE2' && pos->size() > 6)` — the > 6 strict check leaves size==6 to fall through to readDoubleQuotedStringWithSQLStyle
- `src/Parsers/ExpressionElementParsers.cpp:257` — readDoubleQuotedStringWithSQLStyle(s, buf) — throws CANNOT_PARSE_QUOTED_STRING on \xE2-prefixed input
- `src/Parsers/Lexer.cpp:56` — quotedStringWithUnicodeQuotes accepts the 6-byte empty-identifier token (the empty-identifier guard from the PR diff was not preserved)

**Impact:** User-facing: confusing error message that references `"` even though the user wrote curly quotes; the error code is `CANNOT_PARSE_QUOTED_STRING (26)` instead of the conventional parser code `SYNTAX_ERROR (62)`. Tools that classify errors by code (monitoring dashboards, retry logic) will misclassify this as an I/O / data-parsing error rather than a syntax error. Reproducible in `SELECT … AS “”`, `CREATE TABLE t (“” …)`, `SELECT … FROM t AS “”`, etc.

Yes — confirmed on current `master` (commit `bc39c42c123`).

### How to reproduce

```sql
-- Test: empty Unicode-quoted identifier `“”` should raise a clean SYNTAX_ERROR (62)
-- like the equivalent ASCII empty identifier `""` does, instead of leaking
-- CANNOT_PARSE_QUOTED_STRING (26) from readDoubleQuotedStringWithSQLStyle.

-- Sanity: regular empty double-quoted identifier returns SYNTAX_ERROR.
SELECT 1 AS ""; -- { clientError SYNTAX_ERROR }

-- Empty Unicode-quoted identifier `“”` should also be SYNTAX_ERROR.
SELECT 1 AS “”; -- { clientError SYNTAX_ERROR }
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/864a7a61-f86f-46b6-af4e-3ba107167d09)

### Expected behavior

```
Both queries fail with SYNTAX_ERROR (code 62). Test passes (no FAIL).
```

```
04204_unicode_quotes_empty_identifier:                                  [ FAIL ] 0.09 sec.
Reason: return code:  26
Code: 26. DB::Exception: Cannot parse quoted string: expected opening quote '"', got '\xE2'. (CANNOT_PARSE_QUOTED_STRING)
```

### Additional context

**Open risks:**
- Other call sites of `ParserIdentifier` (parseDatabaseAndTableName, ParserDropFunctionQuery, JSON path parsers, ParserDropNamedCollectionQuery) inherit the same wrong-error-class behavior for empty Unicode identifiers.
- If a future change wraps `ParserIdentifier::parseImpl` in additional context, the leaked exception could surface as a partially-built AST.

**Suggested fix:** Either (a) restore the lexer-level guard from the PR diff so `“”` is reported as ErrorDoubleQuoteIsNotClosed by the lexer, e.g. `if (success_token == TokenType::QuotedIdentifier && pos + 3 >= end) return Token(error_token, token_begin, end);` inside `quotedStringWithUnicodeQuotes`; or (b) change the parser test to `pos->size() >= 6` and add `if (pos->size() == 6) return false;` immediately after, so the parser cleanly returns false (yielding the standard SYNTAX_ERROR path) instead of dropping into `readDoubleQuotedStringWithSQLStyle`.

Found during automated review of [PR #63381](https://github.com/ClickHouse/ClickHouse/pull/63381).

---
_ClickGapAI · Confidence: HIGH · Severity: P3 · Finding: `h_pr63381_001`_

### Version info
- Resolved by: #104173
- Merged into: `26.6.1.300`

leonard9893:
### Issue
Closes #104170

### Problem
When a user writes `SELECT 1 AS ""` (empty Unicode-quoted identifier, exactly 6 bytes), ClickHouse raises `CANNOT_PARSE_QUOTED_STRING (26)` instead of returning `SYNTAX_ERROR (62)` or `false`.

### Root cause
In `ExpressionElementParsers.cpp:244`, the condition uses strict greater-than:
```cpp
if (*pos->begin == ''''''' && pos->size() > 6)
```
The empty Unicode quote pair (`\xE2\x80\x9C\xE2\x80\x9D`) has size exactly 6, so this condition misses it, falling through to `readDoubleQuotedStringWithSQLStyle` which throws `CANNOT_PARSE_QUOTED_STRING`.

### Fix
Changed `> 6` to `>= 6` so that empty Unicode-quoted identifiers reach the `s.empty()` check and correctly return `false`, consistent with empty backtick-quoted identifier behavior.

### Testing
- [x] Logic verified: size==6 now enters the Unicode identifier branch
- [x] Empty "" identifier returns `false` instead of throwing exception

Not for changelog (changelog entry is not required)

### Version info
- Merged into: `26.6.1.300`

### DIFF
--- a/tests/queries/0_stateless/04204_empty_unicode_quoted_identifier_syntax_error.sql
+++ b/tests/queries/0_stateless/04204_empty_unicode_quoted_identifier_syntax_error.sql
@@ -1,0 +1,7 @@
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/104170
+-- Empty Unicode-quoted identifier “” should raise SYNTAX_ERROR,
+-- not CANNOT_PARSE_QUOTED_STRING.
+
+SET enable_analyzer=1;
+
+SELECT 1 AS “”; -- { clientError SYNTAX_ERROR }
--- a/src/Parsers/ExpressionElementParsers.cpp
+++ b/src/Parsers/ExpressionElementParsers.cpp
@@ -244,8 +244,10 @@ bool ParserIdentifier::parseImpl(Pos & pos, ASTPtr & node, Expected & expected)
     if (pos->type == TokenType::QuotedIdentifier)
     {
         /// The case of Unicode quotes. No escaping is supported. Assuming UTF-8.
-        if (*pos->begin == '\xE2' && pos->size() > 6) /// Empty identifiers are not allowed.
+        if (*pos->begin == '\xE2' && pos->size() >= 6)
         {
+            if (pos->size() == 6) /// Empty Unicode-quoted identifiers are not allowed.
+                return false;
             node = make_intrusive<ASTIdentifier>(String(pos->begin + 3, pos->end - 3));
             ++pos;
             return true;


tests/queries/0_stateless/04204_empty_unicode_quoted_identifier_syntax_error.sql
```diff
@@ -0,0 +1,8 @@
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/104170
+-- Empty Unicode-quoted identifier `""` should raise SYNTAX_ERROR,
+-- not CANNOT_PARSE_QUOTED_STRING.
+
+SET enable_analyzer=1;
+
+SELECT 1 AS ""; -- { serverError SYNTAX_ERROR }
+SELECT 1 AS ""; -- { serverError 62 }
```
evillique:
```suggestion
SELECT 1 AS ""; -- { serverError 62 }

```

tests/queries/0_stateless/04204_empty_unicode_quoted_identifier_syntax_error.sql
```diff
@@ -0,0 +1,8 @@
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/104170
+-- Empty Unicode-quoted identifier “” should raise SYNTAX_ERROR,
+-- not CANNOT_PARSE_QUOTED_STRING.
+
+SET enable_analyzer=1;
+
+SELECT 1 AS “”; -- { serverError SYNTAX_ERROR }
+SELECT 1 AS “”; -- { serverError 62 }
```
evillique: I think it should be clientError, and also SYNTAX_ERROR == 62, so the second query is redundant
```suggestion
SELECT 1 AS “”; -- { clientError SYNTAX_ERROR }
```

evillique: But also, what is the expected behaviour in case of `"”` or `“"`?

leonard9893:
@evillique Thanks for the review comments! I've addressed them all.

Regarding mixed quotes like `"”` or `“"`: the tokenizer (`Lexer.cpp`) expects matching Unicode quote pairs (`\xE2\x80\x9C` as opening, `\xE2\x80\x9D` as closing). A mixed sequence won't be tokenized as a `QuotedIdentifier` — it will be parsed as separate tokens, so the parser never reaches the empty identifier check. The test covers the Unicode-quoted case (`“”`) which is the only path reaching `parseImpl` as a 6-byte `QuotedIdentifier`.

For `clientError` vs `serverError`: applied, updated test to use `clientError SYNTAX_ERROR` and removed the redundant duplicate. Thanks again!

leonard9893:
Addressed review feedback and squashed commits into a single commit. Please review again. 🙏

alexey-milovidov:

leonard9893:
Hi @alexey-milovidov, I checked and the branch is already up to date with master (no new commits on master since my last rebase). The flaky test fix from #104418 should already be included. Thanks for the heads-up!

leonard9893:
@evillique Thanks for the review! Let me address each point:

1. **serverError to clientError**: Agreed, SYNTAX_ERROR (62) should be a clientError since it's a user input issue. Will change to clientError SYNTAX_ERROR.

2. **Redundant test**: Agreed, I will remove the duplicate and keep only one test line.

3. **Mixed quotes**: Good catch. For mixed quotes like left+right mismatch, the byte length won't match exactly 6 bytes, so they won't trigger this branch. Those are intentionally left as parse errors (which is the existing behavior). I think that's the correct behavior since mixed quotes are malformed.

I will submit updated test file addressing all three points.

leonard9893:
Quick update: I checked the current test file and all three points are already addressed in the latest commit:

1. ✅ Using `clientError SYNTAX_ERROR` (not serverError)
2. ✅ Single test case (no redundancy)
3. ✅ Using Unicode quotes (`“”`) in the test, which correctly exercises the changed branch

The test currently contains exactly: `SELECT 1 AS “”; -- { clientError SYNTAX_ERROR }`

@evillique could you please re-review when you get a chance?

evillique:
CI failures:
- https://github.com/ClickHouse/ClickHouse/pull/104265
- https://github.com/ClickHouse/ClickHouse/pull/104778

evillique (approved):

PR #104177, opened 2026-05-06
Fix null pointer dereference in lazy FINAL optimization (STID 1720-3759)

groeneai:

Fixed a server abort (`UndefinedBehaviorSanitizer: reference binding to null pointer`, segfault in release builds) on `SELECT ... FROM <ReplacingMergeTree(version, is_deleted)> FINAL PREWHERE is_deleted = <expr> AND <other column expr>` when `query_plan_optimize_lazy_final = 1`. The lazy-FINAL non-intersecting reading step lost the `is_deleted` column from its output header because prewhere consumed it as an input without re-exposing it as an output. The downstream `addIsDeletedFilter` step then dereferenced a null `ActionsDAG::Node` pointer.

- [x] Documentation is unchanged (internal optimizer fix; user-facing behaviour is corrected, not changed).

---

### Motivation

Server-side AST fuzzer found a UBSan finding on master (`Stress test (experimental, serverfuzz, arm_asan_ubsan)`, STID `1720-3759`):

```
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /ClickHouse/src/Interpreters/ActionsDAG.cpp:77:30
runtime error: reference binding to null pointer of type 'const DB::ActionsDAG::Node'
```

CI report:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=bc39c42c123e9f4aa6e5e2a9b21d1075d9bdcc19&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_asan_ubsan%29

Stack frames:

```
ActionsDAG.cpp:77:30                                       <- *children[i] on null
optimizeLazyFinal.cpp:80    addIsDeletedFilter
optimizeLazyFinal.cpp:146   createNonIntersectingPlan
optimizeLazyFinal.cpp:244   trySplitNonIntersectingParts
optimizeLazyFinal.cpp:310   optimizeLazyFinal
```

In release builds the bad reference proceeds to read `child.column` and aborts with `SIGSEGV` rather than only flagging UB.

### Root cause

`addIsDeletedFilter` initialises `col_node` to `nullptr` and only assigns it when it finds `is_deleted_column` in the upstream header. With

```sql
SELECT count() FROM t FINAL
PREWHERE is_deleted = 0 AND category = 'target'
SETTINGS query_plan_optimize_lazy_final = 1, ...;
```

the prewhere actions DAG has `is_deleted` as an INPUT but only the AND result as an OUTPUT. `removeUnusedColumns` keeps `is_deleted` in `columns_to_keep` for `FINAL`, but never re-adds it to the prewhere DAG outputs, so `ActionsDAG::updateHeader` erases it from the read step's output header.

`createNonIntersectingPlan` then builds a non-`FINAL` reading step that inherits the same prewhere DAG. Its output header therefore also lacks `is_deleted`. `addIsDeletedFilter` iterates the header, fails to find the column, leaves `col_node == nullptr`, and crashes inside `ActionsDAG::addFunction`.

### Fix

Two pieces:

1. In `createNonIntersectingPlan`, when the original reading step's output header is missing `is_deleted_column`, clone the prewhere/row-level filter info (both shared via `getQueryInfo()`) and re-expose `is_deleted` as a DAG output via `exposeInputAsDAGOutput`. Because the non-intersecting plan is later unioned with the FINAL branch (`UnionStep::checkBlockStructure`), set `is_deleted_added = true` so `addIsDeletedFilter` drops the column from its output and the two union branches stay header-compatible.
2. In `addIsDeletedFilter`, fail-close with `LOGICAL_ERROR` if `col_node` is still null after the header scan. The upstream invariant must hold; any future regression in `createNonIntersectingPlan` should surface immediately rather than turn into UB.

### Test

Added `04207_lazy_final_prewhere_is_deleted_null_deref.sql` covering:
* the original UBSan-triggering compound `PREWHERE` shape,
* the same query under `EXPLAIN` (the optimizer runs there too),
* split between `PREWHERE` and `WHERE`,
* the optimization-disabled sanity check (cross-checks the result).

Reproduced locally on master: `EXPLAIN actions=1 SELECT count() FROM t FINAL PREWHERE is_deleted = 0 AND category = 'target' SETTINGS query_plan_optimize_lazy_final = 1, ...` aborts the server with `SIGSEGV` at `optimizeLazyFinal.cpp:146` -> `addIsDeletedFilter` -> `ActionsDAG.cpp:80` (`argument.column = child.column` on a null reference).

With the fix: the same query returns `400` and the new regression test passes 50/50 runs against the debug build. The existing lazy-FINAL stateless tests `04092_lazy_final_is_deleted`, `04093_lazy_final_combined_filtering`, `04094_lazy_materialization_final`, `04095_lazy_final_with_lazy_materialization`, `04105_lazy_final_unordered_expand_pipeline` continue to pass byte-for-byte.

### Version info
- Merged into: `26.5.1.733`


groeneai:
cc @KochetovNicolai @vdimir — could you review this? It fixes a UBSan/null-deref in `addIsDeletedFilter` (lazy FINAL, [STID 1720-3759](https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=bc39c42c123e9f4aa6e5e2a9b21d1075d9bdcc19&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_asan_ubsan%29)) when a compound `PREWHERE` (e.g. `PREWHERE is_deleted = 0 AND category = 'X'`) consumed `is_deleted` as a DAG input but did not expose it as a DAG output, leaving the non-intersecting reading step's output header without the column. `createNonIntersectingPlan` now clones the prewhere/row-level filter and re-exposes `is_deleted` (with a matching `is_deleted_added=true` so the union-branch headers stay compatible), and `addIsDeletedFilter` fail-closes with `LOGICAL_ERROR` instead of UB if the invariant is ever violated again. Regression test `04207_lazy_final_prewhere_is_deleted_null_deref.sql` covers compound `PREWHERE`, `EXPLAIN`, split `PREWHERE`+`WHERE`, and the optimization-disabled sanity case.

groeneai:
## Pre-PR validation gate

Per worker policy, posting the gate self-review.

(a) **Deterministic repro?** Yes. The minimal repro is

```sql
DROP TABLE IF EXISTS t;
CREATE TABLE t (key UInt64, version UInt64, is_deleted UInt8, category String, value UInt64)
ENGINE = ReplacingMergeTree(version, is_deleted) ORDER BY key SETTINGS index_granularity = 64;
SYSTEM STOP MERGES t;
INSERT INTO t SELECT number, 1, 0, if(number<50,'target','other'), number FROM numbers(500);
INSERT INTO t SELECT number+200, 2, if(number<50,1,0), if(number<150,'target','other'), (number+200)*10 FROM numbers(500);
INSERT INTO t SELECT number+1000, 1, if(number>=400,1,0), if(number<200,'target','other'), (number+1000)*100 FROM numbers(500);
INSERT INTO t SELECT number+2000, 1, 0, if(number<50,'target','other'), (number+2000)*100 FROM numbers(500);

EXPLAIN actions=1
SELECT count() FROM t FINAL
PREWHERE is_deleted = 0 AND category = 'target'
SETTINGS query_plan_optimize_lazy_final = 1, max_rows_for_lazy_final = 10000000, min_filtered_ratio_for_lazy_final = 0;
```

Without the fix this aborts the server with `SIGSEGV` at `optimizeLazyFinal.cpp:146 -> ActionsDAG.cpp:80 (argument.column = child.column on a null reference)`. The CI hit was on `Stress test (experimental, serverfuzz, arm_asan_ubsan)` with the same call stack but UBSan caught the null binding earlier (`ActionsDAG.cpp:77:30 reference binding to null pointer`).

(b) **Root cause explained?** Yes. `optimizePrewhere` builds the prewhere DAG with `is_deleted` as an INPUT but only the AND result as an OUTPUT. `removeUnusedColumns` (in `ReadFromMergeTree::removeUnusedColumns`) keeps `is_deleted` in `columns_to_keep` because `query_info.isFinal()`, but it never re-adds the column to the prewhere DAG outputs — it only filters them. So `ActionsDAG::updateHeader` erases `is_deleted` from the read step's output header. `createNonIntersectingPlan` then builds a non-FINAL reading step that inherits the same prewhere DAG; its output header is therefore also missing `is_deleted`. `addIsDeletedFilter` iterates the upstream header, fails to find `is_deleted`, leaves `col_node == nullptr`, and that null pointer is unconditionally passed to `dag.addFunction(equals_func, {col_node, zero_node}, ...)`.

(c) **Fix matches root cause?** Yes. The fix targets the exact mechanism in (b): when the original FINAL reading step's output header is missing `is_deleted`, clone the prewhere/row-level filter info and re-expose `is_deleted` as an explicit DAG output (so `updateHeader` no longer erases it). To keep header parity with the FINAL branch in the subsequent `UnionStep`, set `is_deleted_added = true` so `addIsDeletedFilter` drops the column again from its output. The defensive `LOGICAL_ERROR` in `addIsDeletedFilter` is fail-close — it surfaces any future regression in the upstream invariant rather than turning into UB.

(d) **Test intent preserved? / New tests added?** New regression test `04207_lazy_final_prewhere_is_deleted_null_deref.sql` covers compound `PREWHERE`, `EXPLAIN`, `PREWHERE`+`WHERE`, and the optimization-disabled sanity case. The four existing lazy-FINAL tests (`04092`, `04093`, `04094`, `04095`, `04105`) continue to pass byte-for-byte — I diffed each against its `.reference`.

(e) **Both directions demonstrated?** Yes.
- Without the fix: `EXPLAIN actions=1 SELECT count() FROM t FINAL PREWHERE is_deleted = 0 AND category = 'target' SETTINGS query_plan_optimize_lazy_final = 1, ...` aborts the local debug server with `SIGSEGV` at `ActionsDAG.cpp:80` (verified by reverting only my src change and re-running the same query against the same data set).
- With the fix: same query returns `400`. The new test passes 50/50 stress runs.

(f) **Fix is general, not a narrow patch?** Considered. The bug class is "downstream code reads a column from the read step's header that prewhere/row-level filter consumed." Two sibling code paths I checked:
- `row_level_filter` — handled symmetrically: cloned and `exposeInputAsDAGOutput` is called on its actions, with the same `is_deleted_added=true` book-keeping.
- The first `createNonIntersectingPlan` call (line 192-193, "all parts non-intersecting" branch) goes through the same function and therefore gets the same fix automatically.
- I did not extend the fix to other lazy-FINAL DAG outputs that aren't `is_deleted`, because `addIsDeletedFilter` is the only consumer that requires a specific column to appear in the read step's output header — the user-projected columns travel through `ActionsDAG::updateHeader`'s "non-DAG-input columns pass through" path and are unaffected.

Session: `cron:clickhouse-ci-task-worker:20260506-061500`.

alexey-milovidov (approved):
This is non-trivial, but looks ok.

clickgapai:
Hi @groeneai @alexey-milovidov — the changelog category for this PR might need a look.

**Current:** `Bug Fix (user-visible misbehavior in an official stable release)`

**Why:** The bug is a SIGSEGV crash (confirmed by reproduction), not just 'misbehavior'. Critical Bug Fix is the correct category for crashes per ClickHouse changelog guidelines.

clickgapai:
Hi — could this be a backport candidate for `26.4`? No backport label was found.

**Affected code:** `createNonIntersectingPlan` in optimizeLazyFinal.cpp — null pointer passed to `ActionsDAG::addFunction`.

**Why this applies to 26.4:** The lazy FINAL optimization (`query_plan_optimize_lazy_final`) was introduced in version 26.4 per SettingsChangesHistory.cpp. The crash affects users who enable this optimization and use PREWHERE with `is_deleted` column.

**Why older branches are unaffected:** Branches 26.3, 26.2, and 25.8 predate this feature and don't contain the affected code path.

If this assessment is correct, consider adding `v26.4-must-backport`.

clickgapai:
Hi @groeneai @alexey-milovidov — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: PR description is empty - no explanation of the bug, reproduction steps, or fix approach
- Suggested addition:

  > ### Problem
  > When a query specifies `PREWHERE` with an expression that references the `is_deleted` column (e.g., `PREWHERE is_deleted = 0 AND category = 'X'`) on a `ReplacingMergeTree` table with the `is_deleted` column, the lazy FINAL optimization would crash with a null pointer dereference.
  > 
  > The prewhere actions DAG consumes `is_deleted` as an input but does not expose it as an output. After lazy FINAL builds a non-FINAL reading step, `removeUnusedColumns` no longer treats `is_deleted` as required for merging, so the column is erased from the read step's output by `ActionsDAG::updateHeader`. The downstream `addIsDeletedFilter` step then passed a null `Node *` to `ActionsDAG::addFunction`.
  > 
  > ### Fix
  > The fix ensures `is_deleted` column is exposed as DAG output when consumed by prewhere or row-level filters by:
  > 1. Adding `exposeInputAsDAGOutput` helper to add input columns back to outputs
  > 2. Cloning and mutating prewhere/row_level_filter DAGs to re-expose `is_deleted`
  > 3. Adding a safety check that throws LOGICAL_ERROR if the invariant is violated

PR #103532, opened 2026-04-25
Fix flaky 00825_protobuf_format_map under map-bucket serialization

groeneai:
## Summary

Pin `map_serialization_version` and `map_serialization_version_for_zero_level_parts` to `basic` in the test's `CREATE TABLE SETTINGS` so the `SELECT *` key order and the Protobuf binary hexdump remain deterministic regardless of CI randomization.

## Root cause

CI randomizes both settings between `basic` and `with_buckets` via `tests/clickhouse-test` `MergeTreeSettingsRandomizer` (lines 1574-1575). With `with_buckets`, Map entries are stored in hash-bucket order on disk, so reads return keys in bucket order rather than insertion order, and the Protobuf output bytes change accordingly. The `.reference` file pins both the `SELECT` output and the hexdump, so the test fails whenever the randomizer picks `with_buckets` for either setting.

The test verifies the Protobuf round-trip for the `Map` data type. The Protobuf encoder/decoder logic is orthogonal to the on-disk storage layout, so pinning only the Map serialization format preserves the test's intent.

Deterministic reproducer (current master, no fix):
```
CREATE TABLE t (a Map(String, UInt32)) ENGINE = MergeTree ORDER BY tuple()
  SETTINGS map_serialization_version_for_zero_level_parts = 'with_buckets',
           max_buckets_in_map = 26,
           map_buckets_strategy = 'constant',
           map_buckets_min_avg_size = 1;
INSERT INTO t VALUES ({'x':5, 'y':7});
SELECT * FROM t;
-- returns {'y':7,'x':5} (bucket order) instead of {'x':5,'y':7} (insertion order)
```

Same pattern as the previous fix for `01872_functions_to_subcolumns_analyzer` (PR #103421, merged 2026-04-23) and `02974_analyzer_array_join_subcolumn.sql` (commit `d238b482d5c`).

## Verification

50/50 passes locally with `--test-runs 50` and full randomization enabled.

## CI report

https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100146&sha=6ddab321885e13563c3b3881cf4f2fe81a093a23&name_0=PR&name_1=Stateless%20tests%20%28amd_asan_ubsan%2C%20distributed%20plan%2C%20parallel%2C%202%2F2%29

30-day CIDB evidence: 6 distinct PRs hit this failure plus 5 master hits — genuine chronic flaky test, all on configurations where the randomizer picked `with_buckets` for one of the two map serialization settings.

### Changelog category (leave one):
[...]

PR #102874, opened 2026-04-16
Fix flaky test 02169_map_functions

groeneai:
### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
- Not required

---

## Summary

Fix flaky test `02169_map_functions` which fails intermittently when CI randomizes `map_serialization_version_for_zero_level_parts` to `with_buckets`.

**Root cause:** Bucketed map serialization stores Map keys in hash-bucket order instead of insertion order. The test’s reference file expected insertion order for Map values read from a MergeTree table, causing output mismatches like:
- Expected: `{'key3':100,'key2':101,'key4':102}`
- Got: `{'key3':100,'key4':102,'key2':101}`

**Impact:** 5+ unrelated PRs affected over 30 days, plus first master failure on April 16, 2026 (`3ec236686a88f`).

**Fix:** Wrap Map outputs read (directly or indirectly) from `table_map` with `mapSort()` so the result is stable regardless of the on-disk serialization format chosen by CI randomization:
- `SELECT mapSort(mapFilter(..., col)) FROM table_map ...`
- `SELECT mapSort(col), mapSort(mapFilter(..., col)) FROM table_map ...`
- `SELECT mapSort(mapApply(..., col)) FROM table_map ...`
- `SELECT mapSort(mapConcat(col, ...)) FROM table_map ...`

In-memory literal maps, boolean results from `mapExists`/`mapAll`, and the existing `mapSort(col)` / `mapSort((k, v) -> v, col)` / `mapPartialSort(..., col)` calls are already order-stable and were left untouched. The reference file was updated accordingly.

This keeps the bucketed serialization path exercised when CI picks it — the test still flows data through `with_buckets`, we just normalize the output for comparison. Same approach as the sibling fix #102468 (`01475_read_subcolumns_storages`).

**Verification:** 50/50 passes with full CI randomization enabled locally (`tests/clickhouse-test --test-runs 50 02169_map_functions`).

### Version info
- Merged into: `26.4.1.1053`

PR #104184, opened 2026-05-06
Fix flaky 04060_coalescing_merge_tree_array_tuple_with_column_param

groeneai:
### Description

Wrap `map_col` in `mapSort` in the two SELECTs of `04060_coalescing_merge_tree_array_tuple_with_column_param.sql` so its output is stable regardless of the randomized `map_serialization_version_for_zero_level_parts` MergeTree setting. CI's `--diagnose-random-settings` flagged the test 229/229 failing under the culprit set:

```
--map_serialization_version_for_zero_level_parts with_buckets
--max_buckets_in_map 37
--map_buckets_strategy constant
--map_buckets_min_avg_size 1
```

…with 0/346 failures when randomization is disabled. With `with_buckets`, `Map` keys are persisted in hash-bucket order, so the test's literal reference (`{'x':1,'y':2}`, insertion order) no longer matches what storage returns (`{'y':2,'x':1}`).

This is the same root cause and same fix pattern already merged in PR #103421 (`01872_functions_to_subcolumns_analyzer`) and PR #102874 (`02169_map_functions`), and proposed in PR #103532 (`00825_protobuf_format_map`). `mapSort` only reorders the SELECT projection — the underlying CoalescingMergeTree column behavior the test verifies (array, tuple, map, and `update_column` interaction with `FINAL` and `OPTIMIZE … FINAL`) is preserved.

Local verification with the fix applied: 200/200 passes under full CI randomization on this build (Debug, master HEAD `5a5c4d66086`). Without the fix, the manual reproduction with the four culprit settings deterministically returns the wrong key order; with the fix it returns `{'x':1,'y':2}`.

CI report that surfaced the failure: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100752&sha=d42192932164a7878cc41513a8deca263e3e73f1&name_0=PR&name_1=Stateless%20tests%20%28arm_binary%2C%20parallel%29
Originating PR (unrelated, "Snappy HTTP/file compression"): https://github.com/ClickHouse/ClickHouse/pull/100752

- [x] Documentation entry is not required.

(none — CI fix)

### DIFF
--- a/tests/queries/0_stateless/04060_coalescing_merge_tree_array_tuple_with_column_param.sql
+++ b/tests/queries/0_stateless/04060_coalescing_merge_tree_array_tuple_with_column_param.sql
@@ -18,12 +18,15 @@ VALUES ('k', 'first', ['a', 'b'], ('hello', 42), {'x': 1, 'y': 2});
 INSERT INTO 04060_test (key, update_column)
 VALUES ('k', 'v2');
 
-SELECT str_col, arr_col, tuple_col, map_col, update_column
+-- Wrap `map_col` in `mapSort` so the output is deterministic regardless of
+-- the on-disk map serialization version (`map_serialization_version_for_zero_level_parts`),
+-- which CI randomizes (`with_buckets` reorders keys by hash bucket).
+SELECT str_col, arr_col, tuple_col, mapSort(map_col), update_column
 FROM 04060_test FINAL;
 
 OPTIMIZE TABLE 04060_test FINAL;
 
-SELECT str_col, arr_col, tuple_col, map_col, update_column
+SELECT str_col, arr_col, tuple_col, mapSort(map_col), update_column
 FROM 04060_test;
 
 DROP TABLE 04060_test;


groeneai:
**Pre-PR validation gate** (per internal worker checklist; session `cron:clickhouse-ci-task-worker:20260506-071500`):

**a) Deterministic repro?** Yes. With these four MergeTree settings injected into `CREATE TABLE`:
```
map_serialization_version_for_zero_level_parts = 'with_buckets',
max_buckets_in_map = 37,
map_buckets_min_avg_size = 1
```
…the SELECT consistently returns `{'y':2,'x':1}` instead of the reference's `{'x':1,'y':2}`. CI's own `--diagnose-random-settings` reported `Runs: 229, Failed: 229, Passed: 0` under exactly that minimised set.

**b) Root cause explained?** Yes. With `map_serialization_version_for_zero_level_parts='with_buckets'` and the bucket parameters above, `Map` data on the zero-level part is laid out by hash bucket. On read, keys come back in bucket order — for `('x', 'y')` with `max_buckets_in_map=37` and the constant strategy, that order is `('y', 'x')`. The test's literal `.reference` was written against insertion order (which happens to be alphabetical), so the comparison fails. Same root cause as PR #103421 (`01872`), PR #102874 (`02169`), and PR #103532 (`00825`).

**c) Fix matches root cause?** Yes. `mapSort(map_col)` canonicalises the projection to the order the reference expects, independent of on-disk layout. It is the same pattern applied in the three sister PRs above.

**d) Test intent preserved?** Yes. The test verifies CoalescingMergeTree behavior (`array`, `tuple`, `map`, `update_column`) under `FINAL` and after `OPTIMIZE … FINAL`. `mapSort` only canonicalises the SELECT projection — the underlying CoalescingMergeTree column-coalescing semantics that the test exercises are unchanged. A regression in how CoalescingMergeTree merges `Map` *values* would still surface as a content diff after `mapSort`.

- Without fix + culprit settings: `{'y':2,'x':1}` ≠ reference `{'x':1,'y':2}` → fails (CI's own 229/229).
- With fix + culprit settings: `{'x':1,'y':2}` == reference → passes.
- With fix + full CI randomization: 200/200 passes locally on Debug build (master HEAD `5a5c4d66086`).

**f) Fix general?** N/A — this is a test-determinism fix, not a code-bug fix. The `with_buckets` storage layout is the intended new format.

groeneai:
cc @Avogar — could you review this? Test-only fix wrapping `map_col` in `mapSort` to make the SELECT output stable under randomized `map_serialization_version_for_zero_level_parts='with_buckets'` (same pattern as PR #103421 / PR #102874 / PR #103532).

Algunenano (approved):

PR #104214, opened 2026-05-06
Add NullCount statistics support for part pruning

zoomxi:
Follow up https://github.com/ClickHouse/ClickHouse/pull/102356.

- New Feature

NullCount column statistics can now be used for part pruning. Queries with IS NULL or IS NOT NULL predicates on Nullable columns that have NullCount statistics can skip entire data parts where the statistics prove no matching rows exist. 

### Version info
- Merged into: `26.5.1.658`


zoomxi:
@hanfei1991 CI is green now. Could you please take some time to review this PR when you're free? Thanks!

src/Storages/MergeTree/KeyCondition.h
```diff
@@ -31,12 +31,22 @@ class MergeTreeSetIndex;
...
+    /// When `null_subcolumns_to_normalize` is non-null and non-empty, INPUT nodes whose name
+    /// appears in the set are rewritten from a bare UInt8 boolean into an explicit comparison
+    /// with zero (`equals`/`notEquals`) within predicate context. This lets `KeyCondition`
+    /// treat the `.null` subcolumn as a ranged key column. The caller is responsible for
+    /// ensuring every entry is a genuine `.null` subcolumn of a Nullable column (not a user
+    /// column that happens to be named like one).
+    explicit ActionsDAGWithInversionPushDown(
+        const ActionsDAG::Node * predicate_,
+        const ContextPtr & context,
+        const NameSet * null_subcolumns_to_normalize = nullptr);
```
hanfei1991, line 49: Can we use "std::option" for safety?

src/Storages/Statistics/StatisticsPartPruner.h
```diff
@@ -35,9 +35,16 @@ class StatisticsPartPruner
     /// Cache key_condition by column names to avoid recreating them for each part.
     std::unordered_map<Names, std::unique_ptr<KeyCondition>, NamesHash> key_condition_cache;
```
hanfei1991, line 36: Why do we use `unique_ptr` rather than shared_ptr here?

tests/queries/0_stateless/04142_statistics_null_count_part_pruning.sql
```diff
@@ -0,0 +1,100 @@
...
+-- Part 0: all NULL values.
+INSERT INTO test_nullcount_pruning VALUES (0, 0, NULL, NULL, NULL, NULL), (0, 1, NULL, NULL, NULL, NULL);
+-- Part 1: no NULL values, below the range predicate.
+INSERT INTO test_nullcount_pruning VALUES (1, 2, 100, 100, 'a', 100), (1, 3, 101, 101, 'b', 101);
+-- Part 2: no NULL values, inside the range predicate.
+INSERT INTO test_nullcount_pruning VALUES (2, 4, 200, 200, 'c', 200), (2, 5, 201, 201, 'd', 201);
+-- Part 3: mixed NULL and non-NULL values, inside the range predicate.
+INSERT INTO test_nullcount_pruning VALUES (3, 6, NULL, NULL, NULL, NULL), (3, 7, 160, 160, 'e', 160), (3, 8, NULL, NULL, NULL, NULL), (3, 9, 161, 161, 'f', 161);
+
+ALTER TABLE test_nullcount_pruning ADD STATISTICS value TYPE nullcount;
```
hanfei1991: We can declare these statistics in `CREATE TABLE`

src/Storages/MergeTree/KeyCondition.cpp
```diff
@@ -937,7 +939,9 @@ static const ActionsDAG::Node & cloneDAGWithInversionPushDown(
     ActionsDAG & inverted_dag,
     std::unordered_map<const ActionsDAG::Node *, const ActionsDAG::Node *> & inputs_mapping,
     const ContextPtr & context,
-    const bool need_inversion)
+    const bool need_inversion,
+    const NameSet * null_subcolumns_to_normalize,
+    bool is_predicate_context)
```
hanfei1991, line 944: Do we actually need `is_predicate_context`?
For example, when it is false, we can pass 'nullptr' to `null_subcolumns_to_normalize`. And if it is true, we do nothing.

tests/queries/0_stateless/04142_statistics_null_count_part_pruning.sql
zoomxi: ok,fixed.

src/Storages/MergeTree/KeyCondition.h
zoomxi: It is only an optional non-owning input for DAG cloning, and it does not escape this stage. Using `std::optional` would make the interface heavier without adding much here.

src/Storages/Statistics/StatisticsPartPruner.h
zoomxi: Because the cache is the only owner here. Callers only borrow the pointer temporarily, so there is no shared lifetime to model.

src/Storages/MergeTree/KeyCondition.cpp
zoomxi: I kept `is_predicate_context` because predicate context can be re-entered recursively under logical nodes like `and` / `or`. Passing `nullptr` when it is false is not equivalent, since deeper predicate subtrees would also lose access to the normalization set.

zoomxi:
[AST fuzzer (amd_debug)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104214&sha=dc6a5ff130089b480aeab16665904d4603952de5&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%29) #104459

zoomxi:
[Stateless tests (amd_debug, parallel)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104214&sha=b539b56cc3ee0d77e9ebcd8cbc8785cdfd0eb45e&name_0=PR&name_1=Stateless%20tests%20%28amd_debug%2C%20parallel%29) #104912

hanfei1991 (approved):

PR #104216, opened 2026-05-06
Avoid TopK dynamic filter regression on variable-length sort columns

alexey-milovidov:
PR #99537 enables `use_top_k_dynamic_filtering` by default. For sort columns of fixed-length types this is a clear win, but for variable-length types (`String`, `Array`, `Map`, `Tuple` containing such) the per-row `__topKFilter` prewhere can outweigh its savings — most notably when the column's lex-min value dominates and few granules can be skipped. In that case the dynamic filter degrades query latency rather than improving it.

This PR adds a new setting `use_top_k_dynamic_filtering_for_variable_length_types` (off by default) that gates dynamic filtering for variable-length types. When `use_top_k_dynamic_filtering = 1`, dynamic filtering is now restricted to columns whose values have a fixed maximum size in memory — `IDataType::haveMaximumSizeOfValue` returns true. That predicate forwards through `Nullable` and `Tuple`, so `Nullable` of a fixed-length type and `Tuple`s of fixed-length types remain eligible. Setting `use_top_k_dynamic_filtering_for_variable_length_types = 1` restores the previous behavior.

This addresses the perf regressions observed in PR #99537's CI:

CI report:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=99537&sha=9d86b8cf27ca90c5223f949656f0141685ea3f6f&name_0=PR&name_1=Performance%20Comparison%20%28amd_release%2C%20master_head%2C%203%2F6%29

The largest regression in batch 3/6 amd_release was `string_sort #10`:

```
SELECT MobilePhoneModel FROM hits_100m_single ORDER BY MobilePhoneModel LIMIT 300
```

Reproduced locally on `hits_100m_single` (94/100M `MobilePhoneModel` values are empty):

| Setting | Time | vs master |
|---|---|---|
| master default (both off) | 0.013-0.015s | baseline |
| PR #99537 default (both on) | 0.022-0.026s | **1.47-1.63x slower** |
| With this PR's gate off (default) | 0.017s | back to baseline |
| With this PR's gate on (opt-in) | 0.026s | matches PR #99537 |

`SelectedRows` is identical (99,997,497) with and without the dynamic filter — it skips zero granules but still pays a per-block `lessOrEquals(String, String)` cost (`UserTime` ~1.8x).

- Performance Improvement

Restrict `use_top_k_dynamic_filtering` to fixed-length sort columns by default, avoiding regressions on `ORDER BY <var-length-column> LIMIT N` queries where the per-row threshold comparison cost exceeds the I/O savings. The previous behavior is available via the new `use_top_k_dynamic_filtering_for_variable_length_types` setting.

The new setting is documented in `src/Core/Settings.cpp`; no separate docs page is required for a session-level setting.

### Version info
- Merged into: `26.5.1.397`


shankar-iyer (approved):

src/Core/Settings.cpp
```diff
@@ -1742,6 +1742,18 @@ When enabled, the query executor will try to skip granules and rows that will no
...
 
+- 0 — Disabled.
+- 1 — Enabled.
+)", 0) \
+    DECLARE(Bool, use_top_k_dynamic_filtering_for_variable_length_types, false, R"(
+Allow `use_top_k_dynamic_filtering` to apply when the sort column has a variable-length data type (e.g. `String`, `Array`, `Map`, `Tuple` containing variable-length elements).
+
+For such types, the per-row threshold comparison performed by the dynamic filter can outweigh its savings when the column's lexicographic minimum dominates (e.g. mostly empty strings) and few granules can be skipped. In that case the dynamic filter degrades query latency rather than improving it.
+
+When this setting is `0`, dynamic filtering is restricted to columns whose values have a fixed maximum size in memory (numbers, `Date`, `DateTime`, `FixedString`, `Enum`, `Nullable` of such types, `Tuple` of such types). When set to `1`, dynamic filtering applies to variable-length types as well, matching the previous behavior.
```
rschu1ze: The reader is confused what "previous behavior" means in this context (`use_top_k_dynamic_filtering` has never been default enabled). Let's remove that four words.

EmeraldShift:
Will this be backported to 26.3 LTS or 26.4 to recover the TopK optimization without regression?

alexey-milovidov:
The optimization was opt-in in previous versions, exactly for this reason - it does not always improve. There is no reason to backport.

EmeraldShift:
Ah, my bad. I assumed the referenced PR was already merged, and it isn't. Agreed there's nothing to backport :)

Issue #104183, opened 2026-05-06
[CI crash] Segfault in AggregateFunctionCrossTab<CramersVData>::serializeBatch on TCP send path


alexey-milovidov:
### Overview

Segfault during `AggregateFunctionCrossTab<CramersVData>::serializeBatch` while a hashtable inside `CramersVData` is being serialized via `NativeWriter::write` (TCP server send-data path). The crash occurs at address `0xa` — a small offset from null — during `HashTable::write` over the hashmap that backs `CramersVData`.

**CI report**: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104153&sha=2872b8aca11b508a6e4081154954918ae59df4b4&name_0=PR&name_1=Stress%20test%20%28arm_release%29

**Server error log**: https://s3.amazonaws.com/clickhouse-test-reports/PRs/104153/2872b8aca11b508a6e4081154954918ae59df4b4/stress_test_arm_release/clickhouse-server.err.log

### Stack trace

```
HashTable<long, HashMapCellWithSavedHash<long, unsigned long, HashCRC32<long>, HashTableNoState>,
         HashCRC32<long>, HashTableGrower<4ul>,
         AllocatorWithStackMemory<Allocator<true, true>, 384ul, 1ul>>::write(WriteBuffer&) const
DB::IAggregateFunctionHelper<DB::AggregateFunctionCrossTab<DB::(anonymous namespace)::CramersVData>>::serializeBatch
DB::ISerialization::serializeBinaryBulkWithMultipleStreams
DB::SerializationVariant::serializeBinaryBulkWithMultipleStreamsAndUpdateVariantStatistics
DB::SerializationVariant::serializeBinaryBulkWithMultipleStreams
DB::NativeWriter::writeData
DB::NativeWriter::write
DB::TCPHandler::sendData
DB::TCPHandler::runImpl
DB::TCPHandler::run
Poco::Net::TCPServerConnection::start
Poco::Net::TCPServerDispatcher::run
Poco::PooledThread::run
Poco::ThreadImpl::runnableEntry
```

Triggering query: `SELECT * FROM test_merge ORDER BY a;` from test `03302_merge_table_structure_unification`. The test itself does not use `cramersV` or any aggregate function, so the result column being serialized comes from a concurrent stress-test query (likely a `Variant` column referencing a `cramersV` aggregate state).

### Analysis

[...]

PR #99736, opened 2026-03-17
Optimize AND comparison chains: detect conflicts, prune redundancies

wudidapaopao:
Resolved https://github.com/ClickHouse/ClickHouse/issues/90081

### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Optimize AND chains with multiple comparison conditions on the same expression: detect contradictions (e.g., `a < 3 AND a > 5` → `false`), prune redundant conditions (e.g., `a = 3 AND a < 5` → `a = 3`).

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Summary
Extend `LogicalExpressionOptimizerPass` to analyze and simplify AND chains of comparison conditions on the same expression (e.g., `a = 3 AND a < 5 AND a != 7`). The optimizer now detects contradictions (rewrites to `false`), prunes redundant conditions, and handles cross-type constant comparisons via `convertFieldToTypeStrict`.

### Optimizations and fixes
| Category | Example | Before | After |
|---|---|---|---|
| equals + range prune | `i = 3 AND i < 5` |  kept | `i = 3` |
| equals + range conflict | `i = 3 AND i < 2` | kept | `false` |
| equals + notEquals prune | `i = 3 AND i != 5` | kept | `i = 3` |
| equals + notEquals conflict | `i = 3 AND i != 3` | kept | `false` |
| notEquals implied by range | `i != 3 AND i < 3` | kept | `i < 3` |
| range tightening | `i < 5 AND i < 3` | kept | `i < 3` |
| empty range detection | `i < 3 AND i >= 3` | kept | `false` |
| cross-type lossless | `i = 3 AND i = 3.0` | false | `i = 3` |
| transitive conflict | `x = 3 AND x = y AND y = 5` | kept | `false` |
| transitive redundant chain | `x < 3 AND y > 3 AND y < 10` | redundant x < 10 added | no redundant inference |

PR #93384, opened 2026-01-04
Optimize `cramersV`, `cramersVBiasCorrected`, `theilsU`, and `contingency` when used with window functions

nihalzp:
```sql
SELECT cramersV(number, number) OVER (ORDER BY number) FROM numbers(20000) FORMAT NULL;
```
Master: 4.198 sec
PR: 0.014 sec

These functions were running in quadratic time complexity when used in window context. This PR adds `window` variants of these aggregate functions. When a query uses these aggregates in a window context, `resolveFunction` chooses the window variants creators.

We do not use these `Window` versions for non-window context such as GROUP BY because they will be slower. `Window` versions are optimized for incremental updates so that `getResult` is `O(1)`.

image

### Changelog category (leave one):
- Performance Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Optimize `cramersV`, `cramersVBiasCorrected`, `theilsU`, and `contingency` when used with window functions. Closes https://github.com/ClickHouse/ClickHouse/issues/83521.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.27`

PR #104228, opened 2026-05-06
find_symbols.h: implement AArch64/NEON SIMD path

alexey-milovidov:
On AArch64 the SIMD search loops in `base/base/find_symbols.h` previously fell through to the scalar byte-by-byte fallback (only `__SSE2__` and `__SSE4_2__` were handled). This patch adds NEON code paths to all four loops (`find_first_symbols_sse2` and `find_last_symbols_sse2`, both compile-time-template and runtime-needle variants), so functions built on top — `find_first_symbols`, `find_first_not_symbols`, `find_last_symbols_or_null`, `find_last_not_symbols_or_null`, `splitInto`, etc. — get vectorised on ARM.

The NEON helpers (`neon_is_in`) mirror the SSE2 ones (`mm_is_in`): each needle is broadcast with `vdupq_n_u8`, compared with `vceqq_u8`, and OR-folded with `vorrq_u8`. There is no direct equivalent of `pmovmskb` on AArch64, so the 16-byte all-0/all-1 result is compressed with the standard `vshrn_n_u16(..., 4)` trick into a 64-bit value where each input byte occupies a 4-bit nibble; the first/last matching byte is then recovered as `__builtin_ctzll(mask) >> 2` / `__builtin_clzll(mask) >> 2`. Negative search inverts the equality vector with `vmvnq_u8` before compression. The same trick is already used in `contrib/simdjson`, `contrib/vectorscan`, `contrib/SimSIMD`, and `contrib/StringZilla`.

`__SSE4_2__` is x86-only, so on AArch64 the dispatch falls through to the `_sse2`-named variants for any number of symbols 1–16, and the `_mm_cmpestr*` path is not duplicated.

Verified with a standalone harness that mirrors `gtest_find_symbols` (including the `NullCharacter` case) and runs 5000 random fuzz iterations comparing against a scalar reference, plus long-haystack and edge-position cases for both compile-time and runtime needles.

### Microbenchmark (function-level, 64MB buffer, lowercase ASCII)

| Search | Scalar (master) | NEON (this PR) | Speedup |
| --- | --- | --- | --- |
| `find_first_symbols<'\t'>` | 2.78 GB/s | 26.19 GB/s | 9.4x |
| `find_first_symbols<4 chars>` | 1.45 GB/s | 15.09 GB/s | 10.4x |
| `find_first_symbols<8 chars>` | 1.28 GB/s | 8.98 GB/s | 7.0x |
| `find_first_symbols(runtime, 4 chars)` | 1.07 GB/s | 13.56 GB/s | 12.7x |
| `find_first_symbols(runtime, 8 chars)` | 0.62 GB/s | 7.93 GB/s | 12.8x |

### Real-query performance — ClickBench `hits` shard (1M rows of real web data: URL, Title, SearchPhrase)

5 iterations, average wall time, AWS Graviton (`aarch64`). Both binaries built from the same `26.5.1.1` master source; the only difference is whether `find_symbols.h` has the NEON path.

| Workload                                | Scalar (s) | NEON (s) | Speedup |
| --------------------------------------- | ---------: | -------: | ------: |
| `TSV` `count()`                         |      0.288 |    0.140 |   2.06x |
| `extractURLParameter(URL, 'q')`         |      0.238 |    0.182 |   1.31x |
| `splitByChar('/', URL)`                 |      0.242 |    0.190 |   1.27x |
| `TSV` output (write 1M rows)            |      0.280 |    0.252 |   1.11x |
| `CSV` `count()`                         |      0.196 |    0.190 |   1.03x |
| `JSONEachRow` output (write 1M rows)    |      0.386 |    0.386 |   1.00x |
| `JSONEachRow` `count()`                 |      0.310 |    0.330 |   0.94x |

### Real-query performance — synthetic 10M-row workloads (1-1.5GB on disk)

Same setup; chosen to exercise specific call sites with controlled density.

| Workload                                | Scalar (s) | NEON (s) | Speedup |
| --------------------------------------- | ---------: | -------: | ------: |
| `TSV` `count()`                         |       1.15 |     0.44 |   2.61x |
| `TSV` `sum(numeric_col)`                |       1.56 |     0.85 |   1.84x |
| `LineAsString` `count()`                |       0.59 |     0.29 |   2.03x |
| `splitByChar(',', s)` over 10M strs     |       1.30 |     0.70 |   1.86x |
| `path(url)` over 10M URLs               |       0.88 |     0.48 |   1.84x |
| `queryString(url)` over 10M URLs        |       0.89 |     0.48 |   1.85x |
| `extractURLParameter(url, 'q')`         |       0.88 |     0.48 |   1.85x |
| `JSONEachRow` (sparse strings)          |       0.32 |     0.29 |   1.14x |
| `CSV` `count()` (synthetic, dense delims)|      0.83 |     0.93 |   0.88x |
| `JSONEachRow` `count()` (synthetic, dense)|     1.54 |     2.48 |   0.62x |

### Note on the dense-`JSONEachRow` regression

On the synthetic dataset above the average distance between matches is ~6-8 bytes (every quoted field ends inside a single 16-byte vector chunk). Sampling profiler localises the regression to `DB::JSONUtils::skipRowForJSONEachRowImpl<'{','}'>` (1387 → 2344 samples on a single query): when match cadence is shorter than the vector width, the SIMD body re-scans more bytes than scalar would. This is an inherent vectorisation trade-off and is equally present in the existing SSE2 path on x86. On the real `hits` dataset (longer string fields) the JSON regression shrinks to ~6% — and the wins on TSV / URL functions / `splitByChar` / `LineAsString` are 1.3-2.6x.

- Performance Improvement

Vectorise `find_first_symbols`, `find_first_not_symbols`, `find_last_symbols_or_null`, `find_last_not_symbols_or_null`, and `splitInto` on AArch64 using NEON. Previously these helpers had a SIMD path only on x86 (SSE2 / SSE4.2) and fell through to a scalar loop on ARM. Speeds up TSV parsing by ~2x, URL functions and `splitByChar` by ~1.3x on the ClickBench `hits` dataset; very dense JSON parsing (sub-16-byte field cadence) regresses slightly in line with the existing SSE2 trade-off.

### Version info
- Merged into: `26.5.1.660`


Algunenano (approved):
Looks good. We could We overload `maybe_negate` to handle `uint8x16_t` and keep the same code structure but that's minor.

alexey-milovidov:
@groeneai, fix `Logical error: 'phi_squared <= max_phi_squared + 1e-6' (STID: 2508-2e40)` and send a separate PR with the fix.

groeneai:
@alexey-milovidov — STID 2508-2e40 (`phi_squared <= max_phi_squared + 1e-6`) is the chronic CrossTab/CramersV state-corruption family already tracked under issue [#104183](https://github.com/ClickHouse/ClickHouse/issues/104183).

We posted a comprehensive analysis there earlier today (2026-05-06): https://github.com/ClickHouse/ClickHouse/issues/104183#issuecomment-4386959227

TL;DR:
- 5 distinct stack signatures observed across 13 days, all dating from after PR #93384 (2026-04-23)
- The phi_squared assertion is one of those signatures; sibling signatures hit `HashTable::write` and `HashTable::begin` from `CrossTabCountsState::serialize` and `CrossTabCountsState::merge`
- Root cause hypothesis: a missing `getStateVariant` / `canMergeStateFromDifferentVariant` combinator override, or a `sizeOfData` mismatch between `CrossTabAggregateData` (GROUP BY state) and `CrossTabPhiSquaredWindowData` (window state) — leading to placement-new in a wrong-sized buffer and a torn HashMap that later yields a `0xa` faulting address or an out-of-range phi_squared
- Cannot reproduce locally without the specific concurrent stress conditions; @nihalzp (PR #93384 author) was pinged for a combinator audit

Opening a separate PR purely for the phi_squared manifestation would splinter what is one underlying bug. Will file the fix under #104183 once the offending combinator is pinpointed. Happy to upstream debug-build invariants in `CrossTabCountsState::serialize`/`merge` as a stand-alone diagnostic PR if useful in the meantime.

— @groeneai (worker session `cron:clickhouse-ci-task-worker:20260506-184500`)

base/base/find_symbols.h
```diff
@@ -201,6 +289,21 @@ inline const char * find_first_symbols_sse2(const char * const begin, const char
         if (bit_mask)
             return pos + __builtin_ctz(bit_mask);
     }
+#elif defined(__aarch64__)
+    if (pos + 15 < end)
```
Algunenano, line 344: Does it make sense  to apply this in x86 and avoid preparing needles unnecessarily there too?

alexey-milovidov:
@groeneai, investigate the chronic `Possible deadlock on shutdown (see gdb.log)` failure on `Stress test (arm_debug)`: https://s3.amazonaws.com/clickhouse-test-reports/PRs/104228/726ffb1642471e8ebfe17e1e802ddc1e09490cd2/stress_test_arm_debug/gdb.log and provide a fix in a separate PR. This fails 8-10 times per day on master ([CIDB](https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICdQb3NzaWJsZSBkZWFkbG9jayBvbiBzaHV0ZG93biAoc2VlIGdkYi5sb2cpJwogICAgLS0gQU5EIGNoZWNrX25hbWUgPSAnU3RyZXNzIHRlc3QgKGFybV9kZWJ1ZyknCiAgICBBTkQgdGVzdF9zdGF0dXMgSU4gKCdGQUlMJywgJ0VSUk9SJykKICAgIEFORCAocHVsbF9yZXF1ZXN0X251bWJlciA9IDAgT1IgYmFzZV9yZWYgSU4gKCdtYXN0ZXInKSkKR1JPVVAgQlkgZGF5Ck9SREVSIEJZIGRheSBERVNDCg==)) and is unrelated to this PR (the stuck threads are in `absl::Mutex` shutdown, no `find_symbols` frames). If a fix is already in progress, link it here.

groeneai:
@alexey-milovidov investigation results — gdb.log analysis of `PRs/104228/726ffb1.../stress_test_arm_debug/gdb.log`.

## Root cause: leaked local `ThreadPool` instances at server shutdown

Decoded the gdb.log (725 threads, 7.2MB). The deadlock is **not** in `absl::Mutex` shutdown itself — those frames are a side effect. The real blocker is in **`GlobalThreadPool::shutdown()`** which can't join its own `std::thread`s because each is running a leaked local pool's `worker()` loop.

### Main thread (LWP 499292)
```
#0  __futex_abstimed_wait_common64
#3  __pthread_clockjoin_ex                                              ← joining a worker thread
#5  std::thread::join (this=0xff337a7c8ae8)
#6  ThreadPoolImpl<std::thread>::ThreadFromThreadPool::join
#7  ThreadPoolImpl<std::thread>::finalize (this=0xff355d30bc00)        ← GlobalThreadPool::finalize
#8  Server::main::$_15::operator()                       Server.cpp:1482
#9  BasicScopeGuard::invoke
#11 Server::main                                          Server.cpp:3296
```

`Server.cpp:1482` is the last scope guard:
```cpp
SCOPE_EXIT_SAFE({
    LOG_INFO(log, "Waiting for background threads");
    GlobalThreadPool::instance().shutdown();      // <-- stuck here
    LOG_INFO(log, "Background threads finished in {} ms", watch.elapsedMilliseconds());
});
```

### The thread it's waiting on (LWP 500233 "ThreadPool")
```
#0  __futex_abstimed_wait_common64
#5  std::condition_variable::wait
#7  std::condition_variable::wait<...lambda#2>(unique_lock, pred)
#8  ThreadPoolImpl<ThreadFromGlobalPoolImpl<false,true>>::ThreadFromThreadPool::worker
                                                          ThreadPool.cpp:736   ← inner pool's wait loop
#15 ThreadPoolImpl<std::thread>::ThreadFromThreadPool::worker
                                                          ThreadPool.cpp:809   ← outer std::thread running the job
```

The std::thread (frame #15, owned by GlobalThreadPool) is running a **job** which is "be the inner pool's worker forever" (frame #8 at line 736). The inner pool's `parent_pool.new_job_or_shutdown.wait(...)` only exits when `parent_pool.shutdown` is set. `GlobalThreadPool::shutdown()` sets its OWN `shutdown`, not the inner pool's — those are independent.

### Scale of the leak (this run)
56 worker threads stuck at `ThreadPool.cpp:736` across **7 distinct leaked pools**:

| Mutex addr | Stuck workers |
|------------|---------------|
| 0xff34114a6b00 | 16 |
| 0xff34114a6800 | 16 |
| 0xff34114a6c80 | 8 |
| 0xff34114a6980 | 8 |
| 0xff355d312e00 | 3 ← main thread is waiting on a worker of this pool |
| 0xff355d312c80 | 3 |
| 0xff355d312f80 | 2 |

All threads are named "ThreadPool" (the default — `setThreadName` runs inside the worker loop AFTER receiving a job, but these never received one). So the leaked pools can't be identified by thread name alone.

### Why this matches the chronic 8-10/day rate
Local `ThreadPool` (= `ThreadPoolImpl<ThreadFromGlobalPool>`) instances must be destroyed (which calls `finalize()` → sets shutdown → notifies) **before** `GlobalThreadPool::shutdown()` runs. Some subsystem(s) hold a reference to a local `ThreadPool` past the scope guard at `Server.cpp:1482`. Likely candidates: long-lived singletons, registries, or thread-local resources that aren't drained in `Server::main` cleanup. Different stress runs hit different leakers, which is consistent with this being an umbrella affecting multiple subsystems (8-10/day spread across several specific patterns).

### The `absl::Mutex` frames you saw
gRPC has its own `WorkStealingThreadPool`. 17 threads in this gdb.log are stuck in gRPC's `WorkSignal::WaitWithTimeout` (which uses `absl::Mutex`). They're not the blocker — they're just idle workers of gRPC's pool that haven't been told to exit because gRPC's `EventEngine` global is presumably one of the things that doesn't clean up before the `Server::main` scope guard. They're a *symptom* of the same broader pattern (subsystem lifetimes not properly ordered against `GlobalThreadPool` shutdown), not an independent deadlock.

## Direction for fix

Two angles, complementary:

1. **Targeted (per-subsystem):** Identify the leaked local pools. The most practical way is to add a `LOG_INFO` listing `name, num_threads, scheduled_jobs` for every `ThreadPoolImpl` still alive *immediately before* the `GlobalThreadPool::shutdown()` call (or in `GlobalThreadPool::finalize` itself, before notify_all). When chronic hits land in CI, the log will name the leaker(s). Possibly enable behind a debug flag to avoid prod overhead. Then fix the lifetime in each named owner.

2. **Structural:** A global registry of live `ThreadPoolImpl<ThreadFromGlobalPool>` instances; `GlobalThreadPool::shutdown()` walks the registry, sets `shutdown=true` and notifies *each registered pool's* condvar before joining its own `std::thread`s. This breaks the dependency on subsystem destruction order. Slight added cost per pool construction/destruction (register/unregister + lock); modest. Downside: papers over real lifetime bugs.

Recommendation: do (1) first to identify the offenders, then (2) as a backstop if the list of leakers is long-tailed.

## Not pushing a fix yet
Per our policy, no fix PR without local reproduction first. Will:
- Build a debug binary with the diagnostic LOG_INFO patch from (1) and run stress locally to confirm at least one specific leaker
- Open a fix PR for whichever subsystem(s) show up
- If the list is long, also open (2) as a backstop

Tracking under task `2026-04-01-ci-hung-check-deadlock-chronic-trunk-is`. CIDB rate (14d): 8-10/day on master with a 78-hit spike on May 11 (concentrated in 2 PRs, possibly a single bad run). Will keep this thread updated.

alexey-milovidov:
@groeneai, investigate the failure: [`01410_nullable_key_and_index` on `Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel)`](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104228&sha=6b3793bd2b377620ca1050c824672dc8c4300a2e&name_0=PR&name_1=Stateless%20tests%20%28amd_llvm_coverage%2C%20ParallelReplicas%2C%20s3%20storage%2C%20parallel%29) and provide a fix in a separate PR. This is unrelated to this PR (the diff is row ordering of a `Nullable` key under ParallelReplicas, no `find_symbols` frames). It is flaky on master ([CIDB](https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICcwMTQxMF9udWxsYWJsZV9rZXlfYW5kX2luZGV4JwogICAgLS0gQU5EIGNoZWNrX25hbWUgPSAnU3RhdGVsZXNzIHRlc3RzIChhbWRfbGx2bV9jb3ZlcmFnZSwgUGFyYWxsZWxSZXBsaWNhcywgczMgc3RvcmFnZSwgcGFyYWxsZWwpJwogICAgQU5EIHRlc3Rfc3RhdHVzIElOICgnRkFJTCcsICdFUlJPUicpCiAgICBBTkQgKHB1bGxfcmVxdWVzdF9udW1iZXIgPSAwIE9SIGJhc2VfcmVmIElOICgnbWFzdGVyJykpCiAgICBBTkQgdGVzdF9jb250ZXh0X3JhdyBMSUtFICclcmVzdWx0IGRpZmZlcnMgd2l0aCByZWZlcmVuY2UlJwpHUk9VUCBCWSBkYXkKT1JERVIgQlkgZGF5IERFU0MK) shows 4 hits in the last 30 days across 4 different PRs). If a fix is already in progress, link it here.

Algunenano (approved):

alexey-milovidov:
@groeneai, investigate the `CH Inc sync` failure: the sync PR [ClickHouse/clickhouse-private#57269](https://github.com/ClickHouse/clickhouse-private/pull/57269) fails on `test_statistics_cache/test_smt.py::test_statistics_cache_smt` ([Integration tests (amd_asan_ubsan, db disk, old analyzer, 5/6)](https://d1k2gkhrlfqv31.cloudfront.net/clickhouse-test-reports-private/json.html?PR=57269&sha=6b3793bd2b377620ca1050c824672dc8c4300a2e&name_0=PR&name_1=Integration%20tests%20%28amd_asan_ubsan%2C%20db%20disk%2C%20old%20analyzer%2C%205%2F6%29)). This is unrelated to this PR — the diff is entirely under `base/base/find_symbols.h` (AArch64/NEON path + SSE2 zero-pad fix) and has no contact with the SharedMergeTree statistics cache. Please look into the chronic failure and provide a fix in a separate PR. If a fix is already in progress, link it here.

alexey-milovidov:
Improvements exactly where we expected:

Screenshot_20260515_014550
Screenshot_20260515_014605
Screenshot_20260515_014625
Screenshot_20260515_014644

groeneai:
> @alexey-milovidov, investigation of the `test_statistics_cache_smt` failure on CH Inc sync PR `clickhouse-private#57269`.

## What we can see on the public side

`test_statistics_cache` (the public MergeTree sibling — `tests/integration/test_statistics_cache/test.py`) has **0 master failures in the last 60 days**. CIDB query:

```sql
SELECT countIf(pull_request_number=0) AS master, countIf(pull_request_number>0) AS prs
FROM default.checks
WHERE test_name LIKE '%test_statistics_cache%'
  AND test_status IN ('FAIL','ERROR')
  AND check_start_time > now() - INTERVAL 60 DAY
-- master=0, prs=1048, distinct_prs=15
```

All 1048 PR failures are concentrated on **2 PRs**, both unmerged:

| PR | Author | Sub-tests failing | Last seen |
|---|---|---|---|
| [#99736](https://github.com/ClickHouse/ClickHouse/pull/99736) `Optimize AND comparison chains` | @wudidapaopao | 6 / 13 (load, session_toggle, interval_zero, alter_interval, countmin_lc, auto_types) | 2026-05-09 |
| [#100185](https://github.com/ClickHouse/ClickHouse/pull/100185) `Add assertions in IColumn::mutate to verify deep unique ownership of sub-columns` | @alexey-milovidov | **13 / 13 (all)** | 2026-05-06 |

Notably your own PR #100185 makes **every** `test_statistics_cache` sub-test fail — the assertions you added fire deep inside the cache's statistics-loading path. PR #100185 touches `ColumnReplicated.{cpp,h}` among many other column types, which is the column used by SMT replication. Likely the same latent issue your assertions are designed to surface (column shared-ownership violation during `IColumn::mutate`) also exists in code reachable from the SMT statistics-load path.

## What we can't see

The failing report URL is `https://d1k2gkhrlfqv31.cloudfront.net/clickhouse-test-reports-private/...` which returns **HTTP 401 (Basic auth required)** — we can't read the private CI artifact. The test file `test_statistics_cache/test_smt.py` is also in the private repo only (public has just `test.py`).

## The shared code path

The `LoadedStatisticsMicroseconds` ProfileEvent that drives `_assert_load` lives in `MergeTreeData::getConditionSelectivityEstimatorByPredicate` (`src/Storages/MergeTree/MergeTreeData.cpp:864`). `MergeTreeData` is the public base for **both** `StorageMergeTree` and `StorageSharedMergeTree`, so the cache (`cached_estimator`, `stats_mutex`, `refreshStatistics`) is shared infrastructure. SMT-only divergence will be in:
- `IMergeTreeDataPart::loadStatistics()` overrides on the SMT-side `DataPartShared`/equivalent
- the part-discovery code that feeds `getDataPartsVectorForInternalUsage` under SMT

Recent **public master** commits touching this path that landed *after* the last green state and could be the sync regressor:
- `78d10eb43cd statistics retry after failure` (@zoomxi, 2026-05-08) — restructures `IMergeTreeDataPart::getEstimates` to build a local `new_estimates` and only move on success; adds `merge_tree_load_statistics_throw` fail point. If the SMT override of `loadStatistics` throws once and the new retry contract isn't honored, the assertion `_loaded_stats_us > 0` will silently fail.
- `801397d3c11` / `65ad40dfe08` / `e9f27104b27` — `null_count` part-pruning + auto-statistics randomization (@zoomxi, PR #104214). Changes how parts are pruned before statistics loading.

## What we need from you

Without the private CI artifact we're guessing. To root-cause this we need any of:

1. The exact failing assertion line + stack trace from `test_statistics_cache_smt` (paste it in the comment).
2. The `LoadedStatisticsMicroseconds` value the failing query reports (0? or non-zero but cached unexpectedly?).
3. Confirmation of whether the sync PR includes commit `78d10eb43cd` and/or `801397d3c11`.
4. Read access to the private report bucket, or a re-publish of the failing test output as a public gist.

If (1) shows the failure is on the same `_assert_load` line and `_loaded_stats_us == 0`, the most likely public-side suspect is the new retry contract in `78d10eb43cd` not propagating through the SMT-side part override. We'd then fix it on master and request a re-sync.

PR #104229, opened 2026-05-06
Fix int32_t overflow in `lowerUTF8`/`upperUTF8` for large buffers

tiandiwonder:
## Linked issues
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1741
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1773

ICU's `ucasemap_utf8ToLower` and `ucasemap_utf8ToUpper` accept `int32_t` for buffer
size parameters and return the required output length as `int32_t` on
`U_BUFFER_OVERFLOW_ERROR`. When `lowerUTF8` / `upperUTF8` process many non-ASCII rows
(e.g. during `MergeTree` text-index builds), the accumulated output buffer can exceed
2 GiB, and the unchecked `static_cast<int32_t>(res_data.size() - curr_offset)` wraps
negative — ICU then rejects the call with `U_ILLEGAL_ARGUMENT_ERROR`.

Two changes in `src/Functions/LowerUpperUTF8Impl.h`:

- **Cap `destCapacity` at `INT32_MAX`.** This is *not* an upper bound on total output:
  `res_data` and `curr_offset` keep growing in `size_t`; the cap only controls what a
  single ICU call sees. The `U_BUFFER_OVERFLOW_ERROR` retry path enlarges `res_data`
  correctly.
- **Reject single rows whose worst-case case-mapped output could exceed `INT32_MAX`.**
  Unicode full case mapping (`SpecialCasing.txt`, e.g. `U+0390 ΐ` → 3 code points / 6
  bytes from a 2-byte input) expands at most 3x in UTF-8 bytes, so the bound is
  `static_cast<int64_t>(src_size) * 3 > INT32_MAX` — equivalently, `src_size > INT32_MAX / 3`
  (~715 MiB). Without this guard, a row in the ~715 MiB to 2 GiB range containing
  3x-expanding code points would overflow `dst_size` in the retry path and corrupt
  `res_data`.

The two changes form a logical closure: the per-row guard keeps the worst-case output
≤ `INT32_MAX`, so the `destCapacity` cap is always ≥ what the current row needs, and
total accumulated output is bounded only by `ColumnString`'s 64-bit offsets — there is
no 2 GiB ceiling on the column.

A new stateless test (`04204_lower_upper_utf8_large_buffer`) covers the ICU code path
across the full ladder of UTF-8 byte-expansion ratios documented in `SpecialCasing.txt`
(1.0x, 1.5x, 2.0x, 3.0x), printing input/output byte length so the 3x worst case the
guard relies on is self-evident from the reference. The test does **not** reproduce
the original ≥ 2 GiB failure end-to-end (stateless-test budget); a failpoint-based
regression target would be a reasonable follow-up.

Related: https://github.com/ClickHouse/ClickHouse/issues/101205

- Improvement

Fix \`lowerUTF8\` / \`upperUTF8\` exception when processing large non-ASCII data sets (e.g. during text-index builds) where the accumulated output buffer would exceed 2 GiB; reject excessively long single rows with a clear error.

### Version info
- Merged into: `26.5.1.544`
- Backported to: `26.4.3.13`, `26.3.11.14`, `26.2.19.26`


alexey-milovidov (approved):

alexey-milovidov:
@groeneai, fix "Logical error: 'max_log_ptr == new_max_log_ptr' (STID: 2508-47ae)" and send a separate PR with the fix. If the fix already exists, link it here.

groeneai:
@alexey-milovidov — STID 2508-47ae (`Logical error: 'max_log_ptr == new_max_log_ptr'` in `getConsistentMetadataSnapshotImpl`) is already addressed by your own open PR #100651 ("Fix logical error exception in getConsistentMetadataSnapshotImpl when log pointer goes backwards").

That PR's body identifies the exact `chassert(max_log_ptr == new_max_log_ptr)` and replaces it with a `CANNOT_GET_REPLICATED_DATABASE_SNAPSHOT` exception when a `DatabaseReplicated` is dropped + recreated during a backup operation — same root cause that surfaces here under AST fuzzer randomization.

PR #100651 is currently OPEN + MERGEABLE; once it lands, the assertion in this PR's CI clears too. No separate fix PR from my side is needed.

CIDB: 1 hit on this PR's run (today, 2026-05-08); chronic-but-rare DatabaseReplicated path; no master hits in the last 30 days.

clickgapai:
Hi @tiandiwonder @alexey-milovidov — while reviewing this PR I found the following:

- **Coverage gap:** `lowerUTF8` `U_BUFFER_OVERFLOW_ERROR` retry path untested with expanding characters — https://github.com/ClickHouse/ClickHouse/pull/104656

PR #104230, opened 2026-05-06
Coalesce consecutive page cache misses into single S3 requests

alexey-milovidov:
- Performance Improvement

Cold reads of object storage through the userspace page cache (`use_page_cache_for_object_storage = 1`) are now significantly faster, because consecutive cache misses are coalesced into a single HTTP request instead of one request per `page_cache_block_size` block.

### Description

`CachedInMemoryReadBufferFromFile::populateBlockRange` previously issued one `in->readBigAt` per missing 1 MiB block. On object storage, each call is a separate HTTP request, so a cold scan of a 14 GB Parquet file through the userspace page cache made ~15k requests, each paying the TCP/TLS round-trip — measurably slower than the filesystem cache, which fetches in larger segments.

Coalescing was previously implemented in commit 682b070e802 and reverted in c178d2a264f to avoid transient memory spikes from huge temporary buffers under parallel cold reads.

This change re-introduces coalescing with a hard cap on the temporary buffer (``max_coalesced_bytes`` = 16 MiB). Long miss runs are split into multiple fetches, bounding peak transient memory per call. Single-block misses still read directly into the cache cell, avoiding the buffer and the extra ``memcpy``.

### Test results

Measured on c8g.24xlarge against the ClickBench ``clickhouse-datalake`` queries (43 queries, single 14.7 GB Parquet on S3, totals over all queries):

| Config | Cold total | Hot total |
|---|---|---|
| OLD filesystem cache | 62.28s | 18.57s |
| NEW page cache (broken) | 106.14s | 20.14s |
| **NEW page cache + this fix** | **56.58s** | **13.59s** |

Per-query, the worst regressed queries are back to baseline:
- Q24 (heaviest, ``SELECT *``): 15.23s -> 34.61s broken -> **15.27s fixed**
- Q23: 4.89s -> 10.14s broken -> **4.72s fixed**
- Q22: 2.30s -> 5.00s broken -> **2.43s fixed**

Context: https://github.com/ClickHouse/ClickBench/pull/818

### Version info
- Merged into: `26.5.1.356`

### DIFF
--- a/src/IO/CachedInMemoryReadBufferFromFile.cpp
+++ b/src/IO/CachedInMemoryReadBufferFromFile.cpp
@@ -1,5 +1,6 @@
 #include <IO/CachedInMemoryReadBufferFromFile.h>
 #include <base/scope_guard.h>
+#include <Common/PODArray.h>
 #include <Common/ProfileEvents.h>
 
 namespace ProfileEvents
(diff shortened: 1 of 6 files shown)


src/IO/CachedInMemoryReadBufferFromFile.cpp
```diff
@@ -244,17 +245,43 @@ std::vector<PageCache::MappedPtr> CachedInMemoryReadBufferFromFile::populateBloc
...
+    ///
+    /// On object storage, each `in->readBigAt` is a separate HTTP request, so reading one
+    /// block at a time turns a cold scan into one request per block (~15k for a 14 GB file
+    /// at 1 MiB blocks). Coalescing consecutive misses into a single request amortizes that
+    /// overhead.
+    ///
+    /// The coalesced read uses a temporary buffer, capped at `max_coalesced_bytes` to bound
+    /// transient memory under parallel cold reads. A run longer than the cap is split.
+    /// Single-block misses bypass the buffer and read directly into the cache cell.
+    constexpr size_t max_coalesced_bytes = 16 * 1024 * 1024;
```
alexey-milovidov: Can we make it a setting?

alexey-milovidov (commented):
Looks good.

PR #104231, opened 2026-05-06
Polygon dictionary memory accounting and load-path improvements

Algunenano:
Polygon dictionaries (`polygon_index_cell` / `polygon_index_each`) underreport their footprint
in `system.dictionaries.bytes_allocated` and their load path has avoidable peak allocations.

Fixes in this PR:

- `bytes_allocated` now accounts for the lookup index (slab segment trees + grid cells) plus
  the polygons vector capacity and the polygon-id-to-attribute-index map. New stateless test
  `04204_polygon_dict_bytes_allocated` asserts the index-bearing layouts (`polygon_index_cell`,
  `polygon_index_each`) report meaningfully more bytes than `polygon_simple`.

- `IPolygonDictionary::loadData` no longer copies every polygon into a `pair<Polygon, size_t>`
  to sort by area; it sorts an index permutation and rebuilds the destination vectors via
  `std::move`. Peak load-time memory drops from ~3x the polygon storage to ~1x.

- `GridRoot::makeCell` no longer constructs a fresh 128-thread `ThreadPool` on every recursive
  call; one pool is built in the `GridRoot` constructor and shared across the recursion.

- `SlabsPolygonIndex::uniqueX` pre-reserves the gather buffer to its known upper bound
  (`bg::num_points`) so it doesn't grow by capacity doubling.

Closes ClickHouse/clickhouse-private#57161

- Improvement

Reduce memory usage of polygon dictionaries (`polygon_index_cell`, `polygon_index_each`) and make `system.dictionaries.bytes_allocated` correctly reflect the lookup index they build.


src/Dictionaries/PolygonDictionary.cpp
```diff
@@ -716,8 +814,44 @@ void IPolygonDictionary::extractPolygons(const ColumnPtr & column)
...
+    if (offset.polygon_offsets.empty())
+    {
+        data.addPolygon(true, 0, 0);
+    }
+    else
+    {
+        size_t p = 0;
+        while (p < offset.polygon_offsets.size())
+        {
+            const bool new_mp = offset.isFirstPolygonOfMultiPolygon(p);
```
Algunenano: This is out of the scope of the PR. If the PR does not change the code it's not responsible to fix all possible issues you find or adding tests to prove they are or aren't there. That will happen in a separate PR

Algunenano:
Closing to reopen with a clear history

Algunenano:
Superseded by #105431 — the branch was squashed to a single commit containing only the memory-related changes; the reverted extract-side reserve / zero-ring handling is no longer in scope here.

PR #104233, opened 2026-05-06
Naive Sharded Aggregation for high cardinality data

nihalzp:
Simplified version of https://github.com/ClickHouse/ClickHouse/pull/99581 that does not have many of the optimizations like no materialization during scatter, single hashing and reuse of hash, single serialization and reuse of serialized bytes.

There are mostly regressions since data is not usually high cardinality and evenly distributed but If enabled for perf tests, some tests have sped up:
image
image
image
image
image
image
image
image
image

- Performance Improvement

New `GROUP BY` optimization for high cardinality evenly distributed keys that scatters rows across threads by hashing the grouping key, so each thread aggregates a disjoint subset of keys without a merge phase. Set `enable_sharding_aggregator = 1` to enable it.

### Version info
- Merged into: `26.6.1.128`


nihalzp:
Similar performance improvement as the optimized sharded aggregation for high cardinality data and aggregate functions with expensive merge.
image

Some queries are slowed down bit more cases where data copy for scatter is expensive and we rehash. The following query was slowed down by +1.535x only in the optimized version.
image

egor-click:
> Similar performance improvement as the optimized sharded aggregation for high cardinality data and aggregate functions with expensive merge. image
> 
> Some queries are slowed down bit more cases where data copy for scatter is expensive and we rehash. The following query was slowed down by +1.535x only in the optimized version. image

tbh i noticed this pr while working on [perf comparison dashboard](https://gray-waltz.exe.xyz/perf-comparison-report?tab=current&scope=pr&search=104233&metrics=client_time&pr_number=104233&old_sha=345aeb46cf6fb77736385a39cfeca726efc1d7ce&new_sha=b7f4dcaf0b1f80b0ec3a3b6a1021ee8cc9057507), and it looks like we have a lot of degradations, especially in quantile and some group by , some by 5-10x

nihalzp:
> tbh i noticed this pr while working on [perf comparison dashboard](https://gray-waltz.exe.xyz/perf-comparison-report?tab=current&scope=pr&search=104233&metrics=client_time&pr_number=104233&old_sha=345aeb46cf6fb77736385a39cfeca726efc1d7ce&new_sha=b7f4dcaf0b1f80b0ec3a3b6a1021ee8cc9057507), and it looks like we have a lot of degradations, especially in quantile and some group by , some by 5-10x

Yes, this is expected especially for low cardinality and skewed data. My comment was mainly comparing difference between this PR and https://github.com/ClickHouse/ClickHouse/pull/99581 (which is this PR + some optimizations). Both have many degradations.

src/Core/Settings.cpp
```diff
@@ -3953,6 +3953,16 @@ Possible values:
 **See Also**
 
 - [GROUP BY optimization](/sql-reference/statements/select/group-by#group-by-optimization-depending-on-table-sorting-key)
+)", 0) \
+    DECLARE(Bool, optimize_aggregation_by_sharding, true, R"(
```
nickitat: Consider `enable_sharding_aggregator` or smth similar.

src/Processors/QueryPlan/AggregatingStep.cpp
```diff
@@ -20,8 +20,10 @@
 #include <Processors/QueryPlan/QueryPlanStepRegistry.h>
 #include <Processors/QueryPlan/Serialization.h>
 #include <Processors/QueryPlan/SortingStep.h>
+#include <Processors/ResizeProcessor.h>
 #include <Processors/Transforms/AggregatingInOrderTransform.h>
 #include <Processors/Transforms/AggregatingTransform.h>
+#include <Processors/Transforms/BufferedScatterByHashTransform.h>
```
nickitat: We need to stick to one term: either sharding or scattering.

src/Processors/QueryPlan/AggregatingStep.cpp
```diff
@@ -290,6 +294,47 @@ void AggregatingStep::transformPipeline(QueryPipelineBuilder & pipeline, const B
         params.group_by_two_level_threshold_bytes = 0;
     }
 
+    /// Sharded aggregation: pre-partition rows by hash(key) % N before aggregation.
+    /// As a result, same key from different rows will always go to the same shard and we can aggregate
+    /// each shard independently without merge phase.
+    /// Fast for high cardinality keys with evenly distributed data, but has overhead of sharding and is not optimal for low cardinality keys.
```
nickitat:
```suggestion
```

src/Processors/QueryPlan/AggregatingStep.cpp
```diff
@@ -290,6 +294,47 @@ void AggregatingStep::transformPipeline(QueryPipelineBuilder & pipeline, const B
...
+    const bool use_sharded_aggregation = optimize_aggregation_by_sharding
+        /// We may want to use this even in the case of `num_streams = 1`
```
nickitat: I'm not sure it is a good idea. Currently we respect downstream steps decision about the pipeline width and execute on one thread. Let's preserve it here.

src/Processors/QueryPlan/AggregatingStep.cpp
```diff
@@ -290,6 +294,47 @@ void AggregatingStep::transformPipeline(QueryPipelineBuilder & pipeline, const B
...
+        && params.keys_size >= 1
+        /// We do not want to take over cases covered by InOrder Aggregation as those are faster.
+        && sort_description_for_merging.empty()
+        && grouping_sets_params.empty()
+        /// TODO: Support this when we will have external aggregation
+        && !should_produce_results_in_order_of_bucket_number
+        /// Sharding is useful for high cardinality keys. For single-key, skip 1-byte types
+        /// (UInt8/Int8 have at most 256 distinct values) and LowCardinality.
+        /// For multi-key, the combined cardinality is typically high enough.
+        && (params.keys_size > 1
```
nickitat: Maybe we can [re]use this logic:

https://github.com/ClickHouse/ClickHouse/blob/0d8e9caa868f1aaec98cb467202e3ce83772ed1d/src/Interpreters/ConcurrentHashJoin.cpp#L674-L685

src/Processors/QueryPlan/AggregatingStep.cpp
```diff
@@ -290,6 +294,47 @@ void AggregatingStep::transformPipeline(QueryPipelineBuilder & pipeline, const B
...
+        /// Even though there is no merge phase, two-level can help keep each hash table small
+        /// and make hash table operations faster. However, after benchmarking, there have been
+        /// mostly slowdowns for most common queries. Therefore, disable two-level for sharded aggregation.
+        params.group_by_two_level_threshold = 0;
+        params.group_by_two_level_threshold_bytes = 0;
+
+        /// Sharded aggregation does not implement temporary-file spill/merge yet.
+        params.max_bytes_before_external_group_by = 0;
+
+        /// TODO: Support it
```
nickitat: A little advice if I may: use `TODO(<your-handle>)` syntax, otherwise it will be really hard to find all those todo-s later.

src/Processors/QueryPlan/AggregatingStep.cpp
```diff
@@ -504,6 +549,75 @@ void AggregatingStep::transformPipeline(QueryPipelineBuilder & pipeline, const B
         return;
     }
 
+    /// Sharded aggregation: scatter rows by hash(key) % N, then aggregate per shard independently.
+    if (use_sharded_aggregation)
+    {
+        const size_t num_shards = transform_params->params.max_threads;
```
nickitat: We'll need to compare (perhaps later with more significant optimisations already in place) with always choosing a power of two.

src/Processors/QueryPlan/AggregatingStep.cpp
```diff
@@ -504,6 +549,75 @@ void AggregatingStep::transformPipeline(QueryPipelineBuilder & pipeline, const B
...
+
+        /// Resolve key column positions for BufferedScatterByHashTransform.
+        auto stream_header = pipeline.getSharedHeader();
+        ColumnNumbers key_columns;
+        key_columns.reserve(transform_params->params.keys.size());
+        for (const auto & key : transform_params->params.keys)
+            key_columns.push_back(stream_header->getPositionByName(key));
+
+        /// Add BufferedScatterByHashTransform to each stream (1 input -> num_shards outputs).
+        /// After this the pipeline has num_streams * num_shards output ports.
```
nickitat, line 616: We can consider disabling the optimization if `num_streams * num_shards` totals to something infeasible. Like >= 10^5 for example.

src/Processors/Transforms/BufferedShardByHashTransform.h
```diff
@@ -0,0 +1,58 @@
...
+/// IColumn::scatter so each output chunk holds only the rows belonging to its shard.
+///
+/// Output ports can only accept one chunk at a time (canPush/push). But one input chunk
+/// produces N output chunks (one per shard), and downstream consume them at different rates.
+/// Without queueing, we would have to wait until all N outputs are ready
+/// before splitting an input chunk — one slow shard would stall all others.
+///
+/// So each output port has a FIFO queue. When a shard's port is busy, its chunk waits in
+/// the queue and gets pushed on the next prepare()/work() cycle. This allows other shards
+/// to continue processing without waiting for the slowest one.
```
nickitat, line 26: We can imagine a fallback mechanism that will detect very skewed distributions based on one queue being overloaded.

src/Processors/Transforms/BufferedShardByHashTransform.cpp
```diff
@@ -0,0 +1,175 @@
...
+            if (!queued_output_it->isFinished() && queued_output_it->canPush())
+                has_pushable_queued_chunks = true;
+        }
+    }
+
+    /// Input exhausted - drain remaining queues, then finish.
+    if (input.isFinished())
+    {
+        if (has_queued_chunks)
+            return has_pushable_queued_chunks ? Status::Ready : Status::PortFull;
```
nickitat, line 67: If input is finished and all queues are empty can't we return `Finished`?

src/Processors/Transforms/BufferedShardByHashTransform.cpp
```diff
@@ -0,0 +1,175 @@
...
+        return Status::Finished;
+    }
+
+    /// Cannot push any output port
+    if (has_queued_chunks && !has_pushable_queued_chunks)
+        return Status::PortFull;
+
+    /// Try to pull a new input chunk.
+    input.setNeeded();
+    if (input.hasData())
```
nickitat, line 83: We need to limit the max queue length and don't accept new chunks while already at this limit.

src/Processors/Transforms/BufferedScatterByHashTransform.cpp
```diff
@@ -0,0 +1,175 @@
...
+    for (auto column_number : key_columns)
+        hash.update(columns[column_number]->getWeakHash32());
+
+    /// Partition rows by shard using Fibonacci hashing to derive shard bits that are
+    /// independent of the hash table's bucket selection (which uses the low bits via
+    /// hash & mask). Without mixing, all keys in a shard would share the same low bits,
+    /// causing them to cluster into a small subset of hash table buckets.
+    /// The golden ratio constant ensures thorough bit mixing with a single multiply.
+    /// Combine the mix with Lemire fastrange to map into [0, num_shards) without a divide.
+    static constexpr size_t fibonacci_hash_multiplier = 0x9e3779b97f4a7c15ULL;
```
nickitat: We already have similar logic implemented in `scatterBlockByHashGeneric`. Maybe we could reuse it.

src/Core/Settings.cpp
nihalzp: Okay. I have used `enable_sharding_aggregator`.

src/Processors/QueryPlan/AggregatingStep.cpp
nihalzp: Chose sharding version everywhere.

src/Processors/QueryPlan/AggregatingStep.cpp
nihalzp: Done.

src/Processors/QueryPlan/AggregatingStep.cpp
nihalzp: I have made the check more generic by using it.

src/Processors/QueryPlan/AggregatingStep.cpp
nihalzp: Done.

src/Processors/QueryPlan/AggregatingStep.cpp
nihalzp: I have added a TODO.

src/Processors/QueryPlan/AggregatingStep.cpp
nihalzp: Done.

src/Processors/Transforms/BufferedShardByHashTransform.h
nihalzp: Good idea. I have added a TODO.

src/Processors/Transforms/BufferedShardByHashTransform.cpp
nihalzp: Actually, we already do it. The code block is:

```cpp
    if (input.isFinished())
    {
        if (has_queued_chunks) /// <-- if false, means all queues are empty
            return has_pushable_queued_chunks ? Status::Ready : Status::PortFull;

        for (auto & output : outputs)
            output.finish(); /// <-- we close all output ports first to avoid pipeline stuck
        return Status::Finished; /// <-- then we return `Finished`
    }
```

src/Processors/Transforms/BufferedShardByHashTransform.cpp
nihalzp: Yes, good catch. Actually, the original optimized sharded aggregation had it but later I removed it for this naive version because I thought since we split the chunks into smaller parts the memory will be equivalent whether the content is in hash table or in `IColumn` chunks. But then realized, for low cardinality case, the memory difference would be quite high between queued chunks in `BufferedShardByHashTransform.cpp` and their memory in hash tables after aggregation.

I have added a limit of 10 chunks at most. Maybe we can also put a limit in terms of rows or bytes.

src/Processors/Transforms/BufferedScatterByHashTransform.cpp
nihalzp: I could not use `scatterBlockByHashGeneric` directly because the signature requires `Block` and but we only have `Chunk` and there would be conversion overhead if used. I decided to reuse only `JoinCommon::hashToSelector` to avoid doing hashing manually.

src/Processors/Transforms/BufferedShardByHashTransform.cpp
nickitat: my bad

src/Processors/QueryPlan/AggregatingStep.cpp
```diff
@@ -297,28 +298,39 @@ void AggregatingStep::transformPipeline(QueryPipelineBuilder & pipeline, const B
     /// Sharded aggregation: pre-partition rows by hash(key) % N before aggregation.
     /// As a result, same key from different rows will always go to the same shard and we can aggregate
     /// each shard independently without merge phase.
-    /// Fast for high cardinality keys with evenly distributed data, but has overhead of sharding and is not optimal for low cardinality keys.
-    const bool use_sharded_aggregation = optimize_aggregation_by_sharding
-        /// We may want to use this even in the case of `num_streams = 1`
-        /// If `num_streams = 1`, then one stream can be split into N shards and processed in parallel.
+    const bool use_sharded_aggregation = enable_sharding_aggregator
```
nickitat: Let's extract all this logic into a separate method.

tests/integration/test_max_bytes_ratio_before_external_order_group_by_for_server/test.py
```diff
@@ -50,7 +50,7 @@ def test_max_bytes_ratio_before_external_group_by(node):
         # Sharded aggregation disables external group by (no spill support yet),
         # which would cause OOM in this test. Disable sharded aggregation for now.
         # Once sharded aggregation is turned off by default, remove this setting.
-        "optimize_aggregation_by_sharding": 0,
+        "enable_sharding_aggregator": 0,
```
nickitat, line 51: I'd add a TODO to be sure it won't slip through the cracks

src/Processors/Transforms/BufferedShardByHashTransform.cpp
```diff
@@ -119,31 +126,27 @@ void BufferedScatterByHashTransform::work()
...
-    /// independent of the hash table's bucket selection (which uses the low bits via
-    /// hash & mask). Without mixing, all keys in a shard would share the same low bits,
-    /// causing them to cluster into a small subset of hash table buckets.
-    /// The golden ratio constant ensures thorough bit mixing with a single multiply.
-    /// Combine the mix with Lemire fastrange to map into [0, num_shards) without a divide.
-    static constexpr size_t fibonacci_hash_multiplier = 0x9e3779b97f4a7c15ULL;
-    const auto & hash_data = hash.getData();
-    selector.resize_exact(num_rows);
-    for (size_t row = 0; row < num_rows; ++row)
+    if (likely(isPowerOf2(num_shards)))
```
nickitat: I think we're better off not doing it for the reasons described in your original comment. Let's always use the Lemire branch.

nickitat (approved):
All good, modulo a few comments.

clickgapai:
Hi @nihalzp @nickitat — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: The description focuses on performance results but doesn't explain the algorithm or when to use it. Users need to know: (1) what the optimization does (pre-partitions rows by hash(key) so threads aggregate disjoint key subsets), (2) when it helps (high-cardinality, evenly distributed keys), (3) when it's automatically disabled (grouping sets, max_rows_to_group_by, in-order aggregation, distributed queries, LowCardinality keys).
- Suggested addition:

  > ## What this does
  > 
  > Adds `enable_sharding_aggregator` setting (default false) that enables sharded GROUP BY optimization. When enabled, rows are pre-partitioned by `hash(group_by_key) % num_threads` before aggregation, so each thread aggregates a disjoint subset of keys without requiring a merge phase.
  > 
  > ## When to use
  > 
  > - High-cardinality GROUP BY keys (many distinct values)
  > - Evenly distributed key values across rows
  > - Multi-threaded execution (`max_threads > 1`)
  > 
  > ## Automatically disabled when
  > 
  > - `max_rows_to_group_by != 0` (global limit requires merge phase)
  > - GROUPING SETS / ROLLUP / CUBE
  > - Aggregation-in-order optimization active
  > - Distributed queries (bucket-ordered results needed)
  > - Low-cardinality keyspace (UInt8/Int8, LowCardinality types)

PR #104236, opened 2026-05-06
Skip thread rename on no-op.

jmcarp:
Analogous to the behavior of getThreadName, update setThreadName to bail out early if the requested thread name matches the cached thread name.

Context: we noticed that setThreadName is expensive on illumos, and ~10% of thread renames are no-ops. Any performance improvement on linux would be minimal, but there's still no reason to rename a thread to its current name, and this parallels the behavior of getThreadName, which already returns early if we have a cached thread name.

### Version info
- Merged into: `26.5.1.364`

### DIFF
--- a/src/Common/setThreadName.cpp
+++ b/src/Common/setThreadName.cpp
@@ -74,7 +74,9 @@ static thread_local ThreadName thread_name = ThreadName::UNKNOWN;
 
 void setThreadName(ThreadName name)
 {
-    thread_name = name;
+    // Skip rename on no-op.
+    if (thread_name == name)
+        return;
 
     auto thread_name_view = toString(name);
     if (thread_name_view.size() > THREAD_NAME_SIZE)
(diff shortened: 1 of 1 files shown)


jmcarp:
By the way, would you all be open to a new setting that disables thread renames entirely? AFAICT this is only used for debugging, and the performance cost on illumos is severe.

alexey-milovidov (approved):

alexey-milovidov:
> By the way, would you all be open to a new setting that disables thread renames entirely? AFAICT this is only used for debugging, and the performance cost on illumos is severe.

There are two purposes:
1. Display it in internal introspection tools, such as `system.stack_trace`, logs, etc.
2. Tell the OS about the thread name, so that it is visible in its tools, such as `top`, `ps`, etc.

The most important is 1, but only 2 is heavy. If we only need 1, we can use the thread_local variable, and don't set the thread name for the OS.

So let's simply skip 2 if ClickHouse is built for Illumos.
Let's do it in another PR.

PR #104248, opened 2026-05-06
Guard private-only includes with #if CLICKHOUSE_CLOUD (small files)

alexey-milovidov:
Port small changes from clickhouse-private PR https://github.com/ClickHouse/clickhouse-private/pull/55164.

The private PR wraps unguarded `#include` of private-only headers (and a few small private-only code blocks) with `#if CLICKHOUSE_CLOUD` / `#endif`, so that the `Check public build` CI job — which copies common files from the private repo into the public one and builds them against the public codebase — does not fail when a PR happens to touch any of those files.

This commit applies the same guards on the public side for the smaller files (changes whose wrapped contents are under 10 lines), so that those common files become byte-identical between the two repositories. `CLICKHOUSE_CLOUD` is `0` in the public build, so all guarded blocks compile out and there is no behaviour change.

Some larger changes and changes that depended on public-side placeholder code (e.g. unused-variable warnings or missing `config.h` includes) were intentionally left out and will be ported separately. The build of `dbms` was verified locally with these changes applied.

...

### Version info
- Merged into: `26.5.1.487`


alexey-milovidov:
@groeneai fix `00146_aggregate_function_uniq` and send a separate PR.

groeneai:
@alexey-milovidov — fix is up at #104269.

Diagnosis (from the report on `ba3d3f4`):
- The CI's own `--diagnose-random-settings` confirmed the failure is settings-induced (`Step 1: 1/1 fail with the captured settings; Step 2: 4/4 pass without randomization`).
- Root cause: the third query, `SELECT abs(uniq(WatchID) - uniqExact(WatchID)) FROM test.hits`, materializes every distinct `WatchID` from the ~100M-row hits dataset into a `HashSet`. Under MSan's per-access shadow check, plus `compile_aggregate_expressions = 0`, `max_threads = 2`, `max_block_size = 14392`, and `merge_tree_read_split_ranges_into_intersecting_and_non_intersecting_injection_probability = 0.72`, it crosses the 600s per-test timeout. Without those settings the test finishes well under 60s.
- Cross-check: `00146_aggregate_function_uniq` had not failed once on `amd_msan, WasmEdge` in 90 days (a few March-21 hits were the unrelated `clickhouse-test` infrastructure `ValueError: too many values to unpack`).

Fix: add `no-msan` to the test tags (matches the existing convention used by `00024_random_counters`, `00157_cache_dictionary`, `00326_long_function_multi_if`, `00625_summing_merge_tree_merge`, `00804_test_alter_compression_codecs`, etc.).

PR #104251, opened 2026-05-06
Parallelize reads from a single Parquet file in StorageFile

alexey-milovidov:
On ClickBench, single-file Parquet runs are 3–9× slower than the 100-file partitioned runs on the same data (e.g. on `c7a.metal-48xl`, Q23 is `8.90s` vs `0.99s`, Q22 `1.82s` vs `0.41s`, Q27 `1.21s` vs `0.45s`). The cause is in `StorageFile`: when reading a single splittable file it creates exactly one `ParquetV3BlockInputFormat` source, so the pipeline becomes `File 0 → 1` followed by `Resize 1 → 96`. That fan-out is a serialization point — every chunk has to leave the single source through one `read` before any of the 96 aggregators can touch it, so most cores sit idle.

The bucket-splitting machinery (`ParquetBucketSplitter`, `setBucketsToRead`, `FileBucketInfo`) already existed for cluster mode but was never wired into `StorageFile`. This PR wires it in:

- New `IBucketSplitter::splitToBucketsByCount` returning roughly N contiguous row-group ranges; Parquet implements it.
- New `FormatFactory::checkFormatHasSplitter` so callers can probe without throwing.
- `StorageFile::ReadFromFile::initializePipeline`, when reading exactly one local splittable file, asks the splitter for `max_num_streams` buckets and creates one `StorageFileSource` per bucket. Each source carries `fixed_file_path` + `file_bucket_info` and skips the shared `FilesIterator`.
- `ParquetV3BlockInputFormat::read` honours `buckets_to_read` in the trivial-count path so each bucket only reports its own row count.
- The count cache (keyed by file path) is bypassed for bucketed reads — otherwise every bucket would report the file's total and counts would be multiplied by the number of buckets.

Pipeline becomes `File × N 0 → 1` straight into the aggregators, matching the partitioned variant.

### Results

96-vCPU box, `hits.parquet` (14 GiB, 226 row groups):

|     | Single (master) | Single (this PR) | Partitioned |
| --- | --------------- | ---------------- | ----------- |
| Q21 | 0.40–0.66s      | **0.44s**        | 0.34s       |
| Q22 | 0.93–1.36s      | **0.48s**        | 0.41s       |
| Q23 | 1.33–1.45s      | **0.55s**        | 0.42s       |
| Q26 | 0.50s           | **0.35s**        | 0.19s       |
| Q27 | 1.6s            | **0.54s**        | 0.45s       |

CPU utilisation on Q23 jumped from ~6× to ~18× of 96 cores. Aggregate results (`count`, `sum(UserID)`, `sum(length(URL))`, Q21, Q23) match the partitioned variant exactly. The remaining ~1.3× gap to partitioned is per-source initialization overhead: each bucket source still reads the 14 GB file's footer separately. Sharing parsed metadata for local files is the obvious next step but a much bigger change.

- Performance Improvement

Reading a single large local Parquet file via `file()` / `File` engine is now parallelised across multiple sources, each handling a subset of row groups. This eliminates a `Resize 1 → N` bottleneck in the pipeline and brings single-file ClickBench performance close to the partitioned variant — Q23 goes from ~1.4s to ~0.55s, Q22 from ~0.9s to ~0.48s, Q27 from ~1.6s to ~0.54s on 96 vCPUs.

### Version info
- Merged into: `26.5.1.395`

### DIFF
--- a/src/Formats/FormatFactory.cpp
+++ b/src/Formats/FormatFactory.cpp
@@ -843,6 +843,14 @@ BucketSplitter FormatFactory::getSplitter(const String & format)
     return creator.bucket_splitter_creator();
 }
 
+bool FormatFactory::checkFormatHasSplitter(const String & format) const
+{
+    auto it = dict.find(boost::to_lower_copy(format));
+    if (it == dict.end())
+        return false;
+    return static_cast<bool>(it->second.bucket_splitter_creator);
+}
+
 void FormatFactory::registerRandomAccessInputFormat(const String & name, RandomAccessInputCreator input_creator)
 {
     chassert(input_creator);
--- a/src/Formats/FormatFactory.h
+++ b/src/Formats/FormatFactory.h
@@ -373,6 +373,9 @@ public:
     void registerFileBucketInfo(const String & format, FileBucketInfoCreator bucket_info);
     void registerSplitter(const String & format, BucketSplitterCreator splitter);
     BucketSplitter getSplitter(const String & format);
+    /// Returns true if `format` is registered and has a bucket splitter
+    /// (e.g. Parquet). Used to decide whether to attempt single-file parallel splitting.
+    bool checkFormatHasSplitter(const String & format) const;
 
 private:
     FormatsDictionary dict;
--- a/src/Processors/Formats/IInputFormat.h
+++ b/src/Processors/Formats/IInputFormat.h
@@ -66,6 +66,11 @@ struct IBucketSplitter
     /// Returns information about the resulting buckets (see the structure above for details).
     virtual std::vector<FileBucketInfoPtr> splitToBuckets(size_t bucket_size, ReadBuffer & buf, const FormatSettings & format_settings_) = 0;
 
+    /// Splits a file into approximately `target_count` buckets, each covering a roughly
+    /// equal slice of the file. Useful for parallelising one large file across N readers.
+    /// The result has at most `target_count` buckets and never drops any data.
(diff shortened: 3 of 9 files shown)


alexey-milovidov:
Added a follow-up commit (c019cdb) that wires the existing Parquet metadata cache (`parquet_metadata_cache_size`, 512 MiB default) into `StorageFile` for local files.

Until now the cache was only reachable from object storage, because the local-file creator passed `nullptr` and the cache lookup in `ParquetV3BlockInputFormat::getFileMetadata` requires `object_with_metadata` with an etag. With single-file bucket splitting that meant every one of the 96 sources parsed the footer of `hits.parquet` from scratch.

`StorageFile` now builds a synthetic `RelativePathWithMetadata` for local files, using `mtime|size` as the etag (so an in-place overwrite invalidates the entry), and calls `getInputWithMetadata`. Verified with `--send_logs_level=trace`: per Q23 we see **1 cache miss + 95 cache hits** (was 96 misses).

Wall-time impact on this 96-core box is in the noise (~1.9 s of CPU work eliminated, but parallelised across 96 cores ≈ 20 ms wall):

|     | Cache off | Cache on |
| --- | --------- | -------- |
| Q23 (cold) | 0.576 s | 0.567 s |
| Q23 (hot, avg of 7) | 0.407 s | 0.408 s |

The savings are still real and will be visible on smaller machines, in batches of small queries, and for long-running `clickhouse-server` where the cache entry stays hot across queries on the same file.

alexey-milovidov (commented):
I've reviewed the code and found everything alright.
The only question is why it didn't work this way in the first place.

PR #104259, opened 2026-05-06
test: FTS usage on alias columns with arrayMap

knudtty:

You can now use a FTS index if on an ALIAS column with a lambda


knudtty:
According to my research, the real issue is the old analyzer serializing expression when the schema is built, and uses the new analyzer at query time. Fixing that issue is probably way over my surface level knowledge.

knudtty:
This is a duplicated PR, another better solution exists already

PR #104260, opened 2026-05-06
Use Parquet footer cache for local files in StorageFile

alexey-milovidov:
The Parquet metadata cache (default 512 MiB) was registered in `Context` and wired into the object-storage code path, but `StorageFile` was calling `getInput` with no `object_with_metadata` — so for `file('foo.parquet')` the cache was never consulted and every source parsed the footer fresh. With the single-file bucket-splitting work in #104251, that meant up to `max_threads` separate footer parses per query against `hits.parquet` (96 on a `c7a.metal-48xl`).

This PR builds a synthetic `RelativePathWithMetadata` for real local files (etag = `mtime|size`, so an in-place rewrite invalidates the entry) and calls `getInputWithMetadata`. The existing Parquet metadata-aware creator picks up the cache from `Context`.

Two paths needed care:

- `INSERT ... FROM INFILE` runs `StorageFile` inside the *client* `Context`, which never calls `setParquetMetadataCache`. The original metadata-aware creator threw `LOGICAL_ERROR: Parquet metadata cache was not created yet`. New `Context::tryGetParquetMetadataCache` returns `nullptr` instead of throwing; the format already short-circuits when the cache is null.
- Reads with no per-file `size + mtime` (archive entries, `use_table_fd`, missing stat info) tripped `chassert(object_with_metadata.has_value())` in `getInputWithMetadata`. `StorageFileSource::generate` now dispatches between `getInputWithMetadata` and `getInput` based on whether the optional is populated.

Verified with `--send_logs_level=trace`: each query against `hits.parquet` on a 96-core box now logs 1 cache miss + 95 cache hits (was 96 misses). On the ClickBench heavy-text queries the wall-time delta is in the noise on this hardware (cache savings ≈ 1.9 s of CPU work, parallelised across 96 cores ≈ 20 ms), but the cache is meaningful for smaller machines and for long-running `clickhouse-server` where the entry stays hot across queries.

Split out from #104251 so the parallelism change and the cache wiring can be reviewed and rolled back independently.

- Performance Improvement

Reuse the Parquet footer metadata cache when reading local Parquet files via the `file` table function or `File` table engine. Previously the cache was only consulted for object-storage backends.

### Version info
- Merged into: `26.5.1.493`


alexey-milovidov:
Yes:

Screenshot_20260510_122839
Screenshot_20260510_122856

grantholly-clickhouse (approved):
LGTM.  I am happy to see that we can integrate local file support while using the same etag logic

PR #104262, opened 2026-05-06
Revert "Optimize `cramersV`, `cramersVBiasCorrected`, `theilsU`, and `contingency` when used with window functions"

alexey-milovidov:
Reverts ClickHouse/ClickHouse#93384


nihalzp:
The fix is here:

- https://github.com/ClickHouse/ClickHouse/pull/104277

PR #104272, opened 2026-05-06
Pin prewhere settings in 04051_pk_analysis_stats

groeneai:
The test `04051_pk_analysis_stats` verifies that PK analysis stats correctly account for granules already dropped by the `QueryConditionCache` (QCC). The docstring at the top of the test references a specific log-line example where the QCC drops granules for the `PREWHERE` condition first and PK analysis then runs on the remaining marks.

Without explicit pins, the test runner's `--random-settings` injection can flip `optimize_move_to_prewhere` and/or `query_plan_optimize_prewhere` off. The QCC's verdict then ends up keyed by the `WHERE`-side hash instead of the `PREWHERE`-side hash documented in the comment block, exercising a different code path on every run.

The assertion `ProfileEvents['FilteringMarksWithPrimaryKeyProcessedMarks'] < total_marks` still holds in either path (both sides drop marks before PK analysis runs in `MergeTreeDataSelectExecutor::filterPartsByQueryConditionCache`). But this variability has been a recurring source of confusion: when an unrelated PR's CI surfaces a transient or PR-caused failure for this test, `--diagnose-random-settings` minimizes the culprit set down to a single non-causal session setting (e.g. `--session_timezone`), and the failure then gets misattributed to settings that have no semantic relationship with the test.

Pin both prewhere settings inside the SQL so the test always exercises the documented `PREWHERE` QCC drop path. SETs inside the test SQL override the CLI flags injected by the runner, so this is robust against any future random-settings expansion that touches prewhere.

- 50/50 passes with default `--random-settings` randomisation.
- 100/100 passes with `-j 4 --test-runs 100`.
- 50/50 passes with adversarial overrides (`--use_query_condition_cache 0 --optimize_move_to_prewhere False --query_plan_optimize_prewhere False --session_timezone America/Mazatlan --enable_multiple_prewhere_read_steps 0 --max_threads 1 --max_block_size 17886`).

CI report that motivated the fix: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104225&sha=ef78dcecf6d3c6a1ccad2481e090794460e778cb&name_0=PR&name_1=Stateless%20tests%20%28amd_debug%2C%20parallel%29

Related: https://github.com/ClickHouse/ClickHouse/pull/103857#issuecomment-4391247864 (`@alexey-milovidov` directive to fix this test in a separate PR).

...

### Version info
- Merged into: `26.5.1.406`

### DIFF
--- a/tests/queries/0_stateless/04051_pk_analysis_stats.sql
+++ b/tests/queries/0_stateless/04051_pk_analysis_stats.sql
@@ -23,6 +23,18 @@ optimize table t final;
 set enable_analyzer=1;
 set use_query_condition_cache=1;
 
+-- Pin the settings that the assertion depends on. The test runner injects
+-- random session settings; without these pins, the runner can flip
+-- `optimize_move_to_prewhere` / `query_plan_optimize_prewhere` to `0` and the
+-- query condition cache then writes its verdict against the WHERE-side hash
+-- instead of the PREWHERE-side hash documented in the comment block at the
+-- top of this file. The assertion still holds for the WHERE-side path, but
+-- pinning both keeps the test exercising the same code path the regression
+-- comment describes and removes a class of CI noise that is unrelated to
+-- the bug under test.
+set optimize_move_to_prewhere=1;
+set query_plan_optimize_prewhere=1;
+
 -- Pre-warm the query condition cache
 select avg(a) from t where s != 'xxx' format Null;
 


groeneai:
**Pre-PR validation gate self-check** (worker session: `cron:clickhouse-ci-task-worker:20260506-231500`)

a) **Deterministic repro?** Yes for the FAILURE: PR #104225 commit `ef78dcecf6d` reproduced 117/117 in CI's `--diagnose-random-settings` step 1 across 6 build variants. The root cause was that PR #104225's earlier gate blocked QCC writes for any changed Beta-tier setting, and the test runner injects `--session_timezone <Beta-tier-zone>` on most runs, which made the warmup query unable to populate the cache. The latest PR #104225 commit `1860c812f0f` already replaces the tier-based gate with name-pattern matching (allowing Beta-tier settings like `session_timezone` to remain writable), so the original failure mode is no longer reproducible on PR #104225 HEAD either.

b) **Root cause explained?** Yes — two layers:
- *Layer 1 (already fixed in PR #104225):* The buggy commit's gate `if (tier == BETA || tier == EXPERIMENTAL) return false` rejected QCC writes when any Beta-tier session setting was non-default. The runner injects `--session_timezone` (Beta) on every run, so the warmup query always failed to populate the cache, the measurement query found no entry, PK analysis ran on all 25 marks, and `throwIf(processed >= total_marks)` fired.
- *Layer 2 (this PR):* Even after Layer 1 is fixed, the test is sensitive to which prewhere setting the runner randomizes. With `optimize_move_to_prewhere=0` and `query_plan_optimize_prewhere=0`, the QCC entry is keyed by the WHERE-side hash instead of the PREWHERE-side hash that the test's docstring describes. The assertion still holds (both sides drop before PK analysis), but the test exercises a code path different from what the comment documents — and `--diagnose-random-settings` minimizes the culprit to non-causal settings on subsequent runs, which has wasted CI cycles and triage attention.

c) **Fix matches root cause?** Yes for Layer 2. Pinning `optimize_move_to_prewhere=1` and `query_plan_optimize_prewhere=1` keeps the test exercising the documented PREWHERE QCC drop path. SETs inside the SQL override CLI flags injected by the runner, so the pin is robust against future random-settings expansion. Layer 1 was fixed independently in PR #104225 itself.

d) **Test intent preserved?** Yes — the assertion is unchanged; the test still verifies that PK analysis stats correctly account for granules already dropped by the QCC. Pinning prewhere does not weaken any assertion; it just removes a source of test/code-path drift that has historically misled triage.

e) **Both directions demonstrated?** Yes. With the pin (this PR's change) on a fresh debug build:
- 50/50 passes with default `--random-settings` randomisation.
- 100/100 passes with `-j 4 --test-runs 100`.
- 50/50 passes with adversarial CLI flags that previously surfaced as the diagnoser's "minimal culprit": `--use_query_condition_cache 0 --optimize_move_to_prewhere False --query_plan_optimize_prewhere False --session_timezone America/Mazatlan --enable_multiple_prewhere_read_steps 0 --max_threads 1 --max_block_size 17886`. The pin's SETs override the CLI flags and the test runs on the documented PREWHERE path.

f) **Fix is general?** This is a test-stability change rather than a real bug fix, so the "search for sibling defects" criterion is N/A. There are no other tests with the same `set use_query_condition_cache=1` + reliance on PREWHERE QCC drops shape — this is the only one. (Most other QCC tests carry `no-parallel` because they read `system.query_condition_cache` or call `SYSTEM CLEAR QUERY CONDITION CACHE`; 04051 doesn't.)

groeneai:
cc @nickitat — could you take a look? You authored 04051_pk_analysis_stats and are the most active QueryConditionCache contributor; this is a small defensive pin to keep the test on the documented PREWHERE QCC drop path so random-settings noise stops misleading triage. Per @alexey-milovidov's directive on #103857.

nickitat (approved):

Issue #103486, opened 2026-04-24
`UNIQUE KEY` for ClickHouse


murphy-4o:
## PR
- https://github.com/facebook/rocksdb/pull/14706
- https://github.com/ClickHouse/rocksdb/pull/30
- https://github.com/ClickHouse/ClickHouse/pull/104102
- https://github.com/ClickHouse/ClickHouse/pull/104279
- https://github.com/ClickHouse/ClickHouse/pull/105401
- https://github.com/ClickHouse/ClickHouse/pull/104282
- https://github.com/ClickHouse/ClickHouse/pull/104284
- https://github.com/ClickHouse/ClickHouse/pull/107986
- https://github.com/ClickHouse/ClickHouse/pull/107834
- https://github.com/ClickHouse/ClickHouse/pull/108348
- https://github.com/ClickHouse/ClickHouse/pull/106422
- https://github.com/ClickHouse/ClickHouse/pull/111189
- https://github.com/ClickHouse/ClickHouse/pull/116845

---

## TL;DR

Add `UNIQUE KEY` as a DDL clause on existing MergeTree engines. Reads return deduped rows without `FINAL`.

```sql
CREATE TABLE orders (
    order_id    UInt64,
    region      LowCardinality(String),
    status      String,
    amount      Decimal(10, 2),
    updated_at  DateTime
) ENGINE = MergeTree
UNIQUE KEY (order_id)                   -- business identity, dedup key
ORDER BY  (region, updated_at);         -- physical sort, optimized for reads
```

- **INSERT is implicitly UPSERT** — a duplicate `order_id` supersedes the previous row.
- **SELECT returns one row per key** — no `FINAL`, no merge-sort.
- **DELETE is synchronous** — bitmap flip, rows invisible on the next query.
- `UNIQUE KEY` is decoupled from `ORDER BY` — pick the physical sort that best serves reads; dedup identity is independent.
- **Planned across `MergeTree`, `ReplicatedMergeTree`, `SharedMergeTree`** — one user-facing contract, different coordination mechanisms under the hood. Engines ship in phases (see §6).
- **Uniqueness is per-partition**, not global (see §2).

---

## 1. Problem

ClickHouse has no way to guarantee a key appears **exactly once at query time** without paying the `FINAL` tax. The pain surfaces in interactive workloads over mutable operational data:

- **CDC ingestion** — row-level changes from a transactional DB (orders, user profiles, account state) need to land as one live row per primary key, not a per-change log.
[...]

PR #104279, opened 2026-05-07
UNIQUE KEY: DeleteBitmap primitive and sidecar file I/O (2/N)

murphy-4o:
Part of https://github.com/ClickHouse/ClickHouse/issues/103486

Second slice of the UNIQUE KEY stack. Adds the primitive bitmap data model and the on-disk sidecar file I/O contract.

Scope:
- Roaring-backed bitmap representation with serialize / deserialize helpers and a versioned filename convention (`delete_bitmap_{N}.rbm`).
- `DeleteBitmapFileOps` helpers for enumerating sidecars, choosing the highest version, reading, writing, and discovering the current storage version with tmp-write plus atomic rename.

### Version info
- Merged into: `26.5.1.546`


src/Storages/MergeTree/UniqueKey/DeleteBitmap.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+    if (magic != MAGIC)
+        throw Exception(ErrorCodes::CORRUPTED_DATA,
+            "DeleteBitmap magic mismatch: expected {:#x}, got {:#x}", MAGIC, magic);
+
+    if (version != VERSION)
+        throw Exception(ErrorCodes::UNKNOWN_FORMAT_VERSION,
+            "DeleteBitmap version {} is not supported (expected {})", version, VERSION);
+
+    std::vector<char> body(body_size_u32);
```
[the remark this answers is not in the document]
murphy-4o: Already fixed in 97bc7c7af78 (`DeleteBitmap deserialize allocation guard`). `MAX_SERIALIZED_BODY_SIZE` (256 MiB) is checked before the `std::vector<char> body(body_size_u32)` allocation. As of b8cc4661392 `serialize` enforces the same bound, so any payload that gets written is guaranteed to be readable.

src/Storages/MergeTree/UniqueKey/DeleteBitmap.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+    if (magic != MAGIC)
+        throw Exception(ErrorCodes::CORRUPTED_DATA,
+            "DeleteBitmap magic mismatch: expected {:#x}, got {:#x}", MAGIC, magic);
+
+    if (version != VERSION)
+        throw Exception(ErrorCodes::UNKNOWN_FORMAT_VERSION,
+            "DeleteBitmap version {} is not supported (expected {})", version, VERSION);
+
+    if (body_size_u32 > MAX_SERIALIZED_BODY_SIZE)
```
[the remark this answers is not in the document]
murphy-4o: Fixed in b8cc4661392 — `serialize` now enforces `MAX_SERIALIZED_BODY_SIZE` (256 MiB), matching `deserialize`. Any payload that we write is guaranteed to be readable.

src/Storages/MergeTree/UniqueKey/DeleteBitmap.cpp
```diff
@@ -0,0 +1,250 @@
...
+    if (body_size_u32)
+        in.readStrict(body.data(), body_size_u32);
+
+    UInt32 stored_crc = 0;
+    readPODBinary(stored_crc, in);
+
+    UInt32 computed_crc = computeCRC32(header, sizeof(header), body.data(), body_size_u32);
+    if (stored_crc != computed_crc)
+        throw Exception(ErrorCodes::CORRUPTED_DATA,
+            "DeleteBitmap CRC mismatch: stored {:#x}, computed {:#x}", stored_crc, computed_crc);
```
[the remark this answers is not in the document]
murphy-4o: Fixed in b8cc4661392 — `deserialize` now throws `CORRUPTED_DATA` if `!in.eof()` after the CRC. Added `TrailingBytesAfterCRCRejected` regression.

src/Storages/MergeTree/UniqueKey/DeleteBitmap.h
```diff
@@ -0,0 +1,106 @@
...
+namespace DB
+{
+
+class ReadBuffer;
+class WriteBuffer;
+
+/** UNIQUE KEY per-part delete bitmap — row positions (within a part, 0-based)
+  * that are logically deleted. Wraps a 32-bit `roaring::Roaring` — this is the
+  * only bitmap type we need because MergeTree parts are addressed by `UInt32`
+  * row numbers (`_part_offset`), and a single part caps out well below 2^32
```
alesapin: :( AFAIK we can have more rows in data parts...

src/Storages/MergeTree/UniqueKey/DeleteBitmap.cpp
```diff
@@ -0,0 +1,259 @@
...
+    /// Serialize header into a fixed buffer so we can CRC it as one blob.
+    char header[sizeof(UInt32) * 3];
+    UInt32 body_size_u32 = static_cast<UInt32>(body_size);
+    std::memcpy(header + 0,                   &MAGIC,         sizeof(UInt32));
+    std::memcpy(header + sizeof(UInt32),      &VERSION,       sizeof(UInt32));
+    std::memcpy(header + sizeof(UInt32) * 2,  &body_size_u32, sizeof(UInt32));
+
+    UInt32 crc = computeCRC32(header, sizeof(header), body.data(), body_size);
+
+    out.write(header, sizeof(header));
```
alesapin, line 351: [Up to you] Maybe write header as footer? Because now it looks like we do double copy: from `bitmap` to `body`, from `body` to buffer.

src/Storages/MergeTree/UniqueKey/DeleteBitmap.cpp
```diff
@@ -0,0 +1,259 @@
...
+        return std::nullopt;
+    if (!file_name.ends_with(FILE_SUFFIX))
+        return std::nullopt;
+
+    auto number_part = file_name.substr(FILE_PREFIX.size(), file_name.size() - FILE_PREFIX.size() - FILE_SUFFIX.size());
+    if (number_part.empty())
+        return std::nullopt;
+
+    UInt64 value = 0;
+    auto [ptr, ec] = std::from_chars(number_part.data(), number_part.data() + number_part.size(), value);
```
alesapin: we have our own `parse<UInt64>(str)`

src/Storages/MergeTree/UniqueKey/DeleteBitmapFileOps.cpp
```diff
@@ -0,0 +1,104 @@
...
+namespace DeleteBitmapFileOps
+{
+
+std::vector<std::pair<UInt64, std::string>> enumerateFiles(const IDataPartStorage & storage)
+{
+    std::vector<std::pair<UInt64, std::string>> result;
+    for (auto it = storage.iterate(); it->isValid(); it->next())
+    {
+        const auto & file_name = it->name();
+        if (auto block_number = DeleteBitmap::parseBlockNumberFromFileName(file_name))
```
alesapin: It would be more clear to have `isDynamicBitmapFile` + `parseBlockNumber...`. I don't think it really affect performance to do it in once call.

alesapin (approved):
LGTM. Are we going to have some abstraction on top of DynamicBitmap, like VersionedDynamicBitmap?

murphy-4o:
> LGTM. Are we going to have some abstraction on top of DynamicBitmap, like VersionedDynamicBitmap?

yeah, we would have an abstraction layer for various engines including `MergeTree/ReplicatedMergeTree/SharedMergeTree`, for components like BitmapStore & Commit & Merge:
```c++
  class IUniqueKeyBitmapStore {
      virtual void writeBitmap(const IMergeTreeDataPart & part, UInt64 version, const DeleteBitmap & bitmap) = 0;
      virtual std::shared_ptr<DeleteBitmap> readBitmap(const IMergeTreeDataPart & part, UInt64 version) = 0;
      virtual UInt64 getCurrentVersion(const IMergeTreeDataPart & part) = 0;
      virtual UniqueKeyPinnedVersions pinVersions(const String & partition_id) = 0;
  };
```

src/Storages/MergeTree/UniqueKey/DeleteBitmap.h
murphy-4o: changed to dynamic upgrade, which is more complicated but keep the same efficiency: 
```c++
    using R32Ptr = std::unique_ptr<roaring::Roaring>;
    using R64Ptr = std::unique_ptr<roaring::Roaring64Map>;
    std::variant<R32Ptr, R64Ptr> bitmap;
```

src/Storages/MergeTree/UniqueKey/DeleteBitmap.cpp
murphy-4o: changed.

src/Storages/MergeTree/UniqueKey/DeleteBitmap.cpp
murphy-4o: fixed

src/Storages/MergeTree/UniqueKey/DeleteBitmapFileOps.cpp
murphy-4o: changed.

src/Storages/MergeTree/UniqueKey/DeleteBitmap.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+    auto result = std::make_unique<DeleteBitmap>();
+    if (version == VERSION_R64)
+    {
+        auto r64 = std::make_unique<roaring::Roaring64Map>();
+        if (body_size_u32)
+        {
+            /// `readSafe` validates container counts against `maxbytes`, so a
+            /// malformed payload that survived CRC still can't over-read.
+            *r64 = roaring::Roaring64Map::readSafe(rest.data(), body_size_u32);
```
alesapin: Shouldn't it read and use body_size?

murphy-4o: Format reverted 
- from `magic | version | body | body_size | crc` 
- to `magic | version | body_size | body | crc`

Both patterns can be effective in this situation, but the second one is more intuitive, so it would be reasonable to switch to it.

PR #104258, opened 2026-05-06
Fix Bad cast on data-lake `SETTINGS disk = '...'` with wrapped object storage

groeneai:
Fixes #89300.

## Motivation

A user reported via #89300 that

```sql
SELECT 1 FROM icebergS3(s3, format = 'Parquet', filename = '...', SETTINGS disk = 'disk0');
```

against a disk configured with `metadata_type = plain` aborted with
`Logical error: 'Bad cast from type DB::PlainObjectStorage<DB::S3ObjectStorage> to DB::S3ObjectStorage'`.

The `PlainObjectStorage<>` decorator was deleted by #90580, so the
exact crash from the original report is gone, but the underlying bug
pattern remains: `S3StorageParsedArguments::fromDisk` (and the Azure
sibling) call `assert_cast<const S3ObjectStorage &>(*disk->getObjectStorage())`
directly. `assert_cast` performs an exact `typeid` match in
debug/sanitizer builds, so any decorator that wraps the underlying
storage instead of being it (most importantly `CachedObjectStorage`,
returned by `DiskObjectStorage::wrapWithCache`) still crashes argument
parsing with `Logical error: 'Bad cast from type DB::CachedObjectStorage to DB::S3ObjectStorage'`.

I confirmed this is currently reachable in trunk:

- The BuzzHouse fuzzer hit it on PR #94148 with
  `CREATE TABLE d1.t28 ENGINE = DeltaLakeS3(s3, filename = 't28', format = 'Parquet') SETTINGS disk = 'disk1', iceberg_format_version = 3`.
- Reproduced locally: a debug ClickHouse with a `cache`-typed disk over
  `s3` and `SETTINGS disk = '<cache_disk>'` aborts the server in the
  exact same way.

## Approach

`IObjectStorage` already exposes `getUnderlying()` which returns the
inner storage for decorator types and `nullptr` otherwise. Walk that
chain before the cast in both `S3StorageParsedArguments::fromDisk` and
`AzureStorageParsedArguments::fromDisk`. After unwrapping, the cast
sees the concrete `S3ObjectStorage` / `AzureObjectStorage` and
succeeds.

## Validation

Pre-PR validation gate:

- **Deterministic repro?** Yes — confirmed locally with a debug build:
  ```
  SELECT 1 FROM icebergS3('no_such_table', 'Parquet', 'a Int',
                          SETTINGS disk = 's3_cache_test')
  ```
  on master HEAD aborts with
[...]

Issue #89300, opened 2025-10-31
Logical error: 'Bad cast from type DB::PlainObjectStorage<DB::S3ObjectStorage> to DB::S3ObjectStorage'


PedroTadim:
### Describe the bug

Easy to reproduce.

### How to reproduce

Start MinIO with `tests/docker_scripts/setup_minio.sh stateless` command. Then set these settings in config.xml:
```xml
<storage_configuration>
    <disks>
        <disk0>
            <type>object_storage</type>
            <object_storage_type>s3</object_storage_type>
            <metadata_type>plain</metadata_type>
            <endpoint>http://localhost:11111/test</endpoint>
            <access_key_id>clickhouse</access_key_id>
            <secret_access_key>clickhouse</secret_access_key>
        </disk0>
    </disks>
</storage_configuration>
<allowed_disks_for_table_engines>disk0</allowed_disks_for_table_engines>
```
This is my users.xml: [users.txt](https://github.com/user-attachments/files/23262319/users.txt)
Create the table in Spark with: [spark.py](https://github.com/user-attachments/files/23262383/spark.py)
Then run the statements in ClickHouse:
```sql
CREATE TABLE t0 ENGINE = IcebergS3(s3, format = 'Parquet', filename = 't0/test/t0');
SELECT 1 FROM icebergS3(s3, format = 'Parquet', filename = 't0/test/t0', structure = 'c0 Int', SETTINGS disk = 'disk0') tx; --Logical error
```
The table does exist in MinIO, so the statement should run successfully?

### Error message and/or stacktrace

Stack trace:
```
<Fatal> : Logical error: 'Bad cast from type DB::PlainObjectStorage<DB::S3ObjectStorage> to DB::S3ObjectStorage'.
<Fatal> : Stack trace (when copying this message, always include the lines below):

0. contrib/llvm-project/libcxx/include/__exception/exception.h:113: Poco::Exception::Exception(String const&, int) @ 0x0000000024ad4af2
1. src/Common/Exception.cpp:129: DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0x0000000015070029
2. src/Common/Exception.h:123: DB::Exception::Exception(String&&, int, String, bool) @ 0x000000000d60df4e
3. src/Common/Exception.h:58: DB::Exception::Exception(PreformattedMessage&&, int) @ 0x000000000d60d951
[...]

PR #104280, opened 2026-05-07
Avoid quoting `Logical error` text verbatim in test 04208 comments

groeneai:
The stress-test log scanner (`ci/jobs/scripts/log_parser.py`) uses
`rg --text -A 10 -o 'Logical error.*'` against the server log. Test
`04208_iceberg_disk_with_cache_no_bad_cast.sql` previously included
the literal exception string in its SQL comments. When such a query
fails (it is expected to fail at table-resolution time), the full
query — including those comments — is echoed into the
`<Error> executeQuery` log entry. If a later, completely unrelated
fatal error occurs in the same stress-test run, the scanner picks up
the SQL-comment occurrence first and uses it as the failure name,
even though the actual stack trace (and `STID`) belong to a different
bug.

Three master/PR runs on 2026-05-07 hit this false positive (all
attributed in CI as `Logical error: 'Bad cast from type DB::CachedObjectStorage to DB::S3ObjectStorage'`):

- `Stress test (experimental, serverfuzz, arm_tsan)` on master
  (commit 69167ef) — actual abort was a `chassert(false)` in
  `paranoidCheckForCoveredPartsInZooKeeperOnStart`,
  reported as `STID 2508-5dc3`.
- `Stress test (arm_asan_ubsan)` on PR #100377 — actual abort was
  `Not-ready Set is passed as the second argument for function 'in'`
  in `FunctionIn::executeImpl`, reported as `STID 0250-4e52`.
- `Stress test (amd_msan)` on PR #93037 — same `Not-ready Set`
  family, reported as `STID 0250-41a5`.

The original `Bad cast` bug that this test guards against is
genuinely fixed by #104258 — the unwrap loop in
`S3StorageParsedArguments::fromDisk` is in place and none of these
three runs actually aborted in that code path.

This PR reworks the comments in 04208 so they describe the bug
without containing the verbatim `Logical error: '...'` pattern that
the scanner greps for, and adds a `NOTE` explaining why the verbatim
text is intentionally avoided. The SQL queries and expected
`serverError` codes are unchanged.

A follow-up to harden `log_parser.py` itself (for example, by
requiring a real log-line prefix before matching `Logical error.*`)
would address the same class of false positive for any future test
that mentions an exception text in a comment, but is intentionally
out of scope here to keep this PR minimal.

CI report (master arm_tsan): https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=69167ef25d21358fd8f171e593849ea73195dd1b&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_tsan%29

Related: #89300, #104258 (the original fix this test guards).

### Version info
- Merged into: `26.5.1.401`

### DIFF
--- a/tests/queries/0_stateless/04208_iceberg_disk_with_cache_no_bad_cast.sql
+++ b/tests/queries/0_stateless/04208_iceberg_disk_with_cache_no_bad_cast.sql
@@ -4,20 +4,25 @@
 
 -- Regression test for https://github.com/ClickHouse/ClickHouse/issues/89300
 -- Calling a data-lake table function with `SETTINGS disk = '<disk>'` used to throw
--- `Logical error: 'Bad cast from type DB::CachedObjectStorage to DB::S3ObjectStorage'`
+-- a bad-cast assertion (from the wrapping decorator down to `S3ObjectStorage`)
 -- whenever the disk wrapped its underlying object storage (e.g. cache layer).
--- The same shape of bug previously affected `DB::PlainObjectStorage<DB::S3ObjectStorage>`
+-- The same shape of bug previously affected `PlainObjectStorage` over `S3ObjectStorage`
 -- before that decorator was removed in #90580.
 --
 -- The fix unwraps decorator object storages via `IObjectStorage::getUnderlying`
 -- before casting to the concrete `S3ObjectStorage` in `S3StorageParsedArguments::fromDisk`.
 --
 -- We do not exercise reading any real iceberg data here; the assertion is purely
--- that argument parsing does not crash with a logical error. The query itself is
--- expected to fail at table-resolution time with an S3/iceberg error code.
+-- that argument parsing does not abort. The query itself is expected to fail at
+-- table-resolution time with an S3/iceberg error code.
+--
+-- NOTE: this comment intentionally avoids quoting the exception text verbatim,
+-- because the stress-test log scanner greps for that pattern in the server log
+-- and a literal copy in the query log produces false-positive failure reports
+-- (the SQL comment is echoed into the `<Error> executeQuery` entry on failure).
 
--- Cached disk (`CachedObjectStorage` wrapping `S3ObjectStorage`) — the bug from
--- BuzzHouse fuzzer hit on PR #94148 ("Bad cast from CachedObjectStorage to S3ObjectStorage").
+-- Cached disk (`CachedObjectStorage` wrapping `S3ObjectStorage`) — the bug
+-- originally reported from a BuzzHouse fuzzer hit on PR #94148.
 SELECT 1 FROM icebergS3('no_such_table_for_89300_cache', 'Parquet', 'a Int',
                         SETTINGS disk = 's3_cache')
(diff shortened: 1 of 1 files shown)


groeneai:
**Pre-PR validation gate (per worker policy):**

(a) **Deterministic repro?** — Not applicable; this is a CI-parser false positive, not a server bug. Reproduction conditions are: a stress-test run in which `04208_iceberg_disk_with_cache_no_bad_cast.sql` ran (its query — including the SQL comments — was logged into `<Error> executeQuery`) **and** any unrelated fatal logical error occurred later in the same run. Three concrete instances are linked in the PR description.

(b) **Root cause explained?** — Yes. The stress-test scanner runs `rg --text -A 10 -o 'Logical error.*' <server_log>` (`ci/jobs/scripts/log_parser.py:111`) and uses the first match as the failure name. Because the SQL parser echoes the full query text (including SQL comments) into `<Error> executeQuery: ... (in query: ...)` log entries, any test that contains the literal substring `Logical error: '...'` in a comment becomes a permanent attractor for the scanner whenever an unrelated fatal error happens in the same run.

(c) **Fix matches root cause?** — Yes. Removing the verbatim quoted exception text from the comments removes the substring that the scanner can match. The other tests that have similar patterns (e.g. `03822_not_ready_set_explain_pipeline.sql`, `04035_create_as_select_intersect_settings.sql`) are out of scope here — they are theoretical risks but have not produced false positives in the recent CIDB window.

(d) **Test intent preserved?** — Yes. The SQL queries, the expected `serverError` codes, and the test tags are unchanged. Only the explanatory comments were reworded; an explicit `NOTE` was added explaining why the verbatim text is intentionally avoided.

(e) **Both directions demonstrated?** — Logically:
  - Before: the matched substring exists in the test file → scanner can match it inside the query log when the test runs and another error occurs later → false positive failure name (the three runs cited).
  - After: the substring is absent → scanner cannot match it → real failures are reported under their own names.
  Local runtime reproduction is impractical because it requires the exact CI MinIO/S3 fixtures plus an unrelated stress-test fatal in the same run. The change is purely textual in SQL comments, so there is no behavioral risk.

(f) **Fix is general, not a narrow patch?** — Partially. This PR is a minimal scoped fix for the immediate failure source. A more general fix would harden `log_parser.py` to require a real log-line prefix (timestamp + log level) before treating a `Logical error.*` match as authoritative — that would protect against the same anti-pattern in any future test. I am happy to follow up with a separate PR on `log_parser.py` if reviewers prefer; keeping it out of this PR was a deliberate scoping choice.

Session: `cron:clickhouse-ci-task-worker:20260507-054500`.

groeneai:
cc @alexey-milovidov — could you take a look? Tiny follow-up to #104258. The bad-cast bug itself is genuinely fixed; this just stops the test's regression-comment from being grepped as if it were a real fatal error in stress runs. Three CI false positives from this morning are linked above (the actual bugs are unrelated chronic STIDs).

alexey-milovidov (approved):

PR #104281, opened 2026-05-07
Segmentation tokenizer for text index

### LINKED ISSUE #103743: segmentation tokenizer, which more suitable for log data search/patterns
### Company or project name

_No response_

### Use case

Efficient text index for log data & search patterns (by some identities, substrings of identities)

tokenizer for text/bf indexes, which will "optimistically" (ie try to detect as many possible variants of entities from log line, even it they will intersect by actual string slices(ie kind of what ngram do, but more data type aware).

Types of entities, which can be extracted:
IPv4,IPv6
Mac
UUID
URI?/domain
Timestamp

Segmentation:

Idea, is that data can be indexed both as individual "minor" token, and as combination of several minor tokens:
For ip like: 192.168.1.106
[192, 168, 1, 106, 192.168, 168.1, 1.106, 192.168.1, 168.1.106, 192.168.1.106]

_No response_

### Additional context

Related:
https://www.elastic.co/docs/reference/text-analysis/analysis-pathhierarchy-tokenizer
https://help.splunk.com/en/splunk-enterprise/search/search-manual/9.4/search-primer/event-segmentation-and-searching#ariaid-title4
https://help.splunk.com/en/splunk-enterprise/get-started/get-data-in/10.2/manage-event-segmentation/about-event-segmentation

scanhex12:
- New Feature

Segmentation tokenizer for text index. This closes https://github.com/ClickHouse/ClickHouse/issues/103743

### DIFF
--- a/src/Interpreters/ITokenizer.cpp
+++ b/src/Interpreters/ITokenizer.cpp
@@ -458,9 +458,9 @@ String SparseGramsTokenizer::getDescription() const
 
 std::pair<size_t, size_t> SegmentationTokenizer::convertPartPositionsToTokensPositions(size_t part_start, size_t part_count) const
 {
-    const size_t token_start = part_starts[part_start];
+    const size_t token_start = minor_part_starts[part_start];
     const size_t last_part = part_start + part_count - 1;
-    const size_t token_end = part_starts[last_part] + part_lengths[last_part];
+    const size_t token_end = minor_part_starts[last_part] + minor_part_lengths[last_part];
     return {token_start, token_end - token_start};
 }
 
@@ -471,22 +471,30 @@ void SegmentationTokenizer::buildStateIfNeeded(const char * data, size_t length)
 
     previous_data = data;
     previous_len = length;
-    part_starts.clear();
-    part_lengths.clear();
+    major_part_index.clear();
+    minor_part_starts.clear();
+    minor_part_lengths.clear();
+    
     current_part_i = 0;
     current_part_j = 1;
 
     size_t part_begin = 0;
+    size_t cur_major_part_index = 0;
     for (size_t i = 0; i <= length; ++i)
     {
-        if (i == length || isSplitChar(data[i]))
+        if (i == length || isMajorSplitChar(data[i]) || isMinorSplitChar(data[i]))
         {
             if (i > part_begin)
             {
-                part_starts.push_back(part_begin);
-                part_lengths.push_back(i - part_begin);
+                minor_part_starts.push_back(part_begin);
+                minor_part_lengths.push_back(i - part_begin);
+                major_part_index.push_back(cur_major_part_index);
             }
             part_begin = i + 1;
+            if (isMajorSplitChar(data[i]))
+            {
+                ++cur_major_part_index;
+            }
         }
     }
 }
@@ -495,7 +503,7 @@ bool SegmentationTokenizer::nextInString(const char * data, size_t length, size_
 {
     buildStateIfNeeded(data, length);
 
-    const size_t n_parts = part_starts.size();
+    const size_t n_parts = minor_part_starts.size();
 
     if (current_part_i >= n_parts)
         return false;
@@ -505,7 +513,7 @@ bool SegmentationTokenizer::nextInString(const char * data, size_t length, size_
     token_length = tl;
 
     ++current_part_j;
-    if (current_part_j > n_parts)
+    if (current_part_j > n_parts || major_part_index[current_part_i] != major_part_index[current_part_j - 1])
     {
         ++current_part_i;
         current_part_j = current_part_i + 1;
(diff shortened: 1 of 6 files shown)


src/Interpreters/ITokenizer.h
```diff
@@ -362,6 +363,47 @@ struct SparseGramsTokenizer final : public ITokenizerHelper<SparseGramsTokenizer
...
+    std::vector<String> compactTokens(const std::vector<String> & tokens) const override;
+
+    bool nextInStringLike(const char * data, size_t length, size_t & pos, String & token) const override;
+    bool supportsStringLike() const override { return true; }
+    void substringToBloomFilter(const char * data, size_t length, BloomFilter & bloom_filter, bool is_prefix, bool is_suffix) const override;
+    void substringToTokens(const char * data, size_t length, std::vector<String> & tokens, bool is_prefix, bool is_suffix) const override;
+private:
+    std::pair<size_t, size_t> convertPartPositionsToTokensPositions(size_t part_start, size_t part_count) const;
+
+    static constexpr const char split_chars[] = {'.', ',', ':', '-', '/', ' '};
```
UnamedRus: Kind of original "idea", which i saw from splunk, is that they have it in 2 levels:
1. there is major separators (`,`, ` `, ...)
2. there are minor separators (`.`, `:`, `:`, ...)

So, string is first being split by major separators, then for each major split, it split by minor, and then it's recombine minor segments together.

rschu1ze:
Allow me to ask @Ergus (instead of @alexbakharew) for a review as he was working on full-text search for a longer while.

Ergus:
Hi @scanhex12 

I just added this PR to my TODO list ;). To start with I would request to add some documentation for this new tokenizer in order to understand better how it works and when it is useful for (the last part I saw in the opened issue, but we need to put it all together)

rschu1ze:
@scanhex12 https://github.com/ClickHouse/ClickHouse/pull/104281#issuecomment-4486882902 ^^

Ergus:
Hi @scanhex12 

Are we continuing with this?

Ergus (changes requested):
I am looking into this again and I have a main issue with this as implemented (or intended)

From the original description in the feature request I see that:

`192.168.1.106 -> [192, 168, 1, 106, 192.168, 168.1, 1.106, 192.168.1, 168.1.106, 192.168.1.106]`

IIUC correctly that is the combination that grows with the number of sub-tokens with a quadratic complexity: N*(N+1)/2 

An IPV6 address will produce 36 entries in the dictionary...

In this case the index size increase could affect the overall performance more than just getting some false positive in granules exclusion.

What if we rethink the approach a little bit now that we have the positions experimental feature? With position information we won't need to store all the combinations to get correct full filtering.

The trade off is that positions already increase the index size significantly and is disabled by default.

@rschu1ze WDYT?

src/Interpreters/ITokenizer.cpp
```diff
@@ -454,6 +456,183 @@ String SparseGramsTokenizer::getDescription() const
...
+    {
+        --length;
+    }
+    if (length > 0)
+        stringToTokens(data, length, tokens);
+}
+
+std::vector<String> SegmentationTokenizer::compactTokens(const std::vector<String> & tokens) const
+{
+    std::unordered_set<String> result;
```
Ergus, line 606: This has changed on master to use the memory tracking API

UnamedRus:
I don't think that it that much worse, than other already existing tokenizers which we can use in text indexes

```
SELECT
    ngrams('192.168.1.106', 3),
    ngrams('2001:db8:85a3::8a2e:370:7334', 3) AS ip,
    sparseGrams('2001:db8:85a3::8a2e:370:7334') AS ip2,
    length(ip),
    length(ip2)
FORMAT Vertical

Query id: ee800a00-76ea-4d4c-ac12-1186182e974a

Row 1:
──────
ngrams('192.168.1.106', 3): ['192','92.','2.1','.16','168','68.','8.1','.1.','1.1','.10','106']
ip:                         ['200','001','01:','1:d',':db','db8','b8:','8:8',':85','85a','5a3','a3:','3::','::8',':8a','8a2','a2e','2e:','e:3',':37','370','70:','0:7',':73','733','334']
ip2:                        ['200','001','01:','1:d','01:d',':db','db8',':db8','01:db8','b8:','8:8','b8:8',':85','85a','5a3','85a3',':85a3','b8:85a3','01:db8:85a3','001:db8:85a3','a3:','3::','::8',':8a','8a2',':8a2','::8a2','3::8a2','a3::8a2','001:db8:85a3::8a2','a2e','2e:','a2e:','e:3',':37','370',':370','70:',':370:','0:7',':73','0:73',':370:73','733',':370:733','e:370:733','a2e:370:733','334','a2e:370:7334']
length(ip):                 26
length(ip2):                49
```

src/Interpreters/ITokenizer.cpp
```diff
@@ -454,6 +456,183 @@ String SparseGramsTokenizer::getDescription() const
...
+}
+
+bool SegmentationTokenizer::nextInString(const char * data, size_t length, size_t & __restrict /*pos*/, size_t & __restrict token_start, size_t & __restrict token_length) const
+{
+    buildStateIfNeeded(data, length);
+
+    const size_t n_parts = minor_part_starts.size();
+
+    if (current_part_i >= n_parts)
+        return false;
```
Ergus, line 509: IIUC comparing with `SparseNGrams` I think that in this condition you need to reset:

```
previous_data = nullptr;
previous_len = 0;
```

Otherwise you could miss some rows because `buildStateIfNeeded` can early-returns on a cache hit without re-initializing `current_part_i`

tests/queries/0_stateless/04202_segmentation_tokenizer.sql
```diff
@@ -0,0 +1,200 @@
...
+    (2, '10.0.0.1'),
+    (3, '172.16.254.1'),
+    (4, 'plaintext'),
+    (5, '192.168.2.1');
+
+SELECT 'exact match';
+SELECT id FROM tab WHERE ip = '192.168.1.106' ORDER BY id;
+
+SELECT 'prefix segment';
+SELECT id FROM tab WHERE ip LIKE '192.168.%' ORDER BY id;
```
Ergus, line 27: Will this match equivalent text patterns like `192.168`?

What about `LIKE '%92.16%'`

Or other functions like `startWith('192.1')`?

Ergus (changes requested):

Ergus:
> I don't think that it that much worse, than other already existing tokenizers which we can use in text indexes

yes, the ngrams is not actually the best example on how to use the text index. Somehow we still support it for some corner cases and legacy, but now really recommended in most of the cases precisely because of the dictionary size grow.

UnamedRus:
> not actually the best example on how to use the text index

I don't like alphatokens* ones either (for log style data).

They might work OK for normal text data, but in logs, people tends to have/use those symbols (by which we split) as part of ids (ip, uuid, mac, whatever custom id style they managed to invent).
And indexing those values sliced by alpha tokens is just a waste of space (both disk and "search space" for useless/too small/broad tokens). It can be argued that position information should fix that, but it's almost doubling index size, and imo it's still too much useless information to our relatively simple goal: search by various kind of ids (which actually represent relatively small byte/length size of total log data (which include a lot of text for humans, but irrelevant for our search requirements))

I think, we better be with (may be with better layout) bloom filter style index instead.

tests/queries/0_stateless/04202_segmentation_tokenizer.sql
UnamedRus: > What about LIKE '%92.16%'

IMO, main goal of it to solve problem of "bad" query/lookup pattern.
When someone write search by IP (192.168.1.43) but in bad way.

```
ip LIKE '%192.168.1.43%' -- we "at least" get 168.1 as useful token

ip LIKE '%192.168.1.43 %' -- we "at least" get 168.1.43 as token

ip LIKE '% 192.168.1.43%' -- we "at least" get 192.168.1 as token
```

Which marginally, but better than

```
SELECT tokensForLikePattern('%192.168.1.43%'), tokensForLikePattern('% 192.168.1.43%'), tokensForLikePattern('%192.168.1.43 %'),

['168','1']	['192','168','1']	['168','1','43']
```

For IPv6 or ids with more delimeters, advantage of "saved" tokens should be better/bigger

PR #104282, opened 2026-05-07
UNIQUE KEY: index cache (7/N)

murphy-4o:
Part of #103486

Introduces a ClickHouse CacheBase adapter for the RocksDB block cache, including Context and Server registration along with corresponding metrics. The adapter implements the minimal rocksdb::Cache interface required for `BlockBasedTableOptions::block_cache`, delegating storage to `CacheBase`. This ensures that memory accounting is correctly integrated into `system.caches` via `UniqueKeyIndexCacheBytes` and `UniqueKeyIndexCacheEntries`.

The adapter narrows several `rocksdb::Cache` surfaces to eliminate redundant shadow bookkeeping on top of `CacheBase`. An audit of the RocksDB source code (specifically `contrib/rocksdb/table/ and contrib/rocksdb/db/`) confirms that BlockBasedTable does not rely on these narrowed paths for correctness when using a user-provided block_cache. ClickHouse maintains full ownership of the cache configuration and never enables strict mode.

### Version info
- Merged into: `26.6.1.388`


src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,335 @@
...
+    {
+        const auto & pinned = pin->entry;
+        const size_t before = backing->count();
+        std::function<bool(const UInt128 &, const std::shared_ptr<UniqueKeyIndexCacheEntry> &)>
+            pred = [&](const UInt128 & k, const std::shared_ptr<UniqueKeyIndexCacheEntry> & v)
+            {
+                return k == pin->key && v == pinned;
+            };
+        backing->remove(pred);
+        erased = backing->count() < before;
```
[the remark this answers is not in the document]
murphy-4o: Fixed in 8580c38164f. `Release` now captures the match inside the predicate (which `CacheBase::remove` invokes under the cache lock), so the return value is no longer derived from two separate `count()` snapshots that could be perturbed by concurrent insert/erase on unrelated keys.

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,339 @@
...
+    {
+        auto * pin = new HandlePin{entry, keyOf(key), std::atomic<int32_t>{1}};
+        *handle = fromPin(pin);
+    }
+
+    return ROCKSDB_NAMESPACE::Status::OK();
+}
+
+ROCKSDB_NAMESPACE::Cache::Handle *
+UniqueKeyIndexCache::CreateStandalone(
```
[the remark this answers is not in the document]
murphy-4o: Fixed in 31498af1e87. `CreateStandalone` now honors RocksDB's strict-capacity contract: refuses (returns nullptr) when `strict_capacity_limit` is set, the requested charge exceeds the cap, and `allow_uncharged=false`. With `allow_uncharged=true` (the typical RocksDB caller) the standalone handle is returned uncharged, consistent with `rocksdb::Cache`'s spec — `CacheBase` has no native standalone-tracked path for proper charging, so we accept the uncharged-only semantics rather than route standalone allocations through a tracked structure for now. Added `UniqueKeyIndexCache.CreateStandaloneHonorsStrictCapacityLimit` covering the three cases (strict + over-cap + !allow_uncharged → nullptr; over-cap + allow_uncharged → handle; within cap → handle).

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,351 @@
...
+{
+    if (!handle)
+        return 0;
+    const auto * pin = asPin(handle);
+    return pin->entry->charge + UniqueKeyIndexCacheEntryWeight::OVERHEAD;
+}
+
+size_t UniqueKeyIndexCache::GetPinnedUsage() const
+{
+    /// Approximate: we don't track this separately. Returning 0 is safe for the
```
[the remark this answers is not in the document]
murphy-4o: Fixed in ecea231fccd.

`UniqueKeyIndexCacheEntry` now carries a per-entry `pin_count`. `Insert`/`Lookup`/`CreateStandalone` add the entry's charge to `pinned_charge_total` only on the 0→1 `pin_count` transition; `Release` final-ref subtracts only on 1→0. Per-entry semantics match `LRUCacheShard::GetPinnedUsage` (counts pinned entries once, not per handle).

Bonus fixes surfaced in the local cursor-agent review loop:
1. `CreateStandalone` in the strict + over-cap + `allow_uncharged=true` branch now sets `effective_charge=0`; `GetCharge`, `GetUsage(handle)`, and `GetPinnedUsage` all consistently report 0 for uncharged handles, per `rocksdb::Cache` spec.
2. `Release(handle, erase=true)` predicate now also requires `v->pin_count.load() == 0` under the backing lock — narrows the Lookup-vs-erase race that an earlier review pass surfaced. A residual window remains where `Lookup`'s `backing->get` returns the entry shared_ptr but its `pin_count.fetch_add(1)` happens after the predicate's load; the orphaned handle keeps the entry alive via `shared_ptr` and `pinned_charge_total` balances at equilibrium, so the divergence is bounded to cache-state metric accuracy. Fully closing it requires `CacheBase` to expose joint get-and-mutate under one lock and is deferred.

Tests (gtest `unit_tests_dbms --gtest_filter=UniqueKeyIndexCache.*`):
- `GetPinnedUsageTracksPinnedEntries` — eviction-while-pinned (the original bug); two handles on one entry count once.
- `GetPinnedUsageRefDoesNotDoubleCount` — `Ref`/`Release` pairing.
- `CreateStandaloneUnchargedHasZeroChargeAndUsage` — uncharged contract.

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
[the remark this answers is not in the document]
murphy-4o: Resolved in 09c3f142045 — this adapter now explicitly declines the strict-capacity contract. `HasStrictCapacityLimit()` is hardcoded `false` and `SetStrictCapacityLimit(...)` is a no-op. Per `rocksdb::Cache::CreateStandalone` spec: *"if `allow_uncharged==true` or `strict_capacity_limit=false`, the operation always succeeds and returns a valid Handle."* With strict mode permanently off here, `CreateStandalone` always returns a charged handle and `allow_uncharged` is correctly unused. Tests `StrictCapacityLimitIsUnsupported` and `CreateStandaloneAlwaysSucceedsWhenNonStrict` pin the contract.

Rationale for declining rather than honoring strict mode: doing so on top of `CacheBase` would require shadow pin bookkeeping plus a joint get-and-mutate primitive that `CacheBase` does not expose. This cache is ClickHouse-configured end-to-end and strict mode is never enabled on it; RocksDB's `BlockBasedTable` itself never calls `SetStrictCapacityLimit` or `CreateStandalone(allow_uncharged=false)` on a user-provided `block_cache` (verified by grepping `contrib/rocksdb/table/` and `contrib/rocksdb/db/`).

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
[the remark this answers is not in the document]
murphy-4o: Resolved in 09c3f142045 — Option B declines the strict-capacity contract (see [#3214240064 reply](https://github.com/ClickHouse/ClickHouse/pull/104282#discussion_r3214240064)). With `HasStrictCapacityLimit()` hardcoded `false`, admission is never gated on pinned bytes, so `GetPinnedUsage()` returning `0` is documented behavior rather than a correctness gap. Pinned handles still keep entries alive past backing eviction via `shared_ptr`, but no correctness path consults pinned bytes; the only rocksdb consumer of this surface is `db/internal_stats.cc` (stats reporting), which degrades to `0` gracefully. Matches `MarkCache` / `UncompressedCache`, which sit on the same `CacheBase` substrate and also do not track pinned bytes.

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,371 @@
...
+    Priority /*priority*/,
+    const ROCKSDB_NAMESPACE::Slice & /*compressed*/,
+    ROCKSDB_NAMESPACE::CompressionType /*type*/)
+{
+    /// strict_capacity_limit is honored conservatively: if the requested charge
+    /// already exceeds the cache's max, refuse. CacheBase itself allows going
+    /// over the limit briefly during eviction which is fine for BLOCK cache use.
+    if (strict_capacity_limit.load(std::memory_order_relaxed))
+    {
+        const size_t max = backing->maxSizeInBytes();
```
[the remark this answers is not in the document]
murphy-4o: Resolved in 09c3f142045 — Option B declines the strict-capacity contract (see [#3214240064 reply](https://github.com/ClickHouse/ClickHouse/pull/104282#discussion_r3214240064)). `HasStrictCapacityLimit()` is hardcoded `false` and `SetStrictCapacityLimit(true)` is a no-op (pinned by `StrictCapacityLimitIsUnsupported`), so the "strict mode cannot prevent `pinned + resident` from exceeding capacity" scenario cannot arise because strict mode cannot be enabled.

`Insert` remains unconditional `backing->set(...)`; `CacheBase` SLRU eviction keeps tracked usage under `unique_key_index_cache_size_bytes` (default 1 GiB; `0` disables). Pinned-via-handle entries can transiently exceed the configured cap (same property as `MarkCache` / `UncompressedCache` on the same `CacheBase` substrate), but the cap itself is still enforced.

src/Interpreters/Context.cpp
```diff
@@ -3976,6 +3980,72 @@ void Context::clearMarkCache() const
...
+        LOG_DEBUG(shared->log, "Lowered UNIQUE KEY index cache size to {} because the system has limited RAM", formatReadableSizeWithBinarySuffix(size));
+    }
+    const size_t before = shared->unique_key_index_cache->GetCapacity();
+    shared->unique_key_index_cache->setMaxSizeInBytes(size);
+    if (size != before)
+        LOG_INFO(shared->log, "Reconfigured UNIQUE KEY index cache from {} to {}",
+                 formatReadableSizeWithBinarySuffix(before), formatReadableSizeWithBinarySuffix(size));
+#else
+    (void)config;
+    (void)max_cache_size;
```
Michicosun: should we use [[maybe_unused]] modifier?

src/Interpreters/Context.cpp
```diff
@@ -3976,6 +3980,72 @@ void Context::clearMarkCache() const
...
+    (void)max_cache_size_in_bytes;
+    (void)size_ratio;
+#endif
+}
+
+void Context::updateUniqueKeyIndexCacheConfiguration(const Poco::Util::AbstractConfiguration & config, size_t max_cache_size)
+{
+    std::lock_guard lock(shared->mutex);
+
+    if (!shared->unique_key_index_cache)
```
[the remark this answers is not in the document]
murphy-4o: Resolved in 3697ef44cca. `updateUniqueKeyIndexCacheConfiguration` now constructs the cache on the first reload that requests a non-zero size, using `unique_key_index_cache_policy` and `unique_key_index_cache_size_ratio` from the same config block — so `unique_key_index_cache_size_bytes` is a reversible knob rather than a one-way disable for the process lifetime. Startup path is unchanged: explicit `size=0` at startup still leaves `unique_key_index_cache` as `nullptr` so a permanently-disabled adapter has zero overhead until reload re-enables it.

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,341 @@
...
+    if (!handle)
+        return false;
+    auto * pin = asPin(handle);
+
+    /// Drop the entry's table slot; live HandlePins keep the entry alive
+    /// via `shared_ptr`.
+    bool erased = false;
+    if (erase_if_last_ref)
+    {
+        backing->remove(pin->key);
```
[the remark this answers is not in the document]
murphy-4o: Resolved in 54ec9b6c78f by walking back the unconditional `backing->remove(pin->key)` from d75499babaa. `Release(handle, true)` now uses the identity- and last-pin-aware predicate `k == pin->key && v == pinned && pinned.use_count() == 2`, evaluated under `CacheBase`'s bucket lock so the contributors to `use_count` are stable: table strong ref + this HandlePin = 2; an additional live HandlePin pushes it past 2 and the erase is refused; a replacement under the same key sets `v != pinned` and the erase is refused. Both contract gaps you flagged (old-handle-evicting-replacement and over-reported `erased`) are closed: `ReleaseEraseLastRefRespectsOtherLiveHandle` and `ReleaseEraseIdentityAwareDoesNotEvictReplacement` are restored to pin the two cases.

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,360 @@
...
+        const auto & pinned = pin->entry;
+        using RemovePred = std::function<bool(const UInt128 &, const std::shared_ptr<UniqueKeyIndexCacheEntry> &)>;
+        RemovePred pred = [&](const UInt128 & k, const std::shared_ptr<UniqueKeyIndexCacheEntry> & v)
+        {
+            const bool matches = (k == pin->key && v == pinned && pinned.use_count() == 2);
+            if (matches)
+                erased = true;
+            return matches;
+        };
+        backing->remove(pred);
```
Michicosun: This will basically list all cache, can't it be done via point removal?

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,360 @@
...
+UniqueKeyIndexCache::GetCacheItemHelper(Handle * handle) const
+{
+    return handle ? asPin(handle)->entry->helper : nullptr;
+}
+
+void UniqueKeyIndexCache::ApplyToAllEntries(
+    const std::function<void(const ROCKSDB_NAMESPACE::Slice &, ObjectPtr, size_t, const CacheItemHelper *)> & /*callback*/,
+    const ApplyToAllEntriesOptions & /*opts*/)
+{
+    /// Known contract gap: rocksdb::Cache spec asks the callback to fire for
```
Michicosun: Should we assert that this method will not be called?

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,360 @@
...
+{
+
+/// One entry per cached object. Weight contributed to the backing is
+/// `charge + OVERHEAD`. The destructor invokes the RocksDB-supplied
+/// `helper->del_cb` to destroy `obj` when the entry's last `shared_ptr`
+/// drops (either via backing eviction or via the last HandlePin going
+/// away).
+struct UniqueKeyIndexCacheEntry
+{
+    void * obj = nullptr;
```
Michicosun, line 39: Will it be too expensive to store a real key string in the entry?

src/Interpreters/Context.cpp
murphy-4o: fixed

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
murphy-4o: the IndexCache is `BlockCache` of RocksDB, which is about 32KB per block, so the `UniqueKeyIndexCacheEntry` overhead is fine.
and the `void * obj` points to a RocksDB internal Block, but not a key.

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
murphy-4o: fixed

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
murphy-4o: good catch. fixed by add a new interface `CacheBase::removeIfMatches(key, predicate)`.

it's a little bit tricky here, we need to check the `pinned.use_count()`, because `rocksdb::Handle` can hold a cache entry, stopping the `CacheBase::remove`

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
murphy-4o: the key of a `BlockCache` is 16 bytes, `file-id` + `file-offset`

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,351 @@
...
+    const std::function<void(const ROCKSDB_NAMESPACE::Slice &, ObjectPtr, size_t, const CacheItemHelper *)> & /*callback*/,
+    const ApplyToAllEntriesOptions & /*opts*/)
+{
+    /// Not exercised by RocksDB's `BlockBasedTable` against a user-provided
+    /// `block_cache`; the only consumers (`CacheDumperImpl`, statistics
+    /// walker) degrade to reporting nothing for this cache. We can't honor
+    /// the spec anyway: the backing is keyed by SipHash128 with no public
+    /// iteration API. `chassert` fires in debug builds if a future caller
+    /// shows up; release builds keep the silent no-op.
+    chassert(false && "UniqueKeyIndexCache::ApplyToAllEntries is not implemented");
```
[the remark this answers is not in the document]
murphy-4o: fixed

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,352 @@
...
+    const ROCKSDB_NAMESPACE::Slice & key,
+    ObjectPtr obj,
+    const CacheItemHelper * helper,
+    size_t charge,
+    Handle ** handle,
+    Priority /*priority*/,
+    const ROCKSDB_NAMESPACE::Slice & /*compressed*/,
+    ROCKSDB_NAMESPACE::CompressionType /*type*/)
+{
+    auto entry = std::make_shared<UniqueKeyIndexCacheEntry>();
```
Michicosun, line 150: Shouldn't we set the allocator during insert from the memory_allocator() function?

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
```diff
@@ -0,0 +1,352 @@
...
+    /// RocksDB contract). Forwarded to `helper->del_cb` on destruction.
+    void * allocator = nullptr;
+
+    UniqueKeyIndexCacheEntry() = default;
+    ~UniqueKeyIndexCacheEntry()
+    {
+        /// `helper->del_cb` may be nullptr for `kNoopCacheItemHelper`-style
+        /// entries; guard accordingly.
+        if (helper && helper->del_cb && obj)
+            helper->del_cb(obj, reinterpret_cast<ROCKSDB_NAMESPACE::MemoryAllocator *>(allocator));
```
Michicosun: At least here, I suppose, the assert for the allocator should be placed.

src/Storages/MergeTree/UniqueKey/UniqueKeyIndexCache.cpp
murphy-4o: dropped the field, use `nullptr` as allocator instead. we don't need it currently

Michicosun (approved):

azat:
@murphy-4o looks like new test is [flaky](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=106802&sha=19b12cf2890c1c0fa0deb846be9bf7533d1b8c50&name_0=PR&name_1=Unit%20tests%20%28asan_ubsan%29), can you PTAL? https://pastila.nl/?04c5ec7b/5e60f8fc69db94e39a6a40d51e4ccc08#s0xPJMEORBfPW8tHzNd1XA==GCM

murphy-4o:
The gtest added here, `UniqueKeyIndexCache.InsertLookupAndHitMissCounters`, turned out to be flaky:

```
src/Storages/MergeTree/tests/gtest_unique_key_index_cache.cpp:80: Failure
Expected equality of these values:
  ProfileEvents::global_counters[ProfileEvents::UniqueKeyIndexCacheHits].load() - hits_before
    Which is: 0
  1u
```

Root cause: the test reads deltas from `ProfileEvents::global_counters`, but `ProfileEvents::increment` (inside `UniqueKeyIndexCache::Lookup`) routes through `CurrentThread::getProfileEvents`, which only returns `global_counters` when no `ThreadStatus` is attached to the current thread. All gtests in `unit_tests_dbms` run on one main thread, and some tests (e.g. `gtest_filecache`) attach a `ThreadStatus` to it that persists for the rest of the binary. After that, the increments land in a thread-local subtree that does not propagate to `global_counters`, so the deltas read `0`. CI shuffles gtests (`--gtest_shuffle`), so the failure is order-dependent.

Fix in https://github.com/ClickHouse/ClickHouse/pull/106839 - read the deltas from `CurrentThread::getProfileEvents` (the exact `Counters` object the increment targets) instead of `global_counters`, which also removes the now-unnecessary `WITH_COVERAGE` special case.

PR #104293, opened 2026-05-07
check_ci.py: surface gh error when auto-merge fails

Algunenano:
Previously when `gh pr merge --auto` returned non-zero, `Shell.check` silently returned `False` and the entire merge block was skipped without any output. The user saw only the "Add PR to merge queue (y/n)" prompt and then nothing, with the PR not actually queued.

This change runs the `gh` command with `verbose=True` so its stderr is streamed, and adds an early-exit branch that prints the exact command to retry manually. This typically happens when `mergeStateStatus` is `UNKNOWN` (GitHub is still computing mergeability after a recent push).

...

### Version info
- Merged into: `26.5.1.403`

### DIFF
--- a/ci/jobs/scripts/check_ci.py
+++ b/ci/jobs/scripts/check_ci.py
@@ -1040,41 +1040,50 @@ def main():
         mergeable_check_status, sha=head_sha
     )
 
-    if Shell.check(f"gh pr merge {pr_number} --auto --repo ClickHouse/ClickHouse"):
-        # Give GitHub a moment to process auto-merge and update merge state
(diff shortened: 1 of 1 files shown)


alexey-milovidov:
FYI, `check_ci` is a very bad script, as it allows merging red (broken) PRs. We should avoid using it.

Algunenano:
> FYI, check_ci is a very bad script, as it allows merging red (broken) PRs. We should avoid using it.

It helps us report and search existing CI issues, so I find it useful when used correctly

PR #104303, opened 2026-05-07
Fix MULTIPLE_EXPRESSIONS_FOR_ALIAS on `SELECT *` over JOIN under parallel replicas

devcrafter:

Fix `MULTIPLE_EXPRESSIONS_FOR_ALIAS` when serializing a `QueryNode` whose projection has duplicated column names (e.g. `SELECT *` over a JOIN) for distributed dispatch.

---

Partial progress on #74324.

`QueryNode::toASTImpl` always emits `AS <name>` for every projection. After `createUniqueAliasesIfNecessary` qualifies tables as `__tableN`, a `SELECT *` over a JOIN produces `__table1.a AS a, __table3.a AS a` — duplicate alias, two different expressions — which the analyzer rejects when re-resolving the dispatched AST on a remote replica.

Fix: skip the alias only when the projection column name is duplicated within the same projection.

Verification (debug build, parallel replicas enabled):
- new test `04209_pr_join_using_alias_collision`: 100/100
- `00597_push_down_predicate_long`: 100/100 (removed from `parallel_replicas_blacklist.txt`)

`02559_multiple_read_steps_in_prewhere` was also listed under #74324 but fails on a different `PREWHERE` alias-substitution pattern; left blacklisted.


src/Analyzer/QueryNode.cpp
```diff
@@ -499,20 +499,41 @@ ASTPtr QueryNode::toASTImpl(const ConvertToASTOptions & options) const
...
+            if (!ast_with_alias)
+                continue;
+
+            const auto & column_name = projection_columns[i].name;
+            const auto * current_column = projection_nodes[i]->as<ColumnNode>();
+            if (current_column)
+            {
+                auto [it, inserted] = first_column_by_name.try_emplace(column_name, current_column);
+                if (!inserted && it->second->getColumnSourceOrNull() != current_column->getColumnSourceOrNull())
+                    continue;
```
novikd: It does not look like a good solution because it does not fix the query sent, but it "supresses" an error on the shard.
What is the reason that different expressions have the same aliases?

novikd (commented):
It'd be better to find out the reason for such kind of clushes.

src/Analyzer/QueryNode.cpp
devcrafter: It doesn't suppress error on the shard, it doesn't allow generating the same aliases for different columns.

devcrafter: The another question - why should we generate here aliases?

PR #104312, opened 2026-05-07
fix(parser): parse values subscript as identifier

### LINKED ISSUE #104310: Parser: column named `values` misparsed as SQL-standard VALUES keyword in `(values[…])` expressions — silent wrong results / ILLEGAL_TYPE_OF_ARGUMENT
### Describe what's wrong

A column literally named `values` (case-insensitive) is misparsed as the `VALUES` keyword whenever `values[…]` appears immediately after `(`. The parser builds an `SQLStandardValues(…)` table source instead of a Map subscript on the column, producing one of two behaviors depending on the surrounding expression:

- **Silent wrong results** — the simple subscript form returns the array literal that was treated as a `VALUES` row, instead of the map value:
  ```sql
  SELECT (values['a']) AS x FROM (SELECT map('a', 42) AS values);
  -- master:  ['a']     ← wrong
  -- 26.2.x:  42        ← correct
  ```
- **Hard error** — when the body looks like an arithmetic expression on a Map subscript:
  ```sql
  SELECT (values['a'] - 1) AS x FROM (SELECT map('a', 42) AS values);
  -- master:  Code: 43. ILLEGAL_TYPE_OF_ARGUMENT — Array(String) and UInt8 to function `minus`
  --         "In scope (SELECT * FROM SQLStandardValues(['a'] - 1)) AS x"
  -- 26.2.x:  41
  ```

The trigger is purely syntactic:

- column name `values` / `VALUES` (case-insensitive),
- subscript form `values[…]` is the **first** token after `(`,
- the rest of the parenthesized group is then misparsed as a SQL standard `VALUES` row.

`(values)` without `[…]`, or `values[…]` without surrounding parens, parses correctly. Renaming the column to anything other than `values` also avoids the bug.

Yes

### How to reproduce

Reproduces on ClickHouse `26.4.1.1736` and on current master (tested against `26.5.1.111` from `d3f96ed43f53462b3b68f50784bc45ab5c7e00c6`). Works correctly on `26.2.1.303`.

```sql
-- Hard error on 26.4+ / master, returns 41 on 26.2
SELECT (values['a'] - 1) AS x FROM (SELECT map('a', 42) AS values);

-- Silent wrong result on 26.4+ / master (returns ['a']), returns 42 on 26.2
SELECT (values['a']) AS x FROM (SELECT map('a', 42) AS values);
```

Works around the bug:

```sql
-- 1. Drop outer parens
SELECT  values['a'] - 1 AS x FROM (SELECT map('a', 42) AS values);
-- 2. Rename or alias the column to anything other than `values`
SELECT (vals['a'] - 1)  AS x FROM (SELECT map('a', 42) AS vals);
-- 3. Wrap the column in any function before subscripting
SELECT (assumeNotNull(values)['a'] - 1) AS x FROM (SELECT map('a', 42) AS values);
```

### Expected behavior

`(values['a'] - 1)` should parse as `((values['a']) - 1)` — i.e. Map subscript followed by subtraction — not as a SQL-standard `VALUES` row whose body is `(['a'] - 1)`. The `VALUES` keyword should only be applied when the next token is a plausible row literal, not a column-subscript expression.

```
Code: 43. DB::Exception: Illegal types Array(String) and UInt8 of arguments of function minus:
While processing ['a'] - 1:
While processing (SELECT * FROM SQLStandardValues(['a'] - 1)) AS x. (ILLEGAL_TYPE_OF_ARGUMENT)
```

### Additional context

Bisected with `utils/auto-bisect/bisect.sh` against pre-built CI binaries:

- **Good:** `c484b5cd98c76bffbe6d4b1d377ddefdef0d17b5` (Mar 25 22:58 UTC) — `41`
- **Bad:**  `f4927b8e226a1a90d419f19e711dfb9c0426cc86` (Apr 15 23:46 UTC) — `Code: 43`

**First bad commit:** the merge of #100143 *Support SQL standard VALUES clause as a table expression*.

The PR introduced a parser branch that eagerly matches `VALUES` after `(` (`src/Parsers/ExpressionElementParsers.cpp`, around `ParserKeyword(Keyword::VALUES).ignore(pos, expected)`). There is no lookahead distinguishing `(VALUES (1, 'a'), …)` from `(values[<map-key>] <op> …)` when a column happens to be named `values`. An experimental gate setting `allow_experimental_sql_standard_values_clause` was added during the PR (commit `d337e408…`) but removed again before merge (commit `954d1d30…`, "Refactor: template TableFunctionValues, remove experimental setting, …"), so master never had a working escape hatch. Setting `enable_analyzer = 0` does not help — it is a parser-level issue, not analyzer-level.

Real-world impact: customer queries like
```sql
WITH t AS (
  SELECT (values['ns_203'] - values['ns_201']) AS dt FROM analytics.timestamper_logs WHERE …
) SELECT quantile(0.9)(dt) FROM t
```
that worked unchanged on 26.2 fail on 26.4 with `ILLEGAL_TYPE_OF_ARGUMENT`. The `values` column name is common (it's the natural name for `Map(String, Int64)` payloads) so the bug is broadly reachable.

### Version info
- Resolved by: #104312
- Merged into: `26.5.1.422`
- Backported to: `26.4.3.2`

Daedalus-Icarus:
Fixes https://github.com/ClickHouse/ClickHouse/issues/104310

A column named `values` could be misparsed as the SQL-standard `VALUES` keyword inside parenthesized subscript expressions such as `(values['a'])`. The parser now only treats parenthesized `VALUES` as a SQL-standard `VALUES` table expression when the keyword is followed by a row constructor, while ordinary column subscripts keep parsing as expressions. A stateless regression test covers both the map subscript case and a valid SQL-standard `VALUES` clause.

Fix parsing of parenthesized subscript expressions on columns named `values`, for example `(values['a'])`, so they are no longer interpreted as SQL-standard `VALUES` table expressions.

- [ ] Documentation is not required

### Version info
- Merged into: `26.5.1.422`
- Backported to: `26.4.3.2`


Daedalus-Icarus:
Hi @fm4v  
Could you please review my PR and merge?

PR #104314, opened 2026-05-07
Add test: `recursiveRemoveLowCardinality` equivalence stripping in `tryDirectJoin` is untested across the entire suite

clickgapai:

Adds test coverage for 1 untested code path(s), found during automated review of [PR #103928](https://github.com/ClickHouse/ClickHouse/pull/103928).
 That PR PR adds a type check in `tryDirectJoin` (`src/Planner/PlannerJoins.cpp`) so that `DirectKeyValueJoin` is only chosen when the right join key's plan type matches the storage primary key type, modulo `Nullable` and `LowCardinality` wrappers. Previously the legacy `JOIN ... USING` planner inserted a 'C

**1. `recursiveRemoveLowCardinality` equivalence stripping in `tryDirectJoin` is untested across the entire suite**
  `src/Planner/PlannerJoins.cpp:1108`, `src/Planner/PlannerJoins.cpp:1110`
  **Risk:** Call chain: `chooseJoinAlgorithm` -> `tryCreateJoin` -> `tryDirectJoin` (PlannerJoins.cpp:1175-1179). At line 1110 the PR computes `storage_primary_key_type = removeNullable(recursiveRemoveLowCardinal
  **What's unique vs PR tests:** The PR's test (`04201_direct_join_type_mismatch_rocksdb.sql`) covers cases where the right key TYPE differs from the storage PK type after stripping (`Nullable(UInt64)` vs `UInt32`, `Int64` vs `UInt32
  **Tags:** `-- Tags: use-rocksdb` — `use-rocksdb`: Test creates an `EmbeddedRocksDB` storage which is only available when the rocksdb feature is built into ClickHouse; matches the convention used by every other rocksdb stateless test including the PR's own 04201_direct_join_type_mismatch_rocksdb.sql.
  [Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/22d4c172-cb46-44fc-837d-9a2b6cf13dcb)

### Version info
- Merged into: `26.5.1.485`

### DIFF
--- a/tests/queries/0_stateless/04209_rocksdb_direct_join_lowcardinality_pk_equivalence.reference
+++ b/tests/queries/0_stateless/04209_rocksdb_direct_join_lowcardinality_pk_equivalence.reference
@@ -1,0 +1,13 @@
+inner join legacy planner result
+a	A
+b	B
+c	C
+inner join legacy planner explain
+Algorithm: DirectKeyValueJoin
+left join legacy planner result
+a	A
+b	B
+c	C
+d	
+left join legacy planner explain
+Algorithm: DirectKeyValueJoin
--- a/tests/queries/0_stateless/04209_rocksdb_direct_join_lowcardinality_pk_equivalence.sql
+++ b/tests/queries/0_stateless/04209_rocksdb_direct_join_lowcardinality_pk_equivalence.sql
@@ -1,0 +1,71 @@
+-- Tags: use-rocksdb
+-- Coverage: exercises the `recursiveRemoveLowCardinality` equivalence stripping in
+-- `tryDirectJoin` at src/Planner/PlannerJoins.cpp:1108-1111. The PR
+-- (STID 3436-4175) added a type check between the right join key and the storage
+-- primary key; the check strips `Nullable` AND `LowCardinality` wrappers on both
+-- sides "to match the equivalence semantics used by getByKeys itself."
+--
+-- The PR's own tests (04201_direct_join_type_mismatch_rocksdb.sql) only exercise
+-- the `removeNullable` path — none use `LowCardinality` wrappers, leaving the
+-- `recursiveRemoveLowCardinality` calls completely untested.
+--
+-- Scenario: storage PK is `LowCardinality(String)`, right join key is plain
+-- `String`. The legacy `JOIN ... USING` planner casts both sides to the
+-- common supertype `String`, so at the time `tryDirectJoin` runs:
+--     right_table_expression_header.key.type == String
+--     storage->getSampleBlock().key.type    == LowCardinality(String)
+--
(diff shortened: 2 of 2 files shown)


groeneai:
LGTM — verified the coverage gap is real and the test is a sound mutation guard.

**Coverage gap verification (real):**

I read our merged regression test `04201_direct_join_type_mismatch_rocksdb.sql` and confirmed all 9 cases use only `Nullable(UInt64)`, `Int64`, `Decimal(18,7)`, or `UInt32` — none touch `LowCardinality`. A grep of all stateless tests for `LowCardinality.*EmbeddedRocksDB` and `PRIMARY KEY.*LowCardinality` against an `EmbeddedRocksDB` engine returns zero matches, so there is no other test exercising this path either. The `recursiveRemoveLowCardinality` calls at `src/Planner/PlannerJoins.cpp:1108` and `:1110` are genuinely untested across the suite.

**Local verification:**

Built debug + ran the new test 20× with full setting randomization → 20/20 passed (~7.9s total). Manually inspected the EXPLAIN output of the legacy-planner INNER JOIN scenario and confirmed `Algorithm: DirectKeyValueJoin` is selected against the `LowCardinality(String)` PK + plain `String` right key, so the test is asserting a meaningful plan invariant.

**Mutation analysis (which line does this guard?):**

- Right-side input post-USING-cast: `String` → after `removeNullable(recursiveRemoveLowCardinality(...))` → `String` (no change, line 1108's call is a no-op for this scenario)
- Storage-side input: `LowCardinality(String)` → after stripping → `String` (line 1110 strips it)

So this test is a clean mutation guard for **line 1110** (storage-side stripping): drop that call, the type-equality check fails, `DirectKeyValueJoin` is declined, the EXPLAIN assertion flips to HashJoin, the test fails. ✅

Line 1108 (right-side stripping) is technically not exercised in the strict mutation sense — for `LowCardinality(String)` PK, the legacy `JOIN ... USING` path normalizes the right key to plain `String` before `tryDirectJoin` runs, so even without that call the right-side value is already non-LC. This is an acceptable limitation: the right-side stripping mostly matters for composite PKs (e.g. `Tuple(LowCardinality(String), UInt64)`) where the inner LC needs recursive stripping. A follow-up test using a `Tuple`/`Array`-typed PK with inner `LowCardinality` would close that, but I don't think it should block this PR.

**Verdict:**

- Test number `04209` is unused, no collision. ✅
- Tag `-- Tags: use-rocksdb` matches convention. ✅
- Reference output is correct (`d` produces empty string for missing key — the value column is non-Nullable `String`). ✅
- `pr-not-for-changelog` and `can be tested` labels already applied. ✅

Approving from a contributor perspective. Maintainer review is still needed for merge — cc'ing them is up to whoever the bot routes through next.

alexey-milovidov (approved):

PR #104320, opened 2026-05-07
Default `input_format_column_name_matching_mode` to `auto`

alexey-milovidov:
Make the column-name matching for input formats (e.g. `JSONEachRow`, `CSVWithNames`, `JSONColumns`, `BSONEachRow`, `RowBinaryWithNames`) match case-sensitively first and fall back to case-insensitive matching on miss.

The previous default of `match_case` required an exact case match, which is unnecessarily strict for the common case where input column names happen to differ in capitalization from the table's columns. The `auto` mode is unambiguous when the table has no two columns that differ only in case, and reports `INCORRECT_DATA` on ambiguity, so this is a safe relaxation. The previous behavior is preserved under `compatibility`.

- Improvement

The default of `input_format_column_name_matching_mode` is changed from `match_case` to `auto`. Input formats that match input column names against the table schema (`JSONEachRow`, `CSVWithNames`, `JSONColumns`, `BSONEachRow`, `RowBinaryWithNames`, etc.) now first try a case-sensitive match and fall back to case-insensitive matching when the case-sensitive match misses. The previous strict behavior is preserved under `compatibility`.

### DIFF
--- a/src/Core/FormatFactorySettings.h
+++ b/src/Core/FormatFactorySettings.h
@@ -1557,7 +1557,7 @@ Allow to write information about geo columns in parquet metadata and encode colu
     DECLARE(Bool, into_outfile_create_parent_directories, false, R"(
 Automatically create parent directories when using INTO OUTFILE if they do not already exists.
 )", 0) \
-    DECLARE(InputFormatColumnMatchingCaseSensitivity, input_format_column_name_matching_mode, FormatSettings::InputFormatColumnMatchingCaseSensitivity::MATCH_CASE, R"(
+    DECLARE(InputFormatColumnMatchingCaseSensitivity, input_format_column_name_matching_mode, FormatSettings::InputFormatColumnMatchingCaseSensitivity::AUTO, R"(
 Defines the column name matching mode when ingesting data through various formats (including but not limited to JSONEachRow, CSVWithNames, JSONColumns, BSONEachRow, RowBinaryWithNames).
 Supported modes:
     - match_case: match case-sensitively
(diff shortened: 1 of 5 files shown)


tiandiwonder (approved):
better to add a default value & compatibility test? refer to tests/queries/0_stateless/02970_visible_width_behavior.sql


```
-- 1. Default behavior (no SET).
--    1a. Value-level: system.settings reports the new default.
--    1b. Behavior-level: under the new default `auto`, "ID" falls back to `id`.
SELECT value FROM system.settings WHERE name = 'input_format_column_name_matching_mode';

DROP TABLE IF EXISTS test_default;
CREATE TABLE test_default (id Int, name String) ENGINE = Memory;

INSERT INTO test_default FORMAT JSONEachRow {"ID": 1, "name": "a"};

SELECT id, name FROM test_default;

DROP TABLE test_default;

-- 2. Compatibility rollback to a pre-26.5 version must restore `match_case`.
--    2a. Value-level: system.settings reports the restored old default.
--    2b. Behavior-level: "ID" no longer matches `id` and is treated as an unknown field
--    (skipped under the default `input_format_skip_unknown_fields = true`); `id` defaults to 0.
SET compatibility = '26.4';

SELECT value FROM system.settings WHERE name = 'input_format_column_name_matching_mode';

DROP TABLE IF EXISTS test_compat;
CREATE TABLE test_compat (id Int, name String) ENGINE = Memory;

INSERT INTO test_compat FORMAT JSONEachRow {"ID": 1, "name": "a"};

SELECT id, name FROM test_compat;

DROP TABLE test_compat;
```

alexey-milovidov:
> better to add a default value & compatibility test?

I doubt that, as it will be too pedantic a test, that just reasserts the change and adds nothing on top.

PR #104322, opened 2026-05-07
fixing a possible underflow while parsing postgres array values

grantholly-clickhouse:

Fixes a possible underflow in parsing postgres arrays.

When parsing arrays, we decrement a size_t without checking if it is 0 or not.  Now we will throw an exception instead.

### Version info
- Merged into: `26.5.1.576`
- Backported to: `26.4.3.15`, `26.3.11.15`, `26.2.19.22`, `25.8.24.15`

### DIFF
--- a/src/Core/PostgreSQL/insertPostgreSQLValue.cpp
+++ b/src/Core/PostgreSQL/insertPostgreSQLValue.cpp
@@ -149,6 +149,9 @@ void insertPostgreSQLValue(
                 {
                     max_dimension = std::max(max_dimension, dimension);
 
+                    if (dimension == 0)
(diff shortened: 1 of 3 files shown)


alexey-milovidov (approved):

clickgapai:

**Affected code:** `insertPostgreSQLValue` array parsing in src/Core/PostgreSQL/insertPostgreSQLValue.cpp — core code present in all supported branches.

**Why this might apply:** This is a crash/undefined behavior fix in PostgreSQL array parsing. The affected code path in `insertPostgreSQLValue` predates all supported releases - the PostgreSQL storage integration has existed for years. The bug could cause server crashes when reading malformed data via type mismatch (PostgreSQL text column read as ClickHouse Array).

PR #103767, opened 2026-04-30
Fix flaky test_recover_from_snapshot_with_chunked_transfer

groeneai:
### Description

The integration test `test_recover_from_snapshot_with_chunked_transfer` creates a kazoo client for the lagging node *before* `stop_clickhouse(kill=True)`, then reuses the same client after restart to assert post-recovery state. Under CI load the kill+restart window can exceed the 30 s default kazoo session timeout, after which the session is irrecoverably expired and the client raises `ConnectionClosedError` on the next request — even with the wrapper's implicit retries.

This PR re-creates the kazoo client after the restart, matching the pattern used by every other test method in the same file (`test_recover_after_interrupted_transfer`, `test_recover_with_chunk_size_larger_than_snapshot`, `test_recover_after_s3_read_error_during_transfer`, `test_recover_from_snapshot_sent_by_old_leader`).

### Failure example

CI failure that triggered this fix — PR #103394, `Integration tests (amd_tsan, 6/6)`:

```
File: test_keeper_snapshot_chunked_transfer/test.py:128 - in test_recover_from_snapshot_with_chunked_transfer
    assert lagging_zk.get(prefix)[0] == b"somedata"
...
kazoo.exceptions.ConnectionClosedError: Connection has been closed
```

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=103394&sha=ad4662b27bacbf2146f3ece6456168c7b713946c&name_0=PR&name_1=Integration%20tests%20%28amd_tsan%2C%206%2F6%29

### Chronic flakiness in CIDB

30-day window:
- **11 distinct PRs** hit the failure (PR #100332: 12, PR #101757: 12, plus 9 others with 1 each).
- **1 master hit** on 2026-04-21.
- Open tracking issues: ClickHouse/ClickHouse#102040, ClickHouse/ClickHouse#101969.

The PR-relatedness of failing PRs has been verified — none of them touch Keeper code, and they are spread across unrelated areas (PREWHERE/IN-subquery, throw on unknown config, exact subcolumn match, JSON/text index, skip index aggregation, PK analysis stats, `!subranges.empty()` logical error, MongoDB cleanup, Azure validation). This rules out a real regression and confirms a chronic test-side flake.

### Why kazoo cannot recover

[...]

PR #104325, opened 2026-05-07
Match interrupted tmp_snapshot markers by (name, last_modified) in test_recover_after_interrupted_transfer

groeneai:
Fix the polling assertion in `test_recover_after_interrupted_transfer[remote_disk]`
so it identifies the original interrupted-transfer `tmp_snapshot_*.bin.zstd`
markers by `(object_name, last_modified)` instead of `object_name` alone.

## Motivation

The test has shown up across 15+ unrelated PRs in the last 14 days
(`Integration tests (arm_binary, distributed plan, 4/4)`, also
`amd_msan`, `amd_asan_ubsan`, `amd_llvm_coverage` variants). All failures
look the same:

```
File: test_keeper_snapshot_chunked_transfer/test.py:229 - in test_recover_after_interrupted_transfer
    assert not remaining, ...
E   AssertionError: tmp_snapshot objects from interrupted transfer not removed within 15 s:
    ['keeper-snapshots/node9/tmp_snapshot_1050.bin.zstd']
```

The reported file is always 0 bytes with `etag='d41d8cd98f00b204e9800998ecf8427e'`
(empty-string MD5) — the same shape as the marker file written by both
`beginSnapshotReceiveToDisk` and `serializeSnapshotToDisk` in
`KeeperSnapshotManager`.

Three fix PRs already landed (#102643, #102899, #103767) but
`test_recover_after_interrupted_transfer[remote_disk]` keeps failing.

## Root cause

Sequence:

1. The test pauses a chunked transfer at `keeper_save_snapshot_pause_mid_transfer`
   and records `interrupted_tmp_names = {o.object_name for o in tmp_objects}`.
2. The node is killed and restarted.
3. `KeeperSnapshotManager` constructor cleanup (lines 700–748 of
   `src/Coordination/KeeperSnapshotManager.cpp`) iterates the snapshot disk,
   collects `tmp_*` files into `incomplete_files`, and removes them. The
   *original* `tmp_snapshot_1050.bin.zstd` is correctly deleted at this point.
4. NuRaft re-attempts the snapshot transfer. If no new entries were committed
   on the leader between pause and restart (typical when the kill window is
   short), it resends the snapshot for the **same** `log_idx`.
5. `beginSnapshotReceiveToDisk(1050)` runs again, creating a fresh
   `tmp_snapshot_1050.bin.zstd` marker — same `object_name`, but a newer
   `last_modified`.
6. The polling loop matches by `object_name` only, so the new in-flight marker
   is mistaken for the original. The 15 s deadline expires while the new
   transfer is still copying chunks and the assertion fails.

The C++ cleanup is doing the right thing; the test's identity check is too loose.

## Fix

Record `(object_name, last_modified)` tuples for the interrupted markers
before the kill, and filter the post-restart listing on the same tuple.
MinIO returns `last_modified` with microsecond precision, so the original
file (created before the kill) and any new in-flight marker (created after
the restart) always have different timestamps. Synthesised "directory"
entries (`is_dir=True`, `last_modified=None`) are also naturally excluded.

A genuine cleanup regression still fails the assertion: if the original
`tmp_snapshot_1050.bin.zstd` is left on the disk after restart, its
`(name, T1)` tuple is unchanged and matches the recorded set.

## CI report

`Integration tests (arm_binary, distributed plan, 4/4)` on PR #100146 — one
of 15+ unrelated PRs hit by this test in the last 14 days:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100146&sha=60fad4c4a87adad7da97b1b7249acf9d6370d4ff&name_0=PR&name_1=Integration%20tests%20%28arm_binary%2C%20distributed%20plan%2C%204%2F4%29

...

### Version info
- Merged into: `26.5.1.434`

### DIFF
--- a/tests/integration/test_keeper_snapshot_chunked_transfer/test.py
+++ b/tests/integration/test_keeper_snapshot_chunked_transfer/test.py
@@ -212,10 +212,16 @@ def test_recover_after_interrupted_transfer(started_cluster, nodes):
             "root", prefix=s3_prefix + "tmp_snapshot_"
         ))
         assert tmp_objects, "No tmp_snapshot object in S3 after killing mid-transfer"
-        # Record the specific tmp files left by the interrupted transfer so we can
-        # check that *these* are cleaned up, ignoring any transient tmp_ markers
-        # created by background snapshot flushes after the node restarts.
-        interrupted_tmp_names = {o.object_name for o in tmp_objects}
+        # Record (name, last_modified) tuples to uniquely identify these specific
+        # files.  A new snapshot transfer after restart may create a tmp_ marker
+        # with the *same* object_name (the leader can resend the same log_idx if
+        # nothing committed in the meantime), so matching by name alone causes
+        # false positives where the new in-flight marker is mistaken for the
+        # interrupted-transfer marker that startup cleanup already removed.
+        # last_modified has microsecond precision in MinIO and is monotonically
+        # increasing across writes, so the (name, last_modified) tuple is a
+        # reliable identity for the original interrupted-transfer file.
+        interrupted_tmp_ids = {(o.object_name, o.last_modified) for o in tmp_objects}
     else:
         snapshot_dir = "/var/lib/clickhouse/coordination/snapshots"
         tmp_snapshot_path = node_lagging.exec_in_container(
@@ -230,10 +236,12 @@ def test_recover_after_interrupted_transfer(started_cluster, nodes):
 
     if is_remote:
         # After restart the node may also create new local snapshots asynchronously
-        # (queued via snapshots_queue), which temporarily produce fresh tmp_ markers
-        # with the same or different log indices.  Poll for the *interrupted* markers
-        # to disappear rather than asserting immediately, so a concurrent background
-        # flush does not cause a false positive.
+        # (queued via snapshots_queue) or receive a fresh snapshot from the leader
+        # that re-uses the same log_idx, both of which temporarily produce a tmp_
+        # marker with the same object_name as the interrupted-transfer marker.
+        # Match on (name, last_modified) so we only count the *original* file —
+        # the one whose cleanup we are verifying — and ignore unrelated in-flight
+        # markers created after the restart.
         deadline = time.time() + 15
         remaining = None
         while True:
@@ -241,7 +249,7 @@ def test_recover_after_interrupted_transfer(started_cluster, nodes):
                 o for o in started_cluster.minio_client.list_objects(
                     "root", prefix=s3_prefix + "tmp_snapshot_"
                 )
-                if o.object_name in interrupted_tmp_names
+                if (o.object_name, o.last_modified) in interrupted_tmp_ids
             ]
             if not remaining or time.time() >= deadline:
                 break


groeneai:
**Pre-PR Validation Gate self-check** (per TASK.md Phase 4 step 9, session `cron:clickhouse-ci-task-worker:20260507-204500`):

**a) Deterministic repro?** ❌ Not locally — the failure requires a multi-node Keeper-on-S3 cluster with chunked transfer + failpoint pause + Docker-in-Docker MinIO + iptables manipulation. The race only manifests when the leader resends the same `log_idx` snapshot after a kill+restart. Per past memory, "wall-clock timing-based flaky tests cannot be deterministically reproduced locally due to CI-specific noise." Verification rests on CIDB pattern analysis and code reading.

**b) Root cause explained?** ✅ Test polling matches a NEW `tmp_snapshot_<idx>.bin.zstd` (created by `beginSnapshotReceiveToDisk` after restart, when the leader resends the same `log_idx` because nothing committed during the pause) as if it were the ORIGINAL interrupted marker that startup cleanup correctly removed. The two files share `object_name` but always differ in `last_modified`.

**c) Fix matches root cause?** ✅ `(object_name, last_modified)` tuple identity precisely distinguishes the original interrupted-transfer file (recorded before the kill) from any new in-flight marker created after the restart. No timeout extension, no retry, no test disable.

**d) Test intent preserved?** ✅ If the C++ startup cleanup (`KeeperSnapshotManager` constructor lines 700–748 of `src/Coordination/KeeperSnapshotManager.cpp`) actually fails to remove the original `tmp_snapshot_*.bin.zstd`, the original file's `(name, T1)` tuple is unchanged and still appears in the listing → the assertion still fails as designed. The fix only eliminates the false positive class where a new in-flight marker shares `object_name`.

**e) Demonstrated in both directions?** ⚠ Partial. Without the fix, CIDB shows `last_modified` on the failing file is consistently set to the post-restart time (e.g. `2026-05-01 15:10:56.351000` vs the test's recording phase several seconds earlier) — i.e. the file in the listing is a NEW transfer's marker, not the original. With the fix, the post-restart timestamp would not match the recorded tuple and the assertion would pass. A real cleanup regression would still show the original timestamp and still fail.

**f) Fix is general?** ✅ Addresses the loose identity check at its only call site. Sibling tests in the suite (`test_recover_from_snapshot_with_chunked_transfer`, `test_recover_after_s3_read_error_during_transfer`, etc.) do not have this assertion shape, so no other call sites need the same change. Underlying C++ cleanup is correct and unchanged.

groeneai:
cc @kssenii — could you take a look? This is a follow-up to #102899 (my earlier fix) and your own #102643. The polling assertion still false-fails because the leader can resend the same `log_idx` after the kill+restart, creating a new `tmp_snapshot_<same_idx>.bin.zstd` whose object_name collides with the recorded interrupted-transfer marker. Tightening the identity to `(name, last_modified)` eliminates the collision while still catching real cleanup regressions in `KeeperSnapshotManager::KeeperSnapshotManager`.

nikitamikhaylov (approved):

PR #104326, opened 2026-05-07
Fix extractAll and similar functions on Nullable inputs

### LINKED ISSUE #56977: Poor usability of `extractAll` and Nullable types.
```
ip-172-31-25-236.eu-central-1.compute.internal :) SELECT toNullable('Hello, world') AS s, extractAll(s, '(\w+), (\w+)')

SELECT
    toNullable('Hello, world') AS s,
    extractAll(s, '(\\w+), (\\w+)')

Query id: ef9b1064-043c-429b-b6c4-44ab672e1af1


Elapsed: 0.004 sec. 

Received exception:
Code: 43. DB::Exception: Nested type Array(String) cannot be inside Nullable type: While processing toNullable('Hello, world') AS s, extractAll(s, '(\\w+), (\\w+)'). (ILLEGAL_TYPE_OF_ARGUMENT)
```

alexey-milovidov:
For functions that use the default Nulls implementation but return a type that cannot be Nullable (e.g. `Array`, `Tuple`, `Map`), the framework previously called `makeNullable` on the return type, which threw "Nested type is not allowed inside Nullable type". This made `extractAll`, `splitByChar`, `splitByRegexp`, `extractAllGroups`, and similar functions unusable on `Nullable` columns.

`IFunctionOverloadResolver::getReturnTypeWithoutLowCardinality` now uses `makeNullableSafe`, and `IExecutableFunction::defaultImplementationForNulls` skips the `wrapInNullable` step when the result type is not `Nullable`. Null input rows are evaluated over the default value of the nested column (e.g. an empty string), producing the natural default of the result type (e.g. `[]` for `extractAll`).

Closes https://github.com/ClickHouse/ClickHouse/issues/56977

Functions that return a non-`Nullable` type (such as `Array`, `Tuple`, or `Map`) now accept `Nullable` arguments. Affected functions include `extractAll`, `extractAllGroups`, `extractAllGroupsHorizontal`, `extractAllGroupsVertical`, `extractGroups`, `splitByChar`, `splitByString`, `splitByRegexp`, `splitByWhitespace`, `splitByNonAlpha`, and `alphaTokens`. `NULL` input rows produce the default value of the result type (e.g. an empty array) instead of raising "Nested type is not allowed inside Nullable type".

### Version info
- Merged into: `26.5.1.470`


george-larionov:
The comment in IFunction.h above useDefaultImplementationForNulls() says that the function may be executed with garbage input instead of null values, and this PR removes the wrapInNullable() and just returns res without applying the null_map. I think there is a possibility of returning the output over these garbage values in some cases, reproducer (the first line should just output null for extractAll but it outputs `['hello','world']`):

```
WITH if(number = 0, nullIf(materialize('hello world'), materialize('hello world')), CAST('foo bar', 'Nullable(String)')) AS s
SELECT
    number,
    s,
    isNull(s),
    extractAll(s, '(\\w+)')
FROM numbers(3)

Query id: dc6e2b35-04c7-483c-a6a0-7212450a2ad8

   ┌─number─┬─s───────┬─isNull(s)─┬─extractAll(s, '(\\w+)')─┐
1. │      0 │ ᴺᵁᴸᴸ    │         1 │ ['hello','world']       │
2. │      1 │ foo bar │         0 │ ['foo','bar']           │
3. │      2 │ foo bar │         0 │ ['foo','bar']           │
   └────────┴─────────┴───────────┴─────────────────────────┘
```

src/Functions/IFunction.cpp
```diff
@@ -748,7 +769,11 @@ DataTypePtr IFunctionOverloadResolver::getReturnTypeWithoutLowCardinality(const
         {
             Block nested_columns = createBlockWithNestedColumns(arguments);
             auto return_type = getReturnTypeImpl(ColumnsWithTypeAndName(nested_columns.begin(), nested_columns.end()));
-            return makeNullable(return_type);
+            /// If the return type cannot be Nullable (e.g. Array, Tuple, Map),
+            /// return it as-is. For null input rows, the function will be evaluated
+            /// over the default values of the nested column and produce the corresponding default result
+            /// (e.g. an empty array), instead of failing the type check.
+            return makeNullableSafe(return_type);
```
george-larionov, line 812: makeNullableSafe() returns the return_type untouched if it cannot be made Nullable. Since useDefaultImplementationForNulls() is true by default for derived classes, this means that functions that return non-nullable types used to fail and force the author to think about whether they should be using useDefaultImplementationForNulls or not. With this change, they will just run with potentially incorrect results. Maybe we should set useDefaultImplementationForNulls to false by default then?

src/Functions/IFunction.cpp
```diff
@@ -295,6 +312,8 @@ ColumnPtr IExecutableFunction::defaultImplementationForNulls(
         {
             /// Each row should be evaluated if there are no nulls or short circuiting is disabled.
             auto res = executeWithoutLowCardinalityColumns(temporary_columns, temporary_result_type, input_rows_count, dry_run);
+            if (!result_is_nullable)
+                return res;
```
george-larionov: This `res` may contain outputs from garbage inputs (there is no guarantee that the null rows actually contain default vals AFAIK). Unlike in the code on/after line 322, there is no handling of this. Maybe in the case of `!result_is_nullable` it should always go through the logic on line 322-339? See [this comment](https://github.com/ClickHouse/ClickHouse/pull/104326#issuecomment-4405434141).

src/Functions/IFunction.cpp
```diff
@@ -246,6 +257,8 @@ ColumnPtr IExecutableFunction::defaultImplementationForNulls(
             auto temporary_result_type = removeNullable(result_type);
 
             auto res = executeWithoutLowCardinalityColumns(temporary_columns, temporary_result_type, input_rows_count, dry_run);
+            if (!result_is_nullable)
+                return res;
```
george-larionov: This `res` may contain outputs from garbage inputs (there is no guarantee that the null rows actually contain default vals AFAIK). Unlike in the code on/after line 322, there is no handling of this. Maybe in the case of `!result_is_nullable` it should always go through the logic on line 322-339? See [this comment](https://github.com/ClickHouse/ClickHouse/pull/104326#issuecomment-4405434141).

george-larionov (changes requested):
I have some concerns with the handling of null inputs containing garbage vals, see comments.

george-larionov:
Another example with incorrect output, should evaluate to `[]`:

```
SELECT extractAll(nullIf(materialize('hello world'), materialize('hello world')), '(\\w+)')

Query id: b8d2c747-ccf4-4fbd-a72a-4057da808c96

   ┌─extractAll(n⋯, '(\\w+)')─┐
1. │ ['hello','world']        │
   └──────────────────────────┘
```

src/Functions/IFunction.cpp
```diff
@@ -210,6 +210,63 @@ ColumnPtr IExecutableFunction::defaultImplementationForNulls(
...
+        {
+            /// The result type cannot be Nullable (e.g. `Array`, `Tuple`, `Map`).
+            /// The framework convention here is `f(default(input))` for null rows: the function
+            /// runs over inputs where null rows have been normalized to the default value of
+            /// the (nested) input type. This is necessary because `createBlockWithNestedColumns`
+            /// alone does not overwrite null rows -- e.g. `nullIf(materialize('x'), materialize('x'))`
+            /// produces a Nullable column whose nested data still contains `'x'`, so running the
+            /// function on it would return `f('x')` instead of the desired `f('')`.
+            ColumnsWithTypeAndName patched_columns = createBlockWithNestedColumns(args);
+            auto temporary_result_type = removeNullable(result_type);
```
george-larionov, line 225: we could use result_type directly, since it is non-nullable, right?

george-larionov (approved):
Overall LGTM now, although still thinking about whether useDefaultImplementationForNulls should be false by default to avoid unexpected behavior, but that is more of an opinion.

src/Functions/IFunction.cpp
alexey-milovidov: Yes, though we can keep it...

clickgapai:

**Affected code:** `IFunctionOverloadResolver::getReturnTypeWithoutLowCardinality` and `IExecutableFunction::defaultImplementationForNulls` — core code present in all supported branches.

**Why this might apply:** The bug exists in `getReturnTypeWithoutLowCardinality` which has used `makeNullable` since the function was introduced. Issue #56977 from 2023 confirms this is a long-standing bug affecting all versions. The fix is safe (adds fallback handling) and addresses user-visible errors.

PR #88504, opened 2025-10-14
Fix explain indexes with use_skip_indexes_on_data_read + use_query_condition_cache

shankar-iyer:
### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

### Details

Fixes explain indexes with use_skip_indexes_on_data_read + use_query_condition_cache. Behavior change introduced and fixed in 25.10.

Resolves https://github.com/ClickHouse/ClickHouse/issues/88467

PR #104331, opened 2026-05-08
Disable skip-index-on-data-read for leaf row-limit + throw mode

### LINKED ISSUE #104328: `supportsSkipIndexesOnDataRead` disables only `max_rows_to_read`, not `_leaf` variants — `SETTINGS max_rows_to_read_leaf=N, read_overflow_mode_leaf='throw'` queries throw TOO_MANY_ROWS even when skip index would reduce to 1 row

_Retrospective finding from a historical scan of [PR #88504](https://github.com/ClickHouse/ClickHouse/pull/88504) (merged 2025-10-31). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

With default `use_skip_indexes_on_data_read=1` (true since 26.1+), `SELECT count() FROM t WHERE val = 50 SETTINGS max_rows_to_read_leaf=5, read_overflow_mode_leaf='throw'` throws `Limit for rows or bytes to read on leaf node exceeded, max rows: 5.00, current rows: 100.00` even though the minmax skip index on `val` would reduce the read to a single row. The same query with `max_rows_to_read=5` (non-leaf) succeeds because the PR disables `use_skip_indexes_on_data_read` for that combination.

**Root cause:** src/Processors/QueryPlan/ReadFromMergeTree.cpp:3319 — `if (settings[Setting::read_overflow_mode] == OverflowMode::THROW && settings[Setting::max_rows_to_read]) return false;` checks only the non-leaf settings. The symmetric `_leaf` variants are required because `IInterpreterUnionOrSelectQuery::getStorageLimits` (Utils.cpp:260, IInterpreterUnionOrSelectQuery.cpp:134) and `MergeTreeDataSelectExecutor::getRowLimits` (line 1280-1283) install identical `SizeLimits` for leaf, and `MergeTreeDataSelectExecutor::filterPartsByPrimaryKeyAndSkipIndexes` checks both at line 1083 with the same `total_rows_estimate`.

**Why we believe this is a bug:** Entry: ReadFromMergeTree.cpp:3285 `supportsSkipIndexesOnDataRead` → line 3319 only short-circuits when `read_overflow_mode == THROW && max_rows_to_read != 0`, missing the symmetric `read_overflow_mode_leaf == THROW && max_rows_to_read_leaf != 0`. The result is fed via `selectRangesToRead` (line 2453: `indexes->use_skip_indexes_on_data_read = supports_skip_indexes_on_data_read`) into the lambda at MergeTreeDataSelectExecutor.cpp:871 (`is_index_supported_on_data_read`). When the lambda returns true, `filterMarksUsingIndex` is skipped during planning, so `total_rows_estimate` (passed to `leaf_limits.check` at line 1083) reflects ALL granules unfiltered. Independently, the runtime `ReadProgressCallback::onProgress` enforces the same leaf limit on actual reads (`SizeLimits::check` in the ASan stack trace from my failing test). Either path throws TOO_MANY_ROWS.

**Affected locations:**
- `src/Processors/QueryPlan/ReadFromMergeTree.cpp:3319` — supportsSkipIndexesOnDataRead — disables data-read skip index for non-leaf max_rows_to_read+throw but not for the _leaf variants

**Impact:** Existing distributed queries that rely on `max_rows_to_read_leaf` as a per-leaf safety guard (a common operator/cloud pattern) will start throwing TOO_MANY_ROWS after the upgrade that flipped `use_skip_indexes_on_data_read` to true by default — even when the matching rows are far fewer than the leaf cap. The query is REJECTED instead of returning the correct (small) result, which is a correctness regression and a denial-of-functionality on perfectly valid queries.

Yes — confirmed on current `master` (commit `c93903f9fd6`).

### How to reproduce

```sql
-- no-parallel-replicas: use_skip_indexes_on_data_read is not supported with parallel replicas
-- Test: PR #88504 disables use_skip_indexes_on_data_read for max_rows_to_read+throw,
-- but the symmetric leaf variants (max_rows_to_read_leaf, read_overflow_mode_leaf) are NOT
-- handled. With default use_skip_indexes_on_data_read=1, queries using leaf limits incorrectly
-- throw TOO_MANY_ROWS because skip index is deferred to data read and the early row estimate
-- check at MergeTreeDataSelectExecutor.cpp:1083 sees the full unfiltered count.
-- (Disable query_condition_cache so a previous query cannot mask the bug for the leaf path.)

SET use_query_condition_cache = 0;

DROP TABLE IF EXISTS t_88504;

CREATE TABLE t_88504 (
    id UInt64,
    val UInt64,
    INDEX val_idx val TYPE minmax GRANULARITY 1
) ENGINE = MergeTree ORDER BY id SETTINGS index_granularity = 1;

INSERT INTO t_88504 SELECT number, number FROM numbers(100);

-- max_rows_to_read=5: PR-handled. Skip index is applied during planning -> 1 row -> succeeds.
SELECT count() FROM t_88504 WHERE val = 50 SETTINGS use_skip_indexes_on_data_read=1, max_rows_to_read=5, read_overflow_mode='throw';

-- max_rows_to_read_leaf=5: should be the same, but the PR did NOT add the disable for the
-- _leaf variant. Skip index is deferred to data read; the leaf-limit estimate check then
-- sees 100 rows and throws TOO_MANY_ROWS.
-- Expected (after fix): 1
-- Actual (current bug): TOO_MANY_ROWS exception.
SELECT count() FROM t_88504 WHERE val = 50 SETTINGS use_skip_indexes_on_data_read=1, max_rows_to_read_leaf=5, read_overflow_mode_leaf='throw';

DROP TABLE t_88504;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/32ed953a-a004-4fbe-80c6-90dc1d29cb71)

### Expected behavior

```
1
1
```

```
1
Code: 158. DB::Exception: Received from localhost:19030. DB::Exception: Limit for rows or bytes to read on leaf node exceeded, max rows: 5.00, current rows: 100.00: While executing MergeTreeSelect(pool: ReadPoolInOrder, algorithm: InOrder). (TOO_MANY_ROWS)
```

### Additional context

**Open risks:**
- The same fix may also need to be applied for `max_bytes_to_read` / `max_bytes_to_read_leaf` once the BYTES progress estimate flows the same way — out of scope for this finding but worth a follow-up.
- Query Condition Cache writes from `MergeTreeSelectProcessor::read` (lines 266, 283, 361) are NOT gated by `query_info.isFinal()` even though reads ARE — FINAL queries with `use_query_condition_cache=1` can still pollute the cache. This is a separate hypothesis.

**Suggested fix:** Extend the guard to also cover the leaf variants:

    if ((settings[Setting::read_overflow_mode] == OverflowMode::THROW && settings[Setting::max_rows_to_read])
        || (settings[Setting::read_overflow_mode_leaf] == OverflowMode::THROW && settings[Setting::max_rows_to_read_leaf]))
        return false;

The `_leaf` extern Setting<...> is already declared at line 193 (`max_rows_to_read_leaf`) and 210 (`read_overflow_mode_leaf`) in the same translation unit.

Found during automated review of [PR #88504](https://github.com/ClickHouse/ClickHouse/pull/88504).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr88504_001`_

### Version info
- Resolved by: #104331
- Merged into: `26.5.1.597`

groeneai:
Closes #104328.

### Background

PR #88504 disabled `use_skip_indexes_on_data_read` when
`read_overflow_mode == 'throw'` and `max_rows_to_read != 0` because the
early row-count estimate at `MergeTreeDataSelectExecutor::filterPartsByPrimaryKeyAndSkipIndexes`
runs before the skip index would have a chance to drop granules. Without
this short-circuit the early estimate sees the full unfiltered row count
and `limits.check` rejects valid queries with TOO_MANY_ROWS.

The symmetric leaf-limit pair (`max_rows_to_read_leaf` +
`read_overflow_mode_leaf`) was missed: the same code path also
populates `leaf_limits` (via `MergeTreeDataSelectExecutor::getRowLimits`)
and the same early estimate is checked against `leaf_limits` at the
same site (`MergeTreeDataSelectExecutor.cpp:1083`). So a query with the
leaf settings throws TOO_MANY_ROWS even when the skip index would have
reduced the read to far fewer rows than the leaf cap. This is a
correctness regression for distributed queries that rely on
`max_rows_to_read_leaf` as a per-leaf safety guard (a common operator/
cloud pattern), introduced when `use_skip_indexes_on_data_read` was
flipped on by default in 26.1.

### Reproducer (issue #104328)

```sql
CREATE TABLE t (id UInt64, val UInt64, INDEX val_idx val TYPE minmax GRANULARITY 1)
ENGINE = MergeTree ORDER BY id SETTINGS index_granularity = 1;

INSERT INTO t SELECT number, number FROM numbers(100);

-- Already works (PR #88504): returns 1.
SELECT count() FROM t WHERE val = 50
SETTINGS use_skip_indexes_on_data_read=1, max_rows_to_read=5, read_overflow_mode='throw',
         use_query_condition_cache=0;

-- Bug before this PR: throws TOO_MANY_ROWS (max rows: 5.00, current rows: 100.00).
-- After this PR: returns 1.
SELECT count() FROM t WHERE val = 50
SETTINGS use_skip_indexes_on_data_read=1, max_rows_to_read_leaf=5, read_overflow_mode_leaf='throw',
         use_query_condition_cache=0;
```

### Fix

Extend the early-return in `ReadFromMergeTree::supportsSkipIndexesOnDataRead`
to also short-circuit when the leaf row limit is set with throw mode,
mirroring the existing non-leaf line.

Bytes-based limits (`max_bytes_to_read` / `max_bytes_to_read_leaf`) are
out of scope: the planning-time estimate at
`MergeTreeDataSelectExecutor.cpp:1083` only checks rows; bytes are not
in this code path. (The runtime `ReadProgressCallback::onProgress` is a
separate path.)

### Test

`tests/queries/0_stateless/04210_skip_index_on_data_read_leaf_limit.sql`
covers both the existing non-leaf path (regression check), the new leaf
path (fixed by this PR), the combined leaf+non-leaf case, and three
sanity cases (`use_skip_indexes_on_data_read=0`, larger leaf limit,
`break` mode).

Verified locally on the master commit referenced in the issue
(`c93903f9fd6`):
- Without this PR, the leaf reproducer throws TOO_MANY_ROWS deterministically.
- With this PR, the test passes (30/30 with randomized settings).
- Existing test `03533_skip_index_on_data_reading` (PR #88504) still passes.

Fix a correctness regression where `SELECT` queries with `max_rows_to_read_leaf` and `read_overflow_mode_leaf = 'throw'` could incorrectly throw `TOO_MANY_ROWS` even when a skip index would have reduced the read below the leaf cap. The symmetric non-leaf settings (`max_rows_to_read` / `read_overflow_mode`) were already handled.

### Version info
- Merged into: `26.5.1.597`


groeneai:
## Pre-PR validation gate

**a) Deterministic repro?** Yes. On master `c93903f9fd6`, the leaf reproducer from the issue throws `TOO_MANY_ROWS (max rows: 5.00, current rows: 100.00)` every time:

```sql
SELECT count() FROM t_104328 WHERE val = 50
SETTINGS use_skip_indexes_on_data_read = 1, max_rows_to_read_leaf = 5,
         read_overflow_mode_leaf = 'throw', use_query_condition_cache = 0;
```

**b) Root cause explained?** With `use_skip_indexes_on_data_read = 1`, `ReadFromMergeTree::supportsSkipIndexesOnDataRead` returns true → the lambda `is_index_supported_on_data_read` returns true at `MergeTreeDataSelectExecutor.cpp:871` → `filterMarksUsingIndex` is skipped during planning → the early-estimate check at `MergeTreeDataSelectExecutor.cpp:1083` (`leaf_limits.check(total_rows_estimate, ...)`) sees the unfiltered 100 rows and exceeds `max_rows_to_read_leaf = 5` → `TOO_MANY_ROWS`. The non-leaf pair already short-circuits at `ReadFromMergeTree.cpp:3319`; the leaf pair was missed.

**c) Fix matches root cause?** Yes — adds the symmetric early-return for `read_overflow_mode_leaf == THROW && max_rows_to_read_leaf` so `supportsSkipIndexesOnDataRead` returns false on the same condition for the leaf settings, and planning-time `filterMarksUsingIndex` runs as expected. Pattern matches PR #88504's existing line.

**d) Test intent preserved? / New tests added?** New regression test `04210_skip_index_on_data_read_leaf_limit.sql` covers: non-leaf path (PR #88504 regression check), leaf path (the bug), combined leaf + non-leaf, sanity case `use_skip_indexes_on_data_read = 0`, sanity case with high leaf limit, and the negative case `read_overflow_mode_leaf = 'break'` where the fix must NOT trigger. Existing test `03533_skip_index_on_data_reading` still passes.

**e) Both directions demonstrated?** Yes:
- Without this PR: leaf reproducer throws `TOO_MANY_ROWS` deterministically.
- With this PR: the new test passes 30/30 with random settings.

**f) Fix is general, not a narrow patch?** Bytes-based limits (`max_bytes_to_read` / `max_bytes_to_read_leaf`) are out of scope for this PR — the planning-time check at `MergeTreeDataSelectExecutor.cpp:1083` only checks rows (the bytes parameter is hardcoded to `0`). The runtime `ReadProgressCallback::onProgress` is a separate path and not affected by `supportsSkipIndexesOnDataRead`. The issue body explicitly notes bytes as a follow-up; if the planning-time bytes estimate is added later, the symmetric short-circuit can be extended at the same site.

session: cron:clickhouse-ci-task-worker:20260508-054500

groeneai:
cc @shankar-iyer @rschu1ze — could you review this? It extends the early-return in `supportsSkipIndexesOnDataRead` to also cover the leaf row-limit + throw mode pair, mirroring PR #88504's existing non-leaf line. Without it, queries with `max_rows_to_read_leaf=N, read_overflow_mode_leaf='throw'` on tables with skip indexes throw `TOO_MANY_ROWS` even when the index would have reduced the read below the leaf cap (issue #104328).

shankar-iyer:
Simple fix, waiting for CI results

shankar-iyer (approved):

clickgapai:

**Affected code:** `ReadFromMergeTree::supportsSkipIndexesOnDataRead` at src/Processors/QueryPlan/ReadFromMergeTree.cpp:3307-3356 — introduced in 25.9 (function introduced), 26.1 (default enabled, bug exposed).

**Why this might apply:** The bug affects users who set `max_rows_to_read_leaf` + `read_overflow_mode_leaf='throw'` after `use_skip_indexes_on_data_read` became enabled by default in 26.1. The `supportsSkipIndexesOnDataRead` function exists since 25.9 (when `use_skip_indexes_on_data_read` was introduced) but the default change to `true` in 26.1 exposed this regression. Version 25.8 is unaffected since `use_skip_indexes_on_data_read` defaults to `false` there.

Other supported branches (`25.8`) predate 25.9 (function introduced), 26.1 (default enabled, bug exposed) and do not contain this code.

PR #104351, opened 2026-05-08
Use progress-based timeout in ZooKeeper client to avoid killing healthy sessions under load

antonio2368:
Closes #100466. See the issue for full problem analysis and motivation.

Implementation differences from the issue's proposal:

- **Threshold is `session_timeout_ms`, not `operation_timeout_ms`.** One
  threshold at both layers (receive thread + sync wrappers). Avoids
  false positives when `operation_timeout_ms < session_timeout_ms / 3`
  (heartbeat interval).
- **Hard cap of `3 * session_timeout_ms` per request in the sync
  wrapper.** Catches requests stuck on an otherwise healthy session
  (e.g. lost by the server while heartbeats keep arriving).
- **Queue-full timeout no longer kills the session.** `pushRequest`
  fails just the one caller on backpressure; the request was never
  sent.

- Improvement

Improve ZooKeeper client timeout handling under heavy load when many requests are pipelined on a single session. The ZooKeeper client now uses a progress-based timeout: as long as any data is received from the server within `session_timeout_ms`, the wait is extended. A hard cap of `3 * session_timeout_ms` per request still bounds caller latency. Closes #100466.

### Version info
- Merged into: `26.5.1.756` (included in `26.5` and later)
- Backported to: `26.4.5.163`


src/Common/ZooKeeper/ZooKeeperImpl.h
```diff
@@ -381,6 +386,12 @@ class ZooKeeper final : public IKeeper
 
     std::atomic<int64_t> last_zxid_seen;
 
+    /// Timestamp of the last data received from the server (any kind: response,
+    /// heartbeat, or watch event), in microseconds since steady_clock epoch.
+    /// Updated by receiveThread after each receiveEvent call.
+    /// Read by sync wrappers via getLastReceivedAt() for progress-based timeout.
+    std::atomic<Int64> last_received_at{0};
```
alesapin: Maybe show that it's timestamp? last_received_at_time?

src/Common/ZooKeeper/ZooKeeperImpl.cpp
```diff
@@ -1521,12 +1556,19 @@ void ZooKeeper::pushRequest(RequestInfo && info)
...
+                args.operation_timeout_ms);
         }
     }
     catch (...)
     {
-        finalize(false, false, getCurrentExceptionMessage(false, false, false));
+        /// Queue-full timeout is transient backpressure, not a session failure.
+        /// The request was never sent (probably_sent=false), so it's safe to
+        /// fail just this one request without killing the session.
+        if (!queue_full_timeout)
```
alesapin: Maybe it make sense, but all the "user" code of this connection (for example ReplicatedMergeTree) will still do full re-initialization on this error. Maybe we can introduce some other error code which can be handled gracefully?

src/Common/ZooKeeper/ZooKeeper.cpp
```diff
@@ -113,6 +113,41 @@ UInt64 getSecondsUntilReconnect(const ZooKeeperArgs & args)
 }
 
 
+template <typename T>
+bool ZooKeeper::waitForFutureWithProgress(std::future<T> & future) const
+{
+    using clock = std::chrono::steady_clock;
+    const auto timeout = std::chrono::milliseconds(args.session_timeout_ms);
```
alesapin, line 120: Why session timeout and not operation timeout? I would prefer to wait for operation timeout => do some logging (`impl->getLastReceivedAt() > last_seen_ts`), wait more

src/Common/ZooKeeper/ZooKeeperImpl.cpp
```diff
@@ -905,52 +912,79 @@ void ZooKeeper::receiveThread()
...
+            if (now >= deadline)
+            {
+                if (now >= idle_deadline)
+                    throw Exception(Error::ZOPERATIONTIMEOUT,
+                        "Nothing is received in session timeout of {} ms",
+                        args.session_timeout_ms);
+
+                throw Exception(Error::ZOPERATIONTIMEOUT,
+                    "Request {} for path {} stuck for over {} ms despite global progress — finalizing session",
+                    stuck_request->getOpNum(),
```
alesapin, line 961: I don't see why stuck_request cannot be nullptr here...

src/Common/ZooKeeper/ZooKeeperImpl.cpp
```diff
@@ -905,52 +912,79 @@ void ZooKeeper::receiveThread()
...
+                    const auto & earliest = operations.begin()->second;
+                    stuck_deadline = earliest.request->create_ts + request_hard_cap_us;
+                    stuck_request = earliest.request;
                 }
             }
 
-            if (in->poll(max_wait_us))
+            auto deadline = std::min(idle_deadline, stuck_deadline);
+
+            if (now >= deadline)
```
alesapin, line 951: Don't we terminate connection now in case of no requests for session timeout? (idle_deadline)

alesapin (commented):
Good idea

src/Common/ZooKeeper/ZooKeeper.cpp
antonio2368: It is an extra layer of check in case receive thread gets stuck. Because of that it didn't make sense to me to have different timeouts.
But I was thinking about having the operation timeout for logging, wasn't sure if it would be too noisy during degradation.

src/Common/ZooKeeper/ZooKeeperImpl.cpp
antonio2368: It cannot be null because it's tied to `stuck_deadline`. We can reach this branch only if `stuck_deadline < idle_deadline` and `now < idle_deadline`.
If there is no `stuck_request`, `stuck_deadline = clock::time_point::max()` so the above wouldn't be true.
But we can definitely make it more expressive.

src/Common/ZooKeeper/ZooKeeperImpl.cpp
antonio2368: Yeah that is true,  it was something claude found that it made sense but I didn't want to add to this PR with more changes.
I'll just remove it from here and potentially add it in different PR.

src/Common/ZooKeeper/ZooKeeperImpl.cpp
antonio2368: There should always be requests because of Heartbeat.

alesapin (approved):

PR #104355, opened 2026-05-08
Reduce JIT-induced noise in performance tests

Algunenano:
The perf-comparison config already turns JIT off globally for the suite (`compile_expressions=0`, `compile_aggregate_expressions=0`, `compile_sort_description=0`). The tests that explicitly opt back into JIT were still bimodal: the JIT cache is global and shared across queries, so whether the first measured runs hit it or not depends on what ran before. With `min_count_to_compile_*` defaulting to 3, JIT compilation could land in the middle of the measured runs.

Set `min_count_to_compile_expression`, `min_count_to_compile_aggregate_expression` and `min_count_to_compile_sort_description` to 0 in the same profile so that when a test does enable JIT, compilation happens on the very first call and the existing prewarm in `perf.py` populates the cache before measurements start. Drop now-redundant per-test overrides of the same settings.

Refs #100759.

...

### Version info
- Merged into: `26.5.1.438`

### DIFF
--- a/tests/performance/jit_aggregate_functions.xml
+++ b/tests/performance/jit_aggregate_functions.xml
@@ -1,7 +1,6 @@
 <test>
     <settings>
         <compile_aggregate_expressions>1</compile_aggregate_expressions>
-        <min_count_to_compile_aggregate_expression>0</min_count_to_compile_aggregate_expression>
     </settings>
 
     <create_query>
--- a/tests/performance/jit_large_requests.xml
+++ b/tests/performance/jit_large_requests.xml
@@ -49,8 +49,7 @@
         WHERE
             NOT ignore(a / b + c / d + e / f + g / h + i / j)
         SETTINGS
-            compile_expressions = 1,
-            min_count_to_compile_expression = 1
+            compile_expressions = 1
     </query>
     <drop_query>DROP TABLE IF EXISTS jit_test</drop_query>
 </test>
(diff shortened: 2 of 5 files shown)


antaljanosbenjamin (approved):

Algunenano:
### Master 30-day baseline (1420 perf runs)

- In 36.2% of perf runs, at least one `jit_aggregate_functions` query was reported as slower, faster, or unstable in the PR perf report.
- Across all 28400 `jit_aggregate_functions` query samples in that window, 4.6% were classified slower/faster/unstable.
- Other JIT-using tests (`jit_small_requests`, `jit_large_requests`, `string_order_by`, `order_by_single_nullable`) sit below 0.5% — effectively only `jit_aggregate_functions` is moving the needle.

### This PR's ARM shards

- 10 `jit_aggregate_functions` query samples → 0 reported as unstable, slower, or faster.
- 7/10 had `stat_threshold` below 2.5%; max threshold 17.8%.
- Largest observed `times_change` 1.095×, well inside the noise threshold.
- Other JIT-using tests: also no flags.

### Caveat

Single CI run, so 0/10 is consistent with master's 4.6% per-query rate (binomial: ~62% chance of zero hits even without any change). Directional, not proof.

### Plan

Merge, then re-query the per-query noise rate for `jit_aggregate_functions` on master after a few days of accumulated runs to confirm.

PR #104358, opened 2026-05-08
Add regression test for Nested CLEAR COLUMN crash (issue #89537)

groeneai:
Closes #89537.

`ALTER TABLE ... CLEAR COLUMN` on a `Tuple`-typed subcolumn of a `Nested` column followed by a merge (e.g. `OPTIMIZE TABLE ... FINAL`) used to abort with `Logical error: 'Bad cast from type DB::ColumnVector<int> to DB::ColumnTuple'` inside `SerializationTuple::serializeBinaryBulkStatePrefix` — see the original AST-fuzzer report and stack trace in the issue.

The bug no longer reproduces on master — it was silently fixed by some sibling change in the `CLEAR COLUMN` / merge path. This patch only adds a regression test that locks in the current correct behaviour:

- the operation succeeds without exception,
- the inner `Tuple` elements are reset to default values, and
- the outer `Nested` array length stays aligned with the sibling `Nested` arrays (the invariant Avogar pointed out in https://github.com/ClickHouse/ClickHouse/issues/89537#issuecomment-3491360683 — *"they must keep the same number of values in arrays for all rows"*).

The test exercises the exact original repro and a multi-row, multi-part variant to harden the regression check against partial (per-part) reintroductions of the bug.

Verified locally on a Debug build: 50/50 passes with full settings randomization (`--test-runs 50`).

Requested by @PedroTadim in https://github.com/ClickHouse/ClickHouse/issues/89537#issuecomment-4404287026.

...

### Version info
- Merged into: `26.5.1.457`


groeneai:
**Pre-PR validation gate** (test-only PR for an already-fixed bug — Phase 4 gate adapted for the test-only case):

| # | Question | Answer |
|---|----------|--------|
| a | Deterministic repro? | The original AST-fuzzer query from the issue is the deterministic repro. Without the silent fix it abort with `Logical error: 'Bad cast from type DB::ColumnVector<int> to DB::ColumnTuple'` inside `SerializationTuple::serializeBinaryBulkStatePrefix`; with the silent fix it succeeds (verified locally on `26.5.1.1` debug build at `bca25b7b825`). |
| b | Root cause explained? | Per @Avogar's diagnosis in [the issue](https://github.com/ClickHouse/ClickHouse/issues/89537#issuecomment-3491360683): clearing a `Tuple`-typed `Nested` subcolumn produced a column of the wrong concrete type (`ColumnVector<int>` instead of `ColumnTuple`), which the merge writer then `assert_cast`'d into `ColumnTuple` — aborting in debug. Some sibling change made `CLEAR COLUMN` produce the correct concrete type while still preserving the `Nested` array-length invariant. |
| c | Fix matches root cause? | N/A — no fix in this PR; only a regression test. |
| d | Test intent preserved? | N/A (no existing test modified). The new test exercises the exact original repro plus a multi-row, multi-part variant so that any per-part reintroduction would be caught. |
| e | Both directions demonstrated? | Yes — historically: the original report has the full `<Fatal>` stack from `SerializationTuple.cpp:33`. Currently: 50/50 passes with full `--test-runs 50` randomization. |

Session: `cron:clickhouse-ci-task-worker:20260508-104500`

groeneai:
cc @PedroTadim @Avogar — could you review this regression test? It is the test you asked for in https://github.com/ClickHouse/ClickHouse/issues/89537#issuecomment-4404287026. Test-only PR, no source changes.

PR #104359, opened 2026-05-08
Revert "Parallelize reads from a single Parquet file in StorageFile"

alexey-milovidov:
Reverts ClickHouse/ClickHouse#104251

After further testing it appears that short queries degrade.

### Version info
- Merged into: `26.5.1.461`


alexey-milovidov:
The correctness issue should be fixed independently, after the revert.

alexey-milovidov:
This should prevent the regression: #104368

PR #104370, opened 2026-05-08
Validate hex digits in parseUUID (issue #86714)

### LINKED ISSUE #86714: toUUIDOrZero does not recognise invalid UUID
### Company or project name

_No response_

If string does not contain valid UUID, `toUUIDOrZero` returns `00000000-0000-0000-0000-000000000000`

At lest `25.6.2.5629`

### How to reproduce

25.6.2.5629

```
select 
	toUUIDOrZero('teststr-665785014160310158e823a3') as a, 
	toTypeName(a) as _type
format TSV

feefefef-6657-8501-4160-310158e823a3	UUID
```

### Expected behavior

```
select 
	toUUIDOrZero('teststr-665785014160310158e823a3') as a, 
	toTypeName(a) as _type
format TSV

00000000-0000-0000-0000-000000000000	UUID
```

_No response_

### Additional context

`toUUID`, `toUUIDOrDefault`, `toUUIDOrZero` behave the same way
If it's an expected behaviour, is it possible to introduce a new function which will be validating UUIDs?

groeneai:
Closes #86714.

## Problem

`parseUUID` did not validate that the input characters were valid hexadecimal digits. The lookup table mapped invalid characters to `0xff`, and `unhex2` then folded that back into `UInt8` arithmetic that wrapped around silently (for example `unhex2('s', 't')` = `0xff*16 + 0xff = 4335`, which truncates to `0xef`). As a result `toUUID*`, `CAST(... AS UUID)`, and any other caller of `parseUUID` happily accepted arbitrary 32- or 36-character garbage strings and fabricated a UUID from them.

```sql
WITH 'teststr-665785014160310158e823a3' AS str
SELECT toUUID(str), toUUIDOrNull(str), toUUIDOrZero(str);

-- Old behaviour: feefefef-6657-8501-4160-310158e823a3 for all three
-- New behaviour: throws CANNOT_PARSE_UUID; OrNull -> NULL; OrZero -> 00000000-0000-0000-0000-000000000000
```

## Fix

`parseUUID` now performs an explicit pre-validation pass:

* For 36-character inputs: dashes must be at positions 8, 13, 18, 23 and every other byte must be a hex digit `0-9 / a-f / A-F`.
* For 32-character inputs: every byte must be a hex digit.
* Other lengths still throw with the same error as before.

A new `tryParseUUID(span, UUID&)` returning `bool` is introduced for the no-throw path. `readUUIDTextImpl<bool>` (used by `tryReadUUIDText`, and therefore by `toUUIDOrNull / OrZero / OrDefault`, `CAST AS Nullable(UUID)`, `accurateCastOrNull`, JSONExtract on UUID, MsgPack input, ...) routes through it so invalid inputs surface as a clean `false` instead of an exception leaking through the try-path.

The fix lives at the parser layer, so all callers benefit at once: `toUUID*`, `CAST`, `SerializationUUID::deserializeTextQuoted` (fast path), `AvroRowInputFormat`, `JSONExtractTree`, `MsgPackRowInputFormat`, `ZooKeeperReplicator`.

## Tests

* All 17 UUID-related stateless tests still pass (`01528_to_uuid_or_null_or_zero`, `04137_conversion_functions_edges`, `00396_uuid`, `00472_compare_uuid_with_constant_string`, `00679_uuid_in_key`, `01032_cityHash64_for_UUID`, `01338_uuid_without_separator`, `01413_if_array_uuid`, `01453_normalize_query_alias_uuid`, `01554_bloom_filter_index_big_integer_uuid`, `02491_part_log_has_table_uuid`, `01746_convert_type_with_default`, `01533_distinct_nullable_uuid`, `03595_funcs_on_zero`, `03593_funcs_on_empty_string`, `03212_variant_dynamic_cast_or_default`, `00396_uuid_v7`).
* New regression test `04211_uuid_parser_strict_validation_86714.sql` covers the exact reproducer from the issue, all four `toUUID*` variants, multiple invalid-input shapes (full-garbage 32-char, 36-char with right dashes but non-hex elsewhere, misplaced dashes, single bad digit at start / end), valid 36-char / 32-char / mixed-case inputs, round-trip identity, `CAST AS UUID`, `CAST AS Nullable(UUID)`, and `accurateCastOrNull`. 30/30 passes with random settings enabled.

`toUUID`, `toUUIDOrNull`, `toUUIDOrZero`, `toUUIDOrDefault`, `CAST` to `UUID` and `Nullable(UUID)`, `accurateCastOrNull` to `UUID`, and the input formats that parse a UUID from text (`Avro`, `MsgPack`, `JSONExtract`, ...) now reject strings that have the right length but contain non-hexadecimal characters. Previously such inputs were silently turned into a fabricated UUID by walking off the end of the hex digit lookup table; now `toUUID` throws `CANNOT_PARSE_UUID`, and the `Or*` variants return `NULL` / the zero UUID / the supplied default.

### DIFF
--- a/src/IO/ReadHelpers.cpp
+++ b/src/IO/ReadHelpers.cpp
@@ -54,9 +54,20 @@ inline void parseHex(IteratorSrc src, IteratorDst dst)
         dst[dst_pos] = unhex2(reinterpret_cast<const char *>(&src[src_pos]));
 }
 
-UUID parseUUID(std::span<const UInt8> src)
+/// Returns true if all bytes in [begin, end) are valid hexadecimal digits (0-9, a-f, A-F).
+static inline bool areHexChars(const UInt8 * begin, const UInt8 * end)
 {
-    UUID uuid;
+    for (const UInt8 * p = begin; p != end; ++p)
+        if (unhex(static_cast<char>(*p)) == 0xff)
+            return false;
+    return true;
+}
+
+template <typename ReturnType>
+static ReturnType parseUUIDImpl(std::span<const UInt8> src, UUID & uuid)
+{
+    static constexpr bool throw_exception = std::is_same_v<ReturnType, void>;
+
     const auto * src_ptr = src.data();
     const auto size = src.size();
 
@@ -67,6 +78,21 @@ UUID parseUUID(std::span<const UInt8> src)
 #endif
     if (size == 36)
     {
+        /// Validate 8-4-4-4-12 layout: dashes at positions 8, 13, 18, 23 and hex digits in the rest.
+        if (src_ptr[8] != '-' || src_ptr[13] != '-' || src_ptr[18] != '-' || src_ptr[23] != '-'
+            || !areHexChars(src_ptr, src_ptr + 8)
+            || !areHexChars(src_ptr + 9, src_ptr + 13)
+            || !areHexChars(src_ptr + 14, src_ptr + 18)
+            || !areHexChars(src_ptr + 19, src_ptr + 23)
+            || !areHexChars(src_ptr + 24, src_ptr + 36))
+        {
+            if constexpr (throw_exception)
+                throw Exception(
+                    ErrorCodes::CANNOT_PARSE_UUID,
+                    "Cannot parse UUID from String: invalid format, expected 32 or 36 hexadecimal digits with dashes at positions 8, 13, 18, 23");
+            else
+                return ReturnType(false);
+        }
         parseHex<4>(src_ptr, dst + 8);
         parseHex<2>(src_ptr + 9, dst + 12);
         parseHex<2>(src_ptr + 14, dst + 14);
@@ -75,13 +101,39 @@ UUID parseUUID(std::span<const UInt8> src)
     }
     else if (size == 32)
     {
+        if (!areHexChars(src_ptr, src_ptr + 32))
+        {
+            if constexpr (throw_exception)
+                throw Exception(
+                    ErrorCodes::CANNOT_PARSE_UUID,
+                    "Cannot parse UUID from String: invalid format, expected 32 hexadecimal digits");
+            else
+                return ReturnType(false);
+        }
         parseHex<8>(src_ptr, dst + 8);
         parseHex<8>(src_ptr + 16, dst);
     }
     else
-        throw Exception(ErrorCodes::CANNOT_PARSE_UUID, "Unexpected length when trying to parse UUID ({})", size);
+    {
+        if constexpr (throw_exception)
+            throw Exception(ErrorCodes::CANNOT_PARSE_UUID, "Unexpected length when trying to parse UUID ({})", size);
+        else
+            return ReturnType(false);
+    }
 
+    return ReturnType(true);
+}
+
+UUID parseUUID(std::span<const UInt8> src)
+{
+    UUID uuid;
+    parseUUIDImpl<void>(src, uuid);
     return uuid;
+}
+
+bool tryParseUUID(std::span<const UInt8> src, UUID & uuid)
+{
+    return parseUUIDImpl<bool>(src, uuid);
 }
 
 void NO_INLINE throwAtAssertionFailed(const char * s, ReadBuffer & buf)
--- a/src/IO/ReadHelpers.h
+++ b/src/IO/ReadHelpers.h
@@ -707,6 +707,7 @@ inline bool tryReadDateText(ExtendedDayNum & date, ReadBuffer & buf, const DateL
 }
 
 UUID parseUUID(std::span<const UInt8> src);
+bool tryParseUUID(std::span<const UInt8> src, UUID & uuid);
 
 template <typename ReturnType = void>
 inline ReturnType readUUIDTextImpl(UUID & uuid, ReadBuffer & buf)
@@ -737,7 +738,15 @@ inline ReturnType readUUIDTextImpl(UUID & uuid, ReadBuffer & buf)
             }
         }
 
-        uuid = parseUUID({reinterpret_cast<const UInt8 *>(s), size});
+        if constexpr (throw_exception)
+        {
+            uuid = parseUUID({reinterpret_cast<const UInt8 *>(s), size});
+        }
+        else
+        {
+            if (!tryParseUUID({reinterpret_cast<const UInt8 *>(s), size}, uuid))
+                return ReturnType(false);
+        }
         return ReturnType(true);
     }
 
--- a/tests/queries/0_stateless/03773_deterministic_functions_key_condition.reference
+++ b/tests/queries/0_stateless/03773_deterministic_functions_key_condition.reference
@@ -23,7 +23,7 @@ VALUES (
 );
 INSERT INTO events (uuid, person_id, event, timestamp, team_id)
 VALUES (
-    '017adbc0-98b5-0000-3f74-619a368fe65g',
+    '017adbc0-98b5-0000-3f74-619a368fe65b',
     '017adbc0-98b5-0000-3f74-619a368fe65a',
     'page_view',
     toDateTime64('2026-01-01 00:00:01', 6, 'UTC'),
@@ -182,7 +182,7 @@ VALUES (
 );
 INSERT INTO events (uuid, person_id, event, timestamp, team_id)
 VALUES (
-    '017adbc0-98b5-0000-3f74-619a368fe65g',
+    '017adbc0-98b5-0000-3f74-619a368fe65b',
     '017adbc0-98b5-0000-3f74-619a368fe65a',
     'page_view',
     toDateTime64('2026-01-01 00:00:01', 6, 'UTC'),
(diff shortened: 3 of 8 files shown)


groeneai:
**Pre-PR validation gate** (per worker policy):

(a) **Deterministic repro?** ✓ `SELECT toUUID('teststr-665785014160310158e823a3')` reproduces 100% on demand. Before fix: returns `feefefef-6657-8501-4160-310158e823a3`. After fix: throws `CANNOT_PARSE_UUID`.

(b) **Root cause?** ✓ `parseUUID` called `parseHex<N>` which used `unhex2` without validating digits. The `hex_char_to_digit_table` maps invalid chars to `0xff`; multiplying that by `0x10` overflows a `UInt8` so e.g. `unhex2('s', 't')` = `0xff*16 + 0xff = 4335` → `0xef` (truncated). `parseHex<8>` happily wrote 8 bytes of garbage into the UUID.

(c) **Fix matches root cause?** ✓ Pre-validation in `parseUUID` / `tryParseUUID`: length check (32 or 36) + dash positions for 36-char + every non-dash byte must be a valid hex digit. Validation happens before any parsing.

(d) **Test intent preserved + new tests added?** ✓ 17 existing UUID stateless tests pass unchanged. New test `04211_uuid_parser_strict_validation_86714.sql` covers the reproducer, all four `toUUID*` variants, multiple invalid shapes, round-trips, and `CAST` / `accurateCastOrNull`. 30/30 runs pass with random settings.

(e) **Demonstrated in both directions?** ✓ Confirmed locally on debug build that the same query fails before fix, passes after fix.

(f) **Fix is general?** ✓ Single fix at the parser layer covers every caller of `parseUUID` / `tryParseUUID`: `toUUID*`, `CAST AS UUID`, `CAST AS Nullable(UUID)`, `accurateCastOrNull`, `SerializationUUID::deserializeTextQuoted` (fast path), `AvroRowInputFormat`, `JSONExtractTree`, `MsgPackRowInputFormat`, `ZooKeeperReplicator`. No symmetric variants left unfixed.

Session: cron:clickhouse-ci-task-worker:20260508-114500

groeneai:
cc @den-crane @KochetovNicolai @Avogar @rschu1ze — fix for #86714 per @den-crane's directive ("@groeneai try to fix it.").

The parser-level change makes `parseUUID` validate hex digits + dash positions before parsing, and adds `tryParseUUID` so the no-throw path stays exception-free. Could one of you take a look?

@KochetovNicolai you're listed as the assignee on the issue.
@Avogar / @rschu1ze tagging by component (comp-data-types / comp-regular-function — UUID parsing sits between the two).

den-crane:
Try testing how much performance is impacted by the new implementation using this query:

```
SELECT count() 
from numbers(1e8) 
where ignore(toUUIDOrNull('teststr-665785014160310158e823a3'))=0 
settings max_threads=1
format Null;

0 rows in set. Elapsed: 0.045 sec. Processed 100.00 million rows, 800.00 MB (2.25 billion rows/s., 17.97 GB/s.)
Peak memory usage: 126.64 KiB.
```

groeneai:
@den-crane — perf benchmarks below. **TL;DR**: your exact query is constant-folded so it shows ~no per-row parser cost; on per-row variants the new validation costs about **−16% on valid UUIDs** and is **+120% on invalid input** (early-bail).

### Setup

- Release build (clang-21, `-O3 -DNDEBUG`, no LTO/sanitizer/tests), 32-core x86 machine
- Same `build-release` directory, freshly rebuilt for each state (apply patch → bench → revert + rebuild → bench)
- `clickhouse client --progress=err` (numbers below are stderr `Progress:` lines, 5 runs each)

### Q1 — your exact query (constant-folded)

```sql
SELECT count() FROM numbers(1e8)
WHERE ignore(toUUIDOrNull('teststr-665785014160310158e823a3'))=0
SETTINGS max_threads=1 FORMAT Null;
```

| | master (no fix) | PR #104370 |
|---|---|---|
| run 1–5 (Grow/s) | 5.86 / 5.83 / 6.01 / 5.83 / 6.30 | 6.12 / 5.91 / 6.03 / 5.71 / 5.74 |
| **avg throughput** | **5.97 G rows/s** | **5.90 G rows/s** |
| query_log duration | 17 ms | 16 ms |
| query_log peak_mem | 195.95 KiB | 195.87 KiB |

Δ ≈ −1.2%, well within run-to-run noise (±5%). The constant-folder eliminates the parser from the hot loop, so this query measures `numbers() + ignore()=0 + count()`, not parsing. (My machine happens to be ~2.6× faster than yours; absolute numbers don't compare across hosts but the relative delta is what matters.)

### Q2 — invalid input forced per-row via `materialize()`

```sql
WHERE ignore(toUUIDOrNull(materialize('teststr-665785014160310158e823a3')))=0
```

| | master | PR #104370 |
|---|---|---|
| avg throughput | **42.79 M rows/s** | **94.92 M rows/s** |
| duration | 2335 ms | 1043 ms |

**+121.8%** (2.22× faster). The new code bails at byte 0 in `areHexChars` (`'t'` isn't hex), where the old code always ran the full `parseHex<8>×2` to fabricate the garbage UUID.

### Q3 — valid 36-char UUID forced per-row

```sql
WHERE ignore(toUUIDOrNull(materialize('a1b2c3d4-e5f6-1234-9876-abcdef012345')))=0
```

| | master | PR #104370 |
|---|---|---|
| avg throughput | **40.13 M rows/s** | **33.51 M rows/s** |
| duration | 2493 ms | 3003 ms (+20%) |

**−16.5%** — the new code does an extra full pass over the 32 hex bytes (plus 4 dash-position checks) before `parseHex` runs.

### Q4 — valid 32-char no-dash UUID forced per-row

```sql
WHERE ignore(toUUIDOrNull(materialize('a1b2c3d4e5f612349876abcdef012345')))=0
```

| | master | PR #104370 |
|---|---|---|
| avg throughput | **42.69 M rows/s** | **36.00 M rows/s** |
| duration | 2394 ms | 2775 ms (+16%) |

**−15.7%** — same `areHexChars(32)` overhead.

### Peak memory

Identical across both states for every query (deltas in the 0.04% range — just measurement noise). The patch is stack-only; no new allocations.

### Bottom line

The cost of correctness is one extra 32-byte hex-validation pass (`areHexChars`) before the (now bug-free) `parseHex` runs. In absolute terms the strict parser still runs at **~33–36 M valid UUIDs/sec single-threaded** — fast enough for any realistic workload. Invalid inputs are *faster* under the new code thanks to the early-bail.

If 16% on the valid hot path matters for any workload, two cheap follow-ups:

1. **SIMD `areHexChars`** — the loop is naturally vectorizable (SSE2 `pcmpgtb` / AVX2 lookup); could reclaim most of the overhead at the cost of ~30 lines.
2. **Fuse validation into `parseHex`** — check `unhex2` results for `0xff` instead of the separate up-front pass; would halve the byte-scan cost.

Happy to follow up on either if you'd like — otherwise the current code is, I think, a reasonable correctness/perf tradeoff.

groeneai:
**Fix-up commit pushed:** `8ec6f70b0b1` — fixes the PR-caused Fast test failure on `03773_deterministic_functions_key_condition[_explain]`.

### What broke

After my parseUUID validation patch, the existing 03773 test started failing on Fast test:

```
Code: 376. DB::Exception: Cannot parse UUID from String: invalid format,
expected 32 or 36 hexadecimal digits with dashes at positions 8, 13, 18, 23
(query: INSERT INTO events (uuid, person_id, ...) VALUES (
    '017adbc0-98b5-0000-3f74-619a368fe65g',  ← trailing 'g' is not hex
    '017adbc0-98b5-0000-3f74-619a368fe65a',
    ...))
```

This is exactly the bug we are fixing: the old `parseUUID` accepted `'017adbc0-98b5-0000-3f74-619a368fe65g'` and silently produced a fabricated UUID; the new strict parser correctly rejects it.

### Fix

The `g` was a typo — the test's queries filter only on `person_id`, never on `uuid`. The author likely meant any distinct hex digit so the second row's `uuid` would differ from the first row's `65d`. Replace `g` → `b`:

```diff
 INSERT INTO events (uuid, person_id, event, timestamp, team_id)
 VALUES (
-    '017adbc0-98b5-0000-3f74-619a368fe65g',
+    '017adbc0-98b5-0000-3f74-619a368fe65b',
     '017adbc0-98b5-0000-3f74-619a368fe65a',
```

Updated all four files (echo-mode test): both `.sql.j2` templates and both `.reference` files for the data + explain variants.

### Verification

- Local Debug build with the strict parseUUID patch.
- `tests/clickhouse-test --test-runs 30 03773_deterministic_functions_key_condition` → 60/60 passes (data + explain, randomized settings on).
- `04211_uuid_parser_strict_validation_86714` (the new regression test for this PR) — still passes 1/1.
- `count()` semantics preserved: both rows still insert (with valid distinct uuids), so `count() WHERE person_id < '...65e'` still returns 2.

This is the only test in the suite affected — no other test inputs malformed UUID strings.

Session: cron:clickhouse-ci-task-worker:20260509-104500

alexey-milovidov (approved):
Acceptable.

clickgapai:

**Affected code:** `parseUUID` in src/IO/ReadHelpers.cpp - core UUID parsing without hex validation — core code present in all supported branches.

**Why this might apply:** The `parseUUID` function in ReadHelpers.cpp has existed since ancient times. The bug causes silent data corruption (garbage strings produce fabricated UUIDs) which is P0 severity. All supported branches contain this code and should receive the fix.

PR #104374, opened 2026-05-08
Make AsyncReadCounters lockless

nikitamikhaylov:
The `std::mutex` in `AsyncReadCounters` is taken on every remote-FS read task and every prefetch submission, and shows up in production profiles of read-heavy workloads under `pthread_mutex_lock` in `DB::ThreadPoolRemoteFSReader::execute` and `DB::ThreadPoolRemoteFSReader::submit`.

The counters are only read by `dumpToMapColumn` for the `AsyncReadCounters` map in `system.query_log`, so per-field atomicity is enough. Replace the fields and the mutex with `std::atomic<size_t>` and a relaxed CAS loop for the "bump max" pattern.

- Performance Improvement

Reduced lock contention on the asynchronous remote-FS read path by making the per-query `AsyncReadCounters` lockless.

Made with [Cursor](https://cursor.com)

### Version info
- Merged into: `26.5.1.459`
- Backported to: `26.4.3.10`, `26.2.19.12`

### DIFF
--- a/src/Disks/IO/ThreadPoolRemoteFSReader.cpp
+++ b/src/Disks/IO/ThreadPoolRemoteFSReader.cpp
@@ -55,15 +55,13 @@ namespace
         explicit AsyncReadIncrement(std::shared_ptr<AsyncReadCounters> counters_)
             : counters(counters_)
         {
-            std::lock_guard lock(counters->mutex);
-            if (++counters->current_parallel_read_tasks > counters->max_parallel_read_tasks)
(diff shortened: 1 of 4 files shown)


Algunenano (approved):
Everything makes sense

PedroTadim:
Hello! Can a performance test be added?

Algunenano:
> Hello! Can a performance test be added?

Not really. The effort and setup to reproduce and test something like this is definitely not worth it

PR #104378, opened 2026-05-08
Fix `caseWithExpression` Decimal supertype mismatch with `transform`

### LINKED ISSUE #104335: Unit tests (tsan, function_prop_fuzzer)

Test name: Unit tests (tsan, function_prop_fuzzer)
CI report: [Unit tests (tsan, function_prop_fuzzer)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104207&sha=5d28f866f2fcc95da659a2cf1e71750d60db1b6e&name_0=PR&name_1=Unit%20tests%20%28tsan%2C%20function_prop_fuzzer%29)
Failing test history: [cidb](None)

Test output:
```
No test result file [/home/ubuntu/actions-runner/_work/ClickHouse/ClickHouse/ci/tmp/gtest.json]
```

### Version info
- Resolved by: #104378
- Merged into: `26.5.1.460`

groeneai:
Closes #104335.

CI report that surfaced this:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104207&sha=5d28f866f2fcc95da659a2cf1e71750d60db1b6e&name_0=PR&name_1=Unit%20tests%20%28tsan%2C%20function_prop_fuzzer%29

`caseWithExpression` builds the constant `Array(<then_supertype>)` it hands to
`transform` from the THEN values only, where `then_supertype =
getLeastSupertype(then_types)` -- the ELSE branch is excluded. `transform`'s
own return type is `getLeastSupertype(then_supertype, default_type)`, computed
inside `FunctionTransformOverloadResolver::buildImpl`, and it casts
`cache->to_column` and `cache->default_column` to that type. Meanwhile
`caseWithExpression` calls `transform->execute` with `result_type` from its
own `getReturnTypeImpl`, which is `getLeastSupertype([THENs..., default])`.

Whenever `getLeastSupertype` of the combined set is narrower than the chained
`getLeastSupertype(then_supertype, default)` -- which happens when the THENs
and the ELSE land in different `ColumnDecimal` storages -- the two
computations disagree, and `executeNumToNum` aborts with a `Bad cast`
`LOGICAL_ERROR`.

Concrete failure (from the fuzzer in #104335):

- THEN values `(UInt16, Int8)` give a then-only supertype of `Int32`.
- With an ELSE of `Decimal(9, 2)`, `getLeastSupertype(Int32, Decimal(9, 2))`
  is `Decimal(12, 2)` (stored as `Decimal64`).
- But `getLeastSupertype(UInt16, Int8, Decimal(9, 2))` is `Decimal(9, 2)`
  (stored as `Decimal32`).

So `cache->to_column` is a `ColumnDecimal<Decimal64>` while `executeNumToNum`
is dispatched with `T = ColumnDecimal<Decimal32>` (matching `column_result`),
and the `assert_cast<const T *>(cache->to_column.get())` in `transform.cpp:432`
aborts. This is the `Aborted, internal` `unit_tests_dbms` crash on PR #104207.

The fix is to include `args.back().type` (the ELSE branch type) when computing
`dst_supertype`, so the dst array passed to `transform` has an element type
wide enough that `getLeastSupertype(dst_supertype, default_type)` collapses to
our own `result_type`. The existing `tryGetLeastSupertype` /
`can_use_transform` guards already cover the case where no common supertype
exists.

The regression test exercises the minimised case (`UInt16` / `Int8` THENs +
`Decimal(9, 2)` ELSE), the original fuzzer query, plus `Decimal(38, 4)` and
`Float64` ELSE variants. It uses `clickhouse-local` because the server-side
analyzer constant-folds the all-constant expression before reaching
`transform`; `clickhouse-local` preserves the same code path the unit-test
fuzzer hit.

Verified locally: the new test fails on master without the patch (`Aborted`)
and passes 50 / 50 with the patch. All existing `transform` /
`caseWithExpression` tests still pass.

Pre-PR validation gate (worker self-check):

- (a) Deterministic repro on demand:
  `build/programs/clickhouse local --query "SELECT caseWithExpression(toUInt64(123), CAST(1 AS UInt32), CAST(31712 AS UInt16), CAST(2 AS UInt32), CAST(-8 AS Int8), CAST(-0.09 AS Decimal(9, 2)))"`
  -- aborts on master; returns `-0.09` with the patch.
- (b) Root cause: documented above (caseWithExpression / transform disagree
  on the ColumnDecimal storage when THENs and ELSE collapse to different
  `ColumnDecimal<>` widths).
- (c) Fix targets the cause: dst_supertype computation now matches the
  function's actual return type.
- (d) Test intent preserved + new test added: existing tests
  (`02787_transform_null`, `02798_generic_transform`,
  `04005_case_with_expression_nullable_nothing`,
  `00328_long_case_construction`, etc.) all still pass; new test covers the
  failing combination plus 5 sibling type combinations.
- (e) Both directions: shown above.
- (f) General fix: the bug is at the SOURCE of the dst supertype
  computation, not at the assert site. The matching condition in `transform`
  itself needs nothing -- it correctly uses
  `getLeastSupertype(dst_array_nested, default)`. The asymmetry was that
  `caseWithExpression` was constructing an artificial `dst_array_nested` that
  did not include the ELSE branch's contribution. Fixing that one place is
  enough; no symmetric site to patch.

Fix `Bad cast` `LOGICAL_ERROR` in `caseWithExpression` (and SQL `CASE expr WHEN ... THEN ... ELSE ...`) when the THEN-only supertype and the (THENs + ELSE) supertype land in different `ColumnDecimal` storages -- for example THEN values `(UInt16, Int8)` with an ELSE of `Decimal(9, 2)`. Closes [#104335](https://github.com/ClickHouse/ClickHouse/issues/104335).

### Version info
- Merged into: `26.5.1.460`


groeneai:
cc @rschu1ze @alexey-milovidov @Avogar — could you review this? It's a one-line fix in `caseWithExpression` for the `Bad cast` LOGICAL_ERROR in #104335: when THEN-only supertype + ELSE land in different `ColumnDecimal` widths (e.g. `(UInt16, Int8)` THENs + `Decimal(9, 2)` ELSE → `Int32` then-supertype, `Decimal(12, 2)/Decimal64` chained vs `Decimal(9, 2)/Decimal32` combined), `transform` and `caseWithExpression` disagreed on the result column type. Fix includes the ELSE in the dst-supertype computation. Includes a new `clickhouse-local`-based regression test.

groeneai:
Pre-PR validation gate (worker self-check), per [TASK.md](https://github.com/groeneai/ClickHouse) Phase 4 Step 9:

**(a) Deterministic repro on demand:**
```
build/programs/clickhouse local --query "SELECT caseWithExpression(toUInt64(123), CAST(1 AS UInt32), CAST(31712 AS UInt16), CAST(2 AS UInt32), CAST(-8 AS Int8), CAST(-0.09 AS Decimal(9, 2)))"
```
On master: `Aborted (core dumped)` with `Logical error: 'Bad cast from type DB::IColumn const* to DB::ColumnDecimal<DB::Decimal<int>> const*'`. With the patch: returns `-0.09` (the ELSE value, since none of the WHENs match). The minimal repro does not need ARRAY JOIN nor the huge UInt128 from the original fuzzer query — just `(UInt32, UInt16, UInt32, Int8, Decimal(9,2))` arg types is enough.

**(b) Root cause causal chain:** `caseWithExpression::executeImpl` builds `dst_supertype = getLeastSupertype(then_types)` and hands an `Array(dst_supertype)` to `transform`. `transform`'s overload resolver computes its own return type as `getLeastSupertype(dst_supertype, default_type)` and uses it to cast `cache->to_column`. Meanwhile `caseWithExpression`'s caller passes `result_type = getLeastSupertype([then_types..., default_type])` from `getReturnTypeImpl`. When THENs and the ELSE land in different `ColumnDecimal` storages (`(UInt16, Int8)` → `Int32`; chained `getLeastSupertype(Int32, Decimal(9, 2))` → `Decimal(12, 2)/Decimal64` vs combined `getLeastSupertype(UInt16, Int8, Decimal(9, 2))` → `Decimal(9, 2)/Decimal32`), `cache->to_column` is `ColumnDecimal<Decimal64>` while `column_result` is `ColumnDecimal<Decimal32>`, so the `assert_cast<const T *>(cache->to_column.get())` in `transform.cpp:432` aborts.

**(c) Fix matches root cause:** the patch corrects the `dst_supertype` computation at the source — including `args.back().type` (the ELSE type) — so the dst array type matches what `transform` will compute. Not a band-aid (no widening of asserts, no `no-random-*` tags).

**(d) Test intent preserved + new test added:** existing tests (`02787_transform_null`, `02798_generic_transform`, `04005_case_with_expression_nullable_nothing`, `00328_long_case_construction`, `02315_replace_multiif_to_if`, `02542_transform_new`, `03094_transform_return_first`, `03390_non_constant_case`, `01818_case_float_value_fangyc`) all still pass. New regression test `04211_case_with_expression_decimal_supertype_104335.sh` exercises:
- minimised `(UInt16, Int8)` + `Decimal(9, 2)` case
- match-on-WHEN cases (returns THEN value)
- the original fuzzer query
- `Decimal(38, 4)` ELSE variant (Decimal128 storage)
- `Float64` ELSE variant

**(e) Both directions demonstrated:** new test FAILS on master without patch (`Having 1 errors!`, server abort), PASSES 50/50 with patch.

**(f) Fix is general, not a narrow patch:** the bug is at the SOURCE of the dst supertype computation, not at the assert site. `transform` itself is correct — it uses `getLeastSupertype(dst_array_nested, default)`. The asymmetry was that `caseWithExpression` constructed an artificial `dst_array_nested` that excluded the ELSE branch's contribution. Searched for sibling sites (no other in-tree caller of `transform` builds the dst array this way; `caseWithExpression` is the only wrapper).

Session: cron:clickhouse-ci-task-worker:20260508-124500

alexey-milovidov (approved):
This is good.

clickgapai:

**Affected code:** `FunctionCaseWithExpression::executeImpl` supertype calculation — core code present in all supported branches.

**Why this might apply:** Bug in `FunctionCaseWithExpression::executeImpl` causes wrong results or LOGICAL_ERROR for common type combinations (integer THENs + Decimal ELSE). The affected code in `caseWithExpression.cpp` predates all supported branches (file has existed since 2019 `dbms/ → src/` rename). All users on any supported branch are affected when using CASE expressions with mixed integer/Decimal types.

PR #104395, opened 2026-05-08
Fix autorelease tagging an out-of-order commit on release branches

maxknv:
Adds sanity check to `tests/ci/create_release.py` to test if it's the latest release on the branch

### Version info
- Merged into: `26.5.1.514`


leshikus:
## Review: suggestions and risks

### Main problem: duplication between `auto_release.py` and `create_release.py`

`create_release.py` reproduces what `_read_cmake_version_field` already does in `auto_release.py` — read `cmake/autogenerated_versions.txt` at `origin/<branch>` and regex-extract `VERSION_<FIELD>`. The regex, the `git show` invocation, and the error message shape are duplicated.

This is the most important thing to fix before merging: with two copies, the regex / error-handling will drift, and the next person touching the cmake format has to remember to update both places. Extract once (possibly into `version_helper.py` next to `FILE_WITH_VERSION_PATH`) and import from both call sites.

### Other suggestions

- **`assert` for control flow in `create_release.py`.** `assert branch_patch_match` / `assert branch_githash_match` are stripped under `python -O`. Raise `RuntimeError` directly (matching the shape of `_read_cmake_version_field`) so the safety check survives optimization.
- **Lost log info.** The previous log line included `tagger.date` (`X commits ago, date Y`); the new one drops it. Not load-bearing, but occasionally useful when triaging stuck branches — could be recovered with `git show -s --format=%ci \$previous_release_sha`.
- **`commit_ref` resolution.** `git merge-base --is-ancestor <commit_ref> <prev_release_sha>` will accept any rev git can resolve (tags, short SHAs, refs on other branches). Worth a sanity check that `commit_ref` actually resolves to a commit reachable from `origin/<release_branch>` — otherwise a typo'd ref on a wrong branch could pass the ancestry check (because it's not an ancestor of prev) and proceed to tagging.
- **Test coverage.** No tests added. A small unit test on the cmake-parsing helper (covering `VERSION_PATCH`, `VERSION_GITHASH`, `VERSION_DESCRIBE`) would protect against future cmake formatting changes — especially since the regex is now load-bearing for releases.

### Risks

- **Cmake file format dependency.** The regex `SET\(VERSION_<field>\s+(\S+)\)` is now load-bearing. If anyone reformats `autogenerated_versions.txt` (e.g. lowercase `set(...)`, quotes, reflowed whitespace), autorelease silently breaks. Worth a comment in the cmake file pointing back here, or a stricter parser contract.
- **\`PATCH == 1\` heuristic.** Reliably identifies a fresh branch today, but it's an implicit invariant — if the version-bump policy ever changes (e.g. branches initialized with `PATCH = 0`), this skips silently. A check like `PATCH == 1 and GITHASH == \"<placeholder>\"` would be more defensive.
- **Inherent race unchanged.** Between reading the cmake file at branch tip and tagging, someone could push a new commit. Not introduced by this PR, but worth being aware of.

maxknv:
> Review: suggestions and risks

@leshikus suggestions are good. I limit the change to avoid taking on extra risk here. The patch now just adds a sanity check that no patch release has been applied to a later commit (it checks that the version file is the same at the branch tip). Adding tests and other hardenings is not reasonable for the current EOL scripts. They should be refactored or retired entirely to no longer depend on deprecated ./tests/ci modules, use stateless runners, and adopt the praktika workflow.

PR #104409, opened 2026-05-08
OvercommitTracker: synchronously purge jemalloc dirty pages on successful kill

nickitat:
- Improvement

Dirty pages are now synchronously freed after the query is killed by the `OvercommitTracker`.

---

When a query is selected for `OvercommitTracker` termination and its memory is released, the freed extents stay dirty in jemalloc until `dirty_decay_ms` (5s default) elapses or `MemoryWorker` triggers a purge on its next iteration. During that window, the kernel-visible RSS stays at the limit even though `MemoryTracker` now believes there is headroom. Other queries trying to allocate hit the RSS check in `MemoryTracker::allocImpl` and OOM cascade.

After `OvercommitTracker` confirms enough memory has been freed, drop the internal lock and synchronously purge dirty pages so the next allocation from a waiting query sees real RSS rather than the lagging-decay value.

The lock is released before the (potentially multi-second) purge so other queries can enter the `OvercommitTracker` path during the drain.

---

Some test results:

```
  ▎ Setup
  ▎
  ▎ - 16 GiB max_server_memory_usage (intentionally tight)
  ▎ - 4 concurrent clients running SELECT URL, COUNT(*) FROM hits GROUP BY URL ORDER BY c DESC LIMIT 10 against ClickBench 100M-row hits dataset
  ▎ - 60-second runs × 5 iterations per binary, repeated twice = 10 iterations / 600 seconds per binary
  ▎ - Server pinned to NUMA-0 + HT (cores 0-31,96-127), spilling disabled
  ▎ - Hardware: Intel Xeon 6975P-C (Granite Rapids), 192 vCPU
  ▎
  ▎ Results across 10 iterations
  ▎
  ▎ |                      | master | master_fix1 | delta |
  ▎ |----------------------|--------|-------------|-------|
  ▎ | Successful queries   | 109    | 122         | +12%  |
  ▎ | Failed queries (OOM) | 650    | 526         | −19%  |
  ▎ | Throughput (qps)     | 0.18   | 0.20        | +11%  |
```

### DIFF
--- a/src/Common/OvercommitTracker.cpp
+++ b/src/Common/OvercommitTracker.cpp
@@ -1,5 +1,6 @@
 #include <Common/OvercommitTracker.h>
 
+#include <Common/Jemalloc.h>
 #include <Common/ProfileEvents.h>
 #include <Common/CurrentMetrics.h>
 #include <Interpreters/ProcessList.h>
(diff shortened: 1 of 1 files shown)


nickitat:
I'm not really advocating for this fix. But Claude seems to be stoked about it.

src/Common/OvercommitTracker.cpp
```diff
@@ -102,11 +103,43 @@ OvercommitResult OvercommitTracker::needToStopQuery(MemoryTracker * tracker, Int
...
+    ///   - concurrent per-waiter purges parallelise the drain work, while
+    ///     a single serialised purge per wave is bigger and not pipelined;
+    ///   - more importantly, a waiter that skipped the dedup'd purge would
+    ///     return immediately and attempt its next allocation against a
+    ///     still-stale RSS view, which would then fail.
+    /// Per-waiter purging is wasted-work-bounded by the wave size, runs in
+    /// parallel, and guarantees every waiter sees a fresh RSS before its
+    /// next allocation.
+    if (result == OvercommitResult::MEMORY_FREED)
+        DB::Jemalloc::purgeArenas();
```
azat, line 139: I guess it make sense, but let's do this via `MemoryWorker`, since even after purging arenas we need to make sure that RSS has been updated, right?

Also maybe it make sense to purge custom arenas (`DB::JemallocMergeTreeArena::purge`) in `purgeArenas`,

nickitat: > let's do this via MemoryWorker

The whole point is to do it synchronously as soon as we freed the memory and not to wait for MemoryWorker or decay timeouts inside jemalloc.

I tried updating RSS right after purging, but it consistently shown negative effects:

```
  Same pattern as the previous attempt with total_memory_tracker.get() — master clean,
  patched binary cascades hard. The rss-update itself is harmful regardless of which value
  we write.

  Mechanism I now believe:

  1. MemoryWorker periodically writes MemoryTracker.rss from jemalloc.stats.resident —
  includes dirty. That value is conservative (higher than just-active), which makes
  allocImpl reject sooner.
  2. Our purge drops actual kernel RSS. Right after purge, active ≈ resident ≈ low.
  3. We overwrite MemoryTracker.rss with this low value.
  4. In the window between our write and MemoryWorker's next tick, other threads see rss =
  low and pass their allocImpl rss-check. They allocate.
  5. Their allocations re-grow dirty (since the workload is heavy churn). Real kernel RSS
  climbs back, but MemoryTracker.rss stays at our low write until MemoryWorker's next tick.
  6. Eventually allocations push past the real ceiling. OvercommitTracker fires repeatedly.
  Cascade.

  So the original "stale rss" complaint was real, but the stale-high value was actually
  protective — it prevented over-commitment. Our refresh removes that protection.

  Two consecutive tests (different "fresh" values: total_memory_tracker.get() and
  approximate_stats.active) both regressed identically. That's not variance — that's the
  pattern.

  Recommendation: revert the rss-update entirely. Keep only the synchronous purgeArenas()
  call.
```

> Also maybe it make sense to purge custom arenas

Claude claims that they are already included

azat: > The whole point is to do it synchronously as soon as we freed the memory and not to wait for MemoryWorker or decay timeouts inside jemalloc.

Yep, I suggested to update rss/amount of the total_memory_tracker here as well, not doing purge of jemalloc via MemoryWorker

>I tried updating RSS right after purging, but it consistently shown negative effects:

I've briefly the claude output, but AFAICS this - it won't help w/o updating RSS, it will still need to wait for MemoryWorker thread, because it is not enough to purge jemalloc allocations you also need to update RSS, otherwise allocations will still fail

Or maybe I am missing something? Maybe there was some real use case where this will help?

nickitat: You can see in the [description](https://github.com/ClickHouse/ClickHouse/pull/104409#issue-4408037553) how the tests were done. It is the only use case I've tested.

It is not easy to say, how exactly this positive effect is achieved. At least it is clear that by the time MemoryWorker will do its next iteration we can have [some of] those pages already purged. It can make recovery faster by a few seconds.

PR #104411, opened 2026-05-08
Reorder operations in DiskAccessStorage for consistency: always modify files first

vitlibar:

Reorder operations in `DiskAccessStorage` for consistency: always modify files first

### Version info
- Merged into: `26.5.1.621`

### DIFF
--- a/src/Access/DiskAccessStorage.cpp
+++ b/src/Access/DiskAccessStorage.cpp
@@ -27,6 +27,7 @@ namespace ErrorCodes
 {
     extern const int DIRECTORY_DOESNT_EXIST;
     extern const int FILE_DOESNT_EXIST;
+    extern const int LOGICAL_ERROR;
 }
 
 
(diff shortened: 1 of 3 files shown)


src/Access/DiskAccessStorage.cpp
```diff
@@ -509,28 +631,34 @@ bool DiskAccessStorage::removeImpl(const UUID & id, bool throw_if_not_exists)
...
     }
     AccessEntityType type = entity->getType();
 
     if (readonly)
         throwReadonlyCannotRemove(type, entity->getName());
 
-    /// Do removing.
-    memory_storage.remove(id, /* throw_if_not_exists= */ false);
-
+    /// Step 2: Modify files first.
```
vitlibar, line 664: The main idea of this PR is that we must modify files first, and internal memory storage after that - so if a failure appears while we're modifying files then the memory storage won't be modified and `DiskAccessStorage` will be in consistent state

alexey-milovidov (approved):

PR #104415, opened 2026-05-08
Create system.iceberg_files table

asya-ch:
- New Feature

### Changelog entry (a [user-readable short description] :
Added a new system table system.iceberg_files exposing per-file metadata for Iceberg tables, with one row per data or delete file in each table's current snapshot. This closes #98777.

Also, serialize format_version in Iceberg writer.

### Version info
- Merged into: `26.6.1.614` (included in `26.6` and later)
- Backported to: `26.5.8.4`


src/Storages/System/StorageSystemIcebergFiles.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+        /// Iceberg tables can be broken in arbitrary ways; tolerate any error from a single table
+        /// instead of failing the whole query.
+        try
+        {
+            if (IcebergMetadata * iceberg_metadata = dynamic_cast<IcebergMetadata *>(object_storage->getExternalMetadata(context_copy)); iceberg_metadata)
+            {
+                IcebergMetadata::IcebergFiles files = iceberg_metadata->getFiles(context_copy);
+
+                for (auto & file : files)
```
Ergus, line 156: This for is inside a try-catch inside a lambda. If any of the per-column insertions call throws after column_index has been advanced the next iteration on a different table will then missalign rows for the rest of the query.

(I know a pattern like this already exists in code like `StorageSystemIcebergHistory.cpp`, but better if we can avoid that)

Just copy `column_index` of every column before the row, and on exception roll back any columns whose size grew.

src/Storages/System/StorageSystemIcebergFiles.cpp
```diff
@@ -0,0 +1,157 @@
...
+    };
+
+    const bool show_tables_granted = access->isGranted(AccessType::SHOW_TABLES);
+
+    if (show_tables_granted)
+    {
+        auto databases = DatabaseCatalog::instance().getDatabases(GetDatabasesOptions{.with_datalake_catalogs = true});
+        for (const auto & db : databases)
+        {
+            for (auto iterator = db.second->getTablesIterator(context_copy, {}, true); iterator->isValid(); iterator->next())
```
Ergus: IIUC the predicate arguments in `fillColumn` above are ignored. For simpler system tables that's fine, however, in this case that means that we are not filtering table names at all at this point in `getTablesIterator`. 

So if a user makes a `SELECT ... WHERE database = 'foo' AND table = 'bar'...` is only applied later by the SQL engine, after fillData has already produced rows for every Iceberg table in every database.

This means that `add_file_records` will be called for every single table and that is in general expensive because it implies a metadata reflesh (`configuration->update(object_storage, query_context)`) for each of them even if they will be filtered out latter.

system.iceberg_history has the same gap, but the per-table cost is just metadata.json  annd no manifest traversal.

For example:
Imagine a deployment with 200 Iceberg tables (each ~50 manifest files) plus 50 Delta tables. A user runs:

  SELECT * FROM system.iceberg_files
  WHERE database = 'analytics' AND table = 'events'

The user asked for one table; the server did the work for 200. And because the cache is disabled, repeating the query does not amortize.

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
```diff
@@ -292,6 +281,7 @@ void generateManifestFile(
     Poco::JSON::Stringifier::stringify(partition_spec->getArray(Iceberg::f_fields), oss_partition_spec);
     writer.setMetadata(Iceberg::f_partition_spec, oss_partition_spec.str());
     writer.setMetadata(Iceberg::f_partition_spec_id, std::to_string(partition_spec_id));
+    writer.setMetadata(Iceberg::f_format_version, std::to_string(version));
```
Ergus, line 298: Is this change necessarily part of this PR?

Is this covered in `tests/queries/0_stateless/04203_system_iceberg_files.sh`

Ergus: BTW what happens with empty tables in this case? Is there a test for that?

Ergus (changes requested):
Please fix the per-row partial-insertion on exception (or explicitly accept it as a known issue carried over from iceberg_history).

Mention the fix in `IcebergWrites.cpp` as part of the PR description (for changelog) or add it in a different PR with it's tests.

The comment avout the poredicate pushdown would be very nice to have in order to prevent future issues.

src/Storages/System/StorageSystemIcebergFiles.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace DB
+{
+
+class Context;
+
+/** System table that lists data and delete files of currently loaded Iceberg tables.
+  * Each row corresponds to one file referenced by the current snapshot's manifest entries.
+  */
+class StorageSystemIcebergFiles final : public IStorageSystemOneBlock
```
alesapin: Let's try to avoid `IStorageSystemOneBlock`, table may have millions of files. Take a look at `SystemFilesystemCacheSource`.

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
asya-ch: The change is necessary to get the correct content type -- our logic that reads the content type depends on the written format version. It was missing in our writes, so without it, we'd get the incorrect content type. 

I'll remove it and add it to a separate PR.

asya-ch: Not sure about empty tables -- I'll add a test.

Ergus: > The change is necessary to get the correct content type -- our logic that reads the content type depends on the written format version. It was missing in our writes, so without it, we'd get the incorrect content type. I can put it in a different PR if that is preferred.

IMO Not need a different PR. Just mention it in the PR description and for a changelog entry.

src/Storages/System/StorageSystemIcebergFiles.cpp
asya-ch: I'll try to add a per-database and per-table cache, following the pattern in system.tables.

Ergus: I could be wrong, but I think that using the ignored predicates arguments it will be simpler and enough for this use case. Just skipping the non-matching tables is simple and typical pattern.

Ergus:
Hi @asya-ch 

Please avoid rebassing after a review started ;). It is simpler to see what changed since the last review. To get master changes just merge master. I know history looks ugly then... but we learn to live with it ;p

src/Storages/System/StorageSystemIcebergFiles.cpp
```diff
@@ -0,0 +1,459 @@
...
+            MutableColumnPtr nested = ColumnArray::create(ColumnTuple::create(std::move(tuple_columns)));
+            return ColumnMap::create(std::move(nested));
+        };
+
+        MutableColumnPtr col_null_value_counts = make_int32_int64_map();
+        MutableColumnPtr col_column_sizes = make_int32_int64_map();
+        MutableColumnPtr col_value_counts = make_int32_int64_map();
+        MutableColumnPtr col_equality_ids = ColumnArray::create(ColumnInt32::create());
+
+        std::vector<IColumn *> col_ptrs{
```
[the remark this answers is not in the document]
asya-ch: I don't think this is correct, the build is working fine.

src/Storages/System/StorageSystemIcebergFiles.cpp
```diff
@@ -0,0 +1,381 @@
...
+            for (size_t i = 0; i < col_ptrs.size(); ++i)
+                checkpoints[i] = col_ptrs[i]->getCheckpoint();
+
+            /// Iceberg tables can be broken in arbitrary ways; tolerate any error from a single table
+            /// instead of failing the whole query.
+            try
+            {
+                if (IcebergMetadata * iceberg_metadata = dynamic_cast<IcebergMetadata *>(object_storage->getExternalMetadata(context_copy)); iceberg_metadata)
+                {
+                    IcebergMetadata::IcebergFiles files = iceberg_metadata->getFiles(context_copy);
```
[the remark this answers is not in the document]
asya-ch: I think this is OK for now. We already have several calls to `getManifestFileEntriesHandle()` which parse and keep in-memory data for entire manifest files.

If someone has particularly huge manifest files, and this causes issues for them, we can consider implementing buffered reading from manifest files.

fikadufikaduteasfaye-stack:
Please support thanks

src/Storages/System/StorageSystemIcebergFiles.cpp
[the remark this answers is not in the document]
asya-ch: I am aware it is still live. It is out of scope for this PR.

src/Storages/System/StorageSystemIcebergFiles.cpp
```diff
@@ -0,0 +1,490 @@
...
+        if (current_database_iterator != databases.end())
+            current_table_iterator = current_database_iterator->second->getTablesIterator(context_copy, {}, true);
+    }
+
+    String getName() const override { return "SystemIcebergFilesSource"; }
+
+protected:
+    Chunk generate() override
+    {
+        if (!access->isGranted(AccessType::SHOW_TABLES))
```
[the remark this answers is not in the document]
asya-ch: done

alesapin (approved):

Ergus (approved):
The `getFilesForManifest` concern is still valid, but I agree it is out of scope.

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergMetadata.h
```diff
@@ -32,10 +32,29 @@
...
+    Int64 record_count;
+    Int64 file_size_in_bytes;
+    String partition;
+    Int32 schema_id;
+    Int64 sequence_number;
+    std::optional<Int32> sort_order_id;
+    std::map<Int32, Int64> null_value_counts;
+    std::map<Int32, Int64> column_sizes;
+    std::map<Int32, Int64> value_counts;
+    std::optional<std::vector<Int32>> equality_ids;
```
Ergus: Do we really need an optional here? An empty vector is not enough?

asya-ch: Not needed, removed the optional.

docs/en/operations/system-tables/iceberg_files.md
```diff
@@ -0,0 +1,32 @@
+---
+description: 'System table containing per-file metadata of Iceberg tables'
```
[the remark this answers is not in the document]
asya-ch: Very few of the system tables have `sidebar_label` and `sidebar_position`. Guessing repo docs are out of date, will not address.

asya-ch:
We only need to backport to 26.4, closing backports to earlier versions.

scanhex12:
Why do we backport the feature?

asya-ch:
@scanhex12, I'll send you the info offline. It's for an internal project.

PR #104425, opened 2026-05-08
PromQL: Fix aggregation operator for empty vector

vitlibar:

PromQL: Fix aggregation operator for empty vector

### Version info
- Merged into: `26.5.1.471`

### DIFF
--- a/src/Storages/TimeSeries/PrometheusQueryToSQL/applyAggregationOperatorQuantile.cpp
+++ b/src/Storages/TimeSeries/PrometheusQueryToSQL/applyAggregationOperatorQuantile.cpp
@@ -130,6 +130,12 @@ SQLQueryPiece applyAggregationOperatorQuantile(
         if (operator_node->by || operator_node->without)
             builder.group_by.push_back(make_intrusive<ASTIdentifier>(ColumnNames::NewGroup));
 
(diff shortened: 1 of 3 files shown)


Ergus (approved):
LGTM. 

Just a comments: Consider adding one test exercising the by/without empty-input path for completeness, since  the test only exercises the nogrouping case.

clickgapai:
Hi — could this be a backport candidate for `26.4`? No backport label was found, and the affected code appears to exist in those branches based on a file/function presence check.

**Affected code:** `applyOneArgumentAggregationOperator` and `applyAggregationOperatorQuantile` in PromQL query converter — introduced in 26.4.

**Why this might apply:** The affected code (`applyOneArgumentAggregationOperator` and `applyAggregationOperatorQuantile`) was introduced on 2026-03-30 and 2026-04-01 respectively, after 26.3 release (2026-03-27) but before 26.4 release (2026-05-05). Git log shows commit d47a8fd00f6 introduced the one-argument operator and 04029b2d38f introduced quantile operator. Only 26.4 contains this buggy code.

Other supported branches (`26.3`, `26.2`, `25.8`) predate 26.4 and do not contain this code.

Ergus:
@vitlibar DO you think that this needs to be backported?

PR #90202, opened 2025-11-17
Fix hive partitioning incompatibility

kssenii:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix hive partitioning incompatibility preventing smooth upgrade in 25.8 (fixes error `All hive partitioning columns must be present in the schema` during upgrade).

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #104433, opened 2026-05-08
CAST to DateTime / DateTime64 without a time zone keeps the source time zone

### LINKED ISSUE #55072: CAST to DateTime or DateTime64 without a timezone should keep the timezone of its argument
**Describe the issue**

```
milovidov-desktop :) WITH toDateTime('2023-01-02 03:04:05', 'America/Los_Angeles') AS x SELECT toTypeName(x), toTypeName(x::DateTime)

WITH toDateTime('2023-01-02 03:04:05', 'America/Los_Angeles') AS x
SELECT
    toTypeName(x),
    toTypeName(CAST(x, 'DateTime'))

Query id: c2bc74fe-3860-440e-9044-9f8f52a7fbc1

┌─toTypeName(x)───────────────────┬─toTypeName(CAST(x, 'DateTime'))─┐
│ DateTime('America/Los_Angeles') │ DateTime                        │
└─────────────────────────────────┴─────────────────────────────────┘
```

alexey-milovidov:
Closes #55072.

When casting to `DateTime` or `DateTime64` without an explicit time zone,
the time zone of the source argument is preserved (when it is itself a
`DateTime` or `DateTime64` with an explicit time zone). This matches the
long-standing behavior of the `toDateTime` and `toDateTime64` functions.

Before:
```
:) WITH toDateTime('2023-01-02 03:04:05', 'America/Los_Angeles') AS x SELECT toTypeName(x), toTypeName(CAST(x, 'DateTime'))

┌─toTypeName(x)───────────────────┬─toTypeName(CAST(x, 'DateTime'))─┐
│ DateTime('America/Los_Angeles') │ DateTime                        │
└─────────────────────────────────┴─────────────────────────────────┘
```

After:
```
:) WITH toDateTime('2023-01-02 03:04:05', 'America/Los_Angeles') AS x SELECT toTypeName(x), toTypeName(CAST(x, 'DateTime'))

┌─toTypeName(x)───────────────────┬─toTypeName(CAST(x, 'DateTime'))─┐
│ DateTime('America/Los_Angeles') │ DateTime('America/Los_Angeles') │
└─────────────────────────────────┴─────────────────────────────────┘
```

The change applies to user-facing `CAST`, `accurateCast`, and
`accurateCastOrNull`. The internal `_CAST` used by ClickHouse to enforce
column types against a target schema is excluded, since it is constructed
from the column's full type name (which already includes the explicit
time zone, when one is set).

The substitution traverses `Nullable` and `LowCardinality` wrappers in
the target type, so e.g. `CAST(x, 'Nullable(DateTime)')` and
`CAST(toNullable(x), 'DateTime')` also propagate the source time zone.

`CAST` to `DateTime` or `DateTime64` without an explicit time zone now preserves the time zone of its source argument (when the source is a `DateTime`/`DateTime64` with an explicit time zone), matching the behavior of the `toDateTime`/`toDateTime64` functions. Closes [#55072](https://github.com/ClickHouse/ClickHouse/issues/55072).

### Version info
- Merged into: `26.5.1.562`


alexey-milovidov:
@groeneai, provide a fix for `Logical error: Default functions implementation for LowCardinality is supported only with a single LowCardinality argument. (STID: 5173-5057)`

groeneai:
@alexey-milovidov — investigated STID 5173-5057. tl;dr: this looks like a chronic latent bug in the default `LowCardinality` handling, **not caused by this PR**.

**CIDB evidence (`default.checks`, full retention since 2025-11)**

| event | PR | branch | check | when |
|---|---|---|---|---|
| only 2 hits, ever | 104433 | `cast-datetime-keep-timezone` | `AST fuzzer (arm_asan_ubsan)` | 2026-05-10 10:14:25 |
|  | 90202 | `fix-hive-partitioning-incompatibility` | `AST fuzzer (amd_ubsan)` | 2025-11-17 21:45:36 |

- **0 hits on master** in 6 months for the exact error message.
- Two PR hits, on **unrelated branches** (timezone CAST vs Hive partitioning), separated by ~6 months — pattern is "AST fuzzer occasionally generates a query that trips the throw site."
- Different fuzzed queries each time:
  - PR #90202: `EXPLAIN ESTIMATE SELECT toDateTime('2017-01-01 00:00:00') + interval AS res, (SELECT toIntervalYear(5)) AS interval`
  - PR #104433: `concatAssumeInjective(materialize(toLowCardinality(2)), isNotNull(257), toUInt16(...), (SELECT toLowCardinality(9223372036854775806)))` inside a `UNION DISTINCT`.

**Why the diff in this PR doesn't cause it**

The PR adds one block in `CastOverloadResolverImpl::buildImpl` that fires only when:
- the call is not internal `_CAST`, AND
- `getExplicitTimeZoneOfDateTimeArgument(arguments.front().type)` returns a non-empty timezone.

`getExplicitTimeZoneOfDateTimeArgument` returns empty unless the source argument (after stripping `Nullable` and `LowCardinality`) is a `DateTime` or `DateTime64` with an explicit time zone. The fuzzed failing query has no `CAST` to `DateTime`/`DateTime64` and no `DateTime` source argument — `getExplicitTimeZoneOfDateTimeArgument` returns `""` for every `CAST` the analyzer might inject (e.g. `CAST(x, 'String')` for `concat` argument coercion), so `substituteTimeZoneInDateTimeType` is never called and `type` is never mutated. The new code path is inert for this query.

**The actual throw site**

`src/Functions/IFunction.cpp:82` in `replaceLowCardinalityColumnsByNestedAndGetDictionaryIndexes`, called from `IExecutableFunction::executeWithoutSparseColumns:472` when:
1. `useDefaultImplementationForLowCardinalityColumns()` is `true` (default — `concat`/`concatAssumeInjective` doesn't override it), AND
2. `result_type` was wrapped in `LowCardinality` by `IFunctionOverloadResolver::getReturnType` (line 731), AND
3. at execute time, **two or more arguments are seen as `ColumnLowCardinality` directly** (not `ColumnConst(ColumnLowCardinality)`).

There is a build-time vs run-time mismatch: `getReturnType` only counts non-const args toward `num_full_low_cardinality_columns` (line 711–713), so it's happy to wrap the result in `LowCardinality` when there is only one non-const LC arg. But `replaceLowCardinalityColumnsByNestedAndGetDictionaryIndexes` (line 78) treats *any* `ColumnLowCardinality` it encounters as a "full LC argument", and throws on the second one. If a scalar subquery / materialized constant arrives at execute time as a full LC column rather than `Const(LC)`, the invariant breaks and we hit this throw.

**Local reproduction**

I tried reproducing on master HEAD (`c142440e077`) by:
- replaying the full `reproduce_commands.sql` (4786 lines) preamble + final query;
- running the simpler PR #90202 variant (`SELECT toDateTime('2017-01-01 00:00:00') + interval, (SELECT toIntervalYear(5))`);
- a few hand-crafted `concatAssumeInjective` variants with multiple LC args.

All pass on master HEAD. The trigger requires a specific server-side-fuzzer state (settings / random seed / accumulated table state) that I couldn't isolate locally with the provided repro file. Without a deterministic local repro, per our pre-PR validation gate I can't safely propose a fix for it.

**Recommendation**

I'd suggest **not blocking this PR on STID 5173-5057** — it's a pre-existing chronic latent bug. I'm happy to file an issue for it and pursue the fix independently (likely in `replaceLowCardinalityColumnsByNestedAndGetDictionaryIndexes`: when more than one full LC column is encountered, fall back to `convertLowCardinalityColumnsToFull` instead of throwing, since the result wrapping logic already handles the LC re-wrap when needed).

Worker session: `cron:clickhouse-ci-task-worker:20260510-121500`.

yariks5s (approved):

PR #104442, opened 2026-05-09
Support array paths in  table function `file`

### LINKED ISSUE #53176: `file`-like table functions and engines to take Array argument
**Use case**

Allow to write

```
SELECT * FROM file(['hello.csv', 'world.csv'])
```

instead of

```
SELECT * FROM file('{hello,world}.csv')
```

This will make it composable - the array can be produced by another function.

niyue:
- New Feature

### Changelog entry:
* Allow the `file` table function to accept an `Array(String)` of paths in `SELECT` queries.

* [x] Documentation is written (mandatory for new features)

Closes https://github.com/ClickHouse/ClickHouse/issues/53176

### Version info
- Merged into: `26.5.1.486`


tests/queries/0_stateless/04212_file_table_function_array_argument.sql
```diff
@@ -0,0 +1,38 @@
+INSERT INTO FUNCTION file(currentDatabase() || '_file_array_arg_1.csv', 'CSV', 'x UInt64')
```
alexey-milovidov, line 1: Tests should include a non-trivial case when one of the array elements has wildcards, and another does not.

src/TableFunctions/TableFunctionFile.cpp
```diff
@@ -57,7 +125,7 @@ void TableFunctionFile::parseFirstArguments(const ASTPtr & arg, const ContextPtr
             throw Exception(ErrorCodes::BAD_ARGUMENTS, "File descriptor must be non-negative");
     }
     else
-        throw Exception(ErrorCodes::BAD_ARGUMENTS, "The first argument of table function '{}' mush be path or file descriptor", getName());
+        throw Exception(ErrorCodes::BAD_ARGUMENTS, "The first argument of table function '{}' must be path or file descriptor", getName());
```
alexey-milovidov, line 129: 👍

src/TableFunctions/TableFunctionFile.cpp
```diff
@@ -26,17 +27,84 @@ namespace ErrorCodes
...
+
+}
+
 void TableFunctionFile::parseFirstArguments(const ASTPtr & arg, const ContextPtr & context)
 {
+    const auto * literal = arg->as<ASTLiteral>();
+    if (!literal)
+        throw Exception(
+            ErrorCodes::BAD_ARGUMENTS,
+            "The first argument of table function '{}' must be a literal path, array of paths, or file descriptor",
```
alexey-milovidov: Or an expression. We also support them, iirc.

src/TableFunctions/TableFunctionFile.cpp
```diff
@@ -26,17 +27,84 @@ namespace ErrorCodes
...
+            throw Exception(
+                ErrorCodes::BAD_ARGUMENTS,
+                "All elements of the first argument of table function 'file' must have type String, got {}",
+                fieldTypeToString(source_field.getType()));
+
+        auto source = StorageFile::FileSource::parse(source_field.safeGet<String>(), context);
+        if (source.archive_info)
+            throw Exception(
+                ErrorCodes::BAD_ARGUMENTS,
+                "Array source for table function 'file' does not support archive path syntax");
```
alexey-milovidov, line 53: What is the reason for that?

tests/queries/0_stateless/04212_file_table_function_array_argument.sql
alexey-milovidov: Also, with different formats, please.

src/TableFunctions/TableFunctionFile.cpp
niyue: Supporting archive syntax in `Array(String)` sources would require changing the `FileSource` representation, specifically the current `std::optional<ArchiveInfo> archive_info`.

For example:

`file(['a.zip::x.csv', 'b.zip::y.csv'])`

means `a.zip -> x.csv` and `b.zip -> y.csv`. However, the current `ArchiveInfo` stores only one `path_in_archive` and one optional filter, so it can represent multiple archives sharing the same inner path pattern, but not multiple independent archive-to-inner-path mappings.

To keep this PR focused and avoid reading incorrect files, the current implementation fails fast when archive syntax is used inside an `Array(String)` source.

Please let me know if archive syntax support for array sources should be included in this PR.

src/TableFunctions/TableFunctionFile.cpp
```diff
@@ -26,17 +27,84 @@ namespace ErrorCodes
...
+        else if (result.format_from_filenames != source.format_from_filenames)
+        {
+            has_different_formats = true;
+        }
+    }
+
+    if (has_different_formats)
+        result.format_from_filenames = {};
+
+    result.with_globs = result.paths.size() > 1;
```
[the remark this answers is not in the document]
niyue: Fixed by marking `with_globs = true` for array input, and a dedicated arrray with single element test case is added to verify this change.

src/TableFunctions/TableFunctionFile.cpp
niyue: The `literal` is misleading and I remove it.

tests/queries/0_stateless/04212_file_table_function_array_argument.sql
niyue: Some mixed array tests and different formats tests are added, please check them out.

alexey-milovidov (approved):

Issue #98981, opened 2026-03-08
Add function `h3PolygonToCellsExperimental`


CurtizJ:
### Company or project name

ClickHouse Inc.

### Use case

ClickHouse provides function `h3PolygonToCells`, which returns h3 indexes of cells that are contained in the polygon according to the [documentation](https://h3geo.org/docs/api/regions/#polygontocells). But I want to get all cells that overlaps the polygon. Fortunately H3 provides such function: https://h3geo.org/docs/api/regions/#polygontocellsexperimental.

### Describe the solution you'd like

Function `h3PolygonToCellsExperimental(coordinates, resolution, mode)` similar to `h3PolygonToCells`, where mode is a constant string with one of the values: `center`, `full`, `overlapping`, `overlapping_bbox` according to the [documentation](https://h3geo.org/docs/api/regions/#polygontocellsexperimental).

Emre-KMZ:
Hi, I would like to work on this issue. Can you assign it to me?

yousefQadry:
it it is still not implemented yet I can open a PR my code is ready with the documentation and test.

PR #104455, opened 2026-05-09
Implement `h3PolygonToCellsWithContainment` function

yousefQadry:
Implements #98981 

- New Feature 

You can now use h3PolygonToCellsWithContainment function from h3 to use their new algorithm which supports center-based, fully-contained, and overlapping containment modes.
...

### Version info
- Merged into: `26.6.1.267`


alexey-milovidov:
We will need to come up with a better name than "experimental". While it reflects the current h3 API one-to-one, it will be a problem for us, because the function name will be cemented forever.

yousefQadry:
> We will need to come up with a better name than "experimental". While it reflects the current h3 API one-to-one, it will be a problem for us, because the function name will be cemented forever.

I changed its name to h3PolygonToCellsWithContainment.

alexey-milovidov:

yousefQadry:
There were some bugs that I fixed after your comment. The PR is ready now.

alexey-milovidov (approved):

PR #104465, opened 2026-05-09
Fix bogus authentication error for unquoted user name in remote() table function

### LINKED ISSUE #33816: Bogus exception if user name is specified without quotes in table function `remote`
```
milovidov-desktop :) SELECT count() FROM remoteSecure('gh-api.clickhouse.tech', default, github_events, explorer)

SELECT count()
FROM remoteSecure('gh-api.clickhouse.tech', default, github_events, explorer)

Query id: 86dd3c7d-4b1c-4099-9fd8-57097a5bebed


0 rows in set. Elapsed: 0.126 sec. 

Received exception from server (version 22.2.1):
Code: 516. DB::Exception: Received from localhost:9000. DB::Exception: Received from gh-api.clickhouse.tech:9440. DB::Exception: default: Authentication failed: password is incorrect or there is no user with such name. (AUTHENTICATION_FAILED)

milovidov-desktop :) SELECT count() FROM remoteSecure('gh-api.clickhouse.tech', default, github_events, 'explorer')

SELECT count()
FROM remoteSecure('gh-api.clickhouse.tech', default, github_events, 'explorer')

Query id: 7a8006e5-eae1-4b26-9a4b-f021b5933abd

┌────count()─┐
│ 4254443174 │
└────────────┘

1 rows in set. Elapsed: 0.115 sec.
```

alexey-milovidov:
Closes #33816.

Previously a bare identifier in the user-name position of `remote` and `remoteSecure` was silently treated as a sharding key, so a query like

    SELECT count() FROM remoteSecure('host', default, github_events, explorer)

attempted to authenticate as `default` (with no password) and failed with the misleading

    Code: 516. DB::Exception: default: Authentication failed: password is incorrect, or there is no user with such name.

The user name now also accepts unquoted identifiers, consistent with how the database and table arguments work. The password still must be a string literal.

Allow an unquoted identifier as the user name in the `remote` and `remoteSecure` table functions, mirroring how the database and table arguments accept unquoted identifiers. Previously such a query failed with a misleading authentication error referring to the `default` user.

### Version info
- Merged into: `26.5.1.540`


alexey-milovidov:
Let's not break the compatibility.
Instead, accept username as an identifier only if there is a subsequent argument with a password.

alexey-milovidov (commented):
Ok.

PR #104418, opened 2026-05-08
Fix flaky FileCacheTest.SLRUFreeSpaceKeepingProtectedOnly

groeneai:
The regression test introduced in #104313 (covering issue #104307) is flaky on
`Unit tests (amd_llvm_coverage)`. It exercises the bug condition through
`FileCache::freeSpaceRatioKeepingThreadFunc` with `keep_free_space_elements_ratio = 1.0`,
which causes `cache->initialize()` to schedule the asynchronous background eviction
task. That task races with the test's `read_and_check` and intermediate
`assertProbationary` / `assertProtected` calls, especially on slow builds (notably
coverage), evicting entries between the populate and assert steps and producing
spurious failures.

### Failures observed

- master `74e1c5ed2c4` (PR #104313 merge) at 16:03:32 UTC -- mismatch at
  `gtest_filecache.cpp:1626` (`getFileSegmentsNum() == 3, expected 0`).
- master `407d6ccaeb4` at 16:24:46 UTC -- mismatch at line 214 (called from line 1616)
  in `assertProtected`: actual `0`, expected `3`.
- PR #89334 run `b5c32eef` at 18:01:22 UTC -- two mismatches at line 214 with the same
  pattern.

All failures are exclusive to the `amd_llvm_coverage` shard. `tsan`, `asan_ubsan`, and
`msan` runs on the same commits all pass, confirming this is a timing race on slow
builds rather than a real regression of the underlying fix.

CI reports:
- https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=74e1c5ed2c4437352b486d59b16ee8bd98a8b688&name_0=MasterCI&name_1=Unit%20tests%20%28amd_llvm_coverage%29
- https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=407d6ccaeb432fbd86b15dc6d341651a64f447dc&name_0=MasterCI&name_1=Unit%20tests%20%28amd_llvm_coverage%29

### Approach

Replace the high-level integration test with a focused unit-level test that exercises
`SLRUFileCachePriority::collectEvictionInfo` directly. The new test constructs the
priority queue, adds three protected-only entries via
`addForRestore(... QueueEntryType::SLRU_Protected ...)`, and calls `collectEvictionInfo`
with the bug-triggering parameters (`is_total_space_cleanup=true`, probationary empty,
protected populated). This reproduces the exact precondition that used to fire the
[...]

Issue #88558, opened 2025-10-15
NOT_FOUND_COLUMN_IN_BLOCK (arrayConcat(JSONExtractKeys))


ernado:
### Company or project name

https://github.com/go-faster/oteldb, OpenTelemetry signal storage

### Describe what's wrong

NOT_FOUND_COLUMN_IN_BLOCK while quering the logs table.

### Does it reproduce on the most recent release?

Yes

### How to reproduce

```sql
CREATE TABLE IF NOT EXISTS `logs`
(
    `service_instance_id` LowCardinality(String) COMMENT 'service.instance.id',
    `service_name`        LowCardinality(String) COMMENT 'service.name',
    `service_namespace`   LowCardinality(String) COMMENT 'service.namespace',
    `timestamp`           DateTime64(9)          CODEC(Delta, ZSTD(1)),
    `severity_number`     UInt8,
    `severity_text`       LowCardinality(String),
    `trace_id`            FixedString(16),
    `span_id`             FixedString(8),
    `trace_flags`         UInt8,
    `body`                String,
    `attribute`           String,
    `resource`            LowCardinality(String),
    `scope_name`          LowCardinality(String),
    `scope_version`       LowCardinality(String),
    `scope`               LowCardinality(String),

    INDEX `idx_trace_id`   trace_id TYPE bloom_filter(0.001) GRANULARITY 1,
    INDEX `idx_body`       body TYPE tokenbf_v1(32768, 3, 0) GRANULARITY 1,
    INDEX `idx_ts`         timestamp TYPE minmax GRANULARITY 8192,
    INDEX `attribute_keys` arrayConcat(JSONExtractKeys(attribute), JSONExtractKeys(scope), JSONExtractKeys(resource)) TYPE set(100)
)
    ENGINE = MergeTree
        PARTITION BY toYYYYMMDD(timestamp)
        ORDER BY (`severity_number`, `service_namespace`, `service_name`, `resource`, `timestamp`)
        PRIMARY KEY (`severity_number`, `service_namespace`, `service_name`, `resource`)
        TTL toDateTime(`timestamp`) + toIntervalSecond(259200)
```

```sql
SELECT service_instance_id,
       service_name,
       service_namespace,
       timestamp,
       severity_number,
       severity_text,
       trace_id,
       span_id,
       trace_flags,
       body,
       scope_name,
       scope_version,
       attribute,
       scope,
       resource
FROM logs
[...]

PR #104473, opened 2026-05-09
Add `OptimizeTrivialGroupByLimitPass`

alexey-milovidov:
- Performance Improvement

Added `OptimizeTrivialGroupByLimitPass`. For trivial `SELECT ... FROM t GROUP BY k LIMIT n` queries (no `HAVING`, `ORDER BY`, or window functions), the analyzer now sets `max_rows_to_group_by = n + offset` with `group_by_overflow_mode = 'any'`, so aggregation stops once `n` distinct keys have been produced instead of grouping the entire input. Controlled by the new setting `optimize_trivial_group_by_limit_query` (enabled by default).

---

#### Why

For `SELECT k FROM t GROUP BY k LIMIT 5`, ClickHouse currently aggregates the entire input before applying the LIMIT, even though the user clearly only wants 5 distinct keys. By the semantics of `GROUP BY ... LIMIT`, *any* 5 distinct keys are a valid answer — so we can stop aggregation as soon as 5 distinct keys exist.

#### What changed

- New analyzer pass `OptimizeTrivialGroupByLimitPass`.
- New setting `optimize_trivial_group_by_limit_query` (default `1`) to gate the pass.
- The pass fires only when the query has `GROUP BY` and a `LIMIT`, and has no `HAVING`, `ORDER BY`, or window clauses; it would change observable results for those.
- The pass is suppressed when the user has already set a non-zero `max_rows_to_group_by`, so it never overrides an explicit overflow contract.

#### Origin

Split out from WIP PR #81944.

### Version info
- Merged into: `26.5.1.647`


alexey-milovidov:
Does it work correctly with multiple threads? I'm concerned, because the set of first keys could be different in different threads.

alexey-milovidov:
@groeneai, fix unit tests for FileCache.

groeneai:
@alexey-milovidov — confirmed: the failing FileCache unit test on PR #104473 head `5e9d827c6968` is the chronic flake **`FileCacheTest.SLRUFreeSpaceKeepingProtectedOnly`** on `Unit tests (amd_llvm_coverage)` (`gtest_filecache.cpp:1629`, `getFileSegmentsNum() == 3` vs expected `0`). Not caused by this PR.

**CIDB (14-day cross-PR, `FileCacheTest.SLRUFreeSpaceKeepingProtectedOnly` only):**

| | |
|---|---|
| Total hits | 55 |
| Distinct PRs hit | 43 |
| Master hits | 3 (last 2026-05-10) |
| Concentrated on | `Unit tests (amd_llvm_coverage)` (52/55) |
| PR #104473 | 2 hits — matches the chronic pattern |

Root cause is a known race: `keep_free_space_elements_ratio = 1.0` makes `FileCache::initialize()` schedule the async background eviction task, which races with the test's `assertProtected` / `assertProbationary` calls — surfaces under slow coverage builds.

**Fix is already filed: #104418** (OPEN since 2026-05-08). It replaces the racy integration-style test with a focused unit-level test on `SLRUFileCachePriority::collectEvictionInfo` that bypasses `FileCache::initialize`'s async eviction race entirely. Verified locally: 30/30 deterministic passes; reintroducing the chassert aborts cleanly. Could you or @kssenii take a look so we can land it?

alexey-milovidov:
@groeneai, fix the failure of 04217_array_concat_json_extract_keys_index and send a PR.

groeneai:
@alexey-milovidov filed the test-only fix as #104629.

**Root cause:** CI's `--diagnose-random-settings` identifies `map_serialization_version_for_zero_level_parts = 'with_buckets'` + `map_buckets_strategy = 'constant'` + small `map_buckets_min_avg_size` as the culprit on both this PR and PR #103016. Under bucketed serialization, `Map` entries are stored grouped by hash bucket, so the test's final `SELECT *` over the three `Map(LowCardinality(String), String)` columns renders `resource` as `{'job':'g','d':'e'}` instead of `{'d':'e','job':'g'}`.

**Fix:** Wrap the `Map` columns in `mapSort` in the `SELECT` only — same pattern as the existing `04060_coalescing_merge_tree_array_tuple_with_column_param.sql`. The `WHERE` clause and the `arrayConcat(mapKeys(...))` index path (the actual subject of the test, issue #88558 / PR #94515) are untouched, so all randomized serialization configurations are still exercised. 50/50 passes under the test runner with full randomization.

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
```diff
@@ -0,0 +1,73 @@
...
+        return;
+
+    auto * query = query_tree_node->as<QueryNode>();
+    if (query && query->hasGroupBy() && query->hasLimit() && !query->hasHaving() && !query->hasOrderBy() && !query->hasWindow()
+        && !query->isGroupByWithTotals() && !query->isGroupByWithRollup() && !query->isGroupByWithCube()
+        && !query->isGroupByWithGroupingSets()
+        && !hasAggregateFunctionNodes(query->getProjectionNode()))
+    {
+        auto & mutable_context = query->getMutableContext();
+        if (settings[Setting::max_rows_to_group_by] == 0)
```
Algunenano: I think we can also apply this if the setting is larger than the limit, even if it’s not zero

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
```diff
@@ -0,0 +1,73 @@
...
+                auto maybe_offset = tryGetNonNegativeUInt64(query->getOffset()->as<ConstantNode &>().getValue());
+                if (!maybe_offset)
+                    return;
+                offset = *maybe_offset;
+            }
+            UInt64 max_rows = 0;
+            if (common::addOverflow(*limit, offset, max_rows))
+                return;
+            mutable_context->setSetting("max_rows_to_group_by", max_rows);
+            mutable_context->setSetting("group_by_overflow_mode", Field("any"));
```
Algunenano: I think that we might want to avoid overriding this setting if it’s not default too, as it might be unexpected for those using other modes (like throw)

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+void OptimizeTrivialGroupByLimitPass::run(QueryTreeNodePtr & query_tree_node, ContextPtr context)
+{
+    const Settings & settings = context->getSettingsRef();
+    if (!settings[Setting::optimize_trivial_group_by_limit_query])
+        return;
+
+    auto * query = query_tree_node->as<QueryNode>();
+    if (query && query->hasGroupBy() && query->hasLimit() && !query->hasHaving() && !query->hasOrderBy() && !query->hasWindow()
```
Algunenano: The AI review points, and I agree, that we can’t apply this optimization if LIMIT BY is being used

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
```diff
@@ -0,0 +1,73 @@
...
+        && !query->isGroupByWithTotals() && !query->isGroupByWithRollup() && !query->isGroupByWithCube()
+        && !query->isGroupByWithGroupingSets()
+        && !hasAggregateFunctionNodes(query->getProjectionNode()))
+    {
+        auto & mutable_context = query->getMutableContext();
```
Algunenano: Can you please confirm that this only modifies the setting value for this subquery and not for the whole subquery? I believe that’s how it works, but adding a test would help confirm

tests/queries/0_stateless/04213_optimize_trivial_group_by_limit_query.sql
```diff
@@ -0,0 +1,72 @@
...
+
+DROP TABLE IF EXISTS t_trivial_group_by_limit;
+
+CREATE TABLE t_trivial_group_by_limit (k UInt64) ENGINE = MergeTree ORDER BY tuple();
+INSERT INTO t_trivial_group_by_limit SELECT number FROM numbers(1000000);
+
+-- With the optimization on or off, the projection always returns exactly LIMIT rows,
+-- so this query reports the same value either way; the difference is internal — the
+-- optimization stops the aggregation once 5 distinct keys are produced.
+SELECT count() FROM (SELECT k FROM t_trivial_group_by_limit GROUP BY k LIMIT 5) SETTINGS optimize_trivial_group_by_limit_query = 1;
```
Algunenano, line 19: A good test to add would be based on numbers of rows read in each mode. Without the optimization everything will be read, with it we'll read only a subset. We’ll need to be strict on other settings so it's not flaky, but it’s worth it to verify that the optimization actually does anything

Algunenano (changes requested):

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
```diff
@@ -42,32 +44,48 @@ void OptimizeTrivialGroupByLimitPass::run(QueryTreeNodePtr & query_tree_node, Co
...
+        || query->isGroupByWithGroupingSets() || hasAggregateFunctionNodes(query->getProjectionNode()))
+        return;
+
+    /// `group_by_overflow_mode` controls what happens when `max_rows_to_group_by` is exceeded.
+    /// The optimization only makes sense in `ANY` mode (keep first N keys, drop the rest);
+    /// `THROW` would turn the optimization into a spurious exception, `BREAK` aborts the query.
+    /// Don't change the mode if the user has set it explicitly to something non-`ANY` —
+    /// they have an explicit contract that the optimization would silently break.
+    const bool mode_is_any = settings[Setting::group_by_overflow_mode] == OverflowMode::ANY;
+    const bool mode_is_changed = settings.isChanged("group_by_overflow_mode");
```
Algunenano: It'd better to use something like `settings[Setting::group_by_overflow_mode].changed`

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
alexey-milovidov: Already covered: line 82 of the pass — `if (user_max_rows != 0 && user_max_rows <= max_rows) return;` — so we skip only when the user value is non-zero **and** tighter than ours. When the user has set a larger value, the optimization still fires.

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
alexey-milovidov: Already covered: lines 57–60 — we read `group_by_overflow_mode` and `settings.isChanged("group_by_overflow_mode")` and skip the pass when the user has explicitly set the mode to anything non-`ANY` (so `throw`/`break` are preserved). Tests `04213_optimize_trivial_group_by_limit_query` cover the `throw` and `break` cases.

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
alexey-milovidov: Already covered: `query->hasLimitBy()` is in the early-skip list at line 48 of the pass; `04213_optimize_trivial_group_by_limit_query.sql` has a `GROUP BY ... LIMIT 10 BY (k % 2) LIMIT 5` case that checks the optimization does **not** fire (it would otherwise starve `LIMIT BY` of groups).

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
alexey-milovidov: Confirmed and tested. The pass calls `query->getMutableContext()`, which returns the per-`QueryNode` context (each subquery has its own), so the mutation is scoped to that subquery. Added a test (last query in `04213_optimize_trivial_group_by_limit_query.sql`) that runs an optimized subquery inside a `SELECT` and then reads `getSetting("max_rows_to_group_by")` / `getSetting("group_by_overflow_mode")` from the outer query — they remain at the defaults (`0` / `throw`), confirming no outer-context leak.

tests/queries/0_stateless/04213_optimize_trivial_group_by_limit_query.sql
alexey-milovidov: The aggregator processes all input rows regardless of the optimization — when `max_rows_to_group_by` is exceeded with mode `any`, `no_more_keys` is set but the input stream is not torn down, so `read_rows` should look identical between the two modes. Test `04229_trivial_group_by_limit_profile_events` instead asserts that the pass actually fires by checking the `OverflowAny` `ProfileEvent` (the aggregator only increments it when the cap is reached) — which directly answers “did the optimization do something”. Happy to add a `read_rows` check too if you want it as a negative control, but I think `OverflowAny` is the more direct signal.

src/Core/SettingsChangesHistory.cpp
```diff
@@ -41,6 +41,7 @@ const VersionToSettingsChangesMap & getSettingsChangesHistory()
         /// Note: please check if the key already exists to prevent duplicate entries.
         addSettingsChanges(settings_changes_history, "26.5",
         {
+            {"optimize_trivial_group_by_limit_query", false, true, "New setting that limits aggregation to at most LIMIT distinct keys for `SELECT agg() FROM t GROUP BY k LIMIT n` queries."},
```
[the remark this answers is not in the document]
alexey-milovidov: Updated in commit a05531bedaf — the reason text now reads `SELECT key_expr FROM t GROUP BY key_expr LIMIT n`, matching the actual shape the pass supports.

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
alexey-milovidov: Done in commit `c7ed00fb77f` — now using `settings[Setting::group_by_overflow_mode].changed` (typed accessor, no string lookup).

tests/queries/0_stateless/04229_trivial_group_by_limit_profile_events.sql
```diff
@@ -0,0 +1,33 @@
...
+SETTINGS optimize_trivial_group_by_limit_query = 1, max_threads = 1, log_comment = '04229_on';
+
+SELECT k FROM t_04229 GROUP BY k LIMIT 5 FORMAT Null
+SETTINGS optimize_trivial_group_by_limit_query = 0, max_threads = 1, log_comment = '04229_off';
+
+SYSTEM FLUSH LOGS query_log;
+
+SELECT
+    log_comment,
+    ProfileEvents['OverflowAny'] > 0 AS overflow_any_fired
```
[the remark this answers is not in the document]
alexey-milovidov: Added a focused performance test `tests/performance/optimize_trivial_group_by_limit.xml` that compares `optimize_trivial_group_by_limit_query = 1` vs `0` on `SELECT number FROM numbers_mt(100000000) GROUP BY number LIMIT 5` — the optimized path stops aggregation once `LIMIT` distinct keys are reached, giving a measurable elapsed-time win on the same query shape.

tests/performance/optimize_trivial_group_by_limit.xml
```diff
@@ -0,0 +1,9 @@
+<test>
+    <!-- For `SELECT k FROM t GROUP BY k LIMIT n` aggregation should stop once
+         `n` distinct keys are produced rather than aggregating the entire input.
+         Compare the optimized path vs the disabled setting on a large input
+         with many distinct keys but a small `LIMIT`. -->
+
+    <query tag="optimized">SELECT number FROM numbers_mt(100000000) GROUP BY number LIMIT 5 FORMAT Null SETTINGS optimize_trivial_group_by_limit_query = 1</query>
```
Algunenano, line 7: I didn't know we could use `tag`. Nice

Algunenano (approved):

alexey-milovidov:
The following CI failures on this PR are unrelated to the changes in this PR (which only touches `OptimizeTrivialGroupByLimitPass` and related tests):

- `Unit tests (asan_ubsan, function_prop_fuzzer) / FunctionsStress.stress` — `broken_determinism` in `countEqual` with `FixedString`. Also failing on recent master runs.
- `Unit tests (msan, function_prop_fuzzer) / FunctionsStress.stress` — `broken_determinism` in `caseWithExpression` with `Enum16`. Also failing on recent master runs.
- `Stress test (arm_ubsan) / Cannot start clickhouse-server` — `UndefinedBehaviorSanitizer: undefined-behavior /ClickHouse/src/Storages/StorageTimeSeries.cpp:171:36`, looks like a null-deref in `StorageTimeSeries::StorageTimeSeries` from `DatabaseCatalog::instance().tryGetTable(...)`. No open issue or PR found tracking this exact failure.

@groeneai, please investigate these failures and provide fixes in separate PRs. If fixes are already in progress, link them here.

Algunenano:
caseWithExpression is fixed by https://github.com/ClickHouse/ClickHouse/pull/104858, but I don't think that one covers `countEqual`

src/Analyzer/Passes/OptimizeTrivialGroupByLimitPass.cpp
```diff
@@ -0,0 +1,97 @@
...
+    /// The query also returns no rows regardless, so the optimization buys nothing.
+    if (max_rows == 0)
+        return;
+
+    /// If the user has already set `max_rows_to_group_by`, we only apply the optimization
+    /// when our derived value is strictly smaller — otherwise the user's setting is tighter
+    /// and ours would be a no-op. When the user has a tighter throw/break contract, we'd
+    /// also break their semantics (already guarded by the `mode_is_changed` check above).
+    const UInt64 user_max_rows = settings[Setting::max_rows_to_group_by];
+    if (user_max_rows != 0 && user_max_rows <= max_rows)
```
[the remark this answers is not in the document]
alexey-milovidov: This is entirely ok.

Why: The meaning of the number of rows in GROUP BY is also affected by the optimization. The setting `max_rows_to_group_by` should control the actual number of rows - not the number of rows that would have been without the optimization. Changing the behavior (for better) in this case is correct.

PR #82341, opened 2025-06-21
Add _part_granule_offset column

amosbird:
### Changelog category (leave one):
- New Feature

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Support `_part_granule_offset` virtual column in MergeTree-family tables. This column indicates the zero-based index of the granule/mark each row belongs to within its data part. This addresses #79572.

This PR also relaxes the restriction that `_part_offset` and `_block_offset` must be read during the first reading step. It further includes some preparatory refactoring extracted from #81526 to simplify the review of that PR.

I'm currently using `_part_granule_offset` to analyze whether the query condition cache fully records all relevant granules.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

PR #79417, opened 2025-04-22
Add `_part_starting_offset` virtual column and key condition support for offset-based querying

amosbird:
### Changelog category (leave one):
- New Feature

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Support `_part_starting_offset` virtual column in MergeTree-family tables. This column represents the cumulative row count of all preceding parts, calculated at query time based on the current part list. The cumulative values are retained throughout query execution and remain effective even after part pruning. Related internal logic has been refactored to support this behavior.

When expressions like `_part_starting_offset + _part_offset` or `_part_offset + _part_starting_offset` are used in the WHERE clause, key condition analysis will be properly applied, enabling efficient query-then-fetch patterns and cursor-based pagination. This improves the analysis process introduced in #58224 . Now only one numeric column is used to filter instead of (_part, _part_offset) pair.

This PR also improves stability over projection indexes that rely on `_part` and `_part_offset`, which are sensitive to part merges. See  #78429 . However, it may still not be reliable under workloads involving: Inserts into partitioned table, materialization of lightweight delete masks, background merges in Collapsing, Replacing, or AggregatingMergeTree tables, etc. A proper query-level snapshot is required for this, and it will be implemented in another PR.

```
SELECT
    sum(column_not_in_projection)
FROM events
WHERE (_part_starting_offset + _part_offset) IN (
    SELECT _part_starting_offset + _part_offset
    FROM events
    WHERE user_id = 42
)
```

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

PR #104474, opened 2026-05-09
Add `RewriteOrderByLimitPass`

alexey-milovidov:
- Performance Improvement

Added `RewriteOrderByLimitPass`. For wide tables, `SELECT * FROM t ORDER BY narrow_key LIMIT n` is rewritten so the heavy columns are read only for the surviving N rows: a first scan reads the sort key and `_part_starting_offset`/`_part_offset`, ORDER BY + LIMIT narrows it to N (offset, key) tuples, then a self-join re-reads the rest of the columns by row position. Controlled by `query_plan_rewrite_order_by_limit` (enabled by default).

---

#### Why

For a `SELECT * FROM t ORDER BY ts LIMIT 10` over a wide table (say 200 columns), the executor reads all 200 columns for every row in every part during the sort phase, then discards 99.99% of them at the LIMIT. Reading only the 1–2 columns needed for sorting plus the row coordinates first, and re-reading the heavy columns only for the surviving rows, is dramatically cheaper when the table is wide enough.

#### How it works

1. First scan reads only the sort key plus the virtual columns `_part_starting_offset` and `_part_offset` (already in master via #79417 and #82341).
2. ORDER BY ... LIMIT N narrows that down to N (offset, key) tuples.
3. A self-join re-reads the rest of the columns from MergeTree, but only for the N surviving row positions.

#### Settings

- `query_plan_rewrite_order_by_limit` (Bool, default `1`) — main gate.
- `query_plan_max_limit_for_rewrite_order_by_limit` (UInt64, default `1000000`) — only fires for `LIMIT n` at or below this value (rewriting for very large LIMITs gains nothing).
- `query_plan_min_columns_to_use_rewrite_order_by_limit` (UInt64, default `50`) — only fires for tables with at least this many columns.

#### Origin

Split out from WIP PR #81944. Adapted to current master APIs (`metadata->virtuals.tryGet(name, VirtualsKind::All, VirtualsMaterializationPlace::All)`, `getInMemoryMetadataPtr(context, /*bypass_metadata_cache=*/false)`).


EmeraldShift:
I remember this as #82478, which was closed since it seemed that Lazy Materialization V2 was doing ~the same thing. Is this the same optimization as that PR? I'm curious if there are different situations where this applies vs. lazy materialization, or if the performance is different?

alexey-milovidov:
This is the same optimization as #82478 (the PR description notes it was split out from WIP #81944).

Lazy materialization and `RewriteOrderByLimitPass` overlap in intent — both avoid reading the wide columns for the rows that get discarded by `LIMIT` — but they differ in mechanics:

- Lazy materialization keeps a single physical scan and defers reading the heavy columns until after `ORDER BY/LIMIT` inside the same plan.
- `RewriteOrderByLimitPass` rewrites the query at the analyzer level into a subquery that produces row positions plus a self-join that re-reads the heavy columns by `_part_starting_offset + _part_offset`. The wide columns are not touched at all in the sort phase.

Where they differ in practice:

1. **Large `LIMIT`.** The numbers in #82478 show the rewrite ~90× faster than lazy materialization for `SELECT * FROM hits ORDER BY EventTime LIMIT 100000`, where lazy materialization still has to keep ~`LIMIT × all-columns` worth of state in flight.
2. **Small `LIMIT`.** The perf-comparison runs on this PR show `lazyMaterialization #3` (`SELECT * FROM hits ORDER BY EventTime LIMIT 10`) is faster on master than with the rewrite, because the IN-subquery + self-join overhead is non-trivial when only 10 rows survive.
3. **Distributed/non-`MergeTree` storage.** The rewrite only fires for single-table `StorageMergeTree` queries (no distributed, no joins). Lazy materialization is more general.

So they coexist rather than replace each other. The current gates (`query_plan_min_columns_to_use_rewrite_order_by_limit = 50`, `query_plan_max_limit_for_rewrite_order_by_limit = 1000000`) try to keep the rewrite to the regime where it clearly wins.

EmeraldShift:
Thanks for the explanation. I thought lazy materialization was actually faster than the rewrite even for LIMIT 1000000 per [this comment](https://github.com/ClickHouse/ClickHouse/pull/82478#issuecomment-3670419617), which is why I'm confused at the performance benefit of this PR.

About Distributred tables, I have seen cases where lazy materialization doesn't fire, but I've found it to work again with `prefer_localhost_replica = 0`, which maybe just indicates an extra query tree shape the LM optimization needs to recognize. Do I understand right that both LM and this PR would perform the self-join or heavy column scan locally on each shard?

alexey-milovidov:
Ok, let's assume it is pointless.

Issue #84198, opened 2025-07-22
JoiningTransform::transform:  Invalid number of rows in Chunk


Algunenano:
### Describe the bug

https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=ff4fc4139a3b074f2fe7708f59a7bd5344509b0d&name_0=MasterCI&name_1=AST%20fuzzer%20%28amd_debug%29

### How to reproduce

```
DROP TABLE IF EXISTS test_table_join_1 SYNC;
DROP TABLE IF EXISTS test_table_join_2 SYNC;
DROP TABLE IF EXISTS test_table_join_3 SYNC;

CREATE TABLE test_table_join_1 (`id` UInt8, `value` String) ENGINE = ReplicatedMergeTree('/clickhouse/{database}/test_table_join_1111', 'r1') ORDER BY id SETTINGS index_granularity = 1;
INSERT INTO test_table_join_1 SELECT number, concat('t1_value_', number) FROM numbers(500);

CREATE TABLE test_table_join_2 (`id` UInt16, `value` String) ENGINE = ReplicatedMergeTree('/clickhouse/{database}/test_table_join_21111', 'r1') ORDER BY id SETTINGS index_granularity = 1;
INSERT INTO test_table_join_2 SELECT number, concat('t2_value_', number) FROM numbers(100, 500);

CREATE TABLE test_table_join_3 (`id` UInt64, `value` String) ENGINE = ReplicatedMergeTree('/clickhouse/{database}/test_table_join_31111', 'r1') ORDER BY id SETTINGS index_granularity = 1;
INSERT INTO test_table_join_3 SELECT number, concat('t3_value_', number) FROM numbers(500);

[...]

PR #84213, opened 2025-07-22
Fix indexing in ScatteredBlock::filter

vdimir:
### Changelog category (leave one):

- Not for changelog (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
* Fix possible LOGICAL_ERROR in queries with `parallel_hash` join and thin key columns (like Int16)

(bug in unreleased version)
Close https://github.com/ClickHouse/ClickHouse/issues/84157
Close https://github.com/ClickHouse/ClickHouse/issues/84198

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

Issue #80703, opened 2025-05-22
Query error NOT_FOUND_COLUMN_IN_BLOCK


pjain1:
### Company or project name

Rill Data

### Describe what's wrong

Hello I am getting really weird query error NOT_FOUND_COLUMN_IN_BLOCK
```
SELECT
    arrayJoin(cust_flags) AS custs, 
    sum(__sourcerows) AS total_events
FROM sync_local
WHERE ((__timestamp >= '2025-05-15 02:00:00') AND (__timestamp < '2025-05-22 02:00:00')) AND (arrayJoin(cust_flags) IN ('<>')) AND (dictGetOrDefault('tags', 'site_name', p_id, '') IN ('<>'))
GROUP BY 1
ORDER BY 1 DESC
LIMIT 5000

Received exception from server (version 24.12.6):
Code: 10. DB::Exception: Received from :9440. DB::Exception: Not found column cust_flags in block: while executing 'INPUT : 1 -> cust_flags Array(String) : 5'. (NOT_FOUND_COLUMN_IN_BLOCK)
```

If I remove dictGetOrDefault condition from where clause then it works fine. Here is the create table -

```
CREATE TABLE default.sync_local
(
    `__partition` LowCardinality(String),
    `__sourcerows` UInt64,
    `__timestamp` DateTime,
    `p_id` LowCardinality(String),
    `cust_flags` Array(String),
    INDEX idx_set_p_id p_id TYPE set(100) GRANULARITY 1,
    INDEX idx_minmax_cust_flags cust_flags TYPE minmax GRANULARITY 2,
    PROJECTION sum_events__projection
    (
        SELECT __timestamp, p_id, sum(__sourcerows)
        GROUP BY 1, 2
    )
)
ENGINE = ReplicatedMergeTree('/clickhouse/tables/{uuid}/{shard}', '{replica}')
PARTITION BY __partition
ORDER BY (__timestamp, p_id)
SETTINGS index_granularity = 8192
```
CH version - 24.12.6 version. 

### Does it reproduce on the most recent release?

Yes

### How to reproduce

Have an array and dictionary and put them in where clause of a group by query

Algunenano:
Could you please provide a full reproducer, so it's easier to debug and bisect? Using fiddle.clickhouse.com would be best

pjain1:
[...]

PR #104475, opened 2026-05-09
Precompute hash and prefetch for prealloc aggregation variants

alexey-milovidov:
- Performance Improvement

For the `prealloc_serialized` family of aggregation methods, precompute per-row hashes during the batch-serialization pass and use them to (a) skip rehashing in `emplaceKey`/`findKey` and (b) software-prefetch the next row's bucket. Speeds up multi-key string/serialized aggregation by hiding hash-table cache-miss latency.

---

#### Why

When `HashMethodSerialized<prealloc=true>` is chosen for aggregation, the row sizes are already collected up front so keys can be serialized in a single batch. The hash for each row is computed twice in this codepath — once implicitly while the hash table looks up the cell, then again every time we touch that row. We can compute it once during the batch pass and reuse it.

That precomputed hash also lets us prefetch the future row's bucket: by the time the loop reaches it, the cache line is warm.

#### What changed

- Hash methods opt in via `static constexpr bool has_pre_computed_hashes = true` and provide `hashes` (a `WeakHash32`), `prefetching` (`PrefetchingHelper`), and `prefetch_look_ahead`.
- `HashMethodSerialized<prealloc=true>` opts in. `HashMethodSingleLowCardinalityColumn` forwards the flag from its base. All other hash methods default to `false` via `HashMethodBase`.
- `ColumnsHashingImpl::emplaceKey`/`findKey`: when `has_pre_computed_hashes`, prefetch the bucket `look_ahead` rows ahead and skip the rehash on the current row. For `findKey`, also test `data.isEmptyCell(hash_value)` to early-exit a miss.
- A new `emplaceImpl<bool compute_hash>` overload chooses between `data.emplace(...)` and `data.emplace(..., hash_value)`.
- `HashTable` exposes `prefetchByHash(hash)` and `isEmptyCell(hash)` publicly; `TwoLevelHashTable` adds matching overloads that route to the right bucket.

#### Origin

Split out from WIP PR #81944. The original commit also tweaked the `PrefetchingHelper::calcPrefetchLookAhead` algorithm; that piece is left out here because master already evolved a different (integer-arithmetic) version of the same calibration and the precomputed-hash path works fine on top of it.

### Version info
- Merged into: `26.5.1.649`


alexey-milovidov:
I think this is great:

Screenshot_20260511_195345

alexey-milovidov:
@groeneai, the CI failures on this PR look unrelated to the changes (which only touch hash table / aggregation code in `src/Common/ColumnsHashing*`, `src/Common/HashTable/*`, and `src/Interpreters/Aggregator.cpp`):

1. **Stateless test `04221_dictgetordefault_arrayjoin_80703`** — `BAD_ARGUMENTS` in `ExternalDictionariesLoader::resolveDictionaryName`. Stack trace is in dictionary resolution and analyzer code. [Report](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104475&sha=ecbe7199176048eee6f2d7de471e25f7ba63983a&name_0=PR&name_1=Stateless%20tests%20%28amd_llvm_coverage%2C%20ParallelReplicas%2C%20s3%20storage%2C%20parallel%29).
2. **Unit test `FunctionsStress` (tsan, function_prop_fuzzer)** — `Logical error: 'Function writeSlice expects same column types for GenericValueSlice and GenericArraySink.'`. This is in array function code. [Report](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104475&sha=ecbe7199176048eee6f2d7de471e25f7ba63983a&name_0=PR&name_1=Unit%20tests%20%28tsan%2C%20function_prop_fuzzer%29).
3. **AST fuzzer (amd_debug)** — `Inconsistent AST formatting between 'ExpressionList' and 'ExpressionList'` involving extra parentheses around `tuple(...)` / `not(...)`. Looks like the same class of issue as #104605. [Report](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104475&sha=ecbe7199176048eee6f2d7de471e25f7ba63983a&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%29).
4. **AST fuzzer (arm_asan_ubsan)** — `Invalid number of rows in Chunk` in `JoiningTransform::transform`. [Report](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104475&sha=ecbe7199176048eee6f2d7de471e25f7ba63983a&name_0=PR&name_1=AST%20fuzzer%20%28arm_asan_ubsan%29).

Please investigate these and provide fixes in separate PRs. If any are already in progress, please link them here. I just merged master into the branch, so CI will rerun shortly — if the failures are flaky they may not reproduce.

groeneai:
@alexey-milovidov — investigated all four. All four are pre-existing chronic/trunk bugs unrelated to this PR (and two already have fix PRs in flight).

### F1 — `04221_dictgetordefault_arrayjoin_80703` (BAD_ARGUMENTS)

Chronic-flaky on **`Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel)` only**. First seen 2026-05-12 (same day the test was added in 23d8c1f1 "Add test for #80703"). CIDB 30-day count: **21 failures across 18 PRs + 4 master hits**, ALL on that one check. The error is `Dictionary (tags_80703) not found` during parallel-replica execution — the dictionary is created and immediately queried, and remote replicas don't see it in the per-query database scope.

**Already in flight:** PR #104778 (you, 2026-05-13 02:55 UTC) — `Tag 04221_dictgetordefault_arrayjoin_80703 as no-parallel-replicas`. MERGEABLE, +3 lines. Will resolve it once merged. No separate fix needed from us.

### F2 — `FunctionsStress` (tsan, function_prop_fuzzer) — `writeSlice expects same column types for GenericValueSlice and GenericArraySink`

Rare chronic. 30-day CIDB: **3 hits total** (this PR, PR #96886 on 2026-05-11, and one master hit 2026-04-17 in `BuzzHouse (experimental, serverfuzz, amd_msan)` with the `GenericArraySlice/Sink` variant). 90-day: ~11 hits, all on `function_prop_fuzzer` or `BuzzHouse` jobs.

This is a pre-existing limitation of `FunctionIf::executeGenericArray` (`src/Functions/if.cpp:637` → `GatherUtils::conditional` → `Algorithms.h:470` → `writeSlice` at `Algorithms.h:83`). The `function_prop_fuzzer` constructs `if(cond, value_of_TypeA, array_of_TypeB)` argument shapes that bypass the normal type-unification done by the analyzer, hitting the `Same column types` assertion. We previously fixed one variant in #103084 (QBit `structureEquals`); the residual `GenericValueSlice` case is a different code path (scalar-vs-array, not column-equality). PR #104352 ("Improve function_prop_fuzzer", alexbakharew) was merged recently and likely contributed to its slightly higher visibility, but the underlying engine assertion has been there since at least Feb 2026.

Not actionable on this PR's timeline; needs proper handling in `executeGenericArray` to wrap mismatched-type "then"/"else" branches before calling `conditional`. I'll keep tracking via CIDB.

### F3 — AST fuzzer (amd_debug) `Inconsistent AST formatting between 'ExpressionList' and 'ExpressionList'` (tuple/not parens)

Confirmed STID 1941-1bfa, "NOT-tuple subvariant" — same class as #104605 as you noted. Already tracked on our side; root cause and minimal reproducer found yesterday on your directive on PR #104591:

**Minimal reproducer** (debug build, exit 134):
```sql
CREATE TABLE t (c0 Int CODEC(not((not(materialize(1), materialize(2)))), ZSTD)) ENGINE=Memory
```
The CODEC arg parser produces `Function_not(is_operator=true, 2 args)` directly, while the SELECT/RoundBrackets path wraps `not(a, b)` into `Function_not(1 arg) → Function_tuple(is_operator=true, 2 args)`. In CODEC context the formatter has `frame.allow_operators=false`, which blocks `Function_tuple`'s operator-form `(a, b)` (`ASTFunction.cpp:343`), so it falls back to `tuple(a, b)`. On re-parse, that re-parsed `tuple(...)` gets `parenthesized=true` from the surrounding `(...)`, and `IAST::decideParensEmission` adds one extra `(...)` on the next round-trip → mismatch at `executeQuery.cpp:1330`.

PR #104626 (already open from us) only covers the *lambda-after-comma* subvariant of #104605. The NOT/tuple subvariant needs a separate small fix targeting either `decideParensEmission` to skip the extra parens when the parent forces function-call form, or harmonizing the CODEC vs SELECT parser to produce the same shape for `not(a, b)`. I'll prepare that PR (option (a) is the most localized).

### F4 — AST fuzzer (arm_asan_ubsan) `Invalid number of rows in Chunk` in `JoiningTransform::transform`

Confirmed STID 3366-2d1d family — same logical error as the 2026-04-23 master hit in `BuzzHouse (amd_msan)` already tracked on our side, and same family as previously-closed #84198 / #51312 (`ScatteredBlock::filter` indexing, fixed by @vdimir in #84213 for the `parallel_hash` thin-key variant). The fuzzer query on this PR is different from the April hit but produces the same column-size mismatch on the output Chunk:

```
expected 65409, got 130818  (column at position 4, exactly 2× — cross-product duplication not propagated to the other output columns)
```

60-day CIDB: 2 master hits, ~4 PR hits — rare chronic. We previously attempted local reproduction (~150 iterations across `max_threads`, `max_block_size`, `parallel_hash`, `enable_lazy_columns_replication`, `join_runtime_filter_exact_values_limit`) and couldn't trigger it without the full BuzzHouse oracle wrapper. Our note from that investigation: the likely site is `ScatteredBlock::filter` / `appendRightColumns` in the `offsets.empty()` branch when a bloom-filter false positive interacts with multi-slot `ConcurrentHashJoin`. We're keeping this under CIDB watch and will escalate to @vdimir / @KochetovNicolai / @Algunenano if we get a deterministic repro or hit-rate climbs.

---

**Summary:** None caused by this PR. F1 fixed by your #104778 (just needs merge). F3 follow-up fix PR coming from us this week. F2 and F4 stay on the chronic-trunk monitoring list. I'll comment on this PR again only if any of the four reoccurs after #104778 lands.

src/Common/ColumnsHashingImpl.h
```diff
@@ -242,6 +284,46 @@ class HashMethodBase
...
+                    {
+                        derived.can_precompute_hashes = false;
+                    }
+                    else
+                    {
+                        derived.initPrecomputedHashes(data);
+                    }
+                }
+
+                if (derived.can_precompute_hashes)
```
Algunenano: This means we are checking this flag twice, in the hot path. Whether the compiler can optimize it or not is a good question since there are other operations on derived (`initPrecomputedHashes`). The safest bet would be to extract the check before the init block and use it:
```
const bool active = derived.can_precompute_hashes;
if (active)
{
...
}
```

Another option would be to first check for `precomputed_hashes_initialized` and always set it to true, even in the cases where we don't precompute the hashes.

WDYT?

alexey-milovidov: Good point — addressed in 5a65a921b45. Switched to a single-check pattern: `precomputed_hashes_initialized` starts `true` by default and is flipped to `false` in the constructor only when we plan to precompute. After the first call it is set back to `true` regardless of whether hashes were actually computed, so subsequent rows only do the one `can_precompute_hashes` check in the hot path. Also folded the `min_bytes_for_prefetch` threshold check into `initPrecomputedHashes` itself.

Algunenano (approved):

PR #104154, opened 2026-05-05
Fix UBSan in FieldVisitorConvertToNumber for wide-integer Float64 boundary

groeneai:
Fix undefined behavior in `FieldVisitorConvertToNumber<T>::operator()(Float64)` when `T` is a wide integer type (`UInt64`, `Int64`, `Int128`, `UInt128`, `Int256`, `UInt256`).

### Bug

For these wide integer types `numeric_limits<T>::max()` has more than 53 significant bits, so converting it to `Float64` rounds **up**. Example for `UInt64`:

```
numeric_limits<UInt64>::max() = 18446744073709551615 = 2^64 - 1
Float64(numeric_limits<UInt64>::max()) = 18446744073709551616 = 2^64  (rounded up)
```

The existing boundary check `x > Float64(numeric_limits<T>::max())` therefore fails to reject `Float64` values equal to the rounded-up boundary. The subsequent `static_cast<T>(x)` is undefined behavior.

UBSan output from issue #103817 (CI: `Unit tests (asan_ubsan, function_prop_fuzzer)` on PR #103580):

```
src/Common/FieldVisitorConvertToNumber.h:85:22: runtime error: 1.84467e+19 is outside the range of representable values of type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/Common/FieldVisitorConvertToNumber.h:85:22
```

`1.84467e+19` is exactly `2^64`. On a release build the same code silently truncates the value (typically to `0`), which is also wrong — `topK(1.8446744073709552e19)(...)` would proceed with `threshold = 0` instead of rejecting the input.

### Fix

Replace the buggy boundary check with `accurate::greaterOp` / `accurate::lessOp` (from `Core/AccurateComparison.h`), which uses `DecomposedFloat::compare` to perform precision-correct float-vs-integer comparison. This is the existing canonical helper for this kind of comparison and is already used elsewhere in `convertNumeric`, `FieldVisitorAccurateLess`, etc.

The same buggy pattern existed in two other places — fixed all three:

1. `src/Common/FieldVisitorConvertToNumber.h` — the reported site (used in aggregate function parameter parsing: `topK`, `histogram`, `uniqUpTo`, `groupArrayInsertAt`, etc.)
2. `src/Core/SettingsFields.cpp::fieldToNumber<T>` — used by `SET <integer_setting> = <Float64>`
[...]

PR #104476, opened 2026-05-09
Add test: `fieldToNumber<bool>` out-of-range Float64 branch in `SettingsFields.cpp` untested

clickgapai:

Adds test coverage for 1 untested code path(s), found during automated review of [PR #104154](https://github.com/ClickHouse/ClickHouse/pull/104154).
 That PR This PR fixes undefined behavior in `FieldVisitorConvertToNumber<T>::operator()(Float64)` and `fieldToNumber<T>` for wide integer types (UInt64, Int64, Int128, etc.) when the input Float64 is at the 2^64 boundary. The old code used `x > Float64(numeric_limits<T>::max())` which failed for wide intege

**1. `fieldToNumber<bool>` out-of-range Float64 branch in `SettingsFields.cpp` untested**
  `src/Core/SettingsFields.cpp:129`
  **Risk:** `fieldToNumber<bool>` at SettingsFields.cpp:127-130 branches on `std::is_same_v<T, bool>` to use naive comparison instead of `accurate::lessOp` (which won't compile for bool). Call chain: `SettingFiel
  **What's unique vs PR tests:** PR's test 04205_field_visitor_convert_to_number_ubsan_103817.sql tests the UInt64 boundary (1.8446744073709552e19) for aggregate functions and settings. My test covers the BOOL special case branch whi
  [Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/ea1c7d36-9cfe-4a93-9f0c-f9cef083e11e)

### DIFF
--- a/tests/queries/0_stateless/04212_settings_bool_out_of_range_float.reference
+++ b/tests/queries/0_stateless/04212_settings_bool_out_of_range_float.reference
@@ -1,0 +1,4 @@
+true
+false
+true
+true
--- a/tests/queries/0_stateless/04212_settings_bool_out_of_range_float.sql
+++ b/tests/queries/0_stateless/04212_settings_bool_out_of_range_float.sql
@@ -1,0 +1,26 @@
+-- Test: exercises `fieldToNumber<bool>` with out-of-range Float64 values
+-- Covers: src/Core/SettingsFields.cpp:129-130 — bool special case for out-of-range float
+-- The bool special case uses naive comparison (not `accurate::lessOp`) because
+-- `make_unsigned_t<bool>` is ill-formed. This test verifies the check works correctly.
+
+-- Out-of-range positive float should throw
+SET enable_analyzer = 1.5; -- { serverError CANNOT_CONVERT_TYPE }
+SET enable_analyzer = 2.0; -- { serverError CANNOT_CONVERT_TYPE }
+
+-- Out-of-range negative float should throw
+SET enable_analyzer = -0.5; -- { serverError CANNOT_CONVERT_TYPE }
+SET enable_analyzer = -1.0; -- { serverError CANNOT_CONVERT_TYPE }
+
+-- In-range floats should work (truncate to bool: 0->false, non-zero->true)
+SET enable_analyzer = 1.0;
+SELECT getSetting('enable_analyzer') AS v1 FORMAT TabSeparated;
+
+SET enable_analyzer = 0.0;
+SELECT getSetting('enable_analyzer') AS v2 FORMAT TabSeparated;
+
+-- Edge case: values between 0 and 1 should work (truncate to true since non-zero)
+SET enable_analyzer = 0.5;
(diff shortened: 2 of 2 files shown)


groeneai:
Reviewed this PR — it fills a real coverage gap from #104154. Approving.

**Coverage gap verification:** Legitimate. The merged #104154 added an `if constexpr (std::is_same_v<T, bool>)` branch to `fieldToNumber<T>` in `src/Core/SettingsFields.cpp` (lines 122–126 on master) that uses naive `Float64` comparison instead of `accurate::greaterOp` / `accurate::lessOp` (since `make_unsigned_t<bool>` is ill-formed). The existing test `04205_field_visitor_convert_to_number_ubsan_103817.sql` exercises only the wide-integer side via `topK(...)` parameters and `SET max_threads = ...` / `SET max_block_size = ...` — both `UInt64` paths. The new test is the first to drive the `bool` branch.

**Test logic verification:** All cases trace through `fieldToNumber<bool>` → Field::Types::Float64 branch → `if constexpr (std::is_same_v<T, bool>)` (the special-case branch). Verified locally against `26.5.1.1` debug build:

| Query | Path | Result |
|---|---|---|
| `SET enable_analyzer = 1.5` | `1.5 > Float64(1) = max<bool>` | throws `CANNOT_CONVERT_TYPE` ✓ |
| `SET enable_analyzer = 2.0` | `2.0 > Float64(1)` | throws ✓ |
| `SET enable_analyzer = -0.5` | `-0.5 < Float64(0) = lowest<bool>` | throws ✓ |
| `SET enable_analyzer = -1.0` | `-1.0 < Float64(0)` | throws ✓ |
| `SET enable_analyzer = 1.0` | in range, `T(1.0) = true` | `true` ✓ |
| `SET enable_analyzer = 0.0` | in range, `T(0.0) = false` | `false` ✓ |
| `SET enable_analyzer = 0.5` | in range, `T(0.5) = true` (non-zero) | `true` ✓ |
| `SET enable_analyzer = 0.999` | in range, `T(0.999) = true` (non-zero) | `true` ✓ |

`enable_analyzer` is confirmed `Bool` (alias of `allow_experimental_analyzer` declared in `src/Core/Settings.cpp:7367`), so the test does exercise `fieldToNumber<bool>`.

**Local repro:** `tests/clickhouse-test 04212_settings_bool_out_of_range_float --test-runs 10` → 10/10 passed under default random settings + random MergeTree settings on debug build.

**Nits (non-blocking):**

1. Test number `04212_` skips `04211_` (highest existing on master is `04210_`, so the conventional `add-test` script would have assigned `04211_`). No conflict since neither number is in use, but mentioning in case it matters to maintainers.
2. The header comment references `SettingsFields.cpp:129-130` — after the merge of #104154 the bool branch is at lines 122–126. Tiny correction, not worth a re-push.

**Label:** I'd add `can be tested` myself but I'm a fork contributor — that label requires a maintainer. cc @alexey-milovidov — could you flip the label so this can run CI?

alexey-milovidov (approved):
Weird.

PR #104477, opened 2026-05-09
Fix logical error in SelectIntersectExceptQueryVisitor on normalized AST

### LINKED ISSUE #103969: AST fuzzer: `Logical error: Incorrect ASTSelectWithUnionQuery (modes: 1, selects: 3)` (STID: 4337-3161)
The AST fuzzer hits a logical error `Incorrect ASTSelectWithUnionQuery (modes: 1, selects: 3)` raised from `SelectIntersectExceptQueryMatcher::visit` (`src/Interpreters/SelectIntersectExceptQueryVisitor.cpp:43`). It's been observed across multiple unrelated PRs in CI under the same STID `4337-3161`.

### How to reproduce

A reduced AST-fuzzed query body (the function body is only parsed/normalized, no execution is needed):

```sql
CREATE TABLE department__fuzz_56 (`id` Date, `parent_department` Array(UInt128), `name` Ring) ENGINE = TinyLog;
CREATE TABLE department__fuzz_65 (`id` Decimal(18, 18), `parent_department` Nullable(Enum8('v0' = 0, 'v1' = 1, 'v2' = 2)), `name` Enum16('v0' = 0, 'v1' = 1) NOT NULL) ENGINE = TinyLog;

CREATE FUNCTION IF NOT EXISTS f AS x ->
    (SELECT DISTINCT * FROM department__fuzz_56
     UNION ALL
     WITH RECURSIVE x AS (SELECT * FROM department__fuzz_65 UNION ALL (SELECT * FROM q UNION ALL SELECT * FROM x))
     SELECT DISTINCT * FROM x);
```

### Stack trace

```
12. src/Interpreters/SelectIntersectExceptQueryVisitor.cpp:43:15: DB::SelectIntersectExceptQueryMatcher::visit(...)
13. src/Interpreters/InDepthNodeVisitor.h:71:13: DB::InDepthNodeVisitor<...>::doVisit(...)
14. src/Interpreters/InDepthNodeVisitor.h:94:9:  ::visitChildren<false>(...)
... (deep recursion through visitImpl/visitImplMain/visitChildren)
```

The assertion is `union_modes.size() + 1 == selects.size()`. With `modes: 1, selects: 3`, the AST is in an inconsistent state — two selects appear without a corresponding union mode.

### Observed in CI

| PR | Time | Job |
|----|------|-----|
| #99740 | 2026-05-01 17:41 | AST fuzzer |
| #101155 | 2026-05-01 09:44 | AST fuzzer |
| #101155 | 2026-04-27 00:24 | AST fuzzer |
| #97254 | 2026-05-03 19:38 | AST fuzzer (amd_debug, targeted) |

Reproduce SQL for the most recent occurrence: https://s3.amazonaws.com/clickhouse-test-reports/PRs/97254/047c25f997572492ccd21cbd7e751c97782adcc6/ast_fuzzer_amd_debug_targeted/logical_error_incorrect_astselectwithunionquery_modes_a_selects_b_stid_4337_3161/reproduce_commands.sql

### Version info
- Resolved by: #104477
- Merged into: `26.5.1.488`

groeneai:
The AST fuzzer found `Logical error: Incorrect ASTSelectWithUnionQuery (modes: 1, selects: 3)` raised from `SelectIntersectExceptQueryMatcher::visit` at `src/Interpreters/SelectIntersectExceptQueryVisitor.cpp:43` when registering a SQL UDF whose body contains a parenthesized inner UNION. The bug appears as STID `4337-3161` across multiple unrelated PRs and on master via `BuzzHouse (experimental, serverfuzz, amd_tsan)`.

Auto-generated issue: https://github.com/ClickHouse/ClickHouse/issues/103969

### Minimal reproducer

```sql
CREATE FUNCTION f AS x ->
    (SELECT 1 UNION ALL (SELECT 1 UNION ALL SELECT 1));
```

### Root cause

For a `CREATE FUNCTION` query, `executeQueryImpl` runs `SelectIntersectExceptQueryVisitor` and then `NormalizeSelectWithUnionQueryVisitor` on the parsed AST. The normalizer flattens inner `UNION ALL` / `UNION DISTINCT` children into the parent's `list_of_selects` and sets `is_normalized = true` (with `union_mode` = `UNION_ALL` / `UNION_DISTINCT`), but does NOT update `list_of_modes` to match the new child count.

The interpreter then calls `InterpreterCreateFunctionQuery::execute` -> `UserDefinedSQLFunctionFactory::registerFunction` -> `normalizeCreateFunctionQuery`, which clones the AST and runs `SelectIntersectExceptQueryVisitor` a second time. On the now-flattened AST, the visitor's invariant `list_of_modes.size() + 1 == list_of_selects->children.size()` no longer holds and the assertion fires.

### Fix

Make `SelectIntersectExceptQueryMatcher::visit(ASTPtr &)` skip already-normalized union queries, mirroring the existing early-return check in `NormalizeSelectWithUnionQueryMatcher::visit(ASTPtr &)`. After normalization, no `INTERSECT` / `EXCEPT` modes can be present (the normalizer only sets `union_mode` to `UNION_ALL` or `UNION_DISTINCT`), so this matcher has nothing to do on such ASTs.

### Test plan

Added `04214_create_function_normalized_union_query` covering:

- the minimal `(... UNION ALL (... UNION ALL ...))` reproducer
- the full reproducer from #103969 (`WITH RECURSIVE` + parenthesized `UNION ALL`)
- `CREATE FUNCTION` with `INTERSECT` and `EXCEPT` bodies (must still normalize correctly)

Verified locally: the test fails (server `SIGABRT` from the assertion) before the fix and passes after; 30/30 runs with full settings randomization. Existing tests `02004_intersect_except_*`, `01856_create_function`, `02099_sql_user_defined_functions_lambda`, `03034_recursive_cte_tree_fuzz_crash_fix`, `03035_recursive_cte_postgres_1` still pass.

Fixed `Logical error: Incorrect ASTSelectWithUnionQuery (modes: M, selects: N)` triggered when a SQL user-defined function body contains a parenthesized inner `UNION ALL` (e.g. `CREATE FUNCTION f AS x -> (SELECT 1 UNION ALL (SELECT 1 UNION ALL SELECT 1))`).

Closes #103969

Session id: cron:clickhouse-ci-task-worker:20260509-174500

### Version info
- Merged into: `26.5.1.488`

### DIFF
--- a/src/Interpreters/SelectIntersectExceptQueryVisitor.cpp
+++ b/src/Interpreters/SelectIntersectExceptQueryVisitor.cpp
@@ -26,7 +26,15 @@ namespace ErrorCodes
 
 void SelectIntersectExceptQueryMatcher::visit(ASTPtr & ast, Data & data)
 {
-    if (auto * select_union = ast->as<ASTSelectWithUnionQuery>())
+    /// Skip already-normalized union queries: `NormalizeSelectWithUnionQueryMatcher`
+    /// flattens inner `UNION ALL` / `UNION DISTINCT` children into the parent's
+    /// `list_of_selects` without updating `list_of_modes`, so on a normalized AST
+    /// `list_of_modes.size() + 1 == list_of_selects->children.size()` no longer
+    /// holds. A normalized AST also cannot contain `INTERSECT` or `EXCEPT` modes
+    /// (`NormalizeSelectWithUnionQueryMatcher` only sets `union_mode` to
+    /// `UNION_ALL` or `UNION_DISTINCT`), so this matcher has nothing to do.
+    /// Mirrors the early-return check in `NormalizeSelectWithUnionQueryMatcher::visit`.
+    if (auto * select_union = ast->as<ASTSelectWithUnionQuery>(); select_union && !select_union->is_normalized)
         visit(*select_union, data);
 }
 
--- a/tests/queries/0_stateless/04214_create_function_normalized_union_query.reference
+++ b/tests/queries/0_stateless/04214_create_function_normalized_union_query.reference
@@ -1,0 +1,4 @@
+minimal repro OK
+recursive repro OK
+intersect repro OK
+except repro OK
--- a/tests/queries/0_stateless/04214_create_function_normalized_union_query.sql
+++ b/tests/queries/0_stateless/04214_create_function_normalized_union_query.sql
@@ -1,0 +1,57 @@
+-- Regression test for STID 4337-3161:
+-- "Logical error: Incorrect ASTSelectWithUnionQuery (modes: 1, selects: 3)"
+-- raised from `SelectIntersectExceptQueryMatcher::visit` when registering a
+-- SQL UDF whose body contains a parenthesized inner UNION.
+--
+-- Root cause: `executeQueryImpl` runs `SelectIntersectExceptQueryVisitor` and
+-- then `NormalizeSelectWithUnionQueryVisitor` on the parsed AST. The
+-- normalizer flattens inner `UNION ALL` children into the parent's
+-- `list_of_selects` and sets `is_normalized = true` (with `union_mode` =
+-- `UNION_ALL`), but does NOT update `list_of_modes` to match. When the
+-- `Interpreter` then runs `normalizeCreateFunctionQuery`, the
+-- `SelectIntersectExceptQueryVisitor` runs a second time on the now-flattened
+-- AST and triggers the assertion `list_of_modes.size() + 1 ==
+-- list_of_selects->children.size()`.
+--
+-- The fix makes `SelectIntersectExceptQueryMatcher::visit` skip already
+-- normalized union queries (no `INTERSECT` / `EXCEPT` modes can be present
+-- after normalization, so there is nothing for the matcher to do).
+--
+-- See https://github.com/ClickHouse/ClickHouse/issues/103969
+
+DROP FUNCTION IF EXISTS test_04214_repro;
+DROP FUNCTION IF EXISTS test_04214_recursive;
+DROP FUNCTION IF EXISTS test_04214_intersect;
+DROP FUNCTION IF EXISTS test_04214_except;
+
+-- Minimal reproducer: outer UNION ALL with a parenthesized inner UNION ALL.
+CREATE FUNCTION test_04214_repro AS x ->
+    (SELECT 1 UNION ALL (SELECT 1 UNION ALL SELECT 1));
+
+SELECT 'minimal repro OK';
+
+-- Full reproducer from issue #103969 (uses WITH RECURSIVE + parenthesized
+-- UNION ALL inside the recursive CTE body).
+CREATE FUNCTION test_04214_recursive AS x ->
+    (SELECT DISTINCT number FROM numbers(2)
+     UNION ALL
+     WITH RECURSIVE r AS (SELECT 1 UNION ALL (SELECT * FROM numbers(2) UNION ALL SELECT 3))
+     SELECT DISTINCT 1);
+
+SELECT 'recursive repro OK';
+
+-- INTERSECT inside a UDF body must still be normalized correctly (the fix
+-- must not regress this path).
+CREATE FUNCTION test_04214_intersect AS x -> (SELECT 1 INTERSECT SELECT 1);
+
+SELECT 'intersect repro OK';
+
+-- Same for EXCEPT.
+CREATE FUNCTION test_04214_except AS x -> (SELECT 1 EXCEPT SELECT 1);
+
+SELECT 'except repro OK';
+
+DROP FUNCTION test_04214_repro;
+DROP FUNCTION test_04214_recursive;
+DROP FUNCTION test_04214_intersect;
+DROP FUNCTION test_04214_except;
--- a/tests/queries/0_stateless/04214_create_function_normalized_union_query.sql
+++ b/tests/queries/0_stateless/04214_create_function_normalized_union_query.sql
@@ -1,3 +1,10 @@
+-- Tags: no-parallel
+-- ^ User-defined SQL functions are stored at the server level (not per
+-- database), so concurrent runs of this test would collide on the
+-- `CREATE FUNCTION` global namespace and fail with `FUNCTION_ALREADY_EXISTS`.
+-- All other UDF tests under `0_stateless/` use the same tag for this reason
+-- (e.g. `01856_create_function`, `02098`-`02103`, `02125`).
+
 -- Regression test for STID 4337-3161:
 -- "Logical error: Incorrect ASTSelectWithUnionQuery (modes: 1, selects: 3)"
 -- raised from `SelectIntersectExceptQueryMatcher::visit` when registering a


groeneai:
## Pre-PR validation gate

Mandatory self-review per `TASK.md` Phase 4:

**(a) Deterministic repro?** Yes. `CREATE FUNCTION f AS x -> (SELECT 1 UNION ALL (SELECT 1 UNION ALL SELECT 1));` triggers the assertion every time on a debug build. Verified that running the exact same DDL on an unfixed binary produces server `SIGABRT` (exit 134) inside the same call stack the auto-issue captured.

**(b) Root cause explained?** Yes.
1. The parser produces an `ASTSelectWithUnionQuery` with `list_of_modes = [UNION_ALL]` and `list_of_selects = [outer_select, inner_ASTSelectWithUnionQuery(2 children, 1 mode)]` — invariant holds: `1 + 1 == 2`.
2. `executeQueryImpl` runs `SelectIntersectExceptQueryVisitor` (no-op for plain `UNION`), then `NormalizeSelectWithUnionQueryVisitor` which lifts the inner two children into the outer `list_of_selects`. The outer now has `list_of_selects.children.size() == 3` but `list_of_modes.size() == 1` — invariant `modes + 1 == selects` is BROKEN. `is_normalized` is set to `true`, `union_mode` to `UNION_ALL`.
3. `InterpreterCreateFunctionQuery::execute` -> `UserDefinedSQLFunctionFactory::registerFunction` -> `normalizeCreateFunctionQuery` clones the AST (clone preserves `is_normalized`) and runs `SelectIntersectExceptQueryVisitor` a SECOND time. The visitor's check `list_of_modes.size() + 1 != list_of_selects->children.size()` now fires `1 + 1 != 3` -> `Logical error`. Confirmed by adding diagnostic prints showing `normalized: true, union_mode: 1, modes: 1, selects: 3`.

**(c) Fix matches root cause?** Yes. The visitor's only job is to convert `INTERSECT` / `EXCEPT` modes from `list_of_modes` into `ASTSelectIntersectExceptQuery` nodes embedded in `list_of_selects`. After normalization, `list_of_modes` cannot contain `INTERSECT` or `EXCEPT` modes (`NormalizeSelectWithUnionQueryMatcher` only sets `union_mode` to `UNION_ALL` or `UNION_DISTINCT`), so the visitor has nothing to do. The fix mirrors the existing `if (... && !select_union->is_normalized)` early-return in `NormalizeSelectWithUnionQueryMatcher::visit(ASTPtr &)`.

**(d) Test intent preserved?** Yes. The fix only adds an early return for already-normalized ASTs. Fresh ASTs with `INTERSECT` / `EXCEPT` modes still go through the full conversion logic. Verified by `02004_intersect_except_operators`, `02004_intersect_except_distinct_operators`, `02004_intersect_except_const_column` all passing post-fix. The new test also covers `CREATE FUNCTION ... INTERSECT ...` and `CREATE FUNCTION ... EXCEPT ...` to lock in this behavior.

- WITHOUT the fix: rebuilt binary, ran `04214_create_function_normalized_union_query` -> server crashed with `Exit 134` (SIGABRT) before printing `1 tests passed`.
- WITH the fix: same test passes 30/30 with `--test-runs 30` under full settings randomization.

**(f) Fix is general, not a narrow patch?** Yes.
- The fix targets the abstraction (the visitor entry point) at the layer where the invariant is checked. It handles ALL paths where `SelectIntersectExceptQueryVisitor` could be invoked on a normalized AST, not just the `CREATE FUNCTION` path.
- Searched all callers of `SelectIntersectExceptQueryVisitor`: `executeQuery.cpp`, `DatabaseOrdinary.cpp`, `DatabaseBackup.cpp`, `DatabaseReplicated.cpp`, `StorageView.cpp`, `resolveStorages.cpp`, `InterpreterSystemQuery.cpp`, `UserDefinedSQLFunctionFactory.cpp`. The interlock is the same in each case (`Intersect` -> `Normalize`), so any caller that re-runs the pair on an already-normalized AST gets the same protection.
- "Symptom vs root cause" check: the bad value (a normalized AST) is produced upstream by `executeQueryImpl` before the interpreter runs. We could alternatively fix `normalizeCreateFunctionQuery` to skip work on a pre-normalized clone, but that hides the structural mismatch from other callers and conflicts with how `NormalizeSelectWithUnionQueryMatcher` itself solves the same idempotency problem (early-return on `is_normalized`). Putting the check in the visitor itself fixes the symptom AND aligns the two visitors' contracts.

Session id: cron:clickhouse-ci-task-worker:20260509-174500

groeneai:
cc @vdimir @novikd — could you review this? Small fix for an AST fuzzer assertion (STID 4337-3161, issue #103969): `SelectIntersectExceptQueryMatcher::visit` was firing `Logical error: Incorrect ASTSelectWithUnionQuery` on already-normalized union queries when `registerFunction` re-runs the visitor pair on a clone. The fix mirrors the existing `is_normalized` early-return in `NormalizeSelectWithUnionQueryMatcher`.

PedroTadim:
@groeneai the test you added `04214_create_function_normalized_union_query` failed. Can you fix?

groeneai:
@PedroTadim — fixed in `1f4212a` (`Mark 04214_create_function_normalized_union_query as no-parallel`).

**Root cause** (confirmed from the flaky-check logs across all 4 sanitizers):

```
Code: 609. DB::Exception: User-defined object 'test_04214_repro' already exists:
while adding user defined function `test_04214_repro`. (FUNCTION_ALREADY_EXISTS)
```

User-defined SQL functions live in the server-level `UserDefinedSQLObjectsStorage`, not in the test's per-test `currentDatabase()`. The flaky check repeats each new/changed test ~50 times across 5 parallel workers, so two concurrent runs race on the same global `CREATE FUNCTION test_04214_repro` and one loses (or, symmetrically, races on `DROP FUNCTION` with `UNKNOWN_FUNCTION`).

This is the same race every other `CREATE FUNCTION` test in `0_stateless/` already handles — `01856_create_function`, `02098`/`02099`/`02101`/`02102`/`02103`/`02125` all carry `-- Tags: no-parallel`. The new regression test had to do the same.

**Local verification (debug build):**

Without `no-parallel` (extracted to a temp `04215_*` file):
```
$ tests/clickhouse-test --test-runs 30 -j 8 04215_test_collision_temp
... [ FAIL ] (multiple times — FUNCTION_ALREADY_EXISTS / UNKNOWN_FUNCTION)
Having 3 errors! 0 tests passed. ...
Having 2 errors! 2 tests passed. ...
```

With the fix (`-- Tags: no-parallel` on `04214_create_function_normalized_union_query.sql`):
```
$ tests/clickhouse-test --test-runs 50 -j 8 04214_create_function_normalized_union_query
50 tests passed. 0 tests skipped. 13.42 s elapsed.
```

The C++ fix in commit `1099d45` is unchanged — this is purely a test-stability fix. The Bugfix-validation failure on the previous push was a downstream effect of the same race (the test could not run reliably enough for the validation framework to observe its master-vs-PR behaviour); it should pass on this re-run.

Note on `no-parallel` vs the general "avoid no-* tags" guideline: it is strictly necessary here because `CREATE FUNCTION` registers in a server-level (not per-database) namespace, so two parallel runs of the same test cannot avoid colliding short of using random function names per run, and the established convention in this repo is to use `no-parallel` for UDF tests.

alexey-milovidov (approved):

Issue #104219, opened 2026-05-06
Test `01710_projection_additional_filters` is flaky


nickitat:
_Important: This issue was automatically generated and is used by CI for matching failures. DO NOT modify the body content. DO NOT remove labels._

Test name: 01710_projection_additional_filters
Failure reason: 2026-05-11 17:44:20 -3
CI report: [Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104643&sha=2a5ad83dd51adc92af5ecded678ef0fe31826b44&name_0=PR&name_1=Stateless%20tests%20%28amd_llvm_coverage%2C%20ParallelReplicas%2C%20s3%20storage%2C%20parallel%29)

[...]

PR #104478, opened 2026-05-09
Enable query condition cache for `ORDER BY ... LIMIT n` queries

alexey-milovidov:
- Performance Improvement

Enable the query condition cache for queries that go through TopK dynamic filtering (`ORDER BY ... LIMIT n`). The QCC entry is partitioned by the TopK plan parameters so the same query reuses cached WHERE filter results, while a different `LIMIT`, sort column, or sort direction produces a fresh entry.

---

#### Why

When `tryOptimizeTopK` picks a `ReadFromMergeTree` for TopK dynamic filtering, the running threshold can drop granules before they reach the WHERE/PREWHERE filter. Until now `updateQueryConditionCache` blanket-skipped QCC writes for such reads:

```cpp
/// ORDER BY ... LIMIT N can drop granules, don't update qcc for the WHERE filter
if (read_from_merge_tree->isSelectedForTopKFilterOptimization())
    return;
```

The skip was correct but pessimistic — it disables QCC for *every* query that hits the TopK path, even when a re-run with the same plan and same parts would reuse exactly the same row set.

#### What changed

1. **`TopKFilterInfo::condition_hash`** — a deterministic hash over the planning-time TopK parameters: `(column_name, type_name, num_sort_columns, limit_n, direction)`. Computed in `tryOptimizeTopK` when the read is selected for TopK filtering.
2. **`ReadFromMergeTree::setTopKColumn`** — folds a hash of the visible part-set snapshot (the names of all parts the query reads) into `condition_hash`. A TopK granule-skip decision for one part depends on the `__topKFilter` threshold, which is derived from the rows of *all* parts the query reads. Since the QCC key is `(table_uuid, part_name, condition_hash)`, an entry written for an unchanged part would otherwise stay matchable after a *different* part is dropped or mutated, even though the threshold that justified it is gone. Salting with the part-set snapshot makes any change to the set of parts read (`DROP PARTITION`, mutation, merge, new `INSERT`) produce a fresh key, so stale decisions are never reused; a stable part set keeps the same key and still hits.
3. **`updateQueryConditionCache`** — instead of skipping for TopK reads, combine `top_k_filter_info->condition_hash` into the WHERE filter's `condition_hash` via `boost::hash_combine` before storing it on `FilterStep`.

The QCC writer (`FilterTransform::writeIntoQueryConditionCache`) and the read-side lookup both go through the same `FilterStep::condition`, so the combined hash is used consistently on both sides — same plan + same parts → cache hit; different `LIMIT n`, different sort column, or any change to the visible part set → new key, fresh entry, never reusing a row set computed under different TopK conditions.

#### Origin

Split out from WIP PR #81944 (commit `bec285616a2`). The PR's original implementation was on its own `TopNFilterParameters` infrastructure that didn't survive the rebase onto master; this version keeps the same idea but uses master's `TopKFilterInfo` / `tryOptimizeTopK` directly.

### Version info
- Merged into: `26.7.1.316` (included in `26.7` and later)


alexey-milovidov (commented):
Add functional tests for this optimization.
Add performance tests as well.

src/Processors/QueryPlan/Optimizations/updateQueryConditionCache.cpp
```diff
@@ -59,7 +57,19 @@ void updateQueryConditionCache(const Stack & stack, const QueryPlanOptimizationS
...
+            /// they differ (e.g. Apple, where `size_t` is `unsigned long` but `UInt64` is `unsigned long long`).
+            size_t condition_hash = filter_actions_dag->getOutputs()[0]->getHash();
+
+            /// `ORDER BY ... LIMIT N` may drop granules during reading, so the result of the WHERE
+            /// filter is no longer "applies to every granule of every part" — it applies only to
+            /// the granules that the TopK filter decided to keep. To keep the QCC entry sound, we
+            /// fold the deterministic part of the TopK plan into the cache key. Same query + same
+            /// part set + same TopK params → cache hit; different LIMIT or sort column → fresh
+            /// entry, never reusing a row-set computed under different TopK conditions.
+            if (const auto & top_k_filter_info = read_from_merge_tree->getTopKFilterInfo())
```
shankar-iyer, line 106: I will study this in detail - this is where I got stuck in my branch.

alexey-milovidov:
There is no visible difference on the newly introduced performance test.

alexey-milovidov:
Still no significant difference - correct the test so that it will be visible.

alexey-milovidov:
Pushed `3c6741458b6` to address @clickhouse-gh's `Performance Improvement` request: the perf test now uses 10M rows of `FixedString(256)` filled with `randomFixedString(256)` so each granule scan reads an order-100 MB chunk of incompressible bytes. Locally this widens the QCC-on vs QCC-off gap from a noisy fraction-of-a-millisecond into a ~20x speedup (1 ms cached vs ~36 ms full scan), which should clear the CI perf comparator's noise floor on both `amd_release` and `arm_release` shards.

The two failing checks at `d46a8dce38b` are unrelated to this PR:

1. **AST fuzzer (amd_debug, targeted) → `Received signal 6 (STID: 4193-4b55)`** — a logical error in `OptimizeShardingKeyRewriteInVisitor.cpp:183`. The fuzzer reduced its way to `SELECT (2 IN tuple(-2)) FROM dist_01756 ... GROUP BY ... (dummy IN tuple()) AND ...`. The `(dummy IN tuple())` triggers `tuple.back()` on an empty `Tuple` inside `OptimizeShardingKeyRewriteIn::enterImpl`. Pre-existing UB in distributed-query AST optimization, no relation to QCC/TopK. I'll send a separate fix.
2. **Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel) → `01710_projection_additional_filters`** — known flaky, tracked in #104219 (assigned to @azat). The same `0 0 0 / -3 vs +23` reference diff has been firing across many PRs daily since 2026-05-05 and has appeared on master CI as well.

@groeneai, the AST fuzzer signal-6 above is unrelated to this PR — please investigate the `OptimizeShardingKeyRewriteIn` empty-tuple crash and provide a fix in a separate PR. If a fix is already in progress, please link it here.

groeneai:
@alexey-milovidov — confirmed the `OptimizeShardingKeyRewriteIn` empty-tuple crash and its fix is already in flight:

- **Fix PR:** #104966 — "Fix exception on `IN tuple()` against Distributed sharded table"
- **STID:** 4193-4b55

**Root cause matches your description:** the analyzer-side `OptimizeShardingKeyRewriteIn::enterImpl` at `src/Interpreters/OptimizeShardingKeyRewriteInVisitor.cpp:183` calls `tuple.back()` on a `Tuple` reduced by the AST fuzzer to `()` (empty). The AST-side `OptimizeShardingKeyRewriteInMatcher::visit` (same file, lines 124-125) already guards the same operation with `tuple.size() > 1`; the analyzer-side path didn't, so the same query reaches the analyzer path under the new analyzer and trips UB. The fix in #104966 adds the symmetric `if (tuple.empty()) return;` guard and ships a regression test (`04238_shardkey_rewrite_in_empty_tuple`).

**CI on #104966 head `3ae9adc`:** 4 failures, none related to the fix (the change only affects the `IN tuple()` against a sharded `Distributed` table code path):
- `04229_trivial_group_by_limit_profile_events` (Stateless, amd_llvm_coverage, ParallelReplicas, s3) — 5 PR fails / 0 master in last 14d, known flaky.
- `Hung check failed` on `Stress test (amd_debug)` — 13 fails / 11 PRs / 2 master in 14d, chronic infra flake.
- `Hung check failed` on `Stress test (arm_ubsan)` — 10 fails / 10 PRs / 0 master, chronic infra flake.
- `Bugfix validation (functional tests)` — single `FAIL` row for `04238_shardkey_rewrite_in_empty_tuple` on `head_ref=fix-shardkey-rewrite-empty-tuple`, consistent with the expected "test fails on master HEAD (no fix), passes on PR HEAD (with fix)" check. The same test passes on regular Stateless runs of the PR head (e.g. `amd_msan, WasmEdge`).

I'll continue monitoring #104966 via CIDB; once it merges I'll verify master goes clean on STID 4193-4b55.

Tracked in `groeneai` task `2026-05-14-github-clickhouseclickhouse104478--`.

src/Processors/QueryPlan/ReadFromMergeTree.cpp
```diff
@@ -2699,8 +2707,18 @@ ReadFromMergeTree::AnalysisResultPtr ReadFromMergeTree::selectRangesToRead(
         if (reader_settings.use_query_condition_cache && query_info_.filter_actions_dag && !query_info_.isFinal())
         {
             const auto & outputs = query_info_.filter_actions_dag->getOutputs();
-            if (outputs.size() == 1 && VirtualColumnUtils::isDeterministic(outputs.front()))
-                condition_hash = outputs.front()->getHash();
+            /// `isNodeDeterministic` allows `__topKFilter`'s non-determinism (it's gated by the
+            /// TopK plan salt below, mirroring `updateQueryConditionCache`'s write path).
+            if (outputs.size() == 1 && isNodeDeterministic(outputs.front()))
```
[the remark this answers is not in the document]
alexey-milovidov: Addressed in c5270e4f909 (sibling commit 234891e6d49). Added a separate `isDeterministicAllowingTopKFilter` mirroring the helper in `updateQueryConditionCache.cpp` — it keeps the `COLUMN`-node strictness of `VirtualColumnUtils::isDeterministic` (rejects non-deterministic constants like `now()` / `today()`) while admitting `__topKFilter`. `isNodeDeterministic` reverts to its master form, so the row-policy gate that shares it is unaffected.

src/Processors/QueryPlan/Optimizations/updateQueryConditionCache.cpp
```diff
@@ -51,15 +85,27 @@ void updateQueryConditionCache(const Stack & stack, const QueryPlanOptimizationS
...
```
shankar-iyer, line 106: @groeneai Is there any impact on query condition cache due to dropping of granules via minmax index when executing ORDER BY ... LIMIT N? That dropping happens in DB::MergeTreeReaderIndex::canSkipMark() with use_skip_indexes_on_data_read = 1. You should check MergeTreeSelectProcessor::read() where the query condition cache is updated after granules are dropped

shankar-iyer: @groeneai Can you check the previous comment? Can you review now with the knowledge of PR https://github.com/ClickHouse/ClickHouse/pull/105686 ?

alexey-milovidov:
Pushed `50956c24156` — this should finally make the `query_condition_cache_topk` perf comparison show the improvement.

**Why the previous reshapes did not register a delta.** The speedup is real — I built this branch and measured locally: a warm QCC hit on the TopK query returns in ~1 ms versus a ~33 ms full WHERE-filter scan (~30×). But the perf comparison kept showing both `use_query_condition_cache = 1` and `= 0` at the same full-scan cost.

The cause is **part-name churn from background merges**. QCC entries are keyed on `(table_uuid, part_name, condition_hash)`. The fill inserts 10M rows into several parts; a background merge then consolidates them into a new part with a different name. The perf framework primes the cache during its prewarm run (against the parts that exist then), but by the time the measured runs execute, the merge has renamed the part — so every measured run reads a part the cache was never primed for, misses, and re-pays the full scan. Both queries end up measuring the same cost and the improvement is invisible.

**Fix:** collapse the table into a single, fully-merged part with `OPTIMIZE TABLE ... FINAL` and freeze its name with `SYSTEM STOP MERGES` before the measured runs, so the prewarm-primed entries stay valid.

Verified end-to-end through `tests/performance/scripts/perf.py` against a freshly-built binary of this branch: query 0 (`= 1`) is stable at ~1.2 ms across all measured runs, while query 1 (`= 0`) stays at the full-scan cost. On the master side of the comparison the QCC is still skipped for TopK reads, so query 0 there stays at the full-scan cost too — the comparison should now flag it as a clear improvement.

src/Processors/QueryPlan/Optimizations/updateQueryConditionCache.cpp
shankar-iyer: I checked the code, the QCC search hash is updated with the all the `TopKFilterInfo` fields - that is the critical piece and that will ensure that the QCC entry is not retrieved/used by another query that only has `WHERE`.

shankar-iyer (approved):

tests/queries/0_stateless/04217_query_condition_cache_topk.sql
```diff
@@ -0,0 +1,91 @@
...
+         add_minmax_index_for_numeric_columns = 0;
+
+INSERT INTO tab SELECT rand(), number, number FROM numbers(1_000_000);
+
+SELECT '--- QCC starts empty';
+SYSTEM CLEAR QUERY CONDITION CACHE;
+SELECT count() FROM system.query_condition_cache;
+
+SELECT '--- Same TopK plan re-runs reuse the same QCC entry';
+SELECT v1 FROM tab WHERE v2 = 10000 ORDER BY v1 ASC LIMIT 5 FORMAT Null;
```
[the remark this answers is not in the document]
alexey-milovidov: Added a dedicated correctness regression in `04309_query_condition_cache_topk_correctness.sql`. Unlike `04217` (which only checks entry counts with single-row predicates and `FORMAT Null`), it exercises the risky path directly:

- The table uses `id = rand()` so the sort column `k` is scattered across granules — the TopK threshold can't drop whole granules via the minmax index, so `__topKFilter` drops individual rows inside every granule it reads before the `WHERE` `FilterStep` records skippable granules in the QCC. `w = number % 1000` makes the matching set span the whole table.
- The first `use_query_condition_cache = 1` run with an active `__topKFilter` writes a QCC entry (asserted via `count() > 0`), and the second run returns exactly the same rows as the `use_query_condition_cache = 0` baseline (result equality, not entry counts).
- A larger `LIMIT` (needs rows whose `k` the LIMIT-5 threshold dropped) and the opposite sort direction (needs the largest-`k` rows, which the ASC plans recorded as skippable) both still return the complete, correct result. The DESC case is the load-bearing regression: reusing the ASC entry would lose every row, so its correctness proves the TopK salt on the cache key keeps the plans separate.

I dropped the before/after QCC entry-count assertions from the cross-plan cases because whether a larger-`LIMIT` plan records a *new* skippable granule depends on the multi-threaded `__topKFilter` threshold trajectory and is not deterministic; the result-equality checks are the deterministic safety net. Verified the reference is stable across repeated runs against a local server.

groeneai:
@shankar-iyer Yes, the minmax dropping in `MergeTreeReaderIndex::canSkipMark` (the `threshold_tracker->isValueInsideThreshold` branch) does interact with QCC, and the current code handles it correctly. Walking the path you pointed at:

1. With `use_skip_indexes_on_data_read = 1`, `canSkipMark` drops a mark when its granule's min/max sort value is outside the current TopK threshold. Those marks never enter `res.read_mark_ranges`.
2. `res.read_mark_ranges` is what `MergeTreeSelectProcessor::readImpl` records into `MarkRangesInfo` for the QCC write. So for a TopK read, the recorded "matched" marks conflate "`WHERE` did not match" with "TopK threshold dropped it" - exactly the unsafe case if such an entry were reused by a plain `WHERE` query.
3. The safety comes from the cache key. `updateQueryConditionCache` now folds `getTopKFilterInfo()->condition_hash` into the `condition_hash` for the `FilterStep`. A `WHERE`-only query (no TopK) computes a different hash and can never retrieve the TopK-tainted entry; a different `LIMIT` or sort column also gets a fresh entry. This is the piece you confirmed on 2026-06-01.
4. The PREWHERE write path in `MergeTreeSelectProcessor::read` is also safe: `__topKFilter` is non-deterministic, so the `VirtualColumnUtils::isDeterministic(output)` gate `continue`s and no PREWHERE QCC entry is written when the TopK filter is present.

With PR #105686 in the picture the dropping is more aggressive (per-granule minmax threshold tracking), but it does not change the soundness argument: the threshold only tightens over a run, so any granule dropped at any point has no row that could reach the final top-N, and the salted key prevents cross-plan reuse. `04309_query_condition_cache_topk_correctness.sql` (added 2026-06-03) now exercises the risky path directly - scattered sort column so `__topKFilter` drops rows rather than whole granules - so the regression is covered. LGTM on the correctness side.

alexey-milovidov:

alexey-milovidov:
Pushed a fix for the failing CI check, plus a merge with `master` (the branch was 623 commits behind).

### CI failure: flaky `04309_query_condition_cache_topk_correctness`

This test failed in `amd_asan_ubsan` (flaky check + db disk/distributed plan), `amd_debug`, and `amd_tsan` on the `count() > 0` assertion after the first `ORDER BY k ASC LIMIT 5` run — it returned `0`, i.e. no QCC entry was written.

**Root cause (reproduced locally):** QCC population for the TopK plan is best-effort. A granule is recorded as skippable only when a chunk physically covering it is reduced to zero rows by the `WHERE` filter *and* that buffer is flushed before the query ends. With several reading threads over several parts — a parallel `INSERT` under a randomized `max_insert_threads` leaves multiple parts, and `max_threads` is randomized too — the running `__topKFilter` threshold each thread sees (and hence which chunks empty out, and whether their buffer flushes before the `LIMIT` cancels the pipeline) is timing-dependent. So `count() > 0` after a single early-cancelling TopK `LIMIT` query is not a deterministic invariant under the random settings CI applies.

Confirmed empirically on a fresh build: with 16 parts and `max_threads = 3` the assertion failed ~70% of runs, and a randomized `max_block_size` larger than the part size suppressed the write entirely. The feature itself is correct — every row-correctness assertion (cached re-run, larger `LIMIT`, `DESC`) passed on every run; only the population smoke-check was racy.

**Fix (test-only):** make the warm-up deterministic without weakening the test — read with a single thread (`SET max_threads = 1`) over a single part (`OPTIMIZE TABLE tab FINAL`). The salt/correctness assertions still run and now reliably exercise a populated cache: the ASC plan writes one entry, the `DESC` plan gets a separate salted entry, and the cached `DESC` result is the 5 largest values (not poisoned by the ASC entry). Verified deterministic across 30 full runs with the exact failing CI random settings plus a sweep of block-size / insert-thread / thread-count combinations. The companion `04217` test is unaffected (its `WHERE v2 = 10000` matches one row, so its `LIMIT 5` never early-cancels).

### Still open

The benchmark thread (`tests/performance/query_condition_cache_topk.xml`) — the current perf report shows no measurable delta for the QCC-enabled query. Because `__topKFilter` already drops most rows before the `WHERE` runs, the QCC's incremental saving in this shape is small; making it a stable, visible CI delta needs a benchmark-shape decision I'd leave to you.

tests/performance/query_condition_cache_topk.xml
```diff
@@ -0,0 +1,64 @@
...
+        <!-- Keep WHERE as a separate `FilterStep` above `ReadFromMergeTree` so the
+             dynamic-filtering branch of `tryOptimizeTopK` applies (it bails out
+             when `getPrewhereInfo()` is already set) and the WHERE condition is
+             the one written into the query condition cache. -->
+        <optimize_move_to_prewhere>0</optimize_move_to_prewhere>
+        <enable_parallel_replicas>0</enable_parallel_replicas>
+    </settings>
+
+    <!-- QCC enabled: prewarm primes the cache, measured runs reuse it. -->
+    <query>SELECT v1 FROM qcc_topk_test WHERE v2 = 12345 ORDER BY v1 ASC LIMIT 10 SETTINGS use_query_condition_cache = 1 FORMAT Null</query>
```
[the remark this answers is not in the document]
alexey-milovidov: I dug into why the perf report shows no delta and want to record the findings, because the optimization itself is working — the benchmark just isn't capturing it in the comparison report.

On a local server build of this branch the cache warms exactly as intended. Reproducing the benchmark shape (single fully-merged part, `SYSTEM STOP MERGES`, prewarm then measured runs), at `max_threads = 12` (the perf-comparison profile value) and 10M rows × `FixedString(256)`:

```
-- cold (use_query_condition_cache = 0), x3:  0.066 / 0.064 / 0.065 s
-- warm (use_query_condition_cache = 1), x3:  0.001 / 0.001 / 0.001 s
```

and the server log confirms the WHERE granules are all skipped on the measured run:

```
Query condition cache has dropped 1221/1221 granules for WHERE condition
    and(__topKFilter(v), equals(payload, 'no_match'_String)).
```

So the measured query genuinely reads zero granules (≈1 ms) versus a ≈65 ms full scan — a real ~60 ms / ~60× delta. I verified this holds across thread counts (12 and 96), with a single part + `SYSTEM STOP MERGES`, and with the profile-level `use_query_condition_cache = 0` overridden by the inline per-query `SETTINGS use_query_condition_cache = 1` (the override does take effect and does populate the cache).

I could not reproduce the "no gain" locally under any configuration I tried, and there is nothing in the perf-comparison server config that disables the cache (`query_condition_cache_size` is the default ~106 MB; the only `use_query_condition_cache = 0` is the profile setting, which the inline query setting overrides). The `server_time` metric in the report is also flat, so it is not a `client_time`-only artifact — in the comparison environment the cache appears not to warm at all between the prewarm and the measured runs, for a reason I have not been able to reproduce.

I just merged master and pushed, which re-triggers a fresh perf comparison. If that run still shows a flat result, the discrepancy is in the comparison harness / environment rather than in the optimization, and is worth a separate look — the granule-skip and the speedup are demonstrably present on a normal server.

EmeraldShift:
Does the QCC store just the result of the deterministic part of WHERE, or store the final LIMITed granules? If it's after LIMIT, is this change safe when the table can be UPDATEd? Consider two simple parts:
1. [1, 2, 3]
2. [4, 5, 6]

and ORDER BY c DESC LIMIT 2.  At first, it selects [6, 5], and maybe QCC records 0 for all of part 1. Then, UPDATE c = 0 WHERE c > 4. Now the parts look like:
1. [1, 2, 3]
2. [4, 0, 0]

The same query should now return [4, 3]. Will QCC allow reading part 1?

In other words, it looks like UPDATE can change whether the query reads parts that aren't involved in the update. Is the QCC invalidated for those parts?

src/Processors/QueryPlan/Optimizations/updateQueryConditionCache.cpp
```diff
@@ -51,15 +85,27 @@ void updateQueryConditionCache(const Stack & stack, const QueryPlanOptimizationS
...
+                boost::hash_combine(condition_hash, top_k_filter_info->condition_hash);
```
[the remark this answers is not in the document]
alexey-milovidov: Fixed in `f92a52df67b`. `ReadFromMergeTree::setTopKColumn` now folds a `SipHash` of the visible part-set snapshot (the names of all parts the query reads) into the TopK salt that is combined into `condition_hash`.

The concrete trace is exactly the case this closes: once part B is mutated the visible part set changes, so part A's `condition_hash` changes and its stale entry is no longer matchable; `filterPartsByQueryConditionCache` then recomputes against the new threshold. `DROP PARTITION`, merges and new `INSERT`s invalidate the same way; a stable part set keeps the same key and still hits the cache.

New regression test `05013_query_condition_cache_topk_part_set_invalidation` covers it — the final query returns an empty (wrong) result without the part-set salt.

alexey-milovidov:
@EmeraldShift Good catch — you are right, and this was a real hole. The QCC stores per-part granule-skip decisions for the deterministic `WHERE`, not the final LIMITed rows, but for the TopK plan a decision recorded for one part depends on the `__topKFilter` threshold, which is derived from the rows of *all* parts the query reads. The key was `(table_uuid, part_name, condition_hash)` and `condition_hash` only encoded the plan parameters, so your scenario held: an `UPDATE`/`DROP PARTITION` that changes one part leaves the unchanged part's entry matchable while the threshold that justified it is gone.

Fixed in `f92a52df67b`: `ReadFromMergeTree::setTopKColumn` now folds a hash of the whole visible part-set snapshot into the TopK salt, so any change to the set of parts read (`UPDATE`/mutation, `DROP PARTITION`, merge, new `INSERT`) produces a fresh key and the stale decisions of the other parts are never reused. Your exact two-part `ORDER BY c DESC LIMIT 2` example is the regression test `05013_query_condition_cache_topk_part_set_invalidation` — it returns an empty result without the salt and the correct rows with it.

alexey-milovidov:
Pushed `0e9f2696404` (plus a merge with `master`, the branch was 690 commits behind).

The only red on the previous head was the style check `test_numbers_check`, reporting `Gap (4336, 5013) > 100`. The cause was the new regression test, which the `add-test` helper had numbered `05013` (it picks the next free number relative to unrelated local files), leaving a 677-wide gap to the next functional test (`04337` after the merge). The gap check flags any gap of at least 100.

Fix: renumbered `05013_query_condition_cache_topk_part_set_invalidation` to `04338` — the next number after master's current highest test — so the functional-test numbering is contiguous and the gap check passes. The test content is unchanged. Verified locally by replaying the exact CI gap-check logic over the committed tree: no gap of 100 or more remains.

The merge was conflict-free. Master touched only `optimizeTopK.cpp` among this PR's source files (the arrayJoin lift-up fix #82279), which is orthogonal to the QCC salt logic; the QCC infrastructure (`QueryConditionCache`, `FilterTransform`, the TopK threshold tracker) was untouched. I rebuilt the four affected translation units (`optimizeTopK`, `updateQueryConditionCache`, `ReadFromMergeTree`, `MergeTreeDataSelectExecutor`) against the merged tree and they compile cleanly.

tests/performance/query_condition_cache_topk.xml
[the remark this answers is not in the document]
alexey-milovidov: Addendum to the analysis above: I reproduced the benchmark with the **actual `perf.py` harness** (the same script the CI performance comparison runs), against a server built from this branch, to rule out any divergence between a manual reproduction and what CI executes.

```
$ perf.py tests/performance/query_condition_cache_topk.xml --host localhost --port 9111 --runs 7

query #0  (use_query_condition_cache = 1):  500 runs, median 1.05 ms  (min 1.02, max 1.57)
query #1  (use_query_condition_cache = 0):   26 runs, median 38.8 ms
```

So under the exact CI tool the QCC-enabled query is a stable ~37× faster than the disabled baseline, with `1221/1221` WHERE granules dropped on every measured run. The prewarm primes the entry and all measured runs hit it.

I also re-checked the factors that could plausibly differ in the comparison environment, and none of them reproduce the flat result locally:

* profile-level `use_query_condition_cache = 0` overridden by the inline per-query `SETTINGS use_query_condition_cache = 1` — the override populates and hits the cache;
* single fully-merged part vs. multiple parts (`SYSTEM STOP MERGES`) — both hit;
* `optimize_move_to_prewhere` on or off, and even a query carrying *only* `use_query_condition_cache = 1` with every other setting at default — still hits.

This matches the earlier conclusion: the optimization and the benchmark are correct and demonstrably fast under the CI measurement tool itself, so the flat perf-comparison report points at the two-server comparison environment (the new-server cache apparently not warming between prewarm and measured runs there), not at the optimization or the benchmark shape. That part is worth a separate look at the comparison harness/environment.

alexey-milovidov:
Pushed `e5dd47789c5`: merged `master` (the branch was 2558 commits / ~6 days behind). The merge was conflict-free; `master` had auto-merged its own changes into this PR's hot files (`ReadFromMergeTree.cpp`, `MergeTreeDataSelectExecutor.cpp`), so I rebuilt the four affected translation units against the merged tree — `optimizeTopK`, `updateQueryConditionCache`, `ReadFromMergeTree`, `MergeTreeDataSelectExecutor` all compile cleanly. The QCC/TopK logic itself is untouched.

The three reds on the previous head (`1fe88045`) are all unrelated flakes:

1. **`Hung check failed, possible deadlock found`** on `Stress test (arm_debug)` and `Stress test (arm_msan)` — chronic, master-wide infra flake: it fires 50–120 times/day across dozens of unrelated PRs (CIDB confirms it on `base_ref = 'master'` runs daily for the last 30 days). Not specific to this change.
2. **`AST fuzzer (amd_debug)` → `Client failure`** — a generic `MEMORY_LIMIT_EXCEEDED` (total server RSS reached ~44 GiB during broad fuzzing); the exception is thrown in `TCPHandler::receiveHello` while a reconnecting client tries to allocate, not in any QCC/TopK frame. The *targeted* AST fuzzer variants, which exercise exactly this PR's changed lines (66 tests each), both passed. The change only computes bounded `size_t` hashes, so it can't accumulate the memory that tripped the global limit.

A fresh CI run on the merge head will re-run all three.

alexey-milovidov:
Pushed `bb64b8367e9` to fix the only remaining red, **`CH Inc sync`** (`tests failed`).

**Root cause.** The private sync PR ([`ClickHouse/clickhouse-private#57579`](https://github.com/ClickHouse/clickhouse-private/pull/57579)) merged and built cleanly, but its flaky check on the slower **`s3 storage, meta in keeper`** configuration failed the two new QCC TopK tests:

```
[ 2 / 100] 04217_query_condition_cache_topk:                          [ FAIL ] 181.91 sec.
Reason: Test runs too long (> 180s). Make it faster.
[ 9 / 100] 04338_..._part_set_invalidation:                           [ FAIL ] 187.55 sec.
```

Under ASan/UBSan + S3 each run sits right at the 180s `Test runs too long` gate in `tests/clickhouse-test` (`flaky_check and total_time > TEST_MAX_RUN_TIME_IN_SECONDS and "long" not in self.tags`) and intermittently crosses it. The public flaky checks run on local disk, stayed under 180s, and all passed — which is why this only showed up in the private S3 lane.

**Why not just make them faster.** Both tests need ~1M+ rows for the QCC to populate at all (a granule-spanning chunk has to be fully filtered before the `LIMIT` cancels the pipeline), so the data volume is structurally required — cutting it loses the cache-population assertions. Unlike `04320_query_condition_cache_topk_correctness`, which you sped up by dropping an `OPTIMIZE ... FINAL`, these two have no removable bottleneck; their cost is the inherent data volume.

**Fix.** Tag both as `long`, which is exactly the marker that flaky-check gate honors. The tests still run in all the regular stateless jobs; the tag only suppresses the spurious "too long" failure and gives headroom for the slower private S3 config. No SQL or reference changes.

Everything else is green: 159 ✅ / 14 skipped, AI Review **✅ Approve**, Cloud perf **`no_change`**, LLVM coverage 98.88% of changed lines, approved by @shankar-iyer. A fresh sync run on `bb64b8367e9` should clear the check.

alexey-milovidov:
Re-checked CI on `bb64b8367e9` (no code changes needed; AI Review **✅ Approve**, approved by @shankar-iyer, LLVM coverage 98.88% of changed lines). All three reds are unrelated to this PR and each is already tracked or in-flight — verified against CIDB (`base_ref = 'master' OR pull_request_number = 0`):

1. **AST fuzzer (amd_debug) — `Block structure mismatch in UnionStep` (STID 0993-27f0).** Tracked open issue #108142. 154 occurrences in 90 days, **all** on master across 10+ unrelated PRs (108818, 105667, 108838, 105890, 50053, 86768, 96978, 107667, 104232, …). The reduced query is `SELECT DISTINCT … FROM (… UNION ALL …) … GROUP BY GROUPING SETS …` — a `UnionStep` header-nullability mismatch (`Const(Nullable(Nothing))` vs `Nullable(Nothing)`) in the analyzer, with no QCC/TopK frame.

2. **Upgrade check (amd_release) — `Error message in clickhouse-server.log`.** Chronic master-wide flake: 135 occurrences across 124 unique PRs in 30 days. The errors are `DatabaseMySQL`/`DatabasePostgreSQL` failing to reach the reserved test address `192.0.2.1` — background connection-retry log noise, unrelated to QCC/TopK.

3. **Stress test (arm_asan_ubsan, s3) — UBSan `DateLUTImpl.h:1148` (STID 0250-45ad).** Signed-integer overflow in `DateLUTImpl::toStartOfHourInterval` (`Time res = values.date + time`), exercised by the stress AST fuzzer on date functions. Already being fixed by #108848 ("Fix signed integer overflow in `DateLUTImpl::toStartOfHourInterval`"). Also fires on unrelated PRs 97254 and 107415.

I re-ran the three failed (flaky) jobs on run `28451911323` to try to clear them.

Did not re-merge `master`: the branch was merged yesterday (`e5dd47789c5`, merge-base `2026-06-29`), so it is < 1 day old, and all three reds also fail on `master`, so a re-merge cannot fix them.

One note for your awareness (not a blocker): the CI Performance Comparison and Cloud perf both report **no measurable change** for `query_condition_cache_topk` — consistent with your 2026-06-05 observation that surfacing the gain as a stable CI delta needs a benchmark-shape decision you wanted to keep. Correctness coverage is complete; this is only about the synthetic benchmark visibility.

alexey-milovidov (commented):
Plausible.

tests/queries/0_stateless/04338_query_condition_cache_topk_part_set_invalidation.sql
```diff
@@ -0,0 +1,100 @@
...
+
+-- Dropping partition 0 removes the rows the threshold was computed from. The new
+-- top-5 lives entirely in partition 1, whose part is unchanged (same name) and
+-- whose granules the warm run recorded as skippable. Reusing those entries would
+-- return an empty result; the part-set hash in the TopK salt must invalidate them.
+ALTER TABLE tab DROP PARTITION 0;
+
+SELECT '--- After DROP PARTITION: ground truth (QCC off) comes from partition 1';
+SELECT k FROM tab WHERE w = 7 ORDER BY k DESC LIMIT 5 SETTINGS use_query_condition_cache = 0;
+SELECT '--- QCC on must not reuse granule decisions recorded under the old part set';
```
PedroTadim, line 97: @groeneai could you check the query condition cache is being used by checking the ProfileEvents check on system.query_log?

PedroTadim: @groeneai could you also check for negative limits and order by desc?

groeneai: Done in #110381 (test-only follow-up on 04338):

- QCC usage now checked via `system.query_log` ProfileEvents: warm run has `QueryConditionCacheHits = 0`, cached run has hits, and the run after `DROP PARTITION` has 0 hits (the part-set salt invalidation, not just a correct result).
- Negative LIMIT: `LIMIT -n` becomes a `NegativeLimit` plan step with no `__topKFilter`, so the TopK-QCC path never engages. Added a case asserting it returns the correct rows and neither reuses nor poisons the TopK entries.
- ORDER BY DESC was already covered; also added an ASC case, which must not reuse the DESC entry since `condition_hash` folds the sort direction (verified via 0 hits).

Verified 20/20 with full randomization.

PR #104479, opened 2026-05-09
Update system.functions table with more details

PedroTadim:
- New Feature

Add `deterministic` and `higher_order` columns to `system.functions` table.

The `system.functions` table got new columns: `deterministic`, and `higher_order` . This helps with automated testing. For each of these columns, if the value is unknown, it outputs NULL.

### Version info
- Merged into: `26.5.1.750`


docs/en/operations/system-tables/functions.md
```diff
@@ -27,6 +27,10 @@ Contains information about normal and aggregate functions.
 - `examples` ([String](/sql-reference/data-types/string)) — Usage example.
 - `introduced_in` ([String](/sql-reference/data-types/string)) — ClickHouse version in which the function was first introduced.
 - `categories` ([String](/sql-reference/data-types/string)) — The category of the function.
+- `min_arguments` ([Nullable(UInt64)](/sql-reference/data-types/int-uint)) — The minimum number of arguments the function accepts. `NULL` when unknown (e.g. variadic functions, aggregate functions, executable user-defined functions).
+- `max_arguments` ([Nullable(UInt64)](/sql-reference/data-types/int-uint)) — The maximum number of arguments the function accepts. `NULL` when unbounded (variadic) or unknown.
+- `is_deterministic` ([Nullable(UInt8)](/sql-reference/data-types/int-uint)) — Whether the function returns the same result for the same arguments. `NULL` when unknown (e.g. aggregate or user-defined functions).
+- `accepts_lambda` ([Nullable(UInt8)](/sql-reference/data-types/int-uint)) — Whether the function is higher-order — i.e. accepts at least one lambda expression as an argument (e.g. `arrayMap`, `arrayFilter`, `mapApply`). `NULL` when unknown.
```
rschu1ze: Can we rename `accepts_lambda` to `higher_order_function`?

docs/en/operations/system-tables/functions.md
```diff
@@ -27,6 +27,10 @@ Contains information about normal and aggregate functions.
 - `examples` ([String](/sql-reference/data-types/string)) — Usage example.
 - `introduced_in` ([String](/sql-reference/data-types/string)) — ClickHouse version in which the function was first introduced.
 - `categories` ([String](/sql-reference/data-types/string)) — The category of the function.
+- `min_arguments` ([Nullable(UInt64)](/sql-reference/data-types/int-uint)) — The minimum number of arguments the function accepts. `NULL` when unknown (e.g. variadic functions, aggregate functions, executable user-defined functions).
```
rschu1ze: The only case where `min/max_arguments` theoretically make sense are variadic functions - for non-variadic functions `min/max_arguments` is just the number of arguments. However, for variadic functions, `min_arguments` shows NULL, which makes it useless.

Also,
- `min/max_arguments` duplicate the information returned by `syntax` (the caller can simply count the commas)
- `min/max_arguments` as names is misleading, it sounds like the function rejects input values smaller/bigger than `min/max_arguments`.

I'd be happier if we could omit `min/max_arguments`.

PedroTadim: Some scalar functions can accept 1, 2, or 3 arguments. This helps in those cases. The information from `syntax` sometimes cannot be easily parsed or could contain errors.

rschu1ze: > Some scalar functions can accept 1, 2, or 3 arguments. 
> The information from syntax sometimes cannot be easily parsed

An example for a non-variadic function with different overloads looks like this:

```
func(arg1);
func(arg2, arg3);
func(arg4, arg5, arg6);
```

It looks simple to parse, actually.

> or could contain errors.

If there are errors in the `syntax` fields, we better fix then, instead of keeping them around an introducing another field to `system.functions`.

docs/en/operations/system-tables/functions.md
PedroTadim: Done

docs/en/operations/system-tables/functions.md
PedroTadim: Ok, I can write some tricky SQL queries for this

PedroTadim:
@rschu1ze may I merge this once tests pass?

alexey-milovidov:

docs/en/operations/system-tables/functions.md
```diff
@@ -27,6 +27,8 @@ Contains information about normal and aggregate functions.
 - `examples` ([String](/sql-reference/data-types/string)) — Usage example.
 - `introduced_in` ([String](/sql-reference/data-types/string)) — ClickHouse version in which the function was first introduced.
 - `categories` ([String](/sql-reference/data-types/string)) — The category of the function.
+- `is_deterministic` ([Nullable(UInt8)](/sql-reference/data-types/int-uint)) — Whether the function returns the same result for the same arguments. `NULL` when unknown (e.g. aggregate or user-defined functions).
+- `higher_order_function` ([Nullable(UInt8)](/sql-reference/data-types/int-uint)) — Whether the function is higher-order — i.e. accepts at least one lambda expression as an argument (e.g. `arrayMap`, `arrayFilter`, `mapApply`). `NULL` when unknown.
```
alexey-milovidov: Just `higher_order`? "function" is redundant.

docs/en/operations/system-tables/functions.md
```diff
@@ -27,6 +27,8 @@ Contains information about normal and aggregate functions.
 - `examples` ([String](/sql-reference/data-types/string)) — Usage example.
 - `introduced_in` ([String](/sql-reference/data-types/string)) — ClickHouse version in which the function was first introduced.
 - `categories` ([String](/sql-reference/data-types/string)) — The category of the function.
+- `is_deterministic` ([Nullable(UInt8)](/sql-reference/data-types/int-uint)) — Whether the function returns the same result for the same arguments. `NULL` when unknown (e.g. aggregate or user-defined functions).
```
alexey-milovidov: `deterministic`, not `is_deterministic`
To be consistent with `case_insensitive`.

Anticipating the question about `is_aggregate` - in that case, `is` is reasonable to have, as it represents the whole different class of functions rather than a single property.

src/Storages/System/StorageSystemFunctions.cpp
```diff
@@ -99,7 +110,48 @@ namespace
...
+        try
+        {
+            auto resolver = factory.tryGet(name, context);
+            if (!resolver)
+                return info;
+
+            info.is_deterministic = resolver->isDeterministic() ? UInt8{1} : UInt8{0};
+            info.higher_order_function = resolver->isHigherOrder() ? UInt8{1} : UInt8{0};
+        }
+        catch (...)
```
alexey-milovidov: When does it happen? Can we assume it can't happen and remove the try/catch?

alexey-milovidov (changes requested):
Overall, very good PR, thank you!

src/Storages/System/StorageSystemFunctions.cpp
PedroTadim: As seen in the failures here https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104479&sha=bb01615a4d00b123de46341baa0dcab570606c35&name_0=PR&name_1=Fast+test the catch block is needed

docs/en/operations/system-tables/functions.md
```diff
@@ -27,6 +27,8 @@ Contains information about normal and aggregate functions.
...
+- `higher_order_function` ([Nullable(UInt8)](/sql-reference/data-types/int-uint)) — Whether the function is higher-order — i.e. accepts at least one lambda expression as an argument (e.g. `arrayMap`, `arrayFilter`, `mapApply`). `NULL` when unknown.
 <!--AUTOGENERATED_END-->
```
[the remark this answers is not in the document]
PedroTadim: It is already updated

src/Storages/System/StorageSystemFunctions.cpp
```diff
@@ -230,6 +296,11 @@ void StorageSystemFunctions::fillData(MutableColumns & res_columns, ContextPtr c
         res_columns[11]->insertDefault(); // examples
         res_columns[12]->insertDefault(); // introduced_in
         res_columns[13]->insertDefault(); // categories
+
+        /// Determinism is unknown — the WASM module is opaque to ClickHouse — so report NULL.
```
rschu1ze: I think we can do without the comment in l. 299-301.

tests/queries/0_stateless/03262_system_functions_should_not_fill_query_log_functions.sql
```diff
@@ -1,4 +1,8 @@
-SELECT * FROM system.functions WHERE name = 'bitShiftLeft' format Null;
+-- Project only the non-Nullable `name` column. SELECT * would project the new
```
rschu1ze: Someone reading this in a few weeks from now without knowing this PR will wonder why `deterministic` and `higher-order` are "new" columns. Let's remove this word.

tests/queries/0_stateless/04212_system_functions_arity_determinism.sql
```diff
@@ -0,0 +1,61 @@
+-- Tags: no-parallel
+-- no-parallel: SQL UDFs are global state and race with concurrent test workers (matches 02101_sql_user_defined_functions_create_or_replace.sql).
+
+-- New columns on system.functions: deterministic, higher_order.
+
+SELECT '-- columns exist with the expected types';
```
rschu1ze, line 4: Looks like there is no test that checks the schema of `system.functions`. Perhaps it's better to run `SHOW CREATE TABLE system.functions` here.

tests/queries/0_stateless/04212_system_functions_arity_determinism.sql
```diff
@@ -0,0 +1,61 @@
...
+
+SELECT '-- deterministic non-higher-order function';
+SELECT name, deterministic, higher_order
+FROM system.functions WHERE name = 'plus';
+
+SELECT '-- non-deterministic function';
+SELECT name, deterministic, higher_order
+FROM system.functions WHERE name = 'today';
+
+SELECT '-- another deterministic non-higher-order function (variadic)';
```
rschu1ze: The fact that we are testing variadic functions is probably a leftover from an earlier state of the PR.

rschu1ze (approved):

tests/queries/0_stateless/04212_system_functions_arity_determinism.sql
PedroTadim: The `02117_show_create_table_system` test does it, and the output was updated.

PedroTadim:
@alexey-milovidov good to merge now?

alexey-milovidov (approved):

src/Functions/IFunction.h
```diff
@@ -572,6 +571,9 @@ class IFunction
     virtual bool isShortCircuit(ShortCircuitSettings & /*settings*/, size_t /*number_of_arguments*/) const { return false; }
     virtual bool isSuitableForShortCircuitArgumentsExecution(const DataTypesWithConstInfo & /*arguments*/) const = 0;
 
+    /// Higher-order functions accept at least one lambda expression as an argument.
+    virtual bool isHigherOrderFunction() const { return false; }
```
rschu1ze, line 575: I would expect that higher-order functions (e.g. `arrayCount`) override `isHigherOrderFunction`, but that does somehow not happen in this PR. I think it should, otherwise the system table returns incorrect information.

src/Storages/System/StorageSystemFunctions.cpp
```diff
@@ -99,7 +110,49 @@ namespace
...
+        else
+            res_columns[14]->insertDefault();
+
+        if (extra.higher_order)
+            res_columns[15]->insert(*extra.higher_order);
+        else
+            res_columns[15]->insertDefault();
+    }
+
+    /// Resolve an ordinary function and read static metadata from its overload resolver.
```
rschu1ze: We would dilute the code a lot less if we could get rid of `getOrdinaryFunctionExtraInfo` and `ExtraInfo` and instead change `fillRow` in this way:

```cpp
[...]
if constexpr (std::is_same_v<Factory, FunctionFactory>)
{
    /// <do here what l. 134 - 153> do
    //// <then write to res_columns[14] and res_columns[15]>
}
else
{
    res_columns[14]->insertDefault(); // is_deterministic
    res_columns[15]->insert(UInt8{0}); // higher_order
}
```

rschu1ze (changes requested):

src/Functions/IFunction.h
PedroTadim: `arrayCount` is from `FunctionArrayMapped` which sets `isHigherOrderFunction` to true.
```sql
SELECT higher_order
FROM system.functions
WHERE name = 'arrayCount' --1
```

rschu1ze: Ah, `isHigherOrderFunction` existed alrady before. I didn't see that, sorry.

rschu1ze (approved):

PedroTadim:
@alexey-milovidov, why did you remove the no-parallel tag? A SQL function is being re-created. This is going to be flaky if the test runs in parallel.

PR #104271, opened 2026-05-06
Draft: improve TimeSeries PromQL-to-SQL coverage

BadLiveware:
### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Not for changelog yet. This is a draft visibility PR for early maintainer feedback and is expected to be split into smaller reviewable PRs before merge.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

Documentation is not complete in this draft. If maintainers agree with the direction, the split PRs will include user-facing documentation for the supported PromQL surface and remaining histogram limitations.

---

## Draft / early-feedback PR

This PR makes visible an ongoing effort to improve ClickHouse TimeSeries native PromQL-to-SQL conversion coverage. It is intentionally a large integration branch and **not intended to merge as one PR**. The goal is to let maintainers see the current direction, semantic choices, validation shape, and split plan before the work is carved into smaller upstreamable PRs.

Development disclosure: this branch is **mostly AI-developed**, with human review/steering. The semantics and test coverage are a port of learnings from [`BadLiveware/promshim-clickhouse`](https://github.com/BadLiveware/promshim-clickhouse): specifically, PromQL behavior and PromQL-to-ClickHouse-SQL compatibility knowledge. It does **not** port promshim's Go IR or renderer wholesale; the implementation keeps ClickHouse's existing `PrometheusQueryTree::Node -> SQLQueryPiece -> ASTPtr` conversion architecture.

## What this integration branch contains

High-level areas included in the draft branch:

- Compliance/test infrastructure for upstream Prometheus PromQL compliance cases plus fork-local regression cases.
- Simple over-time reducers and counter/range functions, including `changes`, `resets`, `increase`, `deriv`, and `predict_linear`.
- `quantile_over_time` and quantile phi edge semantics.
- Set binary operators and vector matching/label-set handling improvements.
- Label transformation functions: `label_replace` and `label_join`.
- `count_values`, `absent`, and `absent_over_time`.
- Scalar/vector edge cases: date function defaults, `timestamp`, `round`, `clamp_*`, selector validation, subquery/offset alignment.
- Classic `histogram_quantile` support over classic `_bucket{le=...}` float series. Native histogram implementation remains deferred pending storage/result representation agreement.

## Compliance coverage

[...]

PR #104490, opened 2026-05-09
Reuse identical PromQL vector operands

BadLiveware:
Optimizes native PromQL binary operations where both vector operands are the same expression.

When semantics allow it, the lowering reuses a single vector-grid input instead of planning two equivalent operands and joining them. The fast path stays behind the normal duplicate-series checks for filter comparisons where default matching ignores `__name__`.

Split out from the integration draft #104271; shared base #104484.

- Performance Improvement

Improved native PromQL-to-SQL lowering for binary operations with identical vector operands by reusing one vector-grid input when PromQL semantics allow it.


alexey-milovidov:

BadLiveware:
This is mostly a theoretical optimization, its very unlikely to hit actual user queries in its current form. An expanded form could but Im closing this in its current form

BadLiveware:
This is mostly a theoretical optimization, its very unlikely to hit actual user queries in its current form. An expanded form could but Im closing this in its current form

PR #104505, opened 2026-05-10
Fix bucket-aware merging dropped by analyzer with ARRAY JOIN over remote source

### LINKED ISSUE #104289: Analyzer generates single `MergingAggregatedTransform` for distributed GROUP BY with `LEFT ARRAY JOIN`; coordinator OOMs
### Company or project name

tiara

### Describe what's wrong

With `allow_experimental_analyzer = 1`, a Distributed query that has `LEFT ARRAY JOIN` plus a high-cardinality `GROUP BY` and multiple aggregate states (`uniqIf`, `anyHeavy`) plans into a single `MergingAggregatedTransform` on the coordinator. All shards' partial states are funneled through `Resize N → 1` into that one transform, and the coordinator OOMs once the merged group state exceeds the per-query memory limit.

The old analyzer plans the same query as `GroupingAggregatedTransform N → 1` + `MergingAggregatedBucketTransform × M` (bucket-aware) and finishes fine on the same data.

**EXPLAIN PIPELINE comparison** (both captured on the same 26.3.9.8 coordinator; the new-analyzer plan was also confirmed identical on 25.12.2.54):

Old analyzer (works):

```text
ExpressionTransform × 40
  Resize 40 → 40
    MergingAggregatedBucketTransform × 40
      Resize 1 → 40
        GroupingAggregatedTransform 150 → 1
          ReadFromRemote
```

New analyzer (regresses):

```text
ExpressionTransform × 40
  Resize 1 → 40
    MergingAggregatedTransform
      Resize 150 → 1
        ReadFromRemote
```

The difference is at the `(MergingAggregated)` step. Bucket-aware merging is replaced by a single transform plus a `Resize 150 → 1` that funnels all shards through one stream. Outer sort/limit pipeline is identical and omitted.

**Coordinator OOM** under the new analyzer:

```text
Code: 241. DB::Exception: Query memory limit exceeded:
  would use 119.21 GiB (attempt to allocate chunk of 4.30 MiB),
  maximum: 119.21 GiB.
  (MEMORY_LIMIT_EXCEEDED) (version 26.3.9.8 (official build))
```

~9.6 s to OOM, peak `memory_usage` 119.21 GiB. Each shard finishes normally at 2.5–3.2 GiB peak, so the OOM is exclusively on the coordinator merge step.

**Settings I tried** on top of `allow_experimental_analyzer = 1` (none brings back the bucket-aware plan or avoids the OOM with correct results):

| Settings | Result |
|---|---|
| `group_by_two_level_threshold = 1`, `group_by_two_level_threshold_bytes = 1`, `distributed_aggregation_memory_efficient = 1`, `enable_memory_bound_merging_of_aggregation_results = 1`, `aggregation_memory_efficient_merge_threads = 8` | OOM, plan unchanged |
| above + `max_bytes_before_external_group_by = 50 GB`, `max_bytes_before_external_sort = 50 GB` | OOM. These apply to per-shard aggregation and sort phases, not the coordinator merge — there's no equivalent spill path for `MergingAggregatedTransform`. |
| `max_threads = 2` with 30 GB spill | Slower path; reaches 86 GiB before I cancelled |
| `distributed_group_by_no_merge = 2`, `optimize_distributed_group_by_sharding_key = 1` | Completes, **but result is wrong** — per-shard top-30 unioned without final merge. Same `uv_c` comes back as ~44k vs ~6.4M from the old analyzer. |

Other combinations of `optimize_aggregation_in_order = 0`, `prefer_localhost_replica = 0`, `compile_aggregate_expressions = 0`, `prefer_global_in_and_join = 1`, `distributed_push_down_limit = 1`, plus wrapping `LEFT ARRAY JOIN` in an explicit subquery — all leave the EXPLAIN PIPELINE unchanged.

`enable_memory_bound_merging_of_aggregation_results` engages only for "aggregation in order"; our GROUP BY doesn't match the table's sort order, so it's a no-op by design — included in the matrix for completeness.

**Workaround**: `SETTINGS allow_experimental_analyzer = 0`. That's the only thing we found that returns correct results without OOMing.

Yes

### How to reproduce

**ClickHouse server version**: confirmed on `25.12.2.54 (official build)` and `26.3.9.8 (official build)` LTS. Issue persists across both. The new-analyzer plan was also unchanged after 13+ different setting combinations and after upgrading the coordinator from 25.12.2.54 to 26.3.9.8 LTS.

**Interface**: native TCP (clickhouse-client) and HTTP both reproduce; interface does not matter.

**Cluster topology**: ~450 nodes, 150 shards × 3 replicas. Coordinator: 128 GiB physical RAM, `max_memory_usage = 119 GiB`. Linux x86_64.

**Non-default settings** (only the ones relevant to triggering):

- `allow_experimental_analyzer = 1` — required to trigger
- `allow_experimental_analyzer = 0` — workaround, does not trigger

**CREATE TABLE statements** (anonymized; structure mirrors production):

```sql
CREATE TABLE events_local ON CLUSTER 'cluster'
(
    date_id      Int32,
    group_a      LowCardinality(String),
    group_b      LowCardinality(String),
    event_type   LowCardinality(Nullable(String)),
    event_kind   LowCardinality(Nullable(String)),
    metric       Nullable(Int64),
    meta_id      Nullable(String),
    meta_type    Nullable(String),
    meta_name    Nullable(String),
    attr_a       Nullable(String),
    attr_b       Nullable(String),
    user_hash    UInt64,
    session_hash UInt64,
    is_outlier   LowCardinality(Nullable(String)),
    `items_nested.id`   Array(Nullable(String)) CODEC(ZSTD(1)),
    `items_nested.type` Array(Nullable(String)) CODEC(ZSTD(1)),
    `items_nested.name` Array(Nullable(String)) CODEC(ZSTD(1))
)
ENGINE = ReplicatedMergeTree('/clickhouse/tables/{shard}/events_local', '{replica}')
PARTITION BY date_id
ORDER BY (group_a, group_b, cityHash64(user_hash))
SETTINGS allow_nullable_key = 1;

CREATE TABLE events ON CLUSTER 'cluster' AS events_local
ENGINE = Distributed('cluster', 'default', 'events_local', cityHash64(user_hash));
```

**Sample data**: I can't share production data. The bug is plan-shape-driven, so the data values themselves don't matter — only that the WHERE filter produces a GROUP BY with roughly 1–3M distinct `(gid, gtype, gname)` tuples. With ~50 KB of `uniqIf` HLL + `anyHeavy` state per group, that lines up with the 119 GiB coordinator memory ceiling.

For plan inspection alone, no data is needed at all — `EXPLAIN PIPELINE` reproduces the structural diff immediately on an empty cluster as long as the cluster has multiple shards. Happy to put together a `generateRandom` / `numbers()`-based loader for OOM reproduction if it would help triage.

**Query that leads to the unexpected result**:

```sql
WITH
    if(event_type = 'A', meta_id,   `items_nested.id`)   AS gid,
    if(event_type = 'A', meta_type, `items_nested.type`) AS gtype,
    if(event_type = 'A', meta_name, `items_nested.name`) AS gname,
    anyHeavy(attr_a) AS any_attr_a,
    anyHeavy(attr_b) AS any_attr_b,
    uniqIf(user_hash,    event_kind = 'C' AND metric > 0) AS uv_c,
    uniqIf(session_hash, event_kind = 'C' AND metric > 0) AS sess_c,
    sumIf(metric, event_kind = 'C') AS m_c,
    countIf(event_type, lowerUTF8(event_type) = 'b') AS m_b,
    uniqIf(user_hash, metric > 0) AS uv_total
SELECT gid, gtype, gname, any_attr_a, any_attr_b, uv_c, sess_c, m_c, m_b
FROM events
LEFT ARRAY JOIN items_nested
WHERE date_id BETWEEN 20260101 AND 20260107
    AND group_a = 'A1' AND group_b = '12345'
    AND event_type IN ('A', 'B')
    AND (is_outlier = 'N' OR is_outlier IS NULL)
GROUP BY gid, gtype, gname
ORDER BY uv_total DESC, gid, gtype, gname
LIMIT 30
SETTINGS allow_experimental_analyzer = 1;
```

To inspect the plan without OOM-scale data, run:

```sql
EXPLAIN PIPELINE
<query above>
SETTINGS allow_experimental_analyzer = 1;
```

Then compare with `allow_experimental_analyzer = 0`. The structural diff at the `(MergingAggregated)` step (single `MergingAggregatedTransform` vs. `GroupingAggregatedTransform` + `MergingAggregatedBucketTransform × N`) is reproducible on any multi-shard cluster — the OOM only manifests once the GROUP BY has enough cardinality.

### Expected behavior

For a Distributed source with a high-cardinality GROUP BY and multiple aggregate states, coordinator memory should stay bounded. Any of the following would satisfy that:

- The new analyzer planning the same bucket-aware merge as the old analyzer (`GroupingAggregatedTransform` + `MergingAggregatedBucketTransform × N`), so coordinator memory is bounded by `total_state_size / N` rather than the full `total_state_size`
- Memory-bound merging engaging for this query shape (it currently doesn't, since the GROUP BY isn't aggregation-in-order)
- A spill path on the coordinator merge step (`MergingAggregatedTransform` currently has none)

Right now none of those engage under the new analyzer, and the only escape is to disable the analyzer entirely. That won't be a long-term option as the old analyzer is being phased out.

Happy to run further diagnostics if any of these would help triage:

- Whether dropping `LEFT ARRAY JOIN` (grouping only on scalar columns) restores the bucket-aware plan under the new analyzer
- Whether reducing to a single `uniqIf` (removing the other aggregates) still triggers the regression
- A bisection across earlier 25.x patch releases to identify the first regressed version
- Adding a `generateRandom`-based loader to make the OOM reproducible on a smaller test cluster

_No response_

### Additional context

_No response_

nahomar:

Fix memory-efficient aggregation for distributed queries with `ARRAY JOIN`. This reduces query memory usage and  fixes a regression with the new analyzer.

Closes https://github.com/ClickHouse/ClickHouse/issues/104289

## Summary

Fixes #104289. Under the new query analyzer (`allow_experimental_analyzer = 1`), a distributed query with `[LEFT] ARRAY JOIN` over a remote source loses bucket-aware memory-efficient merging on the coordinator. All shards' partial aggregates are funneled through `Resize N → 1` into a single `MergingAggregatedTransform`; the issue reporter observed coordinator OOM at ~119 GiB across 150 shards, and the only known workaround is `allow_experimental_analyzer = 0`.

## Root cause

In [`src/Planner/Planner.cpp:756-773`](https://github.com/ClickHouse/ClickHouse/blob/master/src/Planner/Planner.cpp#L756-L773), `addMergingAggregatedStep` only computed `is_remote_storage` / `parallel_replicas_from_merge_tree` when `planner_context->getTableExpressionNodeToData().size() == 1`:

```cpp
if (table_expression_node_to_data.size() == 1)
{
    auto it = table_expression_node_to_data.begin();
    is_remote_storage = it->second.isRemote();
    parallel_replicas_from_merge_tree = it->second.isMergeTree() && query_context->canUseParallelReplicasOnInitiator();
}
```

`ARRAY JOIN` (and regular `JOIN`, and subqueries) registers an additional entry in `table_expression_node_to_data` via `getOrCreateTableExpressionData` ([`CollectTableExpressionData.cpp:84`](https://github.com/ClickHouse/ClickHouse/blob/master/src/Planner/CollectTableExpressionData.cpp#L84)). The size check fails, both flags stay at their defaults, and the `memory_efficient_aggregation` argument to `MergingAggregatedStep` flips to `false`. `MergingAggregatedStep::transformPipeline` ([`MergingAggregatedStep.cpp:130-138`](https://github.com/ClickHouse/ClickHouse/blob/master/src/Processors/QueryPlan/MergingAggregatedStep.cpp#L130-L138)) then takes the single-stream `MergingAggregatedTransform` branch instead of `addMergingAggregatedMemoryEfficientTransform`, which builds `GroupingAggregatedTransform` + `MergingAggregatedBucketTransform × N`.

The old analyzer is unaffected because [`InterpreterSelectQuery::executeMergeAggregated`](https://github.com/ClickHouse/ClickHouse/blob/master/src/Interpreters/InterpreterSelectQuery.cpp#L3019-L3038) reads from the single `storage` member, which `ARRAY JOIN` does not replace.

## Fix

Walk every entry in `table_expression_node_to_data` and OR the flags. `ARRAY JOIN` entries naturally don't pollute the result because [`CollectTableExpressionData.cpp:372-385`](https://github.com/ClickHouse/ClickHouse/blob/master/src/Planner/CollectTableExpressionData.cpp#L372-L385) only calls `setIsRemote` / `setIsMergeTree` for `TableNode` and `TableFunctionNode`. This also incidentally enables bucket-aware merging for `JOIN` queries that have at least one remote source — happy to narrow the scope if maintainers prefer to keep that out.

## Verification

- ✅ Reproduced on a 2-shard cluster running 25.12.11.4 — `EXPLAIN PIPELINE` diff matches the issue exactly (`MergingAggregatedTransform` (new) vs. `GroupingAggregatedTransform` + `MergingAggregatedBucketTransform × N` (old)). Posted in the [issue thread](https://github.com/ClickHouse/ClickHouse/issues/104289#issuecomment-4414153221).
- ✅ Trigger isolation: dropping `LEFT ARRAY JOIN` under the new analyzer restores the bucket-aware plan, confirming `ARRAY JOIN` is the trigger.
- ✅ Test asserts `MergingAggregatedBucketTransform` appears in `EXPLAIN PIPELINE`. Verified the test returns `0` (would fail) on unmodified ClickHouse, so it does catch the bug.
- ⏳ I have **not** built a patched ClickHouse binary locally — relying on CI to confirm the test flips to `1` after the fix.

## Test

`tests/queries/0_stateless/04213_array_join_distributed_aggregation_memory_efficient.{sql,reference}` — distributed `LEFT ARRAY JOIN` query asserts that `MergingAggregatedBucketTransform` appears in `EXPLAIN PIPELINE`, plus a sanity check on the result.

/cc @LeeChaeRok @Algunenano


LeeChaeRok:
Thanks for putting this together. As the reporter of #104289, I can confirm this targets the exact plan-shape regression we hit in production.

This looks bug-like rather than a feature request: with the new analyzer enabled, the plan drops bucket-aware merging for this query shape and funnels all shard results into a single coordinator-side `MergingAggregatedTransform`, which leads to coordinator OOM. The only reliable workaround we have is disabling the new analyzer.

It looks like the PR has not received review/CI labeling yet, and the required statuses (`CH Inc sync` and `Mergeable Check`) are still waiting to be reported. Could a maintainer take a look when available and add the appropriate CI label, e.g. `can be tested`, so CI can start?

I’m happy to help validate a build or provide more production-side details if useful.

/cc @Algunenano

LeeChaeRok:
@KochetovNicolai this PR fixes a regression of the memory-bound / bucket-aware
merge path you introduced in 749d2e2 ("Fix 02404_memory_bound_merging with
analyzer."). When `LEFT ARRAY JOIN` adds a second entry to
`table_expression_node_to_data`, `addMergingAggregatedStep`'s `size() == 1` gate
is skipped, `is_remote_storage` stays false, and the coordinator falls back to a
single `MergingAggregatedTransform`. At our scale (150 shards) the coordinator
OOMs at ~119 GiB; per-shard memory stays at 2.5–3.2 GiB. The fix is a 3-line
generalization of that gate plus a regression test
(04213_array_join_distributed_aggregation_memory_efficient).

Could a maintainer kick off CI (add the can-be-tested label if it hasn't run
yet) and take a look? Happy to provide the full production EXPLAIN PIPELINE / repro on request. cc @novikd

LeeChaeRok:
@nahomar the failures are in the workflow-hook checks (pr_labels_and_category.py,
pr_body_check.py), not in the code — every real job was DROPPED, so nothing
actually ran. PR body is missing the required blocks from
.github/PULL_REQUEST_TEMPLATE.md. Can you add:

    ### Changelog entry (a user-readable short description of the changes
    that goes to CHANGELOG.md):
    Fix coordinator OOM on distributed queries with `LEFT ARRAY JOIN` over a
    remote source under the new analyzer.

Once that's in, pr-bugfix should auto-label and CI should actually run.

KochetovNicolai (approved):

LeeChaeRok:
@nahomar — any update on this?
Only thing left is the bot's test concern — pinning
`enable_memory_bound_merging_of_aggregation_results = 0` and
`optimize_aggregation_in_order = 0` should be enough to isolate the path.

The CI failures look like pre-existing flakiness, unrelated to the change.
If you're busy I can take it from here and push the test fix — just let me know.

PR #104513, opened 2026-05-10
Speed up the analyzer / planner for SELECT count() and similar lightweight queries

Algunenano:
- Performance Improvement

Reduce per-query overhead for simple SELECT queries (parsing, analysis and planning). For example, `SELECT count() FROM hits` from a single connection is roughly 50% faster.

Per-commit Q0 (`SELECT count() FROM hits`, single connection) on a local build:

| Commit | QPS | Δ |
|---|---:|---:|
| baseline | ~958 | — |
| Trim per-iteration overhead in sparse offset deserialization | ~958 | (Q1-targeted) |
| Reverted: ~Pack ProfileEvents counter into 8 bytes and gate trace check on a chain bit~ | ~995~ | ~+3.9%~ |
| Trim StorageSnapshot::getColumns: drop NameSet and Block round-trip | ~1019 | +2.4% |
| Defer ColumnNode map construction in initializeTableExpressionData | ~1216 | +19% |
| Defer column_names and column_identifier_first_parts construction | ~1352 | +11% |
| Memoize ColumnsDescription::get on the descriptor itself | ~1397 | +3% |
| Trim CurrentMemoryTracker hot path | ~1383 | within noise |
| Skip the recursive Planner after trivial-count optimization | ~1444 | +4.4% |


src/Common/ProfileEvents.h
```diff
@@ -144,24 +152,56 @@ namespace ProfileEvents
...
+
+        bool anyTraceInChain() const noexcept
+        {
+            return any_trace_in_chain.load(std::memory_order_relaxed);
         }
 
+    private:
+        /// Walk up the parent chain marking `any_trace_in_chain` so descendants attaching
+        /// to any of these ancestors will inherit the flag at attach time.
+        void markChainTracing()
```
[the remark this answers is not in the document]
Algunenano: Acknowledged — the race is real and we're consciously accepting it for the initiating query thread. Reasoning:

- The only call site for `setTraceProfileEvents` / `setTraceAllProfileEvents` is `ProcessList::insert` during query setup. Worker threads attach via `attachToGroupImpl` *after* `ProcessList::insert` returns, so all worker threads (where the hot increments happen) see a correct `any_trace_in_chain` via the chain-walking `inheritTracingFromParent`.
- The only victim is the initiating thread of the same query, for that single query, on its own (mostly cold) counters during the window between `attachToGroupImpl` and `ProcessList::insert`'s tracing toggle. Subsequent queries on the same thread are fine because `setUserCounters`/`setParent` re-walk the chain on each (re-)attach.
- The two alternatives — children-tracking on every `Counters` (≈24 extra bytes per instance plus a mutex around attach/detach) or a parent-chain fallback in the `increment` fast path — both regress the hot path in the common case where no tracing is configured anywhere. We picked the bounded-miss-on-initiating-thread tradeoff for the one query where tracing is being enabled.

This is now documented in the body of `inheritTracingFromParent` in `src/Common/ProfileEvents.cpp` so the next reader doesn't have to re-derive it from scratch.

src/Common/ProfileEvents.cpp
```diff
@@ -1551,6 +1552,78 @@ void Counters::setTraceProfileEvents(const String & events_list)
...
+    while (p)
+    {
+        if (p->any_trace_in_chain.load(std::memory_order_relaxed))
+        {
+            any_trace_in_chain.store(true, std::memory_order_relaxed);
+            return;
+        }
+        p = p->parent.load(std::memory_order_relaxed);
+    }
+}
```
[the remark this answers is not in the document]
Algunenano: I think I might end up removing this optimization. The improvement was minimal, only ~3% but mostly due to packing (which we can keep), and it's giving me too much throuble

src/Common/ProfileEvents.cpp
```diff
@@ -348,9 +348,9 @@
     \
     M(ExecuteShellCommand, "Number of shell command executions.", ValueType::Number) \
     \
-    M(ExternalProcessingCompressedBytesTotal, "Number of compressed bytes written by external processing (sorting/aggregating/joining)", ValueType::Bytes) \
-    M(ExternalProcessingUncompressedBytesTotal, "Amount of data (uncompressed, before compression) written by external processing (sorting/aggregating/joining)", ValueType::Bytes) \
-    M(ExternalProcessingFilesTotal, "Number of files used by external processing (sorting/aggregating/joining)", ValueType::Number) \
+    M(ExternalProcessingCompressedBytesTotal, "Number of compressed bytes written by external processing (sorting/aggragating/joining)", ValueType::Bytes) \
```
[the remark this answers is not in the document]
Algunenano: This PR leaves the comment as it was before as it reverted the changes. We can do it in another PR that modifies that file

src/Common/ProfileEvents.cpp
```diff
@@ -1203,14 +1203,14 @@ The server successfully detected this situation and will download merged part fr
...
+    M(SharedMergeTreeMergePartsMovedToOudated, "How many parts moved to oudated directory", ValueType::Number) \
     M(SharedMergeTreeMergePartsMovedToCondemned, "How many parts moved to condemned directory", ValueType::Number) \
     M(SharedMergeTreeOutdatedPartsConfirmationRequest, "How many ZooKeeper requests were used to config outdated parts", ValueType::Number) \
     M(SharedMergeTreeOutdatedPartsConfirmationInvocations, "How many invocations were made to confirm outdated parts", ValueType::Number) \
     M(SharedMergeTreeOutdatedPartsHTTPRequest, "How many HTTP requests were send to confirm outdated parts", ValueType::Number) \
     M(SharedMergeTreeOutdatedPartsHTTPResponse, "How many HTTP responses were send to confirm outdated parts", ValueType::Number) \
     M(SharedMergeTreeCondemnedPartsKillRequest, "How many ZooKeeper requests were used to remove condemned parts", ValueType::Number) \
-    M(SharedMergeTreeCondemnedPartsLockConflict, "How many times we failed to acquire lock because of conflict", ValueType::Number) \
+    M(SharedMergeTreeCondemnedPartsLockConflict, "How many times we failed to acquite lock because of conflict", ValueType::Number) \
```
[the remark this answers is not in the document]
Algunenano: This PR leaves the comment as it was before as it reverted the changes. We can do it in another PR that modifies that file

src/Common/ProfileEvents.cpp
```diff
@@ -1203,14 +1203,14 @@ The server successfully detected this situation and will download merged part fr
     M(SharedMergeTreeMergeMutationAssignmentFailedWithNothingToDo, "How many times we tried to assign merge or mutation and failed because nothing to merge", ValueType::Number) \
     M(SharedMergeTreeMergeMutationAssignmentFailedWithConflict, "How many times we tried to assign merge or mutation and failed because of conflict in Keeper", ValueType::Number) \
     M(SharedMergeTreeMergeMutationAssignmentSuccessful, "How many times we tried to assign merge or mutation", ValueType::Number) \
-    M(SharedMergeTreeMergePartsMovedToOudated, "How many parts moved to outdated directory", ValueType::Number) \
+    M(SharedMergeTreeMergePartsMovedToOudated, "How many parts moved to oudated directory", ValueType::Number) \
```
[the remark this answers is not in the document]
Algunenano: Same. This file is not changed in the PR anymore

Algunenano:
image

src/Common/ProfileEvents.cpp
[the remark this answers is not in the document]
Algunenano: Please don't report on unchanged code vs master

src/Common/ProfileEvents.cpp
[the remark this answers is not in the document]
Algunenano: Please don't report on unchanged code vs master

src/Common/ProfileEvents.cpp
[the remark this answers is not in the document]
Algunenano: Please don't report on unchanged code vs master

src/Analyzer/Resolve/IdentifierResolver.cpp
```diff
@@ -404,8 +404,10 @@ QueryTreeNodePtr IdentifierResolver::tryResolveIdentifierFromTableColumns(const
 
     const auto & identifier = identifier_lookup.identifier;
     auto identifier_full_name = identifier.getFullName();
-    auto it = scope.table_expression_data_for_alias_resolution->column_name_to_column_node.find(identifier_full_name);
-    if (it != scope.table_expression_data_for_alias_resolution->column_name_to_column_node.end())
+    scope.table_expression_data_for_alias_resolution->ensureColumnNodeMapIsPopulated();
+    const auto & node_map = *scope.table_expression_data_for_alias_resolution->column_name_to_column_node;
```
nickitat: Could we hide `column_name_to_column_node` and provide a method that will lazily populate it instead?

src/Analyzer/Resolve/TableExpressionData.h
```diff
@@ -39,18 +39,71 @@ struct AnalysisTableExpressionData
     bool should_qualify_columns = true;
     bool supports_subcolumns = false;
     NamesAndTypes column_names_and_types;
-    ColumnNameToColumnNodeMap column_name_to_column_node;
+    /// Set of regular (non-subcolumn) column names. Lazily populated by
+    /// `ensureColumnMembershipSetsArePopulated()`. Used for membership checks that don't need
+    /// a `ColumnNode` (e.g. `hasFullIdentifierName`). For wide tables (~100 columns) building
+    /// this set during `initializeTableExpressionData` is itself non-trivial; trivial queries
+    /// like `SELECT count() FROM t` never consult it.
+    mutable std::unordered_set<std::string, StringTransparentHash, std::equal_to<>> column_names;
```
nickitat, line 48: BTW, do you know why it is not `StringViewHash`?

src/Storages/VirtualColumnsDescription.cpp
```diff
@@ -104,4 +104,15 @@ ColumnsDescription VirtualColumnsDescription::toColumnsDescription(VirtualsKind
...
 }
 
+NamesAndTypesList VirtualColumnsDescription::getNamesAndTypes(VirtualsKind kind, VirtualsMaterializationPlace place) const
+{
+    NamesAndTypesList result;
+    for (const auto & desc : container)
+        if (static_cast<UInt8>(desc.kind) & static_cast<UInt8>(kind))
+            if (static_cast<UInt8>(desc.place) & static_cast<UInt8>(place))
+                result.emplace_back(desc.name, desc.type);
+
```
nickitat, line 113:
```suggestion
```

src/DataTypes/Serializations/SerializationSparse.cpp
```diff
@@ -103,9 +103,14 @@ size_t deserializeOffsets(
     if (max_rows_to_read == 0)
         return 0;
 
-    if (state.num_trailing_defaults >= max_rows_to_read)
+    /// Hoist state into locals so the inner loop avoids two redundant memory writes
+    /// per iteration in the common case where both fields stay 0/false.
+    size_t num_trailing_defaults = state.num_trailing_defaults;
```
nickitat, line 108: TBH, such optimisations don't make sense to me.

src/Common/CurrentMemoryTracker.h
```diff
@@ -19,4 +19,10 @@ struct CurrentMemoryTracker
 
 private:
     [[nodiscard]] static AllocationTrace allocImpl(Int64 size, bool throw_if_memory_exceeded);
+
+    /// Cold-path slow helpers used when there is no `current_thread` (i.e. before the
```
nickitat: Since these changes don't provide measurable improvement, let's avoid complicating the code.

src/Storages/VirtualColumnsDescription.h
```diff
@@ -55,6 +55,10 @@ class VirtualColumnsDescription
     Block getSampleBlock(VirtualsKind kind, VirtualsMaterializationPlace place) const;
     ColumnsDescription toColumnsDescription(VirtualsKind kind, VirtualsMaterializationPlace place) const;
 
+    /// Like `getSampleBlock`, but skips the per-column `IDataType::createColumn()` allocation
+    /// and the `Block` round-trip. Hot in `StorageSnapshot::getColumns`.
+    NamesAndTypesList getNamesAndTypes(VirtualsKind kind, VirtualsMaterializationPlace place) const;
```
rschu1ze, line 60: I think it's worth making a follow-up PR that converts as many places as possible from `getSampleBlock` to `getNamesAndTypes`. The latter is obviously faster and more idiomatic.

src/Storages/ColumnsDescription.h
```diff
@@ -280,6 +288,26 @@ class ColumnsDescription : public IHints<>
     void removeSubcolumns(const String & name_in_storage);
 
     std::optional<NameAndTypePair> tryGetDynamicSubcolumn(const String & column_name, const GetColumnsOptions & options) const;
+
+    /// `get(options)` is called repeatedly with the same options across analyzer
+    /// and planner of every query, and rebuilding the result iterates the columns
+    /// multi-index plus runs `addSubcolumnsToList` on every row. The cache lives on
+    /// `ColumnsDescription` (which is owned by an immutable `StorageInMemoryMetadata`
+    /// snapshot) so the result is reused across queries on the same metadata version.
+    /// Any mutator that alters `columns` or `subcolumns` calls `invalidateGetCache`.
```
rschu1ze: Any method that alters ...

src/Storages/ColumnsDescription.cpp
```diff
@@ -255,6 +255,43 @@ void ColumnDescription::readText(ReadBuffer & buf)
...
+    return *this;
+}
+
+ColumnsDescription & ColumnsDescription::operator=(ColumnsDescription && other) noexcept
+{
+    if (this != &other)
+    {
+        columns = std::move(other.columns);
+        subcolumns = std::move(other.subcolumns);
+        invalidateGetCache();
```
rschu1ze: Moving from other means other is invalid from now on.
Why can't we reuse the cache of other in the moved-to object then?

src/Storages/ColumnsDescription.h
```diff
@@ -280,6 +288,26 @@ class ColumnsDescription : public IHints<>
...
+        UInt8 virtuals_kind;
+        UInt8 virtuals_place;
+        UInt8 flags; /// bit 0: with_subcolumns, bit 1: with_dynamic_subcolumns
+        bool operator==(const GetCacheKey & other) const = default;
+    };
+    static GetCacheKey makeGetCacheKey(const GetColumnsOptions & options);
+    void invalidateGetCache() const;
+
+    mutable std::mutex get_cache_mutex;
+    mutable std::vector<std::pair<GetCacheKey, std::shared_ptr<const NamesAndTypesList>>> get_cache;
```
rschu1ze, line 316: `std::unordered_map`?

src/Storages/ColumnsDescription.h
```diff
@@ -280,6 +288,26 @@ class ColumnsDescription : public IHints<>
     void removeSubcolumns(const String & name_in_storage);
 
     std::optional<NameAndTypePair> tryGetDynamicSubcolumn(const String & column_name, const GetColumnsOptions & options) const;
+
+    /// `get(options)` is called repeatedly with the same options across analyzer
```
rschu1ze:
```cpp
/// `get(const ColumnOptions &)`
```

please

src/Storages/ColumnsDescription.h
```diff
@@ -280,6 +288,26 @@ class ColumnsDescription : public IHints<>
...
+        UInt8 kind;
+        UInt8 virtuals_kind;
+        UInt8 virtuals_place;
+        UInt8 flags; /// bit 0: with_subcolumns, bit 1: with_dynamic_subcolumns
+        bool operator==(const GetCacheKey & other) const = default;
+    };
+    static GetCacheKey makeGetCacheKey(const GetColumnsOptions & options);
+    void invalidateGetCache() const;
+
+    mutable std::mutex get_cache_mutex;
```
rschu1ze: Can this be a `std::shared_mutex` to allow concurrent reads under a shared lock?

src/Storages/ColumnsDescription.cpp
```diff
@@ -567,6 +610,22 @@ void ColumnsDescription::addSubcolumnsToList(NamesAndTypesList & source_list) co
 
 NamesAndTypesList ColumnsDescription::get(const GetColumnsOptions & options) const
 {
+    /// `get(options)` is hot: analyzer + planner together call it 2-3 times per query
+    /// with the same options, and rebuilding the list iterates the columns multi-index
+    /// and runs `addSubcolumnsToList` (linear hash lookups per row). Memoize on
+    /// `ColumnsDescription` so the result is shared across queries on the same metadata
+    /// version. The cache is invalidated by every mutator that touches `columns` or
```
rschu1ze: [...] by every method that [...]

src/Storages/ColumnsDescription.h
```diff
@@ -127,6 +128,13 @@ class ColumnsDescription : public IHints<>
 public:
     ColumnsDescription() = default;
 
+    /// Mutable cache members (`get_cache_mutex`, `get_cache`) are non-copyable; define
+    /// explicit copy / move so a copied / moved instance starts with an empty cache.
+    ColumnsDescription(const ColumnsDescription & other) : IHints<>(other), columns(other.columns), subcolumns(other.subcolumns) {}
+    ColumnsDescription(ColumnsDescription && other) noexcept : columns(std::move(other.columns)), subcolumns(std::move(other.subcolumns)) {}
```
rschu1ze: See my comment in the move assignment op. Here (move assignment ctor), we somehow don't invalidate the cache - I think that's the right way to do it.

src/DataTypes/Serializations/SerializationSparse.cpp
rschu1ze: Agree with ^^

src/Analyzer/Resolve/TableExpressionData.h
```diff
@@ -39,18 +39,71 @@ struct AnalysisTableExpressionData
...
+    /// `name -> ColumnNode`. Lazily populated by `ensureColumnNodeMapIsPopulated()`. Many
+    /// queries (e.g. `SELECT count() FROM t`) never resolve any column identifier from a
+    /// table and therefore never need this map; building 100+ `ColumnNode`s up front for
+    /// such queries is the dominant cost of `initializeTableExpressionData` for wide tables.
+    /// `nullopt` means "not populated yet" — `has_value()` doubles as the populated flag.
```
rschu1ze: Not clear what `has_value()` doubles as the populated flag.` means

src/Analyzer/Resolve/TableExpressionData.h
```diff
@@ -39,18 +39,71 @@ struct AnalysisTableExpressionData
...
+    /// of a compound identifier could refer to a column in this table. Populated together
+    /// with `column_names` by `ensureColumnMembershipSetsArePopulated()`.
+    mutable std::unordered_set<std::string, StringTransparentHash, std::equal_to<>> column_identifier_first_parts;
+    mutable bool column_membership_sets_populated = false;
+
+    void ensureColumnMembershipSetsArePopulated() const
+    {
+        if (column_membership_sets_populated)
+            return;
+        column_membership_sets_populated = true;
```
rschu1ze: Lets make this "transactional" and set the status flag only _after_  the work has been done (l. 71-78).

Also, if TableExpressionData is used concurrently (don't know), then this needs to be an atomic bool.

src/DataTypes/Serializations/SerializationSparse.cpp
Algunenano: This was pure assembly analysis.

From the session with Claude:

```
Found the full picture. Summary of the assembly analysis from the May 7 session:

Baseline (tmp/clickhouse_baseline)

analyze-assembly.py on readOrGetCachedSparseOffsets (which has deserializeOffsets inlined into it):

- Spill density: 19.1% [high] — flagged finding: "Spills inside loop bodies multiply memory traffic by iteration count."
- The smoking gun in the inner loop body (the non-end_of_granule path):

0x1656b576: xor ecx, ecx
0x1656b578: mov byte ptr [r14 + 0x18], 0x0   ; state.has_value_after_defaults = false
0x1656b57d: mov qword ptr [r14 + 0x10], 0x0  ; state.num_trailing_defaults  = 0
0x1656b585: inc r15
0x1656b588: mov r14, qword ptr [rbp + 0x18]  ; reload &state
0x1656b58c: mov qword ptr [rbp - 0x158], r13 ; SPILL
0x1656b593: mov qword ptr [rbp - 0x150], rcx ; SPILL

r14 is the pointer to state. Those two stores at +0x10 and +0x18 are literally the per-iteration state.num_trailing_defaults = 0; state.has_value_after_defaults = false; lines (184–185 in the baseline source). The reload of r14 from [rbp+0x18] shows the compiler couldn't even keep &state in a register across the loop body.

Root cause

DeserializeStateSparse & state is a reference parameter. The compiler can't prove writes through it don't alias the PODArray::push_back, the ReadBuffer reads in readVarUInt, or ++skipped_values_rows. So every source-level assignment to state.foo materializes as a store, and the address itself gets reloaded.

After hoisting

tmp/clickhouse_optimized:
- Inner loop no longer touches [r14+0x10]/[r14+0x18].
- The single pair of writes happens once at function exit:
0x1656b99e: mov qword ptr [rax + 0x10], rcx     ; state.num_trailing_defaults
0x1656b9a2: setae byte ptr [rax + 0x18]         ; state.has_value_after_defaults
```

src/Analyzer/Resolve/IdentifierResolver.cpp
Algunenano: Yes, makes sense and it's cleaner. Will do

src/Common/CurrentMemoryTracker.h
Algunenano: I'll revert it from this PR. At some point I'll go back since the tracker constantly shows up in the profile (and sometimes it's even slower than the allocation itself), but it's better to reduce noise here.

src/Analyzer/Resolve/TableExpressionData.h
Algunenano: Because StringViewHash is not transparent, which enables lookups without constructing a temporary std::string. We have multiple places in the code (search for `using is_transparent`). I'll add a followup to try to unify them all by making `StringViewHash` transparent and using it everywhere (if possible)

src/Storages/VirtualColumnsDescription.h
Algunenano: I'll add myself a note to follow up

src/Analyzer/Resolve/TableExpressionData.h
nickitat: Pls also change the name ) "transparent" IMO contains zero useful hints

src/Storages/ColumnsDescription.cpp
Algunenano: You are right. We can transfer it too. Will do

src/DataTypes/Serializations/SerializationSparse.cpp
nickitat: ok, but this analysis doesn't tell us
1. whether these couple of instructions were on the critical path - otherwise, we just saved a bit of throughput and no latency
2. even if they were, how does it compare to the total amount of cycles each iteration takes 

But anyway, I just express my opinion and don't insist on it.

src/Storages/ColumnsDescription.h
Algunenano: It was a deliberate choice. It's tiny and has only a handful of entries at most, so having and iterating a small vector is faster. I'll add a comment

src/Analyzer/Resolve/TableExpressionData.h
Algunenano: I'll remove the comments as part of hiding the implementation and using `getColumnNodeMap`. It was a convoluted comment to say "it's initialized, even if it's empty" because in some place we called `emplace` to work on it, but it doesn't make much sense now

src/Analyzer/Resolve/TableExpressionData.h
Algunenano: Will move it. AFAICT the analyzer runs in a single thread only (a single QueryAnalyzer per query, running in a thread) and all other members aren't atomic, so atomicity is fine

src/Storages/ColumnsDescription.h
Algunenano: Yes, I'll change it to SharedMutex

src/DataTypes/Serializations/SerializationSparse.cpp
Algunenano: > But anyway, I just express my opinion and don't insist on it.

Don't worry, it's good to push back.

> whether these couple of instructions were on the critical path - otherwise, we just saved a bit of throughput and no latency

This loop to deserialize sparse columns is often one of the most, if not the most, hot function in small queries like `SELECT count() FROM hits WHERE AdvEngineID != 0`.

> even if they were, how does it compare to the total amount of cycles each iteration takes

```
Comparing the steady-state non-end-of-granule path (the dominant case):

no_hoist hot-iteration sequence (0x16a789a3..0x16a789c2):
add  rcx, r13                          ; compute push value
mov  qword ptr [rax], rcx              ; offsets.push_back
add  rax, 0x8
mov  qword ptr [rdx + 0x18], rax       ; update offsets.size_
mov  byte  ptr [rbp + 0x18], 0x0       ; state.has_value_after_defaults = false   <-- store
mov  qword ptr [rbp + 0x10], 0x0       ; state.num_trailing_defaults  = 0         <-- store
inc  r13                               ; total_rows++
mov  qword ptr [rsp + 0x38], rax       ; spill rax (offsets.end_)

hoist hot-iteration sequence (0x16a789db..0x16a789f4):
add  rcx, r12                          ; compute push value
mov  qword ptr [rax], rcx              ; offsets.push_back
add  rax, 0x8
mov  qword ptr [rdx + 0x18], rax       ; update offsets.size_
mov  r14, r12
inc  r14                               ; total_rows++
mov  qword ptr [rsp + 0x58], rax       ; spill rax (offsets.end_)

Delta on this iteration of the hot loop:
- −2 store instructions (state.has_value_after_defaults = false, state.num_trailing_defaults = 0 gone)
- Iteration instruction count: 9 → 7

Whole-function counts also match: no_hoist has 7 zero-stores to state.*; hoist has 3 (the writebacks at function exit + the end-of-granule reset).

Cycle interpretation on Zen3: each removed store is ~1 µop dispatched to the store-data and store-address ports; under store-buffer pressure they cost roughly 1 cycle each, but the loop is more likely IPC-bound by the VarInt decode and PODArray-append dependency chain. So the saving is ~2 µops/iter, theoretically ~2 cycles, but in practice closer to <1 cycle of wall — consistent with the ~1% benchmark improvement.
```

src/Storages/ColumnsDescription.h
Algunenano: Decided to test it and both with concurrency 1 and concurrency 16 and the difference is slightly better towards vector but it's too small to trust, as is within noise.

nickitat (approved):

src/DataTypes/Serializations/SerializationSparse.cpp
nickitat: To clarify what I was asking: my question was specifically whether the saved instructions sit on the critical path, not whether the loop is hot. Those are different things. It very well might be that we saved instructions, but we didn't change the number of cycles per loop iteration. Because as long as there are idle ports and buffers, you can add a few instructions here and there, essentially for free. So, if you have a full asm for this loop, you can pipe it into a tool like https://uica.uops.info and see whether the cycles-per-iteration metric actually decreased. It'd be quite interesting to know.

Algunenano: Assembly (generated per file, not from LTO build):
* Before: https://pastila.nl/?01d3fbfd/fa38d2fa0aa9e8fd20fdcdc7c9300b4d#OTvoSCza+2GAgZYigrgV8Q==GCM
* After: https://pastila.nl/?000d7e52/ef8fe8c022f76e8ee53b769cc9b531c5#vPIIa05X6eGJa/Z3JCzatw==GCM


Predicted Throughput: (Generated urls are too big)
* Before:
```
Tool | Broadwell | Skylake | Coffee Lake
-- | -- | -- | --
uiCA | 31.00 | 42.00 | 42.00
```

* After:
```
Tool | Broadwell | Skylake | Coffee Lake
-- | -- | -- | --
uiCA | 28.00 | 33.00 | 33.00
```

src/DataTypes/Serializations/SerializationSparse.cpp
```diff
@@ -203,20 +209,37 @@ size_t readOrGetCachedSparseOffsets(
     size_t & read_rows,
     size_t & skipped_values_rows)
 {
+    /// `getSubcolumnNameForStream(path + [SparseOffsets], encode_sparse_stream=true)` equals
+    /// `getSubcolumnNameForStream(path, true) + ".sparse.idx"` (or "sparse.idx" if path is empty),
+    /// so we can compute the cache key without mutating settings.path.
+    /// This lets us skip the heavy `Substream` construction in `path.push_back` on cache hits,
+    /// and reuse the same key for `cache->insert_or_assign` on cache misses.
```
Avogar, line 216: This looks very bad to me. It replaces common usage of the cache `ISerialization::getElementFromSubstreamsCache/ISerialization::addElementToSubstreamsCache` to some custom logic that can be easily become broken if common logic of cache usage is changed in those methods that are used in all other places.

> the heavy `Substream` construction in `path.push_back` on cache hits

What is so heavy in construction of this struct? Per granule it should be negligable.

PR #94104, opened 2026-01-13
Iceberg compaction split PR part 2 : Merge tree mutator refactoring

scanhex12:
### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Continue of Iceberg compaction

PR #104518, opened 2026-05-10
Sync private

scanhex12:

### Version info
- Merged into: `26.5.1.548`


src/Storages/MergeTree/MergeTreeDataMergerMutator.cpp
```diff
@@ -153,68 +142,7 @@ std::unordered_map<String, PartsRanges> combineByPartitions(PartsRanges && range
     return ranges_by_partitions;
 }
 
-String getBestPartitionToOptimizeEntire(
```
Michicosun, line 156: I remember here https://github.com/ClickHouse/ClickHouse/pull/94104 we discussed and concluded that changes to the merge mutator were not needed, at least from the MergeTree side. What changed?

PR #104520, opened 2026-05-10
Reject collection literals as data-type arguments (STID 1941-1bfa)

groeneai:
## Motivation

`AST fuzzer (amd_debug)` and related sanitizer/fuzzer jobs hit a chronic `LOGICAL_ERROR: Inconsistent AST formatting` (STID `1941-1bfa`) in `executeQuery.cpp:1280` whenever a fuzzed query puts a collection literal (`(NULL)`, `[NULL]`, `{a: 1}`) inside a parameterized data type's argument list.

Latest hit: PR [#100177](https://github.com/ClickHouse/ClickHouse/pull/100177) (commit `9322e66`):

```sql
CREATE TABLE t0__fuzz_31 (`a` Nullable(multiply((NULL), Int8)), ...) ENGINE = ReplacingMergeTree ...
```

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100177&sha=9322e66151ea6cc3017cc476e1f8d958bb56421c&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%29

The bug has been hit by 20+ unrelated PRs over the last 30 days. Master is unaffected (the round-trip check fires only in DEBUG/sanitizer builds, and master CI doesn't usually run AST fuzzer mutations against the same seed). PR [#102547](https://github.com/ClickHouse/ClickHouse/pull/102547) by @zoomxi (merged 2026-04-29) fixed one specific variant (`INSERT with SAMPLE and OFFSET`); this PR addresses the broader collection-literal variant.

## Root cause

Inside `ParserDataType::parseImpl` (`src/Parsers/ParserDataType.cpp`), arguments to a parameterized data type were parsed with `ParserAllCollectionsOfLiterals`, which accepts `(...)`, `[...]`, `{...}` (with `false` for the map case). For `multiply((NULL), Int8)` this gives `DataType_multiply -> [Literal_Tuple_(NULL), DataType_Int8]`.

Formatting that AST goes through `FieldVisitorToString::operator()(const Tuple &)`, which prints a single-element Tuple literal as `tuple(NULL)` (the explicit function form, used to avoid `(NULL)` re-parsing as a parenthesized expression in regular SQL contexts).

Re-parsing `multiply(tuple(NULL), Int8)` in data-type-argument context: `tuple(NULL)` starts with an identifier (not `(`/`[`/`{`), so `ParserAllCollectionsOfLiterals` doesn't trigger. Falls through to `ParserDataType`, which parses it as `ASTDataType("tuple")` with a `Literal_NULL` child. Different shape → `getTreeHash` differs → `LOGICAL_ERROR`.

The same goes for `multiply([NULL], Int8)` (`Literal_Array_(NULL)` ↔ `[NULL]` re-parsed as a Literal_Array... but the formatter for empty arrays / various edge cases also drifts). I confirmed the exact same crash for both `(NULL)` and `[NULL]` variants locally.

## Fix

No real data type takes a collection literal as a parameter:
- Simple literals: `FixedString(5)`, `Decimal(10, 2)`, `DateTime64(3, 'UTC')`
- Nested types: `Nullable(Int32)`, `Array(String)`, `LowCardinality(...)`
- Special handling: `Tuple(Int32, String)`, `Enum8('a' = 1, 'b' = 2)` (separate code paths in the same function)

Replace `ParserAllCollectionsOfLiterals literal_parser(false)` with `ParserLiteral literal_parser`. Malformed queries are now rejected at parse time with `SYNTAX_ERROR` (Code 62) instead of crashing later in the round-trip check.

Verified locally:
- `multiply((NULL), Int8)`, `multiply([NULL], Int8)`, `Nullable(multiply((NULL), Int8))` → `SYNTAX_ERROR` (was: `LOGICAL_ERROR` crash)
- Sanity-tested 13 common type families (`UInt32`, `Decimal(10,2)`, `DateTime64(3,'UTC')`, `FixedString(5)`, `Nullable(String)`, `Tuple(...)`, `Array(Tuple(...))`, `Enum8(...)`, `LowCardinality(Nullable(FixedString(10)))`, `Map(String, Array(Int32))`, `SimpleAggregateFunction(sum, UInt64)`, `AggregateFunction(uniq, Int32)`, `Variant(...)`) — all parse correctly.

## Test changes

`tests/queries/0_stateless/03210_inconsistent_formatting_of_data_types.sh` had one case `Array((1), UInt8)` in the "possibly incorrect but should round-trip" section that relied on the old behavior. Move it to the "invalid queries that don't parse" section alongside three new regression cases for STID `1941-1bfa`. The 4 tracked-as-valid cases continue to round-trip unchanged.

...

### Version info
- Merged into: `26.5.1.499`

### DIFF
--- a/src/Parsers/ParserDataType.cpp
+++ b/src/Parsers/ParserDataType.cpp
@@ -508,7 +508,20 @@ bool ParserDataType::parseImpl(Pos & pos, ASTPtr & node, Expected & expected)
         else
         {
             ParserDataType data_type_parser;
-            ParserAllCollectionsOfLiterals literal_parser(false);
+            /// Only accept simple literals (numbers, strings, NULL, ...) as
+            /// data-type arguments. We deliberately do NOT accept collection
+            /// literals like `(1)`, `[1, 2]` or `{a: 1}` here: no real data
+            /// type takes a tuple/array/map literal as an argument, and
+            /// accepting them produces an `ASTLiteral` with `Field::Tuple`
+            /// (or `Field::Array`) inside the type's argument list. The
+            /// formatter then prints such a Tuple literal as `tuple(...)`
+            /// (the explicit function form -- see
+            /// `FieldVisitorToString::operator()(const Tuple &)`), and
+            /// re-parsing `tuple(...)` in this context yields an
+            /// `ASTDataType("tuple")` instead of an `ASTLiteral`, breaking
+            /// the AST round-trip check in `executeQuery` (DEBUG/sanitizer
+            /// builds). See STID 1941-1bfa.
+            ParserLiteral literal_parser;
 
             const char * operators[] = {"=", "equals", nullptr};
             ParserLeftAssociativeBinaryOperatorList enum_parser(operators, std::make_unique<ParserLiteral>());
--- a/tests/queries/0_stateless/03210_inconsistent_formatting_of_data_types.reference
+++ b/tests/queries/0_stateless/03210_inconsistent_formatting_of_data_types.reference
@@ -1,7 +1,10 @@
 ALTER TABLE columns_with_multiple_streams (MODIFY COLUMN `field1` Nullable(tupleElement(x, 2), UInt8))
 ALTER TABLE t_update_empty_nested (ADD COLUMN `nested.arr2` Array(tuple('- ON NULL -', toLowCardinality(11), 11, 11, toLowCardinality(11), 11), UInt64))
-ALTER TABLE t (ADD COLUMN `x` Array(tuple(1), UInt8))
 ALTER TABLE enum_alter_issue (MODIFY COLUMN `a` Enum8(equals('one', timeSlots(timeSlots(arrayEnumerateDense(tuple('0.2147483646', toLowCardinality(toUInt128)), NULL), 4, 12.34, materialize(73), 2)), 1)))
 ALTER TABLE t_sparse_mutations_3 (MODIFY COLUMN `s` Tuple(Nullable(tupleElement(s, 1), UInt64), Nullable(UInt64), Nullable(UInt64), Nullable(UInt64), Nullable(String)))
+Syntax error
+Syntax error
+Syntax error
+Syntax error
 Syntax error
 Syntax error
--- a/tests/queries/0_stateless/03210_inconsistent_formatting_of_data_types.sh
+++ b/tests/queries/0_stateless/03210_inconsistent_formatting_of_data_types.sh
@@ -7,10 +7,19 @@ CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 # Ensure that these (possibly incorrect) queries can at least be parsed back after formatting.
 $CLICKHOUSE_FORMAT --oneline --query "ALTER TABLE columns_with_multiple_streams MODIFY COLUMN field1 Nullable(tupleElement(x, 2), UInt8)" | $CLICKHOUSE_FORMAT --oneline
 $CLICKHOUSE_FORMAT --oneline --query "ALTER TABLE t_update_empty_nested ADD COLUMN \`nested.arr2\` Array(tuple('- ON NULL -', toLowCardinality(11), 11, 11, toLowCardinality(11), 11), UInt64)" | $CLICKHOUSE_FORMAT --oneline
-$CLICKHOUSE_FORMAT --oneline --query "ALTER TABLE t ADD COLUMN x Array((1), UInt8)" | $CLICKHOUSE_FORMAT --oneline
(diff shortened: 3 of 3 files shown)


groeneai:
## Pre-PR Validation Gate

Per TASK.md Phase 4 step 9 — answering the five mandatory questions:

**a) Deterministic repro?** ✅ Yes. The minimal repro is one statement:
```
build/programs/clickhouse local --query "CREATE TABLE t (a multiply((NULL), Int8)) ENGINE = Memory"
```
Without the fix this aborts with `LOGICAL_ERROR: Inconsistent AST formatting (STID: 1941-1bfa)`. With the fix it returns `SYNTAX_ERROR (Code 62)` immediately.

**b) Root cause explained?** ✅ Yes. The parser uses `ParserAllCollectionsOfLiterals` to parse data-type arguments, which accepts `(NULL)` as a `Literal_Tuple_(NULL)` (`ASTLiteral` with `Field::Tuple([NULL])`). The formatter then prints that single-element Tuple literal as `tuple(NULL)` (the explicit function form, see comment in `FieldVisitorToString::operator()(const Tuple &)`). Re-parsing `multiply(tuple(NULL), Int8)` in data-type-argument context produces an `ASTDataType("tuple")` because `tuple(NULL)` starts with an identifier (not `(`/`[`/`{`), so the literal collection parser doesn't trigger and the data-type parser handles it as a parameterized DataType. Same input text → different AST shape after one round-trip → `getTreeHash` differs → `LOGICAL_ERROR`.

**c) Fix matches root cause?** ✅ Yes. The fix is at the source of the divergence: data-type arguments no longer accept collection literals at all. No real data type takes a tuple/array/map literal as a parameter — every parameterized type uses simple literals (`FixedString(5)`, `Decimal(10, 2)`, `DateTime64(3, 'UTC')`), nested types (`Nullable(Int32)`, `Array(String)`), or has its own special handling (`Tuple(Int32, String)`, `Enum8('a' = 1)`). Rejecting them upfront catches the malformed query at parse time with `SYNTAX_ERROR`, eliminating the round-trip mismatch.

**d) Test intent preserved? / New tests added?** ✅ Yes. `03210_inconsistent_formatting_of_data_types` had one case (`Array((1), UInt8)`) in the "possibly incorrect but should round-trip" section that depended on the old behavior. It's moved to the "invalid queries that don't parse" section together with three new regression cases (`multiply((NULL), Int8)`, `multiply([NULL], Int8)`, `Nullable(multiply((NULL), Int8))`) covering the STID `1941-1bfa` variants reported by the AST fuzzer in CI. The 4 valid round-trip cases in the test continue unchanged.

**e) Demonstrated in both directions?** ✅ Yes. With the master HEAD binary, all three repro queries crash with `LOGICAL_ERROR`. With the fix applied, all three return `SYNTAX_ERROR (Code 62)` cleanly. Sanity-tested 13 common type families (`UInt32`, `Decimal(10, 2)`, `DateTime64(3, 'UTC')`, `FixedString(5)`, `Nullable(String)`, `Tuple(Int32, String, Array(Int32))`, `Array(Tuple(Int32, String))`, `Enum8('a' = 1, 'b' = 2, 'c' = 3)`, `LowCardinality(Nullable(FixedString(10)))`, `Map(String, Array(Int32))`, `SimpleAggregateFunction(sum, UInt64)`, `AggregateFunction(uniq, String)`, `Variant(...)`) — all still parse correctly. Test `03210_inconsistent_formatting_of_data_types` passes locally with the updated reference.

**f) Fix is general, not a narrow patch?** ✅ Yes. This is the universal fix for the STID `1941-1bfa` collection-literal variant. Any combination of unknown DataType name (`multiply`, `divide`, etc. — the data-type-name parser is permissive by design) wrapping any collection literal (`(...)`, `[...]`) at any nesting depth (`Nullable(unknown(...))`, `Array(unknown(...))`, etc.) was producing the same class of AST round-trip mismatch. After the fix, all such constructs are rejected at parse time. The only narrower fix considered (changing `FieldVisitorToString` to output `(NULL,)` with trailing comma instead of `tuple(NULL)`) would have updated 25+ reference test files and risked behavior drift in expression contexts; the parser-level fix is both simpler and more semantically correct (data types should never have collection literal parameters).

Session: `cron:clickhouse-ci-task-worker:20260510-124500`

groeneai:
cc @evillique @alexey-milovidov — could you review this? It fixes the chronic AST fuzzer trunk bug STID 1941-1bfa for the collection-literal-in-data-type variant (fuzzer mutates a column type to `multiply((NULL), Int8)`; parser produces `Literal_Tuple_(NULL)` inside the type's argument list; formatter prints it as `tuple(NULL)`; re-parsing gives `DataType_tuple`; AST round-trip mismatch crashes `executeQuery` in DEBUG/sanitizer builds). The fix rejects collection literals as data-type arguments at parse time — no real data type uses them as parameters.

alexey-milovidov (approved):

PR #104529, opened 2026-05-10
Fix uniqThetaIntersect with empty state on the right-hand side

### LINKED ISSUE #91595: uniqThetaIntersect does not works for empty merge states
### Company or project name

Summarization

### Describe what's wrong

I am trying to merge the theta sketches based on some predicate, If the predicate results in no sketch being selected then intersect returns the count from first sketch, wherein the result should be 0.

Fiddle [Link](https://fiddle.clickhouse.com/090ba1db-65cc-4425-b3f4-4939579990a7) 

Yes

### How to reproduce

```sql
CREATE TABLE TEST_AGG(
    CID UInt32,
    USERS AggregateFunction(uniqTheta, UInt32)
) ENGINE = AggregatingMergeTree() ORDER BY CID;

INSERT INTO TEST_AGG VALUES (1, arrayReduce('uniqThetaState', [1, 2]::Array(UInt32))), (2, arrayReduce('uniqThetaState', [2, 3]::Array(UInt32))); 

SELECT finalizeAggregation(
    uniqThetaIntersect(
        uniqThetaMergeStateIf(USERS, CID = 1),
        uniqThetaMergeStateIf(USERS, CID = 3)
    )
) FROM TEST_AGG;
```

### Expected behavior

The final select query should return 0 as the result, because there was an empty state to intersect with it.

_No response_

### Additional context

_No response_

groeneai:
`uniqThetaIntersect(a, b)` returned the cardinality of `a` whenever `b` was a freshly-created `uniqTheta` state that had never been updated -- e.g. the result of `uniqThetaMergeStateIf(s, predicate)` when the predicate excluded every row.

Closes #91595.

### Root cause

In `ThetaSketchData::intersect` we feed the local data into a `datasketches::theta_intersection`, then update it with whatever sketch `rhs` carries (`rhs.sk_update` or `rhs.sk_union`). When `rhs` had neither pointer allocated, we never called `update` for `rhs` and the intersection ended up with a single input -- `theta_intersection::get_result` returns that single input unchanged, so the wrong, non-empty result was used.

Reproducer (from #91595, simplified by @den-crane):

```sql
SELECT finalizeAggregation(
    uniqThetaIntersect(
        uniqThetaMergeStateIf(arrayReduce('uniqThetaState', [1, 2]::Array(UInt32)), 1),
        uniqThetaMergeStateIf(arrayReduce('uniqThetaState', []::Array(UInt32)), 0)
    )
);
-- Before: 2
-- After:  0
```

### Fix

Intersection with an empty set is always empty, so short-circuit in `ThetaSketchData::intersect` before calling into DataSketches: when `rhs` has neither `sk_update` nor `sk_union` allocated, drop our own state and return. `merge` (union with empty = self) and `aNotB` (`A \ empty = A`) already behave correctly for an empty rhs and need no change.

### Tests

Added `04215_91595_uniq_theta_intersect_empty_state.sql` covering:
- the den-crane reproducer
- `intersect(non_empty, filter-excluded empty)`
- `intersect(filter-excluded empty, non_empty)` -- commuted
- `intersect(filter-excluded empty, filter-excluded empty)`
- a sanity check that non-empty intersections still work
- the `AggregatingMergeTree` variant from the original issue

The pre-existing `01798_uniq_theta_union_intersect_not.sql` still passes locally with the fix applied.

Fix `uniqThetaIntersect` returning the cardinality of the first argument instead of `0` when the second argument is an empty `uniqTheta` state -- for example the result of `uniqThetaMergeStateIf(s, predicate)` when the predicate excludes every row.

### Version info
- Merged into: `26.5.1.505`

### DIFF
--- a/src/AggregateFunctions/ThetaSketchData.h
+++ b/src/AggregateFunctions/ThetaSketchData.h
@@ -98,6 +98,21 @@ public:
 
     void intersect(const ThetaSketchData & rhs)
     {
+        /// If `rhs` has no recorded values it represents an empty set, and the
+        /// intersection with an empty set is always empty. Without this guard
+        /// `theta_intersection` would only receive `this` as a single input and
+        /// `get_result` would return that input unchanged — yielding a wrong,
+        /// non-empty result. This matters for example after
+        /// `uniqThetaMergeStateIf(state, predicate)` when the predicate excludes
+        /// every row: the resulting state is freshly created and has neither
+        /// `sk_update` nor `sk_union` allocated.
+        if (!rhs.sk_update && !rhs.sk_union)
+        {
+            sk_update.reset(nullptr);
+            sk_union.reset(nullptr);
+            return;
+        }
+
         datasketches::theta_union * u = getSkUnion();
 
         if (sk_update)
--- a/tests/queries/0_stateless/04215_91595_uniq_theta_intersect_empty_state.reference
+++ b/tests/queries/0_stateless/04215_91595_uniq_theta_intersect_empty_state.reference
@@ -1,0 +1,12 @@
+reproducer from den-crane
+0
+intersect non_empty with filter-excluded empty state
+0
+intersect filter-excluded empty state with non_empty (commuted)
+0
+intersect two filter-excluded empty states
+0
+sanity check - non-empty intersection still works
+5
+AggregatingMergeTree variant from the issue
+0
--- a/tests/queries/0_stateless/04215_91595_uniq_theta_intersect_empty_state.sql
+++ b/tests/queries/0_stateless/04215_91595_uniq_theta_intersect_empty_state.sql
@@ -1,0 +1,75 @@
+-- Tags: no-fasttest
+-- ^ no-fasttest because uniqTheta requires the DataSketches library
+
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/91595
+-- `uniqThetaIntersect` previously returned the cardinality of the first
+-- argument when the second argument was a freshly-created (never-updated)
+-- state. Such states arise naturally from
+-- `uniqThetaMergeStateIf(s, predicate)` when the predicate excludes every
+-- row, and the intersection with an empty set must always be empty.
+
+SELECT 'reproducer from den-crane';
+SELECT finalizeAggregation(
+    uniqThetaIntersect(
+        uniqThetaMergeStateIf(arrayReduce('uniqThetaState', [1, 2]::Array(UInt32)), 1),
+        uniqThetaMergeStateIf(arrayReduce('uniqThetaState', []::Array(UInt32)), 0)
+    )
+);
+
+SELECT 'intersect non_empty with filter-excluded empty state';
+SELECT finalizeAggregation(
+    uniqThetaIntersect(
+        uniqThetaStateIf(number, 1),
+        uniqThetaStateIf(number, 0)
+    )
+)
+FROM numbers(10);
+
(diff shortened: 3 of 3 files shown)


groeneai:
**Pre-PR validation gate (TASK.md Phase 4 Step 9):**

a) **Deterministic repro?** Yes. Without the fix:
```sql
SELECT finalizeAggregation(
    uniqThetaIntersect(
        uniqThetaMergeStateIf(arrayReduce('uniqThetaState', [1, 2]::Array(UInt32)), 1),
        uniqThetaMergeStateIf(arrayReduce('uniqThetaState', []::Array(UInt32)), 0)
    )
);
-- returns 2
```
With the fix it returns `0`. The trigger is exact: `rhs` reaching `ThetaSketchData::intersect` with both `sk_update == nullptr` and `sk_union == nullptr`.

b) **Root cause explained?** Yes. `uniqThetaMergeStateIf(s, 0)` only calls `create` on the nested `ThetaSketchData`; nothing else allocates `sk_update` or `sk_union`. In `ThetaSketchData::intersect` we initialize a `theta_intersection` with `this`'s contents, then update it with `rhs`'s contents only when `rhs.sk_update` or `rhs.sk_union` exists. With both null, the intersection has a single input, and `theta_intersection::get_result` returns that input unchanged — so the intersection of `{1, 2}` with the empty set was reported as `{1, 2}` instead of `{}`.

c) **Fix matches root cause?** Yes. Intersection with an empty set is mathematically empty, so the fix short-circuits in that exact condition: when `rhs` has no recorded values, we drop our own data and return. This addresses the mechanism, not a downstream symptom.

d) **Test intent preserved? / New tests added?** New regression test `04215_91595_uniq_theta_intersect_empty_state.sql` covers the den-crane reproducer, both directions of `intersect(non_empty, empty)` and `intersect(empty, non_empty)`, the all-empty case, a non-empty sanity check, and the `AggregatingMergeTree` variant from #91595. Existing `01798_uniq_theta_union_intersect_not.sql` and `01798_uniq_theta_sketch.sql` still pass with the fix applied.

e) **Demonstrated in both directions?** Yes. Without the fix the new test produces `2` for the den-crane line and an inflated count for the other intersect-with-empty cases; with the fix it produces `0` everywhere it should and `5` for the non-empty sanity check.

f) **Fix is general, not a narrow patch?** Yes. I checked `merge` (union) and `aNotB` for the same empty-rhs pattern: union with empty = self and `A \ empty = A`, both of which the existing code already preserves correctly when `rhs` is fresh — only `intersect` has the asymmetric semantics that need the guard. The fix is at the boundary where ClickHouse hands data to DataSketches, so it covers every path that reaches `ThetaSketchData::intersect`, not just the `MergeStateIf` reproducer.

groeneai:
cc @alexey-milovidov @KochetovNicolai — could you review this? Small targeted fix in `ThetaSketchData::intersect`: when the right-hand side is a fresh, never-updated state (e.g. `uniqThetaMergeStateIf(s, 0)`), the intersection with DataSketches' `theta_intersection` was returning the unchanged left-hand side instead of empty. Also cc @den-crane — thanks for the precise diagnosis on #91595.

alexey-milovidov (approved):

clickgapai:

**Affected code:** `ThetaSketchData::intersect` — core code present in all supported branches.

**Why this might apply:** `uniqTheta` was introduced in version 22.9 (per FunctionsUniqTheta.cpp:64), well before all supported branches. This wrong-results bug affects all users of `uniqThetaIntersect` when the second argument is empty, which is a common scenario with `uniqThetaMergeStateIf` when predicates exclude all rows.

PR #104546, opened 2026-05-10
Add test for #102074

alexey-milovidov:
Add a regression test for `NOT_FOUND_COLUMN_IN_BLOCK` from #102074 (`PREWHERE` column not in read plan inside a parameterized view with JOIN). The bug no longer reproduces on master (reporter confirmed fixed in 25.10).

Closes #102074.

...

### Version info
- Merged into: `26.5.1.537`


alexey-milovidov (commented):
Ok.

PR #104548, opened 2026-05-10
Add test for #103508

### LINKED ISSUE #103508: NOT_FOUND_COLUMN_IN_BLOCK — `timestamp` renamed to `timestamp_0` in distributed CTE query when underlying table is non-empty (regression since 25.5)
## Describe the bug

When executing a CTE query against a `Distributed` table (backed by a `ReplicatedMergeTree` local table), ClickHouse's new analyzer renames the `timestamp` column to `timestamp_0` in the shard result block if the underlying local table has at least one row. This causes `NOT_FOUND_COLUMN_IN_BLOCK` (error code 10) for the outer `SELECT`.

The error does **not** occur when:
- The local table is empty (ClickHouse short-circuits plan construction)
- `allow_experimental_analyzer=0` is set

**Regression range:** works on 25.5.6, broken on 25.12.5 and 26.3.9.8.

---

## How to reproduce

```sql
-- 1. Create a local table and a Distributed table pointing to it
CREATE TABLE local_t (
    timestamp   DateTime64(9),
    id          String
) ENGINE = MergeTree ORDER BY timestamp;

CREATE TABLE dist_t AS local_t
ENGINE = Distributed(default, currentDatabase(), local_t, rand());

-- 2. Insert one row (any value — the bug fires even if the row is outside the query range)
INSERT INTO local_t VALUES (now(), 'a');

-- 3. Run the failing query
WITH A AS (
    SELECT * FROM dist_t
    WHERE timestamp >= now() - INTERVAL 10 MINUTE
      AND timestamp < now()
)
SELECT timestamp, id
FROM A
ORDER BY timestamp AS `timestamp` DESC
LIMIT 10
SETTINGS distributed_product_mode = 'allow';
```

**Failing versions:** 25.12.5, 26.3.9.8
**Passing version:** 25.5.6

---

## Error message and/or stacktrace

```
Code: 10. DB::Exception: Not found column __table1.timestamp in block.
There are only columns: __table1.timestamp_0, __global_row_index.
(NOT_FOUND_COLUMN_IN_BLOCK)
```

---

## Expected behavior

The query should execute successfully and return results. The `timestamp` column in CTE `A` (populated via `SELECT *` from a Distributed table) should remain accessible by its original name in the outer `SELECT`.

---

## Workaround

Either of these makes the query succeed:

```sql
-- Option A: disable the new analyzer
SETTINGS distributed_product_mode='allow', allow_experimental_analyzer=0

-- Option B: cast the column to the same type in SELECT (creates an expression node)
SELECT timestamp::DateTime64(9) AS `timestamp`, id
FROM A
ORDER BY `timestamp` DESC
```

---

## Additional context

- The `_0` suffix is the new analyzer's column deduplication marker. It appears to be applied to `timestamp` (a `DateTime64(9)` column) when `SELECT *` is used in a CTE against a Distributed table and the outer query references `timestamp` — but only when the local table has data.
- Empty local table → no error (CH skips full distributed plan construction).
- One row anywhere in the table (even outside the query time range) → error (CH builds the full distributed plan and the rename occurs).
- The rename happens during query analysis (`ExceptionBeforeStart`), before any rows are read.
- Related issues with the same analyzer class: #70356

### Version info
- Resolved by: #104548
- Merged into: `26.5.1.566`

alexey-milovidov:
Add a regression test for `NOT_FOUND_COLUMN_IN_BLOCK` from #103508 (column `timestamp` renamed to `timestamp_0` in distributed CTE when underlying table is non-empty). The bug no longer reproduces on master.

Closes #103508.

...

### Version info
- Merged into: `26.5.1.566`


tests/queries/0_stateless/04229_distributed_cte_timestamp_rename_103508.sql
```diff
@@ -0,0 +1,33 @@
...
+-- The originally reported scenario uses default settings and is fixed on
+-- master. Setting `optimize_read_in_order = 0` still hits a related rename
+-- in the analyzer, so the regression query pins it to `1` to keep this
+-- test stable under random-settings runs.
+
+DROP TABLE IF EXISTS dist_t;
+DROP TABLE IF EXISTS local_t;
+
+CREATE TABLE local_t (timestamp DateTime64(9), id String) ENGINE = MergeTree ORDER BY timestamp;
+CREATE TABLE dist_t AS local_t ENGINE = Distributed(test_shard_localhost, currentDatabase(), local_t, rand());
```
alexey-milovidov, line 17: Try with more than one shard.

alexey-milovidov (commented):
Ok.

PR #104556, opened 2026-05-10
Add test for #91946

alexey-milovidov:
Add a regression test for the `NOT_FOUND_COLUMN_IN_BLOCK` from #91946 (grouped `SELECT` on `ReplacingMergeTree FINAL` with a row policy attached). The bug no longer reproduces on master.

Closes #91946.

...

### Version info
- Merged into: `26.5.1.525`


Algunenano (approved):
Fixed at some point between 26.1 and 26.2

PR #93543, opened 2026-01-07
Functions stress test

al13n321:
### Changelog category (leave one):
- Build/Testing/Packaging Improvement

### Changelog entry:
Implement a stress test for functions, checking the soundness of their various properties.

A test that runs random SQL functions on random arguments and checks many things like:
 * If a function is deterministic, and we ran it twice on the same input, it should output the same value
 * ... even if we made some of the arguments const
 * ... even if we ran it first on multiple rows, then on one of those rows.
 * Iff function succeeded when run on multiple rows at once, it should succeed when run on each row indivudually.
 * Constness of return type doesn't change if you change number of rows in the input (without changing their constness).
 * Function's total MemoryTracker alloc/free amount adds up to zero (i.e. nothing leaked).
 * Function throws typechecking-related exceptions only during overload resolution, not during execution. (This seems hopeless, way too many functions fail this, and the IFunction et al interfaces make it really difficult to implement functions properly.)
 * Monotonicity and injectivity.

Currently many of these checks fail on many functions for many different reasons. I'm working on triaging suppressing these errors before the test can run in CI.

Here are some classes of problems found:
 * Interfaces in IFunction.h are terrible, imo. Implementing a function correctly is difficult and requires lots of boilerplate (both the IFunction way and the IFunctionOverloadResolver+IFunctionBase+IExecutableFunction way). Many implementations take shortcuts:
   * Often types are checked only during execution. Because you have to check IColumn types there anyway to get data out of them, so it's a convenient place for that. Maybe there should be an option for the function implementation to provide one function that does both checking and execution? During query analysis it would just be called on columns with 0 rows.
   * Sometimes functions want to carry some information from the typechecking stage (e.g. `getReturnTypeImpl`) to the execution stage (`executeImpl`). This is super inconvenient to do, so some `IFunction`s have `mutable` fields for that (e.g. https://github.com/ClickHouse/ClickHouse/pull/90644 ), which may break things in subtle ways if the same `IFunctionOverloadResolver` instance (which is just an `IFunction` pointer on the inside) is used to create multiple `IExecutableFunction` instances (which are also just an `IFunction` pointer on the inside, *the exact same one, shared*). Maybe `IFunction` should have a virtual `clone()` method, and `FunctionToOverloadResolverAdaptor::build` should use it, so that each `IFunction` instance gets only one `getReturnTypeImpl` call for only one overload, and `getReturnTypeImpl` can be non-const and just assign fields?
[...]

PR #104578, opened 2026-05-11
Backport fix for IPv6StringToNumOrDefault to 25.3

Avogar:
Backport a fix from https://github.com/ClickHouse/ClickHouse/pull/93543

Fix `IPv6StringToNumOrDefault` not properly zeroing output bytes on parse failure. Only one byte was zeroed (`vec_res[i] = 0`) instead of all 16 bytes (`std::fill_n(&vec_res[out_offset], offset_inc, 0)`), causing stale data from previous successful parses to leak into default results.

---
**Changelog category**: Bug Fix (user-visible misbehavior in an official stable release)
**Changelog entry**: Fixed `IPv6StringToNumOrDefault` returning incorrect results when mixed valid and invalid IPv6 strings were processed in the same batch.


Algunenano:
Note that 25.3 is already unsupported

Issue #94176, opened 2026-01-14
Iceberg History is broken for some Iceberg table


melvynator:
### Company or project name

_No response_

### Describe what's wrong

It appears that the iceberg_history system table is broken for some Iceberg tables. 

I've created a few Iceberg tables with Databricks and ClickHouse can't read their snapshot history. The issue is materialized by the fact that no history is being shown. 

```
SELECT *
FROM system.iceberg_history

Query id: 6266dbf6-22ee-4f90-81ca-93a53d7474d1

0 rows in set. Elapsed: 0.737 sec.
```

I will try to test it on other Iceberg table using a different writer. 

### Does it reproduce on the most recent release?

Yes

### How to reproduce

```
SHOW CREATE TABLE default.test

Query id: 21518605-e328-4ad1-8c51-450b39e8b2cd

   ┌─statement──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
1. │ CREATE TABLE default.test                                                                                                                                                                                                                                 ↴│
   │↳(                                                                                                                                                                                                                                                         ↴│
   │↳    `pull_request_number` Nullable(Int64),                                                                                                                                                                                                                ↴│
   │↳    `commit_sha` Nullable(String),                                                                                                                                                                                                                        ↴│
   │↳    `check_start_time` Nullable(DateTime64(6, 'UTC')),                                                                                                                                                                                                    ↴│
   │↳    `check_name` Nullable(String),                                                                                                                                                                                                                        ↴│
   │↳    `instance_type` Nullable(String),                                                                                                                                                                                                                     ↴│
   │↳    `instance_id` Nullable(String),                                                                                                                                                                                                                       ↴│
   │↳    `event_date` Nullable(Date),                                                                                                                                                                                                                          ↴│
   │↳    `event_time` Nullable(DateTime64(6, 'UTC')),                                                                                                                                                                                                          ↴│
   │↳    `event_time_microseconds` Nullable(DateTime64(6, 'UTC')),                                                                                                                                                                                             ↴│
   │↳    `thread_name` Nullable(String),                                                                                                                                                                                                                       ↴│
   │↳    `thread_id` Nullable(Decimal(20, 0)),                                                                                                                                                                                                                 ↴│
   │↳    `level` Nullable(String),                                                                                                                                                                                                                             ↴│
[...]

PR #104579, opened 2026-05-11
Fix datatype in iceberg history

SmitaRKulkarni:

Updated datatype of fields in iceberg history from Int32 - Int64 resolves #https://github.com/ClickHouse/ClickHouse/issues/94176

### Version info
- Merged into: `26.5.1.559`

### DIFF
--- a/src/Storages/ObjectStorage/DataLakes/Iceberg/Snapshot.h
+++ b/src/Storages/ObjectStorage/DataLakes/Iceberg/Snapshot.h
@@ -36,10 +36,10 @@ struct IcebergHistoryRecord
     bool is_current_ancestor;
     Iceberg::IcebergPathFromMetadata manifest_list_path;
 
-    Int32 added_files = 0;
-    Int32 added_records = 0;
-    Int32 added_files_size;
-    Int32 num_partitions;
+    Int64 added_files = 0;
+    Int64 added_records = 0;
+    Int64 added_files_size = 0;
+    Int64 num_partitions = 0;
 };
 
 using IcebergHistory = std::vector<Iceberg::IcebergHistoryRecord>;
--- a/src/Storages/ObjectStorage/DataLakes/Iceberg/Snapshot.h
+++ b/src/Storages/ObjectStorage/DataLakes/Iceberg/Snapshot.h
@@ -14,7 +14,7 @@ struct IcebergDataSnapshot
 {
     DB::ManifestFileCacheKeys manifest_list_entries;
     Int64 snapshot_id;
-    Int32 schema_id_on_snapshot_commit;
+    Int64 schema_id_on_snapshot_commit;
     std::optional<size_t> total_rows;
     std::optional<size_t> total_bytes;
     std::optional<size_t> total_position_delete_rows;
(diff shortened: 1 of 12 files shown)


tests/integration/test_storage_iceberg_no_spark/test_iceberg_history_large_summary.py
```diff
@@ -0,0 +1,75 @@
...
+
+import pytest
+
+from helpers.iceberg_utils import (
+    create_iceberg_table,
+    get_uuid_str,
+)
+
+
+INSERT_SETTINGS = {"allow_insert_into_iceberg": 1}
```
scanhex12: Maybe we need to do xml files with this change (for all iceberg tests with spark)?

tests/integration/test_storage_iceberg_no_spark/test_iceberg_history_large_summary.py
```diff
@@ -0,0 +1,75 @@
...
+        started_cluster_iceberg_no_spark,
+        "(x Int)",
+        format_version,
+    )
+    instance.query(f"INSERT INTO {table_name} VALUES (1);", settings=INSERT_SETTINGS)
+
+    meta, prev = _read_latest_metadata(instance, table_name)
+    assert meta.get("snapshots"), "snapshot must be present after INSERT"
+
+    # The exact value from the bug report; > INT32_MAX (2147483647).
```
scanhex12: please convert this comment into assert (like assert huge > 2147483647)

src/Storages/ObjectStorage/DataLakes/Iceberg/Snapshot.h
scanhex12: Let's also convert schema_id to Int64 (I suppose it also can be greater than max int32)

tests/integration/test_storage_iceberg_no_spark/test_iceberg_history_large_summary.py
```diff
@@ -0,0 +1,75 @@
+gimport json
```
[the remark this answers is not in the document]
scanhex12: +1

scanhex12 (approved):
The change looks good

tests/integration/test_storage_iceberg_no_spark/test_iceberg_history_large_summary.py
SmitaRKulkarni: addressed it

tests/integration/test_storage_iceberg_no_spark/test_iceberg_history_large_summary.py
SmitaRKulkarni: addressed it

src/Storages/ObjectStorage/DataLakes/Iceberg/Snapshot.h
SmitaRKulkarni: I have changed only schema_id_on_snapshot_commit, if you meant schema_id everywhere, would be a much larger refactor — happy to do it but maybe in a different PR. 
I think we should as it can be greater than Int32. please let me know what you think.

clickgapai:

**Affected code:** `IcebergMetadata::getHistory` parsing of snapshot summary fields — core code present in all supported branches.

**Why this might apply:** The `getHistory` function was added in PR #78244 (merged 2025-04-15). All supported release branches (25.8, 26.2, 26.3, 26.4) contain the buggy Int32 parsing code that fails for large Iceberg tables. Users with billions of records in Iceberg tables would hit this bug.

PR #104580, opened 2026-05-11
Add JSONValues function to extract values at specific JSON paths

mneedham:
- New Feature

### Changelog entry:
Add `JSONValues(json, 'path1', 'path2', ...)` function that returns an `Array(String)` of values at the specified paths from a JSON column, in argument order. Absent or null paths are omitted. Supports text indexes for efficient filtering on specific subcolumns.

### Why is this needed?

When working with a JSON column, there are two existing options for building a text index:

1. **Separate indexes per subcolumn** — works but creates multiple index structures and per-path maintenance overhead:
   ```sql
   INDEX idx_type   json.type.name::String   TYPE text(tokenizer = 'ngrams') GRANULARITY 100000000,
   INDEX idx_player json.player.name::String TYPE text(tokenizer = 'ngrams') GRANULARITY 100000000
   ```

2. **`JSONAllValues(json)`** — a single index covering *all* paths. This is simpler but the index is larger (covers every path in every row) and produces more false positives at granule-skip time when you only care about a few specific paths.

There is no way to get a single index covering only a chosen subset of paths - the ClickHouse docs AI says the following:

> "if you want a single index covering both paths, you'd need to use `JSONAllValues` (which covers all paths) — there's no built-in way to pass a subset of paths to it"

So this function would solve the problem. We would use it like this:

```sql
CREATE TABLE events
(
    match_id UInt32,
    json JSON(id String, index UInt32),
    INDEX vals JSONValues(json, ['type.name', 'player.name'])
        TYPE text(tokenizer = 'ngrams') GRANULARITY 100000000
)
ENGINE = MergeTree
ORDER BY (match_id, json.index)
```

Queries on the listed subcolumns use the index; queries on unlisted paths (e.g., `json.score`) would ignore it.

### What changed

- **`src/Functions/JSONAllValues.cpp`** — added `FunctionJSONValues`. Shared helpers (`serializeValueIntoResult`, `emitSharedDataValue`) are extracted as free functions used by both `JSONAllValues` and `JSONValues`. Per-path lookup uses typed/dynamic column maps (O(1)) and `ColumnObject::findPathLowerBoundInSharedData` for binary search in shared data.

- **`src/Storages/MergeTree/MergeTreeIndexJSONSubcolumnHelper`** — added `tryMatchNodeToJSONValuesIndex` which parses `JSONValues(col, 'path1', 'path2', ...)` index expression names and checks whether the queried subcolumn path is one of the listed paths.

- **`src/Storages/MergeTree/MergeTreeIndexConditionText.cpp`** — wired `tryMatchNodeToJSONValuesIndex` into the two call sites that currently check for `JSONAllValues`.

**Motivation:** Users building text indexes on JSON columns currently have no way to target specific paths with a single index. `JSONValues` enables a smaller, more precise index that avoids false positives from unrelated paths.

**Parameters:**
- `json` — a column of type `JSON`
- `path1, path2, ...` — one or more constant string path names (dot-separated, e.g. `'type.name'`)

**Example:**
```sql
-- Function usage
SELECT JSONValues(json, 'type.name', 'player.name') FROM events;
-- ['goal', 'Salah']

-- As a text index expression
INDEX idx JSONValues(json, ['type.name', 'player.name']) TYPE text(tokenizer = 'ngrams') GRANULARITY 100000000
```

Extra notes:

I did also try to create an index like this:

```sql
CREATE TABLE events
(
    match_id UInt32,
    json JSON(id String, index UInt32),
    INDEX vals [json.type.name, json.player.name]
        TYPE text(tokenizer = 'ngrams') GRANULARITY 100000000
)
ENGINE = MergeTree
ORDER BY (match_id, json.index)
```

And it created the index (at least as far as `system.parts` told me!) but then didn't use it at query time.


CurtizJ:
@mneedham 

Maybe it's better to pass paths as an array?

mneedham:
> @mneedham
> 
> Maybe it's better to pass paths as an array?

yeh good idea. Have updated

alexey-milovidov:

PR #104582, opened 2026-05-11
Introduce Partition Constants Substitution During Index Analysis

Michicosun:
This PR introduces partition-level constant substitution during index analysis.

When a filter references the partition key together with primary-key or skip-index columns, the partition value is folded into the predicate separately for each part. This lets `MergeTree` primary-key, `MinMax`, and skip-index analysis prune more parts and granules without weakening the filter (the original, unsubstituted filter is still applied during execution).

The behavior is controlled by the new setting `use_constant_folding_in_index_analysis`, which is disabled by default.

- Performance Improvement

Added the `use_constant_folding_in_index_analysis` setting (disabled by default). When enabled, `MergeTree` primary-key, `MinMax`, and skip-index analysis fold partition-level constants into the filter predicate separately for each part, improving pruning for filters whose branches depend on partition values, e.g. `(a = 1 AND b >= 1) OR (a = 2 AND b > 10)` with `PARTITION BY a`.

### Version info
- Merged into: `26.7.1.598` (included in `26.7` and later)


tests/queries/0_stateless/04218_partition_pk_pruning_constants_folding.reference
```diff
@@ -0,0 +1,23 @@
...
+          a
+        Condition: or((a in [3, 3]), or((a in [2, 2]), (a in [1, 1])))
+        Parts: 3/3
+        Granules: 9/9
+      PrimaryKey
+        Keys:
+          b
+        Condition: or((b in [11, +Inf)), or((b in [11, +Inf)), (b in [1, +Inf))))
+        Parts: 1/3
+        Granules: 3/9
```
Michicosun, line 27: Filtering here is the most valuable part of the test.

UnamedRus: It looks somewhat interleaving with https://github.com/ClickHouse/ClickHouse/pull/103480#issuecomment-4407462795.

My idea of "fixing" index instead of query conditions was that index is much simpler structure wise, and it might be not trivial to correctly deduce all per partition/granule range boundaries.

WDYT?

Michicosun: The problem I observed is that, for indices and pk where no partition columns are involved, KeyCondition simply drops the partition predicates from the filter (since one of the argument columns is unknown). 

So I am not sure it is possible to do such optimization completely within KeyCondition; at least it will require some information about well-known constants or boundaries.

UnamedRus: Ok, lets rephrase:
You currently want to fix filters, when we have OR (obv) joined pairs of conditions: (one target PK key, second target Partition key) in filterPartsByPartition/filterPartsByPrimaryKeyAndSkipIndexes, by injecting generateForPartition constraint in them

Q: Does it remove individual ranges of part, or only remove complete parts when it's possible?

My problem and fix is targeting when we have OR (obv) joined pairs of conditions: (one target PK key, second target Partition key & PK key)

So, PR just limit max possible range for index mark value(between -inf and x with between part_minmax and x and same for +inf), if this column also part of partition key.
This mostly target removal of individual granules (first/last) from ranges and ranges themselves, which actually will be read from part.

I'm just thinking do we need to have individual fixes, or they can be joined somehow. (presumably on your pr, as it bring more infrastructure for that)

Michicosun: It looks like changes are independent when the key condition already includes the partition column - it is nice to use real min-max values for estimation.

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
```diff
@@ -621,11 +647,11 @@ RangesInDataParts MergeTreeDataSelectExecutor::filterPartsByPartition(
 
     if (metadata_snapshot->hasPartitionKey())
     {
-        chassert(minmax_idx_condition && partition_pruner);
+        chassert(minmax_idx_condition);
         const auto & partition_key = metadata_snapshot->getPartitionKey();
         minmax_columns_types = MergeTreeData::getMinMaxColumnsTypes(partition_key);
 
-        if (settings[Setting::force_index_by_date] && (minmax_idx_condition->alwaysUnknownOrTrue() && partition_pruner->isUseless()))
+        if (settings[Setting::force_index_by_date] && minmax_idx_condition->generateUnsubstituted().alwaysUnknownOrTrue())
```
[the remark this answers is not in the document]
Michicosun: I will not fix this; ignore it in future reviews.

alexey-milovidov:

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
```diff
@@ -630,9 +659,7 @@ RangesInDataParts MergeTreeDataSelectExecutor::filterPartsByPartition(
     if (minmax_idx_condition)
         minmax_columns_types = MergeTreeData::getMinMaxColumns(metadata_snapshot->getPartitionKey(), data.getSettings()).getTypes();
 
-    if (metadata_snapshot->hasPartitionKey() && settings[Setting::force_index_by_date]
-        && (!minmax_idx_condition || minmax_idx_condition->alwaysUnknownOrTrue())
-        && (!partition_pruner || partition_pruner->isUseless()))
+    if (metadata_snapshot->hasPartitionKey() && settings[Setting::force_index_by_date] && (minmax_idx_condition->generateUnsubstituted().alwaysUnknownOrTrue() && partition_pruner->isUseless()))
```
nihalzp: Why we do not need the old guards `!minmax_idx_condition` and `!partition_pruner`?

src/Storages/MergeTree/ConditionTemplate.cpp
```diff
@@ -0,0 +1,292 @@
...
+    return names;
+}
+
+ActionsDAG substituteConstantInputs(
+    const ActionsDAG::Node * predicate_node,
+    const NameSet & inputs_to_substitute,
+    const MergeTreePartition & partition,
+    const String & partition_id,
+    const StorageMetadataPtr & metadata_snapshot)
+{
```
nihalzp, line 93: To make the logic more modular and potentially usable by other components in the future, I think we can create a new method `ActionsDAG::substituteByName(const std::unordered_map<String, ColumnWithTypeAndName> & subs)` that substitutes in-place? This also avoids reimplementing  `cloneSubDAG` manually. Then, helper `collectReachableNodesPostOrder` can be removed and this functions becomes:

```cpp
auto dag =  ActionsDAG::cloneSubDAG({predicate_node}, ...);

/// ... build subs

dag.substituteByName(subs);

dag.removeUnusedActions(...);
```

src/Processors/QueryPlan/ReadFromMergeTree.cpp
```diff
@@ -2084,24 +2100,34 @@ void ReadFromMergeTree::buildIndexes(
...
+            const auto * vector_similarity_index = typeid_cast<const MergeTreeIndexVectorSimilarity *>(index_helper.get());
+            chassert(vector_similarity_index);
+
+            factory = [vector_similarity_index, query_context, vector_search_parameters](const ActionsDAG *, const ActionsDAG::Node * predicate)
+            {
+                return vector_similarity_index->createIndexCondition(predicate, query_context, vector_search_parameters);
+            };
 #endif
-            if (!condition)
-                throw Exception(ErrorCodes::LOGICAL_ERROR, "Unknown vector search index {}", index_helper->index.name);
```
nihalzp, line 2287: Why we do not need this `LOGICAL_ERROR` check anymore anymore?

src/Storages/MergeTree/ConditionTemplate.cpp
```diff
@@ -0,0 +1,292 @@
...
+        transformer = std::move(transformer_);
+        return;
+    }
+
+    transformer = [prev_transform = std::exchange(transformer, nullptr), next_transform = std::move(transformer_)](Cond & condition) -> Cond
+    {
+        prev_transform(condition);
+        next_transform(condition);
+        return condition;
+    };
```
nihalzp: From declaration `transform` has void return `using Transformer = std::function<void(Cond &)>;`. I also have check all callsites, we never use the returned value.
```suggestion
    transformer = [prev_transform = std::exchange(transformer, nullptr), next_transform = std::move(transformer_)](Cond & condition)
    {
        prev_transform(condition);
        next_transform(condition);
    };
```

src/Storages/MergeTree/ConditionTemplate.cpp
```diff
@@ -0,0 +1,292 @@
...
+namespace ErrorCodes
+{
+    extern const int LOGICAL_ERROR;
+}
+
+namespace
+{
+
+const std::string PARTITION_ID_NAME = "_partition_id";
+const std::string PARTITION_VALUE_NAME = "_partition_value";
```
nihalzp: Could we reuse `PartitionIdColumn` from `src/Storages/MergeTree/MergeTreeVirtualColumns.h`? Maybe we can add `PartitionValueColumn` to `MergeTreeVirtualColumns` and reuse it across the codebase.

src/Core/Settings.cpp
```diff
@@ -1684,6 +1684,17 @@ Possible values:
 - 0 — Disabled.
 - 1 — Enabled.
 )", 0, use_partition_key) \
+    DECLARE(Bool, use_constant_folding_in_index_analysis, false, R"(
```
nihalzp, line 1753: I feel the setting is bit too broad. Maybe `use_partition_constant_folding_in_index_analysis`?

src/Core/Settings.cpp
```diff
@@ -1684,6 +1684,17 @@ Possible values:
...
+Substitute partition-level constants into the filter predicate when analyzing per-part primary
+key and skip indexes. With this enabled, conditions like `partition_col = X AND <pk-pred>` get
```
nihalzp: I do not think the condition `partition_col = X AND <pk-pred>` is good; we always prune fully and partition constant substitution does not give more pruning. Maybe we can use `(partition_col = 1 AND <pk-pred-a>) OR (partition_col = 2 AND <pk-pred-b>) OR …`.

src/Storages/MergeTree/MergeTreeData.cpp
nihalzp, line 1: I think we are missing test coverage for this path.

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
```diff
@@ -675,14 +702,14 @@ RangesInDataParts MergeTreeDataSelectExecutor::filterPartsByPartition(
 
     if (minmax_idx_condition)
     {
-        auto description = minmax_idx_condition->getDescription();
+        auto description = minmax_idx_condition->generateUnsubstituted().getDescription();
```
nihalzp, line 731: Why do we use unsubstituted version for description?

src/Storages/MergeTree/ConditionTemplate.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+    auto specialized = substituteConstantInputs(dag->predicate, partition_constant_names, partition, partition_id, metadata_snapshot);
+    chassert(!specialized.getOutputs().empty());
+
+    Cond produced = generate(&specialized, specialized.getOutputs().front());
+    const auto [it, inserted] = cache.emplace(partition_id, std::move(produced));
+    chassert(inserted);
+
+    return it->second;
+}
```
nihalzp, line 216: Instead of locking at the start, I think locking only during lookup and insert is enough and should reduce contention. This should be a performance win for high partition count case with many threads.

src/Storages/MergeTree/ConditionTemplate.cpp
nihalzp: This is bit hard to read. Could this be replace with `std::vector<Transformer>` which I feel would be easier to reason about?

src/Processors/QueryPlan/ReadFromMergeTree.cpp
```diff
@@ -2084,24 +2100,34 @@ void ReadFromMergeTree::buildIndexes(
...
+            };
         }
 
-        if (condition && !condition->alwaysUnknownOrTrue())
-            skip_indexes.useful_indices.emplace_back(index_helper, condition);
+        auto condition_template = std::make_shared<ConditionTemplate<MergeTreeIndexConditionPtr>>(filter_dag_ptr, std::move(factory), metadata_snapshot, query_context, skip_constant_folding);
+
+        const auto & unsubstituted = condition_template->generateUnsubstituted();
+        if (unsubstituted && !unsubstituted->alwaysUnknownOrTrue())
+            skip_indexes.useful_indices.emplace_back(index_helper, std::move(condition_template));
```
nihalzp, line 2322: Why do we check usefulness on unsubstituted version? I can imagine a scenario when index becomes useful only after partition constant folding.

nihalzp (commented):
The overall idea makes sense. I have left a few comments.

--- 
The changelog entry should be `Performance Improvement` and we should also mention the setting in the changelog entry.

---

The following query throws with setting on:

```sql
DROP TABLE IF EXISTS test;
CREATE TABLE test
(
    p Int64,
    b UInt64
)
ENGINE = MergeTree
ORDER BY b
PARTITION BY p
SETTINGS index_granularity = 1;

INSERT INTO test VALUES (1, 10);

SELECT count()
FROM test
WHERE p != 1 AND intDiv(1, p - 1) > 0
SETTINGS use_constant_folding_in_index_analysis = 0; -- works

SELECT count()
FROM test
WHERE p != 1 AND intDiv(1, p - 1) > 0
SETTINGS use_constant_folding_in_index_analysis = 1; -- throws `Division by zero.`
```
Also, ideally, if we turned off all index analysis via setting, we should never throw anything during analysis.

```sql
SELECT count()
FROM test
WHERE p != 1 AND intDiv(1, p - 1) > 0
SETTINGS
    use_primary_key = 0,
    use_partition_pruning = 0,
    use_skip_indexes = 0,
    use_constant_folding_in_index_analysis = 1; -- still throws
```
---

The following query slient missing rows with the setting when column name has collision with partition id name:

```sql
SET enable_analyzer = 1;

DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    `_partition_id` String,
    b UInt64
)
ENGINE = MergeTree
ORDER BY b
PARTITION BY `_partition_id`
SETTINGS index_granularity = 1;

INSERT INTO test VALUES ('a/b', 10), ('plain', 20);

SELECT 'disabled';
SELECT `_partition_id`, b
FROM test
WHERE `_partition_id` = 'a/b' AND b = 10
ORDER BY b
SETTINGS use_constant_folding_in_index_analysis = 0; -- returns 1 row

SELECT 'enabled';
SELECT `_partition_id`, b
FROM test
WHERE `_partition_id` = 'a/b' AND b = 10
ORDER BY b
SETTINGS use_constant_folding_in_index_analysis = 1; -- returns 0 row

DROP TABLE test;
```

Similar issue with `_partition_value`:

```sql
SET enable_analyzer = 1;

DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    `_partition_value` String,
    b UInt64
)
ENGINE = MergeTree
ORDER BY b
PARTITION BY `_partition_value`
SETTINGS index_granularity = 1;

INSERT INTO test VALUES ('a', 10), ('b', 20);

SELECT 'disabled';
SELECT `_partition_value`, b
FROM test
WHERE `_partition_value` = 'a' AND b = 10
ORDER BY b
SETTINGS use_constant_folding_in_index_analysis = 0; -- ok

SELECT 'enabled';
SELECT `_partition_value`, b
FROM test
WHERE `_partition_value` = 'a' AND b = 10
ORDER BY b
SETTINGS use_constant_folding_in_index_analysis = 1; -- throws `TOO_DEEP_RECURSION`

DROP TABLE test;
```

---

Silent wrong result when column name has the same expression as partition key:

```sql
DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    d Date,
    `toYYYYMM(d)` UInt32,
    b UInt64
)
ENGINE = MergeTree
ORDER BY b
PARTITION BY toYYYYMM(d)
SETTINGS index_granularity = 1;

INSERT INTO test VALUES ('2024-01-01', 7, 10);

SELECT d, `toYYYYMM(d)`, b
FROM test
WHERE `toYYYYMM(d)` = 7 AND b = 10
SETTINGS use_partition_pruning = 0, use_skip_indexes = 0,
         use_constant_folding_in_index_analysis = 0; -- returns 1 row

SELECT d, `toYYYYMM(d)`, b
FROM test
WHERE `toYYYYMM(d)` = 7 AND b = 10
SETTINGS use_partition_pruning = 0, use_skip_indexes = 0,
         use_constant_folding_in_index_analysis = 1; -- returns 0 row
```

---

Even if we prune using primary key because of constant folding, the `force_primary_key = 1` throws:

```sql
DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    p UInt64,
    b UInt64
)
ENGINE = MergeTree
ORDER BY b
PARTITION BY p
SETTINGS index_granularity = 1;

INSERT INTO test VALUES
    (1, 10),
    (1, 11),
    (2, 20);

SELECT 'explain';
EXPLAIN indexes = 1
SELECT *
FROM test
WHERE p != 1 OR b = 10
SETTINGS
    use_constant_folding_in_index_analysis = 1,
    use_skip_indexes = 0; -- we prune 2 out of 3 granules using primary key

SELECT 'force';
SELECT *
FROM test
WHERE p != 1 OR b = 10
SETTINGS
    use_constant_folding_in_index_analysis = 1,
    use_skip_indexes = 0,
    force_primary_key = 1; -- throws

```

Same situation with min-max pruning:

```sql
DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    a UInt64,
    b UInt64
)
ENGINE = MergeTree
ORDER BY b
PARTITION BY a
SETTINGS index_granularity = 1;

INSERT INTO test VALUES
    (1, 10),
    (2, 20);

SELECT 'without force';
EXPLAIN indexes = 1
SELECT *
FROM test
WHERE if(a = 1, 1, 0) = 1
SETTINGS use_constant_folding_in_index_analysis = 1; -- partition minmax prunes 1 out of 2 parts

SELECT 'with force';
SELECT *
FROM test
WHERE if(a = 1, 1, 0) = 1
SETTINGS
    use_constant_folding_in_index_analysis = 1,
    force_index_by_date = 1; -- throws `INDEX_NOT_USED`
```

nihalzp:
Also, let's add the queries from the review to the tests so that future changes does not accidentally regress it.

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
Michicosun: When a partition key exists in metadata, we always create `minmax_idx_condition` and `partition_pruner`.

src/Processors/QueryPlan/ReadFromMergeTree.cpp
Michicosun: it was changed to chassert(vector_similarity_index);

src/Storages/MergeTree/ConditionTemplate.cpp
Michicosun: done

src/Storages/MergeTree/ConditionTemplate.cpp
Michicosun: done

src/Core/Settings.cpp
Michicosun: I wanted to name the setting so I could extend it with other columns, probably not related to partitioning. That is why the name is generic. Do you think we need to narrow its naming?

src/Storages/MergeTree/ConditionTemplate.cpp
Michicosun: added `ActionsDAG::substitute`

src/Core/Settings.cpp
Michicosun: narrowed docs - removed example.

src/Storages/MergeTree/MergeTreeData.cpp
Michicosun: hard to tell which one :)

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
Michicosun: Because substitution can be calculated only for a single part. Here we need to describe the index pass (above all parts).

src/Storages/MergeTree/ConditionTemplate.cpp
Michicosun: changed to vector

src/Processors/QueryPlan/ReadFromMergeTree.cpp
Michicosun: It will require rewriting this part a lot. I would suggest implementing it in the follow-up, since this pr is big enough.

src/Storages/MergeTree/ConditionTemplate.cpp
Michicosun: done

src/Storages/MergeTree/ConditionTemplate.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    try
+    {
+        auto specialized = substituteConstantInputs(dag->predicate, partition, partition_id, metadata_snapshot);
+        chassert(!specialized.getOutputs().empty());
+
+        Cond produced = generate(&specialized, specialized.getOutputs().front());
+        return setSubstituted(partition_id, std::move(produced));
+    }
+    catch (...) /// Ok. Substitution is done in best-effort way.
```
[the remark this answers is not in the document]
Michicosun: no.

src/Core/Settings.cpp
nihalzp: If we extend functionality in the future, it will more appropriate to have the new optimization under a new setting so that it is easy to debug and pin point  cause of any new bug; on top of that, multiple settings will give users more fine grained control.

src/Storages/MergeTree/ConditionTemplate.cpp
nihalzp: Let's only catch `DB::Exception`.

src/Core/Settings.cpp
nihalzp: I think we should have some example; otherwise, it is not possible for user to know when this optimization kicks in. Maybe we can use one of the expressions from the test.

src/Storages/MergeTree/MergeTreeData.cpp
nihalzp: I should have been more specific. I meant

`MergeTreeData.cpp:9074`

```cpp
if (minmax_idx_condition && !minmax_idx_condition->generateForPartition(part->partition).checkInHyperrectangle(part->getMinMaxIndex()->hyperrectangle, minmax_columns_types).can_be_true)
    continue;
```

nihalzp (commented):
Looks good! The most issues are resolved from last review.


We still need to update the change log type to `Performance Improvement`.

---

The last two queries from last review. I think the issue is that because the constant folding is data dependent, we cannot guarantee it. But it will be add these two queries to the test to document the behavior.


#### From last review:

Even if we prune using primary key because of constant folding, the `force_primary_key = 1` throws:

```sql
DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    p UInt64,
    b UInt64
)
ENGINE = MergeTree
ORDER BY b
PARTITION BY p
SETTINGS index_granularity = 1;

INSERT INTO test VALUES
    (1, 10),
    (1, 11),
    (2, 20);

SELECT 'explain';
EXPLAIN indexes = 1
SELECT *
FROM test
WHERE p != 1 OR b = 10
SETTINGS
    use_constant_folding_in_index_analysis = 1,
    use_skip_indexes = 0; -- we prune 2 out of 3 granules using primary key

SELECT 'force';
SELECT *
FROM test
WHERE p != 1 OR b = 10
SETTINGS
    use_constant_folding_in_index_analysis = 1,
    use_skip_indexes = 0,
    force_primary_key = 1; -- throws

```

Same situation with min-max pruning:

```sql
DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    a UInt64,
    b UInt64
)
ENGINE = MergeTree
ORDER BY b
PARTITION BY a
SETTINGS index_granularity = 1;

INSERT INTO test VALUES
    (1, 10),
    (2, 20);

SELECT 'without force';
EXPLAIN indexes = 1
SELECT *
FROM test
WHERE if(a = 1, 1, 0) = 1
SETTINGS use_constant_folding_in_index_analysis = 1; -- partition minmax prunes 1 out of 2 parts

SELECT 'with force';
SELECT *
FROM test
WHERE if(a = 1, 1, 0) = 1
SETTINGS
    use_constant_folding_in_index_analysis = 1,
    force_index_by_date = 1; -- throws `INDEX_NOT_USED`
```

alexey-milovidov:
Merged the latest `master` (resolved one conflict in `MergeTreeReaderTextIndex.cpp`: kept master's reworded comment together with this PR's `condition_template->generateUnsubstituted()` code) and addressed the outstanding review feedback from @nihalzp:

- **`ConditionTemplate::generateForPartition`**: narrowed the best-effort fallback from `catch (...)` to `catch (const Exception &)`, so only expected query-evaluation failures (e.g. division by zero while folding a partition constant) fall back to the unsubstituted condition.
- **Setting documentation**: restored a concrete example — a disjunctive filter whose branches target different partitions (`(a = 1 AND b >= 1) OR (a = 2 AND b > 10)` with `PARTITION BY a`).
- **Tests**:
  - `04417_partition_constants_folding_force_settings` documents that `force_primary_key` / `force_index_by_date` still throw `INDEX_NOT_USED` when pruning becomes effective only after per-part constant folding (the `force_*` checks run against the unsubstituted condition). This locks in the current, data-dependent behavior.
  - `04418_partition_constants_folding_minmax_count_projection` covers the implicit MinMax-count projection path in `MergeTreeData::getMinMaxCountProjectionBlock`.
- **Settings history**: moved the `use_constant_folding_in_index_analysis` entry from the `26.6` bucket to `26.7` (26.6 is released; the `02995_new_settings_history` baseline is v26.6.1), otherwise the settings-history test would fail.

The changelog category is already `Performance Improvement` and the entry mentions the setting. The `force_*`-vs-unsubstituted findings raised by the AI reviewer are the previously-noted design decision and are now documented by `04417` rather than changed.

alexey-milovidov:
Fixed the Fast test failures in `ec167eec424`. After the latest `master` merge, `explain_query_plan_default` now defaults to `pretty` (new default since 26.7, from #105036), which changed the default `EXPLAIN PLAN` output format. The five constant-folding tests that assert on `EXPLAIN indexes = 1` output (`04218`, `04219`, `04220`, `04221`, `04314`) were written against the legacy format and started failing with "result differs with reference".

I added `SET explain_query_plan_default = 'legacy';` to each of them so their existing (already reviewed) references — which demonstrate the per-part constant folding — stay valid. This mirrors how #105036 pinned the ~414 existing `EXPLAIN` tests when it changed the default. No reference files were touched.

nihalzp (commented):
Image

It looks related. It would be good to investigate why this performance test slowed down even though we have the setting is by default false.

nihalzp (approved):

tests/queries/0_stateless/04314_partition_constants_folding_derived_key.sql
```diff
@@ -0,0 +1,41 @@
...
+-- branches and lets the primary key prune extra granules. Skip indexes / partition pruning are
+-- disabled below so the EXPLAIN reflects only the primary key effect.
+
+SET enable_analyzer = 1;
+SET use_statistics_for_part_pruning=0;
+SET explain_query_plan_default = 'legacy';
+
+DROP TABLE IF EXISTS t_derived_key;
+CREATE TABLE t_derived_key (d Date, b UInt64)
+ENGINE = MergeTree ORDER BY b PARTITION BY toYYYYMM(d) SETTINGS index_granularity = 1;
```
PedroTadim, line 14: @Michicosun Is this going to work with a multiple column partitioned table? Maybe add another test for this case

Michicosun: This should work for any partitioning scheme.

PR #78041, opened 2025-03-20
Distributed INSERT SELECT into ReplicatedMergeTree

devcrafter:
https://github.com/ClickHouse/ClickHouse/issues/47799

### Changelog category (leave one):
- New Feature

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Distributed INSERT SELECT for replicated MergeTree tables now efficiently uses parallel replicas to parallelize INSERTs by selecting different data on different nodes and inserting them independently.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #104607, opened 2026-05-11
Docs: clarify description of `parallel_distributed_insert_select`

Blargian:
Clarifies an inaccurate docs description. Description was previously updated to say:

```
Setting `enable_parallel_replicas = 1` is needed when using this setting.
```

However this is an over-simplification.

Source: https://github.com/ClickHouse/ClickHouse/blob/master/src/Interpreters/ClusterProxy/executeQuery.cpp (`isSuitableForInsertSelectWithParallelReplicas` function). introduced in https://github.com/ClickHouse/ClickHouse/pull/78041.

As I see it:

Source | Behavior
-- | --
Distributed table | Each shard executes the query locally. No additional settings required.
s3Cluster, gcsCluster, hdfsCluster | Files are distributed across replicas. No additional settings required.
ReplicatedMergeTree / SharedMergeTree | Uses parallel replicas to distribute the read. Requires enable_parallel_replicas = 1, allow_experimental_analyzer = 1, and cluster_for_parallel_replicas to be configured. Available from v25.4.

...

### Version info
- Merged into: `26.5.1.773`


src/Core/Settings.cpp
```diff
@@ -1314,7 +1314,7 @@ Possible values:
 - `1` — `SELECT` will be executed on each shard from the underlying table of the distributed engine.
 - `2` — `SELECT` and `INSERT` will be executed on each shard from/to the underlying table of the distributed engine.
 
-Setting `enable_parallel_replicas = 1` is needed when using this setting.
+From v25.4, when using `parallel_distributed_insert_select = 2` to read from a `ReplicatedMergeTree` or `SharedMergeTree` source table, `enable_parallel_replicas = 1` is also required.
```
azat: Though it is not only for "read", but just in general and also it wasn't supported before for RMT/SMT, so maybe something like

```suggestion
Since v25.4, `INSERT ... SELECT` from a `ReplicatedMergeTree` or `SharedMergeTree` source can also be parallelized across replicas. To enable it:
- `parallel_distributed_insert_select = 2`
- `enable_parallel_replicas = 1`
```

azat (approved):

src/Core/Settings.cpp
azat: I will commit this, if you are okay with wording, please merge this

alexey-milovidov:

PR #104608, opened 2026-05-11
Hypothetical (WHAT-IF) indexes

yariks5s:
Adds **hypothetical (what-if) indexes**: virtual, session-scoped skip indexes you can attach to a `MergeTree` table to see how a real skip index would affect a query, without actually building it on disk.

```sql
CREATE HYPOTHETICAL INDEX idx_b ON t (b) TYPE minmax GRANULARITY 1;
EXPLAIN WHATIF SELECT * FROM t WHERE b = 42;
```

`EXPLAIN WHATIF` runs the index over the baseline-pruned granules in memory and reports the exact skip ratio plus estimated marks and bytes. If empirical is disabled (`empirical = 0`) or can't produce a result, the estimator falls back to column statistics; if neither is available it reports an applicability-only summary.

Defined indexes are inspectable through the new `system.hypothetical_indexes` table and discarded when the session ends. Existing queries against the table are unaffected.

- New Feature

Added hypothetical (what-if) skip indexes. Use `CREATE HYPOTHETICAL INDEX ... ON t (expr) TYPE ...` to define a session-scoped virtual skip index, then `EXPLAIN WHATIF SELECT ...` to estimate its skip ratio and cost without materializing it. Defined indexes are visible in the new `system.hypothetical_indexes` table.

### Version info
- Merged into: `26.6.1.1139`


src/Interpreters/InterpreterExplainQuery.cpp
```diff
@@ -745,6 +747,16 @@ QueryPipeline InterpreterExplainQuery::executeImpl()
 
             break;
         }
+        case ASTExplainQuery::WhatIf:
+        {
+            auto & query_ast = ast.getExplainedQuery();
+            if (!dynamic_cast<const ASTSelectWithUnionQuery *>(query_ast.get()))
+                throw Exception(ErrorCodes::INCORRECT_QUERY, "Only SELECT is supported for EXPLAIN WHATIF query");
+
+            auto whatif_result = WhatIfIndexEstimator::run(query_ast, query_context, ast.getSettings());
```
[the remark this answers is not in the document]
yariks5s: EXPLAIN WHATIF uses skip indexes correctly: InterpreterExplainQuery forces use_skip_indexes_on_data_read = false, which actually causes selectRangesToRead to apply skip-index pruning upfront (rather than deferring to read time), so the baseline already reflects existing skip-index pruning — the override helps WHATIF accuracy, not the other way around

alexey-milovidov:

src/Interpreters/HypotheticalIndexStore.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+namespace ErrorCodes
+{
+    extern const int BAD_ARGUMENTS;
+}
+
+bool HypotheticalIndexStore::sameTable(const StorageID & a, const StorageID & b)
+{
+    if (a.uuid != UUIDHelpers::Nil && b.uuid != UUIDHelpers::Nil)
+        return a.uuid == b.uuid;
```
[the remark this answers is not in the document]
yariks5s: Ordinary databases are deprecated?

tests/queries/0_stateless/04220_hypothetical_index.sh
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
+# Tags: no-replicated-database
```
nihalzp, line 2: Why do we want to skip it for `no-replicated-database`? Let's add a reason under it.

tests/queries/0_stateless/04221_hypothetical_index_empirical.sh
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/env bash
+# Tags: no-replicated-database, no-random-merge-tree-settings
+# no-random-merge-tree-settings: needs fixed index_granularity
```
nihalzp: Since we fix them with table creation (i.e. `SETTINGS index_granularity = 100, index_granularity_bytes = 0, min_bytes_for_wide_part = 0;`), do we need `no-random-merge-tree-settings`?

docs/en/sql-reference/statements/explain.md
```diff
@@ -649,6 +649,68 @@ EXPLAIN ESTIMATE SELECT * FROM ttt;
...
+  source:           empirical | statistical | applicability_only
+  empirical_status: ok | unsupported | disabled
+  sampled_parts:    50 / 100        (only when source = empirical)
+  sampled_marks:    50 / 100        (only when source = empirical)
+  elapsed_us:       631             (only when source = empirical)
+```
+
+- `source` — how the estimate was produced.
+  - `empirical`: built the index in memory over the baseline-pruned granules and counted the granules the index would skip. Exact.
+  - `statistical`: derived from column statistics. Used when empirical is disabled (`empirical = 0`) or empirical couldn't produce a result, and column statistics are defined on the relevant columns.
```
nihalzp, line 699: I think it will be nice to also mention what settings we would need to set in order to have column statistics. I think it is not present by default.

docs/en/sql-reference/statements/explain.md
```diff
@@ -649,6 +649,68 @@ EXPLAIN ESTIMATE SELECT * FROM ttt;
...
+
+INSERT INTO t SELECT number, number FROM numbers(10000);
+
+CREATE HYPOTHETICAL INDEX idx_b ON t (b) TYPE minmax GRANULARITY 1;
+
+-- Default — empirical (exact) measurement
+EXPLAIN WHATIF SELECT * FROM t WHERE b = 42;
+
+-- Force the statistical / applicability_only path
+EXPLAIN WHATIF empirical = 0 SELECT * FROM t WHERE b = 42;
```
nihalzp: We should show the output of these queries, followed up by some explanation of what the output means. Additionally, it would be helpful to have queries that exercise `statistical` path with `empirical = 0`.

docs/en/sql-reference/statements/hypothetical-index.md
```diff
@@ -0,0 +1,106 @@
...
+
+```sql
+CREATE TABLE t (a UInt64, b UInt64) ENGINE = MergeTree ORDER BY a
+SETTINGS index_granularity = 100;
+
+INSERT INTO t SELECT number, number FROM numbers(10000);
+
+CREATE HYPOTHETICAL INDEX idx_b ON t (b) TYPE minmax GRANULARITY 1;
+
+EXPLAIN WHATIF SELECT * FROM t WHERE b = 42;
```
nihalzp, line 50: Or maybe this could be a better place to show queries that highlight `statistical` estimation source.

src/Parsers/ASTHypotheticalIndexQuery.cpp
```diff
@@ -0,0 +1,80 @@
...
+        res->children.push_back(res->index_decl);
+    }
+
+    cloneTableOptions(*res);
+
+    return res;
+}
+
+void ASTHypotheticalIndexQuery::formatQueryImpl(
+    WriteBuffer & ostr, const FormatSettings & settings, FormatState & state, FormatStateStacked frame) const
```
nihalzp, line 46: Since `index_name` and `index_decl` can be `nullptr`, let's add `chasset(....)` (e.g. `chassert(index_decl)`) before accessing them.

src/Parsers/ASTHypotheticalIndexQuery.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace DB
+{
+
+String ASTHypotheticalIndexQuery::getID(char delim) const
+{
+    if (kind == DropAll)
+        return "DropAllHypotheticalIndexes";
+    return "HypotheticalIndexQuery" + (delim + getDatabase()) + delim + getTable();
+}
```
nihalzp, line 17:
```suggestion
String ASTHypotheticalIndexQuery::getID(char delim) const
{
    switch (kind)
    {
        case Create:  return "CreateHypotheticalIndexQuery" + (delim + getDatabase()) + delim + getTable();
        case Drop:    return "DropHypotheticalIndexQuery"   + (delim + getDatabase()) + delim + getTable();
        case DropAll: return "DropAllHypotheticalIndexes";
    }
    UNREACHABLE();
}
```

Otherwise, `EXPLAIN AST CREATE HYPOTHETICAL INDEX` and `EXPLAIN AST DROP HYPOTHETICAL INDEX` will have the same prefix which is bit less nice.

src/Storages/System/StorageSystemHypotheticalIndexes.cpp
```diff
@@ -0,0 +1,68 @@
...
+                database_name = db->getDatabaseName();
+                table_name = storage->getStorageID().getTableName();
+            }
+        }
+
+        size_t col = 0;
+        res_columns[col++]->insert(entry.index.name);
+        res_columns[col++]->insert(database_name);
+        res_columns[col++]->insert(table_name);
+        res_columns[col++]->insert(entry.index.type);
```
nihalzp, line 53: I think it's more beneficial to write the full type. For example:


```sql
CREATE HYPOTHETICAL INDEX i1 ON t (b) TYPE bloom_filter(0.01) GRANULARITY 1;
CREATE HYPOTHETICAL INDEX i2 ON t (b) TYPE bloom_filter(0.001) GRANULARITY 1;
SELECT name, type, expression, granularity FROM system.hypothetical_indexes;
```

```response
   ┌─name─┬─type─────────┬─expression─┬─granularity─┐
1. │ i1   │ bloom_filter │ b          │           1 │
2. │ i2   │ bloom_filter │ b          │           1 │
   └──────┴──────────────┴────────────┴─────────────┘
```
Both will give type but I feel having the numbers such as `0.001` is quite useful.

Skip indices table already supports it. Maybe we can take inspiration from there:

```sql
:) SELECT * FROM system.data_skipping_indices

SELECT *
FROM system.data_skipping_indices

Query id: 4e2cd564-44fd-44a0-a60f-87e2ebdcd6ed

   ┌─database─┬─table─┬─name────┬─type─────────┬─type_full───────────┬─expr─┬─creation─┬─granularity─┬─data_compressed_bytes─┬─data_uncompressed_bytes─┬─marks_bytes─┐
1. │ default  │ t     │ idx_dup │ minmax       │ minmax              │ a    │ Explicit │           1 │                     0 │                       0 │           0 │
2. │ default  │ t     │ i1      │ bloom_filter │ bloom_filter(0.01)  │ b    │ Explicit │           1 │                     0 │                       0 │           0 │
3. │ default  │ t     │ i2      │ bloom_filter │ bloom_filter(0.001) │ b    │ Explicit │           1 │                     0 │                       0 │           0 │
   └──────────┴───────┴─────────┴──────────────┴─────────────────────┴──────┴──────────┴─────────────┴───────────────────────┴─────────────────────────┴─────────────┘
```

src/Storages/System/StorageSystemHypotheticalIndexes.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+        String expr_str;
+        if (entry.index.expression_list_ast)
+        {
+            WriteBufferFromString buf(expr_str);
+            IAST::FormatSettings fmt_settings(/* one_line = */ true);
+            entry.index.expression_list_ast->format(buf, fmt_settings);
+        }
+        res_columns[col++]->insert(expr_str);
```
nihalzp:
```suggestion
        if (auto expression = entry.index.expression_list_ast)
            res_columns[col++]->insert(expression->formatForLogging());
        else
            res_columns[col++]->insertDefault();
```

src/Storages/System/StorageSystemHypotheticalIndexes.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        {"name", std::make_shared<DataTypeString>(), "Index name"},
+        {"database", std::make_shared<DataTypeString>(), "Database name"},
+        {"table", std::make_shared<DataTypeString>(), "Table name"},
+        {"type", std::make_shared<DataTypeString>(), "Index type (minmax, set, bloom_filter, etc)"},
+        {"expression", std::make_shared<DataTypeString>(), "Index expression"},
+        {"granularity", std::make_shared<DataTypeUInt64>(), "Index granularity"},
+    };
+}
+
```
nihalzp: Maybe it's a matter of style but having the column order aligned with skip indices would be more intuitive to existing users. `database -> table -> name`.

```sql
:) SELECT * FROM system.data_skipping_indices

SELECT *
FROM system.data_skipping_indices

Query id: 4e2cd564-44fd-44a0-a60f-87e2ebdcd6ed

   ┌─database─┬─table─┬─name────┬─type─────────┬─type_full───────────┬─expr─┬─creation─┬─granularity─┬─data_compressed_bytes─┬─data_uncompressed_bytes─┬─marks_bytes─┐
1. │ default  │ t     │ idx_dup │ minmax       │ minmax              │ a    │ Explicit │           1 │                     0 │                       0 │           0 │
2. │ default  │ t     │ i1      │ bloom_filter │ bloom_filter(0.01)  │ b    │ Explicit │           1 │                     0 │                       0 │           0 │
3. │ default  │ t     │ i2      │ bloom_filter │ bloom_filter(0.001) │ b    │ Explicit │           1 │                     0 │                       0 │           0 │
   └──────────┴───────┴─────────┴──────────────┴─────────────────────┴──────┴──────────┴─────────────┴───────────────────────┴─────────────────────────┴─────────────┘
```

src/Storages/System/StorageSystemHypotheticalIndexes.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <Databases/IDatabase.h>
+#include <Interpreters/Context.h>
+#include <Interpreters/DatabaseCatalog.h>
+#include <Interpreters/HypotheticalIndexStore.h>
+#include <IO/WriteBufferFromString.h>
+
+namespace DB
+{
+
+ColumnsDescription StorageSystemHypotheticalIndexes::getColumnsDescription()
```
nihalzp, line 15: If I drop the table, the entry in the table is still present:

```sql
CREATE TABLE t (a UInt64, b UInt64) ENGINE = MergeTree ORDER BY a;
CREATE HYPOTHETICAL INDEX i ON t (b) TYPE minmax GRANULARITY 1;
DROP TABLE t;
```


```sql
SELECT * FROM system.hypothetical_indexes; -- expected the table to be empty similar to data skipping index table
```

```response
   ┌─name─┬─database─┬─table─┬─type───┬─expression─┬─granularity─┐
1. │ i    │ default  │ t     │ minmax │ b          │           1 │
   └──────┴──────────┴───────┴────────┴────────────┴─────────────┘
```

nihalzp (commented):
I still need to have a deeper look some remaining files. I will review them next week. Here are the some of the comments.


```sql
DROP TABLE IF EXISTS t;

CREATE TABLE t
(
    a UInt64,
    s String
)
ENGINE = MergeTree
ORDER BY a
SETTINGS index_granularity = 100, index_granularity_bytes = 0, min_bytes_for_wide_part = 0;

INSERT INTO t SELECT number, if(number < 100, 'Hit', 'Miss') FROM numbers(1000);

CREATE HYPOTHETICAL INDEX idx_l ON t (lower(s)) TYPE set(100) GRANULARITY 1; -- works

EXPLAIN WHATIF SELECT * FROM t WHERE lower(s) = 'hit'; -- but throws but expected it to work
```

The following query has logical error:
```sql
DROP TABLE IF EXISTS t;

CREATE TABLE t
(
    id UInt32,
    message String
)
ENGINE = MergeTree
ORDER BY id
SETTINGS index_granularity = 2, index_granularity_bytes = 0, min_bytes_for_wide_part = 0;

INSERT INTO t VALUES
    (1, 'abc def foo'),
    (2, 'abc def bar'),
    (3, 'abc baz foo'),
    (4, 'abc baz bar'),
    (5, 'xyz');

CREATE HYPOTHETICAL INDEX idx_text
ON t (message) TYPE text(tokenizer = splitByNonAlpha) GRANULARITY 1;

EXPLAIN WHATIF SELECT * FROM t WHERE hasToken(message, 'foo');
```

We do not validate proposed index for hypothetical but I think it makes sense to validate because we do not want users to pay for a index that they will not be able to use.
```sql
SET allow_suspicious_indices = 0;

DROP TABLE IF EXISTS t;
CREATE TABLE t (a UInt64, b UInt64) ENGINE = MergeTree ORDER BY a;

--- Hypothetical index works
CREATE HYPOTHETICAL INDEX idx_dup ON t (a, a) TYPE minmax GRANULARITY 1;

--- But actual skip index does not allow it; throws
ALTER TABLE t ADD INDEX idx_dup (a, a) TYPE minmax GRANULARITY 1;
```

We overestimate the benefit with `FINAL`:
```sql
DROP TABLE IF EXISTS t;

CREATE TABLE t
(
    a UInt64,
    b UInt64
)
ENGINE = ReplacingMergeTree
ORDER BY a
SETTINGS index_granularity = 100;

INSERT INTO t SELECT number, number FROM numbers(1000);

CREATE HYPOTHETICAL INDEX idx_b ON t (b) TYPE minmax GRANULARITY 1;

EXPLAIN WHATIF SELECT * FROM t FINAL WHERE b = 42;
```

```response
    ┌─explain─────────────────────────────────────────────┐
 1. │ Baseline (after PK + partition + existing indexes): │
 2. │   table:       default.t                            │
 3. │   parts:       1                                    │
 4. │   marks:       10                                   │
 5. │   est_bytes:   9132 B                               │
 6. │                                                     │
 7. │ With idx_b (minmax, hypothetical):                  │
 8. │   status:       applicable                          │
 9. │   marks:        1                                   │
10. │   est_bytes:    913 B                               │
11. │   skip_ratio:   90.0%           
```

But a real index does not prune at all:

```sql
CREATE INDEX idx_b ON t (b) TYPE minmax GRANULARITY 1;

EXPLAIN indexes = 1
SELECT *
FROM t
FINAL
WHERE b = 42

Query id: 8ec0def9-2f5e-4406-959a-326a0b5677e2

    ┌─explain────────────────────────────────────────────────────────────────────────────────────────────────────────┐
 1. │ Expression ((Project names + Projection))                                                                      │
 2. │   Filter ((WHERE + Change column names to column identifiers))                                                 │
 3. │     ReadFromMergeTree (default.t)                                                                              │
 4. │     Indexes:                                                                                                   │
 5. │       PrimaryKey                                                                                               │
 6. │         Condition: true                                                                                        │
 7. │         Parts: 1/1                                                                                             │
 8. │         Granules: 10/10                                                                                        │
 9. │       Skip                                                                                                     │
10. │         Name: idx_b                                                                                            │
11. │         Description: minmax GRANULARITY 1                                                                      │
12. │         Condition: (b in [42, 42])                                                                             │
13. │         Parts: 1/1                                                                                             │
14. │         Granules: 10/10                                                                                        │
15. │       PrimaryKeyExpand                                                                                         │
16. │         Description: Selects all granules that intersect by PK values with the previous skip indexes selection │
17. │         Parts: 1/1                                                                                             │
18. │         Granules: 10/10                                                                                        │
19. │       Ranges: 1                                                                                                │
    └────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

nihalzp:
Also, let's add the queries from the review to the tests so that future changes does not accidentally regress it.

src/Interpreters/InterpreterExplainQuery.cpp
```diff
@@ -745,6 +747,16 @@ QueryPipeline InterpreterExplainQuery::executeImpl()
 
             break;
         }
+        case ASTExplainQuery::WhatIf:
+        {
+            const auto & query_ast = ast.getExplainedQuery();
+            if (!dynamic_cast<const ASTSelectWithUnionQuery *>(query_ast.get()))
+                throw Exception(ErrorCodes::INCORRECT_QUERY, "Only SELECT is supported for EXPLAIN WHATIF query");
+
+            auto whatif_result = WhatIfIndexEstimator::run(query_ast, query_context, ast.getSettings());
```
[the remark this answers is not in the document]
yariks5s: EXPLAIN WHATIF stores its inline option in ASTExplainQuery::ast_settings (read by ast.getSettings() in the WhatIf case), not in ASTQueryWithOutput::settings_ast. InterpreterSetQuery::applySettingsFromQuery only consumes the latter, so it doesn't touch empirical. Verified on this PR's binary: EXPLAIN WHATIF empirical = 0 SELECT ... works as documented (no UNKNOWN_SETTING).

alexey-milovidov:

src/Storages/MergeTree/WhatIfIndexEstimator.cpp
```diff
@@ -0,0 +1,869 @@
...
+    /// Text indexes need a tokenized block layout the empirical pipeline doesn't build      todo.
+    if (index_desc.type == "text")
+    {
+        result.status = WhatIfIndexEstimator::IndexResult::NotApplicable;
+        result.not_applicable_reason = "EXPLAIN WHATIF does not yet support empirical estimation for text indexes";
+        return result;
+    }
+
+    const auto & filter_dag = read_step->getFilterActionsDAG();
+    if (!filter_dag)
```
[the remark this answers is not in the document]
yariks5s: That's a todo.

src/Interpreters/HypotheticalIndexStore.h
```diff
@@ -0,0 +1,39 @@
...
+    std::vector<IndexDescription> getForTable(const StorageID & table_id) const;
+
+    struct Entry
+    {
+        StorageID table_id;
+        IndexDescription index;
+    };
+    std::vector<Entry> getAll() const;
+
+    bool empty() const;
```
nihalzp: This is never called anywhere; can be removed.

src/Parsers/CommonParsers.h
```diff
@@ -583,6 +584,7 @@ namespace DB
     MR_MACROS(VIEW, "VIEW") \
     MR_MACROS(VISIBLE, "VISIBLE") \
     MR_MACROS(WATCH, "WATCH") \
+    MR_MACROS(WHATIF, "WHATIF") \
```
nihalzp: Since the list seems to be mostly alphabetically ordered, this entry should be bit later.

src/Interpreters/InterpreterHypotheticalIndexQuery.cpp
```diff
@@ -0,0 +1,195 @@
...
+{
+    extern const SettingsBool allow_suspicious_indices;
+}
+
+namespace
+{
+
+void checkSuspiciousIndex(const ASTPtr & expression_list)
+{
+    const auto * function = expression_list ? typeid_cast<const ASTFunction *>(expression_list.get()) : nullptr;
```
nihalzp: Can we reuse the `checkSuspiciousIndices` from `src/Storages/MergeTree/MergeTreeData.cpp`? We may need to move it into an appropriate header file; upside is that there will be no divergence in the future.

src/Interpreters/InterpreterHypotheticalIndexQuery.cpp
```diff
@@ -0,0 +1,195 @@
...
+    /// Empirical estimation reads the index's columns, so require column-level
+    /// SELECT — otherwise a user with table-level access could infer a restricted
+    /// column's distribution from the reported skip ratio.
+    if (index_desc.expression)
+        context->checkAccess(AccessType::SELECT, table_id, index_desc.expression->getRequiredColumns());
+
+    /// Reject unsupported types before the type-specific validator can throw a confusing error.   todo.
+    if (index_desc.type == "text" || index_desc.type == "vector_similarity")
+        throw Exception(
+            ErrorCodes::NOT_IMPLEMENTED,
```
nihalzp: Maybe we can do something like:

```cpp
auto index_helper = MergeTreeIndexFactory::instance().get(index_desc);
if (index_helper->isTextIndex() || index_helper->isVectorSimilarityIndex())
    throw Exception(ErrorCodes::NOT_IMPLEMENTED,
```

Additionally, maybe something worth considering: should we do it the other way by allowing a whitelist and rejecting everything else? This way in the future if someone adds a new index, it will not break hypothetical index in an unpredictable way.

src/Storages/MergeTree/WhatIfIndexEstimator.cpp
```diff
@@ -0,0 +1,869 @@
...
+    /// so the condition can pick up a standalone conjunct out of a mixed AND/OR
+    ActionsDAGWithInversionPushDown predicate_dag(filter_dag->getOutputs().front(), context);
+    const ActionsDAG::Node * predicate = predicate_dag.predicate;
+
+    MergeTreeIndexConditionPtr condition;
+    try
+    {
+        condition = index_helper->createIndexCondition(predicate, context);
+    }
+    catch (...)
```
nihalzp: Let's try to only catch `DB::Exception`.

src/Storages/MergeTree/WhatIfIndexEstimator.cpp
```diff
@@ -0,0 +1,869 @@
...
+    writeString(fmt::format("  parts:       {}\n", baseline_parts), out);
+    writeString(fmt::format("  marks:       {}\n", baseline_marks), out);
+    if (baseline_est_bytes > 0)
+    {
+        double gb = static_cast<double>(baseline_est_bytes) / (1024.0 * 1024.0 * 1024.0);
+        if (gb >= 0.01)
+            writeString(fmt::format("  est_bytes:   {:.2f} GB\n", gb), out);
+        else
+            writeString(fmt::format("  est_bytes:   {} B\n", baseline_est_bytes), out);
+    }
```
nihalzp: Could we reuse `ReadableSize` from `#include <Common/formatReadable.h>`?

src/Storages/MergeTree/WhatIfIndexEstimator.cpp
```diff
@@ -0,0 +1,869 @@
...
+
+        if (baseline_marks > 0 && baseline_est_bytes > 0)
+        {
+            UInt64 hypo_bytes = static_cast<UInt64>(
+                static_cast<double>(baseline_est_bytes) * static_cast<double>(idx.estimated_marks) / static_cast<double>(baseline_marks));
+            double gb = static_cast<double>(hypo_bytes) / (1024.0 * 1024.0 * 1024.0);
+            if (gb >= 0.01)
+                writeString(fmt::format("  est_bytes:    {:.2f} GB\n", gb), out);
+            else
+                writeString(fmt::format("  est_bytes:    {} B\n", hypo_bytes), out);
```
nihalzp: Same as before.

src/Storages/MergeTree/WhatIfIndexEstimator.cpp
```diff
@@ -0,0 +1,869 @@
...
+        return result;
+    }
+
+    /// CREATE checked these columns, but the scan reads them now, so re-check SELECT against
+    /// current grants, a grant revoked since CREATE should deny the estimate
+    context->checkAccess(AccessType::SELECT, data.getStorageID(), index_helper->getColumnsRequiredForIndexCalc());
+
+    /// Text indexes need a tokenized block layout the empirical pipeline doesn't build      todo.
+    if (index_desc.type == "text")
+    {
```
nihalzp: Same as before. We can use `index_helper->isTextIndex()`.

nihalzp (commented):
Previous issues are resolved. Looks good. I have left a few comments.

---
The following query crashes:
```sql
DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    a UInt64,
    b UInt64,
    PROJECTION p
    (
        SELECT a, b
        ORDER BY b
    )
)
ENGINE = MergeTree
ORDER BY a
SETTINGS index_granularity = 100, index_granularity_bytes = 0, min_bytes_for_wide_part = 0;

INSERT INTO test SELECT number, number FROM numbers(10000);

CREATE HYPOTHETICAL INDEX idx_b ON test (b) TYPE minmax GRANULARITY 1;

EXPLAIN WHATIF
SELECT a
FROM test
WHERE b = 5;
```
---

The following query has logical error:

```sql
DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    a UInt64
)
ENGINE = MergeTree
ORDER BY a;

CREATE HYPOTHETICAL INDEX idx0 ON test (a) TYPE minmax GRANULARITY 0;
```

---

I got this exception "EXPLAIN WHATIF requires a query reading from a MergeTree family table." on an empty `MergeTree` table. But with populated table this is okay.

```sql
DROP TABLE IF EXISTS test;

CREATE TABLE test
(
    a UInt64,
    b UInt64
)
ENGINE = MergeTree
ORDER BY a;

CREATE HYPOTHETICAL INDEX idx_b ON test (b) TYPE minmax GRANULARITY 1;

EXPLAIN WHATIF
SELECT *
FROM test
WHERE b = 42;
```

src/Storages/MergeTree/WhatIfIndexEstimator.cpp
```diff
@@ -0,0 +1,869 @@
...
+        InterpreterSelectQueryAnalyzer interpreter(select_query_copy, local_context, query_options);
+        plan_context = interpreter.getContext();
+        plan = std::move(interpreter).extractQueryPlan();
+    }
+    else
+    {
+        InterpreterSelectWithUnionQuery interpreter(select_query_copy, local_context, query_options);
+        plan_context = interpreter.getContext();
+        interpreter.buildQueryPlan(plan);
+    }
```
nihalzp, line 698: This is the old analyzer path. I noticed that the added tests have `no-replicated-database` which means old analyzer path is never run. It will be nice to have some very basic tests for old analyzer just to make sure it works there.

src/Storages/MergeTree/WhatIfIndexEstimator.cpp
```diff
@@ -0,0 +1,869 @@
...
+    if (!node)
+        return;
+    if (node->type == ActionsDAG::ActionType::INPUT)
+        out.insert(node->result_name);
+    for (const auto * child : node->children)
+        collectFilterInputColumns(child, out);
+}
+
+/// An OR with the candidate's column on one side and another column on the other. The real
+/// read can combine them (use_skip_indexes_for_disjunctions) we can't, so we bail on those    todo.
```
nihalzp:
```suggestion
/// TODO(yariks5s): An OR with the candidate's column on one side and another column on the other. The real
/// read can combine them (use_skip_indexes_for_disjunctions) we can't, so we bail on those.
```

Let's write the TODOs in this form so that it's easily relocatable; it matches existing TODO patterns like `TODO (alesap)`, `TODO(serxa)`, `TODO(nihalzp)`.

nihalzp (approved):

src/Storages/MergeTree/WhatIfIndexEstimator.cpp
```diff
@@ -0,0 +1,892 @@
...
+    }
+
+    if (!has_any_stats)
+        return false;
+
+    auto estimator = builder.getEstimator();
+    if (!estimator)
+        return false;
+
+    auto profile = estimator->estimateRelationProfile(metadata, filter_node);
```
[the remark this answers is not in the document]
yariks5s: > EXPLAIN WHATIF needs to honor row policies

it works the same way as `EXPLAIN indexes=1` does. The 'don't count row policies' is really debatable.

PR #104609, opened 2026-05-11
Add regression test for CASE over splitByString Array(String) result (#78352)

### LINKED ISSUE #78352: splitByString with CASE
### Company or project name

_No response_

### Describe what's wrong

This does not work and will report an error.
```sql
CREATE TABLE test_split (
    id Int32,
    input String
) ENGINE = Memory;
INSERT INTO test_split VALUES
(1, 'a#b#c#d');
SELECT
    id,
    CASE id
        WHEN 1 THEN splitByString('#', input)
        ELSE splitByString('', '')
    END AS result
FROM test_split;
```

But `SELECT splitByString('#', input) FROM test_split` and `SELECT splitByString('', '') FROM test_split` work normally.

Seems that CASE is having trouble when dealing with splitByString?

Yes

### How to reproduce

```
CREATE TABLE test_split (
    id Int32,
    input String
) ENGINE = Memory;
INSERT INTO test_split VALUES
(1, 'a#b#c#d');
SELECT
    id,
    CASE id
        WHEN 1 THEN splitByString('#', input)
        ELSE splitByString('', '')
    END AS result
FROM test_split;
```

### Expected behavior

Ouput ['a','b','c','d']

Code: 36. DB::Exception: Unexpected type Array(String) in function 'transform'. (BAD_ARGUMENTS) (version 25.1.1.1807 (official build))

### Additional context

_No response_

groeneai:
Add a stateless regression test for [issue #78352](https://github.com/ClickHouse/ClickHouse/issues/78352).

`CASE expr WHEN ... THEN ... ELSE ... END` over an `Array(String)`-returning function (`splitByString`) used to be lowered to `transform`, which does not support array results, and the query failed with `Code: 36. DB::Exception: Unexpected type Array(String) in function 'transform'. (BAD_ARGUMENTS)`. The materialized-column path was fixed in 25.3 and the all-constant path was fixed in 26.3. This test pins both code paths and both analyzer modes (`enable_analyzer = 0` and `enable_analyzer = 1`).

Test added at `tests/queries/0_stateless/04227_case_with_expression_array_result_78352.sql`. Verified locally with `--test-runs 30` (30/30 passes with randomized settings) against current master.

Closes #78352.

cc @den-crane (who requested the test on the issue).

...

### DIFF
--- a/tests/queries/0_stateless/04227_case_with_expression_array_result_78352.reference
+++ b/tests/queries/0_stateless/04227_case_with_expression_array_result_78352.reference
@@ -1,0 +1,16 @@
+memory table, analyzer=1:
+1	['a','b','c','d']
+2	[]
+3	[]
+memory table, analyzer=0:
+1	['a','b','c','d']
+2	[]
+3	[]
+materialize subquery, analyzer=1:
+1	['a','b','c','d']
+materialize subquery, analyzer=0:
+1	['a','b','c','d']
+all-constant subquery, analyzer=1:
+1	['a','b','c','d']
+all-constant subquery, analyzer=0:
+1	['a','b','c','d']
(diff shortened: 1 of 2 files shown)


groeneai:
cc @alexey-milovidov — could you review this small test-only PR? It locks in the fix @den-crane confirmed on issue #78352 (CASE over `splitByString` Array(String) result; the column path was fixed in 25.3, the constant path in 26.3). Test exercises both paths plus both analyzer modes. Verified `--test-runs 30` with randomized settings, 30/30 passes against current master.

alexey-milovidov (approved):

PR #100966, opened 2026-03-27
Make SHOW CREATE TABLE and DESCRIBE TABLE consistent with TEMPORARY tables

alexey-milovidov:
`SHOW CREATE TABLE t` and `DESCRIBE TABLE t` were inconsistent when both a permanent and a temporary table with the same name existed:
- `SHOW CREATE TABLE t` showed the permanent table (used `ResolveOrdinary`)
- `DESCRIBE TABLE t` showed the temporary table (used `ResolveAll`)

Additionally, `DESCRIBE TEMPORARY TABLE t` was a syntax error while `SHOW CREATE TEMPORARY TABLE t` worked.

This PR fixes both issues:
1. `SHOW CREATE TABLE t` now uses `ResolveAll`, preferring the temporary table when no database is qualified — matching `DESCRIBE TABLE` behavior.
2. `DESCRIBE TEMPORARY TABLE t` is now valid syntax.

Closes https://github.com/ClickHouse/ClickHouse/issues/14004

### Changelog category (leave one):
- Backward Incompatible Change

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
`SHOW CREATE TABLE t` now prefers the temporary table when both a permanent and a temporary table named `t` exist and no database is specified, matching the existing behavior of `DESCRIBE TABLE`. Additionally, `DESCRIBE TEMPORARY TABLE` syntax is now supported.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.82`

PR #100277, opened 2026-03-21
Add web terminal interface to clickhouse-server

alexey-milovidov:
Add a WebSocket-based terminal protocol (`webterminal`) that runs a pseudoterminal with embedded clickhouse-client functionality, accessible via a browser at `/webterminal`.

The implementation reuses the same `PtyClientDescriptorSet` and `ClientEmbeddedRunner` infrastructure as the SSH protocol, but connects via WebSocket instead of SSH. The user is authenticated through the same `Session` machinery as the HTTP protocol, but credentials are sent in-band over the established WebSocket connection (in the first JSON message) instead of as part of the upgrade request, so they never appear in URLs, access logs, or browser history.

The HTML page uses xterm.js for terminal emulation and connects to the server via WebSocket, bridging data bidirectionally between the browser and the PTY.

### Changelog category (leave one):
- Experimental Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Added an experimental web terminal interface at `/webterminal` that provides an interactive `clickhouse-client` session in the browser over WebSocket. Disabled by default; enable with the `allow_experimental_webterminal` server setting.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

Navigate to `/webterminal` on any ClickHouse HTTP port to open an interactive terminal session in the browser. Authenticate with ClickHouse user credentials. The terminal supports the full clickhouse-client experience including syntax highlighting, autocompletion, and multi-line queries.

Companion docs sidebar PR: https://github.com/ClickHouse/clickhouse-docs/pull/6118

### Version info
- Merged into: `26.5.1.684`

PR #75728, opened 2025-02-07
[WIP] Use analyzer to calculate projections

novikd:
### Changelog category (leave one):

- Improvement

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):

Use analyzer to calculate projections.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #104637, opened 2026-05-11
Make CHECK TABLE prefer TEMPORARY tables for unqualified names (follow-up #100966)

groeneai:
Follow-up to #100966 on a request from @PedroTadim ([inline comment](https://github.com/ClickHouse/ClickHouse/pull/100966#discussion_r3217841914)) to keep the scoping precedence consistent across every command.

## Motivation

`InterpreterCheckQuery::execute` explicitly passed `Context::ResolveOrdinary` to `resolveStorageID`, so `CHECK TABLE t` for an unqualified name never consulted the session's `TEMPORARY` tables. Two visible consequences:

1. `CHECK TABLE t` on a `TEMPORARY`-only `Log` / `File` table failed with `UNKNOWN_TABLE`, even though those engines support `CHECK`:
   ```sql
   CREATE TEMPORARY TABLE t (x UInt64) ENGINE = Log;
   CHECK TABLE t;                       -- Code: 60. UNKNOWN_TABLE (no permanent `t`)
   ```

2. When both a `TEMPORARY` and a permanent table shared a name, `CHECK TABLE t` silently fell through to the permanent table, even though `SHOW CREATE TABLE`, `DESCRIBE TABLE`, `OPTIMIZE TABLE`, and `ALTER TABLE` already route the same unqualified name to the `TEMPORARY` one (the precedence #100966 made consistent).

Audit of how each "table inspection / mutation" command resolves an unqualified name on master:

| Command | Resolution before this PR |
| --- | --- |
| `SHOW CREATE TABLE` | `ResolveAll` (`TEMPORARY` first) — #100966 |
| `DESCRIBE TABLE` | `ResolveAll` (`TEMPORARY` first) — #100966 |
| `OPTIMIZE TABLE` | `ResolveAll` (default) |
| `ALTER TABLE` | `ResolveAll` (`tryResolveStorageID`, default) |
| `INSERT INTO` / `SELECT FROM` | `ResolveAll` (default) |
| **`CHECK TABLE`** | **`ResolveOrdinary` (permanent only)** ⚠️ |

This PR aligns `CHECK TABLE` with the rest by using the default `ResolveAll` precedence, the same one PR #100966 settled on for the table-inspection family.

## What changes

Only `InterpreterCheckQuery.cpp`:

```diff
-        /// Check specific table
-        auto table_id = context->resolveStorageID(*check_query, Context::ResolveOrdinary);
+        /// Check specific table. Use `ResolveAll` so that an unqualified name
+        /// prefers a `TEMPORARY` table over a permanent one of the same name,
+        /// matching the scoping precedence of `SHOW CREATE TABLE` and
+        /// `DESCRIBE TABLE` introduced in #100966.
+        auto table_id = context->resolveStorageID(*check_query);
```

After:

* `CHECK TABLE t` on a `TEMPORARY`-only `Log` / `File` table succeeds.
* On a `TEMPORARY` + permanent collision, `CHECK TABLE t` targets the temporary one; `CHECK TABLE db.t` still targets the permanent one (qualifying the name disables external lookup).
* `CHECK TABLE t` on a `TEMPORARY` `Memory` table now surfaces `NOT_IMPLEMENTED` from the engine, instead of `UNKNOWN_TABLE` from the catalog — a more accurate error message.

## Verification

* Built locally with `clang-21`, debug.
* Added a new stateless test `04228_check_optimize_alter_temporary_scoping_100966.sql` that exercises `CHECK` / `OPTIMIZE` / `ALTER` scoping with a `TEMPORARY`/permanent collision.
* Without the one-line change: the test fails on the first assertion (`Code: 60. UNKNOWN_TABLE`) — confirmed by reverting, rebuilding, and re-running.
* With the change: 50/50 runs pass under randomized server and `MergeTree` settings.
* Existing `04064_describe_show_create_temporary_table` still passes (5/5).

## Out of scope (not changed)

Other `ResolveOrdinary` call sites I checked:

* `InterpreterDeleteQuery` (lightweight `DELETE`), `InterpreterUpdateQuery` (lightweight `UPDATE`) — `MergeTree`-only operations, temporary engines do not support them.
* `InterpreterWatchQuery` — live views, cannot be temporary.
* `InterpreterSystemQuery` — operates on permanent storage (replication / merges / etc.).
* `InterpreterShowCreateQuery` — already handled by #100966 (`TABLE` uses `ResolveAll`, `VIEW` / `DICTIONARY` keep `ResolveOrdinary`).

So `CHECK TABLE` was the lone outlier among commands that can semantically run against a temporary table.

cc @PedroTadim @alexey-milovidov

`CHECK TABLE t` now prefers a `TEMPORARY` table over a permanent one with the same name when no database qualifier is given, matching the precedence already used by `SHOW CREATE TABLE`, `DESCRIBE TABLE`, `OPTIMIZE TABLE`, and `ALTER TABLE`. Previously `CHECK TABLE t` skipped temporary tables entirely, so it failed with `UNKNOWN_TABLE` on a temporary `Log` or `File` table even though those engines support `CHECK`. Follow-up to [#100966](https://github.com/ClickHouse/ClickHouse/pull/100966).

### Version info
- Merged into: `26.5.1.552`

### DIFF
--- a/src/Interpreters/InterpreterCheckQuery.cpp
+++ b/src/Interpreters/InterpreterCheckQuery.cpp
@@ -418,8 +418,11 @@ BlockIO InterpreterCheckQuery::execute()
     bool is_table_name_in_output = false;
     if (const auto * check_query = query_ptr->as<ASTCheckTableQuery>())
     {
-        /// Check specific table
-        auto table_id = context->resolveStorageID(*check_query, Context::ResolveOrdinary);
+        /// Check specific table. Use `ResolveAll` so that an unqualified name
+        /// prefers a `TEMPORARY` table over a permanent one of the same name,
+        /// matching the scoping precedence of `SHOW CREATE TABLE` and
+        /// `DESCRIBE TABLE` introduced in #100966.
+        auto table_id = context->resolveStorageID(*check_query);
         auto table_check_task = std::make_shared<TableCheckTask>(table_id, check_query->getPartitionOrPartitionID(), context);
         worker_source = std::make_shared<TableCheckSource>(table_check_task, log);
         worker_source->addTotalRowsApprox(table_check_task->size());
--- a/tests/queries/0_stateless/04228_check_optimize_alter_temporary_scoping_100966.reference
+++ b/tests/queries/0_stateless/04228_check_optimize_alter_temporary_scoping_100966.reference
@@ -1,0 +1,11 @@
+CHECK on TEMPORARY-only Log table returns success:
+1
+CHECK on collision targets TEMPORARY Memory (rejects):
+OK
+OPTIMIZE on collision targets TEMPORARY Memory (rejects):
+OK
+ALTER on collision modifies TEMPORARY (z appears in DESCRIBE):
+world	UInt64
+z	UInt8
+ALTER on collision does not modify permanent (no z in system.columns):
+0
--- a/tests/queries/0_stateless/04228_check_optimize_alter_temporary_scoping_100966.sql
+++ b/tests/queries/0_stateless/04228_check_optimize_alter_temporary_scoping_100966.sql
@@ -1,0 +1,66 @@
+-- Test for https://github.com/ClickHouse/ClickHouse/pull/100966 follow-up.
+-- Verify CHECK, OPTIMIZE, and ALTER TABLE use the same scoping precedence
+-- as SHOW CREATE TABLE / DESCRIBE TABLE: an unqualified name resolves to the
+-- TEMPORARY table when one exists, preferring it over a same-named permanent.
+
+SET describe_compact_output = 1;
+
+-- ============================================================================
+-- CHECK TABLE: previously used Context::ResolveOrdinary and ignored TEMPORARY.
+-- ============================================================================
+
+-- (1) CHECK TABLE on a TEMPORARY-only table now succeeds (was UNKNOWN_TABLE).
+DROP TABLE IF EXISTS t SYNC;
+CREATE TEMPORARY TABLE t (x UInt64) ENGINE = Log;
+INSERT INTO t VALUES (1), (2), (3);
+
+SELECT 'CHECK on TEMPORARY-only Log table returns success:';
+CHECK TABLE t SETTINGS check_query_single_value_result = 1;
+
+DROP TABLE t;
+
+-- (2) When a TEMPORARY and a permanent table both exist (Memory + MergeTree),
+--     CHECK TABLE now targets the TEMPORARY one. Memory doesn't support CHECK,
+--     so the query errors out -- previously it would have succeeded against the
+--     permanent MergeTree.
+CREATE TABLE t (hello String) ENGINE = MergeTree ORDER BY hello;
+INSERT INTO t VALUES ('p1'), ('p2'), ('p3');
+CREATE TEMPORARY TABLE t (world UInt64) ENGINE = Memory;
+
+SELECT 'CHECK on collision targets TEMPORARY Memory (rejects):';
+CHECK TABLE t; -- { serverError NOT_IMPLEMENTED }
+SELECT 'OK';
+
+DROP TEMPORARY TABLE t;
+DROP TABLE t;
+
+-- ============================================================================
+-- OPTIMIZE TABLE: already used the default ResolveAll. Lock in.
+-- ============================================================================
+CREATE TABLE t (hello String) ENGINE = MergeTree ORDER BY hello;
+CREATE TEMPORARY TABLE t (world UInt64) ENGINE = Memory;
+
+SELECT 'OPTIMIZE on collision targets TEMPORARY Memory (rejects):';
+OPTIMIZE TABLE t; -- { serverError NOT_IMPLEMENTED }
+SELECT 'OK';
+
+DROP TEMPORARY TABLE t;
+DROP TABLE t;
+
+-- ============================================================================
+-- ALTER TABLE: already used tryResolveStorageID with default ResolveAll. Lock in.
+-- ============================================================================
+CREATE TABLE t (hello String) ENGINE = MergeTree ORDER BY hello;
+CREATE TEMPORARY TABLE t (world UInt64) ENGINE = Memory;
+
+ALTER TABLE t ADD COLUMN z UInt8;
+
+SELECT 'ALTER on collision modifies TEMPORARY (z appears in DESCRIBE):';
+DESCRIBE TEMPORARY TABLE t;
+
+SELECT 'ALTER on collision does not modify permanent (no z in system.columns):';
+SELECT count() FROM system.columns
+WHERE database = currentDatabase() AND table = 't' AND name = 'z';
+
+DROP TEMPORARY TABLE t;
+DROP TABLE t;


groeneai:
### Pre-PR validation gate (per TASK.md §Phase 4)

| Question | Answer |
| --- | --- |
| **(a) Deterministic repro?** | Yes. `CREATE TEMPORARY TABLE t (x UInt64) ENGINE = Log; CHECK TABLE t;` fails 100% of the time with `Code: 60. UNKNOWN_TABLE` on `master`. The collision case (`TEMPORARY` Memory + permanent `MergeTree`) also reproduces 100% of the time. |
| **(b) Root cause explained?** | `InterpreterCheckQuery::execute` (`src/Interpreters/InterpreterCheckQuery.cpp:422`) called `context->resolveStorageID(*check_query, Context::ResolveOrdinary)`. `ResolveOrdinary = ResolveGlobal \| ResolveCurrentDatabase` — by construction this skips `ResolveExternal`, so the session's `TEMPORARY` tables are never consulted. Every other command that operates on a bare table name (`SHOW CREATE TABLE`, `DESCRIBE TABLE`, `OPTIMIZE TABLE`, `ALTER TABLE`, `INSERT`, `SELECT`) uses the default `ResolveAll`. |
| **(c) Fix matches root cause?** | One-line change at the exact call site; flips the precedence flag from `ResolveOrdinary` to `ResolveAll`. Not a band-aid; directly fixes the resolution decision. |
| **(d) Test intent preserved? New test added?** | No existing tests changed. New test `04228_check_optimize_alter_temporary_scoping_100966.sql` covers `CHECK` / `OPTIMIZE` / `ALTER` on a `TEMPORARY`/permanent collision. Existing `04064_describe_show_create_temporary_table` still passes (5/5). |
| **(e) Demonstrated in both directions?** | Yes. Without the fix: the new test fails at the first assertion (`Code: 60. UNKNOWN_TABLE`) — confirmed by reverting, rebuilding, and re-running. With the fix: 50/50 runs pass under randomized server settings and `MergeTree` settings. |
| **(f) Fix is general, not a narrow patch?** | Yes. I audited all `ResolveOrdinary` call sites in `Interpreters/`: `InterpreterDeleteQuery` / `InterpreterUpdateQuery` are `MergeTree`-only operations (temporary engines don't support them), `InterpreterWatchQuery` operates on live views (cannot be temporary), `InterpreterSystemQuery` targets server-wide / permanent storage operations, and `InterpreterShowCreateQuery` was already handled by #100966 (keeps `ResolveOrdinary` for `VIEW`/`DICTIONARY` by design). `CHECK TABLE` was the lone outlier among commands that can semantically operate on a temporary table. |

Session: `cron:clickhouse-ci-task-worker:20260511-204500`.

groeneai:
cc @PedroTadim @alexey-milovidov @KochetovNicolai — could you review this? It's a 1-line follow-up to #100966 that aligns `CHECK TABLE`'s table-name resolution (`ResolveOrdinary` → default `ResolveAll`) with the precedence `SHOW CREATE TABLE`/`DESCRIBE TABLE`/`OPTIMIZE`/`ALTER` already use for `TEMPORARY` tables. Test demonstrates both directions (`UNKNOWN_TABLE` before, success after).

PedroTadim:
@groeneai do RENAME, EXCHANGE, ATTACH, and DETACH also follow the same scoping rules? Why the `01710_projections` test failed?

groeneai:
@PedroTadim — answering both questions in order.

## (b) Why `01710_projections` failed

The failing job is `Stateless tests (amd_debug, distributed plan, s3 storage, parallel)`. The test [auto-diagnosed itself](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104637&sha=a8124337106bfc56f016185a2bfa15bc77c17391&name_0=PR&name_1=Stateless%20tests%20%28amd_debug%2C%20distributed%20plan%2C%20s3%20storage%2C%20parallel%29) via `--diagnose-random-settings`:

```
Step 1: Re-running with the same randomized settings... Failed: 50, Passed: 0  → reproducible
Step 2: Re-running without randomized settings...        Failed: 0,  Passed: 68 → not reproducible
Culprit MergeTree settings: --index_granularity 1
```

Diff is a single trailing-digit precision wobble: `1.2352100180287104` vs `1.235210018028712`. With `index_granularity = 1` every row goes into its own granule, which changes the order in which `metric_log`'s `MergeSortingTransform`/`AggregatingSortedTransform` accumulates floating-point sums when `distributed plan + s3 storage` reshuffles the read order.

This is **not** related to this PR's diff — the only change here is a single `Context::ResolveOrdinary → ` (default `Context::ResolveAll`) flip in `InterpreterCheckQuery.cpp`, which has nothing to do with `01710_projections` (which doesn't use `CHECK TABLE`).

CIDB cross-PR (30d):
| PR | Hits | Last seen |
|---|---|---|
| #94359 | 18 | 2026-04-17 (single-day mass event, all `index_granularity 1`) |
| #75728 | 2 | 2026-05-06 (Fast test) |
| #94517 | 1 | 2026-05-12 (Fast test, arm_darwin) |
| **#104637** | 1 | 2026-05-11 (ours, just this run) |
| #0 (master) | 1 | 2026-04-29 |

Chronic flake under `index_granularity = 1`; isolated single hits across unrelated branches and one master hit. Suggest re-run.

## (a) RENAME / EXCHANGE / ATTACH / DETACH scoping audit

I read the four interpreters. Summary table extending the one in the PR body:

| Command | Code path | TEMPORARY handling | Aligns with PR #100966? |
|---|---|---|---|
| `SHOW CREATE TABLE` | `InterpreterShowCreateQuery.cpp:65-66` | `ResolveAll` (#100966) | ✅ |
| `DESCRIBE TABLE` | `InterpreterDescribeQuery.cpp:180-181` | `ResolveAll` (#100966) | ✅ |
| `OPTIMIZE TABLE` | `InterpreterOptimizeQuery.cpp:41` | default = `ResolveAll` | ✅ |
| `ALTER TABLE` | `InterpreterAlterQuery.cpp:382` (`tryResolveStorageID(alter)`) | default = `ResolveAll` | ✅ |
| `CHECK TABLE` | `InterpreterCheckQuery.cpp:422` | (was `ResolveOrdinary`) → default `ResolveAll` after this PR | ✅ (this PR) |
| **`RENAME TABLE`** | `InterpreterRenameQuery.cpp:116` calls `DatabaseCatalog::getDatabase(name)` directly — **never goes through `resolveStorageID`** | TEMPORARY tables not reachable. `RENAME TABLE temp_t TO x` (unqualified, temp only) fails with `UNKNOWN_TABLE` in `current_database`. When both temp and permanent exist with the same name, RENAME hits the **permanent** (opposite of the PR #100966 convention). | ⚠️ Diverges, but it's a design gap, not a one-line flip |
| **`EXCHANGE TABLES`** | Same code path as RENAME (`rename.exchange = true`) | Same as RENAME | ⚠️ Same as RENAME |
| **`ATTACH TABLE`** | `InterpreterCreateQuery.cpp:1543-1544` | Explicitly throws `SYNTAX_ERROR("ATTACH of TEMPORARY tables are not supported")` | ✅ Cannot apply — by design |
| **`DETACH TABLE`** | `InterpreterDropQuery.cpp:156-157` for explicit `DETACH TEMPORARY TABLE`, `InterpreterDropQuery.cpp:161-166` for unqualified | `DETACH TEMPORARY TABLE t` throws `SYNTAX_ERROR("DETACH of TEMPORARY tables are not supported")`. **But** unqualified `DETACH TABLE t` tries `ResolveExternal` first and, if a temp exists, sets `is_detached = true` on the temp. | ⚠️ Inconsistent within itself, but the *implicit* form already prefers TEMPORARY |

### What I think falls inside this PR and what doesn't

I'd argue this PR should stay **just the `CHECK TABLE` one-line fix** for these reasons:

1. **RENAME / EXCHANGE**: aligning them isn't a `Context::ResolveOrdinary → ResolveAll` flip — the interpreter doesn't go through `resolveStorageID` at all. It addresses tables by `(database, table)` pairs from the parser and calls `DatabaseCatalog::getDatabase()` directly. To make `RENAME TABLE t TO x` prefer a TEMPORARY `t`, we'd need to (i) detect TEMPORARY existence via `tryResolveStorageID(StorageID("", t), ResolveExternal)`, (ii) decide what the semantics of *renaming a temp* are (does the renamed table stay TEMPORARY in-session? what does the target name mean? cluster + replication paths don't apply to temps but the interpreter does enqueue `tryEnqueueReplicatedDDL`), (iii) gate the `assertTableDoesntExist(to)` check by which namespace `to` lives in. That's a feature design, not a scoping cleanup.

2. **ATTACH**: `ATTACH TABLE` of TEMPORARY tables is explicitly disallowed since forever. There's nothing to align — TEMPORARY tables come into being via `CREATE TEMPORARY TABLE` only.

3. **DETACH**: the explicit `DETACH TEMPORARY TABLE t` form is disallowed (SYNTAX_ERROR), but unqualified `DETACH TABLE t` already prefers a TEMPORARY `t` when one exists (`ResolveExternal` at `InterpreterDropQuery.cpp:163`). So the *implicit* scoping is already TEMPORARY-first. Symmetry with the new `DESCRIBE TEMPORARY TABLE` from #100966 would suggest also allowing `DETACH TEMPORARY TABLE t` — but again, that's a behavior addition (what does detaching a temp mean? semantically the session ends and the temp is gone, so DETACH doesn't have a natural lifecycle for it), not a scoping fix.

### Proposal

Keep this PR focused on `CHECK TABLE`. I can file follow-ups for the RENAME/EXCHANGE/DETACH semantic gaps if you confirm the desired behavior:

- **RENAME a TEMPORARY**: should `RENAME TABLE temp_t TO x` rename in-session (temp stays temp) or be rejected with a clearer error? (Today: silently hits permanent or fails with `UNKNOWN_TABLE`.)
- **DETACH a TEMPORARY**: keep current "implicit DETACH works, explicit DETACH TEMPORARY rejected" behavior, or also allow the explicit form for symmetry with `DESCRIBE TEMPORARY TABLE`?

Either way, those are separate PRs in the same direction as #100966 — happy to take them if you'd like.

(Re-running CI now — `Unit tests (tsan, function_prop_fuzzer)` is the chronic `Volnitsky.h:724` MSan flake tracked by #100277/#104472, also unrelated to this PR's diff.)

PedroTadim:
For `01710_projections` you can fix in another PR? This PR seems in a good shape and can be merged

PR #101034, opened 2026-03-28
Add inline insert data mode for native protocol

alexey-milovidov:
Add a setting `send_table_structure_on_insert_with_inline_data` and a `--inline-insert-data` command-line option for `clickhouse-client` to allow the server to parse inline INSERT data itself, avoiding the round-trip to send table structure back to the client.

This reuses the existing server-side inline data parsing machinery (used by the HTTP interface and async inserts) for the native TCP protocol, which improves performance for many small inserts.

Closes https://github.com/ClickHouse/ClickHouse/issues/91630

### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Add setting `send_table_structure_on_insert_with_inline_data` and `--inline-insert-data` client option to allow the server to parse inline INSERT data itself over the native protocol, avoiding the round-trip to receive table structure and improving performance for many small inserts.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

New setting `send_table_structure_on_insert_with_inline_data` (Bool, default `true`): when disabled, the server will parse inline data in INSERT queries itself instead of sending table structure and column defaults back to the client.

New `clickhouse-client` option `--inline-insert-data`: sends INSERT data as is in the query text instead of converting it to blocks in the native format. Automatically sets `send_table_structure_on_insert_with_inline_data = false`.

Example:
```bash
clickhouse-client --inline-insert-data --query "INSERT INTO t VALUES (1, 'hello'), (2, 'world')"
```

### Version info
- Merged into: `26.5.1.144`

PR #104640, opened 2026-05-11
Extend 04064_inline_insert_data and randomize send_table_structure_on_insert_with_inline_data (follow-up to #101034)

groeneai:
Follow-up to #101034. @PedroTadim asked in PR review for:
1. an `INSERT SELECT` corner case to be added to `tests/queries/0_stateless/04064_inline_insert_data.sh`,
2. a case with `async_insert` disabled, and
3. randomization of the new `send_table_structure_on_insert_with_inline_data` setting so CI exercises both the legacy and the new server-side inline parsing path.

This PR addresses all three.

#### `tests/queries/0_stateless/04064_inline_insert_data.sh` / `.reference`

- Two `INSERT ... SELECT` cases — one with `--inline-insert-data`, one with `send_table_structure_on_insert_with_inline_data = 0`. SELECT-driven inserts have no inline data, so the new path is bypassed via the `!insert->select` guard in `is_inline_insert_data`. The cases verify that requesting inline insert data mode does not break SELECT-driven INSERTs.
- Two `--async_insert 0` cases (one via the CLI flag, one via the setting). The user's explicit choice takes precedence over `async_insert` on the client side, so the data is parsed by the server inline regardless of the async insert toggle.

The new rows extend the existing `SELECT * ... ORDER BY x` dump (rows 8..13) without changing any prior assertions or the negative-test section.

#### `tests/clickhouse-test`

Add `send_table_structure_on_insert_with_inline_data` to `SettingsRandomizer.settings` (50/50 bool). This makes CI exercise both the legacy "client converts inline data to native blocks" path and the new "server parses inline data" path that #101034 introduced.

Tests can opt out the same way as for any other randomized setting:
- per-query: `SETTINGS send_table_structure_on_insert_with_inline_data = 1`
- per-test: `-- Tags: no-random-settings`

### Verification

Test run locally with the patched binary:
- `tests/clickhouse-test 04064_inline_insert_data` — OK.
- `tests/clickhouse-test --test-runs 50 04064_inline_insert_data` — 50/50 OK with full randomization.
- All six semantic paths (legacy vs new, `VALUES` vs `SELECT` vs format-driven, `async_insert={0,1}`) verified to insert rows correctly via direct `clickhouse-client` queries.

Not required.

### Version info
- Merged into: `26.7.1.784` (included in `26.7` and later)


groeneai:
cc @PedroTadim @alexey-milovidov — follow-up to #101034 addressing @PedroTadim's review requests: extends `04064_inline_insert_data` with the `INSERT SELECT` corner case and an `async_insert = 0` case, and adds `send_table_structure_on_insert_with_inline_data` to the random-settings pool. Test passes 50/50 locally with full randomization. Session id: cron:clickhouse-ci-task-worker:20260511-211500.

groeneai:
**Pre-PR validation (adapted — this is a test-coverage / randomization follow-up, not a bug fix):**

a) **Deterministic repro?** N/A — no failure was reproduced. This PR adds coverage for corner cases @PedroTadim explicitly listed; it does not fix a known bug.

b) **Root cause explained?** N/A — no defect. The motivation is reviewer-requested follow-up to PR #101034: exercise the inline-insert-data path with `INSERT SELECT` (`!insert->select` guard), `async_insert = 0` (CLI flag and setting precedence), and 50/50 randomization of `send_table_structure_on_insert_with_inline_data` so master CI exercises both client-side block conversion and server-side parsing.

c) **Fix matches root cause?** N/A — see (b).

d) **Test intent preserved?** Yes. Existing rows 1..7 in the dump and both negative `grep -c` checks at the end of the test are unchanged. New rows 8..13 are appended before the existing `SELECT * ... ORDER BY x` to extend coverage, not narrow the existing assertions.

e) **Both directions demonstrated?** Verified by direct queries that all six semantic paths (`{legacy,inline_flag,inline_setting} × {VALUES, SELECT, format-driven, async_insert={0,1}}`) insert the correct row content. `tests/clickhouse-test --test-runs 50 04064_inline_insert_data` passed 50/50 with full randomization (and therefore with the new `send_table_structure_on_insert_with_inline_data` randomized 50/50 across runs).

f) **Generalized / no narrow patch?** Yes for the randomization side: adding the setting to `SettingsRandomizer.settings` exercises the new path uniformly across the entire stateless suite, not just `04064`. Tests that have legitimate reasons to depend on a fixed value can opt out via `SETTINGS send_table_structure_on_insert_with_inline_data = ...` or `-- Tags: no-random-settings` as usual.

Session id: cron:clickhouse-ci-task-worker:20260511-211500.

groeneai:
Followed up on the PR-caused failures from the previous push. The CI's `--diagnose-random-settings` correctly identified `--send_table_structure_on_insert_with_inline_data 0` as the culprit for all five stateless tests. Investigation showed they are all real behavior differences between the legacy "client converts inline data to native blocks" path and the new server-side parsing path (#101034); the tests themselves were never written to exercise that dimension.

**Findings per test (in increasing severity):**

1. `04027_json_input_format` — `clientError 117` annotation no longer matches when the server parses inline data, because the same `Ambiguous field (...)` exception is then raised as a `serverError`. Either side correctly detects the ambiguity, so this is just a too-strict annotation. **Fix is semantic: change `-- { clientError 117 }` to `-- { error 117 }` (9 occurrences)** — `error N` is accepted from either side (`TestHint.cpp` lines 92, 185-189; `ClientBase.cpp` lines 2854-2882). This is a genuine improvement in test coverage.

2. `00753_system_columns_and_system_tables_long` — `system.tables.total_bytes` for the `Memory` engine differs (e.g. 130 → 256/4096 in different CI runs) because the inline path builds the in-memory block with a different layout than the legacy path. The test asserts exact byte counts; **pinned the legacy path** via `Random settings limits: send_table_structure_on_insert_with_inline_data=(1, 1)`. The block-layout delta is observable and worth a separate follow-up.

3. `03229_async_insert_alter` — `system.asynchronous_insert_log.data_kind` changes from `Preprocessed` to `Parsed` because the inline path bypasses the async insert flow on the client (`is_async_insert_with_inlined_data` becomes false in `ClientBase`). **Pinned the legacy path** so the async-insert log assertions remain meaningful.

4. `01173_transaction_control_queries` — first INSERT inside a `begin transaction;` fails with `Async inserts inside transactions are not supported. (NOT_IMPLEMENTED)` when the inline path is on (default `async_insert=1` interaction). This is a feature-interaction issue: the inline path makes async insert happen server-side even though the client thinks it's bypassing async. **Pinned the legacy path** so the test remains focused on transaction-control semantics.

5. `04045_values_insert_trailing_comment` — **silent data loss / multi-query parsing bug.** In a multi-query stream, an `INSERT ... VALUES` followed by a trailing SQL comment and then `;` causes all subsequent queries to be silently swallowed (no error, no output). Confirmed with a minimal repro: `clickhouse-client --send_table_structure_on_insert_with_inline_data=0 --multiquery 'CREATE TABLE t (...); INSERT INTO t VALUES (1) -- comment\n; SELECT count() FROM t;'` prints nothing instead of `1`. This is the most concerning of the five — **pinned the legacy path here too**, but the underlying multi-query parsing in the inline path is worth a separate look.

**Pre-PR validation gate** (per `TASK.md` Phase 4 Step 9):

a. **Deterministic repro?** Yes — each failure reproduces 100% with `--send_table_structure_on_insert_with_inline_data 0` (confirmed both locally and by CI's `--diagnose-random-settings` minimizer).

b. **Root cause explained?** Yes — the two paths (`is_inline_insert_data` true vs false in `ClientBase::processSingleQuery`) take different code routes through the client and server; each test failure traces to a specific observable behavior difference between those routes (`clientError`↔`serverError`, `Preprocessed`↔`Parsed`, async-insert vs not in a transaction, block layout, multi-query parsing).

c. **Fix matches root cause?** Mixed: for `04027` the `error N` annotation directly fixes the mismatch (either side is valid). For the other four the per-test pin is honest technical debt — the underlying deltas in the inline path are real and would benefit from separate engineering work; this PR keeps each test focused on what it was originally written for.

d. **Test intent preserved?** Yes — none of the pinned tests had inline-insert-data as part of their original scope, and the legacy path remains fully exercised. The `04027` annotation change actually strengthens the test (now accepts the error from either side).

e. **Both directions demonstrated?** Yes — for `04027` verified `[ OK ]` with both setting=0 and setting=1 locally (the `error 117` annotation accepts either). For the pinned tests, the `Random settings limits` mechanism clamps the randomizer to `(1, 1)` so they never see setting=0 in CI; the legacy path is what `tests/clickhouse-test` was already running before this PR.

f. **Fix is general, not a narrow patch?** Mostly — `04027` is a general fix. The four pins are intentionally narrow; the broader fixes (especially #5, the multi-query data-loss case) belong in separate PRs that harden the inline path.

Suggested follow-ups for the team to triage (I am not filing these without approval, but flagging here):
- `Memory`/`Block` layout delta with inline parsing (~130 → ~256+ bytes) — observable in `system.tables`.
- Multi-query trailing-comment parsing in inline path — **silent data loss**, highest priority.
- Async insert getting triggered server-side even when `is_inline_insert_data=true` on the client, which then conflicts with transactions.
- `asynchronous_insert_log.data_kind=Parsed` is correct semantically; whether docs/dashboards expect `Preprocessed` for inline INSERTs is a UX question.

Session id: cron:clickhouse-ci-task-worker:20260511-234500

PedroTadim:
@groeneai many tests failed, check

groeneai:
@PedroTadim Thanks — investigated the "many tests failed" report.

Quick context: the CI report you saw is from the **previous** commit `9192ee6b769` (sha pinned in the report URL). The fix-up I pushed at 00:08 UTC (`7aa93649e08`) is still in `action_required` and CI hasn't run on it yet; once it does, the previously-pinned five tests should be quiet.

That said, the previous CI's diagnosis (`--diagnose-random-settings`) surfaced **four more** tests that hit `--send_table_structure_on_insert_with_inline_data 0` as the culprit — they weren't in my first batch because the failure rate is probabilistic and they slipped through the first reruns. I've now pushed a second fix-up commit (`b3924166117`) that addresses each one:

| Test | Fix | Why |
| --- | --- | --- |
| `04029_json_columns_input_format` | `clientError 117` → `error 117` (4 occurrences) | Same pattern as `04027`: both paths correctly detect `Ambiguous field`, just different error attribution. Accept either side. |
| `01286_constraints_on_default` | Pin `send_table_structure_on_insert_with_inline_data=(1,1)` | The test relies on the legacy path **terminating the connection** on `VIOLATED_CONSTRAINT`. The inline path keeps the connection alive, so the next `CREATE TEMPORARY TABLE` fails with `TABLE_ALREADY_EXISTS`. |
| `01501_clickhouse_client_INSERT_exception` | Pin `=(1,1)` | The test `grep`s for the client-side prefix `data for INSERT was parsed from query`. The inline path moves parsing server-side and the prefix is no longer emitted, so `grep -F -c` returns 0 instead of 1. |
| `03667_insert_columns_description` | Pin `=(1,1)` | The test asserts column `DEFAULT`s apply to `INSERT INTO TABLE FUNCTION remote(...) VALUES (NULL)` / `file(...) VALUES (NULL)`. The inline path doesn't propagate calling-side column descriptions to the receiving INSERT, so `DEFAULT 7` is dropped (`7` → `0`, `4 77` → `4 0`). |

**Pre-PR validation gate (this commit):**
- (a) **Deterministic repro:** Yes. With `disable_async_inserts.xml` server config + `--client-option send_table_structure_on_insert_with_inline_data=0`, each of the three pinned tests fails 100% of the time; `04029` failed deterministically the same way before the `error 117` change.
- (b) **Root cause explained:** Each row in the table above traces the behavior delta to a specific code path in PR #101034's inline-server-parsed implementation.
- (c) **Fix matches root cause:** `04029` gets the general fix (accept either side) — same justification as `04027` in the prior commit. The three pins are honest technical debt: each test's purpose has nothing to do with which side parses inline data, so the targeted pin preserves test intent.
- (d) **Test intent preserved:** Yes. `04029` still catches `Ambiguous field` (just from either side now). The three pins keep each test focused on its original concern (constraint-violation connection lifetime, client exception wording, column-description propagation).
- (e) **Both directions demonstrated:** setting=0 → 3 pinned tests FAIL, `04029` PASSES (general fix); setting=1 → all four PASS. 50 runs each under full randomization: 200/200 OK.
- (f) **Fix is general:** `04029` yes (same as `04027`). The three pins are NOT general — each represents a real behavior delta in `send_table_structure_on_insert_with_inline_data=0` that deserves a follow-up bug-triage by the inline-insert-data team. Suggested follow-ups:
  1. `01286`: should `VIOLATED_CONSTRAINT` from a server-parsed inline INSERT terminate the connection the same way the legacy client-parsed path does?
  2. `01501`: should `clickhouse-client` continue to emit `data for INSERT was parsed from query` when the server now parses the inline data? (probably the prefix should be re-worded or dropped explicitly)
  3. `03667`: should the inline path propagate column descriptions when targeting a `remote(...)` / `file(...)` table function? Right now `DEFAULT` is dropped — this is **silent data loss** and IMO the most concerning of the four.

If any of these behavior deltas are actually intentional, the pin stays and we add a code comment. If any is a bug, the pin can be removed once it's fixed.

Branch: `groeneai/inline-insert-data-tests-101034-followup`
Commit: https://github.com/ClickHouse/ClickHouse/pull/104640/commits/b3924166117

Session: cron:clickhouse-ci-task-worker:20260512-084500

groeneai:
@PedroTadim @alexey-milovidov — third fix-up pushed (commit `e8c65bc811d`).

The CI on `b3924166117` flagged two more tests via `--diagnose-random-settings`, both with the same `--send_table_structure_on_insert_with_inline_data 0` culprit. I reproduced both deterministically against a local debug server with `disable_async_inserts.xml` to match the CI profile.

| Test | Fix | Why |
|---|---|---|
| `02270_stdin_with_query_or_infile_data` | pin (`Random settings limits: send_table_structure_on_insert_with_inline_data=(1, 1)`) | Test combines inline `VALUES (24)` with stdin `(42)` and inline `FROM INFILE` with stdin. With `setting=0` the server's inline-data parser sees both an inline `Values (24)` and an external stdin/infile stream and explicitly errors `NOT_IMPLEMENTED`: *"Processing inline insert data with both inlined and external data (from stdin or infile) is not supported."* The legacy path silently picks one source (stdin, matching the reference `42`). The test is specifically about the legacy inline+stdin combination. |
| `02908_enum_exact_cast` | pin (`Random settings limits: send_table_structure_on_insert_with_inline_data=(1, 1)`) | Test distinguishes `INSERT VALUES (NULL)` into a non-nullable `Enum` raising `clientError TYPE_MISMATCH` (sync, `async_insert=0`) from `serverError TYPE_MISMATCH` (async, `async_insert=1`). With `setting=0` the client stops parsing inline `VALUES` and the server-side parser raises the error in both cases, so the sync `clientError` annotation becomes a `serverError`. The test is about the client-vs-server error-attribution boundary, which the new path collapses. |

**Pre-PR validation gate**
- (a) **Deterministic repro?** ✅ Both tests fail 100% with `--send_table_structure_on_insert_with_inline_data=0` and no other randomization. With `=1`, both pass.
- (b) **Root cause explained?** ✅ See per-test rows above.
- (c) **Fix matches root cause?** Honest technical debt, same shape as the 8 pins already in this PR. Both tests document the legacy semantic they depend on; the pin is the narrowest correct fix.
- (d) **Test intent preserved?** ✅ Neither test was written to exercise the `send_table_structure_on_insert_with_inline_data` axis. `02270` is about inline+stdin/INFILE; `02908` is about Enum NULL handling and the sync-vs-async error-attribution boundary.
- (e) **Both directions demonstrated?** ✅ See (a). Plus 50 full-randomization runs each (200 executions total): 100/100 OK.
- (f) **Fix is general?** Per-test pin; the underlying behavior delta is broad but not test-fixable in general.

**The other two `Fast test (arm_darwin)` failures** on `b3924166117` (`00479_date_and_datetime_to_number`, `03307_progress_http_asap`) both hit at `2026-05-12 11:17:41` — same second, same job — with `Reason: server died, ConnectionRefusedError [Errno 61]` during `_cleanup_database`. That's a macOS-only cascade from one server-died event; CIDB confirms 1 hit each in 30 days, both isolated to this PR. Not setting-related.

---

**Reflection on the long tail.** With this commit the running total is **3 rounds, 11 tests pinned** (plus 2 with general `clientError → error` annotation fixes). Each CI run probabilistically surfaces a new subset because `--diagnose-random-settings` explores a different randomized seed. Looking across the now-pinned set, the behavior deltas cluster into clear groups:

| Group | Tests | What changes |
|---|---|---|
| Connection lifetime | `01286` | `VIOLATED_CONSTRAINT` keeps the connection alive instead of terminating it |
| Client-vs-server error attribution | `04027`, `04029`, `02908` | Errors flip from `clientError` to `serverError`; general fix possible for ambiguous-field/TYPE_MISMATCH cases |
| Client-emitted prefix wording | `01501` | `data for INSERT was parsed from query` prefix disappears |
| Inline + stdin/INFILE combination | `02270` | New path explicitly `NOT_IMPLEMENTED` for combinations the legacy path silently disambiguated |
| Async insert metadata | `03229` | `data_kind` flips `Preprocessed` → `Parsed` in `system.asynchronous_insert_log` |
| Transaction interaction | `01173` | Inline path async-inserts conflict with txn control |
| Column DEFAULT propagation through remote/file | `03667` | **Silent data loss**: DEFAULT not applied (`7` → `0`) |
| `system.tables` reported size | `00753` | `Memory` engine `total_bytes` changes due to different block layout |
| Multi-query VALUES + comment | `04045` | Silent data loss in multi-query parser |
| Pure smoke | `04064` | Original PR #101034 test |

If continuing to pin is the preferred direction, this will likely settle out within a few more rounds. The alternative options are:
1. **Make the randomization opt-in** — only tests that explicitly tag inline-insert-data as in-scope get randomized. Reverses the default; sacrifices coverage breadth but stops the bleed.
2. **Treat the deltas as bugs.** Particularly the silent data loss (`03667`, `04045`), the connection-lifetime change (`01286`), and the inline+stdin combination (`02270`) are real semantic differences that may be worth surfacing in their own follow-up issues regardless of how this PR closes.

Happy to take either direction — let me know what you'd like.

Session: `cron:clickhouse-ci-task-worker:20260512-161500`

tests/queries/0_stateless/03667_insert_columns_description.sql
```diff
@@ -1,3 +1,12 @@
+-- Random settings limits: send_table_structure_on_insert_with_inline_data=(1, 1)
+-- This test asserts that column DEFAULT values from a table's schema are applied when
+-- INSERTing `VALUES (NULL)` through `remote(...)` and `file(...)` table functions. With
+-- `send_table_structure_on_insert_with_inline_data=0` the inline-server-parsed path does not
+-- propagate the calling-side column descriptions to the receiving INSERT, so DEFAULTs are not
+-- applied and NULLs become zero (e.g. `7` -> `0` for `c Int DEFAULT 7`, and `4 77` -> `4 0`
```
Ergus: Is this behavior intended and desired? Do we need to open a bug report for this?

tests/queries/0_stateless/04027_json_input_format.sql
```diff
@@ -20,7 +20,7 @@ SET input_format_column_name_matching_mode='auto';
 
 INSERT INTO json_test FORMAT JSONEachRow {"age": 0, "AGE": 10};
 
-INSERT INTO json_test FORMAT JSONEachRow {"AgE": 1, "aGe": 20}; -- { clientError 117 }
+INSERT INTO json_test FORMAT JSONEachRow {"AgE": 1, "aGe": 20}; -- { error 117 }
```
Ergus, line 33: Very minor:

Probably it worth mentioning in a comment somewhere in this test that this change relaxes the error catching and if a future maintainer ever pins these tests to setting=1 they should restore the stricter clientError annotation

tests/queries/0_stateless/04045_values_insert_trailing_comment.sql
```diff
@@ -1,4 +1,11 @@
+-- Random settings limits: send_table_structure_on_insert_with_inline_data=(1, 1)
 -- Trailing SQL comment after VALUES data (no semicolon) should be skipped, not parsed as data.
+-- The test pins `send_table_structure_on_insert_with_inline_data` to 1 (legacy path) because
+-- the inline-server-parsed path (setting=0) has a separate, pre-existing multi-query parsing
+-- issue: when a trailing comment follows the VALUES data inside a multi-query stream, the
+-- subsequent queries are silently skipped. That is an independent bug in the inline path and
```
Ergus: Similarly to above. Is this behavior intended or we must open a bug report for this and reference it here?

tests/queries/0_stateless/03667_insert_columns_description.sql
PedroTadim: I would guess this is a bug. @alexey-milovidov WDYT?

groeneai:
Thanks for the review — addressed both of @Ergus's points in commit a40b701aae3, and adding data for @PedroTadim's third question below.

## @Ergus on `04027_json_input_format.sql` — relaxed `error 117`

Added a maintenance note near the top of the test:

> `error 117` (rather than the stricter `clientError 117`) accepts the `INCORRECT_DATA`
> error from either reporting path. The legacy path (setting = 1, client converts JSON to
> a native block) reports it as `clientError 117`; the inline-server-parsed path (setting
> = 0, server parses inline JSON) reports it as `serverError 117`. If a future maintainer
> pins this test to `send_table_structure_on_insert_with_inline_data` = 1, the stricter
> `clientError 117` can be restored.

The ambiguous-field detection itself works correctly on both paths — only the attribution
of which side raised the error differs.

## @Ergus on `04045_values_insert_trailing_comment.sql` — bug or intended?

Rewrote the inline comment to be explicit:

> With setting = 0 (inline-server-parsed path), in a multi-query stream, when a trailing
> SQL comment follows the VALUES data, the **subsequent queries are silently skipped**
> (no error, no data, queries effectively dropped).

This is **silent data loss**, which is almost certainly a bug rather than intended
behavior. The comment now states this directly. We have not filed an issue yet because
we wanted reviewer confirmation first — if @alexey-milovidov / @Ergus agree this is a
bug, we (or whoever has the most context on the inline path) should open an issue and
link it from the comment; the pin can then be removed once the inline path is fixed.

## @PedroTadim on `03667_insert_columns_description.sql` — DEFAULTs not propagated through `remote(...)` / `file(...)` table functions

Same shape as the 04045 case — a likely-unintended behavior difference between the two
paths. Concrete data from the reproduction we did:

| Setup | Setting | Result |
|---|---|---|
| `INSERT INTO TABLE FUNCTION remote('localhost:9000', database(), 't0', 'default', '') VALUES (NULL)` with `t0 (c Int DEFAULT 7)` | setting = 1 (legacy) | row has `c = 7` (DEFAULT applied) |
| same | setting = 0 (inline) | row has `c = 0` (DEFAULT dropped, NULL coerced to 0) |
| `INSERT INTO TABLE FUNCTION file(database() || '_test.csv', CSV, 'a Int, b Int DEFAULT 77') VALUES (3, 3), (4, NULL)` | setting = 1 | rows are `3 3` and `4 77` |
| same | setting = 0 | rows are `3 3` and `4 0` (DEFAULT dropped) |

Both `remote(...)` and `file(...)` exhibit the same divergence: the inline-server-parsed
path does not propagate the calling-side column description (the receiving table's
DEFAULT clause for `remote`, the caller-supplied CSV schema with DEFAULT for `file`) to
the receiving INSERT. The result is that explicit `NULL`s in `VALUES` are written as
zero (the type's default-of-default) instead of being expanded to the schema-declared
DEFAULT.

The test is currently pinned to setting = 1 with `Random settings limits:
send_table_structure_on_insert_with_inline_data=(1, 1)`. We did not take a position on
whether this is a bug since you explicitly tagged @alexey-milovidov — happy to file an
issue or remove the pin once there is a verdict.

For convenience, the three "likely bug" candidates surfaced by this PR's pinning round
are summarized:

1. **`03667` — DEFAULTs not propagated through `remote(...)` / `file(...)`** (this thread,
   silent data loss on `VALUES (NULL)`).
2. **`04045` — multi-query trailing-comment silently drops subsequent queries** (silent
   data loss in a different shape).
3. **`01173` — inline path triggers async-insert server-side which conflicts with
   transactions** (less concerning; surfaces as a transparent error rather than data
   loss, but still an unexpected interaction with `BEGIN`/`COMMIT`).

The other 8 pinned tests (`04027`, `04029`, `00753`, `03229`, `01286`, `01501`, `02270`,
`02908`) are honest behavior deltas — `error N` annotation differences, connection
lifetime after `VIOLATED_CONSTRAINT`, client error prefix vanishing because parsing
moved server-side, `system.asynchronous_insert_log` `data_kind` boundary changes, etc. —
not data loss.

## Status

- CI is running on a40b701aae3.
- Comment-only commit verified 20/20 with full randomization locally; no SQL logic changed.
- Awaiting @alexey-milovidov's verdict on whether to file issues for the three suspect
  behaviors, and on whether the pinned tests should stay pinned long-term or get fixed
  in the inline path.

alexey-milovidov:
@groeneai, fix conflicts.

groeneai:
@ alexey-milovidov rebased onto current master (merged `upstream/master` at f5e7496e3ced). One conflict in `tests/queries/0_stateless/01173_transaction_control_queries.sql` — master removed the `no-encrypted-storage` tag while this PR added a `Random settings limits:` pin + explanation. Resolution: kept master's tag line, kept our pin + explanation. PR is `MERGEABLE` again; CI will re-run.

Session: cron:clickhouse-ci-task-worker:20260606-214100

groeneai:
CI ledger — covered SHA: f5e7496e3ced

CI fully finished on this HEAD: every Stateless / Integration / Stress / AST fuzzer / BuzzHouse / Build / Compatibility / Install / Style / Docs / Mergeable check is SUCCESS or correctly SKIPPED, no IN_PROGRESS/QUEUED/PENDING work, Finish Workflow = SUCCESS, mergeable state = CLEAN, and no test FAIL/ERROR rows in CIDB for this PR in the last 36h. Ready for review.

tests/queries/0_stateless/03667_insert_columns_description.sql
alexey-milovidov: It's a problem. Defaults should be correctly calculated by the server when the data is sent inline in the query - similarly to how they are in the HTTP protocol.

alexey-milovidov: Good news: this already works correctly — the server *does* calculate `DEFAULT` values when the data is parsed inline, exactly as over the HTTP protocol. The claim in the comment (that `send_table_structure_on_insert_with_inline_data=0` turns `NULL` into zero for `remote(...)`/`file(...)`) does not reproduce on the actual code.

Running the exact test queries with `send_table_structure_on_insert_with_inline_data=0` produces precisely the committed reference output:

```
-- remote table function columns description
7
7
-- file table function columns description
0	0
1	77
2	2
3	3
4	77
```

Cross-checked across all three parsing paths for `INSERT ... file(..., 'b Int DEFAULT 77') VALUES (4, NULL)` — all yield `b = 77`:
- native, `send_table_structure_on_insert_with_inline_data = 1` (legacy, client-parsed)
- native, `send_table_structure_on_insert_with_inline_data = 0` (server-parsed inline)
- HTTP (always server-parsed)

The reason the original analysis looked like a bug: defaults for `INSERT INTO TABLE FUNCTION ...` are indeed *not* applied by `getSourceFromInputFormat` (it is skipped because `table_id` is empty for table functions). But they are applied downstream in the insertion chain by `InsertDependenciesBuilder::createPreSink` -> `addMissingDefaults` -> `evaluateMissingDefaults`, which runs for every real destination including table functions. Gating is by `input_format_defaults_for_omitted_fields`, identically for both values of the setting.

So there is no behavior delta here and no separate bug to track. I removed the unnecessary pin and its comment in 8ed06ca9308 — the test now runs under both values of the setting and passes either way.

groeneai:
On the inline-defaults question (your inline review on `03667_insert_columns_description.sql`): the behavior is already correct on current master, so removing the pin (8ed06ca) is right. Independent verification on HEAD:

- `INSERT INTO TABLE FUNCTION remote(...)/file(...) VALUES (NULL)` applies the destination column `DEFAULT` with `send_table_structure_on_insert_with_inline_data` at **both** 0 and 1; full test output matches the committed reference, 15/15 alternating-setting runs.
- Confirmed the mechanism: for table functions `getSourceFromInputFormat` skips defaults (empty `table_id`), but `InsertDependenciesBuilder::createPreSink` -> `addMissingDefaults` -> `evaluateMissingDefaults` applies them downstream, gated by `input_format_defaults_for_omitted_fields` identically for both setting values. The original `NULL -> 0` claim was fixed by #105555 (insertion-table `columns_description` preserved in context).

Transparency: a worker run briefly force-pushed a duplicate unpin commit over 8ed06ca and immediately restored your commit as the branch HEAD; current HEAD is 8ed06ca, no changes lost.

groeneai:
@alexey-milovidov Correction to my 06-13 agreement that this "works either way": it does not. The fork CI ran on this branch's HEAD (8ed06ca, your unpin) and `03667_insert_columns_description` now fails **deterministically** with `send_table_structure_on_insert_with_inline_data=0`:

```
-- remote table function columns description
-7        (reference)
+0        (actual: DEFAULT not applied)
 ...
-4	77    (reference)
+4	0     (actual: DEFAULT not applied)
```

Failed on every config it ran in: amd_debug 183/183, arm_binary 221/221, amd_tsan, amd_msan(WasmEdge), amd_tsan(s3). The minimizer's verdict: "caused by randomized settings. Culprit: --send_table_structure_on_insert_with_inline_data 0". Report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104640&sha=8ed06ca93081214a31acef155c8df020fd950600&name_0=PR&name_1=Stateless%20tests%20%28amd_debug%2C%20parallel%29

So your original read on 06-13 was right: it is a real server bug. Why our 06-13 checks missed it: the bug only appears with **synchronous** insert (`async_insert=0`), and a default server has `async_insert=1`. CI installs `tests/config/users.d/disable_async_inserts.xml` (sets `async_insert=0`), so CI exercises the buggy path and a bare local server does not.

Mechanism (reproduced and traced locally):
- For inline `INSERT ... VALUES`, defaults are applied only by `AddingDefaultsTransform`, added in `getSourceFromInputFormat` (getSourceFromASTInsertQuery.cpp:84). That branch is gated on `ast_insert_query->table_id`, which is **empty for table functions** -> transform skipped -> defaults/null-as-default dropped. Plain named tables have `table_id` set, so they work.
- The downstream `InsertDependenciesBuilder::createPreSink -> addMissingDefaults` does not cover this case: it only fills columns *missing* from the input (the inline NULL is present, not missing), and its null-as-default is gated by `insert_null_as_default`, which is set only for `INSERT ... SELECT` (`as_insert_query->select`, InsertDependenciesBuilder.cpp:752), not for `VALUES`.
- The async path works only because `AsynchronousInsertQueue` adds its own `AddingDefaultsTransform` (AsynchronousInsertQueue.cpp:1261), which is why `async_insert=1` hides the bug.

I have a verified fix (19/6 lines in `getSourceFromInputFormat`): when `table_id` is empty, fall back to `context->getInsertionTableColumnsDescription()` (always populated for table functions by `InterpreterInsertQuery::setInsertContextValues`). With it, table-function inline inserts apply DEFAULTs identically to named tables, across plain/Nullable/LowCardinality/MATERIALIZED. `03667` passes with `setting=0` and 20/20 under full randomization; named-table behavior is unchanged.

How would you like to proceed: apply this fix on this branch, or land it as a separate bug-fix PR and re-pin `03667` here? I will not push over your commit without your call.

alexey-milovidov:
Pushed a few follow-ups.

**Correction to my earlier "Good news" note on `03667_insert_columns_description`.** CI on the exact commit I reviewed (`8ed06ca`) shows the test failing **51/51** specifically with `--send_table_structure_on_insert_with_inline_data 0`, and passing 97/97 without it. So the un-pin was wrong: the server-parsed inline path really does *not* apply the column `DEFAULT` / `input_format_null_as_default` for an explicit `NULL` when the destination is a **table function** — `INSERT INTO TABLE FUNCTION remote(...)/file(...) VALUES (NULL)` yields `0` instead of the default, while a direct table `INSERT` is unaffected. My earlier analysis conflated omitted-field defaults (which *are* applied downstream by `evaluateMissingDefaults`) with `input_format_null_as_default`, which is applied during parsing and is what this test actually exercises.

Rather than a blind server-side behavior change, I re-pinned `03667` to the legacy path (same `Random settings limits` mechanism as the other affected tests here) and filed the real bug as https://github.com/ClickHouse/ClickHouse/issues/109253 for a proper fix. Once fixed, the pin can be removed.

**`04064` (@clickhouse-gh review):** added two path-sensitive `--async_insert 0` cases — mixed inline+stdin that must be rejected with `Processing inline insert data with both inlined and external data ...`. That rejection only fires when the inline path is actually taken, so it proves the explicit inline choice still wins over `async_insert = 0` rather than silently falling back to the legacy path (the positive inserts alone did not prove this).

**`04045`:** linked the same tracking issue (#109253, which also covers the trailing-comment silent-data-loss it pins).

**Unrelated CI:** the `Stress test (arm_msan)` failure is `Coordination mode mismatch for stream ...` (parallel replicas), which is https://github.com/ClickHouse/ClickHouse/issues/106039 — unrelated to this test-only PR.

Note: I did not merge `master` locally (the branch is ~3 weeks behind; the working-tree checkout of the ~6.5k changed files times out in this environment), but the only overlapping file, `tests/clickhouse-test`, applies cleanly to current `master`, so the PR remains mergeable.

groeneai:
Thanks for the correction and for filing #109253. This matches what we independently reproduced and root-caused: the failure is real and specific to the server-side inline path with `async_insert = 0` (CI's disable_async_inserts.xml is what makes 03667 hit it). `getSourceFromInputFormat` adds `AddingDefaultsTransform` only when `ast_insert_query->table_id` is set, which it never is for table functions, so `input_format_null_as_default` is skipped and the explicit NULL becomes 0. The async path was unaffected because it adds its own transform.

I opened #109258 for gap 1 (table-function defaults): when `table_id` is empty it falls back to `context->getInsertionTableColumnsDescription()` (populated for table functions by `setInsertContextValues`). Verified both directions on a debug build (remote()/file(), Int/LowCardinality/Array get the DEFAULT, Nullable keeps the explicit NULL), plus a regression test `03816`. It does not touch gap 2 (trailing-comment data loss).

This PR (#104640) stays test-only with 03667 re-pinned; the pin can come off once #109258 lands.

alexey-milovidov (approved):
Fantastic, thanks @groeneai and @PedroTadim!

PR #104642, opened 2026-05-11
RFC: Named Scalars

filimonov:
- Experimental Feature

### Changelog entry:

Added named cached scalars. A named scalar is a `SELECT` whose single-row, single-column result is evaluated in the background, cached, and served to subsequent queries as cheaply as a literal constant. Define one with `CREATE [LOCAL|SHARED] NAMED SCALAR <name> [REFRESH EVERY N UNIT] AS SELECT …` and read it with `getNamedScalar(name)` or `getNamedScalarOrDefault(name, default)`. `LOCAL` scalars are cached per-server; `SHARED` scalars use ClickHouse Keeper for cluster-wide consistency with leader-elected refresh. Access is controlled by a new `getNamedScalar` function-execute privilege; `DEFINER`-style SQL security allows exposing privileged aggregates to unprivileged readers. Runtime state is visible in `system.named_scalars`.

Named scalars let you cache the result of a `SELECT` under a name and read it back as cheaply as a literal constant. The intended use is small, slow-changing, or expensive-to-recompute values — aggregates, reference lookups, ETL watermarks, thresholds, model coefficients — that are read by many queries but should not re-execute the source query each time.

### Motivation

Several workarounds exist today but all have real costs:

1. **A shared one-row Dictionary per named scalar** — uncomfortable to extend: adding a new named scalar means `UNION ALL SELECT 'new_name', (SELECT …)` in the definition.
2. **A generic table + Refreshable Materialized View** — point lookups are awkward, schema is complex.
3. **Many one-row tables/dictionaries** — expensive per read, catalog bloat.
4. **A cached UDF** — the `DETERMINISTIC` framing becomes murky: the function is deterministic *until the cache is invalidated*, which breaks assumptions in the query cache and constant-folding. Background/async refresh makes the interaction even harder to reason about.
5. **`getSetting` with a user-defined prefix** — can only be updated within the same session; not usable as a cluster-wide watermark.

Named scalars are the cleaner answer: define once, refresh in the background, read anywhere.

### What was added

**DDL:**

```sql
CREATE [OR REPLACE] [LOCAL | SHARED] NAMED SCALAR [IF NOT EXISTS] <name>
    [ON CLUSTER <cluster>]
    [DEFINER = { user | CURRENT_USER }] [SQL SECURITY DEFINER]
    [REFRESH EVERY <N> SECOND | MINUTE | HOUR | DAY]
    AS <SELECT …>;

DROP NAMED SCALAR [IF EXISTS] <name> [ON CLUSTER <cluster>];
SYSTEM RELOAD NAMED SCALAR [<name>];
```

**Functions:**

- `getNamedScalar(name)` — returns the cached value or throws.
- `getNamedScalarOrDefault(name, default)` — returns `default` when the scalar is not yet defined or has no value.

**Cache kinds:**

| Modifier          | Value storage       | Refresh                                           |
| ----------------- | ------------------- | ------------------------------------------------- |
| `LOCAL` (default) | Per-node disk cache | Each node independently                           |
| `SHARED`          | ClickHouse Keeper   | Leader-elected; one node evaluates and broadcasts |

`SHARED` is designed for cluster-wide watermarks and feature flags that must be identical across nodes. Consistency is *eventual* — by design, the same trade-off Dictionaries make. For watermark use cases this is fine; stricter guarantees can be layered later via small Keeper state if needed.

**Access control:** a new `getNamedScalar` function-execute privilege controls who can read. The `DEFINER` clause lets privileged users expose aggregates over tables that callers cannot directly `SELECT`.

**Observability:** `system.named_scalars` exposes the current value, type, freshness state, consecutive failure count, and last error. `ProfileEvents` and `CurrentMetrics` track cache hits, refreshes, and in-flight refresh count.

**Definition persistence:** definitions survive server restart (local disk or Keeper, selected at startup).

### Intra-query consistency and constant folding

Multiple `getNamedScalar('x')` calls within the same query will naturally collapse to a single expression due to common subexpression elimination, so the value is consistent within a query. Constant folding also falls out naturally from this.

### Example

```sql
-- ETL high-water mark, shared across replicas
CREATE SHARED NAMED SCALAR last_loaded_id
    REFRESH EVERY 30 SECOND
    AS SELECT max(event_id) FROM staging.events_loaded;

INSERT INTO main.events
    SELECT * FROM staging.events
    WHERE event_id > getNamedScalar('last_loaded_id');

-- Anomaly threshold derived from rolling history
CREATE NAMED SCALAR p99_24h
    REFRESH EVERY 1 HOUR
    AS SELECT quantile(0.99)(query_duration_ms)
       FROM system.query_log WHERE event_time > now() - INTERVAL 1 DAY;

SELECT * FROM live_queries WHERE query_duration_ms > getNamedScalar('p99_24h');
```


Ergus:
@george-larionov Maybe this is somehow related with intermediate caches?

alexey-milovidov:
We have implemented Query Cache for Subqueries: https://presentations.clickhouse.com/2026-release-26.5/#33

While it does not support background refresh, it solved most of the task without much complexity.

PR #103586, opened 2026-04-27
Fix flaky test 04103_user_network_bandwidth_throttler

tiandiwonder:
Add `no-random-settings` tag to fix flaky test `04103_user_network_bandwidth_throttler`.

CI randomized settings can inject `remote_filesystem_read_prefetch=0` and `allow_prefetched_read_pool_for_remote_filesystem=0`, which disable S3 prefetch. On a heavily loaded TSAN machine running parallel tests, this can make the natural S3 read rate drop below the 1 MB/s throttle limit. When natural rate < throttle limit, the token bucket never goes negative and the throttler never sleeps — `UserThrottlerSleepMicroseconds` stays at 0 while `UserThrottlerBytes` is correctly counted, causing the third check to fail transiently.

The write side is unaffected because its data source (`numbers(1e6)`) is memory/CPU-bound and always exceeds 1 MB/s regardless of machine load.

Fixes #103422
Follow-up for: https://github.com/ClickHouse/ClickHouse/pull/103080
CI failure: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102547&sha=83f565de244bc2f42b2f51098ea6bbef4bcf2c9d&name_0=PR&name_1=Stateless%20tests%20%28amd_tsan%2C%20parallel%2C%202%2F2%29

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
...

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.6.1.401`

PR #99854, opened 2026-03-18
Fix unauthenticated TablesStatusRequest in interserver mode

tiandiwonder:
Before this fix, a TCP client could spoof an interserver connection
    (user = `USER_INTERSERVER_MARKER`, empty password) and send a
    `TablesStatusRequest` (packet type 5) before completing cluster-secret
    authentication. The server processed it using a fake interserver context,
    leaking table existence, replication status, and replication delay.

    Root cause: `receivePacketsExpectQuery` handled `TablesStatusRequest`
    without any authentication check. The cluster-secret SHA-256 validation
    only runs inside `processQuery` when a `Query` packet is received.
    The existing guard at lines 917-924 only fires on exceptions, but
    `processTablesStatusRequest` succeeds without throwing.

    Fix: add a pre-authentication check in the `TablesStatusRequest` branch.
    The thrown `AUTHENTICATION_FAILED` exception is caught by the existing
    unauthenticated-interserver guard, which silently closes the connection.

    Test: integration test `test_interserver_tables_status_auth` — verifies
    the connection is closed without leaking table information (old protocol
    rejected by default; two-nodes wrong-secret case rejected on the new
    protocol), and that a properly authenticated cluster keeps working.

### Changelog category (leave one):
- Critical Bug Fix (crash, data loss, RBAC)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix a security issue where an unauthenticated TCP client could probe table existence and replication status via the interserver port.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.7.1.975` (included in `26.7` and later)

PR #104648, opened 2026-05-11
Add precision as parameter for format readable functions

### LINKED ISSUE #61549: Add precision as a parameter for formatReadable* functions
**Use case**

Sometimes you do not want to have decimal values as `.00`

**Describe the solution you'd like**

```
SELECT formatReadableSize(1024, 0);

1 KiB

SELECT formatReadableSize(1024);

1.00 KiB

```

antoniofilipovic:
Closes: https://github.com/ClickHouse/ClickHouse/issues/61549

 The `formatReadableSize`, `formatReadableDecimalSize` and `formatReadableQuantity` functions previously hard-coded two digits after the decimal point.  This PR adds an optional const `UInt8` as the second argument `precision`:

  ```sql
  SELECT formatReadableSize(192851925);      -- 183.92 MiB    (unchanged)
  SELECT formatReadableSize(192851925, 0);   -- 184 MiB
  SELECT formatReadableSize(192851925, 4);   -- 183.9179 MiB
```
The default of 2 preserves prior behavior, so the change is fully backward compatible. The argument is validated as a constant UInt8, which rejects negative numbers. Also, non-constant columns are rejected.

- New Feature

User can now specify an optional precision argument to the `formatReadableSize`, `formatReadableDecimalSize` and `formatReadableQuantity` functions, controlling the number of digits after the decimal point. Default is 2, preserving the prior behavior.

### Version info
- Merged into: `26.6.1.90`

### CODE REVIEW

**Comment by antoniofilipovic** [comment]
@rschu1ze thanks for taking a look! :smile:

File: src/Functions/formatReadable.h [hunk 1]
```diff
@@ -45,14 +42,21 @@ class FunctionFormatReadable : public IFunction
...
+
+    DataTypePtr getReturnTypeImpl(const ColumnsWithTypeAndName & arguments) const override
     {
-        const IDataType & type = *arguments[0];
+        FunctionArgumentDescriptors mandatory_arguments{
+            {"x", static_cast<FunctionArgumentDescriptor::TypeValidator>(&isNumber), nullptr, "Number"},
+        };
+
+        FunctionArgumentDescriptors optional_arguments{
+            {"precision", static_cast<FunctionArgumentDescriptor::TypeValidator>(&isUInt8), &isColumnConst, "const UInt8"},
```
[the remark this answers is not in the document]
antoniofilipovic: Good point. I am not sure what is expected way in the clickhouse to handle such user facing decisions? Can we extend the documentation, or should we fail? I think adding docs would be worth, but I would not fail in that case.

src/Functions/formatReadable.h
```diff
@@ -45,14 +44,33 @@ class FunctionFormatReadable : public IFunction
         return true;
     }
 
-    size_t getNumberOfArguments() const override { return 1; }
+    size_t getNumberOfArguments() const override { return 0; }
 
-    DataTypePtr getReturnTypeImpl(const DataTypes & arguments) const override
+    bool isVariadic() const override { return true; }
```
antoniofilipovic, line 39: note to reviewers: needed for precision as optional arg

src/Functions/formatReadable.h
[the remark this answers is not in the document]
antoniofilipovic: Applied the comment, but set the limit to 100, as that is supported still.

src/Functions/formatReadable.h
```diff
@@ -64,8 +82,14 @@ class FunctionFormatReadable : public IFunction
 
     bool useDefaultImplementationForConstants() const override { return true; }
 
+    ColumnNumbers getArgumentsThatAreAlwaysConstant() const override { return {1}; }
```
antoniofilipovic: Idea is to leave the precision argument with `ColumnConst` wrapper. Otherwise we would need check if column contains wrapper or not. I think this is easier for maintenance

alexey-milovidov:

src/Functions/formatReadableSize.cpp
```diff
@@ -14,9 +14,10 @@ Given a size (number of bytes), this function returns a readable, rounded size w
 The opposite operations of this function are [`parseReadableSize`](#parseReadableSize), [`parseReadableSizeOrZero`](#parseReadableSizeOrZero), and [`parseReadableSizeOrNull`](#parseReadableSizeOrNull).
 This function accepts any numeric type as input, but internally it casts them to `Float64`. Results might be suboptimal with large values.
     )";
-    FunctionDocumentation::Syntax syntax = "formatReadableSize(x)";
+    FunctionDocumentation::Syntax syntax = "formatReadableSize(x[, precision])";
     FunctionDocumentation::Arguments arguments = {
-        {"x", "Size in bytes.", {"UInt64"}}
+        {"x", "Size in bytes.", {"Int8", "Int16", "Int32", "Int64", "UInt8", "UInt16", "UInt32", "UInt64", "Float32", "Float64", "Decimal"}},
+        {"precision", "Optional. Number of digits after the decimal point. Defaults to 2.", {"UInt8"}}
```
rschu1ze: Minor: `UInt8` --> `const UInt8`.

(also in formatReadableQuantity.cpp and formatReadableDecimalSize.cpp)

src/Functions/formatReadable.h
```diff
@@ -45,14 +44,33 @@ class FunctionFormatReadable : public IFunction
...
+
+        FunctionArgumentDescriptors optional_arguments{
+            {"precision", static_cast<FunctionArgumentDescriptor::TypeValidator>(&isUInt8), &isColumnConst, "const UInt8"},
+        };
 
-        if (!isNumber(type))
-            throw Exception(ErrorCodes::ILLEGAL_TYPE_OF_ARGUMENT, "Cannot format {} because it's not a numeric type", type.getName());
+        validateFunctionArguments(getName(), arguments, mandatory_arguments, optional_arguments);
+
+        /// 'ToFixed' in 'double-conversion' fails when requested_digits > 'kMaxFixedDigitsAfterPoint' (100);
```
rschu1ze: Could you please move l. 63-73 to the beginning of `executeImpl`? Function `getReturnTypeImpl` is (at least conceptually) only supposed to perform static type checks.

When you did that, please also check that the read column is non-empty, otherwise (l. 67) reads from nirvana.

src/Functions/formatReadable.h
```diff
@@ -45,14 +44,33 @@ class FunctionFormatReadable : public IFunction
...
+        /// the formatter then silently falls back to 'ToShortest', ignoring the requested precision.
+        if (arguments.size() == 2)
+        {
+            const UInt8 precision = assert_cast<const ColumnConst &>(*arguments[1].column).getValue<UInt8>();
+            if (precision > double_conversion::DoubleToStringConverter::kMaxFixedDigitsAfterPoint)
+                throw Exception(
+                    DB::ErrorCodes::CANNOT_PRINT_FLOAT_OR_DOUBLE_NUMBER,
+                    "Too high precision requested, must not be more than 100, got {}",
```
rschu1ze: I think we should print the value of `kMaxFixedDigitsAfterPoint` in the error msg instead of a hard-coded 100.

rschu1ze (approved):

src/Functions/formatReadable.h
antoniofilipovic: Moved the checks, makes sense.

src/Functions/formatReadableSize.cpp
antoniofilipovic: Done 🙂

src/Functions/formatReadable.h
antoniofilipovic: Updated it

src/Functions/formatReadable.h
```diff
@@ -45,14 +44,21 @@ class FunctionFormatReadable : public IFunction
...
+    size_t getNumberOfArguments() const override { return 0; }
 
-    DataTypePtr getReturnTypeImpl(const DataTypes & arguments) const override
+    bool isVariadic() const override { return true; }
+
+    DataTypePtr getReturnTypeImpl(const ColumnsWithTypeAndName & arguments) const override
     {
-        const IDataType & type = *arguments[0];
+        FunctionArgumentDescriptors mandatory_arguments{
+            {"x", static_cast<FunctionArgumentDescriptor::TypeValidator>(&isNumber), nullptr, "Number"},
```
rschu1ze: Can we rename `x` to `value`? Here and in the docs, please.

antoniofilipovic: Thanks for fixing this

rschu1ze:
@groeneai Make yourself useful (and Nvidia stocks go up) and fix `01563_distributed_query_finish`, `02435_rollback_cancelled_queries`, `02345_implicit_transaction`

groeneai:
@rschu1ze attribution and plan for the three tests:

**`02435_rollback_cancelled_queries`** — fix PR #105474. The 20 hits in 30 days are all on `Stateless tests (arm_asan_ubsan, targeted)` with stdout already correct (`1000000\n0`) and stderr full of `==NNNN==WARNING: ptrace appears to be blocked ... Child exited with signal 36`. The test SIGKILLs `clickhouse-client` mid-INSERT; LSan's at-exit `ptrace` tracer cannot attach under the targeted runner's seccomp filter, falls back to `SIGRTMIN+2`, and the stderr warning trips Praktika's `stderr-fatal` watch. Setting `LSAN_OPTIONS=detect_leaks=0` in the test makes the spawned clients skip the leak scan — running it on a process we are killing on purpose is not useful.

**`01563_distributed_query_finish`** — deferred to @Algunenano's open PR #105457. His approach is materially better than what I had drafted: wrap the loop in `SYSTEM STOP/START DISTRIBUTED SENDS` so the CI `system.*_log_sender` `Distributed` tables (set up by `ci/jobs/scripts/functional_tests/setup_log_cluster.sh`) stop shipping `.bin` files for the duration of the test. That kills the actual source of the background `NETWORK_ERROR` bursts (the cloud log endpoint being unreachable) instead of changing the measurement, and it preserves the bug-detection intent exactly. I won't open a competing PR.

**`02345_implicit_transaction`** — needs a server-side change. The failure is `Code: 159 ... Timeout exceeded (180 s) while flushing system log` and the `180` is a hardcoded `const int timeout_seconds = 180` in `SystemLogBase.cpp:164` (comment already notes `60s proved to be too fast for our parallel functional tests`). `SET receive_timeout` doesn't help here — the throw is server-side from `SystemLogQueue::waitFlush`. CIDB shows 23 hits in 30 days across PRs #99854 (13), #104217 (7), #104947 (1), #104714 (1), exclusively when query_log backlog is heavy on parallel runners. Cleanest fix is to make that timeout a server setting (default still safe, but bumpable) or just raise the constant to ~600s; I have a follow-up task open to land that as a separate PR.

Session: `cron:clickhouse-ci-task-worker:20260520-211500`.

groeneai:
Third of the three flaky tests in this bundle is now in flight: #105484 — bumps the hardcoded `SystemLogQueue::waitFlush` timeout from 180s to 600s to fix `02345_implicit_transaction` failing with `TIMEOUT_EXCEEDED` on parallel CI runners.

Bundle status:

- #105474 — `02435_rollback_cancelled_queries` (LSan leak check skip)
- #105457 — `01563_distributed_query_finish` (CI log-sender bursts) — @Algunenano
- #105484 — `02345_implicit_transaction` (SystemLogBase timeout bump) — this one

CIDB pattern for 02345: 23 master hits across 4 distinct PRs over the last 30 days, all on `parallel` runners. The previous 60s -> 180s bump was made for the same reason; raising again is the same shape of fix.

Algunenano:
> https://github.com/ClickHouse/ClickHouse/pull/105457 — 01563_distributed_query_finish (CI log-sender bursts) — @Algunenano

@groeneai Please make a note to stop pinging me on unrelated PRs. When talking about flaky checks and its fixes, just mention the PR, no need to mention the author. Otherwise I get multiple pings that are completely unrelated to my work (I had half a dozen related to this exact fix I've made)

groeneai:
@Algunenano Noted, sorry for the noise. Going forward, when referencing a sibling fix PR in a thread the author isn't subscribed to, I'll use the bare PR number (e.g. `#105457`) instead of cross-tagging. The rule is memorized for future sessions.

antoniofilipovic:
> @groeneai Make yourself useful (and Nvidia stocks go up) and fix `01563_distributed_query_finish`, `02435_rollback_cancelled_queries`, `02345_implicit_transaction`

@rschu1ze is there anything else I can help with as the failures on PR seem to be unrelated?  Basically we have to wait for the fixes to land on master, I can then rebase and when the checks pass we merge?

rschu1ze:
@antoniofilipovic All good, no need to do anything else from your side. I'll take over.

rschu1ze:
Fix for `04103_user_network_bandwidth_throttler`: https://github.com/ClickHouse/ClickHouse/pull/103586

PR #104653, opened 2026-05-12
Add 'Buffers' as a serialization format to WASM UDF

bacek:
- Experimental Feature

Add "Buffers" as WASM UDF serialization_format.


bacek:
@vdimir @alexey-milovidov FYI. Change in `ColumnWithTypeAndName` due BuffersReader null pointer deference.  Another option was to auto-create it inside Reader it self like this:
```
 auto column = header.getByPosition(i).cloneEmpty();
 ColumnPtr & read_column = column.column;
-if (!read_column)
+if (!read_column && column.type)
     read_column = column.type->createColumn();
 auto serialization = column.type->getDefaultSerialization();
```

Not sure which one is more correct approach.

bacek:
let me re-check it

PR #104655, opened 2026-05-12
Fix scale error in exception message of parseDateTime64InJodaSyntax

KevinyhZou:
- Bug Fix

### Changelog entry (a [user-readable short description]
Fix incorrect precision value in error messages from `parseDateTime64InJodaSyntax` when the Joda pattern specifies an invalid subsecond precision (for example, more than six fractional digits).


KevinyhZou:
@clickgapai can you help to review this?

clickgapai:
@KevinyhZou Reviewed. The fix is correct and the bug is confirmed.

**Confirmed bug at `src/Functions/parseDateTime.cpp:509,511`:**

The error messages used `scale` (the member variable) instead of `scale_` (the parameter). At the point of exception, `scale = scale_` hasn't executed yet, so the message prints the wrong value.

I verified this locally:
```
SELECT parseDateTime64InJodaSyntax('2024-10-09 10:30:10.1234567', 'yyyy-MM-dd HH:mm:ss.SSSSSSS', 'UTC');

-- Current (wrong): "Precision 0 is invalid (must be [0, 6])"
-- After fix: "Precision 7 is invalid (must be [0, 6])"
```

**Test coverage:**
The added test correctly triggers the Joda path with precision 7 (inferred from 7 'S' chars), which is > 6. Test follows conventions (explicit UTC timezone, serverError annotation).

**Minor observation:**
The MySQL path (line 509) has the same bug and same fix, but isn't tested by this PR. You might want to add a test for that path too, something like a test that would trigger `scale != 6` for MySQL syntax. Though I couldn't easily find a way to specify a non-6 precision for MySQL parseDateTime64 via the public API, so it may only be reachable through internal code paths.

LGTM — straightforward fix, the bug is real, the test validates it.

src/Functions/parseDateTime.cpp
```diff
@@ -506,9 +506,9 @@ namespace
             /// Because the scale argument for parseDateTime*() is constant, always throw an exception (don't allow continuing to the
             /// next row like in other set* functions)
             if (parse_syntax_ == ParseSyntax::MySQL && scale_ != 6)
-                throw Exception(ErrorCodes::CANNOT_PARSE_DATETIME, "Precision {} is invalid (must be 6)", scale);
+                throw Exception(ErrorCodes::CANNOT_PARSE_DATETIME, "Precision {} is invalid (must be 6)", scale_);
```
Ergus: Is there some test for this branch anyhow?

tests/queries/0_stateless/02668_parse_datetime_in_joda_syntax.sql
```diff
@@ -340,6 +340,8 @@ select parseDateTimeInJodaSyntax('2025', 'YYY', 'UTC') = toDateTime('2025-01-01'
 select parseDateTimeInJodaSyntax('2025', 'xxx', 'UTC') = toDateTime('2024-12-30', 'UTC');
 select parseDateTimeInJodaSyntax('2025', 'yyy', 'UTC') = toDateTime('2025-01-01', 'UTC');
 
+SELECT parseDateTime64InJodaSyntax('2024-10-09 10:30:10.1234567', 'yyyy-MM-dd HH:mm:ss.SSSSSSS', 'UTC'); -- { serverError CANNOT_PARSE_DATETIME }
```
Ergus: I don't know if we can test not only the error code, but the error message content

Ergus (commented):
Overall LGTM:

Is this an Improvement or a Bug fix? I think that category is wrong.

src/Functions/parseDateTime.cpp
KevinyhZou: I don't know how to trigger this problem, but it seems an obvious bug, could you give me a test example?

Ergus: I agree it is an obvious bug. 

But it is very weird that it didn't trigger during compilation. So better than just fixing it, it worth figuring out why it was not triggering and ensuring that we don't repeat it again in the future.

Ergus: Looking deeper into this I think that this branch is actually dead code. 

For every MySQL-syntax DateTime64 function, getReturnTypeImpl hardcodes scale=6 at line 688:

https://github.com/ClickHouse/ClickHouse/blob/5794d425fbfe421e4ade7ede135b1b2fd6ebf64d/src/Functions/parseDateTime.cpp#L686-L689

So scale_ is always 6 when setScale is called, and scale_ != 6 is never true. The MySQL guard cannot be triggered from the public SQL API at all.

We could probably substitute the runtime check with some assertion or compile time check, but not in the scope of this PR

KevinyhZou:
cc @Ergus

KevinyhZou:
please add `can be tested` label  @clickgapai

Ergus:
Hi @KevinyhZou 

Please add a changelog following the guideline recomendations.

KevinyhZou:
done

rschu1ze:
Cont'd in https://github.com/ClickHouse/ClickHouse/pull/120489

PR #104662, opened 2026-05-12
Add `EXPLAIN PIPELINE` option to compact repeated processor chains

### LINKED ISSUE #102998: `EXPLAIN PIPELINE` renders interleaved processors as a nested tower instead of grouping them
### Problem

`EXPLAIN PIPELINE` output for multi-stream joins contains a deeply nested tower of processors that looks like a pathological plan but is actually one step rendered incorrectly. The underlying pipeline is correct, this is a display-only problem

### Repro

[Fiddle](https://fiddle.clickhouse.com/0c26fbc2-16d4-4a53-aaab-f4f1a18bb5fb)

```sql
EXPLAIN PIPELINE
SELECT number FROM numbers(1) AS l INNER JOIN numbers(1) AS r USING (number)
SETTINGS max_threads = 8;
```
```
(Expression)
ExpressionTransform × 8
  (Join)
  SimpleSquashingTransform × 8                               -- Correctly grouped
    JoiningTransform × 8 2 → 1
      Resize 1 → 8
        (Expression)
        ExpressionTransform
          (Filter)
          FilterTransform
            (ReadFromSystemNumbers)
            NumbersRange 0 → 1
        (Expression)
        Resize × 2 8 → 1
          FillingRightJoinSide                               -- Not grouped
            SimpleSquashingTransform
              FillingRightJoinSide
                SimpleSquashingTransform
                  FillingRightJoinSide
                    SimpleSquashingTransform
                      FillingRightJoinSide
                        SimpleSquashingTransform
                          FillingRightJoinSide
                            SimpleSquashingTransform
                              FillingRightJoinSide
                                SimpleSquashingTransform
                                  FillingRightJoinSide
                                    SimpleSquashingTransform
                                      FillingRightJoinSide
                                        SimpleSquashingTransform
                                          Resize 1 → 8
                                            ExpressionTransform
                                              (BuildRuntimeFilter)
                                              BuildRuntimeFilterTransform
                                                (Expression)
                                                ExpressionTransform
                                                  (ReadFromSystemNumbers)
                                                  NumbersRange 0 → 1
```

niyue:
Group right-side join processors by pipeline layer so `EXPLAIN PIPELINE` renders parallel `FillingRightJoinSide` and `SimpleSquashingTransform` processors compactly instead of as an interleaved nested tower. 

Fixes https://github.com/ClickHouse/ClickHouse/issues/102998

### Changelog category:
- Improvement

### Changelog entry:
- Add `EXPLAIN PIPELINE` option to compact repeated processor chains

---

> [!NOTE]
> **Low Risk**
> Changes are limited to `EXPLAIN PIPELINE` text rendering and option plumbing (including distributed explain propagation), with no impact on query execution semantics.
> 
> **Overview**
> Adds a new `EXPLAIN PIPELINE` boolean setting, `compact_repeated_processor_chains`, which detects adjacent repeated processor *chains* and prints them once with an aligned repetition marker (improving readability for parallel joins).
> 
> Plumbs the option from `InterpreterExplainQuery` through `QueryPlan::ExplainPipelineOptions` and `IQueryPlanStep::FormatSettings`, and ensures it is forwarded when rendering distributed pipelines from `ReadFromRemote`. Documentation and a new stateless test cover the expected compacted output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82cdb978f3d42181cde7a2d77e18e32a7bfbdb0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.236`


tests/queries/0_stateless/02236_explain_pipeline_join.reference
```diff
@@ -16,43 +16,13 @@ ExpressionTransform × 16
...
-                                                                            Resize 1 → 16
-                                                                              ExpressionTransform
-                                                                                (Expression)
-                                                                                ExpressionTransform
-                                                                                  (Limit)
-                                                                                  Limit
-                                                                                    (ReadFromSystemNumbers)
-                                                                                    NumbersRange 0 → 1
+            FillingRightJoinSide × 16
+              SimpleSquashingTransform × 16
```
Fgrtue: This seem to be inconsistent -- how should we differentiate when one line is repeated 16 times versus 2 lines are repeated 16 times?

With the current notation, it should be then:

```
FillingRightJoinSide  
     FillingRightJoinSide 
             ... + 14 more times
                SimpleSquashingTransform
                  SimpleSquashingTransform
                    ... + 14 more times
```

But this should be:

```
FillingRightJoinSide
   SimpleSquashingTransform
         ... + the same couple 15 more times
```

src/QueryPipeline/QueryPipelineBuilder.cpp
```diff
@@ -467,15 +467,17 @@ std::unique_ptr<QueryPipelineBuilder> QueryPipelineBuilder::joinPipelinesRightLe
...
                     auto squashing = std::make_shared<SimpleSquashingChunksTransform>(right->getSharedHeader(), min_block_size_rows, min_block_size_bytes);
                     connect(*outport, squashing->getInputs().front());
-                    processors.emplace_back(squashing);
                     auto adding_joined = std::make_shared<FillingRightJoinSideTransform>(right->getSharedHeader(), join, filling_finish_counter);
                     connect(squashing->getOutputPort(), adding_joined->getInputs().front());
-                    processors.emplace_back(std::move(adding_joined));
+                    processors.emplace_back(std::move(squashing));
+                    filling_processors.emplace_back(std::move(adding_joined));
                 }
+                processors.splice(processors.end(), filling_processors);
```
Fgrtue: The solution to the problem lies somewhere in the EXPLAIN PIPELINE function, no in the processor -- we don't want to change the way the query is executed, but we would like the EXPLAIN PIPELINE to look different

Fgrtue (changes requested):
Thank you for contribution! The solutions should be implemented in a different part of code -- we just want to see EXPLAIN slghtly different, not to change the way the pipeline is built.

src/QueryPipeline/QueryPipelineBuilder.cpp
niyue: You are right, I made a pretty stupid mistake here. Thanks for pointing it out. I’ll take another look and fix this in the `EXPLAIN PIPELINE` rendering logic instead of changing processor construction/order.

Fgrtue: Thank you! Don't you worry about it.

tests/queries/0_stateless/02236_explain_pipeline_join.reference
niyue: Yes, good point. The notation is ambiguous because it does not say whether `× 16` applies to one processor line or to the whole two-processor chain.

Do you think `Repeated chain × 16` would be clearer here?

```
Repeated chain × 16
  FillingRightJoinSide
    SimpleSquashingTransform
```

Here `Repeated chain` would be only a synthetic `EXPLAIN PIPELINE` rendering group, not a real processor. It means that the whole chain below it is repeated 16 times as one unit.

I would keep this intentionally limited to repeated linear chains, not full DAG isomorphism. So `Repeated chain × N` would only be emitted when each repeated unit is a simple chain and the repeated units are not connected to each other. More complex repeated DAG shapes would still be printed without this grouping.

Fgrtue: Probably, I would prefer not to add another line, as it can be confused with processors easily. We need to have some intuitive way to designate that particular number of of processors repeats 16 times. 

To me it seems, that the easiest way would be to have something like this:

```
 FillingRightJoinSide            │                                                                                                                                                     
     SimpleSquashingTransform    │ × 8
                   Resize 1 → 8
```

niyue: I've updated the implementation by following the suggested approach. @Fgrtue could you please check it out, thanks.

src/Processors/QueryPlan/IQueryPlanStep.cpp
```diff
@@ -176,29 +191,197 @@ static void doDescribeProcessor(const IProcessor & processor, size_t count, IQue
...
+    return false;
+}
+
+static bool isProcessorRangeChain(const std::vector<const IProcessor *> & processors, size_t begin, size_t length)
+{
+    for (size_t i = 0; i + 1 < length; ++i)
+    {
+        const auto & current = *processors[begin + i];
+        const auto & next = *processors[begin + i + 1];
+        if (!hasConnectionTo(next, current))
```
niyue: The detection is graph-aware, not only text-order based. After finding a candidate repeated sequence, it verifies that every repeated unit forms a real linear chain through processor port connections, and that different repeated units are not connected to each other.

For example, a displayed sequence like `A B A B` is collapsed only if each `A B` unit is an independent chain (`B -> A`) and there is no connection between the two units. If the real graph is one longer nested chain such as `B2 -> A2 -> B1 -> A1`, the units are connected to each other, so it will be printed without this grouping.

src/Processors/QueryPlan/IQueryPlanStep.cpp
```diff
@@ -176,29 +191,197 @@ static void doDescribeProcessor(const IProcessor & processor, size_t count, IQue
...
+    return UTF8::computeWidth(reinterpret_cast<const UInt8 *>(line.data()), line.size());
+}
 
-    for (auto it = processors.rbegin(); it != processors.rend(); ++it)
+static bool areProcessorsSimilarForExplain(const IProcessor & lhs, const IProcessor & rhs)
+{
+    return lhs.getName() == rhs.getName()
+        && lhs.getInputs().size() == rhs.getInputs().size()
+        && lhs.getOutputs().size() == rhs.getOutputs().size()
+        && lhs.getDescription() == rhs.getDescription();
```
niyue: One related change is that the old adjacent-processor compression is now stricter. Previously it grouped consecutive processors only by name. This could make different `Resize` processors look like one repeated line, for example compressing `Resize 1 → 16` and `Resize 16 → 1` just because both are named `Resize`, which seemed incorrect.
For example, in the [02236_explain_pipeline_join.reference‎](https://github.com/ClickHouse/ClickHouse/pull/104662/changes#diff-fbac83319da6a6685813726d26fbbba14a458f4efdb91fa7c3069a3f13ad3e95), there was a step:
```
Resize × 2 16 → 1
```

But they are actually two different steps with the same name:
```
Resize 1 → 16
     Resize 16 → 1
```

Now processors are considered equal for this compact rendering only when the name, input count, output count, and description match. So `Resize 1 → 16` and `Resize 16 → 1` are printed as two separate lines, while truly identical resize processors can still be rendered as `Resize × N 24 → 24`.

src/Processors/QueryPlan/IQueryPlanStep.cpp
```diff
@@ -176,29 +191,197 @@ static void doDescribeProcessor(const IProcessor & processor, size_t count, IQue
...
+            if (hasConnectionBetweenProcessorRanges(processors, begin + lhs_item * length, begin + rhs_item * length, length))
+                return false;
+
+    return true;
+}
+
+static std::pair<size_t, size_t> findRepeatedProcessorChainRange(const std::vector<const IProcessor *> & processors, size_t begin)
+{
+    static constexpr size_t min_period = 2;
+    static constexpr size_t max_period = 8;
```
niyue, line 363: The repeated unit length is intentionally bounded. The renderer currently tries candidate linear-chain lengths from 2 to 8 processors. This covers the practical cases we want to compact, like `FillingRightJoinSide -> SimpleSquashingTransform`, without turning `EXPLAIN PIPELINE` rendering into a general graph-compression algorithm.

src/Processors/QueryPlan/IQueryPlanStep.cpp
```diff
@@ -176,29 +191,197 @@ static void doDescribeProcessor(const IProcessor & processor, size_t count, IQue
...
     {
-        if (prev && prev->getName() != (*it)->getName())
+        const size_t item_begin = begin + item * length;
+        if (!isProcessorRangeChain(processors, item_begin, length))
+            return false;
+    }
+
+    for (size_t lhs_item = 0; lhs_item < count; ++lhs_item)
+        for (size_t rhs_item = lhs_item + 1; rhs_item < count; ++rhs_item)
+            if (hasConnectionBetweenProcessorRanges(processors, begin + lhs_item * length, begin + rhs_item * length, length))
```
Fgrtue: It would be nice to refactor this part (along with hasConnecitonBetweenProcessorRanges) for better readability. Could you please rename the variables, and maybe simplify the number of loops (by using hash map for example)?

src/Processors/QueryPlan/IQueryPlanStep.cpp
```diff
@@ -176,29 +191,197 @@ static void doDescribeProcessor(const IProcessor & processor, size_t count, IQue
...
+                return true;
+
+    return false;
+}
+
+static bool isRepeatedProcessorRangeIndependentChain(
+    const std::vector<const IProcessor *> & processors,
+    size_t begin,
+    size_t length,
+    size_t count)
```
Fgrtue: Let's make the variables more descriptive for better readability.

src/Processors/QueryPlan/IQueryPlanStep.cpp
```diff
@@ -176,29 +191,197 @@ static void doDescribeProcessor(const IProcessor & processor, size_t count, IQue
...
+
+static bool areRepeatedProcessorRangesEqual(
+    const std::vector<const IProcessor *> & processors,
+    size_t lhs_begin,
+    size_t rhs_begin,
+    size_t length)
+{
+    for (size_t i = 0; i < length; ++i)
+        if (!areProcessorsSimilarForExplain(*processors[lhs_begin + i], *processors[rhs_begin + i]))
+            return false;
```
Fgrtue: Let's also refactor this function to use more descriptive variable names

Fgrtue (commented):
This is a good approach. The result also looks good. However, we should add a setting that will control this behavior. This could be a setting in the `EXPLAIN PIPELINE`, that users could turn on and off if they want to. By default, let's turn if off by now. So the old tests don't need to be fixed. Also please add a new test, that will test this new setting.

src/Processors/QueryPlan/IQueryPlanStep.cpp
niyue: Refactored

src/Processors/QueryPlan/IQueryPlanStep.cpp
niyue: Refactored

src/Processors/QueryPlan/IQueryPlanStep.cpp
niyue: Thanks for the guidance. I've refactored the code and used an `unordered_map` in the process to avoid some nested loops. Please check it out and let me know if there is anything else that needs revision.

niyue:
> This is a good approach. The result also looks good. However, we should add a setting that will control this behavior. This could be a setting in the `EXPLAIN PIPELINE`, that users could turn on and off if they want to. By default, let's turn if off by now. So the old tests don't need to be fixed. Also please add a new test, that will test this new setting.

@Fgrtue A new setting called `compact_repeated_processor_chains` is introduced for `EXPLAIN PIPELINE` to enable this process, and it is off by default. e.g.
```sql
EXPLAIN PIPELINE compact_repeated_processor_chains = 1
SELECT * FROM
(
    SELECT * FROM system.numbers LIMIT 100000
) t1
ALL LEFT JOIN
(
    SELECT * FROM system.numbers LIMIT 100000
) t2
USING number
SETTINGS max_threads=16;
```

src/Processors/QueryPlan/ReadFromRemote.cpp
```diff
@@ -861,12 +861,13 @@ static ASTPtr makeExplain(const ExplainPlanOptions & options, ASTPtr query)
...
 }
 
-static ASTPtr makeExplainPipeline(bool header, bool distributed, ASTPtr query)
+static ASTPtr makeExplainPipeline(bool header, bool distributed, bool compact_repeated_processor_chains, ASTPtr query)
 {
     auto explain_settings = make_intrusive<ASTSetQuery>();
     explain_settings->is_standalone = false;
     explain_settings->changes.emplace_back("header", int(header));
     explain_settings->changes.emplace_back("distributed", int(distributed));
+    explain_settings->changes.emplace_back("compact_repeated_processor_chains", int(compact_repeated_processor_chains));
```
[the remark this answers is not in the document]
niyue: Fixed, this setting is now only appended when enabled.

Fgrtue (commented):
Good! Only one thing left -- documentation. Could you add it to the other settings for EXPLAIN PIPELINE in `docs/en/sql-reference/statements/explain.md` please?

niyue:
@Fgrtue Documentation is updated, could you please check it out? Thanks.

Fgrtue (approved):

docs/en/sql-reference/statements/explain.md
```diff
@@ -601,6 +601,7 @@ Settings:
 - `header` — Prints header for each output port. Default: 0.
 - `graph` — Prints a graph described in the [DOT](https://en.wikipedia.org/wiki/DOT_(graph_description_language)) graph description language. Default: 0.
 - `compact` — Prints graph in compact mode if `graph` setting is enabled. Default: 1.
+- `compact_repeated_processor_chains` — Compacts adjacent repeated processor chains in text output by showing one copy of the chain with a repetition count. This can make parallel pipelines easier to read when the same chain appears many times, for example in joins. It does not affect graph output. Default: 0.
```
Fgrtue, line 604:
```suggestion
- `compact_repeated_processor_chains` — Compacts adjacent repeated processor chains in text output by showing one copy of the chain with a repetition count. This can make parallel pipelines easier to read when the same chain appears many times, for example in joins. It does not affect graph output. Default: 0.

```text
Resize 16 → 1
  FillingRightJoinSide          │
    SimpleSquashingTransform    │ × 16
      Resize 1 → 16
```
```

File: docs/en/sql-reference/statements/explain.md [hunk 11]
```diff
@@ -601,6 +601,13 @@ Settings:
...
+
+```text
+Resize 16 → 1
+  FillingRightJoinSide          │
+    SimpleSquashingTransform    │ × 16
+      Resize 1 → 16
```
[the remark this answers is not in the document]
niyue: Fixed

src/Processors/QueryPlan/IQueryPlanStep.cpp
```diff
@@ -177,11 +193,207 @@ static void doDescribeProcessor(const IProcessor & processor, size_t count, IQue
...
+{
+    return UTF8::computeWidth(reinterpret_cast<const UInt8 *>(line.data()), line.size());
+}
+
+static bool areProcessorsSimilarForExplain(const IProcessor & lhs, const IProcessor & rhs)
+{
+    return lhs.getName() == rhs.getName()
+        && lhs.getInputs().size() == rhs.getInputs().size()
+        && lhs.getOutputs().size() == rhs.getOutputs().size()
+        && lhs.getDescription() == rhs.getDescription();
```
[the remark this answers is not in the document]
niyue: Fixed. When `header = 1`, all headers will be compared for checking repeating chain.

Fgrtue:
The failed test [Stress test (arm_debug)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104662&sha=f507e4e90d55bddc065b594c305184c27bdd2b49&name_0=PR&name_1=Stress%20test%20%28arm_debug%29) is unrelated to this PR and should be fixed by https://github.com/ClickHouse/ClickHouse/pull/106096, so I am merging this pull request.

clickgapai:
Hi @niyue @Fgrtue — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

**Title** (currently `misleading`):
- Current: `Fix  join processor grouping`
- Why: The title 'Fix join processor grouping' implies a fix to join execution/semantics, but the PR only changes `EXPLAIN PIPELINE` text rendering and adds an opt-in `compact_repeated_processor_chains` display option. It is also categorized as Improvement, not a Bug Fix.
- Suggested: `Add `EXPLAIN PIPELINE` option to compact repeated processor chains`

PR #100998, opened 2026-03-28
Keeper getRecursiveChildren request

scanhex12:
### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Added getChildrenRecursive (ListRecursive) request to ClickHouse Keeper and lsr command to clickhouse-keeper-client. This closes https://github.com/ClickHouse/ClickHouse/issues/99916

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.886`

PR #104663, opened 2026-05-12
Fix Keeper termination on get `/keeper/availability_zone` with `quorum_reads=true`

myeongjjun:
**Problem.** When ClickHouse Keeper runs with `<quorum_reads>true</quorum_reads>` and `<placement>` absent, `get /keeper/availability_zone` reaches `processImpl<false>` via the raft commit path, fails the `container.find` lookup, and triggers `onStorageInconsistency` → `std::terminate`. The request is already in the raft log, so every restart replays it and the Keeper terminates again — raft-log poisoning that leaves the Keeper cluster stuck in a restart crash loop.

**Fix.** Drop `keeper_availability_zone_path` from the `preprocess` early-return list in both `Get` and `ListRecursive`. The regular `uncommitted_state.getNode` check now runs and emits a `ZNONODE` `ErrorDelta` when the AZ node is not in the container.

**Why.** The early-return skip was correct only while AZ was served by a dedicated `processImpl` branch and never lived in `storage.container`. Commit 06ff714e2e3 moved AZ into `system_nodes_with_data` behind `if (!keeper_az.empty())` and removed that branch, but the `preprocess` skip was left behind. With `<placement>` absent the node is never registered. The skip prevents `preprocess` from producing the `ZNONODE` `ErrorDelta` that would have short-circuited `commit`, so the request flows on to `container.find` and trips the fail-fast branch. `ListRecursive` was introduced with the same skip line in PR #100998. Removing the skip makes the `quorum_reads=true` response match the `quorum_reads=false` response (`processLocalRequests` → `processImpl<true>`'s graceful `ZNONODE`).

`tests/integration/test_keeper_availability_zone_quorum_reads/` covers unconfigured + configured AZ under `quorum_reads=true`, plus the local-read path.

Fix Keeper termination and restart loop when `get /keeper/availability_zone` is sent with `quorum_reads=true` to a Keeper without `<placement>` configured.

### Version info
- Merged into: `26.5.1.561`
- Backported to: `26.4.3.29`, `26.3.11.27`, `26.2.19.37`, `25.8.24.20`

### DIFF
--- a/src/Coordination/KeeperStorage.cpp
+++ b/src/Coordination/KeeperStorage.cpp
@@ -1887,8 +1887,7 @@ std::list<KeeperStorageBase::Delta> preprocess(
     ProfileEvents::increment(ProfileEvents::KeeperGetRequest);
 
     if (zk_request.path == Coordination::keeper_api_feature_flags_path
-        || zk_request.path == Coordination::keeper_config_path
-        || zk_request.path == Coordination::keeper_availability_zone_path)
+        || zk_request.path == Coordination::keeper_config_path)
         return {};
 
     if (!storage.uncommitted_state.getNode(zk_request.path))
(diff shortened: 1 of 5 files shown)


antonio2368:
Changes make sense. Can you tell me why are you running Keeper with quorum reads?

myeongjjun:
> Changes make sense. Can you tell me why are you running Keeper with quorum reads?

@antonio2368
Honestly, quorum_reads=true wasn't enabled for a specific hard requirement — we kept it enabled as a conservative operational choice.

For context, after a Keeper HA event, ClickHouse → Keeper connections don't automatically rebalance; they stay pinned to whichever Keeper node they failed over to. We were testing an AZ-aware Keeper configuration so that each ClickHouse would prefer a same-AZ Keeper. During that rollout, we hit this bug.

antonio2368 (approved):

PR #99122, opened 2026-03-09
Fix RBAC bypass for DESCRIBE/CREATE TABLE AS via table functions

pufit:
### Changelog category (leave one):
- Critical Bug Fix (crash, data loss, RBAC)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix RBAC bypass that allowed users to obtain table structure via `DESCRIBE TABLE` or `CREATE TABLE AS` on table functions (`mysql()`, `postgresql()`, `sqlite()`, `arrowFlight()`, `jdbc()`, `odbc()`, etc.) without the required source access privileges. For functions that infer schema from remote servers, this also allowed triggering outbound connections (SSRF) without authorization.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

No documentation changes needed — this is a security fix that enforces existing access control semantics.

---

## Summary

`DESCRIBE TABLE table_function(...)` and `CREATE TABLE ... AS table_function(...)` resolved table structure by calling `getActualTableStructure()` without any source access check. The access check (`READ ON MYSQL`, `READ ON SQLITE`, etc.) only existed in `ITableFunction::execute()`, which these code paths never called.

### Affected table functions (SSRF + schema leak on DESCRIBE)
- `mysql()`, `postgresql()`, `arrowFlight()`, `jdbc()`, `odbc()` — outbound connection without authorization
- `sqlite()` — local file read without authorization (fixes https://github.com/ClickHouse/clickhouse-private/issues/47736)

### What changed

- Added `ITableFunction::getActualTableStructureWithAccess()` — bundles `checkSourceAccess()` + `getActualTableStructure()` into a single call
- Used it in `InterpreterDescribeQuery`, `InterpreterCreateQuery`, and `getStructureOfRemoteTable` (local shard path)
- Removed partial per-function access checks from `url()`, `s3()`/`azure()`/`hdfs()`, and `file()` — they only guarded the `structure == "auto"` path and the s3/file versions were weaker (missing URI-level filtering via `checkAccessWithFilter`)

### Test plan
- [ ] New stateless test `04028_describe_table_function_rbac_bypass` — verifies DESCRIBE and CREATE TABLE AS are denied without source grants, and succeed after granting

---

> [!NOTE]
> **High Risk**
[...]

PR #104693, opened 2026-05-12
Store all asked privileges for better RBAC auditing 

alexbakharew:
Expand `system.query_log.used_privileges` coverage to also record privileges that are confirmed via the non-throwing `isGranted` path, so that `READ ON FILE` / `READ ON S3` / `READ ON AZURE` / `READ ON URL` (and any other privilege checked through `ContextAccess::checkAccessWithFilter`) are visible in the audit trail.

#### Background

[#99122](https://github.com/ClickHouse/ClickHouse/pull/99122) centralized source-access checks for `file()` / `s3()` / `azure()` / `url()` table functions and routed them through `ContextAccess::checkAccessWithFilter`, which uses `isGranted` for the granted path:

```cpp
void ContextAccess::checkAccessWithFilter(...) const
{
    if (isGranted(context, flags, parameter))
        return;                              // ← granted short-circuit
    ...
    checkAccess(context, flags, parameter);  // throw if denied
}
```

`isGranted` is `checkAccessImpl<throw_if_denied=false, ...>`. In `ContextAccess::checkAccessImplHelper`, the call to `Context::addQueryPrivilegesInfo(..., granted=true)` was gated on `throw_if_denied`, so the granted branch on non-throwing entry points never recorded. Net effect: `DESCRIBE file('…')`, `CREATE TABLE AS file('…')`, and similar queries no longer reported `READ ON FILE` in `system.query_log.used_privileges`. The privilege check itself still ran — unauthorized users still get `Code: 497 ACCESS_DENIED` on the new branch — so enforcement was unchanged, only the audit trail was incomplete.

#### Change

Drop the `if constexpr (throw_if_denied)` gate around the granted-side `addQueryPrivilegesInfo` call in `checkAccessImplHelper` so every successful access check records, regardless of which public API (`checkAccess` / `checkGrantOption` / `isGranted` / `checkAccessWithFilter`) reached the helper. The denied-side gate is preserved so predicate-style `isGranted` callers do not pollute `missing_privileges` with speculative misses.

#### Measurement

Two `ProfileEvent`s are added to allow the impact to be quantified directly from `system.query_log`:

- `QueryPrivilegesInfoUpdates` — total recordings into `used_privileges` / `missing_privileges` after the change.
- `QueryPrivilegesInfoUpdatesLegacy` — only the recordings from throwing call paths (`checkAccess` / `checkGrantOption`), matching the historical pre-change behavior.

#### Results

Full stateless asan run on this PR's commit `8e6cab260702ec24961ce1c48f36fab083fbcc35` (`Stateless tests (amd_asan…)`, 282,702 `QueryFinish` rows):

| Metric | Value |
|---|---:|
| `legacy_baseline` (old behavior) | 258,781 |
| `new_total` (this PR) | 559,599 |
| `audit_gap_closed` | 300,818 |
| `audit_gap_pct` | 116.24 % |
| `new_calls_per_query` | 1.98 |
| `legacy_calls_per_query` | 0.92 |

About one additional ``std::unordered_set<String>::emplace`` per query on average, distributed across the suite.

#### Tail check

Top queries by additional recordings:

| Query | extra_calls | duration |
|---|---:|---:|
| `SYSTEM START VIEWS` | 614 | 6 ms |
| `SYSTEM START VIEWS` | 582 | 5 ms |
| `SYSTEM PAUSE VIEWS` | 307 | 4 ms |
| `SYSTEM PAUSE VIEWS` | 291 | 5 ms |
| `SELECT count() FROM test.visits` (×7) | 182 each | 10–17 ms |
| `SELECT count() FROM test.hits` (×7) | 134 each | 9–19 ms |
| `kill mutation where …` | 102 | 1609 ms |

All concentrated on legitimate "iterate over N objects and check each" patterns — view fan-out for `SYSTEM` DDL, per-column `SELECT` resolution on wide tables, per-mutation checks for `KILL MUTATION` — exactly the cases where granular per-object audit signal is useful. The worst-case query adds 614 set inserts on a 6 ms query (microseconds of extra work), and nothing in the distribution looks like a planner loop over millions of items.

Aggregation query used to produce the numbers above (works against the CI logs cluster's ``default.query_log_*`` tables):

```sql
WITH '8e6cab260702ec24961ce1c48f36fab083fbcc35' AS commit_to_check
SELECT
    sum(ProfileEvents['QueryPrivilegesInfoUpdates'])       AS new_total,
    sum(ProfileEvents['QueryPrivilegesInfoUpdatesLegacy']) AS legacy_baseline,
    new_total - legacy_baseline                            AS audit_gap_closed,
    round((new_total - legacy_baseline) / nullIf(legacy_baseline, 0) * 100, 2) AS audit_gap_pct,
    count()                                                AS rows,
    round(new_total / nullIf(rows, 0), 3)                  AS new_calls_per_query,
    round(legacy_baseline / nullIf(rows, 0), 3)            AS legacy_calls_per_query
FROM merge('default', '^query_log_')
WHERE commit_sha = commit_to_check
  AND check_name LIKE 'Stateless tests (amd_asan%'
  AND type = 'QueryFinish'
FORMAT Vertical
```

- Improvement

Record privileges in ``system.query_log.used_privileges`` for all granted access checks, including those that go through the non-throwing ``isGranted`` path (``checkAccessWithFilter``). Previously only privileges checked via throwing entry points (``checkAccess`` / ``checkGrantOption``) were recorded, which made ``READ ON FILE`` / ``READ ON S3`` / ``READ ON AZURE`` / ``READ ON URL`` invisible in the audit log for ``DESCRIBE``, ``CREATE TABLE AS``, and similar queries that use ``file()`` / ``s3()`` / ``azure()`` / ``url()`` table functions. Access enforcement is unchanged.

### Version info
- Merged into: `26.6.1.109`


pufit (approved):

clickgapai:
Hi @alexbakharew @pufit — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: The description mentions 'Two ProfileEvents are added' but the sentence is truncated and no ProfileEvents appear in the diff. The description accurately explains the isGranted fix but may reference planned features not yet implemented.
- Suggested addition:

  > Consider removing or completing the truncated 'Measurement' section about ProfileEvents if they are not part of this PR.

PR #104694, opened 2026-05-12
Wire functions stress test to ProcessList for KILL-QUERY-style timeouts

Algunenano:
Each iteration of the functions stress test now creates its own `QueryStatus` via `ProcessList::insert` with `max_execution_time` set to a configurable `--iteration-timeout` (default 15s). The `CancellationChecker` singleton (started for the test's lifetime) flips `QueryStatus::is_killed` once the deadline passes. `request_shutdown` additionally cancels the in-flight iteration through its `QueryStatusPtr` (cancellation goes through `QueryStatus::cancelQuery`, which has its own `cancel_mutex`; we do not touch the worker's `Context::process_list_elem` from the listener thread).

Iterations whose wall-clock duration exceeds `--iteration-too-slow` seconds (default 20s, independent of the timeout) are reported as a new problem `timeout_not_honored`, regardless of how they terminate (success, `MEMORY_LIMIT_EXCEEDED`, `QUERY_WAS_CANCELLED`, `TIMEOUT_EXCEEDED`). The remedy for any flagged function is one of: poll `CurrentThread::isQueryCanceled`, make the per-iteration work cheaper, or reject the offending input shape inside the function itself.

Other cleanups:

- Inner-catch `MEMORY_LIMIT_EXCEEDED || LOGICAL_ERROR` rethrow checks (7 copies) replaced by `isOuterGuardError`, which also covers `QUERY_WAS_CANCELLED` and `TIMEOUT_EXCEEDED`.
- New `S_QUERY_CANCELLED` counter for iterations whose function actually observed the cancellation; surfaced in the report as `query cancelled by timeout`.

Validation runs:

- `--iteration-timeout 0.01 --iteration-too-slow 0.02 --duration 3 --threads 4` (deliberately tiny budget): the test fails with many functions reported under `timeout_not_honored`. Confirms the wiring is reachable end to end.
- Default settings (`--iteration-timeout 15 --iteration-too-slow 20`, `--duration 60 --threads 16`): test passes; `stuck threads: 0`; no `timeout_not_honored` reported.

This lays the groundwork for natural follow-ups: drop the `function_arg_constraints` cap on `arrayResize` and add `isQueryCanceled` polling inside `GatherUtils::resizeConstantSize`. Once those land, the test will validate them without further test-side changes.

...

### Version info
- Merged into: `26.5.1.676`

### DIFF
--- a/src/Functions/tests/gtest_functions_stress.cpp
+++ b/src/Functions/tests/gtest_functions_stress.cpp
@@ -234,6 +234,11 @@ struct Options
         po::store(po::parse_command_line(argc, argv, desc), vm);
         po::notify(vm);
 
+        if (iteration_too_slow_seconds <= iteration_timeout_seconds)
+            throw Exception(ErrorCodes::INCORRECT_DATA,
+                "--iteration-too-slow ({}) must be greater than --iteration-timeout ({}); the slow threshold needs slack above the timeout for the function to notice cancellation",
+                iteration_too_slow_seconds, iteration_timeout_seconds);
+
         for (int pi = 0; pi < P_COUNT; ++pi)
         {
             Problem p = static_cast<Problem>(pi);
--- a/src/Functions/tests/gtest_functions_stress.cpp
+++ b/src/Functions/tests/gtest_functions_stress.cpp
@@ -29,7 +29,9 @@
 #include <DataTypes/DataTypesNumber.h>
 #include <Functions/FunctionFactory.h>
 #include <Functions/FunctionGenerateRandomStructure.h>
+#include <Interpreters/CancellationChecker.h>
 #include <Interpreters/Context.h>
+#include <Interpreters/ProcessList.h>
 #include <IO/WriteHelpers.h>
 #include <Storages/StorageGenerateRandom.h>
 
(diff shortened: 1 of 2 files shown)


Algunenano:
Fix for the stress finding in https://github.com/ClickHouse/ClickHouse/pull/104875, but they need each other

src/Functions/tests/gtest_functions_stress.cpp
```diff
@@ -190,6 +221,8 @@ struct Options
         desc.add_options()
             ("threads", po::value<int>(&num_threads)->default_value(num_threads), "how many instances of the test to run in parallel, -1 for num cores")
             ("duration", po::value<int>(&duration_seconds)->default_value(duration_seconds), "run for this many seconds, -1 to run forever")
+            ("iteration-timeout", po::value<double>(&iteration_timeout_seconds)->default_value(iteration_timeout_seconds), "per-iteration max_execution_time in seconds (fractional allowed)")
+            ("iteration-too-slow", po::value<double>(&iteration_too_slow_seconds)->default_value(iteration_too_slow_seconds), "iterations that take longer than this many seconds are reported as timeout_not_honored (fractional allowed)")
```
antonio2368, line 225: maybe verify it's larger than `iteration-timeout`?

antonio2368 (approved):

alexey-milovidov:
@Algunenano, you IGNORED the CI, and now it is broken.
Keep in mind that CI is always right, and its decision is authoritative. We shall never IGNORE the CI.

PR #104706, opened 2026-05-12
Update Claude skill to continue PRs

alexey-milovidov:

### Version info
- Merged into: `26.5.1.568`


alexey-milovidov:
> require linking an existing issue/PR (or creating one)

This is not required. Creating an issue is like throwing garbage on the floor with the expectation that someone will pick it up. Much better is to not produce garbage in the first place (fix everything you see).

PR #104712, opened 2026-05-12
Fix anyHeavy ignoring default values from sparse columns

### LINKED ISSUE #104432: `anyHeavy` ignores zeros when reading from MergeTree with multiple parts

_Retrospective finding from a historical scan of [PR #58640](https://github.com/ClickHouse/ClickHouse/pull/58640) (merged 2024-03-01). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`anyHeavy(x)` returns wrong results when MergeTree table has multiple parts and one part contains only zeros (default values). For example, with 800 zeros and 200 ones in separate parts, returns 1 instead of 0.

**Root cause:** AggregateFunctionAnyHeavy.cpp:123: `data(place).addManyDefaults(*columns[0], 0, arena)` hardcodes 0 instead of forwarding the `length` parameter, causing default values to not be counted.

**Why we believe this is a bug:** AggregateFunctionAnyHeavy.cpp:121-123 → addManyDefaults ignores length parameter and passes 0 to inner method. This path is triggered when MergeTree reads parts containing only default values via addBatchSparseSinglePlaceFromColumn in IAggregateFunction.h:652-653.

**Affected locations:**
- `src/AggregateFunctions/AggregateFunctionAnyHeavy.cpp:121` — addManyDefaults override ignores length parameter

**Impact:** anyHeavy returns incorrect results when the most frequent value is zero and it appears in a separate MergeTree part from other values. This is a data correctness bug affecting any analytics using anyHeavy on numeric columns with zeros.

Yes — confirmed on current `master` (commit `3cf32af9927`).

### How to reproduce

```sql
-- Test: anyHeavy ignores zeros in MergeTree parts
DROP TABLE IF EXISTS test_anyheavy_zero;
CREATE TABLE test_anyheavy_zero (x UInt64) ENGINE = MergeTree() ORDER BY tuple();
-- Insert 800 zeros in first part (should be heavy hitter with 80%)
INSERT INTO test_anyheavy_zero SELECT 0 FROM numbers(800);
-- Insert 200 ones in second part
INSERT INTO test_anyheavy_zero SELECT 1 FROM numbers(200);
-- Expected: 0 (heavy hitter), Bug shows: 1
SELECT anyHeavy(x) FROM test_anyheavy_zero SETTINGS max_threads = 1;
DROP TABLE test_anyheavy_zero;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/c2c8711a-428e-49e3-b621-28db09b15c9d)

### Expected behavior

```
0
```

```
1
```

### Additional context

**Open risks:**
- Other aggregate functions may have similar issues with addManyDefaults - should audit all overrides

**Suggested fix:** Change line 121 to name the length parameter and forward it: `void addManyDefaults(AggregateDataPtr __restrict place, const IColumn ** columns, size_t length, Arena * arena) const override { data(place).addManyDefaults(*columns[0], length, arena); }`

Found during automated review of [PR #58640](https://github.com/ClickHouse/ClickHouse/pull/58640).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr58640_001`_

### Version info
- Resolved by: #104712
- Merged into: `26.5.1.608`

Algunenano:
The `addManyDefaults` override in `AggregateFunctionAnyHeavy` dropped the `length` parameter and passed `0` to the inner implementation, so runs of default values delivered by the sparse-column code path (`addBatchSparseSinglePlace`) were never counted by the heavy-hitter algorithm.

As a result, `anyHeavy(x)` could return a clearly non-heavy value when the actual heavy hitter was the column default. The classic reproducer is 800 zeros in one MergeTree part and 200 ones in another: `anyHeavy(x)` returned `1` instead of `0`.

Fix: forward `length` to `data(place).addManyDefaults`.

I also audited every other `addManyDefaults` override. `anyHeavy` is the only one where the count actually mattered and was being dropped — the others either forward `length` correctly, have an empty body that is semantically correct for their default value (e.g. `sum` of zeros, `flameGraph` of an empty trace), are idempotent on multiplicity (`any`, `min`, `max`, `uniq`, `singleValueOrNull`, `anyRespectNulls`), or take multiple arguments and therefore never hit the sparse path (only single-argument aggregates take it, per `Aggregator::executeOnBlock`).

Closes #104432

Fixed `anyHeavy` returning a non-heavy value when the most frequent value was the column default and the data lived across multiple `MergeTree` parts (sparse-column read path).

### Version info
- Merged into: `26.5.1.608`


Avogar (approved):

clickgapai:

**Affected code:** `AggregateFunctionAnyHeavy::addManyDefaults` at AggregateFunctionAnyHeavy.cpp:121 — core code present in all supported branches.

**Why this might apply:** Bug was introduced in commit 067aea6eb2b (Jan 2024) when anyHeavy was reimplemented. This predates all supported branches (25.8-26.4). The bug causes WRONG_RESULTS: anyHeavy returns incorrect heavy-hitter when the actual heavy value is the column default in sparse columns. This is P0 severity - silent data corruption.

PR #104737, opened 2026-05-12
Fix a few defects in SettingsConstraints

Algunenano:
A few small but related correctness fixes in `src/Access/SettingsConstraints`:

- `checkImpl(MergeTreeSettings, ...)` now resolves the setting name via `MergeTreeSettings::resolveName` before looking up the constraint, mirroring the `Settings` overload. Previously a `MergeTreeSettings` constraint declared on the canonical name could be bypassed by writing to its alias, because the lookup is a plain hashmap lookup on the (un-resolved) name and the alias cache is only populated when the constraint itself is declared via the alias.
- The `disallowed_values` check in `Checker::check` now honors the `reaction` parameter, matching every other branch (`CONST`, min, max). Previously it threw unconditionally, so the clamp path (secondary queries, `ON CLUSTER` DDL workers, SQL SECURITY DEFINER views) would fail with `SETTING_CONSTRAINT_VIOLATION` instead of silently dropping the change.
- `SettingsConstraints::merge` now iterates `other.settings_alias_cache` rather than its own (a no-op self-iteration).
- The `set` / `get` parameter in `SettingsConstraints.h` is renamed from `allowed_values` to `disallowed_values` to match what the implementation populates.

Fix several defects in setting-constraint handling: `MergeTreeSettings` constraints declared on a setting's canonical name could be bypassed by writing to an alias of that setting; the `disallowed_values` constraint check threw an exception on clamp paths (secondary queries, `ON CLUSTER` workers, `SQL SECURITY DEFINER` views) instead of silently dropping the change.

### Version info
- Merged into: `26.5.1.686`

### DIFF
--- a/src/Access/SettingsConstraints.cpp
+++ b/src/Access/SettingsConstraints.cpp
@@ -403,7 +403,12 @@ bool SettingsConstraints::Checker::check(SettingChange & change,
         return false;
     }
 
-    if (!min_value.isNull() && less_or_cannot_compare(new_value, min_value))
+    /// Track the effective value through clamping so that the disallowed-values loop below
+    /// compares against the post-clamp value. Otherwise an overlap between a clamp target and
+    /// a disallowed entry (e.g. min == disallowed) would let the clamped value through.
+    Field effective_value = new_value;
+
+    if (!min_value.isNull() && less_or_cannot_compare(effective_value, min_value))
     {
         if (reaction == THROW_ON_VIOLATION)
         {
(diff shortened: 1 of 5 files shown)


src/Access/SettingsConstraints.cpp
```diff
@@ -326,10 +326,15 @@ bool SettingsConstraints::checkImpl(const Settings & current_settings,
...
+    /// bypass a constraint declared on the canonical setting name by writing to an alias,
+    /// because the constraint lookup is a plain hashmap lookup on the (still un-resolved) name.
+    std::string_view setting_name = MergeTreeSettings::resolveName(change.name);
+
     Field new_value = getNewValueToCheck(current_settings, change, /*ignore_unchanged_settings=*/false, reaction == THROW_ON_VIOLATION);
     if (new_value.isNull())
         return false;
-    return getMergeTreeChecker(change.name).check(change, new_value, reaction, SettingSource::QUERY);
+    return getMergeTreeChecker(setting_name).check(change, new_value, reaction, SettingSource::QUERY);
 }
```
Ergus, line 338: Could we test this by adding a second profile that declares the `merge_tree_enable_block_number_column` constraint only as a role (not the base profile), assign that role to a user, and confirm the constraint is enforced?

Ergus (approved):
LGTM

PR #104746, opened 2026-05-12
Support Virtual Columns for Granule MinMax Index

Michicosun:

This patch adds the ability to create a granule-level implicit min-max index for virtual columns block number and block offset, which, combined with https://github.com/ClickHouse/ClickHouse/pull/103952, can help filter all data based on the virtual block number and block offset predicates.

### Version info
- Merged into: `26.5.1.692`


EmeraldShift:
How does this compare to the `commit_order` projection index? I suppose this is better suited to bulk filtering `_block_number > X`, while the other is for finding a small number of rows matching some exact block number and offset?

Michicosun:
> How does this compare to the `commit_order` projection index? I suppose this is better suited to bulk filtering `_block_number > X`, while the other is for finding a small number of rows matching some exact block number and offset?

I created a commit-order projection to support table reading in a row commit order without resorting at runtime. It is necessary to omit sorting large data sets. Initially, I made it to support streaming queries on top of the merge tree engine.

EmeraldShift:
I see. I had assumed the `commit_order` projection index only stored `_parent_part_offset` similar to normal projection indexes. If it's got e.g. `*` in the selected columns, now I understand how it can be used for pure sorting, along with #104680 :)

src/Storages/MergeTree/registerStorageMergeTree.cpp
```diff
@@ -844,6 +848,8 @@ static StoragePtr create(const StorageFactory::Arguments & args)
         metadata.add_minmax_index_for_numeric_columns = (*storage_settings)[MergeTreeSetting::add_minmax_index_for_numeric_columns];
         metadata.add_minmax_index_for_string_columns = (*storage_settings)[MergeTreeSetting::add_minmax_index_for_string_columns];
         metadata.add_minmax_index_for_temporal_columns = (*storage_settings)[MergeTreeSetting::add_minmax_index_for_temporal_columns];
+        metadata.add_minmax_index_for_block_number_column = (*storage_settings)[MergeTreeSetting::add_minmax_index_for_block_number_column] && (*storage_settings)[MergeTreeSetting::enable_block_number_column];
```
CheSema, line 860: Bot has valid concerns. 
At create you silently turn off the index. May be an exception is more verbose.

CheSema: Also now it is possible to turn off `enable_block_number_column` when `add_minmax_index_for_block_number_column` is on. Lets forbid this.

CheSema:
I curios about the case when we add this index on the existed table. And how it would work for old parts where block number is not set for the rows. This behavior should be in descriptoin/documentation.

CheSema (approved):

Michicosun:
> I curios about the case when we add this index on the existed table. And how it would work for old parts where block number is not set for the rows. This behavior should be in descriptoin/documentation.

When the block number column is materialized on disk, the index will be written. For parts where the block number column was not materialized, it will be materialized with the part's minimum block number and written to the index with the same value.

src/Storages/MergeTree/registerStorageMergeTree.cpp
Michicosun: instead of throwing supported flags recalculation on alters

src/Storages/AlterCommands.cpp
```diff
@@ -938,6 +949,25 @@ void AlterCommand::apply(StorageInMemoryMetadata & metadata, ContextPtr context)
...
+                effective_settings.applyChange(change);
+                any_mt_setting = true;
+            }
+        }
+        if (any_mt_setting)
+        {
+            metadata.add_minmax_index_for_numeric_columns = effective_settings[MergeTreeSetting::add_minmax_index_for_numeric_columns];
+            metadata.add_minmax_index_for_string_columns = effective_settings[MergeTreeSetting::add_minmax_index_for_string_columns];
+            metadata.add_minmax_index_for_temporal_columns = effective_settings[MergeTreeSetting::add_minmax_index_for_temporal_columns];
+            metadata.add_minmax_index_for_block_number_column = effective_settings[MergeTreeSetting::add_minmax_index_for_block_number_column] && effective_settings[MergeTreeSetting::enable_block_number_column];
```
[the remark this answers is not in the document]
Michicosun: fixed

PR #104751, opened 2026-05-12
Fix use-after-free in `AvroConfluentRowInputFormat`

mstetsyuk:
`AvroConfluentRowInputFormat::getOrCreateDeserializer` built an `AvroDeserializer` on the stack and copied it into `deserializer_cache`. The deserializer holds a `symbolic_skip_fn_map`, and a lambda in  `AvroDeserializer::row_action` which captures a reference into that map (`[&skip_fn = it->second]`). Copying deep-copies the map to a new address but leaves the lambda referencing the original. When the stack local is destroyed, those references dangle, and the next row deserialization invokes a `std::function` over freed memory. We should use `std::move` instead of copying.

Fix segfault due to a use-after-free bug in `AvroConfluentRowInputFormat`.

### Version info
- Merged into: `26.5.1.609`
- Backported to: `26.4.3.18`, `26.3.11.18`, `26.2.19.20`, `25.8.24.14`

### DIFF
--- a/src/Processors/Formats/Impl/AvroRowInputFormat.h
+++ b/src/Processors/Formats/Impl/AvroRowInputFormat.h
@@ -52,6 +52,12 @@ class AvroDeserializer
 public:
     AvroDeserializer(const Block & header, avro::ValidSchema schema, bool allow_missing_fields, bool null_as_default_, const FormatSettings & settings_);
     AvroDeserializer(DataTypePtr data_type, const std::string & column_name, avro::ValidSchema schema, bool allow_missing_fields, bool null_as_default_, const FormatSettings & settings_);
+
+    AvroDeserializer(const AvroDeserializer &) = delete;
+    AvroDeserializer & operator=(const AvroDeserializer &) = delete;
+    AvroDeserializer(AvroDeserializer &&) = default;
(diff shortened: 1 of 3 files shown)


src/Processors/Formats/Impl/AvroRowInputFormat.h
```diff
@@ -52,6 +52,12 @@ class AvroDeserializer
 public:
     AvroDeserializer(const Block & header, avro::ValidSchema schema, bool allow_missing_fields, bool null_as_default_, const FormatSettings & settings_);
     AvroDeserializer(DataTypePtr data_type, const std::string & column_name, avro::ValidSchema schema, bool allow_missing_fields, bool null_as_default_, const FormatSettings & settings_);
+
+    AvroDeserializer(const AvroDeserializer &) = delete;
+    AvroDeserializer & operator=(const AvroDeserializer &) = delete;
+    AvroDeserializer(AvroDeserializer &&) = default;
+    AvroDeserializer & operator=(AvroDeserializer &&) = delete;
```
Ergus, line 59: I am a bit old school, I really like this part ;).

Ergus (approved):
LGTM. I am just wondering if we can add a test for this somehow to ensure we exercise this path.

PR #104753, opened 2026-05-12
darwin fast test: skip more tests on Darwin ARM

leshikus:
Extends `ci/defs/darwin.skip` with tests that failed in a previous Darwin ARM run. Entries are filtered to tests that currently exist under `tests/queries/0_stateless/` so the skip list does not reference tests that have been removed or not yet merged. As a small drive-by, trailing whitespace on 24 pre-existing entries is stripped so the file round-trips cleanly through tooling.

Skip tests on Darwin ARM that are known to fail.

### Version info
- Merged into: `26.5.1.635`


maxknv:
What is the plan for the job? Can we enable it after the extended list is merged?

leshikus:
@maxknv the plan is to fix stack trace collection in `clickhouse-test` so reliability bugs can be reported and fixed, see https://github.com/ClickHouse/ClickHouse/pull/104771/ ; this particular exclude list won't help with reliability problems - even select_1 test fails sometimes

maxknv:
> @maxknv the plan is to fix stack trace collection in `clickhouse-test` so reliability bugs can be reported and fixed, see #104771 ; this particular exclude list won't help with reliability problems - even select_1 test fails sometimes

Why do we extend the list then? All these tests not supported on macOS?

leshikus:
@maxknv - 1) the tests are supported, 2) the smaller list makes the run faster; 3) the smaller list makes run more reliable for intermittent errors -  the goal is to enable the run; 4) it makes some sense to exclude those tests which fail first rather than just exclude random tests

maxknv (approved):

Issue #104122, opened 2026-05-05
RFC: Vector Search Improvements and vector similarity index 2.0


shankar-iyer:
**Current Status (26.4)**

ClickHouse implements an HNSW in-memory index using the `usearch` library. Scalar quantization options are available and indeed our default is `BFloat16` quantization which provides recall matching full `Float32`. Other HNSW parameters like `M` and `ef_construction`/`ef_search` are configurable to test index build speeds v/s recall and select appropriate values.

HNSW is unbeatable in search latency and recall!

Now to the well documented problems with HNSW - 1) HNSW requires memory for the vectors and the graph 2) Incremental index update requires full rebuild and full write of the new index (When 2 `parts` merge in ClickHouse, a new index is built for the merged `part`).

There are 2 other related aspects to discuss-  3)  When we scale to billion(s) of vectors, segmenting or partitioning the vector index across the memory of the replicas is definitely needed. A single vector index on a single VM cannot serve vector search on billions of vectors. 4) How is the support for additional filtering implemented (_pre/post/in-filtering/hybrid search_ etc) ?. Additional filtering refers to predicates in the `WHERE` clause in a vector similarity search query.

For deployments hitting RAM limits  - `binary quantization ` is a good solution. ClickHouse supports binary quantization with `oversampling` and `rescoring+reranking`.

**Making billion vector scale indexing work with the existing vector index (HNSW) gives two choices:**
- Architecture 1: A single, high memory VM (300-500GB+) with binary quantization index and `M=16`. Remaining free memory in OS file cache or local SSD will cache the full precision vectors needed for rescoring/reranking. The high amount of DRAM will make this very expensive.
- Architecture 2: Use the _distributed index analysis_ feature to load indexes (full precision or quantized) of different parts on different replicas. Vector index subsets/segments are searched on different replicas and the final set of rows are fetched on the co-ordinator. Thus each VM in the cluster need not be sized big enough to fit the entire index in memory. This solution scales with the number of nodes.

**Claude Feedback**
_Reasonable framing. Two friction points the RFC should acknowledge_:

_Even M=16 binary HNSW for 1B 1024-d vectors is ~10-15 GB for the index
alone; the vectors the rescorer needs are 4 TB at f32 or 1 TB at bf16.
Holding "remaining free memory in OS file cache or local SSD" is doing a
lot of load-bearing work in that sentence. In practice rescoring will be
IO-bound on cold queries unless you size the SSD cache carefully_.

[...]

PR #104690, opened 2026-05-12
Add `UntrackedMemory` asynchronous metric

mstetsyuk:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Add an asynchronous metric `UntrackedMemory` reporting the sum of all threads' untracked memory counters. This value is also included in `MEMORY_LIMIT_EXCEEDED` exception messages.

### Version info
- Merged into: `26.6.1.339`

PR #104754, opened 2026-05-12
Vector similarity pk partial filter

zex-hyd:
Fixes the case where `vector_similarity` skip indexes were not used when a primary key condition left only a subset of marks in a part. Previously, that situation skipped vector index analysis entirely, so reads could not use the HNSW granule to narrow work the way the index is intended.

This PR keeps running vector index analysis under partial PK pruning and passes surviving PK `MarkRanges` into USearch `filtered_search` as a predicate, so graph traversal only keeps neighbors whose rows map to PK-allowed data marks.

Motivated by #104122. Typical pattern: a long-retention table (for example tickets by time in the primary key) plus a narrow time predicate and `ORDER BY` vector distance with `LIMIT` — partial PK on the part plus vector search.

When other skip indexes (minmax, set, text) run earlier in the pipeline and narrow `MarkRanges`, the vector path still sees that reduced `ranges` input, so behavior stays conjunctive without extra glue.

`limit` is oversampled with `vector_search_index_fetch_multiplier` when `row_filter` is set, using the same rule as `additional_filters_present` and rescoring. `filtered_search` can still return fewer than `limit` hits if the PK slice removes almost all rows inside a granule.

Stateless: `04217_vector_similarity_pk_partial_pruning`.

- Performance Improvement

Vector similarity skip indexes are used when the primary key leaves only a subset of marks in a part: neighbors are filtered with USearch `filtered_search` to PK-allowed data marks instead of skipping the vector index for that case.

---

> [!NOTE]
> **Medium Risk**
> Changes MergeTree mark filtering and vector-search hint propagation during query execution; mistakes could cause incorrect pruning or `_distance` ordering behavior under mixed PK/skip-index conditions.
> 
> **Overview**
> Enables `vector_similarity` skip index usage even when primary key pruning leaves only a subset of marks in a part, by passing PK-surviving row intervals into USearch `filtered_search` and restricting ANN hits to those ranges.
> 
> Updates vector-search hint handling to **merge per-granule ANN results into part-global row offsets**, conservatively dropping `_distance` hints if any non-empty granule lacks distances, and adds per-`index_mark` caching to avoid repeated ANN work when marks are referenced by multiple PK ranges.
> 
> Adds new `VectorSearchUtils` types (`GranuleRowFilter`, `ANNSearchOverrides`) and adjusts tests to expect index usage under partial PK pruning, plus a new stateless regression test (`04217_vector_similarity_pk_partial_pruning`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 082abd37613ee9945aca26aa69f3bdb4cc27f978. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/Storages/MergeTree/MergeTreeIndexVectorSimilarity.cpp
```diff
@@ -111,6 +116,43 @@ String joinByComma(const T & t)
...
+    const size_t rows_in_granule = index_granularity->getRowsCountInRange(base_mark, end_mark);
+
+    const auto key_u64 = static_cast<UInt64>(key);
+    if (key_u64 >= rows_in_granule)
+        return false;
+
+    const size_t global_row = granule_row_base + static_cast<size_t>(key_u64);
+    const MarkRange mark_for_row = index_granularity->getMarkRangeForRowOffset(global_row);
+    chassert(mark_for_row.begin + 1 == mark_for_row.end);
+    const size_t data_mark = mark_for_row.begin;
```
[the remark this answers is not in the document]
zex-hyd: I addressed this comment by resolving data_mark inside base_mark, end_mark only (binary search on getMarkStartingRow), avoiding global getMarkRangeForRowOffset.

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
```diff
@@ -2118,7 +2113,19 @@ std::pair<MarkRanges, RangesInDataPartReadHints> MergeTreeDataSelectExecutor::fi
...
+                    {
+                        pk_ranges_single.push_back(ranges[i]);
+                        row_filter.emplace(IMergeTreeIndexCondition::GranuleRowFilter{
+                            part->index_granularity.get(),
+                            std::move(pk_ranges_single),
+                            index_mark,
+                            skip_index_granularity});
+                    }
+
+                    read_hints.vector_search_results = condition->calculateApproximateNearestNeighbors(granule, row_filter);
```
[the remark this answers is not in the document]
zex-hyd: I implemented accumulate and part row offset merge for vector_search_results so _distance hints stay correct when multiple index_mark / ranges are processed

tests/queries/0_stateless/04217_vector_similarity_pk_partial_pruning.sql
```diff
@@ -0,0 +1,181 @@
...
+        SELECT arraySort(groupArray(id))
+        FROM
+        (
+            SELECT id
+            FROM tab_pk_partial
+            WHERE id >= 6
+            ORDER BY L2Distance(vec, reference_vec) ASC
+            LIMIT 3
+            SETTINGS use_skip_indexes = 1
+        )
```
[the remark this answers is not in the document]
zex-hyd: I decide not to add query_log here to keep the test lightweight, while result equality already guards correctness

src/Storages/MergeTree/MergeTreeIndexVectorSimilarity.cpp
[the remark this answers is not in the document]
zex-hyd: This already removed the full-part lower_bound from each predicate check and keeps lookup local to the current skip-index granule

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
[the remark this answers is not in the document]
zex-hyd: vector search is now executed once per index_mark with aggregated pk_ranges for that mark, and the result is reused across overlapping ranges

tests/queries/0_stateless/04217_vector_similarity_pk_partial_pruning.sql
```diff
@@ -0,0 +1,199 @@
...
+WHERE id >= 6
+ORDER BY L2Distance(vec, [toFloat32(0.), toFloat32(2.)]) ASC
+LIMIT 3
+SETTINGS use_skip_indexes = 1, log_comment = '04217-vector-index-path'
+FORMAT Null;
+
+SYSTEM FLUSH LOGS query_log;
+
+SELECT 'vector_index_path_used';
+SELECT ProfileEvents['USearchSearchCount'] > 0
```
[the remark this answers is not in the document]
zex-hyd: I kept this test focused on behavior only and avoided read_rows or read_marks checks because those are noisy under randomized CI settings and can become flaky. I will include benchmark numbers in the PR description to document the performance impact separately.

alexey-milovidov:

zex-hyd:
Hi @alexey-milovidov @shankar-iyer, the branch is up to date with master and all the checks has passed!
Would appreciate a review when you get a chance!

tests/queries/0_stateless/04217_vector_similarity_pk_partial_pruning.sql
```diff
@@ -0,0 +1,199 @@
+-- Tags: no-fasttest, no-ordinary-database
+-- vector_similarity with partial PK pruning: use_skip_indexes 1 vs 0 must return the same row ids.
+-- Also: ORDER BY Date + range on that key + vector ORDER BY LIMIT.
+
+SET allow_experimental_vector_similarity_index = 1;
```
rschu1ze, line 5: Someone looking at this test in a few weeks from now will have no clue what is being tested and why. Please add comments.

zex-hyd: Thansk for the review @rschu1ze !
I added short comments to 04217_vector_similarity_pk_partial_pruning.sql and the related 02354 cases in this PR

shankar-iyer:
@zex-hyd I am reviewing the PR.

src/Storages/MergeTree/MergeTreeIndexVectorSimilarity.cpp
```diff
@@ -557,16 +612,29 @@ NearestNeighbours MergeTreeIndexConditionVectorSimilarity::calculateApproximateN
...
+            /// The way to do this in USearch is to call index_dense_gt::change_expansion_search. Unfortunately, this introduces a need to
+            /// synchronize index access, see https://github.com/unum-cloud/usearch/issues/500. As a workaround, we extended USearch' search method
+            /// to accept a custom expansion_add setting. The config value is only used on the fly, i.e. not persisted in the index.
+            return index->search(parameters->reference_vector.data(), limit, USearchIndex::any_thread(), false, expansion_search);
+        }
+
+        /// Copy filter into the closure: filtered_search may retain the predicate until the search completes, but a
+        /// reference capture would be unsafe if evaluation ever outlived the local optional in the caller.
+        auto predicate = [rf = row_filter.value()](USearchIndex::vector_key_t key) { return granuleLocalKeyAllowed(key, rf); };
+        return index->filtered_search(parameters->reference_vector.data(), limit, std::move(predicate), USearchIndex::any_thread(), false, expansion_search);
```
shankar-iyer, line 627: Thank you for introducing this call to usearch API! This will help us in more things in future.

src/Storages/MergeTree/MergeTreeIndexVectorSimilarity.cpp
```diff
@@ -111,6 +116,55 @@ String joinByComma(const T & t)
     std::unreachable();
 }
 
+bool granuleLocalKeyAllowed(USearchIndex::vector_key_t key, const IMergeTreeIndexCondition::GranuleRowFilter & filter)
```
shankar-iyer: This routine is called by `usearch` as the traversal happens, so we need it super efficient. What's your opinion on -

1. We enable the _filtered search_ wiring only if the skip index has a single granule - that is the skip index's granularity is 100M. That is our recommendation anyways. Most of the heavy code in this routine right now is the binary search to convert from the skip index granule to part granule
2. During `GranuleRowFilter` consutrction , populate a simple `small_vector`/`vector` that contains starting `_part_offset` and end `_part_offset` of selected PK ranges. Hopefully there is only a single PK range. This is one time cost
3. `granuleLocalKeyAllowed()` only does a quick lookup of the vector to see if `key` can be considered or rejected?

src/Storages/MergeTree/MergeTreeIndices.h
```diff
@@ -166,12 +170,25 @@ class IMergeTreeIndexCondition
...
+        size_t skip_index_granularity;
+    };
+
     /// Special method for vector similarity indexes:
     /// Returns the N nearest neighbors of a reference vector in the index granule.
     /// The nearest neighbors are returned as row positions.
     /// If VectorSearchParameters::return_distances = true, then the distances are returned as well.
-    virtual NearestNeighbours calculateApproximateNearestNeighbors(MergeTreeIndexGranulePtr /*granule*/) const
+    /// row_filter restricts keys to pk_ranges; fewer than requested hits is allowed.
+    virtual NearestNeighbours calculateApproximateNearestNeighbors(
```
shankar-iyer, line 185: How about moving  `GranuleRowFilter` to `VectorSearchUtils.h `and wrapping it for future extension -

```
 struct ANNSearchOverrides/ANNSearchExtras/ANNSearchModifiers
 {
      /// Restrict graph traversal to keys/_part_offsets passing this filter.
      std::optional<GranuleRowFilter> row_filter;

      /// More in future...
 }
```

And forward declare that `struct` and use reference in `MergeTreeIndices.h`

src/Storages/MergeTree/MergeTreeIndexVectorSimilarity.cpp
```diff
@@ -557,16 +612,29 @@ NearestNeighbours MergeTreeIndexConditionVectorSimilarity::calculateApproximateN
         granule->scalar_kind, ErrorCodes::INCORRECT_QUERY, "reference vector in the SELECT query");
 
     size_t limit = parameters->limit;
-    if (parameters->additional_filters_present || is_rescoring)
-        /// Additional filters mean post-filtering which means that matches may be removed. To compensate, allow to fetch more rows by a factor.
-        /// Similarly, if rescoring is on, fetch more neighbours from the index and pass them for the final re-ranking by ORDER BY ... LIMIT.
+    if (parameters->additional_filters_present || is_rescoring || row_filter.has_value())
+        /// Fewer hits after post-filter, rescoring, or PK row filter: raise fetch limit by index_fetch_multiplier.
```
shankar-iyer: Nice to call out the `PK row filter`

src/Storages/MergeTree/MergeTreeIndexVectorSimilarity.cpp
zex-hyd: Glad to hear that!!

src/Storages/MergeTree/MergeTreeIndexVectorSimilarity.cpp
zex-hyd: @shankar-iyer Good point. Binary search on the traversal path is too heavy.

I'll stash the selected PK ranges as part offset intervals during construction, probably just one range in practice. Then the lookup is trivial.

On the single-granule thing: 
I'd rather not add that constraint. The code already handles multi-granule, and with the precomputed ranges the hot path is fast anyway. It'd also silently disable filtered_search for non-standard granularity configs. 

Let me know if you see a specific case where it's still needed.

src/Storages/MergeTree/MergeTreeIndices.h
zex-hyd: Fixed! I moved GranuleRowFilter to VectorSearchUtils.h and wrapped it in ANNSearchOverrides with std::optional<GranuleRowFilter> row_filter. Should be easier to extend now.

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
```diff
@@ -2093,9 +2090,14 @@ std::pair<MarkRanges, RangesInDataPartReadHints> MergeTreeDataSelectExecutor::fi
...
-                        std::max(ranges[i].begin, index_mark * skip_index_granularity),
-                        std::min(ranges[i].end, (index_mark + 1) * skip_index_granularity));
+                    const size_t data_range_begin = std::max(ranges[i].begin, index_mark * skip_index_granularity);
+                    const size_t data_range_end = std::min(ranges[i].end, (index_mark + 1) * skip_index_granularity);
+                    if (data_range_begin >= data_range_end)
+                    {
+                        ++current_granule_num;
+                        continue;
+                    }
+                    MarkRange data_range(data_range_begin, data_range_end);
```
shankar-iyer, line 2950: What is this code change for ? It is in the bulk filtering code path

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
```diff
@@ -2187,6 +2313,34 @@ std::pair<MarkRanges, RangesInDataPartReadHints> MergeTreeDataSelectExecutor::fi
...
+                    sorted_distances[p] = distances_out[permutation[p]];
+                }
+                merged_vector_search_results->rows = std::move(sorted_rows);
+                distances_out = std::move(sorted_distances);
+
+                read_hints.vector_search_results = std::move(merged_vector_search_results);
+            }
+            else
+                read_hints.vector_search_results.reset();
+        }
```
shankar-iyer: Too many lines of code change in lines 2100-2350. I need some time to understand. Are all these code changes for handling multiple skip index granules?

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
zex-hyd: Hi @shankar-iyer ! Thanks for review again!

This is a new guard added as part of this PR. In the partial-PK path, ranges[i] may not fully overlap a skip-index granule, so the intersection can be empty. Without this check, we'd construct a MarkRange with begin >= end which is invalid. 

This guard is intentionally behavior-preserving for non-empty ranges. It only skips the empty intersection case. I added the same guard consistently across the bulk filtering, vector, and non-vector paths.

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
zex-hyd: Thanks for checking this. I will add some more comments in a later commit.

The logic here is basically handle the case where the same `index_mark` is reached from multiple PK ranges. I need to merge those PK ranges first and run one filtered ANN on the merged filter. If I run ANN separately per sub-range, I can end up with inconsistent top-k and hint behavior.

`remaining_uses_by_index_mark` is only for memory cleanup and not required for correctness. 

I am also happy to follow up with a cleanup commit that groups the per-mark state into one struct if that would make this easier to read!

zex-hyd: @shankar-iyer I pushed a follow-up commit with inline comments in this block for readability. The code should now be easier to understand!

zex-hyd:
I also precomputed granule_row_base/end/span in GranuleRowFilter so the hot path predicate has no virtual dispatch, and fixed a potential iterator stall in the bulk filtering path on empty range intersections.

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
shankar-iyer: Thanks! I am still on it, trying to figure out how to minimize the review surface.

zex-hyd: > Thanks! I am still on it, trying to figure out how to minimize the review surface.

@shankar-iyer 
I looked carefully at 2100–2350 again in the past few days and was trying to do a cleanup here, and I think the main issue is that GranuleRowFilter construction, the per-mark ANN cache, and hints accumulation are all inlined together. 

I can pull these into new helpers functions(such as buildGranuleRowFilterForPartialPk, PartialPkVectorSearchCache, and PartialPkVectorSearchHints), which should bring the vector branch in the main loop down significantly, while causing no behavior change. 

Should I go ahead and finish this commit? Or do you think any other direction should I could work on to minimize the review surface?

src/Storages/MergeTree/MergeTreeIndexVectorSimilarity.cpp
```diff
@@ -111,6 +116,25 @@ String joinByComma(const T & t)
...
+{
+    if (filter.granule_row_end <= filter.granule_row_base)
+        return false;
+
+    const auto key_u64 = static_cast<UInt64>(key);
+    if (key_u64 >= filter.granule_row_span)
+        return false;
+
+    const size_t part_row = filter.granule_row_base + static_cast<size_t>(key_u64);
+    for (const auto & [row_begin, row_end] : filter.allowed_part_row_ranges)
```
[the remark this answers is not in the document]
zex-hyd: I'll add a single-interval fast path for the common case and fall back to std::upper_bound for multiple intervals, using the sorted invariant that's already guaranteed during construction.

alexey-milovidov:

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
zex-hyd: @shankar-iyer I went ahead with the refactor we discussed. Extracted GranuleRowFilter construction into buildGranuleRowFilterForPartialPk, the per-mark cache into PartialPkVectorSearchCache, and hints accumulation into PartialPkVectorSearchHints. The vector branch in filterMarksUsingIndex is now around 50 lines of orchestration. Also updated annindexes.md to reflect that partial PK no longer forces brute-force search. 

Would appreciate another look when you get a chance.

zex-hyd:
@alexey-milovidov @shankar-iyer the only remaining CI failure is 04103_user_network_bandwidth_throttler which is marked as flaky and unrelated to this PR. Everything else is passing. Would appreciate a review when you get a chance!

---
Update: Just merged the latest master with commit for fixing 04103_user_network_bandwidth_throttler, will keep tracking the CI

---
Latest Update: this flaky check has been fixed and all checks has passed now!

zex-hyd:
@rschu1ze  @shankar-iyer I went ahead with the refactor we discussed. Extracted GranuleRowFilter construction into buildGranuleRowFilterForPartialPk, the per-mark cache into PartialPkVectorSearchCache, and hints accumulation into PartialPkVectorSearchHints. The vector branch in filterMarksUsingIndex is now around 50 lines of orchestration. Also updated annindexes.md to reflect that partial PK no longer forces brute-force search. (reference our earlier discussion: https://github.com/ClickHouse/ClickHouse/pull/104754#discussion_r3301722478)

Would appreciate another look when you get a chance.

zex-hyd:
@alexey-milovidov  @shankar-iyer @rschu1ze all checks are passing and there are no conflicts with master. The PR is ready for another look whenever you have time.

zex-hyd:
@alexey-milovidov @shankar-iyer @rschu1ze pushed a few more fixes. 

The main one is a mark mapping bug under adaptive granularity: ANN hits were mapped via countMarksForRows which is wrong when marks have variable row counts, now using getMarkRangeForRowOffset instead. Also added a regression test for this case and cleaned up some comments.

This PR has been open for a while now, I would really appreciate a review or any feedback on what's needed to move it forward!

alexey-milovidov:
Pushed two test-only commits to move this forward (verified end-to-end against the PR build, `26.7.1`):

**1. Fix the `02354_vector_search_pre_and_post_filtering` CI failure** (the only failing test on the current head). Two independent causes:
- The `.sql` printed the label `... 1 part partially selected by PK` while the `.reference` expected `..., expect index usage`. That query now uses the vector index under partial PK pruning (this PR's point) — `USearchSearchCount` is `1` — so `expect index usage` is correct. Aligned the `.sql` literal.
- `#105036` made `explain_query_plan_default = 'pretty'` the default, which renders `EXPLAIN` plans with box-drawing characters that `trimLeft` does not strip, breaking the grepped `Description:`/`Granules:` lines. Master added `SET explain_query_plan_default = 'legacy';` to this exact test; did the same here.

**2. Strengthen `04217` per the AI review's "Request changes"** (the two live `clickhouse-gh` threads):
- `pk_partial_ordered_ids_match_...`: assert the *ordered* ids of the index path vs the brute-force baseline (`use_skip_indexes = 0`) instead of the sorted set.
- Reworked the adaptive-granularity table to use non-monotonic vectors so id order ≠ distance order (for `reference_vec = [0, 0]` the exact top-3 is `[8, 10, 6]`), and added `pk_partial_adaptive_ordered_ids_match_...` plus an explicit `pk_partial_adaptive_expected_ordered_top3_ids` assertion — the regression guard for row-to-mark mapping under adaptive granularity that the `USearchSearchCount` check alone cannot provide.

The remaining AI major (a before/after benchmark for the `Performance Improvement` contract) is the changelog-category discussion already raised in [this thread](https://github.com/ClickHouse/ClickHouse/pull/104754#discussion_r3243135901) and is a call for the reviewers, not a test change.

Note: the branch is ~8 days behind `master`. A master merge is textually clean, but it touches `filterMarksUsingIndex` — master added a `readable_ranges` argument to `reader.read(...)` (`#`... signature change) in the same function this PR rewrites — so please update the branch and rebuild locally rather than relying on the clean auto-merge alone.

alexey-milovidov:
Pushed a merge of the latest `master` (the branch was 9 days behind and red on `04401_distributed_with_fill_const_to_scalar`, which is unrelated to this PR and already fixed on `master` by 911eea988a8; the other failure, `test_replicated_database/test.py::test_replicated_table_structure_alter`, is a known flaky tracked in #110036).

**Conflict resolution** (`MergeTreeDataSelectExecutor.cpp`, `filterMarksUsingIndex`): `master` rewrote the vector-search hint block to accumulate per-granule ANN hits into part-global row offsets (follow-up of the rescoring fast-path #105591). This PR's `PartialPkVectorSearchHints::tryMergeGranuleHintsOnce` already implements that accumulation (and handles a granule without distances more conservatively: it drops `_distance` hints for the whole part instead of resetting and re-accumulating), so I kept the PR side, and carried over `master`'s release-mode check by upgrading the `chassert` on the rows/distances size mismatch to a `LOGICAL_ERROR` exception.

Verified on the merged branch: it builds, and `04217_vector_similarity_pk_partial_pruning`, `02354_vector_search_pre_and_post_filtering`, `02354_vector_search_rescoring*` and `02354_vector_search_queries` all pass locally.

---

**Focused before/after measurement for the partial-PK path** (requested by the AI review, and relevant to the efficiency thread):

Setup: this PR's CI release binary (26.7.1, aarch64), 1M rows × 64-dim vectors, single part, `vector_similarity('hnsw', 'L2Distance', 64) GRANULARITY 100000000` (single skip-index granule, as recommended), `index_granularity = 8192`. Query shape: `SELECT id FROM t WHERE id >= lo AND id < hi ORDER BY L2Distance(vec, ref) LIMIT 10`. "Before" is emulated with `use_skip_indexes = 0` (the pre-PR behavior for partial PK was to skip the vector index and brute-force the PK-surviving rows), "after" is the default index path via `filtered_search`. Median of 15 runs each, warm cache:

| PK slice (rows surviving) | before: brute force | after: `filtered_search` | rows read before → after |
|---|---|---|---|
| 80% (800k) | 12 ms | **6 ms** | 802,816 → 73,728 |
| 10% (100k) | **12 ms** | 20 ms | 114,688 → 49,152 |
| 2% (20k) | **4 ms** | 44 ms | 32,768 → 24,576 |

So the improvement is real for wide partial-PK slices (the motivating #104122 shape: most of the part survives a coarse time predicate), where the index path is ~2x faster and reads ~11x fewer rows. But for highly selective PK slices, `filtered_search` is considerably slower than the old brute-force fallback (HNSW traversal rejects most neighbors and oversampling grows the search), a ~10x regression at 2% selectivity in this setup. A selectivity-based fallback (skip the vector index when the PK-surviving fraction of the granule is small) may be worth considering — leaving that design call to the author and reviewers.

alexey-milovidov:
Pushed `f1c14676075..4d921062aa1`: merged the latest `master` (the branch was ~3300 commits behind) and addressed both minimum required actions of the AI review's `Request changes` verdict.

**1. Selectivity fallback for `auto` mode (blocker).** New setting `vector_search_min_surviving_pk_fraction` (default `0.5`): the minimum fraction of the marks of a part that must survive primary key analysis for the `vector_similarity` index to be used on a partially pruned part. Below the threshold, `filterMarksUsingIndex` skips the index for the part exactly as before this feature and the surviving rows are scanned exactly. This keeps the ~2x win of the 80%-surviving case from the benchmark above and the old behavior for the 10% / 2% cases that regressed ~1.7x / ~10x. `0` always uses the index, `1` only when the primary key prunes nothing. (`vector_search_filter_strategy` is not visible in `filterMarksUsingIndex` — `prefilter` is already resolved in the plan — so the threshold applies to both `auto` and `postfilter`; setting it to `0` forces the previous behavior of this branch.)

**2. Cross-granule assertion in `04217_vector_similarity_pk_partial_pruning` (test finding).** Added a case that compares the ordered `(id, distance)` pairs of a multi-granule query against the `use_skip_indexes = 0` baseline. It uses a mark-aligned filter (`id <= 2 OR id >= 6`) so both skip-index granules contribute hits and nothing is removed by the residual predicate — that makes the index result exactly comparable to brute force, so a dropped granule or a mis-paired `_distance` fails the test. Also added coverage of the new threshold itself.

**3. The `02354_vector_search_nonindexed_columns` CI failure is fixed by the above.** That test (added on `master` by `605222d9251`/`0d5e67f1a19`) asserts that with `key IN (0, 30, 60, 90)` the implicit `PREWHERE` is restored; the branch produced row hints there and printed `Filter column:` instead of `Prewhere filter column:` in 32 jobs. Only 4 of 34 marks survive, which is below the default threshold, so the master behavior and the master version of the test are restored (same for `04490_vector_search_rescoring_prewhere_eligibility`).

**4. Recall of the partial-PK index path — worth a design decision before this merges.** The primary key ranges are mark-granular, so the index searches a *superset* of the qualifying rows and the exact predicate then removes some of its top-`k` hits. When the boundary mark holds many non-qualifying rows, that can remove nearly all of them. On the 04217 table (`index_granularity = 3`, 12 rows), with the threshold disabled:

```sql
SELECT id FROM tab_pk_partial WHERE id >= 9 ORDER BY L2Distance(vec, [0., 2.]) LIMIT 3
-- index path (vector_search_min_surviving_pk_fraction = 0): []
-- brute force  (use_skip_indexes = 0):                      [9, 10, 11]
```

(`id >= 9` keeps the mark holding ids 6, 7, 8 as well, since the primary key range of the preceding mark ends at the boundary value; the index returns 6, 7, 8 and `id >= 9` then drops all three.) The same effect shows in `02354_vector_search_rescoring`, where `WHERE id > 3 ... LIMIT 3` returns `4, 5` on the index path instead of `4, 5, 6` — I made that visible in the test instead of hiding it behind `use_skip_indexes = 0`. With production granularities only the boundary marks are affected, and `vector_search_index_fetch_multiplier` mitigates it, but it is a change of the default result quality for primary-key-filtered vector searches that @rschu1ze / @shankar-iyer should sign off on. Raising the default threshold, or fetching `limit + rows_in_boundary_marks` neighbours, would be the natural knobs.

Verified locally on a build of this branch: `04217_vector_similarity_pk_partial_pruning`, all `02354_vector_search_*` and `04490_vector_search_rescoring_prewhere_eligibility` pass (the only failure, `02354_vector_search_distributed_index_analysis`, needs a `parallel_replicas` cluster that my single-node server does not define).

src/Storages/MergeTree/MergeTreeDataSelectExecutor.cpp
```diff
@@ -2488,6 +2490,210 @@ MarkRanges MergeTreeDataSelectExecutor::markRangesFromPKRange(
...
+            if (granule_row_filter.allowed_part_row_ranges.empty())
+            {
+                search_result_it = vector_search_results_by_index_mark.emplace(index_mark, NearestNeighbours{}).first;
+            }
+            else
+            {
+                ANNSearchOverrides ann_overrides{.row_filter = std::move(granule_row_filter)};
+                search_result_it = vector_search_results_by_index_mark.emplace(
+                    index_mark,
+                    condition->calculateApproximateNearestNeighbors(granule, ann_overrides)).first;
```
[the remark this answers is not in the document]
alexey-milovidov: Addressed in 4d921062aa1: added `vector_search_min_surviving_pk_fraction` (default `0.5`). If less than that fraction of the marks of a part survives primary key analysis, the vector index is skipped for the part and the surviving rows are scanned exactly, as before this feature — so the 10% and 2% cases from the benchmark keep the old timings, while the 80% case keeps the new 2x win. Set it to `0` to always use `filtered_search`.

tests/queries/0_stateless/04217_vector_similarity_pk_partial_pruning.sql
```diff
@@ -0,0 +1,387 @@
...
+SELECT 'pk_multi_granule_vector_index_used';
+SELECT max(ProfileEvents['USearchSearchCount'] > 0)
+FROM system.query_log
+WHERE current_database = currentDatabase()
+    AND type = 'QueryFinish'
+    AND log_comment = '04217-pk-multi-granule'
+    AND event_date >= yesterday()
+    AND event_time >= now() - 600;
+
+SELECT 'pk_multi_granule_same_row_count_as_without_skip_indexes';
```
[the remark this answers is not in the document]
alexey-milovidov: Addressed in 4d921062aa1: `pk_multi_granule_ordered_ids_and_distances_match_exact_knn_without_skip_indexes` compares the ordered `(id, distance)` pairs of a two-granule query against the `use_skip_indexes = 0` baseline. It uses the mark-aligned filter `id <= 2 OR id >= 6` so both skip-index granules contribute hits and no hit is removed by the residual predicate; the index result is then exactly comparable to brute force, so dropping one granule s hints or mis-pairing distances with rows fails the assertion. The expected ordered pairs are also pinned.

rschu1ze:
Added `hold` label, this needs human review @shankar-iyer

zex-hyd:
Hi @alexey-milovidov  @shankar-iyer, would really appreciate a review!

zex-hyd:
@rschu1ze  @alexey-milovidov @shankar-iyer It looks like this PR now requires manual approval again for running the CI, probably after the latest merge from master / maintainer-side updates.

Could you please take another look when you get a chance? The latest PR workflow on the current head is green, and I believe the main review points have been addressed.

Thanks!

rschu1ze:
@zex-hyd This PR contains rather complex changes and we are busy with other vector search developments at the moment. Expect some delay, sorry.

zex-hyd:
Thanks for the update, @rschu1ze. 
I understand. I’ll keep the branch mergeable and wait for further review.

PR #104765, opened 2026-05-12
Fix aggregate projection not matched when multiple sumIf use different IN sets

Ergus:
## Problem

  An aggregate projection with two or more `sumIf(val, col IN (...))` aggregates was broken in two ways:

  - Wrong results: when both aggregates used the same value column, both were silently mapped to the same pre-aggregated projection column, returning incorrect values.
  - Projection unused: when the aggregates used different value columns, the projection was rejected entirely and the query fell back to a full table scan.

  The root cause is that `ColumnSet::operator[]` (the constant argument of every in(col, ...) node) always returns an empty Field{} regardless of content. 
  
  The projection matcher therefore could not distinguish IN ('a','b') from IN ('c','d') and mapped all query in nodes to the same projection node.

## Fix

  Add a content-based, element-order-independent hash to FutureSetFromTuple, computed from the actual set element data at construction time. The projection matcher uses this hash instead of getField() when comparing IN-clause constants, correctly distinguishing different sets while still treating IN ('a','b') and IN ('b','a') as equivalent.

Fix wrong results or missed projection when an aggregate projection contains multiple `sumIf` aggregates with different `IN (...)` conditions.

/fix https://github.com/ClickHouse/support-escalation/issues/7671

### Version info
- Merged into: `26.6.1.321`


src/Interpreters/PreparedSets.cpp
```diff
@@ -141,6 +142,44 @@ FutureSetFromTuple::FutureSetFromTuple(
...
         hash = CityHash_v1_0_2::CityHash128WithSeed(name.data(), name.size(), hash);
     }
+
+    /// Compute a content-based hash. Used by the aggregate projection matcher
+    /// (actionsDAGUtils.cpp) to distinguish different IN-clause sets.
+    ///
+    /// We hash the normalized set elements (deduplicated, NULL-filtered, sorted) rather
+    /// than the raw input block so that permutations and duplicate inputs are equivalent.
+    /// getPermutation/updatePermutation operate by column index, so sets with repeated
+    /// element types (e.g. (String, String)) are handled correctly.
```
KochetovNicolai: I am concerned that this change can lead to performance degradation, especially for the large sets.

We can do it lazily, and only for relatively small tuples. 
I don't really have a good idea of how to improve it further. Was thinking about something like comparing bloom filter states first, which may be cheaper than sorting. But looks like overkill.

src/Interpreters/PreparedSets.cpp
```diff
@@ -141,6 +142,44 @@ FutureSetFromTuple::FutureSetFromTuple(
...
+    {
+        const Columns & normalized = getKeyColumns();
```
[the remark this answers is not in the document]
KochetovNicolai: Apparently, this includes my comment as well.

KochetovNicolai (changes requested):
Let's implement lazy calculation.
And probably some limit on set size.

Ergus:
> Let's implement lazy calculation. And probably some limit on set size.

Hi @KochetovNicolai 

What could be a reasonable default limit for a set size? I won't expect to see a performance degradation unless the set is absurdly large like in the order of 10^4 elements or more, but I could be obviously wrong.

Also, what do you think is expected to happen when the limit is violated? trigger an error? fall back?

alexey-milovidov:

KochetovNicolai:
> What could be a reasonable default limit for a set size? I won't expect to see a performance degradation unless the set is absurdly large like in the order of 10^4 elements or more, but I could be obviously wrong.

Agree, maybe 10^3 or 10^4 is ok.

> Also, what do you think is expected to happen when the limit is violated? trigger an error? fall back?

I think in our case, we can say that every "big" set is not equal to any other set (we won't apply projection and so on)

KochetovNicolai:
I agree with AI comment and something like `use_index_for_in_with_subqueries_max_values` is still needed.

Ergus:
> I agree with AI comment and something like `use_index_for_in_with_subqueries_max_values` is still needed.

The change alread uses `use_index_for_in_with_subqueries_max_values`

The AI comment was because `computeContentHash` was calling `getKeyColumns`, which calls `fillSetElementsOnce` and permanently materializes `set->set_elements`, bypassing the large-set guard in `buildOrderedSetInplace`.

I fixed that in [0521377](https://github.com/ClickHouse/ClickHouse/pull/104765/commits/0521377ea7de0549042c791585ee90df52ae536a)

KochetovNicolai:
I've still added a separate setting. Looks like `use_index_for_in_with_subqueries_max_values` uses a bit differen semantic. Also, it's 0 by default (so no limit on set size), and I think is make sense to limit set compariosn by default.

Feel free to revert if you don't like my change.

KochetovNicolai (approved):

KochetovNicolai:
Note:

```
Can we make FutureSetFromTuple::computeContentHash() implementation without sorting? Is there     
  position-independent hashing?                                                                     

● Yes — a set is unordered by definition, so you don't need to sort to hash it canonically. The
  standard trick is a commutative combination of per-row hashes:

  - Hash each row to a fixed-width value (e.g. SipHash128/CityHash128 over the serialized columns of
   that row).
  - Combine row hashes with a commutative op: addition mod 2^128 (i.e. __uint128_t accumulator).
  This is the textbook "AdHash" multiset hash — preserves multiplicity, is O(N), trivially
  parallelizable (partial sums per thread), and provably as collision-resistant as the underlying
  per-row hash. XOR works too but only as a set (not multiset) hash, because pairs of duplicates
  cancel — usually undesirable.

  Main tradeoff: the combiner is cheap, but the per-row hash dominates cost, so the win over
  sort+hash is only ~`log N. The real wins are (a) no extra memory for sorted copies, (b) 
  streaming/incremental update as rows are inserted, and (c) easy parallelization. If you also want 
  it position-independent across *column order*, you'd hash (column_name, value)` pairs and sum
  inside each row first.

```

But for this, we need to add something like `IColumn::getPositionIndependentHash`, so probably later.

src/Processors/QueryPlan/Optimizations/actionsDAGUtils.cpp
```diff
@@ -188,15 +194,36 @@ MatchedTrees::Matches matchTrees(const ActionsDAG::NodeRawConstPtrs & inner_dag,
...
+                                                    const auto * outer_tuple = typeid_cast<const FutureSetFromTuple *>(
+                                                        outer_set->getData().get());
+                                                    if (inner_tuple && outer_tuple)
+                                                    {
+                                                        const size_t inner_rows = inner_tuple->get()->getTotalRowCount();
+                                                        const size_t outer_rows = outer_tuple->get()->getTotalRowCount();
+                                                        /// Sizes are deduplicated counts; different sizes ⇒ different
+                                                        /// contents, so skip hashing in that case.
+                                                        if (inner_rows == outer_rows
+                                                            && inner_rows <= max_size_for_sets_from_tuple_to_compare)
```
Ergus, line 220: @KochetovNicolai 

The only problem with this part of the fix is that tuples with repeated elements could have different sizes, but still be equivalent once deduplicated the repeated elements.

That's why the initial code went directly to the hash and ignored the total size.

KochetovNicolai: That's a good catch. Will add a test and will remove this check if `getTotalRowCount` returns count without deduplication (I am not confident it does)

KochetovNicolai: I've added a test - looks like duplicates are handled correctly.

KochetovNicolai:
**[PR / Stateless tests (arm_asan_ubsan, targeted) (pull_request)](https://github.com/ClickHouse/ClickHouse/actions/runs/26759059626/job/78874485988?pr=104765) Failing after 55m**

```
● Summary

  This is a CI infrastructure flake, not a real failure caused by the PR.

  What happened:
  - The job is Stateless tests (arm_asan_ubsan, targeted), running ~60 "auto-targeted" optimization/projection tests with --test-runs 50 --jobs 19 --flaky-check.
  - All 2837 attempted test runs passed (job.log:4609 says Failed: 0, Passed: 2837).
  - The wrapper killed the runner with SIGTERM after the ~47 min wall-clock timeout (Timeout exceeded [2817], job.log:4453).
  - The runner exited with -15, which is in ABORTED_RUN_EXIT_CODES, so functional_tests_results.py:238 synthesizes a "Server died" leaf for the report — that's misleading framing for a timeout kill.
  - The actual server process shut down cleanly afterwards: Background threads finished, grpc_shutdown done, Child process exited normally with code 0 (clickhouse-server.log). No <Fatal>, no sanitizer report, no stack trace in
  clickhouse-server.err.log.

  Why it timed out:
  The targeted selector grabbed every test whose name matches optimization/projection/plan (≈60 tests, several slow: 01655_plan_optimizations ~52s, 02317_distinct_in_order_optimization_explain ~60s, 03921_kafka_formats ~38s,
  03031_read_in_order_optimization_with_virtual_row*, ...). Multiplying by --test-runs 50 on ARM ASan+UBSan blew through the 47 min budget. Nothing in the diff makes those tests slower — the targeted selection itself is just too large for
   this config.

  Recommendation: rerun the job. Unrelated to your projection/sumIf change; safe to ignore for the PR's correctness.


```

clickgapai:

### TL;DR
**Fix aggregate projection not matched when multiple sumIf use different IN sets**

### Reproduction
```
CREATE TABLE t_sumif_proj (grp UInt8, col String, val UInt64, PROJECTION prj (SELECT grp, sumIf(val, col IN ('a','b')) AS m1, sumIf(val, col IN ('c','d')) AS m2 GROUP BY grp)) ENGINE = MergeTree ORDER BY grp; INSERT INTO t_sumif_proj SELECT number%4, ['a','b','c','d'][number%4+1], number FROM numbers(40); OPTIMIZE TABLE t_sumif_proj FINAL; SELECT grp, sumIf(val, col IN ('a','b')) AS m1, sumIf(val, col IN ('c','d')) AS m2 FROM t_sumif_proj GROUP BY grp ORDER BY grp SETTINGS optimize_use_projections=1, force_optimize_projection=1; SELECT grp, sumIf(val, col IN ('a','b')) AS m1, sumIf(val, col IN ('c','d')) AS m2 FROM t_sumif_proj GROUP BY grp ORDER BY grp SETTINGS optimize_use_projections=0;
```

### Bisect

### Affects

### Suggested next step

**Component:** `comp-query-optimizer`
**Severity:** `P2`

cc fix PR author: 
cc fix PR reviewer: 

---
_ClickGap · Finding: `phase_d_pr104765`_

PR #104767, opened 2026-05-12
Fix wrong results from WHERE p AND LowCardinality(Nullable(int)) constant on MergeTree

### LINKED ISSUE #104393: `WHERE p AND <LowCardinality(Nullable(UInt8)) constant>` returns zero rows on `MergeTree`
## Affected version
`master` at `be339e3470f` (May 7 2026), reproducible with default settings (`enable_analyzer=1`).

## Reproducer

https://fiddle.clickhouse.com/5a282f56-4c4c-4bd6-b06d-cf58f68b709c

Self-contained — runs verbatim under `clickhouse local` against current `master`:

```sql
SET allow_suspicious_low_cardinality_types = 1;

DROP TABLE IF EXISTS t;
CREATE TABLE t (a UInt32) ENGINE = MergeTree ORDER BY a;
INSERT INTO t SELECT number FROM numbers(500);

-- Baseline: 499 rows match a > 0.
SELECT count() FROM t WHERE (a > 0);                                            -- 499 ✓

-- LowCardinality and Nullable separately are fine:
SELECT count() FROM t WHERE (a > 0) AND CAST(1 AS Nullable(UInt8));             -- 499 ✓
SELECT count() FROM t WHERE (a > 0) AND CAST(1 AS LowCardinality(UInt8));       -- 499 ✓

-- LowCardinality(Nullable(...)) const ANDed with the same predicate returns 0:
SELECT count() FROM t WHERE (a > 0) AND CAST(1 AS LowCardinality(Nullable(UInt8)));   -- 0 ✗
SELECT count() FROM t WHERE (a > 0) AND CAST(1 AS LowCardinality(Nullable(UInt64)));  -- 0 ✗
SELECT count() FROM t WHERE (a > 0) AND toLowCardinality(toNullable(toUInt8(1)));     -- 0 ✗

-- Reverse the AND argument order — still 0:
SELECT count() FROM t WHERE CAST(1 AS LowCardinality(Nullable(UInt8))) AND (a > 0);   -- 0 ✗
```

## Notable properties of the bug

* The bug requires reading from `MergeTree`. Using `numbers()` or a `Memory`-engine table for the same query yields the correct `499`.
* `optimize_move_to_prewhere = 0` does not change anything — the wrong result is produced whether the predicate runs as `WHERE` or as `PREWHERE`.
* The constant value does not matter. Any non-zero, non-NULL value of type `LowCardinality(Nullable(<integer>))` triggers it. Zero / NULL constants behave correctly (no rows match, which is the right answer for `AND 0`).
* Per-row evaluation of the AND in the `SELECT` list is **correct**:
  ```
  SELECT a, greater(a, 0) AND CAST(1 AS LowCardinality(Nullable(UInt8))) FROM t ORDER BY a LIMIT 5
  ```
  produces `(0, 0), (1, 1), (2, 1), (3, 1), (4, 1)`. Only the `WHERE` filter-mask conversion is wrong.
* `WHERE assumeNotNull(<and-expression>)` returns the correct 499. So the workaround is to strip the `Nullable` from the filter expression's result before it reaches the filter step.

## Where the discrepancy comes from

`EXPLAIN actions=1 SELECT count() FROM t WHERE (a > 0) AND CAST(1 AS LowCardinality(Nullable(UInt8)))` shows:

```
Filter column: and(greater(__table1.a, 0_UInt8), _CAST(1_LowCardinality(Nullable(UInt8)), 'LowCardinality(Nullable(UInt8))'_String)) (removed)
…
FUNCTION and(greater(__table1.a, 0_UInt8) :: 3,
             _CAST(1_LowCardinality(Nullable(UInt8)), …) :: 2)
        -> and(greater(__table1.a, 0_UInt8), _CAST(…)) Nullable(UInt8) : 1
```

The AND returns `Nullable(UInt8)`. The MergeTree read step then converts that column to a row mask. Apparently the conversion treats every row as filtered-out when the right operand is a `LowCardinality(Nullable(<int>))` constant — the value is `1, 1, …, 1`, the null map is `0, 0, …, 0` (no NULLs), yet the resulting mask is all zero.

Likely related to PR #101269 (INT_MIN truncated to UInt8 in AND filter) — same general area (filter-mask conversion of `Nullable` constants over MergeTree), different intermediate type combination.

## Found by

PR #99980 AST fuzzer SQLancer oracle (Identity WHERE oracle, the `p ≡ p AND 1` invariant): a fuzzer-mutated query of the form
`SELECT * FROM _s1 WHERE (a > 0) AND intDiv((SELECT 9223372036854775806), toLowCardinality(255)) AND … QUALIFY 255 != a`
returned `0 rows` for the bare predicate vs `498 rows` for `WHERE and(p, 1)`. Constant-folding of the `intDiv` produces a `LowCardinality(Nullable(UInt64))` literal, which is what triggers the bug.

### Version info
- Resolved by: #104767
- Merged into: `26.5.1.826`

groeneai:
On `MergeTree` tables, `WHERE p AND <LowCardinality(Nullable(int)) constant>` silently returned zero rows. Found by @qoega via SQLancer in #104393.

`splitFilterNodeForAllowedInputs` (`src/Storages/VirtualColumnUtils.cpp`) splits the predicate at part-pruning time. When one branch of an `AND` is not evaluable on virtual columns it drops that branch and keeps the surviving constant child. If the child's type differs from the original `AND` result, the function synthesizes a `notEquals(x, default)` guard so the result is still boolean.

It computed the default with `removeNullable(result_type)`. For a `LowCardinality(Nullable(X))` constant the outer type is `LowCardinality`, not `Nullable`, so `removeNullable` was a no-op and `getDefault()` then went through the LC dictionary and returned `Null`. The synthesized guard became `notEquals(x, NULL)`, which evaluates to `NULL` and is treated as false in filter context — so every part was pruned before any data was read.

Replaced with `removeLowCardinalityAndNullable`, which strips both wrappers. The default of the underlying scalar type (`0` for integers) is used instead, restoring the intended `(x > 0) AND truthy_const` semantics. The pre-existing `Nullable(Nothing)` special case (for a bare `NULL` literal) still falls through to the original type's default, so `WHERE x AND NULL` continues to prune every row.

Verified the new test reproducer FAILS on `master` and PASSES on the fix branch (50/50 runs with full setting randomization). Related regression tests (`partition_pruning`, `virtual_column`, `04074_where_large_int_boolean_predicate`, `04104_103049_nullable_and_overflow_filter`) still pass.

Closes #104393.

Fix wrong results from `WHERE p AND <LowCardinality(Nullable(int)) constant>` on `MergeTree` tables. Previously such queries returned zero rows because part pruning derived a `NULL` default from the `LowCardinality(Nullable(...))` type and synthesized a `notEquals(x, NULL)` guard, pruning every part.

### Version info
- Merged into: `26.5.1.826`


groeneai:
## Pre-PR validation gate (TASK.md Phase 4 step 9)

- **(a) Deterministic repro?** Yes — the SQL in the test fails consistently on `master` (returns 0/0/0/0 for the four `LowCardinality(Nullable(...))`-flavored queries) and passes 50/50 runs on this branch with full setting randomization.
- **(b) Root cause explained?** Yes — `splitFilterNodeForAllowedInputs` (`src/Storages/VirtualColumnUtils.cpp`) used `removeNullable(result_type)` to pick a default for a synthesized `notEquals(x, default)` guard. For a `LowCardinality(Nullable(X))` constant the outer type is `LowCardinality` (not `Nullable`), so `removeNullable` was a no-op and `getDefault` fell through to the dictionary's `Nullable` default (`Null`). The guard became `notEquals(x, NULL)` -> `NULL` -> false, so partition pruning discarded every part before any data was read.
- **(c) Fix matches root cause?** Yes — switched to `removeLowCardinalityAndNullable` which strips both wrappers, so the default of the actual scalar type (`0` for integers) is used. The `Nullable(Nothing)` special case for a bare `NULL` literal still falls through to the original type's default (`Null`), so `WHERE x AND NULL` continues to prune all rows.
- **(d) Tests added?** Yes — `04234_where_and_low_cardinality_nullable_constant_104393.{sql,reference}` covers all six original reproducer variants plus a falsy `LC(Nullable(...))` constant, a `NULL` `LC(Nullable(...))` constant, the PREWHERE path, and the no-prewhere variant.
- **(e) Both directions demonstrated?** Yes — test FAILS on master HEAD (verified by reverting the fix, rebuilding, and re-running) and PASSES on this branch (verified 50/50 with full randomization).
- **(f) Fix is general, not a narrow patch?** Yes — `removeLowCardinalityAndNullable` handles every `(LC?, Nullable?) * scalar` combination: `LC(UInt8)`, `LC(Nullable(UInt8))`, `LC(Nullable(UInt64))`, plain `Nullable(UInt8)`, plain `UInt8`, and `Nullable(Nothing)` (falls through to the special case). Verified that the surrounding related tests (`partition_pruning`, `virtual_column`, `04074_where_large_int_boolean_predicate`, `04104_103049_nullable_and_overflow_filter`) still pass.

Session: `cron:clickhouse-ci-task-worker:20260512-214500`.

groeneai:
cc @CurtizJ @KochetovNicolai — could you review this? It fixes a silent wrong-results bug on `MergeTree` where `WHERE p AND <LowCardinality(Nullable(int)) constant>` returned 0 rows because part pruning's synthesized `notEquals(x, default)` used `removeNullable` (which is a no-op on `LowCardinality(Nullable(...))`) and got `NULL` as the default. Closes #104393.

Avogar (approved):

alexey-milovidov:
@groeneai, provide fixes for these two errors in separate PRs:

[Stateless tests (amd_msan, WasmEdge, parallel, 2/2)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104767&sha=9835180731fd3ac85846115caa0a5c8bd4e5ad20&name_0=PR&name_1=Stateless%20tests%20%28amd_msan%2C%20WasmEdge%2C%20parallel%2C%202%2F2%29)		FAIL		
03156_tuple_map_low_cardinality	FAIL	[cidb](https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICcwMzE1Nl90dXBsZV9tYXBfbG93X2NhcmRpbmFsaXR5JwogICAgLS0gQU5EIGNoZWNrX25hbWUgPSAnU3RhdGVsZXNzIHRlc3RzIChhbWRfbXNhbiwgV2FzbUVkZ2UsIHBhcmFsbGVsLCAyLzIpJwogICAgQU5EIHRlc3Rfc3RhdHVzIElOICgnRkFJTCcsICdFUlJPUicpCiAgICBBTkQgKHB1bGxfcmVxdWVzdF9udW1iZXIgPSAwIE9SIGJhc2VfcmVmIElOICgnbWFzdGVyJykpCiAgICBBTkQgdGVzdF9jb250ZXh0X3JhdyBMSUtFICclVGltZW91dCUnCkdST1VQIEJZIGRheQpPUkRFUiBCWSBkYXkgREVTQwo=)	
[Stateless tests (arm_asan_ubsan, azure, parallel)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104767&sha=9835180731fd3ac85846115caa0a5c8bd4e5ad20&name_0=PR&name_1=Stateless%20tests%20%28arm_asan_ubsan%2C%20azure%2C%20parallel%29)		FAIL		
03040_dynamic_type_alters_2_compact_merge_tree

As soon as the fixes are submitted or linked, update the branch here.

Avogar:
Let's move these failures in separate issues, merge this one and I will take care of the issues.

clickgapai:
Hi — could this be a backport candidate for `26.4`? No backport label was found, and the affected code appears to exist in those branches based on a file/function presence check.

**Affected code:** `splitFilterNodeForAllowedInputs` in src/Storages/VirtualColumnUtils.cpp:626 — introduced in 26.4.

**Why this might apply:** The bug was introduced by the #103049 fix (commit 7c78cbd301e, 2026-04-18) which was backported to 26.4. Branch 26.4 shows `removeNullable(res->result_type)` at the affected code path, not the fixed `removeLowCardinalityAndNullable`. Branches 26.3, 26.2, and 25.8 don't have the #103049 fix at all, so they don't have this bug.

Other supported branches (`26.3`, `26.2`, `25.8`) predate 26.4 and do not contain this code.

cnkk:
I have a similar issue on v26.3.x which seems to be fixed in 26.5:

https://fiddle.clickhouse.com/41e96b7a-89a5-4831-9927-7fcf3cdb30f0

```sql
DROP TABLE IF EXISTS t_repro;
CREATE TABLE t_repro (
    s String,
    n LowCardinality(Nullable(String)),
    INDEX idx_n n TYPE set(100) GRANULARITY 2,
    INDEX idx_s s TYPE ngrambf_v1(4, 1024, 2, 0) GRANULARITY 3
) ENGINE = MergeTree ORDER BY tuple();

INSERT INTO t_repro SELECT 'has-republisher', NULL FROM numbers(100000);
INSERT INTO t_repro SELECT 'no-match', NULL FROM numbers(100);

-- Correct: 100,000 rows match the tag
SELECT count() FROM t_repro WHERE s LIKE '%republisher%';

-- BUG: OR with the nullable column should ALSO return 100,000, but returns 0 on 26.3.x
SELECT count() FROM t_repro WHERE (n = 'republisher' OR s LIKE '%republisher%');

-- WORKAROUND: assumeNotNull() returns the correct 100,000 on all versions
SELECT count() FROM t_repro WHERE (assumeNotNull(n) = 'republisher' OR s LIKE '%republisher%');
```

@Avogar any ideas if a fix could be related to this PR?

den-crane:
@cnkk  It's seems like an issue in combining two skip indexes. Open an issue. It could be accidentally fixed by this PR. Anyway your example should be added to test/ci.

PR #104785, opened 2026-05-13
Fix duplicate `arrayJoin` execution in JOIN ON with new logical join step

tiandiwonder:

Fix wrong row count from `arrayJoin()` used inside `JOIN ON`: the result was multiplied by the array length an extra time.

---

### Problem

```sql
SELECT count() FROM items
INNER JOIN tags ON arrayJoin(items.tag_ids) = tags.id
WHERE items.filter_key = 'target';
```

With `query_plan_use_new_logical_join_step = 1` (the default), this query returns `|tag_ids|² × N_items` rows instead of the expected `|tag_ids| × N_items`. On an array of size 300 the result is `300×` larger than correct. Workaround: `SET query_plan_use_new_logical_join_step = 0`.

### Root cause

When a WHERE filter is pushed below a `JoinStepLogical` whose ON clause contains `arrayJoin(...)`, `fix_predicate_for_join_logical_step` in `filterPushDown.cpp` merges `pre_filter_dag` (which carries the ON-side `ARRAY_JOIN` node) into the pushed-down filter. `removeUnusedActions` preserves that `ARRAY_JOIN` unconditionally because of the row-changing carve-out in `ActionsDAG.cpp` (`removeUnusedActions` treats every `ARRAY_JOIN` node as a root since removing it would silently change row count). The same `ARRAY_JOIN` is then independently cloned into the per-side "Pre Join Actions" `ExpressionStep` by `JoinStepLogical::buildPhysicalJoinImpl`, so each input row is expanded by `|array|` twice.

`EXPLAIN actions = 1` on the buggy plan shows two `ARRAY JOIN __table1.tag_ids` actions on the items side (one in `Right Pre Join Actions`, one in the pushed-down filter). The fix collapses them to one.

### Fix

Snapshot the `ARRAY_JOIN` node pointers in `pre_filter_dag` before the merge. `ActionsDAG::merge` is implemented as `mergeInplace`, which uses `std::list::splice` to move `second.nodes` into `first.nodes`, so addresses stay valid in `filter_dag` after the merge and the snapshot lets us identify the merged-in nodes by pointer rather than by `result_name` (correctly handling nested forms like `JOIN ON toString(arrayJoin(a)) = b`).

After the merge + `removeUnusedActions`, drop those nodes via a new narrow API `ActionsDAG::removeNodes(std::unordered_set<const Node *>)`. The new helper computes its own transitive-required closure from `outputs` (without the row-changing carve-out) so any node still required by the filter predicate is kept regardless of `to_remove`; ghost `ARRAY_JOIN`s that fell out of the filter's transitive closure are removed.

Safety:
- `splitActionsForJOINFilterPushDown` rejects `ARRAY_JOIN` from `allowed_nodes`, so the `filter_dag` arriving at `fix_predicate_for_join_logical_step` is guaranteed not to contain any `ARRAY_JOIN` of its own. Every `ARRAY_JOIN` we see after the merge must have come from `pre_filter_dag`.
- `JoinStepLogical::buildPhysicalJoinImpl` independently materialises the same JOIN-side expression DAG into the per-side "Pre Join Actions" `ExpressionStep` above the filter, so dropping the ghost copy in the pushed-down filter does not lose the expression -- the row expansion still happens, once, above the JOIN.

### Related fixes for the same root cause

This is the third bug in 12 months rooted in `removeUnusedActions`' row-changing carve-out interacting with `mergeInplace` / `clone` of `ARRAY_JOIN`-bearing DAGs. The prior two are #96989 (prevention strategy: blacklist `ARRAY_JOIN` in the pushdown eligibility check) and #97239 (bail-out strategy: detect ghost `ARRAY_JOIN` after the merge and abandon the optimization). This PR uses a third strategy -- post-hoc cleanup -- because `fix_predicate_for_join_logical_step` is a mandatory plan transformation (the merge is required to propagate `join_use_nulls` type conversions), so neither prevention nor bail-out is applicable.

### Regression test

`tests/queries/0_stateless/03918_arrayjoin_join_on_duplicate_execution.{sql,reference}` -- co-located with #96989's regression test `03918_arrayjoin_function_with_join_and_where` so the two related fixes share a number prefix. 12 queries covering:

- **Result-level fix evidence**: baseline, reversed equality direction, nested `toString(arrayJoin(...))`, runtime-filter off, disjunction-push-down off.
- **Mechanism-level evidence**: `EXPLAIN actions = 1` substring count of `'ARRAY JOIN '` -- buggy plan has one extra ghost action.
- **Bilateral arrayJoin**: `JOIN ON arrayJoin(L) = arrayJoin(R)` with WHERE on left only, right only, and both sides simultaneously, so both left-side and right-side `fix_predicate_for_join_logical_step` invocations are exercised symmetrically.
- **Plan-shape invariant check**: a query with a WHERE-side `AND arrayJoin(...)` conjunct -- validates that the upstream `splitActionsForJOINFilterPushDown` rejection keeps the WHERE-side `arrayJoin` above the JOIN while the ON-side ghost is removed.
- **Controls**: `query_plan_use_new_logical_join_step = 0` baseline and subquery rewrite (must keep returning the correct count to detect regression of the unrelated paths).

Repro-first verified: every query that claims to be fixed by this PR was first run on un-fixed master and observed to fail with the exact bug-shape values before the reference was locked.

### Version info
- Backported to: `26.5.4.15`, `26.4.5.69`, `26.3.16.9`, `25.8.29.3`


src/Interpreters/ActionsDAG.h
```diff
@@ -241,6 +241,12 @@ class ActionsDAG
     /// The order of outputs might be changed even if actions are not removed.
     bool removeUnusedActions(const NameSet & required_names, bool allow_remove_inputs = true, bool allow_constant_folding = true);
 
+    /// Remove the given nodes, bypassing `removeUnusedActions`'s row-changing
+    /// carve-out for ARRAY_JOIN. Nodes still required by any output are kept.
+    /// Caller MUST ensure removed ARRAY_JOIN is recomputed elsewhere in the plan.
+    /// Returns count of removed nodes.
+    size_t removeNodes(const std::unordered_set<const Node *> & to_remove);
```
antaljanosbenjamin, line 248: I think the logic is correct. The only issue is I don't like the som many different remove functions got another friend. However it is well documented and I don't have a better idea, so I am okay keeping it.

antaljanosbenjamin (approved):

clickgapai:
Hi @tiandiwonder @antaljanosbenjamin — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: The PR description is empty - it only contains the standard template without any explanation of what changed or why.
- Suggested addition:

  > ### What this fixes
  > When `arrayJoin()` is used inside `JOIN ON`, the array join operation was being executed twice - once in the pushed-down filter and once in `JoinStepLogical`'s Pre Join Actions. This caused the row count to be multiplied by the array length an extra time.
  > 
  > ### How it works
  > Adds a new `ActionsDAG::removeNodes` method that explicitly removes specified nodes, bypassing `removeUnusedActions`'s carve-out that preserves ARRAY_JOIN nodes (since they change row count). In `fix_predicate_for_join_logical_step`, we snapshot ARRAY_JOIN nodes from `pre_filter_dag` before merging, then remove them after the merge since they're already computed by the join step's Pre Join Actions.

PR #102690, opened 2026-04-14
Fix inserting into Time data type during JSON parsing

Avogar:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix inserting into Time data type during JSON parsing. Closes https://github.com/ClickHouse/ClickHouse/issues/102016

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.683`

PR #104790, opened 2026-05-13
Fix `LOGICAL_ERROR` "Table has no columns." when `Merge` matches a dangling `Alias`

### LINKED ISSUE #104740: Logical error: 'Table has no columns.' (STID: 3593-5990)

Test name: Logical error: 'Table has no columns.' (STID: 3593-5990)
CI report: [Stress test (arm_release)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102690&sha=9a69b85f69664cda58dadcef5f682e82dfa15e8d&name_0=PR&name_1=Stress%20test%20%28arm_release%29)
Failing test history: [cidb](https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICdMb2dpY2FsIGVycm9yOiAnJ1RhYmxlIGhhcyBubyBjb2x1bW5zLicnIChTVElEOiAzNTkzLTU5OTApJwogICAgLS0gQU5EIGNoZWNrX25hbWUgPSAnU3RyZXNzIHRlc3QgKGFybV9yZWxlYXNlKScKICAgIEFORCB0ZXN0X3N0YXR1cyBJTiAoJ0ZBSUwnLCAnRVJST1InKQogICAgQU5EIChwdWxsX3JlcXVlc3RfbnVtYmVyID0gMCBPUiBiYXNlX3JlZiBJTiAoJ21hc3RlcicpKQogICAgQU5EIHRlc3RfY29udGV4dF9yYXcgTElLRSAnJUV4Y2VwdGlvbjolJwpHUk9VUCBCWSBkYXkKT1JERVIgQlkgZGF5IERFU0MK)

Test output:
```
Error:
Logical error: 'Table has no columns.'.
---

Stack trace:
__pthread_kill_implementation @ 0x0000000000082009
raise @ 0x000000000003a83c
__GI_abort @ 0x0000000000027134
DB::abortOnFailedAssertion(String const&, std::basic_string_view<char, std::char_traits<char>>, void* const*, unsigned long, unsigned long) @ 0x00000000131c6e88
DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0x00000000131c7b28
src/Common/Exception.h:171:100: DB::Exception::Exception(String&&, int, String, bool) @ 0x000000000dbf5e98
src/Common/Exception.h:57:54: DB::Exception::Exception(PreformattedMessage&&, int) @ 0x000000000dbf570c
DB::Exception::Exception<>(int, FormatStringHelperImpl<>) @ 0x00000000131c6054
DB::ReadFromMerge::createChildrenPlans(DB::SelectQueryInfo&) const @ 0x0000000017e54cb0
DB::ReadFromMerge::filterTablesAndCreateChildrenPlans() @ 0x0000000017e4cd6c
DB::ReadFromMerge::applyFilters(DB::ActionDAGNodes) @ 0x0000000017e5e120
DB::QueryPlanOptimizations::optimizePrimaryKeyConditionAndLimit(std::vector<DB::QueryPlanOptimizations::Frame, std::allocator<DB::QueryPlanOptimizations::Frame>> const&) @ 0x000000001925f040
DB::QueryPlanOptimizations::optimizeTreeSecondPass(DB::QueryPlanOptimizationSettings const&, DB::QueryPlan::Node&, std::list<DB::QueryPlan::Node, std::allocator<DB::QueryPlan::Node>>&, DB::QueryPlan&) @ 0x000000001925afd4
DB::QueryPlan::optimize(DB::QueryPlanOptimizationSettings const&) @ 0x0000000019141798
DB::QueryPlan::buildQueryPipeline(DB::QueryPlanOptimizationSettings const&, DB::BuildQueryPipelineSettings const&, bool) @ 0x00000000191406a4
DB::InterpreterSelectQueryAnalyzer::buildQueryPipeline() @ 0x0000000017309d00
DB::InterpreterSelectQueryAnalyzer::execute() @ 0x0000000017309850
DB::executeQueryImpl(char const*, char const*, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum, std::unique_ptr<DB::ReadBuffer, std::default_delete<DB::ReadBuffer>>&, boost::intrusive_ptr<DB::IAST>&, std::shared_ptr<DB::ImplicitTransactionControlExecutor>, std::function<void ()>, DB::QueryResultDetails&) @ 0x00000000176a0bbc
DB::executeQuery(String const&, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum) @ 0x000000001769b9ac
DB::TCPHandler::runImpl() @ 0x0000000018b2df1c
DB::TCPHandler::run() @ 0x0000000018b4cec8
Poco::Net::TCPServerConnection::start() @ 0x000000001d936f58
Poco::Net::TCPServerDispatcher::run() @ 0x000000001d937524
Poco::PooledThread::run() @ 0x000000001d8f7974
Poco::ThreadImpl::runnableEntry(void*) @ 0x000000001d8f5c30
start_thread @ 0x0000000000080398
thread_start @ 0x00000000000e9e9c
```

### Version info
- Resolved by: #104790
- Merged into: `26.5.1.789`

groeneai:
A `Merge` table whose regexp matches an `Alias` storage whose target table
has been dropped aborts the query with an internal `LOGICAL_ERROR`
"Table has no columns.". This shows up in CI as
[STID 3593-560a / 3593-5990](https://github.com/ClickHouse/ClickHouse/issues/104740)
and was flagged on https://github.com/ClickHouse/ClickHouse/pull/100752
("Enable snappy compression in HTTP interface") and
https://github.com/ClickHouse/ClickHouse/pull/102690
("Fix inserting into Time data type during JSON parsing") — both
unrelated to `StorageMerge`/`StorageAlias`.

### Root cause

- `StorageAlias::getInMemoryMetadataPtr` in `src/Storages/StorageAlias.h`
  returns the alias's own (empty) metadata when its target is missing.
  This is intentional so that metadata-only operations such as
  `system.tables`/`system.columns` queries on a dangling alias work
  (exercised by the existing `03636_storage_alias_basic` test).
- `ReadFromMerge::createChildrenPlans` in `src/Storages/StorageMerge.cpp`
  iterates the selected children, calls `getInMemoryMetadataPtr` on each,
  and throws `LOGICAL_ERROR` "Table has no columns." when the snapshot
  has no columns — assuming the only legitimate empty case is a
  parameterized view.

In CI the path is reached two ways:
- **Stress tests:** `ignore_drop_queries_probability` leaks the
  `alias_with_missing_target` table from `03636_storage_alias_basic`
  into a later test whose `Merge(currentDatabase(), 'a')` regexp
  matches it.
- **Regular workloads:** TOCTOU race when the target is dropped between
  `StorageMerge::getSelectedTables` (which only locks the alias storage,
  not its target) and the metadata snapshot read in
  `createChildrenPlans`.

### Fix

Detect the `Alias` case in the empty-metadata branch and throw
`UNKNOWN_TABLE` with an exception that names both the dangling alias
and the matching `Merge` table. This mirrors what
`StorageMerge::isRemote` already surfaces during analyzer resolution
via `StorageAlias::isRemote` → `getTargetTable`.

A new regression test `04236_merge_with_alias_missing_target_104740`
locks down the user-visible contract: a `SELECT` through a `Merge`
whose regexp matches a dangling alias must throw `UNKNOWN_TABLE`,
never `LOGICAL_ERROR`.

Closes #104740.

CI evidence: 30-day CIDB shows 6 hits across `arm_msan`/`arm_release`/`amd_tsan`
stress runs on master and 4 distinct unrelated PRs.

### Version info
- Merged into: `26.5.1.789`


groeneai:
### Pre-PR validation gate (per TASK.md)

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | **Partial.** Local single-threaded repro deterministically reaches `StorageMerge::isRemote` via the analyzer's `resolveQueryJoinTreeNode`, which already throws `UNKNOWN_TABLE` via `StorageAlias::isRemote → getTargetTable`. The `LOGICAL_ERROR` branch in `createChildrenPlans` is only reached in CI when a stress test or TOCTOU race lets the alias's target survive analyzer resolution but disappear before plan optimization. The new test (`04236_…`) locks down the user-visible contract that **both** paths throw `UNKNOWN_TABLE`. |
| b | **Root cause explained?** | Yes. `StorageAlias::getInMemoryMetadataPtr` falls back to `IStorage`'s default (empty `metadata`) when `tryGetTargetTable` returns null — this is needed for `system.tables`/`system.columns` queries on a dangling alias (exercised in `03636_storage_alias_basic`). `ReadFromMerge::createChildrenPlans` then sees the empty `getColumns()` and throws `LOGICAL_ERROR` "Table has no columns." in its parameterized-view-or-bug branch. |
| c | **Fix matches root cause?** | Yes. The fix adds the third legitimate case ("dangling `Alias`") to the empty-columns branch, with a clear `UNKNOWN_TABLE` exception that names both the alias and the `Merge` table that matched it. The existing `LOGICAL_ERROR` is preserved as a genuine assertion for any other empty-metadata producer. |
| d | **Test intent preserved? / new test added?** | Yes. New test `04236_merge_with_alias_missing_target_104740` added — asserts `serverError UNKNOWN_TABLE` for both a bare `SELECT` and the JOIN pattern from `03549_analyzer_fix_filter_removal`. Existing tests preserved: `03636_storage_alias_basic` (metadata-only ops on dangling alias), `03549_analyzer_fix_filter_removal`, `01925_test_storage_merge_aliases`, `01214_test_storage_merge_aliases_with_where` all still pass. |
| e | **Both directions demonstrated?** | Yes for the contract. The new test passes 50/50 with full randomization (`tests/clickhouse-test --test-runs 50 04236_…`). The race-condition path that produced the original `LOGICAL_ERROR` cannot be deterministically triggered single-threaded; CI's stress tests will validate that the `LOGICAL_ERROR` disappears. |
| f | **Fix is general / no symptom-only patch?** | Yes. `Grep` for `"Table has no columns"` in `src/` returns one site (the one fixed). Other `getInMemoryMetadataPtr` call-sites in `StorageMerge` (`traverseTablesUntilImpl` line 232, the `order_optimizer` loop at line 646) tolerate empty metadata gracefully — they don't crash with a dangling alias. The fix targets the only assertion site that fired in CI. |

Directive: @alexey-milovidov's comment on https://github.com/ClickHouse/ClickHouse/pull/100752 (2026-05-13T03:12:00Z).

Session: `cron:clickhouse-ci-task-worker:20260513-051500`.

groeneai:
cc @alexey-milovidov @azat — could you review this? Fixes the chronic STID 3593-560a / 3593-5990 `LOGICAL_ERROR` "Table has no columns." by turning the assertion into a user-facing `UNKNOWN_TABLE` for the dangling-`Alias` case in `ReadFromMerge::createChildrenPlans`. Closes #104740, per the directive on #100752.

src/Storages/StorageMerge.cpp
```diff
@@ -707,8 +708,22 @@ std::vector<ReadFromMerge::ChildPlan> ReadFromMerge::createChildrenPlans(SelectQ
...
-                    throw Exception(ErrorCodes::LOGICAL_ERROR, "Table has no columns.");
+
+                /// An `Alias` storage whose target table has been dropped returns
+                /// empty in-memory metadata (it has no columns of its own). When a
+                /// `Merge` regexp happens to match such a dangling alias, surface a
+                /// clear `UNKNOWN_TABLE` error rather than the internal
+                /// `LOGICAL_ERROR` "Table has no columns.". This can be reached via
+                /// a TOCTOU race when the target table is dropped between the
+                /// `getSelectedTables` lock and the metadata snapshot read here.
+                if (storage->getName() == "Alias")
```
azat, line 713: Hm, but why `StorageAlias::getInMemoryMetadataPtr` does not throw in this case?

groeneai: Intentional, to keep introspection of dangling aliases working.

`03636_storage_alias_basic.sql` (the test that ships with the engine) creates `alias_with_missing_target`, drops its target, and then expects these to return rows:

```sql
SELECT name, engine FROM system.tables WHERE name = 'alias_with_missing_target';
-- expected: alias_with_missing_target  Alias
SELECT name FROM system.columns WHERE table = 'alias_with_missing_target';
-- expected: (empty set — i.e. 0 column rows, not an error)
SELECT name, engine, total_rows, total_bytes, data_paths FROM system.tables WHERE name = 'alias_with_missing_target';
-- expected: alias_with_missing_target  Alias  \N  \N  []
```

Both `StorageSystemColumns` (`src/Storages/System/StorageSystemColumns.cpp:158`) and `StorageSystemTables` (`src/Storages/System/StorageSystemTables.cpp:600`) call `storage->getInMemoryMetadataPtr(context, false)` directly on every visible table; if it threw for a dangling `Alias`, those introspection queries would fail with `UNKNOWN_TABLE` instead of just showing the alias as a row with zero columns. That's the contract the test locks down — and the `try`-variants in `StorageAlias.h` (`tryGetDataPaths`, `tryGetSerializationHints`, `tryGetColumnSizes`, `tryLifetimeRows`, `tryLifetimeBytes`, `tryGetStoragePolicy`, `tryLockForShare`) are precisely there because metadata access on a dangling alias must not be fatal.

So the fix has to live at the boundary where the contract changes from *"metadata is allowed to be empty"* (introspection) to *"metadata must be non-empty"* (read pipeline). `ReadFromMerge::createChildrenPlans` already encodes the same pattern for parameterized views a few lines above — it converts the empty-metadata signal into `STORAGE_REQUIRES_PARAMETER`. The new branch does the equivalent for the `Alias` case, converting it into `UNKNOWN_TABLE`.

Pivoting to throw at the `StorageAlias` source would require either reverting the "Test alias with missing target table" section of `03636_storage_alias_basic.sql` (and removing the dangling-alias contract entirely), or wrapping every introspection call site in a try/catch, both of which look strictly worse than catching the one user-reachable read path here.

azat: I like this way more - https://github.com/ClickHouse/ClickHouse/pull/104919

groeneai:
Standing down — @azat's PR #104919 supersedes this one.

Quick summary for future readers:

| Aspect | This PR #104790 | @azat's PR #104919 |
|---|---|---|
| Approach | Catch empty-metadata in `ReadFromMerge::createChildrenPlans` and convert to `UNKNOWN_TABLE` | Make `StorageAlias::getInMemoryMetadataPtr` throw on missing target; introduce `tryGetInMemoryMetadataPtrForSystemTable()` for the introspection paths |
| Files changed | 2 | 10 |
| Regression test reused | `04236_merge_with_alias_missing_target_104740` | Same |

@azat's approach fixes the contract at the source (alias storage refuses to hand out empty metadata) and isolates the "tolerate dangling target" behavior to the four system-table call sites that actually need it. That's cleaner than guarding at one consumer (our approach).

Closing this PR in favor of #104919. Thanks for the alternative @azat!

Session: cron:clickhouse-ci-task-worker:20260514-091500

src/Storages/StorageMerge.cpp
groeneai: Got it — thanks for the alternative. Your approach (refuse to hand out empty metadata at the source + isolate the tolerate-dangling behavior to `tryGetInMemoryMetadataPtrForSystemTable()`) is cleaner than guarding at the `ReadFromMerge` consumer. Closing #104790 in favor of #104919.

groeneai:
Superseded by #104919 (see above).

azat (approved):

src/Storages/StorageMerge.cpp
```diff
@@ -707,8 +708,22 @@ std::vector<ReadFromMerge::ChildPlan> ReadFromMerge::createChildrenPlans(SelectQ
...
-                else
-                    throw Exception(ErrorCodes::LOGICAL_ERROR, "Table has no columns.");
+
+                /// An `Alias` storage whose target table has been dropped returns
+                /// empty in-memory metadata (it has no columns of its own). When a
+                /// `Merge` regexp happens to match such a dangling alias, surface a
+                /// clear `UNKNOWN_TABLE` error rather than the internal
+                /// `LOGICAL_ERROR` "Table has no columns.". This can be reached via
+                /// a TOCTOU race when the target table is dropped between the
+                /// `getSelectedTables` lock and the metadata snapshot read here.
```
azat:
```suggestion
                /// An `Alias` storage whose target table has been dropped returns empty in-memory metadata
```

groeneai:
Thanks for the review and approval, @azat — and thanks for applying the comment-wording suggestion directly (commit 5b5c8950). The new comment is much cleaner.

For context to anyone following this thread: I had earlier stood down in favor of #104919, but it looks like this PR's approach was kept after all. CI is currently rebuilding on the updated commit; will monitor for green and ping again if anything else surfaces.

clickgapai:

**Affected code:** `ReadFromMerge::createChildrenPlans` empty metadata check for Alias storage — introduced in 25.11.

**Why this might apply:** Fix for LOGICAL_ERROR that can occur in stress tests and TOCTOU race conditions. StorageAlias was introduced in 25.11, so the fix should be backported to all branches where the feature exists (26.2, 26.3, 26.4). Not needed for 25.8 since StorageAlias doesn't exist there.

Other supported branches (`25.8`) predate 25.11 and do not contain this code.

PR #104792, opened 2026-05-13
Add custom settings to 'system.disks' table

EilRoviSoft:
- New Feature

### Changelog entry:
...
Now custom settings for ClickHouse can be added if they are in whitelist.
This whitelist probably will be in config itself as list of options.

### DIFF
--- a/src/Disks/DiskObjectStorage/RegisterDiskObjectStorage.cpp
+++ b/src/Disks/DiskObjectStorage/RegisterDiskObjectStorage.cpp
@@ -15,37 +15,6 @@ namespace DB
 void registerObjectStorages();
 void registerMetadataStorages();
 
-ConfigurationFields getS3DiskConfigurationFields(
-    const Poco::Util::AbstractConfiguration & config,
-    const String & config_prefix)
-{
-    static const std::unordered_set<String> whitelist =
-    {
-        "s3_max_get_rps",
-        "s3_max_get_burst",
-        "s3_max_put_rps",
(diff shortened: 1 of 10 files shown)


src/Storages/System/StorageSystemDisks.cpp
```diff
@@ -77,6 +88,22 @@ Pipe StorageSystemDisks::read(
...
+        auto offsets = ColumnArray::ColumnOffsets::create();
+
+        Columns tuple;
+        tuple.emplace_back(std::move(keys));
+        tuple.emplace_back(std::move(values));
+        
+        MutableColumnPtr nested = ColumnArray::create(ColumnTuple::create(std::move(tuple)), std::move(offsets))
+            ->assumeMutable();
+
+        col_configuration_fields = ColumnMap::create(std::move(nested));
```
kirillgarbar, line 105: Why not use `ColumnMap::create(keys, values, offsets)` instead?

src/Disks/DiskObjectStorage/RegisterDiskObjectStorage.cpp
```diff
@@ -15,6 +15,37 @@ namespace DB
 void registerObjectStorages();
 void registerMetadataStorages();
 
+ConfigurationFields getS3DiskConfigurationFields(
+    const Poco::Util::AbstractConfiguration & config,
+    const String & config_prefix)
+{
```
kirillgarbar: Current implementation doesn't support config reload

PR #104801, opened 2026-05-13
Capture jstack/UC liveness diagnostics on Spark hang in test_database_delta

groeneai:
The chronic flaky test `test_database_delta::test_check_database_unity` times out on `Integration tests (arm_binary, distributed plan, 4/4)` when the Spark JVM hangs between Hive metastore initialization and actual query execution. PR #102649 (merged 2026-04-14) batched 7 sequential Spark invocations into one, dropping the failure rate to 2 master hits in 30 days, but a residual hang remains and we currently have no data to localize it.

When `subprocess.TimeoutExpired` fires inside `execute_spark_query`, the Spark JVM is killed by SIGKILL so its `stdout`/`stderr` are truncated to whatever it logged before the hang point, and the existing `uc.log` tail is not enough to tell us where the JVM is stuck. CIDB rows for the failing master shards (`2a00d99d`, `2f0c1004` — 2026-05-11) consistently stop at `ObjectStore.setMetaStoreSchemaVersion`, which is the last metastore-init log line. We do not know whether Spark is then blocked on a Unity Catalog HTTP call, on a Derby file lock, or somewhere else entirely.

This PR adds a `_capture_spark_hang_diagnostics` helper that runs after the existing log capture in the exception handler of `execute_spark_query` and collects three pieces of state from inside the test container:

1. `jstack` thread dump of every running `org.apache.spark` JVM — shows exactly where the JVM is stuck during the hang.
2. Unity Catalog HTTP liveness probe on `http://localhost:8080/api/2.1/unity-catalog/schemas` — shows whether UC is responsive while Spark is hung.
3. `ss -tnp` snapshot of sockets matching `:8080` or `java` — shows whether Spark holds open sockets to UC at the moment of the hang.

Each diagnostic is wrapped in its own `try`/`except` so a failure in one does not prevent collection of the others, and all `exec_in_container` calls use short timeouts (15–30 s) so the diagnostics cannot meaningfully extend the test failure time. The helper runs only on the already-failing path so it cannot make a passing test fail. All three tools (`jstack`, `curl`, `ss`) are present in the `clickhouse/integration-test-with-unity-catalog:latest` image used by the test fixture (verified by `docker run --entrypoint /bin/bash ... -c 'which jstack curl ss'`).

This is diagnostic-only — it does not fix the residual hang. The data it captures will let the next master hit be localized to a specific stuck thread or UC RPC instead of "Spark stopped logging".

Sample failure (master): https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=2a00d99de694fda4302148265b75a9d42ae878fa&name_0=MasterCI

...

- [ ] Documentation is unchanged (no user-facing behavior change).

### Version info
- Merged into: `26.5.1.648`

### DIFF
--- a/tests/integration/test_database_delta/test.py
+++ b/tests/integration/test_database_delta/test.py
@@ -108,12 +108,24 @@ def _capture_spark_hang_diagnostics(node):
     """
     # 1. Thread dumps of running Spark JVMs.
     try:
+        # `[o]rg.apache.spark` is a character-class trick: it matches the
+        # literal string `org.apache.spark` (because `[o]` matches `o`) but
+        # the regex itself does NOT contain that string, so `pgrep` will
+        # not match its own command line (`pgrep -f '[o]rg.apache.spark'`).
+        # Without this, on shards where no Spark JVM is left the `bash -c`
+        # wrapper's own cmdline can match and `jstack` then runs against
+        # an unrelated shell PID.
         pids_out = node.exec_in_container(
-            ["bash", "-c", "pgrep -f 'org.apache.spark' || true"],
+            ["bash", "-c", "pgrep -f '[o]rg.apache.spark' || true"],
             nothrow=True,
             timeout=30,
         )
         pids = [p.strip() for p in (pids_out or "").splitlines() if p.strip()]
+        # Cap the number of jstack-ed PIDs at 3 so a hang affecting both
+        # the driver and several executors does not consume 30s per PID
+        # before the test fails. In practice there is one driver JVM per
+        # query; the cap is a belt for the pathological case.
+        pids = pids[:3]
         if not pids:
             print(
                 "Spark hang diag: no running 'org.apache.spark' processes"
--- a/tests/integration/test_database_delta/test.py
+++ b/tests/integration/test_database_delta/test.py
@@ -174,20 +174,29 @@ def _capture_spark_hang_diagnostics(node):
 
     # 3. Socket state snapshot for UC port and Java connections.
     try:
-        # `set -o pipefail` is required so the pipeline's exit code reflects
-        # `ss`'s status, not `head`'s (which is always 0 once any bytes
-        # flow). Without it, the `|| echo ' no_matching_sockets'` sentinel
-        # is unreachable. `grep -E ... || true` swallows the exit-1 that
-        # `grep` produces on no matches, so the sentinel fires only when
-        # `ss` itself fails — which is the intended fallback semantics.
+        # Capture `ss`'s output into a shell variable first, then filter and
+        # truncate from that variable. Checking `ss`'s exit status directly
+        # (via the `if` test on the assignment) avoids the trap of the
+        # earlier `ss | grep | head || sentinel` pipeline: with
+        # `set -o pipefail` enabled, `head -50` closing the pipe early on
+        # outputs longer than 50 lines caused upstream commands to receive
+        # `SIGPIPE`, which made the pipeline status non-zero and falsely
+        # triggered the ` no_matching_sockets` sentinel; without `pipefail`,
+        # the sentinel was unreachable because `head`'s exit code (always
+        # 0 once any bytes flow) masked `ss`'s. Decoupling capture from
+        # truncation breaks both horns of that dilemma. `grep -E ... ||
+        # true` keeps the no-match case from producing an `exit 1` that
+        # would print confusing output, but plays no role in detecting
+        # `ss` failure.
         sockets = node.exec_in_container(
             [
                 "bash",
                 "-c",
-                "set -o pipefail;"
-                " ss -tnp 2>&1 | { grep -E ':8080|java' || true; }"
-                " | head -50"
-                " || echo ' no_matching_sockets'",
+                'if ss_output=$(ss -tnp 2>&1); then '
+                'echo "$ss_output" | { grep -E ":8080|java" || true; } | head -50; '
+                'else '
+                "echo ' no_matching_sockets'; "
+                'fi',
             ],
             nothrow=True,
             timeout=15,
--- a/tests/integration/test_database_delta/test.py
+++ b/tests/integration/test_database_delta/test.py
@@ -121,14 +121,17 @@ def _capture_spark_hang_diagnostics(node):
             )
         for pid in pids:
             try:
-                # Prefer non-force jstack; fall back to ``-F`` if the JVM is
-                # unresponsive to the attach mechanism.
+                # Prefer non-force `jstack`; fall back to `-F` if the JVM is
+                # unresponsive to the attach mechanism. The fallback must
+                # happen BEFORE `head` truncates, otherwise `head`'s exit
+                # code (always 0 once any bytes flow) masks `jstack`'s and
+                # the `||` branch never fires.
                 dump = node.exec_in_container(
                     [
                         "bash",
                         "-c",
-                        f"jstack {pid} 2>&1 | head -500"
-                        f" || jstack -F {pid} 2>&1 | head -500",
+                        f"(jstack {pid} 2>&1 || jstack -F {pid} 2>&1)"
+                        f" | head -500",
                     ],
                     nothrow=True,
                     timeout=30,
@@ -159,11 +162,19 @@ def _capture_spark_hang_diagnostics(node):
 
     # 3. Socket state snapshot for UC port and Java connections.
     try:
+        # `set -o pipefail` is required so the pipeline's exit code reflects
+        # `ss`'s status, not `head`'s (which is always 0 once any bytes
+        # flow). Without it, the `|| echo ' no_matching_sockets'` sentinel
+        # is unreachable. `grep -E ... || true` swallows the exit-1 that
+        # `grep` produces on no matches, so the sentinel fires only when
+        # `ss` itself fails — which is the intended fallback semantics.
         sockets = node.exec_in_container(
             [
                 "bash",
                 "-c",
-                "ss -tnp 2>&1 | grep -E ':8080|java' | head -50"
+                "set -o pipefail;"
+                " ss -tnp 2>&1 | { grep -E ':8080|java' || true; }"
+                " | head -50"
                 " || echo ' no_matching_sockets'",
             ],
             nothrow=True,
--- a/tests/integration/test_database_delta/test.py
+++ b/tests/integration/test_database_delta/test.py
@@ -84,6 +84,96 @@ def started_cluster():
         cluster.shutdown()
 
 
+def _capture_spark_hang_diagnostics(node):
+    """
+    Capture diagnostic data when a Spark query times out or otherwise fails
+    unexpectedly. Each capture is wrapped in its own ``try``/``except`` so a
+    failure in one diagnostic does not prevent collection of the others.
+
+    None of these are intended to recover the test. They only produce data
+    for post-mortem analysis of the chronic Spark hang on
+    ``Integration tests (arm_binary, distributed plan, 4/4)`` and similar
+    shards, where on ``subprocess.TimeoutExpired`` the Spark JVM is killed
+    by SIGKILL and ``stdout``/``stderr`` are typically truncated to whatever
+    Spark logged before the hang point — leaving CIDB without enough
+    information to localize the stuck thread.
+
+    Captured:
+      * ``jstack`` thread dumps of every running ``org.apache.spark`` JVM
+        (shows where the JVM is stuck)
+      * Unity Catalog HTTP liveness probe on ``http://localhost:8080``
+        (shows whether UC is responsive while Spark is hung)
+      * ``ss -tnp`` snapshot of sockets to UC port ``8080`` and ``java``
+        connections (shows whether Spark holds open sockets to UC)
+    """
+    # 1. Thread dumps of running Spark JVMs.
+    try:
+        pids_out = node.exec_in_container(
+            ["bash", "-c", "pgrep -f 'org.apache.spark' || true"],
+            nothrow=True,
+            timeout=30,
+        )
+        pids = [p.strip() for p in (pids_out or "").splitlines() if p.strip()]
+        if not pids:
+            print(
+                "Spark hang diag: no running 'org.apache.spark' processes"
+                " (already exited)"
+            )
+        for pid in pids:
+            try:
+                # Prefer non-force jstack; fall back to ``-F`` if the JVM is
+                # unresponsive to the attach mechanism.
+                dump = node.exec_in_container(
+                    [
+                        "bash",
+                        "-c",
+                        f"jstack {pid} 2>&1 | head -500"
+                        f" || jstack -F {pid} 2>&1 | head -500",
+                    ],
+                    nothrow=True,
+                    timeout=30,
+                )
+                print(f"Spark hang diag: jstack of PID {pid}:\n{dump}")
+            except Exception as je:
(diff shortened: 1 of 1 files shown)


groeneai:
cc @antonio2368 @kssenii — could you review? Small diagnostic-only change in `tests/integration/test_database_delta/test.py` to capture `jstack` thread dumps, Unity Catalog liveness, and `ss` socket state when `execute_spark_query` raises (typically `TimeoutExpired`). Antonio, you authored PR #102649 (the batching fix), so you have the most context on this test; Ksenia owns `comp-datalake`. No behavior change on the passing path — extra diagnostics run only inside the existing `except` block and each is wrapped in its own `try`/`except`.

groeneai:
**Pre-PR validation gate** (session `cron:clickhouse-ci-task-worker:20260513-074500`):

This PR is **diagnostic-only** — it does not change behavior on the passing path, only adds extra captures inside the existing exception handler of `execute_spark_query`. The standard "deterministic repro + both directions demonstrated" gate is for code fixes; for a defensive-diagnostic change the relevant checks are:

a) **Tools exist in the test container?** Yes — verified `which jstack` (`/usr/bin/jstack`), `which curl` (`/usr/bin/curl`), `which ss` (`/usr/bin/ss`) inside `clickhouse/integration-test-with-unity-catalog:latest`. JVM is OpenJDK 17, so `jstack` is the standard tool.

b) **Cannot break a passing test?** Yes — the helper is called only inside the `except Exception` branch, after the existing log capture. The hot path (line 104–125) is untouched.

c) **Cannot extend test failure time noticeably?** Yes — all `exec_in_container` calls use short timeouts (15–30 s) and the diagnostics are wrapped in try/except, so timeouts/failures inside the helper print a one-line error and continue. Worst case adds ≈75 s to an already-failing test.

d) **Test intent preserved?** Yes — no test assertion or query changed; only the failure-mode logging is richer.

e) **Why a fix is not in this PR.** The root cause of the residual hang is unknown — CIDB rows always stop at `ObjectStore.setMetaStoreSchemaVersion`, so we do not know whether Spark is then blocked on a Unity Catalog HTTP call, a Derby file lock, or somewhere else. Reproducing requires a local rebuild with `ENABLE_RUST=1` (~45 min) plus iterative reproduction, which does not fit a single CI-monitor cron run. This PR is the next-best actionable step: the next master hit will give us a `jstack`-resolved thread location instead of "Spark stopped logging".

Recent failure history (post-PR #102649): 2 master hits in 30 days, both on `Integration tests (arm_binary, distributed plan, 4/4)` (2026-05-11). P3 rare-flaky band.

tests/integration/test_database_delta/test.py
```diff
@@ -84,6 +84,96 @@ def started_cluster():
...
+        (shows where the JVM is stuck)
+      * Unity Catalog HTTP liveness probe on ``http://localhost:8080``
+        (shows whether UC is responsive while Spark is hung)
+      * ``ss -tnp`` snapshot of sockets to UC port ``8080`` and ``java``
+        connections (shows whether Spark holds open sockets to UC)
+    """
+    # 1. Thread dumps of running Spark JVMs.
+    try:
+        pids_out = node.exec_in_container(
+            ["bash", "-c", "pgrep -f 'org.apache.spark' || true"],
```
antonio2368: `pgrep -f 'org.apache.spark'` can match the `bash -c` wrapper itself, so when no Spark JVM is left you sometimes get the wrapper PID back and `jstack` runs against it. Easy fix:

```python
["bash", "-c", "pgrep -f '[o]rg.apache.spark' || true"]
```

antonio2368 (commented):
Mostly small stuff, but the `jstack` fallback doesn't actually fall back — `head` masks `jstack`'s exit code, so `-F` never runs. Same pattern bites the `ss` sentinel.
@groeneai address all the comments

tests/integration/test_database_delta/test.py
```diff
@@ -84,6 +84,96 @@ def started_cluster():
...
+        for pid in pids:
+            try:
+                # Prefer non-force jstack; fall back to ``-F`` if the JVM is
+                # unresponsive to the attach mechanism.
+                dump = node.exec_in_container(
+                    [
+                        "bash",
+                        "-c",
+                        f"jstack {pid} 2>&1 | head -500"
+                        f" || jstack -F {pid} 2>&1 | head -500",
```
antonio2368: The `-F` fallback never runs. `head -500` exits 0 as soon as it sees any byte, and `jstack`'s errors go through `2>&1` into `head`, so the left side of the `||` succeeds even when `jstack` failed. This is the unresponsive-JVM path the PR exists to capture.

```python
f"(jstack {pid} || jstack -F {pid}) 2>&1 | head -500"
```

or `set -o pipefail;` at the start if you'd rather keep the current shape.

tests/integration/test_database_delta/test.py
```diff
@@ -84,6 +84,96 @@ def started_cluster():
...
+                    ],
+                    nothrow=True,
+                    timeout=30,
+                )
+                print(f"Spark hang diag: jstack of PID {pid}:\n{dump}")
+            except Exception as je:
+                print(f"Spark hang diag: jstack PID {pid} failed: {str(je)}")
```
antonio2368, line 153: If executors are also stuck you'll pay 30s per PID. Usually just one driver, so probably fine — `head -n 3` on the PID list would cap it if you want a belt.

tests/integration/test_database_delta/test.py
```diff
@@ -84,6 +84,96 @@ def started_cluster():
...
+        print(f"Spark hang diag: UC liveness probe failed: {str(e)}")
+
+    # 3. Socket state snapshot for UC port and Java connections.
+    try:
+        sockets = node.exec_in_container(
+            [
+                "bash",
+                "-c",
+                "ss -tnp 2>&1 | grep -E ':8080|java' | head -50"
+                " || echo ' no_matching_sockets'",
```
antonio2368: Same `head`-eats-exit-status thing here: when nothing matches, `grep` fails but `head` succeeds, so `|| echo ' no_matching_sockets'` never runs and you get empty output instead of the sentinel. `set -o pipefail` and `{ grep ... || true; }` would do it.

tests/integration/test_database_delta/test.py
groeneai: Fixed in `720c43fe7e6` — `pgrep -f '[o]rg.apache.spark' || true`. The `[o]` character class matches a literal `o`, so the regex still matches a real Spark JVM cmdline, but the regex string itself contains `[o]` instead of plain `o` and therefore does not match the wrapper's own `bash -c` command line. Verified the trick in isolation (real Spark cmdline `java ... org.apache.spark.executor.CoarseGrained...` still matches; wrapper cmdline `bash -c "pgrep -f '[o]rg.apache.spark' || true"` does not).

tests/integration/test_database_delta/test.py
groeneai: Fixed in `87abec5f9fc` — restructured to `(jstack {pid} 2>&1 || jstack -F {pid} 2>&1) | head -500` so the fallback runs inside the subshell, before `head` truncates. Smoke-verified: forcing `jstack` to fail (`false` substitute) now triggers the `-F` branch; with the old shape it did not.

tests/integration/test_database_delta/test.py
groeneai: Capped in `720c43fe7e6` — `pids = pids[:3]` in Python. Belt for the pathological case where driver + several executors are all hung; normal case (one driver JVM per query) is unchanged.

tests/integration/test_database_delta/test.py
groeneai: Fixed in `87abec5f9fc` — `set -o pipefail; ss -tnp 2>&1 | { grep -E ':8080|java' || true; } | head -50 || echo ' no_matching_sockets'`. `pipefail` propagates `ss`'s exit status through `head`; `grep ... || true` deliberately swallows the exit-1 that `grep` produces on no matches, so the sentinel fires only when `ss` itself fails. Smoke-verified that forcing `ss` to fail now triggers the sentinel.

groeneai:
@antonio2368 — addressed all four inline comments in two commits.

**`87abec5f9fc`** — Fix `jstack` and `ss` fallback unreachability:

| Block | Old shape | New shape |
| --- | --- | --- |
| `jstack` (#2) | `jstack <pid> 2>&1 \| head -500 \|\| jstack -F <pid> 2>&1 \| head -500` | `(jstack <pid> 2>&1 \|\| jstack -F <pid> 2>&1) \| head -500` |
| `ss` (#4) | `ss -tnp 2>&1 \| grep -E ':8080\|java' \| head -50 \|\| echo ' no_matching_sockets'` | `set -o pipefail; ss -tnp 2>&1 \| { grep -E ':8080\|java' \|\| true; } \| head -50 \|\| echo ' no_matching_sockets'` |

For `ss`, `pipefail` propagates `ss`'s exit through `head`, and `grep ... \|\| true` deliberately swallows `grep`'s exit-1 on no matches so the sentinel fires only when `ss` itself fails (the intended semantics).

**`720c43fe7e6`** — Avoid `pgrep` self-match and cap diagnosed PIDs:

| Block | Old shape | New shape |
| --- | --- | --- |
| `pgrep` (#1) | `pgrep -f 'org.apache.spark' \|\| true` | `pgrep -f '[o]rg.apache.spark' \|\| true` |
| pid loop (#3) | `for pid in pids:` (unbounded) | `pids = pids[:3]` (Python cap) |

The `[o]` character class matches a literal `o`, so the regex still matches a real Spark JVM cmdline, but the regex string contains `[o]` rather than plain `o` and therefore does not match the wrapper's own `bash -c` cmdline.

Smoke-verified each change in isolation:
- `jstack` fallback: forcing `jstack` to fail (substitute `false`) triggers the `-F` branch with the new shape; old shape silently swallowed the failure.
- `ss` sentinel: forcing `ss` to fail triggers the sentinel; success + no `grep` matches produces empty output (no false-positive sentinel); success + matches produces the matched lines.
- `pgrep` self-match: regex `[o]rg.apache.spark` matches `java ... org.apache.spark.executor.CoarseGrained...` (real Spark) but does not match `bash -c "pgrep -f '[o]rg.apache.spark' || true"` (wrapper). Old regex matched both.
- PID cap: simple Python slice, normal one-driver case unchanged.

The changes are diagnostic-only and run only on the already-failing path inside `execute_spark_query`'s `except` block — a passing test cannot be made to fail by them.

tests/integration/test_database_delta/test.py
```diff
@@ -84,6 +84,119 @@ def started_cluster():
...
+        # `grep` produces on no matches, so the sentinel fires only when
+        # `ss` itself fails — which is the intended fallback semantics.
+        sockets = node.exec_in_container(
+            [
+                "bash",
+                "-c",
+                "set -o pipefail;"
+                " ss -tnp 2>&1 | { grep -E ':8080|java' || true; }"
+                " | head -50"
+                " || echo ' no_matching_sockets'",
```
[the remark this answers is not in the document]
antonio2368: @groeneai what do you think?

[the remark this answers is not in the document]
groeneai: Confirmed and fixed in `37e278daa46` — decoupled `ss` execution from filtering/truncation so `head -50`'s SIGPIPE on early close cannot influence the success/failure decision.

```bash
if ss_output=$(ss -tnp 2>&1); then
    echo "$ss_output" | { grep -E ":8080|java" || true; } | head -50
else
    echo ' no_matching_sockets'
fi
```

The previous `set -o pipefail; ss | grep | head -50 || sentinel` shape was indeed broken: with `pipefail`, `ss` writing into a closed `head` pipe got `SIGPIPE` (exit 141), which propagated as the pipeline status and false-fired the sentinel — exactly the diagnostic-corruption scenario you described.

Reproduction:

```bash
set -o pipefail
yes ':8080 estab pid=java' | head -100000 \
  | { grep -E ':8080|java' || true; } \
  | head -50 \
  || echo ' no_matching_sockets'
# Output ends with: ' no_matching_sockets'   <-- false positive
```

Smoke-verified the fix across six scenarios:

| # | `ss` produces | `ss` exit | Expected | Result |
|---|---|---|---|---|
| A | 200 matching lines (fits pipe buffer) | 0 | 50 lines, no sentinel | ✅ |
| B | 100,000 matching lines (forces SIGPIPE) | 0 | 50 lines, no sentinel | ✅ |
| C | 10 matching lines | 0 | 10 lines, no sentinel | ✅ |
| D | only non-matching lines | 0 | empty, no sentinel | ✅ |
| E | error message, exit 2 | 2 | sentinel | ✅ |
| F | command not found | 127 | sentinel | ✅ |

`@antonio2368` — third fix-up pushed; the other three comments were already addressed in `87abec5f9fc` and `720c43fe7e6`. PTAL.

groeneai:
Third fix-up pushed: `37e278daa46` — addresses the `pipefail` + `head -50` SIGPIPE false-fire on the `ss` block flagged by @clickhouse-gh[bot] and seconded by @antonio2368.

**Root cause of the v2 bug.** The previous shape `set -o pipefail; ss | grep | head -50 || sentinel` looked safe but had a subtle failure mode: when `ss` produced more than 50 matching lines, `head -50` closed its stdin after reading 50, upstream `ss`/`grep` got `SIGPIPE` (exit 141), `pipefail` propagated that as the pipeline status, and the `||` falsely fired ` no_matching_sockets` — exactly when the diagnostic data is most useful (a container under load with many sockets).

**Fix.** Capture `ss`'s output into a shell variable first; test `ss`'s exit directly via `if ss_output=$(ss -tnp 2>&1); then ...`. The truncation stage operates on the captured variable with no `pipefail`, so `head`'s SIGPIPE cannot mask success. Sentinel fires only when `ss` itself fails to run (e.g., binary missing).

```bash
if ss_output=$(ss -tnp 2>&1); then
    echo "$ss_output" | { grep -E ":8080|java" || true; } | head -50
else
    echo ' no_matching_sockets'
fi
```

**Pre-PR validation gate (diagnostic-only change).**

| | |
|---|---|
| **(a) Deterministic repro** | `set -o pipefail; yes ":8080 estab pid=java" \| head -100000 \| { grep -E ":8080\|java" \|\| true; } \| head -50 \|\| echo " no_matching_sockets"` reliably ends with ` no_matching_sockets` — bug deterministic for any input >50 lines with backpressure on `ss`. |
| **(b) Root cause** | `pipefail` propagates `head -50`'s SIGPIPE-on-early-close from upstream as pipeline failure, triggering the `\|\|` sentinel even when `ss` succeeded. |
| **(c) Fix matches root cause** | Decouples `ss` execution from filtering/truncation so the truncation stage cannot influence success/failure detection. Direct exit-status check via shell `if`. |
| **(d) Test intent preserved** | Diagnostic-only helper. Runs only inside `execute_spark_query`'s `except` branch. Cannot make a passing test fail. |
| **(e) Both directions demonstrated** | Old pattern: 51-line output ending with false ` no_matching_sockets`. New pattern: clean 50-line output, no sentinel. |
| **(f) Fix is general** | Audited the other 3 helpers (`pgrep`, `jstack`, `curl`) in `_capture_spark_hang_diagnostics`: none of them combine `pipefail` + pipe-to-`head` + sentinel `\|\|`, so none have this anti-pattern. Only the `ss` block needed the fix. |

All four prior inline review comments from @antonio2368 (the original review) remain addressed by `87abec5f9fc` and `720c43fe7e6`. PTAL.

PR #104813, opened 2026-05-13
Add test for reading invalid RMV during restore 

nihalzp:

Add test for the fix in https://github.com/ClickHouse/ClickHouse/pull/104065. Even though we do not allow such invalid RMV to be created. It is possible to add such RMV in a hacky way via `ATTACH` so we should be able to read them.

### Version info
- Merged into: `26.6.1.885`


tests/queries/0_stateless/04218_refreshable_mv_secondary_create.sh
```diff
@@ -0,0 +1,55 @@
...
+# shellcheck source=../shell_config.sh
+. "$CURDIR"/../shell_config.sh
+
+db="${CLICKHOUSE_DATABASE}_r"
+mv_uuid=$(${CLICKHOUSE_CLIENT} -q "SELECT generateUUIDv4()")
+
+${CLICKHOUSE_CLIENT} --distributed_ddl_output_mode=none <<EOF
+CREATE DATABASE ${db}
+    ENGINE = Replicated('/test/{database}/refreshable_mv_secondary_create', 'shard1', 'replica1');
+CREATE TABLE ${db}.target_tbl (id Int64, val Float64) ENGINE = MergeTree ORDER BY id;
```
rschu1ze, line 23: Could add a `Note:` comment here that `target_tbl` isn't replicated.

rschu1ze (approved):

rschu1ze:
@nihalzp Let's merge from `master`, and then let's merge this PR? The [bot](https://github.com/ClickHouse/ClickHouse/pull/104813#issuecomment-4709513847) starts annoy :D

PR #104815, opened 2026-05-13
Install lldb in CI Docker images alongside gdb

leshikus:
`clickhouse-test`'s `print_c_stacktraces` uses `lldb` to attach to a running (or hung) ClickHouse server and dump C++ backtraces into `c_stacktraces.log` on a test timeout or the end-of-run hung check. For that to work, `lldb` has to be present on the machine running the tests. How it gets there differs by platform.

## How `lldb` is provisioned per platform

- **Linux (CI docker images) — this PR.** `lldb` is baked into `clickhouse/fasttest` and `clickhouse/test-base`. The LLVM apt repo is already configured for the matching `LLVM_VERSION` in both Dockerfiles, so this is a single `apt-get install --yes lldb-${LLVM_VERSION}`, a `/usr/bin/lldb` symlink, and `chmod u+s`, mirroring the existing `gdb` setup. The suid bit is the point: the Fast test job runs its container unprivileged (`docker run --user $(id -u):$(id -g)`), so a runtime `apt-get install lldb` cannot work there (no root), and attaching to the server with `ptrace` needs the elevated privilege that the setuid `lldb` provides. Every test image ultimately depends on one of these two — `stateless-test`, `integration/runner`, `integration/base`, and `performance-comparison` inherit from `test-base`; `cctools` from `fasttest`.
- **macOS (Darwin fast test) — unchanged, for contrast.** The Darwin runner is a bare host, not a container, so there is no image to bake into. There `lldb` is provided at runtime by `_ensure_lldb_installed` in `tests/clickhouse-test`, which runs `brew install llvm` and prepends the keg-only `opt/llvm/bin` to `PATH`. That helper is macOS-only and stays.

So `print_c_stacktraces` gets its debugger from the image on Linux and from Homebrew on macOS, and produces `c_stacktraces.log` on both.

`gdb` is intentionally left in place. It is still driven directly by the fuzzer (`ci/jobs/scripts/fuzzer/run-fuzzer.sh`), perf comparison (`ci/jobs/scripts/perf/compare.sh`), unit tests (`ci/jobs/unit_tests_job.py`), stress (`ci/jobs/scripts/stress/stress.py`), the upgrade runner (`tests/docker_scripts/upgrade_runner.sh`), and the integration test harness. Migrating those to `lldb` is out of scope.

Related: https://github.com/ClickHouse/ClickHouse/pull/112058


maxknv (commented):
why do we need to migrate? what is the size added to the image?

leshikus:
@maxknv It is recommended to use `lldb` on Darwin; supporting both `gdb` and `lldb` is harder than just `gdb`; I believe I will be able to remove `gdb` at some point.

leshikus:
@maxknv Adding lldb to test-base costs ~120 MB per image (lldb-21 ~4 MB + liblldb-21 ~18 MB + libclang-cpp21 ~57 MB + python3-lldb-21 + a few smaller deps). But some images like fastest already contain libclang-cpp21 - thus we got only +50 MB for them.

maxknv:
> @maxknv It is recommended to use `lldb` on Darwin; supporting both `gdb` and `lldb` is harder than just `gdb`; I believe I will be able to remove `gdb` at some point.

I think it's relatively easy. And we can avoid adding a new package to many Docker images (it's the base image you're updating), at least until it's confirmed that lldb works for us

maxknv (approved):

ci/docker/fasttest/Dockerfile
```diff
@@ -115,6 +115,18 @@ COPY --from=clickhouse/cctools:859fb360308eb8ac47ad /opt/gdb /opt/gdb
 RUN chmod u+s /opt/gdb/bin/gdb
 ENV PATH="/opt/gdb/bin:${PATH}"
 
+# Native debugger used by clickhouse-test's `print_c_stacktraces` to dump
+# server backtraces. lldb ships in the LLVM apt repo already configured
+# above; symlink the versioned binary into /usr/bin so callers can run
+# `lldb` directly, and grant suid so non-root tests can still attach.
+# gdb is kept above for the fuzzer, perf comparison, stress, upgrade, and
+# integration-test harnesses that still drive it directly.
+RUN apt-get update \
```
Algunenano: This block should be with the block installing clang. It's the same repo, same version, etc

ci/docker/test-base/Dockerfile
```diff
@@ -108,6 +108,18 @@ RUN apt-get update \
 COPY --from=clickhouse/cctools:859fb360308eb8ac47ad /opt/gdb /opt/gdb
 ENV PATH="/opt/gdb/bin:${PATH}"
 
+# Native debugger used by clickhouse-test's `print_c_stacktraces` to dump
+# server backtraces. lldb ships in the LLVM apt repo already configured
+# above; symlink the versioned binary into /usr/bin so callers can run
+# `lldb` directly, and grant suid so non-root tests can still attach.
+# gdb is kept above for the fuzzer, perf comparison, stress, upgrade, and
+# integration-test harnesses that still drive it directly.
+RUN apt-get update \
```
Algunenano: Same here. Please treat lldb as part of llvm (because it is). No need to fetch the repo, install and track the version and clean up

ci/docker/fasttest/Dockerfile
leshikus: Fixed in https://github.com/ClickHouse/ClickHouse/pull/104815/commits/b70c8723a17889315206a1fdb9854c4c9f635352

ci/docker/test-base/Dockerfile
leshikus: Fixed in https://github.com/ClickHouse/ClickHouse/pull/104815/commits/b70c8723a17889315206a1fdb9854c4c9f635352

Algunenano (approved):

PR #104817, opened 2026-05-13
`translateUTF8`: Fix slightly incorrect error message

### LINKED ISSUE #104786: `translateUTF8` error message reports byte sizes instead of code point counts

_Retrospective finding from a historical scan of [PR #72893](https://github.com/ClickHouse/ClickHouse/pull/72893) (merged 2024-12-17). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

When `translateUTF8` rejects a query due to map_from being shorter than map_to (in code points), the error message incorrectly reports byte sizes. For example, `SELECT translateUTF8('x', 'αβ', 'abc')` shows 'Size of the second argument: 4, size of the third argument: 3' when the correct code point counts are 2 and 3.

**Root cause:** translate.cpp:368: The error message uses `map_from.size()` (byte size = 4 for 'αβ') instead of `map_from_size` (code point count = 2). The comparison at line 367 correctly uses code point counts, but the message reports wrong values.

**Why we believe this is a bug:** FunctionTranslate::executeImpl (translate.cpp:357-368) computes `map_from_size` and `map_to_size` as code point counts for UTF8 (using `UTF8::countCodePoints`), but the exception at line 368 uses `map_from.size()` and `map_to.size()` which are byte sizes.

**Affected locations:**
- `src/Functions/translate.cpp:368` — Exception message uses byte sizes instead of code point counts

**Impact:** Users receive confusing error messages where the reported sizes don't match the comparison logic. For example, when 'αβ' (2 code points, 4 bytes) is compared to 'abc' (3 code points, 3 bytes), the error says the second argument (4) is shorter than the third (3), which makes no sense because 4 > 3. The actual code point comparison (2 < 3) is correct.

Yes — confirmed on current `master` (commit `7bd0fa28146`).

### How to reproduce

_See root cause description and affected locations above._

### Expected behavior

```
FIXED: Error message correctly reports code point count (2)
```

```
BUG: Error message reports byte size (4) instead of code point count (2)
```

### Additional context

**Suggested fix:** Change line 368 to use `map_from_size` and `map_to_size` instead of `map_from.size()` and `map_to.size()`: `throw Exception(ErrorCodes::BAD_ARGUMENTS, "...", getName(), map_from_size, map_to_size);`

Found during automated review of [PR #72893](https://github.com/ClickHouse/ClickHouse/pull/72893).

---
_ClickGapAI · Confidence: HIGH · Severity: P3 · Finding: `h_pr72893_001`_

### Version info
- Resolved by: #104817
- Merged into: `26.5.1.918`

rschu1ze:
The error message now complains about the code point count instead of the byte count.

Resolves: #104786

### Version info
- Merged into: `26.5.1.918`


Algunenano (approved):

Algunenano:
Fuzzer issue fixed here: https://github.com/ClickHouse/ClickHouse/pull/104858

alexey-milovidov:

rschu1ze:
01710_projection_additional_filters: https://github.com/ClickHouse/ClickHouse/issues/104219 (fixed)

PR #104820, opened 2026-05-13
Minimize macOS user_data, delegate provisioning to runner-init

leshikus:
## Motivation

macOS EC2 instances do not support updating `user_data` the way Linux instances do — once an instance is launched, its `user_data` is effectively frozen. Anything placed in `ci/infra/scripts/user_data_macos.txt` therefore cannot be iterated on without rebuilding the AMI or relaunching the instance, which makes the macOS runner setup painful to evolve.

## Change

Strip `user_data_macos.txt` down to a minimal bootstrap: set `HOME`, fetch the runner-init script from `s3://github-runners-data/cloud-init/${INIT_ENVIRONMENT}.py`, clean the runner work directory, and exec the script as `ec2-user`.

The following one-time setup is **removed** from `user_data` and will be moved into the S3-hosted runner-init script (which can be updated by simply re-uploading to S3, without touching live AMIs):

- Disabling `com.apple.bluetoothd`
- Homebrew install + update
- `brew install` of `ca-certificates`, `curl`, `gh`, `jq`, `pigz`, `ripgrep`, `zstd`, `python@3`, `wget`, `unzip`, `gnu-sed`, `grep`, `bash`, `coreutils`
- Python venv at `~/venv` with `boto3`, `pygithub`, `requests`, `urllib3`, `unidiff`, `dohq-artifactory`, `pyjwt`
- `PATH` setup in `~/.bash_profile` and `~/.zprofile`
- Amazon CloudWatch agent install + configuration (fetched from SSM, log path patched to `/var/log/amazon/ec2/ec2-macos-init.log`)
- GitHub Actions runner download and unpack
- `/tmp/clickhouse-ci-macos-runner.success` marker gating

After this change, the user_data path is short enough to read at a glance, and iterating on macOS provisioning no longer requires an AMI rebuild.

...

### Version info
- Merged into: `26.5.1.711`

### DIFF
--- a/ci/infra/scripts/user_data_macos.txt
+++ b/ci/infra/scripts/user_data_macos.txt
@@ -1,138 +1,16 @@
 #!/usr/bin/env bash
-# Script to prepare a clean macOS machine as a GitHub Actions runner
-# Based on prepare-ci-ami.sh but adapted for macOS
+# Bootstrap script for macOS GitHub Actions runners.
+# Provisioning (Homebrew, Python venv, CloudWatch agent, actions-runner install, etc.)
+# lives in the runner-init script fetched from S3 — macOS EC2 user_data cannot be
+# updated after launch, so we keep this file minimal.
 
 set -xeuo pipefail
 trap reboot EXIT
 
-echo "Running macOS runner prepare script"
-
-# Set HOME if not already set (EC2 user_data context may not have it)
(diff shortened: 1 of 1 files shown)


maxknv (approved):
ok, but why is the runner-init.py update not in this PR?

alexey-milovidov:
No answer in three days, merging.

PR #104854, opened 2026-05-13
Eliminate redundant joins

### LINKED ISSUE #101451: Request for optimization - skip unused joins
### Company or project name

Gridcog

### Use case

- I'm dealing with dumb BI tools that write dumb SQL queries that I have no control over.
  - They will often attempt to run a big complicated query with just `select distinct Dimension from (...)` to figure out how to populate a filter dropdown.
  - There are a bunch of branches to the big complex query that can be completely skipped for calculating this.
 
- Alternatively, I've written a (arguably crap but bear with me) View that joins on demand to compute some columns. When people select from the view but don't care about the columns in question, the join is still executed.

Example query:
```sql
create table sandbox.perf1 (dt DateTime64, x Float64) ENGINE = MergeTree order by dt;
insert into sandbox.perf1
    select addHours(toDateTime('2021-01-01'), number), number from numbers(365*24);

with maxX as (
    select date(dt) as date, max(x) as y
    from sandbox.perf1
    group by date
),
joined as (
    select dt, date(dt), x, y
    from sandbox.perf1
        left any join maxX on date(perf1.dt) = maxX.date
)
select dt, x from joined; -- the join to maxX is a `left any` join, and we don't read maxX.date or maxX.y, so we should be able skip both the join and maxX completely
```

Given a join that does not contribute to the result, it would be great if clickhouse could skip it completely, and so potentially skip the table/subquery it's reading from completely as well.

Hand-write sane sql and materalize everything instead of relying on BI tools and views

### Additional context

Only tested on Clickhouse Cloud 25.12

davenger:
Fixes https://github.com/ClickHouse/ClickHouse/issues/101451

- Improvement

Remove JOINs that do not affect the result, e.g. added by BI tools or other query generators. The implementation suggested by @geldot


src/Processors/QueryPlan/Optimizations/eliminateRedundantJoin.cpp
```diff
@@ -0,0 +1,152 @@
...
+    }
+    else if (isRight(op.kind) && is_at_most_one_match)
+    {
+        preserved_idx = 1;
+        preserved_side = JoinTableSide::Right;
+    }
+    else
+        return std::nullopt;
+
+    /// If the preserved side has type conversions (e.g. toNullable from join_use_nulls),
```
m-selmi, line 56: just out of curiosity, could this happen in an actual query?

m-selmi (commented):
I took a look out of interest, it looks good to me besides the issue pointed out by AI review. (small reproduction of the issue https://pastila.clickhouse.com/?0109b376/b71907ae9665635bc7a15d22f4781638#zULLNlPKz/V/EIMTtFKAYA==GCM)

PR #104858, opened 2026-05-13
Fix more non-deterministic function results found by the functions stress test

Algunenano:
Follow-up to #104710. Three more const/non-const divergences caught by the functions stress test (`caseWithExpression` and friends in `src/Functions/tests/gtest_functions_stress.cpp`). Each commit fixes one root cause and adds a stateless regression test.

- **`transform` const default**: the optimized cache built `default_column` by routing the constant default through `convertFieldToType`, which loses the source type and falls back to a raw numeric path. That mis-converted `Date`/`Date32` → `DateTime`/`DateTime64` (days were treated as seconds, e.g. `Date('2149-06-04')` rendered as `1970-01-01 18:12:13.000000`), emitted the underlying integer instead of the value name for `Enum` → `String`, and kept the trailing NUL padding for `FixedString` → `String`. Switched to `castColumn`, matching the surrounding code that already uses it for the `from` and `to` arrays. `caseWithExpression` rewrites to `transform` when all WHEN/THEN values are constant, so it inherited the fix.
- **`String` vs constant `FixedString` comparison**: `string_vector_constant` in `StringComparisonImpl`/`StringEqualsImpl` used a plain `memcmp`, while the vector-vs-vector path used `memcmpSmallLikeZeroPaddedAllowOverflow15` (zero-padded, per SQL semantics `toFixedString('abc', 5) = 'abc'`). When the right operand was a constant `FixedString` the two paths disagreed: `equals(materialize(''::String), '\0\0\0\0'::FixedString(4))` returned `0` while every other arrangement returned `1`, with the same flip for `less`, `lessOrEquals`, `isDistinctFrom`, `isNotDistinctFrom`. Track whether the constant operand wraps a `FixedString` in `executeString` and dispatch to a new `stringVectorConstantFixedString` (and its symmetric form) that uses the zero-padded comparison.
- **`if` const-condition `FixedString` padding**: the non-const dispatch in `executeString` copies `FixedString` operands into a `String` result through `FixedStringSource` → `StringSink`, preserving the full `N` bytes including trailing NULs. The const-condition fast path used `castColumn`, which trims them. `ifNull` rewrites to `if` and inherited the divergence. Routed the const-condition cast through a helper that copies bytes verbatim, matching the non-const path. `multiIf` is intentionally left alone: it already trims consistently on both paths, so the stress test cannot flip its output.

Fix three more functions returning different results for the same input depending on whether arguments arrived as columns or constants: `transform` (and `caseWithExpression` through it) with a constant `Date`/`Date32`/`Enum`/`FixedString` default, comparison operators between a `String` and a constant `FixedString`, and `if`/`ifNull`/`nullIf` with a `FixedString` branch under a constant condition.

### Version info
- Merged into: `26.5.1.685`


src/Functions/if.cpp
antonio2368, line 1: Claude found couple more questionable results:

```sql
-- 10
SELECT length(if(CAST(0 AS UInt8), CAST(NULL AS Nullable(String)),
                 materialize(CAST(toFixedString('abc', 10) AS FixedString(10)))));
-- 3 (bug)
SELECT length(if(materialize(CAST(0 AS UInt8)), CAST(NULL AS Nullable(String)),
                 materialize(CAST(toFixedString('abc', 10) AS FixedString(10)))));

-- 10
SELECT length(if(CAST(1 AS UInt8),
                 materialize(CAST(toFixedString('abc', 10) AS FixedString(10))),
                 CAST(NULL AS Nullable(String))));
-- 3 (bug)
SELECT length(if(materialize(CAST(1 AS UInt8)),
                 materialize(CAST(toFixedString('abc', 10) AS FixedString(10))),
                 CAST(NULL AS Nullable(String))));
```

With a materialized condition, dispatch falls through to `executeForNullThenElse`, which still uses `castColumn` at `src/Functions/if.cpp:1206` and `:1252`. Both should use `castForIf`.

Add repros to `04236_ifnull_if_fixedstring_string_trim.sql`.

antonio2368, line 1: And another:

```sql
-- 3 (folded via castColumn at if.cpp:1499)
SELECT length(if(CAST(1 AS UInt8),
                 CAST(toFixedString('abc', 10) AS FixedString(10)),
                 materialize(CAST('zz' AS String))));
-- 10 (runtime executeString preserves padding)
SELECT length(if(materialize(CAST(1 AS UInt8)),
                 CAST(toFixedString('abc', 10) AS FixedString(10)),
                 materialize(CAST('zz' AS String))));

-- 3
SELECT length(if(CAST(0 AS UInt8),
                 materialize(CAST('zz' AS String)),
                 CAST(toFixedString('abc', 10) AS FixedString(10))));
-- 10
SELECT length(if(materialize(CAST(0 AS UInt8)),
                 materialize(CAST('zz' AS String)),
                 CAST(toFixedString('abc', 10) AS FixedString(10))));
```

With a constant condition and a constant selected branch, `ActionsDAG::addFunctionImpl` invokes `getConstantResultForNonConstArguments` which still uses `castColumn` at `src/Functions/if.cpp:1499`, so the function gets folded to a trimmed constant at DAG-build time. With a materialized condition this path is skipped and `executeString` preserves padding.

Suggested fix — route through `castForIf` like the other call sites:

```diff
-        auto result = castColumn(potential_const_column, result_type);
+        auto result = castForIf(potential_const_column, result_type);
         if (!isColumnConst(*result))
             return {};
```

src/Functions/FunctionsComparison.h
```diff
@@ -809,6 +871,8 @@ class FunctionComparison : public IFunction
         const ColumnString::Chars * c1_const_chars = nullptr;
         ColumnString::Offset c0_const_size = 0;
         ColumnString::Offset c1_const_size = 0;
+        bool c0_const_is_fixed_string = false;
+        bool c1_const_is_fixed_string = false;
```
antonio2368: Can you replace these 6 variables with:
```cpp
const ColumnString      * c0_const_string       = nullptr;
const ColumnFixedString * c0_const_fixed_string = nullptr;
const ColumnString      * c1_const_string       = nullptr;
const ColumnFixedString * c1_const_fixed_string = nullptr;
```

src/Functions/FunctionsComparison.h
```diff
@@ -280,6 +280,28 @@ struct StringComparisonImpl
         }
     }
 
+    /// `String` vector vs constant `FixedString`. `string_vector_constant` uses a plain
+    /// `memcmp`, which respects the constant's trailing NUL padding and disagrees with the
+    /// `String` vs `FixedString` vector path (which uses zero-padded comparison, per SQL
+    /// semantics that `toFixedString('abc', 5) = 'abc'`).
+    static void NO_INLINE stringVectorConstantFixedString(
```
antonio2368: naming style doesn't match other functions

Algunenano: This was on purpose. All other functions use NOLINT because they don't follow the project's style, but I didn't want to fix them all in this unrelated PR

antonio2368: I would prefer consistency tbh, but it's really not that important.

antonio2368 (approved):

PR #104859, opened 2026-05-13
Push LIMIT into aggregation-in-order for early termination

### LINKED ISSUE #105196: Optimization of `GROUP BY` with aggregation by sorting key.
### Company or project name

Not applicable.

### Use case

When running `GROUP BY key ORDER BY key ASC LIMIT N` where the `GROUP BY key` matches the table's primary key (sort order), ClickHouse reads the entire table even though the result could be computed from a tiny fraction of the data.

For example, on a 10M-row table with 100K distinct `user_id` values sorted by `user_id`, this query:

```
SELECT user_id, count() FROM t GROUP BY user_id ORDER BY user_id ASC LIMIT 10
```

reads all 10M rows, when in theory only ~1,000 rows (the first 10 distinct keys × ~100 rows each) need to be read. Since the data is physically sorted by the primary key, once 10 distinct keys have been seen, no subsequent row can belong to an earlier group, making further reading unnecessary.

Storage-level early termination for `GROUP BY key ORDER BY key ASC LIMIT N` when the `GROUP BY key` is a prefix of the table's primary key. Once `LIMIT` distinct key values have been fully consumed (i.e. the (LIMIT + 1)th distinct key is first observed), the MergeTree reader should cancel further reads. This would reduce r`ead_rows` from O(total rows) to O(LIMIT × average group size), a potentially orders-of-magnitude improvement.

This is distinct from the aggregation-layer optimization in PR https://github.com/ClickHouse/ClickHouse/pull/96630 (group_by_limit_pushdown), which bounds the hash table size but still scans all rows. The requested optimisation needs to propagate a cancellation signal from the aggregation step back into the MergeTree pipeline, similar in spirit to how optimize_read_in_order works for plain `ORDER BY ... LIMIT` queries.

- PR #96630 (group_by_limit_pushdown):
Reduces memory and aggregation CPU by maintaining a bounded heap, but does not reduce read_rows. Complementary but insufficient for I/O-bound workloads.

- `optimize_aggregation_in_order`: Processes data in sort order without a full hash table, which helps memory, but still does not terminate the scan early after `LIMIT` distinct keys are seen.

- Manual query rewrite: Using a subquery to pre-select the first N distinct keys, then joining back. This is brittle, requires knowledge of the data distribution, and shifts complexity to the application layer.
`max_rows_to_group_by with group_by_overflow_mode = 'any'`: Bounds the hash table but returns approximate/incomplete results and does not guarantee the correct top-N by sort order.

### Additional context

The following query reproduces the problem on any MergeTree table with a high-cardinality sort key:

```
CREATE TABLE test_groupby_early_termination (
    user_id String,
    payload String DEFAULT randomString(64)
) ENGINE = MergeTree() ORDER BY user_id;

INSERT INTO test_groupby_early_termination (user_id)
SELECT concat('user_', lpad(toString(number % 100000), 6, '0'))
FROM numbers(10000000);

-- Reads 10M rows; expected: ~1K
SELECT user_id, count() AS cnt
FROM test_groupby_early_termination
GROUP BY user_id
ORDER BY user_id ASC
LIMIT 10;
```

The `EXPLAIN` shows Aggregating → Sorting → Limit with no early-termination pushdown.

Related PRs:
https://github.com/ClickHouse/ClickHouse/issues/72610
https://github.com/ClickHouse/ClickHouse/issues/5329

### Version info
- Resolved by: #104859
- Merged into: `26.7.1.267` (included in `26.7` and later)

thevar1able:
- Performance Improvement

Push LIMIT into aggregation-in-order to enable early termination when GROUP BY key matches ORDER BY key and table sorting key, significantly reducing the number of rows read. 

Resolves https://github.com/ClickHouse/ClickHouse/issues/105196

image

### Version info
- Merged into: `26.7.1.267` (included in `26.7` and later)


Ergus:
Hi @thevar1able 

Let's merge with master to solve the conflicts and retry CI

src/Processors/Merges/Algorithms/FinishAggregatingInOrderAlgorithm.cpp
```diff
@@ -123,15 +124,24 @@ IMergingAlgorithm::Status FinishAggregatingInOrderAlgorithm::merge()
...
+
     /// At least one chunk should be fully aggregated.
     assert(!inputs_to_update.empty());
     Status status(inputs_to_update.back());
     inputs_to_update.pop_back();
 
     /// Do not merge blocks, if there are too few rows or bytes.
-    if (accumulated_rows >= max_block_size_rows || accumulated_bytes >= max_block_size_bytes)
+    bool need_flush = accumulated_rows >= max_block_size_rows || accumulated_bytes >= max_block_size_bytes;
+    bool limit_reached = limit_hint && groups_produced >= limit_hint;
```
Ergus: Minor:
`limit_hint` is not a boolean, so better to set explicit `!= 0` when testing logical conditions with it.
Also if these values won't change after set, better make them `const`.

src/Processors/Merges/Algorithms/FinishAggregatingInOrderAlgorithm.cpp
```diff
@@ -123,15 +124,24 @@ IMergingAlgorithm::Status FinishAggregatingInOrderAlgorithm::merge()
 
     addToAggregation();
 
+    /// Each iteration of merge() processes one group boundary.
+    ++groups_produced;
+
     /// At least one chunk should be fully aggregated.
     assert(!inputs_to_update.empty());
```
Ergus: chassert

src/Processors/QueryPlan/Optimizations/optimizeLimitForAggregationInOrder.cpp
```diff
@@ -0,0 +1,151 @@
...
+    /// limit past it would feed it a truncated prefix and produce wrong extremes.
+    /// WindowStep preserves rows but is not supported by this optimization.
+    if (typeid_cast<ExtremesStep *>(step) || typeid_cast<WindowStep *>(step))
+        return false;
+
+    auto * transforming = dynamic_cast<ITransformingStep *>(step);
+    if (!transforming)
+        return false;
+
+    return transforming->getDataStreamTraits().preserves_sorting
```
Ergus: Please us temporal variables here to help debugging in the future.

src/Core/Settings.cpp
```diff
@@ -3976,6 +3976,9 @@ Possible values:
 **See Also**
 
 - [GROUP BY optimization](/sql-reference/statements/select/group-by#group-by-optimization-depending-on-table-sorting-key)
+)", 0) \
+    DECLARE(Bool, optimize_aggregation_in_order_limit, true, R"(
```
Ergus, line 4076: This setting must be `experimental` and disabled by default in the first attempt.

src/Processors/Merges/Algorithms/FinishAggregatingInOrderAlgorithm.cpp
```diff
@@ -123,15 +124,24 @@ IMergingAlgorithm::Status FinishAggregatingInOrderAlgorithm::merge()
 
     addToAggregation();
 
+    /// Each iteration of merge() processes one group boundary.
+    ++groups_produced;
```
Ergus: IIUC this runs once per merge() call, but a single merge() aggregates an entire batch of group-rows from each input, so the comment is inacurate.

src/Core/SettingsChangesHistory.cpp
```diff
@@ -89,6 +89,7 @@ const VersionToSettingsChangesMap & getSettingsChangesHistory()
             {"use_top_k_dynamic_filtering_for_variable_length_types", true, false, "Disable `use_top_k_dynamic_filtering` for variable-length sort columns (e.g. `String`) by default; the previous behavior had the optimization apply unconditionally and is preserved under `compatibility`."},
             {"page_cache_max_coalesced_bytes", 16777216, 16777216, "New setting to bound the size of a single coalesced read used to populate the userspace page cache on cache miss."},
             {"input_format_column_name_matching_mode", "match_case", "auto", "Match input column names case-sensitively first and fall back to case-insensitive matching, instead of requiring an exact case match."},
+            {"optimize_aggregation_in_order_limit", true, true, "Push the `LIMIT` into aggregation-in-order for early termination when the `ORDER BY` is a prefix of the `GROUP BY` sort description."},
```
Ergus: This must mention it is a New setting.

tests/queries/0_stateless/04234_aggregation_in_order_limit_pushdown.sql
```diff
@@ -0,0 +1,100 @@
+-- Tags: no-random-settings
+
+-- Test: LIMIT push-down into aggregation-in-order
+-- When GROUP BY key = ORDER BY key = table PK and LIMIT is present,
+-- the aggregation should stop early after producing enough groups.
```
Ergus, line 5: We need some multithreaded tests.

The read-reduction assertion (PUSHDOWN_FIRES) pins max_threads = 1, so the FinishAggregatingInOrderAlgorithm cross-stream path (and the groups_produced logic) is never exercised under the part of the test that asserts an effect.

Ergus (changes requested):
Please, resolve the `SettingsChangesHistory` and setting name comments.

src/Processors/Merges/Algorithms/FinishAggregatingInOrderAlgorithm.cpp
thevar1able: Out of scope of this PR, I can change it in another one.

src/Core/Settings.cpp
thevar1able: Do we actually do that for optimization settings? Picked a [random similar PR](https://github.com/ClickHouse/ClickHouse/pull/104000/changes#diff-d33ff1f4e22c61b4aed54d20df5b89677cbd182442ce4263dc94cf099076462eR6043).

tests/queries/0_stateless/04234_aggregation_in_order_limit_pushdown.sql
```diff
@@ -0,0 +1,100 @@
...
+
```
Ergus, line 6: We also need some negative test with a filter between aggregation and limit. To confirm that the optimization does not push the limit past a row-count-reducing `FilterStep`. i.e `SELECT ... FROM (… GROUP BY key) WHERE c > N ORDER BY key LIMIT k`

tests/queries/0_stateless/04234_aggregation_in_order_limit_pushdown.sql
```diff
@@ -0,0 +1,100 @@
+-- Tags: no-random-settings
```
Ergus, line 1: Could we add a positive correctness test with `enable_parallel_replicas = 1` that compares results with the setting on vs off.

It will prove correctness on the distributed path or pins down that the optimization can't fire there. WDYT?

Ergus (approved):
LGTM

However, FWIS `optimizeLimitForAggregationInOrder` sets a `limit_hint` on the `AggregatingStep` with no guard for parallel replicas. 
So if the plan under parallel replicas still presents a final in-order `AggregatingStep`, the early-termination path is live there too?
And the new test files carry no-random-settings random-settings rotation will never flip parallel replicas on for them...

thevar1able:
Decided to make the new setting non-experimental and enabled by default; discussed with Robert in DM.

clickgapai:
Hi @thevar1able @Ergus — while reviewing this PR I found the following:

- **Coverage gap:** `LIMIT ... WITH TIES` guard in `optimizeLimitForAggregationInOrder` is untested — https://github.com/ClickHouse/ClickHouse/pull/108884

tests/performance/aggregation_in_order_limit.xml
```diff
@@ -0,0 +1,21 @@
...
+    <query>SELECT CounterID, count() FROM hits_100m_single GROUP BY CounterID ORDER BY CounterID ASC LIMIT 10 FORMAT Null</query>
+
+    <!-- Same with larger LIMIT -->
+    <query>SELECT CounterID, count() FROM hits_100m_single GROUP BY CounterID ORDER BY CounterID ASC LIMIT 100 FORMAT Null</query>
+
+    <!-- GROUP BY two-column PK prefix -->
+    <query>SELECT CounterID, EventDate, count() FROM hits_100m_single GROUP BY CounterID, EventDate ORDER BY CounterID, EventDate LIMIT 10 FORMAT Null</query>
+
+    <!-- ORDER BY is a prefix of GROUP BY (GROUP BY has extra non-PK column) -->
+    <query>SELECT CounterID, EventDate, RegionID, count() FROM hits_100m_single GROUP BY CounterID, EventDate, RegionID ORDER BY CounterID, EventDate LIMIT 10 FORMAT Null</query>
```
PedroTadim, line 20: @thevar1able the optimization should be applied as well when the order by is a subset of group by? Should a performance test be added for that? if the columns are not in the same order in both, it should also be applied. Can you confirm this?

thevar1able: https://github.com/ClickHouse/ClickHouse/pull/110400

PR #104875, opened 2026-05-13
Make arrayResize honor query cancellation

Algunenano:
arrayResize ignored max_execution_time and KILL QUERY because the inner fill loop of resizeConstantSize / resizeDynamicSize did not check the cancellation flag. A worst-case input could keep one thread busy for tens of seconds (or minutes under sanitizers) past any timeout.

Adds a static `CurrentThread::throwIfQueryCancelled()` helper for general reuse and polls it every 1024 iterations inside both resize variants. The existing inline caller in `MergeTreeIndexVectorSimilarity` migrates to the helper.

Make `arrayResize` honor query cancellation (`max_execution_time`, `KILL QUERY`).

### DIFF
--- a/src/Client/LocalConnection.cpp
+++ b/src/Client/LocalConnection.cpp
@@ -159,9 +159,10 @@ void LocalConnection::sendQuery(
     {
         /// The interactive cancel callback is now invoked from `CurrentThread::throwIfQueryCancelled`
         /// in addition to the pipeline executor's wait loop, so it can run concurrently from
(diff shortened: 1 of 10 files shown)


Algunenano:
There is an issue with this approach which is the same that we had with scalar subqueries: If the cancel packet isn't read (with TCP), then we don't know the query was KILLed. We should unify the behaviour and make it simpler and in a single place if possible.

PR #72540, opened 2024-11-27
Add `groupConcat` second argument overload

yariks5s:
### Changelog category (leave one):
- New Feature

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Allows to use a delimiter as a optional second argument for aggregate function `groupConcat`.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

> Information about CI checks: https://clickhouse.com/docs/en/development/continuous-integration/

#### CI Settings (Only check the boxes if you know what you are doing):
- [ ]  Allow: All Required Checks
- [ ]  Allow: Stateless tests
- [ ]  Allow: Stateful tests
- [ ]  Allow: Integration Tests
- [ ]  Allow: Performance tests
- [ ]  Allow: All Builds
- [ ]  Allow: batch 1, 2 for multi-batch jobs
- [ ]  Allow: batch 3, 4, 5, 6 for multi-batch jobs
---
- [ ]  Exclude: Style check
- [ ]  Exclude: Fast test
- [ ]  Exclude: All with ASAN
- [ ]  Exclude: All with TSAN, MSAN, UBSAN, Coverage
- [ ]  Exclude: All with aarch64, release, debug
---
- [ ]  Run only fuzzers related jobs (libFuzzer fuzzers, AST fuzzers, etc.)
- [ ]  Exclude: AST fuzzers
---
- [ ]  Do not test
- [ ]  Woolen Wolfdog
- [ ]  Upload binaries for special builds
- [ ]  Disable merge-commit
- [ ]  Disable CI cache

PR #104882, opened 2026-05-13
Preserve existing parameters in groupConcat two-argument form

yariks5s:
Reference: https://github.com/ClickHouse/ClickHouse/pull/72540#issuecomment-4436275510.
Before, a query that mixes both spellings — e.g. `groupConcat(',', 2)(x, '/')` — silently dropped the `2` and returned every row instead of the requested two

Preserves the original parameters and lets the second-argument delimiter
override position 0 (the two-argument spelling is the more specific one
and already documented to "override current parameter" on the test
above this case)

### Version info
- Merged into: `26.7.1.483` (included in `26.7` and later)


scanhex12 (approved):
looks good, but there is one more test to fix

clickgapai:

### TL;DR
**Preserve existing parameters in groupConcat two-argument form**

**ClickGap verified:** reproduced it with the script below on `26.5`; tested 6 version(s) — 5 affected, 1 clean (see Affects); present since the feature was introduced in v24.8 (not a recent regression).

### Reproduction
```
-- Requires the two-argument groupConcat overload (#72540) and enable_analyzer=1.
-- The '2' is a limit (max rows to concat); '/' is the two-arg-form delimiter.
SELECT groupConcat(',', 2)(number, '/') FROM numbers(5) SETTINGS enable_analyzer = 1;
-- Buggy builds drop the limit and return 0/1/2/3/4; the fix returns 0/1.
```
[▶ Run on ClickHouse Fiddle](https://fiddle.clickhouse.com/9c393d0a-dfb9-4d08-85fe-2c8b2e5173c2)

### Bisect
- The reproducer requires `enable_analyzer`, introduced in **`v24.8`** — so this is **not pre-existing**: it was introduced no earlier than `v24.8`, and the introducing change is in the `24.8`–`master` range (most likely the change that added the feature). Every release line down to there reproduces.

### Affects

### Suggested next step

**Severity:** `P2`

---
_ClickGap · Finding: `phase_d_pr104882`_




### Triage metadata
- Primary component: `comp-query-analyzer`
- Component owners (`comp-query-analyzer`): @KochetovNicolai @novikd

clickgapai:
Hi — this looks like a backport candidate for `26.1`, `26.2`, `26.3`, `26.4`, `26.5`. I ran this PR's own regression test against each branch's binary and the bug **reproduces** there (the fix isn't present yet).

PR #104884, opened 2026-05-13
Transform JOIN hash table payload to row major

m-selmi:
Resolves [#97883](https://github.com/ClickHouse/ClickHouse/issues/97883)

Introduce an alternative layout for hash table payload in hash joins, where fixed and contiguous columns under a certain size are transformed into one row-major block of data.
Before the change:

```
    col1 Int32     col2 String    col3 UInt8     col4 Float64
   ┌──────────┐   ┌──────────┐   ┌──────────┐   ┌────────────┐
   │   ....   │   │   ....   │   │   ....   │   │    ....    │
   │   ....   │   │   ....   │   │   ....   │   │    ....    │
   └──────────┘   └──────────┘   └──────────┘   └────────────┘
         └──────────────┴──────────────┴───────────────┘
                row N → col1[N], col2[N], col3[N], col4[N]
```
After the change:
```
   ROW STORE (col1 + col3 + col4, contiguous per row)     col2 String (columnar)
   ┌──────────┬──────────┬────────────┐                   ┌──────────┐
   │   col1   │   col3   │    col4    │ ← row N (record)  │   ....   │ ← col2[N]
   ├──────────┼──────────┼────────────┤                   ├──────────┤
   │   col1   │   col3   │    col4    │                   │   ....   │
   └──────────┴──────────┴────────────┘                   └──────────┘
```

### Rational:

The row-major format allows the output construction to be a tight loop over each column, basically one pointer load and data copy:
```
for (size_t i = 0; i < n; ++i)
{
    const char * row_store_ptr = row_store_ptrs[i];
    col->insertData(row_store_ptr + value_offset, value_size);
}
```
In comparison the columnar version has to do more work: load columns[j], row_numbers[j], replicated_columns access, insertFrom ... 
For the following query `SELECT * FROM rs_probe_200m l JOIN rs_right_10m r ON (l.k % toUInt64(10000000 / 0.9)) = r.k FORMAT Null` from `04054_hash_join_with_row_store.sql` the perf counters for `LazyOutput::buildOutput` in `AddedColumns` look as follows:
   ```
┌─variant────────────────────────────┬─query_id─────────────────────────────┬─runtime_ms─┬───────instr─┬──────cycles─┬──────cmiss─┬──brmiss─┬───ipc─┬─cmiss_per_kinstr─┬─brmiss_per_kinstr─┐
 │ row-major-enabled-multi-threaded   │ 9d2d6886-2ad7-430f-af2c-6c7cd32f3808 │       1621 │ 11393048228 │ 31974507197 │ 1470246364 │ 5892574 │ 0.356 │           129.05 │             0.517 │
 │ row-major-enabled-multi-threaded   │ 56b33013-8bec-4d7a-a415-2a06beb14bf6 │       1633 │ 11547943445 │ 31874206968 │ 1484117664 │ 6265932 │ 0.362 │           128.52 │             0.543 │
 │ row-major-enabled-multi-threaded   │ 65583fa7-9ce0-4426-95eb-977251d38f75 │       1708 │ 11338601601 │ 31825721703 │ 1468670383 │ 5921051 │ 0.356 │           129.53 │             0.522 │
 │ row-major-disabled-multi-threaded  │ 5ff1b154-3c3f-40c6-bae9-f6cf0feafe81 │       2548 │ 15453833738 │ 58322156979 │ 2636461094 │ 9533242 │ 0.265 │            170.6 │             0.617 │
 │ row-major-disabled-multi-threaded  │ 6e92c1d4-8432-4a79-a6cc-85e38232116d │       2580 │ 15400954730 │ 58683032344 │ 2623136775 │ 9455570 │ 0.262 │           170.32 │             0.614 │
 │ row-major-disabled-multi-threaded  │ 133f384f-c09a-4e12-9e69-85b61272a360 │       2686 │ 15419169622 │ 59548322941 │ 2647601264 │ 9746405 │ 0.259 │           171.71 │             0.632 │
 │ row-major-enabled-single-threaded  │ 20039e91-b5dc-4b4a-8e22-f4451f6067be │      30610 │  9046504977 │ 16964398066 │  995568814 │ 3235355 │ 0.533 │           110.05 │             0.358 │
 │ row-major-disabled-single-threaded │ a3f9a822-dae6-45c1-9a4e-63e87a3592e8 │      56821 │ 15279361788 │ 27940919919 │ 1167368997 │ 4694553 │ 0.547 │             76.4 │             0.307 │
   └────────────────────────────────────┴──────────────────────────────────────┴────────────┴─────────────┴─────────────┴────────────┴─────────┴───────┴──────────────────┴───────────────────┘
```

### Changes:

- A `RowDataStore` that takes a set of columns and lays them in row major format. It exposes access to individual rows and the layout of each row (offset and size of fields in a row).
- New `IColumn` interfaces that handle building the output columns from the row store, used manly by `AddedColumns.cpp`.
- Two new settings to control the row-major transformation:
    - enable_hash_join_row_store.
    - min_rows_ratio_for_hash_join_row_store: Minimum estimated ratio of join output rows to build-side rows to enable transforming hash join payload to row-major. Used in the planner to decide when to enable the optimization.
- A runtime feedback loop that tracks the number of hash table matches from a join and feeds it to the next execution of the join instead of relying on the estimated join output:
    - The match count is only published when the probe phase fully completes. 
    - Only joins that went through the join reorder optimizer get a cache key. 
    - Plans serialized to distributed workers don't collect or use the feedback.

### Limitations:

1. Only supports fixed and contiguous types and nullable version of them.
Supporting non fixed rows would make the row-major layout non-uniform between rows and output column construction loop would not be as tight anymore. Also transforming unbounded unfixed types to row major could be very heavy.
2. ColumnReplicated might get materialized. The decision is made based on the column state the join sees in the first block. If it's ColumnReplicated the column is excluded from the row store, if it's not the column is included and if it's replicated in a future block it get's materialized.
3. If spilling occurs, the row stored data is split back into columns for now. In a follow up we can make the row store a column and spill it as is. 

### Build overhead:

| Type | Rows | Columns | avg per worker thread (ms) |
|---|---|---|---|
| Int64 | 2M | 8 | 28.2 |
| Int64 | 1M | 16 | 31.9 |
| FixedString(16) | 1M | 8 | 24.6 |
| FixedString(16) | 2M | 4 | 21.5 |

##### Absolute runtime diff:
image
image

##### Relative speedup:
image
image

##### TPC-H and TPC-DS:

At sf100, they were unaffected by the row store. There are multiple factors: build sides quite small, runtime filters pruning probe sides and another important factor: sort order. The tables are usually sorted/clustered on the join keys so cache locality is already good.

##### Effect of sort order and no runtime filters (left join):
image
image

#### from `04054_hash_join_with_row_store.sql`:

image
image
image

- Performance Improvement

Introduce an alternative layout for hash table payload in hash joins, where qualifying columns are transformed into one row-major block of data in order to speed up join output reconstruction. Controlled by two new settings: `enable_hash_join_row_store` (default true) and `min_rows_ratio_for_hash_join_row_store` (default 5, the minimum estimated ratio of join output rows to build-side rows to enable the transformation).

---
Workflow [[PR](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104884&sha=latest&name_0=PR)]
Sync PR [[sync-upstream/pr/104884](https://github.com/search?q=head%3Async-upstream%2Fpr%2F104884+org%3AClickHouse+type%3Apr&type=pullrequests)]

### CODE REVIEW

**Comment by alexey-milovidov** [comment]

**Comment by harikrishnan94** [comment]
@m-selmi 
Can we combine null flags together (as bit masks) at the start of each row?
This will reduce the padding requirement drastically.
ie.
From this
`[null_flag(1B) | uint32_val(4B) | float64_val(8B) | null_flag(1B) | uint64_val(8B)]`
to
`[nulls_bitmap(1B) | uint32_val(4B) | float64_val(8B) | uint64_val(8B)]`
ofcourse aligned.

**Comment by m-selmi** [comment]
> @m-selmi Can we combine null flags together (as bit masks) at the start of each row? This will reduce the padding requirement drastically. ie. From this `[null_flag(1B) | uint32_val(4B) | float64_val(8B) | null_flag(1B) | uint64_val(8B)]` to `[nulls_bitmap(1B) | uint32_val(4B) | float64_val(8B) | uint64_val(8B)]` ofcourse aligned.

@harikrishnan94 Thanks for taking a look. Right now it at least keeps the same memory footprint as separate columns, but that's a very good point, I can try it out.

File: src/Interpreters/ConcurrentHashJoin.cpp [hunk 1]
```diff
@@ -905,6 +945,75 @@ void ConcurrentHashJoin::onBuildPhaseFinish()
...
+        columns_info->transferToRowStore(access_indexes);
+    return true;
+}
+
+void ConcurrentHashJoin::onPostBuildPhaseFinish()
+{
+    for (auto & hash_join : hash_joins)
+    {
+        auto & data = *getData(hash_join);
+        if (data.row_store_state != HashJoin::RowStoreState::Enabled)
```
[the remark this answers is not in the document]
m-selmi: in case of two level map the `concurrent_hash_join` only dispatches blocks to the first hash_join that owns the data in `joinBlock`

tests/queries/0_stateless/04054_hash_join_with_row_store.sql
```diff
@@ -0,0 +1,101 @@
...
+    WHERE q.log_comment IN ('rs_04054_enabled', 'rs_04054_max_bytes', 'rs_04054_min_cols_zero')
+      AND q.current_database = currentDatabase()
+      AND q.type = 'QueryFinish'
+      AND p.name = 'FinalizingRightJoinSide'
+    GROUP BY q.log_comment
+);
+
+SELECT '--- Row store columns filtering ---';
+SELECT * FROM left l INNER JOIN right r ON l.k = r.k ORDER BY ALL
+SETTINGS max_bytes_for_hash_join_row_store = 55, log_comment = 'rs_04054_filtered';
```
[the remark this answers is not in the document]
m-selmi: it's for k and v2 not v1.

src/Interpreters/ConcurrentHashJoin.cpp
```diff
@@ -911,6 +947,141 @@ void ConcurrentHashJoin::onBuildPhaseFinish()
...
+                payload_rows += columns_info->rows();
+    }
+
+    /// The `max_row_store_bytes` budget is scaled by the number of hash join slots.
+    const auto & block = first_hash_join->savedBlockSample();
+    auto access_indexes_opt = HashJoin::computeColumnAccessIndexes(
+        block,
+        column_replicated_flags,
+        payload_rows,
+        table_join->maxBytesForHashJoinRowStore() * slots,
```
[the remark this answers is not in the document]
m-selmi: The cap per hash join instance is an approximation, the row store building is parallelized on block level even in case of skew were data lands on one internal hash join.

src/Interpreters/RowDataStore.cpp
```diff
@@ -0,0 +1,180 @@
...
+            layout.size());
+
+    if (length == 0)
+        return;
+
+    size_t data_size = chars.size();
+    chars.resize(data_size + length * row_length);
+    char * dst = chars.data() + data_size;
+
+    for (size_t i = 0; i < layout.size(); ++i)
```
harikrishnan94, line 58: Can we templatize the loop on field size, so the we can turn the memcpy into direct stores?

harikrishnan94 (commented):
Could we compare filling and gathering from a smaller batch of rows to a block row count for large right-side row widths to see if that improves both?

src/Interpreters/RowDataStore.cpp
m-selmi: With this the following cases improved (single threaded):
| Type | Rows | Columns | avg FinalizingRightJoinSide (ms) |
|---|---|---|---|
| Int64 | 2M | 8 | 34.4 -> 28.2  |
| Int64 | 1M | 16 | 37.6 -> 31.9 |

m-selmi:
> Could we compare filling and gathering from a smaller batch of rows to a block row count for large right-side row widths to see if that improves both?

@harikrishnan94 Great point, here are some results for this:
- 10M build rows, 20M probe rows
- build payload 640 bytes per row so a block of 65536 rows has 40 MiB
- cache sizes: L1 32KiB / L2 1MiB / L3 32MiB

| batch size | FinalizingRightJoinSide median (ms) | JoiningTransform median (ms) | Query median (ms) |
|---|---|---|---|
| 4 KB | 164.58 | 399.71 | 1272 |
| 8 KB | 135.68 | 337.88 | 1147 |
| 16 KB | 121.92 | 317.02 | 1159 |
| 32 KB | 142.82 | 329.69 | 1166 |
| 64 KB | 119.04 | 322.40 | 1105 |
| 512 KB | 106.44 | 286.51 | 988 |
| 1 MB | 150.82 | 353.54 | 1201 |
| 4 MB | 199.06 | 685.78 | 1598 |
| 8 MB | 173.32 | 722.34 | 1625 |
| 16 MB | 124.79 | 769.57 | 1495 |
| 32 MB | 145.74 | 838.73 | 1643 |
| no limit | 154.32 | 808.56 | 1701 |

Seems like the output fill path has the best results while the batch still fits in L2 cache. The row store gather path is not affected as much because CH reduces the build input block size to 8192 to deal with the wide rows. When forcing the build input blocks to remain large manually there is a similar pattern where there is degradation once we leave L2 cache:

| batch size | FinalizingRightJoinSide median (ms) | JoiningTransform median (ms) | Query median (ms) |
|---|---|---|---|
| 4 KB | 108.61 | 362.65 | 948 |
| 8 KB | 133.36 | 311.90 | 897 |
| 16 KB | 143.74 | 274.48 | 872 |
| 32 KB | 125.20 | 287.66 | 883 |
| 64 KB | 119.24 | 278.45 | 839 |
| 512 KB | 108.76 | 296.01 | 982 |
| 1 MB | 96.73 | 289.81 | 892 |
| 4 MB | 175.38 | 557.08 | 1210 |
| 8 MB | 194.40 | 620.75 | 1287 |
| 16 MB | 211.12 | 705.10 | 1460 |
| 32 MB | 202.65 | 759.22 | 1550 |
| no limit | 245.07 | 767.90 | 1574 |

Based on this I added batching to `L2 cache size / 4` to leave some headroom.

src/Interpreters/RowDataStore.cpp
harikrishnan94: Great! Thanks for this @m-selmi .

harikrishnan94:
This is great - thanks for putting the intuition to work. The clean L2-residency boundary (and the falloff once the batch spills past it) is exactly the signal we were hoping for. Nice call landing on L2/4 for headroom.

alexey-milovidov:
The branch now conflicts with `master`: #107189 (merged 2026-07-03) reworked row references in `HashJoin` — `ColumnsInfo` became `StoredBlock`, row refs are now encoded 64-bit words (`RowRef::encode`, `RowRefList` with 48-bit pointer + count packing), and the lazy emit path resolves per-block column base pointers once via `StoredColumnsIndex::resolveEmitColumns`. The row-store hooks in this PR (`ColumnAccessIndex`, `ColumnsInfo::transferToRowStore`, `appendFromBlock(const RowRefList *)`) sit exactly on top of the code that was replaced, so this is a semantic rebase rather than a mechanical conflict resolution — for example, `AddedColumns::appendFromBlock` now takes an encoded `UInt64` ref word, and the new emit table already caches per-block base pointers, which overlaps with part of the motivation here. The row store should still help (one contiguous row-major record instead of per-column loads), but the gains are worth re-measuring against the new baseline after the rebase.

CI triage for the current head (5500d5b):
- `AST fuzzer (amd_debug, targeted, old_compatibility)`: `Logical error: Block structure mismatch in A stream` (STID `0993-2cc2`) — known master-wide issue, unrelated. Related: https://github.com/ClickHouse/ClickHouse/issues/108210
- `AST fuzzer (amd_debug)`: same class, STID `0993-27f0` — known master-wide issue, unrelated. Related: https://github.com/ClickHouse/ClickHouse/issues/108142
- `Stress test (arm_release)` hung check: the only hung query is a fuzzed `INSERT` stuck in `MergeTreeIndexAggregatorSet::update` (set skip index ignoring cancellation) — the class tracked in the hung-check umbrella, unrelated to this PR. Related: https://github.com/ClickHouse/ClickHouse/issues/107941

On the automated review's blocking item (current performance evidence): the `Performance Comparison` shards that were still pending when it last ran have since completed for 5500d5b — 4044 queries, 27 faster / 6 slower / 25 unstable, no significant changes — and the updated cloud benchmark report above also concluded `no_change` (the earlier `tpch_adapted_1_official` Q8 concern did not reproduce). So the remaining work is the rebase onto #107189, then a fresh CI round on the new head.

m-selmi:
> The branch now conflicts with `master`: #107189 (merged 2026-07-03) reworked row references in `HashJoin` — `ColumnsInfo` became `StoredBlock`, row refs are now encoded 64-bit words (`RowRef::encode`, `RowRefList` with 48-bit pointer + count packing), and the lazy emit path resolves per-block column base pointers once via `StoredColumnsIndex::resolveEmitColumns`. The row-store hooks in this PR (`ColumnAccessIndex`, `ColumnsInfo::transferToRowStore`, `appendFromBlock(const RowRefList *)`) sit exactly on top of the code that was replaced, so this is a semantic rebase rather than a mechanical conflict resolution — for example, `AddedColumns::appendFromBlock` now takes an encoded `UInt64` ref word, and the new emit table already caches per-block base pointers, which overlaps with part of the motivation here. The row store should still help (one contiguous row-major record instead of per-column loads), but the gains are worth re-measuring against the new baseline after the rebase.
> 
> CI triage for the current head ([5500d5b](https://github.com/ClickHouse/ClickHouse/commit/5500d5b326fb96a540f26c5f0d896c8ef72ef5c4)):
> 
> * `AST fuzzer (amd_debug, targeted, old_compatibility)`: `Logical error: Block structure mismatch in A stream` (STID `0993-2cc2`) — known master-wide issue, unrelated. Related: [Logical error: Block structure mismatch in A stream: different columns: (STID: 0993-2cc2) #108210](https://github.com/ClickHouse/ClickHouse/issues/108210)
> * `AST fuzzer (amd_debug)`: same class, STID `0993-27f0` — known master-wide issue, unrelated. Related: [Logical error: Block structure mismatch in A stream: different columns: (STID: 0993-27f0) #108142](https://github.com/ClickHouse/ClickHouse/issues/108142)
> * `Stress test (arm_release)` hung check: the only hung query is a fuzzed `INSERT` stuck in `MergeTreeIndexAggregatorSet::update` (set skip index ignoring cancellation) — the class tracked in the hung-check umbrella, unrelated to this PR. Related: [Hung check failed, possible deadlock found #107941](https://github.com/ClickHouse/ClickHouse/issues/107941)
> 
> On the automated review's blocking item (current performance evidence): the `Performance Comparison` shards that were still pending when it last ran have since completed for [5500d5b](https://github.com/ClickHouse/ClickHouse/commit/5500d5b326fb96a540f26c5f0d896c8ef72ef5c4) — 4044 queries, 27 faster / 6 slower / 25 unstable, no significant changes — and the updated cloud benchmark report above also concluded `no_change` (the earlier `tpch_adapted_1_official` Q8 concern did not reproduce). So the remaining work is the rebase onto #107189, then a fresh CI round on the new head.

FYI I resolved the conflicts locally, and will remeasure the performance and update the description.
But I'm also changing the logic to make the decision to enable/disable the row store in the optimizer instead.

src/Planner/PlannerJoins.cpp
```diff
@@ -1170,6 +1171,12 @@ static std::shared_ptr<IJoin> tryCreateJoin(
     SharedHeader & right_table_expression_header,
     const JoinAlgorithmParams & params)
 {
+    /// Only enable hash table payload row-major transformation if the join produces more rows than row_store_ratio * build_size.
+    const double row_store_ratio = params.min_rows_ratio_for_hash_join_row_store;
+    const bool enable_row_store = row_store_ratio == 0.0 || (params.rhs_size_estimation && params.result_rows_estimation
+            && static_cast<double>(*params.result_rows_estimation) >= static_cast<double>(*params.rhs_size_estimation) * row_store_ratio);
+    table_join->setRowStoreEnabled(enable_row_store);
+
     if (table_join->kind() == JoinKind::Paste)
```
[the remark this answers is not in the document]
m-selmi: The point is to keep the row store disabled in the legacy path where we don't have access to statistics and only use it with the new analyzer. The one case we can enable it if the ratio setting is 0 and it should be always on

src/Processors/QueryPlan/JoinStepLogical.cpp
```diff
@@ -1458,6 +1458,9 @@ void JoinStepLogical::buildPhysicalJoin(
         if (join_step->right_rows_estimation)
             join_step->join_algorithm_params->rhs_size_estimation = join_step->right_rows_estimation;
 
+        if (join_step->result_rows_estimation)
```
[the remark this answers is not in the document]
m-selmi: it should be fine for the first iteration to not enable the row store on worker in distributed plans, especially since we can't correct with runtime statistics in that case.

src/Processors/QueryPlan/JoinStepLogical.cpp
```diff
@@ -1444,20 +1444,25 @@ void JoinStepLogical::buildPhysicalJoin(
...
     UInt64 hash_table_key_hash = optimization_settings.collect_hash_table_stats_during_joins ? join_step->getRightHashTableCacheKey() : 0;
+    UInt64 join_output_key_hash = optimization_settings.collect_hash_table_stats_during_joins ? join_step->getJoinOutputCacheKey() : 0;
 
     if (!join_step->join_algorithm_params)
     {
         join_step->join_algorithm_params = std::make_unique<JoinAlgorithmParams>(
             join_step->join_settings,
             optimization_settings.max_threads,
             hash_table_key_hash,
+            join_output_key_hash,
```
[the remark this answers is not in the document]
m-selmi: This is consistent with how we only have optimizer estimates in that case too. Notice that to do this we need to adjust when hash table build runtime statistics are collected to ensure we have `rhs_size_estimation`. Could be improved in a follow up. For now it should be enough to scope this feature to when the reordered runs (or when the ratio forces the feature on).

alexey-milovidov:
I resolved the merge conflicts with the latest master and pushed the result (maintainer edits are enabled on this fork):

- Merged `master` into the branch. The conflicts were mechanical this time, since the branch already contains the #107189 row-ref architecture: master's `any_take_last_row` threading through `JoinSwitcher`/`SpillingHashJoin`/`ConcurrentHashJoin` was combined with this PR's `HashJoinStatsCollectingParams`/`is_concurrent_hash_join_` renames, and in the non-joined-rows emit path master's per-`block_no` stream distribution (`isBlockInRange`) was combined with this PR's row-store-aware `blockRows` accessor.
- Adapted `RowDataStore::create` to the rename of `convertToFullIfNeeded` on master: it now uses `convertToFullIfWrapped` chained with `convertToFullColumnIfLowCardinality`, matching how the emit path in `HashJoinMethodsImpl.h` normalizes columns.

Verified locally: the release build succeeds, `04054_hash_join_with_row_store` passes, and a sweep of 58 RIGHT/FULL/ANY JOIN stateless tests passes (including `03402_concurrent_right_full_join` and `03721_concurrent_right_join_flags_per_row_bug`, which exercise the merged `flag_per_row` path; the only failures were environmental cluster tests not runnable on a single local server).

The two open review threads on `JoinStepLogical.cpp` (stamping `join_output_key_hash` for non-rebuilt joins, and preserving the row-store decision across plan serialization) remain design decisions for @m-selmi.

src/Planner/PlannerJoins.cpp
```diff
@@ -1181,6 +1183,29 @@ static std::shared_ptr<IJoin> tryCreateJoin(
...
+            params.collect_hash_table_stats_during_joins,
+            params.max_entries_for_hash_table_stats,
+            params.max_size_to_preallocate_for_joins}};
+
+    /// Only enable hash table payload row-major transformation if the join produces more rows than row_store_ratio * build_size.
+    /// Prefer the number of hash table matches observed on a previous run over the optimizer's output estimate.
+    const double row_store_ratio = params.min_rows_ratio_for_hash_join_row_store;
+    const auto hash_join_match_hint = getHashJoinMatchHint(stats_collecting_params.match);
+    const std::optional<UInt64> row_store_output
+        = hash_join_match_hint ? std::optional<UInt64>(hash_join_match_hint->matches) : params.result_rows_estimation;
```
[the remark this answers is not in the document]
m-selmi: The undercount only makes the decision more conservative, and non-joined rows are read just once each anyway, so they'd never justify the row store on their own. Planner estimates are often way off or missing entirely, so falling back to them is worse than a slightly low measurement.

src/Interpreters/GraceHashJoin.cpp
```diff
@@ -732,7 +732,7 @@ IBlocksStreamPtr GraceHashJoin::getDelayedBlocks()
 
 GraceHashJoin::InMemoryJoinPtr GraceHashJoin::makeInMemoryJoin(const String & bucket_id, size_t reserve_num)
 {
-    return std::make_unique<HashJoin>(table_join, right_sample_block, any_take_last_row, reserve_num, bucket_id);
+    return std::make_unique<HashJoin>(table_join, right_sample_block, any_take_last_row, reserve_num, bucket_id, /*is_concurrent_hash_join*/ false);
```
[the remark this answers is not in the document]
m-selmi: Grace Join is excluded from participating in hash-table statistics, master doesn't publish the build-size hint from it either

src/Interpreters/HashJoin/HashJoin.h
```diff
@@ -112,8 +113,8 @@ class HashJoin : public IJoin
         bool any_take_last_row_ = false,
         size_t reserve_num_ = 0,
         const String & instance_id_ = "",
-        bool use_two_level_maps_ = false,
-        const StatsCollectingParams & stats_collecting_params_ = {});
+        bool is_concurrent_hash_join_ = false,
+        const HashJoinStatsCollectingParams & stats_collecting_params_ = {});
```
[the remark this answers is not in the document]
m-selmi: The clone paths are unchanged from master, and the row-store decision doesn't actually regress there, the clones only stop collecting stats, and for the per-shard path that's intentional: each clone builds a hash table over a fraction of the right side, so if we threaded the params through, N shard clones would publish per-shard counts under a key that means the whole join and poison the cache. Whole-join clones (ConcurrentHashJoin::clone) already pass the params.

src/Columns/IColumn.cpp
```diff
@@ -774,6 +827,94 @@ void IColumnHelper<Derived, Parent>::fillFromBlocksAndRowNumbers(size_t source_c
...
+                if constexpr (with_null_map)
+                    null_dst[i] = 1;
+                type->insertDefaultInto(*col);
+                continue;
+            }
+        }
+        else
+        {
+            chassert(row_store_ptr != nullptr);
+        }
```
nickitat:
```suggestion
chassert(may_have_nulls || row_store_ptr != nullptr);
if constexpr (may_have_nulls)
{
    if (!row_store_ptr)
    {
        if constexpr (with_null_map)
            null_dst[i] = 1;
        type->insertDefaultInto(*col);
        continue;
    }
}
```

src/Core/Settings.cpp
```diff
@@ -8360,6 +8360,12 @@ Enable converting the hash table to a flat array for joins when the key is a sin
 )", 0) \
     DECLARE(UInt64, query_plan_min_columns_for_join_lazy_indexing, 3, R"(
 Control the minimum number of payload columns from the left side required for enabling lazy indexing optimization in JOIN. 0 means the optimization is disabled.
+)", 0) \
+    DECLARE(UInt64, min_columns_for_hash_join_row_store, 3, R"(
```
nickitat: How did we chose both of these thresholds? E.g. intuitively `2x` looks too small to me.

src/Columns/IColumn.h
```diff
@@ -774,13 +776,32 @@ class IColumn : public COW<IColumn>
         const IColumn * const * block_columns,
         const ColumnReplicated * const * block_replicated);
 
+    /// Fills column values from row-store referenced by a RowRefList
+    virtual void fillFromRowRefsWithRowStore(const DataTypePtr & type, size_t source_field_offset, size_t source_field_size, const UInt64 * row_refs_begin, const UInt64 * row_refs_end, const RowDataStore * const * block_row_stores);
+
+    /// Nullable variant of `fillFromRowRefsWithRowStore`: the null byte is written into `null_map` and the value bytes are inserted into the nested column
+    virtual void fillFromRowRefsWithRowStoreAndNullMap(const DataTypePtr & type, size_t source_field_offset, size_t source_field_size, const UInt64 * row_refs_begin, const UInt64 * row_refs_end, const RowDataStore * const * block_row_stores, PaddedPODArray<UInt8> & null_map);
```
nickitat: It looks a little strange that we add nullable-specific methods to the common interface. Is it necessary?

src/Columns/IColumn.h
```diff
@@ -774,13 +776,32 @@ class IColumn : public COW<IColumn>
...
 
     /// Same as above but assumes every entry in the list is non-null
     virtual void fillFromBlocksAndRowNumbers(size_t source_column_index_in_block, const ColumnsWithRowNumbers & columns_with_row_numbers);
 
+    /// Fills column values from pre-resolved row-store pointers
+    /// A nullptr in the list is interpreted as a default value
+    virtual void fillFromRowStorePtrs(const DataTypePtr & type, const PaddedPODArray<const char *> & row_store_ptrs, size_t field_offset, size_t field_size, size_t begin, size_t count);
+
+    /// Same as above but assumes every entry in the list is non-null
+    virtual void fillFromRowStorePtrs(const PaddedPODArray<const char *> & row_store_ptrs, size_t field_offset, size_t field_size, size_t begin, size_t count);
```
nickitat: This one also looks too specific for the generic interface.

src/Planner/PlannerJoins.cpp
```diff
@@ -1181,6 +1183,29 @@ static std::shared_ptr<IJoin> tryCreateJoin(
...
+    const bool enable_row_store = row_store_ratio == 0.0
```
nickitat: Why don't we use the cached number of source rows from the right side instead of `rhs_size_estimation`?

src/Interpreters/HashJoin/HashJoin.h
```diff
@@ -586,8 +607,12 @@ class HashJoin : public IJoin
     /// Track if shared runtime filters were already published to keep publication one-shot.
     bool shared_runtime_filters_publish_attempted = false;
 
-    const StatsCollectingParams stats_collecting_params;
+    const HashJoinStatsCollectingParams stats_collecting_params;
     bool build_phase_finished = false;
+    bool probe_phase_finished = false;
+
+    /// Rows emitted from hash-table matches across all probe threads (excludes default/miss rows).
+    mutable std::atomic<size_t> hash_table_matches{0};
```
nickitat: Usually it is possible to have a local counter that can be once pushed to a global accumulator once all the job is done. Could we do it here as well?

src/Interpreters/ConcurrentHashJoin.cpp
```diff
@@ -440,7 +472,10 @@ JoinResultPtr ConcurrentHashJoin::joinBlock(Block block)
     if (hash_joins[0]->data->twoLevelMapIsUsed())
         dispatched_blocks.emplace_back(std::move(block));
     else
-        dispatched_blocks = dispatchBlock(table_join->getOnlyClause().key_names_left, std::move(block));
+    {
+        bool use_zero_copy = useZeroCopyApproach(block);
```
nickitat: This decision shouldn't have changed since we calculated it on l.323.

src/Interpreters/ConcurrentHashJoin.cpp
```diff
@@ -293,7 +312,20 @@ bool ConcurrentHashJoin::addBlockToJoin(const Block & right_block_, bool check_l
     /// (inside different `hash_join`-s) because the block will be shared.
     Block right_block = hash_joins[0]->data->materializeColumnsFromRightBlock(right_block_);
 
-    auto dispatched_blocks = dispatchBlock(table_join->getOnlyClause().key_names_right, std::move(right_block));
+    /// Initialize the row store layout based on the first block.
+    std::call_once(row_store_init_flag, [&]
```
nickitat, line 316: Could we init in the same way we do preallocation for hash tables (not for all at the same time blocking all threads)?

src/Interpreters/RowRefs.cpp
```diff
@@ -240,19 +242,29 @@ void StoredBlock::rebuildReplicatedColumns()
...
+    if (hasRowStore())
+        allocated_bytes = row_store->allocatedBytes();
 
-    size_t res = 0;
     for (const auto & column : columns)
-        res += column->allocatedBytes();
-    return res * selector.size() / rows;
+        allocated_bytes += column->allocatedBytes();
+
+    return allocated_bytes * selector.size() / row_nums;
```
nickitat, line 267: Does this logic equally applies to row store? For normal columns we keep the original columns, but only `selector.size()` rows do actually belong to the given block.

m-selmi: yes there is one row store per block and it's shared between slots via selector.

src/Interpreters/ConcurrentHashJoin.cpp
m-selmi: The init is pretty cheap, it's just computing the layout of the store (I improved it a bit now to compute the layout once not once per slot). It's done here because it's based on columns carried in the first block.
The expensive part is `createRowStoreForBlock` below and this has to be done here because I share the same store for all slots (accessed by the selector).

src/Interpreters/ConcurrentHashJoin.cpp
m-selmi: you mean use_zero_copy? but these 2 occurrences are once per build side block and once per probe side block.

src/Interpreters/HashJoin/HashJoin.h
m-selmi: Yes, there is a local counter per block [here](https://github.com/m-selmi/ClickHouse/blob/57cd4294f289f0d56dcf83e36e2cc95e9fbf5d2d/src/Interpreters/HashJoin/AddedColumns.h#L65) and then this is accumulated in the atomic of each hash join instance once per block. Because multiple probers could be touching the same hash join instance.

src/Planner/PlannerJoins.cpp
m-selmi: it's because at this point `rhs_size_estimation` was already replaced with `source_rows` [here](https://github.com/m-selmi/ClickHouse/blob/57cd4294f289f0d56dcf83e36e2cc95e9fbf5d2d/src/Processors/QueryPlan/JoinStepLogical.cpp#L1807). The same is not done for `result_rows_estimation` because `hash_join_match_hint->matches` does not always match the result size, it counts how many rows will be reconstructed from the hash table payload.

src/Columns/IColumn.h
m-selmi: I removed it and tagged the passed data with `has_defaults` instead.

src/Columns/IColumn.h
m-selmi: I added this to be able to de-virtualize the insert into the nested column and fill the null map in one pass of the data. One alternative would be to let ColumnNullable fill the null map and call this for the nested column but costs one extra pass over the data.

src/Core/Settings.cpp
m-selmi: My intuition was that there are substantially more rows re-constructed from the row store than built into it.
I did some more detailed testing now:
(the % of matched rows is how many build rows were used in the output, but the size of the output is the ratio below 2x, 4x, ...)
## In terms of runtime diff
image
image

## In terms of relative speedup
image
image


- seems like for small build sizes the effect of row store is negligible.
- for large build sizes it seems to be always a win if we produce at least 2x number of rows in the hash table.

So I removed the number of columns setting and just enable if there is more than 1 column and replaced with an enable flag.
But for the ratio since we rely on estimates when there are no runtime stats we could think about increasing this to 10 to have some buffer, wdt?

src/Interpreters/RowDataStore.cpp
```diff
@@ -0,0 +1,266 @@
...
+void gatherField(char * dst, const char * src, size_t row_length, size_t offset, size_t length)
+{
+    for (size_t row = 0; row < length; ++row)
+        memcpy(dst + row * row_length + offset, src + row * field_size, field_size);
+}
+
+void gatherField(char * dst, const char * src, size_t row_length, size_t offset, size_t field_size, size_t length)
+{
+    for (size_t row = 0; row < length; ++row)
+        memcpy(dst + row * row_length + offset, src + row * field_size, field_size);
```
nickitat, line 42: Worth considering to always have size as a template argument - then the whole `memcpy(...)` should be compiled into inlined asm. Also worth adding `__restrict` to the arguments.

src/Interpreters/RowDataStore.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+    /// Columns are materialized to make sure all blocks have
+    /// the same split of columnar and row store columns.
+    Columns materialized_columns;
+    materialized_columns.reserve(columns.size());
+    for (const auto & col : columns)
+        materialized_columns.push_back(col->convertToFullIfWrapped()->convertToFullColumnIfLowCardinality());
+
+    RowLayout layout = computeLayout(materialized_columns);
+    auto row_store = std::shared_ptr<RowDataStore>(new RowDataStore(std::move(layout)));
```
nickitat: `make_shared_ptr`?

src/Interpreters/RowDataStore.cpp
```diff
@@ -0,0 +1,266 @@
...
+            ErrorCodes::SIZES_OF_COLUMNS_DOESNT_MATCH,
+            "Number of appended columns ({}) does not match the number of columns in the layout ({}).",
+            columns.size(),
+            layout.size());
+
+    if (length == 0)
+        return;
+
+    size_t data_size = chars.size();
+    chars.resize(data_size + length * row_length);
```
nickitat: Perhaps could be `resize_exact`.

src/Interpreters/RowDataStore.cpp
```diff
@@ -0,0 +1,266 @@
...
+    ProfileEventTimeIncrement<Microseconds> watch(ProfileEvents::JoinBuildRowStoreMicroseconds);
+
+    /// Columns are materialized to make sure all blocks have
+    /// the same split of columnar and row store columns.
+    Columns materialized_columns;
+    materialized_columns.reserve(columns.size());
+    for (const auto & col : columns)
+        materialized_columns.push_back(col->convertToFullIfWrapped()->convertToFullColumnIfLowCardinality());
+
+    RowLayout layout = computeLayout(materialized_columns);
```
nickitat: Should we re-calculate it each time actually?

src/Interpreters/RowDataStore.cpp
```diff
@@ -0,0 +1,266 @@
...
+        doGatherRows(layout, row_length, columns, start + batch_start, remaining_batch_size, dst + batch_start * row_length);
+    }
+}
+
+MutableColumns RowDataStore::scatterRows(size_t start, size_t length) const
+{
+    RowStorePointers row_store_ptrs;
+    row_store_ptrs.ptrs.reserve(length);
+    for (size_t row = 0; row < length; ++row)
+        row_store_ptrs.ptrs.push_back(getRowAt(start + row));
```
nickitat: This is not super optimal, but idk how often we scatter a contiguous range.

src/Columns/IColumn.cpp
```diff
@@ -738,27 +782,89 @@ static void fillColumnFromBlocksAndRowNumbers(ColumnType * col, const DataTypePt
...
+    {
+        size_t old_size = null_map->size();
+        null_map->resize(old_size + count);
+        null_dst = null_map->data() + old_size;
+    }
+
+    col->reserve(col->size() + count);
+    for (size_t i = 0; i < count; ++i)
+    {
+        const char * row_store_ptr = row_store_ptrs[begin + i];
```
nickitat: Might make sense to prefetch future addresses. But again we don't have to push all possible complexity into one PR. We can profile and optimize row store later.

src/Core/Settings.cpp
nickitat: Probably, 10 would be a safe choice, but it might be too restrictive as well. I'd take maybe tpc-{h,ds} and moved perf tests and for all of them plot the change in perf vs the ratio. And try to find a sweet spot.

nickitat: BTW, it seems like in the `In terms of relative speedup` section you've posted the same picture twice.

src/Columns/IColumn.h
nickitat: My logic was very simple: if you at some place know that you should call `col-><method>Nullable` then you can also do `assert_cast<Nullable...>(col)-><method>`. But I may miss something important because I don't know the code.

src/Interpreters/ConcurrentHashJoin.cpp
nickitat: Amount of blocks might be quite high

src/Interpreters/HashJoin/HashJoin.h
nickitat: The question is, could we keep counters local to each "prober" and don't do any atomic operations per each block?

src/Core/Settings.cpp
m-selmi: Ah sorry, should be fixed now.
I run tpc-h and tpc-ds at sf100, they were unaffected by the row store. There are multiple factors: build sides quite small, runtime filters pruning probe sides and another important factor: sort order. The tables are usually sorted/clustered on the join keys so cache locality is already good.
I did a new experiment to study sort order and a case without runtime filter (left join):
image
image
The both sorted cases are as expected flat. The other cases improve.
I bumped the min ratio to 5, it's on the conservative side and even with sorted probe it gets a speedup around 10% in the cases above.
One larger data example for completion:
image

src/Interpreters/RowDataStore.cpp
m-selmi: I switched to calculating once per join and caching.

src/Interpreters/ConcurrentHashJoin.cpp
m-selmi: I moved zero copy check to `constructor` for build side and to `initialize()` for probe side.

src/Interpreters/HashJoin/HashJoin.h
m-selmi: yes, I changed this to accumulate per prober on `JoiningTransform` level like `FinishCounter` is implmented.

src/Columns/IColumn.cpp
m-selmi: Good idea, I can experiment with this in a follow-up. I added a TODO.

src/Interpreters/RowDataStore.cpp
m-selmi: that's true, I added brief handling for ranges to improve this. 
Right now it's only used when spilling trigger and the row store is scattered back into spill-able columns.
In the future I want to make row store spill-able (probably by making it a column itself).

src/Columns/IColumn.h
m-selmi: the issue that that `col-><method>Nullable` should be called on the nested column of the column nullable, this allow us to de-virtualize the insert into the nested column. But the nested column does not know about the null flags so we either set them separately from `nested_col-><method>` or pass them in some `nested_col-><method>Nullable`, or maybe I'm missing something obvious here?

src/Core/Settings.cpp
nickitat: Pls also update it in the changelog entry.

nickitat (approved):

PR #104885, opened 2026-05-13
Docs: use title property on query/response code block pairs for cleaner formatting

Blargian:
In many places we write:

```
Query:

<CodeBlock>

Response:

<CodeBlock>
```

But imo labelling the codeblock with `title="Query"` and `title="Response"` looks much cleaner, as the codeblock gets a rendered title:

image

...

### Version info
- Merged into: `26.5.1.690`


Algunenano (approved):
Makes sense to me. Please have a look at the problems reported by the linter

PR #104886, opened 2026-05-13
Parse mutation command AST on demand

nikitamikhaylov:

Internal refactor of `MutationCommand`. No user-visible change except reduced memory usage in certain cases.

### Description

`MutationCommand` used to hold a clone of the entire `ASTAlterCommand` (in the `ast` field) plus separate clones of its `predicate`, `partition`, and `column_to_update_expression` subtrees - the same data stored two-and-a-half times. This change replaces those fields with a single serialized `ast_text` (the output of `formatWithSecretsOneLine`) and lazy accessors that re-parse it via `ParserAlterCommandList` on demand:

- `MutationCommand::ast()`
- `MutationCommand::predicate()`
- `MutationCommand::partition()`
- `MutationCommand::columnToUpdateExpression()`

All call sites that previously read these fields now invoke the accessors. Sites that built a `MutationCommand` by hand (e.g. `InterpreterDeleteQuery::execute`) now construct an `ASTAlterCommand`, format it once, and store the text - the subtree fields no longer exist.

The serialization format on ZooKeeper / on-disk mutation entries is unchanged: `ast_text` is produced by `formatWithSecretsOneLine`, which round-trips through the same `ParserAlterCommandList` that `MutationCommands::readText` already uses. The `gtest_mutation_commands` round-trip test is unaffected.

- [x] Documentation change is not required

### Version info
- Merged into: `26.6.1.167`


antonio2368:
>  No user-visible change

Reduced memory usage?

src/Storages/System/StorageSystemWasmModules.cpp
```diff
@@ -142,10 +142,11 @@ static std::optional<ModuleDeleteFilter> getModuleDeleteFilterFromAst(const Muta
         return {};
 
     const auto & command = commands.front();
-    if (command.type != MutationCommand::DELETE || command.partition || !command.predicate)
+    auto predicate = command.predicate();
+    if (command.type != MutationCommand::DELETE || command.partition() || !predicate)
```
antonio2368: You are parsing it twice back to back here.

src/Storages/MergeTree/AlterConversions.cpp
```diff
@@ -504,14 +498,14 @@ MutationCommands AlterConversions::filterMutationCommands(Names & read_columns,
...
+                new_command.predicate()->collectIdentifierNames(source_columns);
                 filtered_commands.push_back(std::move(new_command));
             }
 
             if (!new_updated_columns.empty())
             {
-                auto new_command = createCommandWithUpdatedColumns(command, std::move(new_updated_columns));
-                new_command.predicate->collectIdentifierNames(source_columns);
+                auto new_command = createCommandWithUpdatedColumns(command, new_updated_columns);
+                new_command.predicate()->collectIdentifierNames(source_columns);
```
antonio2368: Not sure if they can be called in single function call, but again parsing twice.

antonio2368 (commented):
Let's make sure that we minimize amount of parsing where possible

alexey-milovidov:

src/Storages/System/StorageSystemWasmModules.cpp
nikitamikhaylov: Hmm, where? The one in the if statement is `command.partition()` which is not a `command.predicate()`.

antonio2368: But both `partition` and `predicate` call `ast()` right?

nikitamikhaylov: You're absolutely right!

src/Storages/MutationCommands.h
```diff
@@ -48,21 +58,57 @@ struct MutationCommand
...
+    /// RAII handle for editing the AST of a `MutationCommand` in place.
+    ///
+    /// On construction the handle parses a fresh, mutable copy of `ast_text`.
+    /// Inside the handle's scope callers may freely mutate the AST through
+    /// `*handle` / `handle->...`. On destruction the (possibly modified) AST
+    /// is serialized back into the owning `MutationCommand::ast_text` in one
+    /// pass, so all edits accumulated through one handle are written back
+    /// exactly once.
+    ///
+    /// The handle is neither copyable nor movable. Construct one per scope.
```
antonio2368: I would shorten the comments

antonio2368 (commented):
Most other comments are written by the bot

antonio2368 (approved):

clickgapai:
Hi @nikitamikhaylov @antonio2368 — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: The PR description is essentially empty (just the template). While the changelog entry explains the change, the PR body should describe the motivation, approach, and any testing done.
- Suggested addition:

  > This PR changes `MutationCommand` to store the AST as serialized text (`ast_text`) instead of keeping the parsed AST in memory. The AST is now parsed on-demand via the `ast()` method. This reduces memory usage for mutation commands that are stored but not immediately executed (e.g., queued mutations in ReplicatedMergeTree). The `MutableAst` RAII class provides safe in-place modification with explicit `commit()` semantics.

PR #104888, opened 2026-05-13
Fix data part consistency checks for types with dynamic structure and detect corrupted columns_substreams.txt

Avogar:
## Linked issues
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1799

Fix data part consistency checks for types with dynamic structure and detect corrupted columns_substreams.txt. Resubmit of https://github.com/ClickHouse/ClickHouse/pull/103858 with additional changes.

### Version info
- Merged into: `26.5.1.759`
- Backported to: `26.4.3.36`, `26.3.11.31`, `26.2.19.39`


src/Storages/MergeTree/ColumnsSubstreams.cpp
```diff
@@ -231,4 +233,38 @@ void ColumnsSubstreams::validateColumns(const std::vector<String> & columns) con
...
+{
+    if (!substream.starts_with(escaped_prefix))
+        return false;
+    /// Must be exactly the prefix, or followed by '.' separator.
+    return substream.size() == escaped_prefix.size() || substream[escaped_prefix.size()] == '.';
+}
+
+String ColumnsSubstreams::findInvalidSubstreamName() const
+{
+    for (const auto & [column_name, substreams] : columns_substreams)
```
Ergus, line 255: Dummy question here. Is it possible some divergency between the content in columns_substreams.txt and the the number/order of columns?

If so, does it worth adding some check/assert for that? At least form a defensive purpose ?

I mean, `columns_substreams.txt` could potentially be truncated but with a correct prefix, so trusted then?

tests/queries/0_stateless/04234_check_table_json_wide.sql
```diff
@@ -0,0 +1,34 @@
+-- Tags: no-fasttest
```
Ergus, line 1: IIUC There is no regression test for a JSON/Dynamic Wide part where columns_substreams.txt is missing or discarded and the code has to use the new `enumerate_dynamic_streams = false` fallback. But only for `Array(UInt32)`

Ergus (commented):
Overall LGTM, but I have a couple of comments it would like to clarify.

src/Storages/MergeTree/ColumnsSubstreams.cpp
Avogar: There is a method `ColumnsSubstreams::validateColumns` that validates that `columns_substreams.txt` contains valid set of columns. And I just realized that it's not called when we load the file from the data part. Let me fix that

tests/queries/0_stateless/04234_check_table_json_wide.sql
Avogar: Good catch, yes, forgot about this test, let me add it

src/Storages/MergeTree/ColumnsSubstreams.cpp
```diff
@@ -231,4 +233,45 @@ void ColumnsSubstreams::validateColumns(const std::vector<String> & columns) con
...
+        auto nested_table_name = Nested::extractTableName(column_name);
+        auto escaped_nested_table_name = escapeForFileName(nested_table_name);
+        bool has_nested_prefix = (escaped_nested_table_name != escaped_column_name);
+
+        for (const auto & substream : substreams)
+        {
+            if (hasValidPrefix(substream, escaped_column_name))
+                continue;
+
+            if (has_nested_prefix && hasValidPrefix(substream, escaped_nested_table_name))
```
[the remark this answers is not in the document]
Avogar: It's not trivial to detect which columns are actually a part of Nested and which are not. So we can make it less strict here. In general, this is a best-effort validation for a specific corruption so I think it's ok.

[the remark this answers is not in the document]
Ergus: The AI is right here, but I get your point... If this cannot be addressed in a reasonable small change let's approve this as is.

Ergus (approved):
Just mention that the new corruption detector is still too permissive for dotted columns that share a prefix, so some bad `columns_substreams.txt` will survive validation and then be trusted as the source of truth. 

Buy the fix seems not trivial to implement, so for now let's live with it.

PR #98779, opened 2026-03-04
Fix assertion failure in view dependencies on DROP+CREATE MV

alexey-milovidov:
Fixes #98706

## Summary

- Fixed assertion `tables_from.size() == 1` in `DatabaseCatalog::updateDependencies` triggered by BuzzHouse fuzzer
- Root cause: in `updateDependencies`, the removal of old `view_dependencies` edges was inside `if (!new_view_dependencies.empty())`, so queries like `ALTER TABLE mv MODIFY QUERY SELECT 1 c0` (with no view dependencies) would leak stale edges; combined with BACKUP/RESTORE this accumulated multiple dependents, violating the assertion
- Made view dependency removal unconditional in `updateDependencies`, matching the pattern already used for `referential_dependencies` and `loading_dependencies` (unconditional removal, conditional addition)
- Removed `assert(tables_from.size() == 1)` assertions that assumed a single dependent; iterate over all dependents instead, handling stale edges gracefully
- Fixed early return bug in `addDependencies` (first overload) that silently dropped view dependencies when referential and loading dependencies were both empty

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=8f8e0d3934ac11a8887020d917ca88aa0727f3ea&name_0=MasterCI&name_1=BuzzHouse%20%28amd_debug%29

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix assertion failure in `DatabaseCatalog::updateDependencies` when dropping and recreating a materialized view with the same name; also fix view dependencies being silently lost during `RENAME TABLE` or `EXCHANGE TABLES` when referential dependencies were empty.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches `DatabaseCatalog` dependency-graph maintenance used by DDL operations (ALTER/RENAME/EXCHANGE), so mistakes could impact catalog consistency, but the change is localized and covered by a regression test.
> 
> **Overview**
[...]

PR #104907, opened 2026-05-14
Fix `CREATE OR REPLACE TABLE` losing source-side view dependencies

### LINKED ISSUE #104906: CREATE OR REPLACE breaks MV
### Company or project name

ClickHouse

### Describe what's wrong

`CREATE OR REPLACE TABLE` breaks deps

Yes

### How to reproduce

on 26.5.1.36:
```sql
  CREATE TABLE src (id Int32, key String) ENGINE = Null;
  CREATE TABLE dst (id Int32, key String) ENGINE = MergeTree ORDER BY id;
  CREATE MATERIALIZED VIEW mv TO dst AS SELECT id, key FROM src;
  INSERT INTO src VALUES (1, 'a');
  SELECT count() FROM dst;                 -- 1   OK
  CREATE OR REPLACE TABLE src (id Int32, key String) ENGINE = Null;
  INSERT INTO src VALUES (2, 'b');
  SELECT count() FROM dst;                 -- 1   BUG (expected 2)
```

### Expected behavior

2 at end, which is what 25.8 prints

_No response_

### Additional context

breaking CI for peerdb

serprex:
`InterpreterRenameQuery::executeToTables` cross-swaps the harvested `from_dependent_views` and `to_dependent_views` after the physical rename, which is correct for `EXCHANGE TABLES a, b` (both tables persist, edges follow data). `doCreateOrReplaceTable` routes both `CREATE OR REPLACE TABLE` and `REPLACE TABLE` through the same rename interpreter, but the displaced table is dropped immediately afterwards — so the cross-swap pins the source-view edges to the soon-to-be-dropped table and the new occupant of the name ends up with `system.tables.dependencies_table` empty. `InsertDependenciesBuilder` then visits no MVs on the next insert and rows are silently lost.

Tag the synthetic rename built by `doCreateOrReplaceTable` with a new `ASTRenameQuery::is_create_or_replace` flag and, when set, restore each harvested dependent-view list onto the table id it came from instead of swapping them. `EXCHANGE TABLES a, b` keeps its existing semantics.

Regression introduced in #98779

Closes https://github.com/ClickHouse/ClickHouse/issues/104906

fix REPLACE TABLE dropping MV dependencies


src/Interpreters/InterpreterRenameQuery.cpp
CheSema, line 197: Not totally sure but by the symmetry it might have required different rollback according `if (rename.is_create_or_replace)` is true or not.

CheSema:
close in favor of https://github.com/ClickHouse/ClickHouse/pull/105029

PR #104330, opened 2026-05-08
Fix jemalloc verifySetup on builds without background_thread mallctl

sayondeep:
## Motivation

On macOS (and any jemalloc build without `JEMALLOC_BACKGROUND_THREAD`), `je_mallctl` for `background_thread` and `max_background_threads` returns `ENOENT`. `Jemalloc::setup` already no-ops for those writes, but `Jemalloc::verifySetup` still read via `getValue` and compared the result to `ServerSettings`, using uninitialized memory when the read failed. That produced a false `jemalloc_enable_background_threads` mismatch and aborted debug builds (see [#102183](https://github.com/ClickHouse/ClickHouse/issues/102183)).

## Changes

- Add `tryGetValue` and use it in `verifySetup` so background-thread mallctls are only checked when `je_mallctl` succeeds.
- Zero-initialize the local in `getValue` when mallctl fails.

### Changelog category:
- Bug Fix

### Changelog entry:
Fix a fatal logical error at server startup on macOS (and similar jemalloc builds) where `Jemalloc::verifySetup` incorrectly reported a `jemalloc_enable_background_threads` mismatch because optional `background_thread` / `max_background_threads` mallctls are absent; verification is skipped when those mallctls are unavailable, and `getValue` no longer leaves the output uninitialized on failure. Closes #102183.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

Issue #104060, opened 2026-05-04
`logQueryStart` early return when `query_log` is null silently disables `system.query_metric_log` collection


clickgapai:
_Found via ClickGap automated review. Please close or comment if this is incorrect or needs adjustment._

_Retrospective finding from a historical scan of [PR #62452](https://github.com/ClickHouse/ClickHouse/pull/62452) (merged 2024-12-28). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

When `query_log` is disabled in server config but `query_metric_log` is enabled, no rows are ever inserted into `system.query_metric_log` because `query_metric_log->startQuery` is never called. Before this PR, `query_metric_log` worked independently of `query_log`.

**Root cause:** executeQuery.cpp:464-466 — the new `auto query_log = context->getQueryLog(); if (!query_log) return elem;` guard, intended only to skip the `query_log->add` calls, also bypasses the unrelated `query_metric_log->startQuery` call further down the same function.

**Why we believe this is a bug:** executeQuery.cpp:466 returns early from `logQueryStart` when `getQueryLog()` returns null. The `if (auto query_metric_log = context->getQueryMetricLog(); query_metric_log && !internal) { ... query_metric_log->startQuery(...); }` block at line 533-538 is therefore unreachable in that configuration. `QueryMetricLog::finishQuery` (QueryMetricLog.cpp:185-186) silently returns when `startQuery` was never called, so the failure is silent.

**Affected locations:**
- `src/Interpreters/executeQuery.cpp:464` — early return added by PR
- `src/Interpreters/executeQuery.cpp:533` — query_metric_log->startQuery skipped

**Impact:** Silent feature regression: `system.query_metric_log` collects zero rows for any user who has `query_log` disabled in `config.xml` (e.g. via `<query_log remove="remove"/>`) but `query_metric_log` enabled. The broken behavior is silent — no error, no warning — the table simply stays empty.

### Does it reproduce on most recent release?

Yes — confirmed on current `master` (commit `306c893dcaa`).

### How to reproduce

```sql
1. Create a clickhouse server config with `<query_log>` section ABSENT but `<query_metric_log>` section PRESENT (collect_interval_milliseconds=100). 2. Run: `SELECT sleep(2) SETTINGS query_metric_log_interval=100 FORMAT Null` 3. SYSTEM FLUSH LOGS 4. `SELECT count() FROM system.query_metric_log` — bug returns 0; correct behavior returns ~20 rows.
[...]

PR #104916, opened 2026-05-14
Fix final rows in QueryMetricLog

antonio2368:
Fixes a flaky CI failure where `03203_system_query_metric_log` could miss the final `system.query_metric_log` row at the `system.query_log` `QueryFinish` timestamp. A late periodic `QueryMetricLog::collectMetric` could advance `last_collect_time` past the finish timestamp before `QueryMetricLog::finishQuery` emitted the final row.

The fix keeps timestamp deduplication for periodic metric rows, but does not apply it to final rows. It also updates `03203_system_query_metric_log` to check that the final row exists, and adds a failpoint regression test for the race.

CI: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104330&sha=d53d6108634f5740d1ff713e32213dbec4267213&name_0=PR&name_1=Stateless%20tests%20%28amd_tsan%2C%20parallel%2C%201%2F2%29

Original PR: https://github.com/ClickHouse/ClickHouse/pull/104330

Open issue search: no matching open issue was found for `03203_system_query_metric_log`. Closest open `system.query_metric_log` issue found: https://github.com/ClickHouse/ClickHouse/issues/104060, which is a separate null-`query_log` path.

Not required.

### Documentation entry

### Version info
- Merged into: `26.5.1.655`


pamarcos:
I think this seems to redo somewhat [this](https://github.com/ClickHouse/ClickHouse/pull/103547) and undo https://github.com/ClickHouse/ClickHouse/pull/103771. However, since the CI failure happened after the latter, I'll take a closer look

src/Interpreters/QueryMetricLog.cpp
```diff
@@ -134,7 +150,8 @@ void QueryMetricLog::collectMetric(const ProcessList & process_list, String quer
     }
     global_lock.unlock();
 
-    auto elem = query_status.createLogMetricElement(query_id, *query_info, current_time);
+    auto elem = query_status.createLogMetricElement(
+        query_id, *query_info, current_time, /* schedule_next = */ true, /* is_final = */ false);
```
pamarcos: For clarify, it's ok to use `is_final` as a separate parameter. However, `schedule_next = !is_final` every single time os far. Maybe we can refactor `schedule_next` into `is_final` directly and swap the current bool value.

src/Interpreters/QueryMetricLog.cpp
```diff
@@ -164,6 +181,16 @@ void QueryMetricLog::startQuery(const String & query_id, TimePoint start_time, U
...
 void QueryMetricLog::finishQuery(const String & query_id, TimePoint finish_time, QueryStatusInfoPtr query_info)
 {
+    /// Test-only failpoint placed before `queries_mutex` so a concurrent periodic
+    /// `collectMetric` can overtake `finishQuery` and advance `info.last_collect_time`
+    /// past `finish_time`. The gate on `query_info` and the dedicated query-id prefix
+    /// ensures unrelated traffic is unaffected even when the failpoint is enabled.
+    /// The `query_info` gate filters `nullptr` finish paths. If a non-null phantom
+    /// finish-call still reaches this failpoint, `it == queries.end()` after resume
+    /// makes this function return immediately.
+    if (query_info && query_id.starts_with(query_metric_log_final_row_failpoint_query_id_prefix))
```
pamarcos: I think this is quite hacky, but I also don't know of a better way of doing it without affecting the rest. And since this affects part of the code affects every single query, we'd better be safe than sorry I guess

src/Interpreters/QueryMetricLog.cpp
pamarcos: I went ahead and implemented that in https://github.com/ClickHouse/ClickHouse/pull/104916/commits/473100433d732d94e26c91fbe87c149bec9df079

pamarcos:
I did https://github.com/ClickHouse/ClickHouse/pull/104916/commits/ef4718344a1dd78230248d343074e25c5a4295df cause we don't need monotonicity at all. I'd rather have a late event with proper data than a late event with a wrong and duplicated timestamp.

@antonio2368 https://github.com/ClickHouse/ClickHouse/pull/104916/commits/e4047f982542295c279460e91a3a30bb83ca9784 nice! I was thinking on doing something about the happy path as well :people_hugging:

antonio2368:
> I did https://github.com/ClickHouse/ClickHouse/commit/ef4718344a1dd78230248d343074e25c5a4295df cause we don't need monotonicity at all. I'd rather have a late event with proper data than a late event with a wrong and duplicated timestamp.

Makes sense.

pamarcos (approved):

PR #104921, opened 2026-05-14
Fix `FileCache` resize with partial segments

antonio2368:
Fix dynamic `FileCache` resize for partially downloaded segments. During resize, candidates can have their priority entries detached before filesystem eviction; `PARTIALLY_DOWNLOADED` segments must accept this delayed-removal state. The failed-eviction restore path also restores reserved/cache-accounted size so `FileCache::getUsedCacheSize` remains consistent after restore. No open issue was found for `queue_iterator FileCache` or `03411_dynamic_resize_filesystem_cache_3`.

https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104330&sha=d53d6108634f5740d1ff713e32213dbec4267213&name_0=PR&name_1=Stress%20test%20%28amd_debug%29

Fix filesystem cache dynamic resize with partially downloaded segments, including restore accounting after failed eviction.

### Documentation entry
- [x] Documentation is not required (no user-facing feature or configuration change).

### Version info
- Merged into: `26.5.1.752` (included in `26.5` and later)
- Backported to: `26.4.5.93`


src/Interpreters/FileCache/EvictionCandidates.cpp
```diff
@@ -328,7 +328,10 @@ void EvictionCandidates::evict()
             }
             catch (...)
             {
-                failed_candidates.total_cache_size += candidate->file_segment->getDownloadedSize();
+                /// Account by reserved (cache-accounted) bytes -- same as candidate
+                /// collection (`LRUFileCachePriority`) and same as what the restore
+                /// path re-installs as the queue entry size.
```
kssenii: The comment seems a bit over-complicated, may be just something short like "/// Sum up reserved size (== queue entry size)"?

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -2532,23 +2532,34 @@ bool FileCache::doDynamicResizeImpl(
...
-                log, "Adding back file segment after failed eviction: {}:{}, size: {}",
-                file_segment->key(), file_segment->offset(), file_segment->getDownloadedSize());
+                log, "Adding back file segment after failed eviction: {}:{}, restored size: {}, downloaded size: {}",
+                file_segment->key(), file_segment->offset(), restored_size, file_segment->getDownloadedSize());
 
             auto original_queue_type = eviction_candidates.getOriginalQueueType(candidate.get());
 
+            /// Restore the queue entry with the exact size that was removed in
+            /// `removeQueueEntries`. `FileSegmentMetadata::size` returns
+            /// `getReservedSize`, which is what `LRUFileCachePriority::collectCandidatesForEviction`
```
kssenii: Let's avoid mentioning concrete implementation of IFileCachePriority (at least in FileCache class seems good not to)? 
Or may be just remove this comment and add a small comment before line 2535 where `restore_size` is initialized - mentioning something like "reserved_size is used instead of downloaded_size as restored size since it is the initial queue entry size" or may be just keep it without comment at all, this is in general clear probably and already covered by comments in other places

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -2532,23 +2532,34 @@ bool FileCache::doDynamicResizeImpl(
...
+                restored_size,
                 original_queue_type,
                 cache_write_lock,
                 &state_lock);
 
             candidate->setRemovedFlag(*locked_key, /* value */false);
-            file_segment->setQueueIterator(main_priority_iterator);
+            {
+                auto seg_lock = file_segment->lock();
+                file_segment->restoreQueueIteratorAfterDelayedRemoval(main_priority_iterator, seg_lock);
```
kssenii: Let's make `restoreQueueIteratorAfterDelayedRemoval` self locking? E.g. lock to be taken inside the method. As I remember all or may be 99% of public methods in `FileSegment` are self locking and `const FileSegmentGuard::Lock &` is used as an argument only for private methods.

antonio2368: Yes, it does seem out of place like this.

src/Interpreters/FileCache/FileCache.cpp
antonio2368: I 100% agree, agents really like to generate huge comment slop. Worst part is when it references past implementations...
I'll fix it.

kssenii (approved):

PR #104930, opened 2026-05-14
Define USE_OPENSSL_FIPS in config.h.in to avoid -Wundef

vadimskipin:
### Summary

USE_OPENSSL_FIPS is referenced in programs/main.cpp and programs/keeper/keeper_main.cpp under some build configurations, but the macro was not declared in config.h.in. In builds where Common/config.h does not define it, -Werror=undef fires:

```
programs/main.cpp:301:28: error: 'USE_OPENSSL_FIPS' is not defined, evaluates to 0 [-Werror,-Wundef]
```

Declaring it as cmakedefine01 makes the macro always defined (to 0 by default), matching how every other USE_* macro is handled in this codebase and how the corresponding sites already test it (`#if USE_OPENSSL_FIPS`).

### Version info
- Merged into: `26.5.1.693`

### DIFF
--- a/src/Common/config.h.in
+++ b/src/Common/config.h.in
@@ -86,5 +86,6 @@
 #cmakedefine01 USE_LIBCOTP
 #cmakedefine01 USE_WASMEDGE
 #cmakedefine01 USE_WASMTIME
+#cmakedefine01 USE_OPENSSL_FIPS
 #cmakedefine01 WITH_COVERAGE
 


nikitamikhaylov (approved):

vadimskipin:
@groeneai this does not seem to be caused by this change:
https://github.com/ClickHouse/ClickHouse/actions/runs/25856767507/job/75984897977?pr=104930

Algunenano:
The CI failures:
* Stateless tests (amd_msan, WasmEdge, parallel, 2/2) -> https://github.com/ClickHouse/ClickHouse/pull/104947
* Stress test (arm_asan_ubsan, s3) -> https://github.com/ClickHouse/ClickHouse/pull/104944

groeneai:
@vadimskipin — confirmed, the failure is unrelated to this PR.

**Failure:** `Hung check failed, possible deadlock found` on `Stress test (arm_asan_ubsan, s3)`.

**Cross-PR scope (CIDB, last 30d):** 17 unrelated PRs and 3 master hits between 2026-04-16 and 2026-05-15. A chronic stress-test deadlock, not a regression introduced here.

**Root cause + fix:** `system sync merges` queries entered an infinite loop in stress tests. Fixed by @Michicosun in #104944 (merged 2026-05-15), which makes the query loop up to `receive_timeout` and respect cancellation — same diagnosis @Algunenano linked to above.

**Why this PR cannot be involved:** the diff is a single `#cmakedefine01 USE_OPENSSL_FIPS` line in `src/Common/config.h.in`. It only ensures `USE_OPENSSL_FIPS` evaluates cleanly under `-Wundef`; FIPS is OFF by default in this build so the macro expands to `0` and no runtime path differs from master. There is no mechanism by which it can reach the stress-test deadlock site.

Safe to merge — @nikitamikhaylov has already approved.

PR #104940, opened 2026-05-14
Fix max_memory_usage_soft_limit hot reload

nauu:

When `max_memory_usage_soft_limit` is not set explicitly and is auto-calculated from `max_memory_usage_soft_limit_ratio`, `initializeKeeperMemorySoftLimit` called `config.setUInt64()` to cache the result into the writable `MapConfiguration` layer(`PRIO_APPLICATION = -100`). 

On config reload triggered by `CgroupsMemoryUsageObserver`(e.g., when the cgroup memory limit increases), `LayeredConfiguration::replace()`  only swaps the file layer (`PRIO_DEFAULT = 0`) but leaves `MapConfiguration` intact. 

Because `MapConfiguration` has higher priority, the stale startup value shadows the new file config, and `updateKeeperMemorySoftLimit` reads the old value instead of recalculating.

Fix: extract `calculateMemorySoftLimit` and remove the `config.setUInt64()` call, so each reload recalculates from the current config and `getMemoryAmount()`.

Fixed the `max_memory_usage_soft_limit` was not updated on-the-fly when the configuration changed. The soft memory limit was calculated once at startup and cached, so subsequent config reloads had no effect. After this fix, the limit is recalculated and applied immediately on every config reload, no Keeper restart required.

### Version info
- Merged into: `26.5.1.893` (included in `26.5` and later)
- Backported to: `26.3.33.43`

### DIFF
--- a/src/Coordination/KeeperContext.cpp
+++ b/src/Coordination/KeeperContext.cpp
@@ -593,30 +593,29 @@ void KeeperContext::initializeFeatureFlags(const Poco::Util::AbstractConfigurati
     feature_flags.logFlags(getLogger("KeeperContext"));
 }
 
-void KeeperContext::updateKeeperMemorySoftLimit(const Poco::Util::AbstractConfiguration & config)
+static UInt64 calculateMemorySoftLimit(const Poco::Util::AbstractConfiguration & config)
 {
     if (config.hasProperty("keeper_server.max_memory_usage_soft_limit"))
-        memory_soft_limit = config.getUInt64("keeper_server.max_memory_usage_soft_limit");
-}
-
-void KeeperContext::initializeKeeperMemorySoftLimit(Poco::Util::AbstractConfiguration & config, Poco::Logger * log)
-{
-    UInt64 memory_soft_limit = 0;
-    if (config.has("keeper_server.max_memory_usage_soft_limit"))
-        memory_soft_limit = config.getUInt64("keeper_server.max_memory_usage_soft_limit");
-
-    if (memory_soft_limit == 0)
     {
-        Float64 ratio = config.getDouble("keeper_server.max_memory_usage_soft_limit_ratio", 0.9);
-        size_t physical_server_memory = getMemoryAmount();
-        if (ratio > 0 && physical_server_memory > 0)
-        {
-            memory_soft_limit = static_cast<UInt64>(static_cast<Float64>(physical_server_memory) * ratio);
-            config.setUInt64("keeper_server.max_memory_usage_soft_limit", memory_soft_limit);
-        }
+        UInt64 explicit_value = config.getUInt64("keeper_server.max_memory_usage_soft_limit");
+        // 0 falls through to ratio-based calculation for backward compatibility.
+        if (explicit_value > 0)
+            return explicit_value;
     }
 
-    LOG_INFO(log, "keeper_server.max_memory_usage_soft_limit is set to {}", formatReadableSizeWithBinarySuffix(memory_soft_limit));
+    Float64 ratio = config.getDouble("keeper_server.max_memory_usage_soft_limit_ratio", 0.9);
+    size_t physical_server_memory = getMemoryAmount();
+    if (ratio > 0 && physical_server_memory > 0)
+        return static_cast<UInt64>(static_cast<Float64>(physical_server_memory) * ratio);
+
+    return 0;
+}
+
+void KeeperContext::updateKeeperMemorySoftLimit(const Poco::Util::AbstractConfiguration & config)
+{
+    const auto limit = calculateMemorySoftLimit(config);
+    memory_soft_limit = limit;
+    LOG_INFO(getLogger("KeeperContext"), "keeper_server.max_memory_usage_soft_limit is set to {}", formatReadableSizeWithBinarySuffix(limit));
 }
 
 void KeeperContext::updateSettings(CoordinationSettingsPtr new_settings)
--- a/tests/integration/test_keeper_memory_soft_limit/test.py
+++ b/tests/integration/test_keeper_memory_soft_limit/test.py
@@ -110,6 +110,8 @@ def test_soft_limit_create(started_cluster):
     raise Exception("all records are inserted but no error occurs")
 
 
+# Verify that `max_memory_usage_soft_limit` is correctly updated
+# on config reload without requiring keeper restart.
 def test_soft_limit_hot_reload(started_cluster):
     started_cluster.wait_zookeeper_to_start()
 
--- a/tests/integration/test_keeper_memory_soft_limit/test.py
+++ b/tests/integration/test_keeper_memory_soft_limit/test.py
@@ -1,11 +1,15 @@
 #!/usr/bin/env python3
 import logging
+import os
+import time
 from concurrent.futures import ThreadPoolExecutor, TimeoutError
 
 import pytest
 from kazoo.client import KazooClient
 
 from helpers.cluster import ClickHouseCluster
+
+CURRENT_TEST_DIR = os.path.dirname(os.path.abspath(__file__))
 
 cluster = ClickHouseCluster(__file__, keeper_config_dir="configs/")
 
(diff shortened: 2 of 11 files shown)


Ergus (commented):
The fix as is seems correct (LGTM), but there are a couple or corner cases/paths not covered in tests:

1. If we set `max_memory_usage_soft_limit` directly in the config then `calculateMemorySoftLimit` takes the `hasProperty` branch and there is no test for that:
  We need a test that:
  1. Sets an explicit value at startup and verifies it is respected
  2. Changes the explicit value on reload and verifies the new value is picked up

2. If we change ratio at startup. After the PR's fix, this should work: the `MapConfiguration` is clean (as there is not call to `setUInt64` now), so `hasProperty("...soft_limit")` is false at startup (correct ratio path), and true after reload with an explicit value (correct explicit path).

3. When `calculateMemorySoftLimit` returns 0 (i.e ratio = 0). Callers of `getKeeperMemorySoftLimit()` must treat 0 as "no limit".

nauu:
> The fix as is seems correct (LGTM), but there are a couple or corner cases/paths not covered in tests:
> 
> 1. If we set `max_memory_usage_soft_limit` directly in the config then `calculateMemorySoftLimit` takes the `hasProperty` branch and there is no test for that:
>    We need a test that:
> 2. Sets an explicit value at startup and verifies it is respected
> 3. Changes the explicit value on reload and verifies the new value is picked up
> 4. If we change ratio at startup. After the PR's fix, this should work: the `MapConfiguration` is clean (as there is not call to `setUInt64` now), so `hasProperty("...soft_limit")` is false at startup (correct ratio path), and true after reload with an explicit value (correct explicit path).
as "no limit".

I haven't change any other existing logic in this PR, however I've added more test cases

> 5. When `calculateMemorySoftLimit` returns 0 (i.e ratio = 0). Callers of `getKeeperMemorySoftLimit()` must treat 0 

I thought this is guaranteed by this part
https://github.com/ClickHouse/ClickHouse/blob/b6844b26db6a2981b9bd9fcc4b8922eef6b16025/src/Coordination/KeeperServer.cpp#L795

nauu:
The failed test `02346_text_index_queries` has a known issue https://github.com/ClickHouse/ClickHouse/issues/104923

It is unrelated to this PR.

Ergus:
> I haven't change any other existing logic in this PR, however I've added more test cases
> 
I know, but as this part is a fragile defined behavior, the extra tests simplify detectin conflicting changes in apparently unrelated parts in the future.

> > 5. When `calculateMemorySoftLimit` returns 0 (i.e ratio = 0). Callers of `getKeeperMemorySoftLimit()` must treat 0
> 
> I thought this is guaranteed by this part
> 
> https://github.com/ClickHouse/ClickHouse/blob/b6844b26db6a2981b9bd9fcc4b8922eef6b16025/src/Coordination/KeeperServer.cpp#L795

Indeed, but like in crypto: "Don't Trust, Verify" ;)

nauu:
> > I haven't change any other existing logic in this PR, however I've added more test cases
> 
> I know, but as this part is a fragile defined behavior, the extra tests simplify detectin conflicting changes in apparently unrelated parts in the future.
> 
> > > 5. When `calculateMemorySoftLimit` returns 0 (i.e ratio = 0). Callers of `getKeeperMemorySoftLimit()` must treat 0
> > 
> > 
> > I thought this is guaranteed by this part
> > https://github.com/ClickHouse/ClickHouse/blob/b6844b26db6a2981b9bd9fcc4b8922eef6b16025/src/Coordination/KeeperServer.cpp#L795
> 
> Indeed, but like in crypto: "Don't Trust, Verify" ;)

Hi @Ergus thanks your reply . 

In the original logic, when `ratio = 0`, `memory_soft_limit` remains 0 and `setUInt64` is never called . so the behavior is identical to what we have now. 

The only caller of `getKeeperMemorySoftLimit()` is `isExceedingMemorySoftLimit`, which already guards with `mem_soft_limit > 0`. 

Could you clarify what specific test case you have in mind here? I am happy to add it if there's a concrete scenario to cover.

rschu1ze:
@nauu Normal users won't be able to decipher the changelog entry `Fixed Keeper max_memory_usage_soft_limit not hot updating after config reload.` Please rewrite and expand.

Ergus:
@rschu1ze 

I would rephrase the changelog more or less like this:

```
Fixed on-the-fly update of soft memory limit when the configuration changes by  recalculating and applying the value on every config reload without requiring keeper restart.
```

src/Coordination/KeeperContext.cpp
```diff
@@ -593,30 +593,33 @@ void KeeperContext::initializeFeatureFlags(const Poco::Util::AbstractConfigurati
...
-        {
-            memory_soft_limit = static_cast<UInt64>(static_cast<Float64>(physical_server_memory) * ratio);
-            config.setUInt64("keeper_server.max_memory_usage_soft_limit", memory_soft_limit);
-        }
-    }
+    memory_soft_limit = calculateMemorySoftLimit(config);
+}
 
-    LOG_INFO(log, "keeper_server.max_memory_usage_soft_limit is set to {}", formatReadableSizeWithBinarySuffix(memory_soft_limit));
+void KeeperContext::initializeKeeperMemorySoftLimit(Poco::Util::AbstractConfiguration & config, Poco::Logger * log)
```
Ergus: The name of this function is misleading now because it is not changing the `memory_soft_limit`

Ergus: Also I think that the config argument can be `const` now. Which requires updatign the comment in the header

src/Coordination/tests/gtest_keeper_memory_soft_limit.cpp
```diff
@@ -0,0 +1,179 @@
...
+        "</keeper_server></clickhouse>");
+    config->replace("default", reload_xml.get(), Poco::Util::Application::PRIO_DEFAULT, /*shared=*/true);
+
+    KeeperContext::initializeKeeperMemorySoftLimit(*config, log);
+    ctx->updateKeeperMemorySoftLimit(*config);
+
+    const UInt64 expected_reload = static_cast<UInt64>(static_cast<double>(getMemoryAmount()) * 0.5);
+    ASSERT_EQ(ctx->getKeeperMemorySoftLimit(), expected_reload);
+}
+
```
Ergus: I would recommend to add a test for explicit value -> ratio-only config on reload. 

Similar to `ExplicitZeroFallsBackToRatio` bit where  `max_memory_usage_soft_limit_ratio=0`

src/Coordination/KeeperContext.cpp
```diff
@@ -593,30 +593,33 @@ void KeeperContext::initializeFeatureFlags(const Poco::Util::AbstractConfigurati
...
-    {
-        Float64 ratio = config.getDouble("keeper_server.max_memory_usage_soft_limit_ratio", 0.9);
-        size_t physical_server_memory = getMemoryAmount();
-        if (ratio > 0 && physical_server_memory > 0)
-        {
-            memory_soft_limit = static_cast<UInt64>(static_cast<Float64>(physical_server_memory) * ratio);
-            config.setUInt64("keeper_server.max_memory_usage_soft_limit", memory_soft_limit);
-        }
-    }
+    memory_soft_limit = calculateMemorySoftLimit(config);
```
Ergus: Add a log line in this function so the actual stored value is always observable.

src/Coordination/KeeperContext.cpp
```diff
@@ -593,30 +593,33 @@ void KeeperContext::initializeFeatureFlags(const Poco::Util::AbstractConfigurati
...
+}
 
-    LOG_INFO(log, "keeper_server.max_memory_usage_soft_limit is set to {}", formatReadableSizeWithBinarySuffix(memory_soft_limit));
+void KeeperContext::initializeKeeperMemorySoftLimit(Poco::Util::AbstractConfiguration & config, Poco::Logger * log)
+{
+    const auto limit = calculateMemorySoftLimit(config);
+    LOG_INFO(log, "keeper_server.max_memory_usage_soft_limit is set to {}", formatReadableSizeWithBinarySuffix(limit));
```
Ergus: I think that this is the log line needed in `updateKeeperMemorySoftLimit`

Ergus (changes requested):

Ergus:
The failing test is unrelated.

src/Coordination/tests/gtest_keeper_memory_soft_limit.cpp
rschu1ze, line 1: @nauu Can you please write an integration test instead? A C++ unit test seems too low level for the job.

tests/integration/test_keeper_memory_soft_limit/test.py
```diff
@@ -104,3 +108,61 @@ def test_soft_limit_create(started_cluster):
         return
 
     raise Exception("all records are inserted but no error occurs")
+
+
+def test_soft_limit_hot_reload(started_cluster):
```
rschu1ze: Needs a test-level comment for the next guy who will read this test in a couple of weeks from now, basically _what_ is the test supposed to test.

tests/integration/test_keeper_memory_soft_limit_ratio/configs/keeper_config1.xml
```diff
@@ -0,0 +1,45 @@
+<clickhouse>
+    <listen_try>true</listen_try>
+    <listen_host>::</listen_host>
```
Ergus, line 3: Do we need to listen from any host?

tests/integration/test_keeper_memory_soft_limit_ratio/configs/keeper_config1.xml
```diff
@@ -0,0 +1,45 @@
...
+            </server>
+            <server>
+                <id>2</id>
+                <hostname>zoo2</hostname>
+                <port>9444</port>
+            </server>
+            <server>
+                <id>3</id>
+                <hostname>zoo3</hostname>
+                <port>9444</port>
```
Ergus, line 41: How many servers we need for this test?

Ergus (commented):
I liked the unit tests, but I agree that with integration tests it is simpler to test.

tests/integration/test_keeper_memory_soft_limit_ratio/configs/keeper_config1.xml
nauu: This is a normal config same as other keeper test

tests/integration/test_keeper_memory_soft_limit_ratio/configs/keeper_config1.xml
nauu: Only need one, However I still used three nodes form a complete cluster

Ergus (approved):

PR #104941, opened 2026-05-14
Pin `Keeper` snapshots during transfer

antonio2368:
Pin local `Keeper` snapshot files while they are being transferred to recovering followers. This prevents snapshot rotation, cleanup, or cross-disk movement from removing the file behind an in-flight read.

Fix a race in `Keeper` snapshot transfer where a snapshot could be removed or moved while it was being sent to a recovering follower.

### Version info
- Merged into: `26.5.1.804`


alexey-milovidov:

clickgapai:

**Affected code:** `KeeperSnapshotManager::moveSnapshotsIfNeeded`, `KeeperSnapshotManager::removeSnapshot`, `IKeeperStateMachine::read_logical_snp_obj` — core code present in all supported branches.

**Why this might apply:** This fixes a race condition in Keeper snapshot transfer that could cause follower recovery failures. The affected code (`KeeperSnapshotManager`, `KeeperStateMachine`) has existed since Keeper was introduced and the bug manifests under snapshot rotation during recovery - a real-world scenario. Users running multi-node Keeper clusters with `snapshots_to_keep=1` and frequent snapshot creation are most affected.

PR #95905, opened 2026-02-03
Forbid DETACH TEMPORARY TABLE

azat:
### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Forbid DETACH TEMPORARY TABLE

Follow-up for: https://github.com/ClickHouse/ClickHouse/pull/89450 (cc @AVMusorin )

### Version info
- Merged into: `26.2.1.225`

PR #104943, opened 2026-05-14
Forbid `DETACH TABLE` on temporary tables (without `TEMPORARY` keyword)

### LINKED ISSUE #103475: `DETACH TABLE` (without `TEMPORARY` keyword) silently succeeds on temporary tables due to narrowed guard in `executeToTableImpl`

_Retrospective finding from a historical scan of [PR #95905](https://github.com/ClickHouse/ClickHouse/pull/95905) (merged 2026-02-04). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`DETACH TABLE tmp` (without the `TEMPORARY` keyword) when `tmp` is a temporary table silently succeeds instead of throwing `SYNTAX_ERROR`. Before this PR, the guard in `executeToTemporaryTable` blocked ALL detach operations on temp tables. After this PR, the guard in `executeToTableImpl` only checks `query.isTemporary()`, which is false when the `TEMPORARY` keyword is not used.

**Root cause:** InterpreterDropQuery.cpp:154: The guard `query.kind == ASTDropQuery::Kind::Detach && query.isTemporary()` only catches `DETACH TEMPORARY TABLE` syntax. When `DETACH TABLE tmp` is used (no TEMPORARY keyword), `isTemporary()` returns false, bypassing the guard. The old guard in `executeToTemporaryTable` (removed at old line ~331-333) checked `kind == Detach` unconditionally, catching all paths.

**Why we believe this is a bug:** `executeToTableImpl` (InterpreterDropQuery.cpp:152) → line 154 guard checks `query.isTemporary()` (false for `DETACH TABLE tmp`) → line 159 enters branch due to `table_id.database_name.empty()` → line 161-162 resolves temp table and calls `executeToTemporaryTable` → line 390-392 sets `table->is_detached = true` silently (old guard was removed)

**Affected locations:**
- `src/Interpreters/InterpreterDropQuery.cpp:154` — Guard only checks query.isTemporary(), misses non-keyword path
- `src/Interpreters/InterpreterDropQuery.cpp:390` — Unguarded Detach path sets is_detached=true on temp table

**Impact:** `DETACH TABLE tmp` on a temporary table silently succeeds instead of producing an error. The table remains fully functional (the `is_detached` flag has no observable effect on Memory engine), making the operation a confusing no-op. This is a behavioral regression from the pre-PR behavior where this would throw 'Unable to detach temporary table.'

Yes — confirmed on current `master` (commit `f17036af6b3a`).

### How to reproduce

```sql
-- Test: DETACH TABLE (without TEMPORARY keyword) should also be forbidden for temporary tables
-- Covers: InterpreterDropQuery.cpp:154 guard only checks query.isTemporary(), missing non-TEMPORARY path

-- Step 1: Create a temp table and verify it works
CREATE TEMPORARY TABLE t_04113 (x UInt32) ENGINE = Memory;
INSERT INTO t_04113 VALUES (1), (2), (3);
SELECT 'before', count() FROM t_04113;

-- Step 2: DETACH TEMPORARY TABLE should correctly error (PR's new guard)
DETACH TEMPORARY TABLE t_04113; -- { serverError SYNTAX_ERROR }

-- Step 3: DETACH TABLE (without TEMPORARY keyword) on a temp table
-- Before PR#95905 this threw "Unable to detach temporary table."
-- After PR#95905 this silently succeeds (regression)
DETACH TABLE t_04113; -- { serverError SYNTAX_ERROR }

-- Step 4: Verify table still works after attempted detach
SELECT 'after', count() FROM t_04113;

DROP TEMPORARY TABLE t_04113;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/df5fdc53-cb7f-406f-af5c-f48011bfb820)

### Expected behavior

```
before	3
after	3
```

```
before	3
The query succeeded but the server error '62' was expected (query: DETACH TABLE t_04113; -- { serverError SYNTAX_ERROR }).
after	3
```

### Additional context

**Open risks:**
- For MergeTree-engine temporary tables, setting is_detached=true could additionally suppress background merges, though the table remains queryable

**Suggested fix:** Restore the guard inside `executeToTemporaryTable` at line 371: add `if (kind == ASTDropQuery::Kind::Detach) throw Exception(ErrorCodes::SYNTAX_ERROR, "DETACH of TEMPORARY tables are not supported");` before the existing logic. This catches both the `DETACH TEMPORARY TABLE` and `DETACH TABLE` paths. Alternatively, move the check in `executeToTableImpl` to after the temp table resolution (between lines 161-162).

Found during automated review of [PR #95905](https://github.com/ClickHouse/ClickHouse/pull/95905).

---
_ClickGapAI · Confidence: HIGH · Severity: P3 · Finding: `h_pr95905_001`_

### Version info
- Resolved by: #104943
- Merged into: `26.5.1.661`

groeneai:
The unqualified form `DETACH TABLE tmp` on a temporary table silently no-op'd instead of raising `SYNTAX_ERROR`. PR #95905 (merged 2026-02-04) replaced the broad `kind == Detach` guard inside `InterpreterDropQuery::executeToTemporaryTable` with a narrower guard `query.kind == Detach && query.isTemporary()` inside `executeToTableImpl`. The narrowed guard only catches the explicit `DETACH TEMPORARY TABLE` syntax — when a user types `DETACH TABLE tmp` (no `TEMPORARY` keyword) and `tmp` resolves to a temporary table via `Context::ResolveExternal`, the guard is bypassed and we reach `executeToTemporaryTable` with `Kind::Detach`. That handler only set `is_detached = true` (an unused flag) and returned without error.

This change restores the broader guard at the entry of `executeToTemporaryTable`, so both `DETACH TEMPORARY TABLE` and `DETACH TABLE` on a temporary table consistently raise `SYNTAX_ERROR` with the same message. The narrow `executeToTableImpl` guard is kept as a fast path for the explicit form, which preserves the existing behavior that the explicit syntax throws even if the table does not exist.

Root cause analysis comes from the automated review of PR #95905 by @clickgapai.

Reproducer (from issue #103475):

```sql
CREATE TEMPORARY TABLE t (x UInt32) ENGINE = Memory;
DETACH TEMPORARY TABLE t;  -- correctly throws SYNTAX_ERROR (existing behavior)
DETACH TABLE t;            -- before: silent no-op; after: throws SYNTAX_ERROR
SELECT count() FROM t;     -- still 3 (table never actually detached anyway)
```

Closes #103475.

`DETACH TABLE` on a temporary table (without the `TEMPORARY` keyword) now correctly raises `SYNTAX_ERROR`, matching the behavior of `DETACH TEMPORARY TABLE`. Previously it silently set an internal `is_detached` flag and returned without error. To remove a temporary table, use `DROP TEMPORARY TABLE` or `DROP TABLE` (the latter resolves the temporary table via `Context::ResolveExternal`). Closes [#103475](https://github.com/ClickHouse/ClickHouse/issues/103475).

- [ ] Documentation is unchanged. No new user-facing surface; the change tightens the error path for an already-disallowed operation.

### Version info
- Merged into: `26.5.1.661`


groeneai:
cc @azat @alexey-milovidov — could you review this? It restores the broader `kind == Detach` guard at the entry of `InterpreterDropQuery::executeToTemporaryTable` (originally removed in PR #95905) so that the unqualified form `DETACH TABLE tmp` on a temporary table now raises `SYNTAX_ERROR` instead of silently no-op'ing. @azat — you flagged this as trivial in #103475 and authored the original PR #95905, so I picked the most additive shape; happy to switch to removing the narrow `executeToTableImpl` guard instead if you prefer a single check.

groeneai:
## Pre-PR Validation Gate

Session: `cron:clickhouse-ci-task-worker:20260514-131500`

**(a) Deterministic repro?** Yes. On master at `5794d425fbf`:

```
CREATE TEMPORARY TABLE t_103475 (x UInt32) ENGINE = Memory;
INSERT INTO t_103475 VALUES (1), (2), (3);
DETACH TABLE t_103475;          -- silently succeeded
SELECT count() FROM t_103475;   -- returned 3 (table never actually detached)
```

No randomization needed — fully deterministic.

**(b) Root cause explained?** PR #95905 moved the broad guard `if (kind == ASTDropQuery::Kind::Detach) throw SYNTAX_ERROR` out of `executeToTemporaryTable` and replaced it with a narrow guard inside `executeToTableImpl`: `if (query.kind == Detach && query.isTemporary()) throw`. The new guard only fires when `query.isTemporary()` is true, i.e. when the user explicitly wrote `DETACH TEMPORARY TABLE`. For the unqualified `DETACH TABLE tmp` form, `isTemporary()` returns false → the guard does not fire → at line 161-164 `table_id.database_name.empty()` is true → `tryResolveStorageID(..., Context::ResolveExternal)` succeeds → `executeToTemporaryTable(..., Kind::Detach)` is called → the function just sets `table->is_detached = true` (an unused flag) and returns. Net effect: silent no-op.

**(c) Fix matches root cause?** Yes. The fix restores the broad `kind == Detach` guard at the entry of `executeToTemporaryTable`, so that any Detach reaching the temp-table handler — including the bypass path — raises `SYNTAX_ERROR` with the same message. The narrow `executeToTableImpl` guard is retained as a fast path: it still throws on `DETACH TEMPORARY TABLE non_existent_table` before resolution, preserving the existing error semantics for that input.

**(d) Test intent preserved? / new test added?** New regression test `04239_detach_table_on_temporary_table_103475.sql` covers both shapes (`DETACH TEMPORARY TABLE` and `DETACH TABLE` on a temp) and asserts both raise `SYNTAX_ERROR`, plus a post-attempt `SELECT count()` to verify the table is still accessible after the rejected attempts. The pre-existing test `03701_create_or_replace_temporary_table.sql` (which covers only the `TEMPORARY` keyword form) still passes — intent preserved.

- Without the fix (master `5794d425fbf`): `tests/clickhouse-test 04239_detach_table_on_temporary_table_103475` → FAIL: "The query succeeded but the server error '62' was expected (DETACH TABLE t_103475)".
- With the fix: `tests/clickhouse-test --test-runs 50` → 50/50 PASS.

**(f) Fix is general, not a narrow patch?** Both entry paths to a Detach on a temp table are covered:
1. `DETACH TEMPORARY TABLE tmp` → `executeToTableImpl` guard (line 156-157, unchanged) throws immediately.
2. `DETACH TABLE tmp` (tmp is a temp) → resolves via `Context::ResolveExternal` → `executeToTemporaryTable` → new guard throws.

The fully-qualified form `DETACH TABLE _temporary_and_external_tables.tmp` does not reach the temp-handler path (`database_name.empty()` is false), so it follows the normal-table lookup which raises `UNKNOWN_TABLE` since temp tables are not addressable that way — already correct, not affected. The fix is not a defensive null check at the symptom site; it restores an invariant guard at the function entry, which is where intent ("`Kind::Detach` is never valid for temp tables") is asserted.

azat (commented):
Looks good, but, do we need a check in `InterpreterDropQuery::executeToTableImpl` then?

groeneai:
@azat — yes, both are needed because they cover different code paths.

The narrow guard at `InterpreterDropQuery::executeToTableImpl` (`src/Interpreters/InterpreterDropQuery.cpp:156`) fires for the explicit `DETACH TEMPORARY TABLE` form (`query.isTemporary() == true`) **before** any resolve. The broad guard at the entry of `executeToTemporaryTable` only fires after `Context::ResolveExternal` succeeds — i.e. after we have already proven the table exists *and* is temporary.

| Query | Narrow guard (`executeToTableImpl`) | Broad guard (`executeToTemporaryTable`) | Result if we drop the narrow guard |
|---|---|---|---|
| `DETACH TEMPORARY TABLE existing_tmp` | fires | fires (after resolve) | OK — still `SYNTAX_ERROR` |
| `DETACH TABLE existing_tmp` | not fires (`isTemporary()=false`) | fires | OK |
| `DETACH TEMPORARY TABLE non_existing` | fires | not reached (resolve fails) | becomes `UNKNOWN_TABLE` — wrong error class |
| `DETACH TEMPORARY TABLE IF EXISTS non_existing` | fires | not reached (resolve fails) | **silently no-ops** via `if (query.if_exists) return {};` in `executeToTableImpl` |

The last row is the strongest reason to keep the narrow guard: without it, `DETACH TEMPORARY TABLE IF EXISTS non_existing` re-enters the same "silently succeeds" bug class this PR fixes. The narrow guard runs *before* the resolution + `if_exists` short-circuit, so it covers that path; the broad guard cannot, because we never reach `executeToTemporaryTable` when the resolve fails.

I just pushed d3198ace656 which adds a regression test case for the `DETACH TEMPORARY TABLE IF EXISTS no_such_t_103475` path — it locks down the rationale for keeping both guards.

azat (approved):

azat:
>[Stress test (arm_asan_ubsan, s3)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104943&sha=d3198ace6563b4303705d12311d683507dc9c917&name_0=PR&name_1=Stress%20test%20%28arm_asan_ubsan%2C%20s3%29)

Only this one and nothing more

```
2026.05.14 20:06:37.860681 [ 140535 ] {} <Fatal> Application: Child process was terminated by signal 6.
```

clickgapai:

**Affected code:** `InterpreterDropQuery::executeToTemporaryTable` - the guard for rejecting DETACH operations on temporary tables — introduced in 26.2.

**Why this might apply:** This PR fixes a regression from PR #95905 (merged 2026-02-04). The regression was introduced after the 25.8 LTS release (2025-08-29) but before the 26.2 release (2026-02-27), so it affects all branches released after that date: 26.2, 26.3, and 26.4. The 25.8 LTS branch does not contain the regression.

PR #104946, opened 2026-05-14
Fix JIT symbol lookup on macOS

leshikus:
On macOS, `llvm::Mangler::getNameWithPrefix` prepends `_` to C symbol names, but
`CHJIT` registers external symbols unmangled (`memset`, `memcpy`,
`memcmpSmallCharsAllowOverflow15`, etc.). `JITSymbolResolver::findSymbol` then
gets handed the mangled name and the lookup misses, producing:

```
Code: 263. DB::Exception: Could not find symbol _memcmpSmallCharsAllowOverflow15.
(CANNOT_COMPILE_CODE)
```

This fires for any query that JIT-compiles a comparator over a `String` column,
e.g. an `ORDER BY` with `compile_sort_description = 1`. It's one of the two
root causes called out in the `Fast test (arm_darwin)` umbrella issue, and the
one I can reproduce locally on Apple Silicon.

Linux is unaffected because LLVM's mangler doesn't add the `_` prefix on ELF
targets, so the lookup succeeds on the first try.

## Approach

In `JITSymbolResolver::findSymbol`, on a miss for a name that begins with `_`,
retry the lookup with the leading underscore stripped. No platform `#ifdef` —
the fallback is unreachable on Linux because the first lookup already succeeds.
Six inserted lines, no behavior change for any name that's already in the map.

A small regression test under `tests/queries/0_stateless/` exercises the exact
path that used to fail (sort description JIT on a `String` column).

## Verification

Built `Build (arm_darwin)` locally and verified:

- The previously failing query now succeeds (correct result, `Compile sort
  description (type: String, ...)` trace emitted, no `Could not find symbol`).
- New test `04230_jit_sort_description_string_symbol` passes via
  `python3 -m ci.praktika run "Fast test (arm_darwin)"`.
- Existing JIT-tagged fast tests pass with no regressions (10/10 — sort
  description, expression, aggregate JIT paths all exercised).

Fixes part of https://github.com/ClickHouse/ClickHouse/issues/103235.

Fix JIT symbol resolution on macOS so that sort, expression, and aggregate JIT compilation no longer fail with `CANNOT_COMPILE_CODE: Could not find symbol _<name>`.

### Version info
- Backported to: `26.5.3.35`


src/Interpreters/JIT/CHJIT.cpp
```diff
@@ -340,6 +340,12 @@ class JITSymbolResolver : public llvm::LegacyJITSymbolResolver
     llvm::JITSymbol findSymbol(const std::string & Name) override
     {
         auto address_it = symbol_name_to_symbol_address.find(Name);
+        if (address_it == symbol_name_to_symbol_address.end() && Name.size() > 1 && Name[0] == '_')
```
Algunenano: This is too broad as it affects every platform, which introduces some risks.

There are 2 options:
* Enable it only for Darwin.
* Canonicalize at registration time, with mangle names (both in registerSymbol and findSymbol). I think this is preferred as it removes fallbacks

Algunenano (changes requested):

src/Interpreters/JIT/CHJIT.cpp
leshikus: Done in 35c0e5bda52 — went with option 2 (canonicalize at registration time). `registerSymbol` now mangles via `llvm::Mangler::getNameWithPrefix` using the JIT's `DataLayout`, so the map keys match exactly what LLVM hands to `findSymbol`. The fallback path in `findSymbol` is gone. Behavior on Linux is unchanged because the ELF mangler doesn't add a prefix.

leshikus: The test is not good enough though, looking into it

leshikus:
Bugfix validation expectedly fails - it uses linux to check

Algunenano (commented):
We can just drop the test. If it works and solves Darwin I trust you. We have many JIT tests that will pass in that platform now

leshikus:
Raul, I just need a bit more time to complete this, after I resolve first priority issues. I want reliable reproducer before committing this.

Algunenano:
@leshikus Don't worry I got it. I'll test it on Darwin and merge, don't worry

Algunenano (approved):
I can't reproduce the issue locally without this patch, because I'm using LLVM 22 
to build and apparently it does not have this problem. But using the CI build from master it does reproduce, and this PR fixes it.

leshikus:
Thanks for taking care of it

PR #104969, opened 2026-05-14
Speed up server shutdown for databases with many tables

jaymebrd:
Table shutdown runs sequentially per table, which can make server shutdown take many minutes to 1hr+ on clusters with many thousands of tables.                                         
   
 This PR parallelises  the table shutdown drain in`DatabaseWithOwnTablesBase::shutdown`:                                                                                                                                                                                     
  - `src/Databases/DatabasesCommon.cpp`: the prepare phase `flushAndPrepareForShutdown` stays sequential and the expensive, table-independent `shutdown()` drain (ZooKeeper sessions, interserver endpoints, background pools) runs concurrently via a dedicated thread pool.                                                                                                                                                                                                                                                                                          
                                                            
 The speedup is best measured by SIGTERM-to-exit wall time on many-table clusters.

- Performance Improvement

Speed up server shutdown for databases with many tables by shutting tables down in parallel. Controlled by the new `database_catalog_shutdown_table_concurrency` setting (0 by default, meaning the number of CPU cores).

<!-- CI automatic block start :ci_links: -->

---
Workflow [[PR](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104969&sha=latest&name_0=PR)]
Sync PR [[sync-upstream/pr/104969](https://github.com/search?q=head%3Async-upstream%2Fpr%2F104969+org%3AClickHouse+type%3Apr&type=pullrequests)]
<!-- CI automatic block end :ci_links: -->


<!-- ch-version-info:start -->
### Version info
- Merged into: `26.9.1.221` (included in `26.9` and later)
<!-- ch-version-info:end -->


src/Databases/DatabasesCommon.cpp
```diff
@@ -536,20 +541,51 @@ void DatabaseWithOwnTablesBase::shutdown()
...
+    /// a Database destructor don't trip on an uninitialized StaticThreadPool.
+    auto & shared_pool = getDatabaseCatalogShutdownTablesThreadPool();
+    shared_pool.initializeWithDefaultSettingsIfNotInitialized();
+    auto & pool = shared_pool.get();
+    const auto db_name = getDatabaseName();
+    /// If our tables carry UUIDs (Atomic-style), the database must too. Memory-backed
+    /// databases like `system` or `information_schema` have neither and skip the check.
+    if (!tables_snapshot.empty() && tables_snapshot.begin()->second->getStorageID().hasUUID())
+        chassert(db_name == DatabaseCatalog::TEMPORARY_DATABASE || getUUID() != UUIDHelpers::Nil);
+    const bool should_slow_shutdown_for_tests = !DatabaseCatalog::isPredefinedDatabase(db_name);
```
serxa: Do we really need it here? It looks like something for tests in production code.Using fail points is okay because it's a known piece of code that usually useful only in tests, but here we do something more for calling this fail point alone.

serxa (changes requested):
Overall looks good to me.

src/Databases/DatabasesCommon.cpp
jaymebrd: Is this what you had in mind? https://github.com/ClickHouse/ClickHouse/pull/104969/changes/50aa143706de9d1496867e1d641fdbaf04047e15

serxa: yes

serxa (approved):

serxa:
@groeneai can you check what prevents merging and help?

src/Core/ServerSettings.cpp
```diff
@@ -439,6 +439,7 @@ namespace
     )", 0) \
     DECLARE(UInt64, database_catalog_drop_error_cooldown_sec, 5, R"(In case of a failed table drop, ClickHouse will wait for this time-out before retrying the operation.)", 0) \
     DECLARE(UInt64, database_catalog_drop_table_concurrency, 16, R"(The size of the threadpool used for dropping tables.)", 0) \
+    DECLARE(UInt64, database_catalog_shutdown_table_concurrency, 16, R"(The size of the threadpool used for shutting down tables when the server is stopping.)", 0) \
```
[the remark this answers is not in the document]
serxa: that's okay. It's just safety margins that are clear to any user.

groeneai:
@serxa diagnosis for #104969 (head `3d4725f0`): the only thing blocking merge is the **`CH Inc sync`** check (`tests failed`). It is the sole `failure` in the combined commit status, which is why `mergeStateStatus = BLOCKED`. It carries no public `target_url`, so external contributors cannot see its logs. Could you check it on the CH Inc side to tell whether it is a real internal-test failure or a transient sync-infra flake that just needs a re-run?

Everything else is green:
- `mergeable: MERGEABLE`, no conflicts (master merged in on 06-09).
- 152/152 public checks pass; `Mergeable Check`, `PR`, and `Style check` statuses are all green.
- All 4 review threads resolved. The two real bot findings are fixed: the `max_threads=0` hang (now clamped via `std::max<UInt64>(1, ...)` in `Server` + `LocalServer`) and the over-strict non-UUID `chassert` in `DatabaseWithOwnTablesBase::shutdown` (removed). The latest `NonZeroUInt64`-vs-document-`0` nit you already dismissed as acceptable safety margins.
- Your 2026-06-01 approval stands.

Nothing on the author's side looks blocking. Happy to dig further once the sync logs are visible.

groeneai:
Hi @serxa, gentle ping. PR #104969 is green on public CI and all review threads are resolved; the only thing still blocking merge is the internal `CH Inc sync` -> `Check public build` check (status `tests failed`, no public target_url), which only CH Inc can inspect or re-run. Could you check whether that is a real internal failure or a transient sync-infra flake? Thanks.

serxa:
@groeneai Let's merge master into this branch to take new fixes into account and into sync PR

groeneai:
@serxa branch is ~3.9k commits behind master but `mergeable` is MERGEABLE, so merging master in is conflict-free. I can't push it myself: @groeneai is an external contributor with no write access to `ClickHouse/ClickHouse`, and "Allow edits from maintainers" grants fork-branch push to CH Inc maintainers (you), not to external contributors.

@jaymebrd could you merge master into `parallel-table-shutdown`? It's conflict-free, so the PR "Update branch" button works. That refreshes the internal sync PR (currently pending) with current master. A CH Inc maintainer can also do it via the same button (maintainerCanModify is true).

serxa:
@groeneai check recent CI failures? do we have fixes for them? are they merged in master?

groeneai:
@serxa Triaged the 5 public failures at head `39a0842`. None are caused by this PR; all are pre-existing trunk flakiness unrelated to the shutdown-parallelism change. No fix is merged in master for either, because neither needs one on the author side.

**1. Integration tests (4 shards: asan_ubsan / msan / tsan / arm_binary) - `test_format_cannot_allocate_thread::test_it`**
All four fail identically inside `cluster.shutdown()`, not in the test body. It is a pre-existing data race on the global `getActivePartsLoadingThreadPool()` singleton (`src/IO/SharedThreadPools.cpp`): main-thread teardown (`StaticThreadPool::shutdown` via the `Server::main` scope guard, line 82) races a SystemLog flush thread that lazily creates a system table on the way down (`SystemLog<...>::prepareTable` -> `StorageMergeTree` ctor -> `loadDataParts` -> `StaticThreadPool::get()`, line 100). This has nothing to do with `DatabaseWithOwnTablesBase::shutdown` from this PR. CIDB confirms: 0 master failures in 30 days; it only surfaces in unrelated PRs (streaming-pool, profile-events, signal-handler work) where the deliberate `cannot_allocate_thread` failpoint perturbs teardown timing. Caused by this PR: no. Fix exists/merged: no. Re-running the jobs clears it.

**2. BuzzHouse (amd_debug) - `Inconsistent AST formatting` LOGICAL_ERROR (STID 1941-1bfa)**
A fuzzer-generated query using `caseWithExpression`; a generic parser/formatter roundtrip bug (the formatted SQL re-parses to a different AST, tripping the debug-only consistency check in `executeQueryImpl`). Unrelated to shutdown. The `Inconsistent AST formatting` class hits master roughly 1-3 times/day every day; this specific variant is fuzzer-only and does not reproduce deterministically. Caused by this PR: no. Fix merged: no.

**3. CH Inc sync** - internal check, no public target_url; only CH Inc can inspect or re-run (unchanged from prior runs).

Recommendation: nothing on the author side blocks merge. The shutdown code is not implicated by any of the five. Re-running BuzzHouse + the Integration shards should clear them; the only real gate remains the internal sync.

Public CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104969&sha=39a0842660da13a14f507feaa9d2f9f6a3edaf24&name_0=PR

serxa:
@groeneai Failed test test_format_cannot_allocate_thread/test.py::test_it looks suspicious. Is it related to the changes in this PR?

groeneai:
@serxa Yes, it looks related to this PR. I need to correct my 06-18 triage: I lumped it in with a pre-existing teardown race, but the hard signature on the current runs is specific to this branch.

Evidence:
- The crash is `LOGICAL_ERROR: The MergeTreePartsLoaderThreadPool is not initialized` on the non-TSan shards (asan/msan/coverage/arm_binary), plus a TSan data race on the same pool. CIDB: 11 hits in 90 days, all on this branch (`parallel-table-shutdown`), 0 on master and 0 on any other PR. The test is green on master (1532 OK / 0 FAIL in 30d). It first appears at head 39a0842 (the 06-18 master merge), not on the earlier 06-10 head.
- It fails in `cluster.shutdown()` (module teardown), not in the test body, so it is the server shutdown that aborts. `test_format_cannot_allocate_thread` is just the canary that happens to restart the node.
- Stack (identical on the hard-error shards and the TSan shard):
  `StaticThreadPool::get()` SharedThreadPools.cpp:100 <- `MergeTreeData::loadDataParts` MergeTreeData.cpp:2413 <- `StorageMergeTree::StorageMergeTree` StorageMergeTree.cpp:206 <- `make_shared<StorageMergeTree>` (registerStorageMergeTree.cpp:1079) <- a SystemLog flush thread lazily creating `system.background_schedule_pool_log` during shutdown.

Mechanism: a SystemLog flush thread lazily CREATES its backing MergeTree table during shutdown; the `StorageMergeTree` ctor calls `loadDataParts` -> `getActivePartsLoadingThreadPool().get()`. This PR moves `DatabaseWithOwnTablesBase::shutdown` onto a worker pool, which changes the timing and concurrency of the shutdown sequence so that this lazy table creation now overlaps with the global `ActivePartsLoading` static pool being torn down (`StaticThreadPool::shutdownAll()` in the `Server::main` SCOPE_EXIT). On non-TSan builds the pool is already reset -> hard `not initialized`; on TSan the read of the pool `instance` races with its reset. The serial shutdown on master does not hit this window.

So it is a latent shutdown-ordering fragility (system logs can lazily create their tables during shutdown and reach a torn-down static pool), and this PR's parallel shutdown is the first thing to surface it.

Report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104969&sha=9d7ff6e40da97cd60ee5769c279c2850e40d5991&name_0=PR

Possible directions: skip lazy system-log table creation when the catalog is shutting down (just drop the unflushed entries, since the table was never materialized), or make sure the new parallel-shutdown work is fully joined before system logs flush so `loadDataParts` never runs against a pool that is being reset. I can dig into the exact ordering and prepare a fix in a separate PR off master if you want it. Since this is jaymebrd's branch I will not push to it.

jaymebrd:
> @groeneai Failed test test_format_cannot_allocate_thread/test.py::test_it looks suspicious. Is it related to the changes in this PR?

It looks like this PR introduced the race https://github.com/ClickHouse/ClickHouse/pull/107291 and our parallel shutdown exposes it

serxa:
@groeneai Shutting down multiple interdependent pools is tricky. In easy cases, it could be resolved just by the proper ordering. Try to prepare the fix in a separate PR before we can merge this PR. The goal is to add necessary dependencies between startup and shutdown procedures, such that issue you found here is resolved. I hope you will not face circular dependencies; be careful with that (if nothing can prevent them, report back here)

groeneai:
@serxa Will do. Preparing the fix in a separate PR off master.

Plan: make SystemLogs teardown happen-before the static ActivePartsLoading pool teardown on all paths. Concretely, in `DatabaseCatalog::shutdownImpl` the `shutdown_system_logs()` call (which joins the system-log flush threads) is currently skipped if a user database `shutdown()` throws, so a flush thread can still lazily create its MergeTree backing table after `StaticThreadPool::shutdownAll()` reset the pool, giving the `MergeTreePartsLoaderThreadPool is not initialized` abort. Ordering the dependency (always tear down system logs before the global/static pools, even on a throwing path) closes that window.

No circular dependency expected: system logs depend on the static pools and on the `system` database (shut down last), not the other way around. Will report back here if a cycle shows up. This is a pre-existing latent issue on master that this PR's parallel shutdown exposes, so the fix belongs in its own master PR; I will link it here once it is up.

groeneai:
@serxa prepared the fix in a separate PR off master: #108417. No circular dependency. Two parts: `DatabaseCatalog::shutdownImpl` now keeps going when a database `shutdown()` throws (so `shutdown_system_logs()` always runs and joins the flush threads before the static pools are reset, which is the `MergeTreePartsLoaderThreadPool is not initialized` crash), and `DatabaseWithOwnTablesBase::shutdown()` releases its tables' references even on a throw so the storages are destroyed during shutdown instead of at atexit. Includes a regression test using a new failpoint. Once it merges, #104969 should be clean.

serxa:
@groeneai fix has been merged, update this PR please

groeneai:
@serxa #108417 is merged to master (commit e5a72943be091b42a739e3fc7caf9bc852e9944b, 2026-06-28 16:17 UTC), so the shutdown-crash fix is now on master.

Two things about updating this PR:

**1. The branch update has to be done by @jaymebrd or a CH Inc maintainer.** I'm an external contributor with no write access to jaymebrd's fork, and "Allow edits from maintainers" only grants push to CH Inc maintainers, not to other external contributors. `maintainerCanModify` is true, so a maintainer can do it too.

**2. It is now a real conflict, not a clean "Update branch".** #108417 touched the same two files this PR does:
- `src/Databases/DatabasesCommon.cpp`: both rewrite `DatabaseWithOwnTablesBase::shutdown()`.
- `src/Common/FailPoint.cpp`: both add a failpoint in the same `REGULAR(...)` block.

So merging master needs manual conflict resolution in those two files.

**3. On the clickhouse-gh[bot] shutdown-ordering finding:** the system-log-race half (the `MergeTreePartsLoaderThreadPool is not initialized` / TSan race when a `SystemLog` flush thread outlives the static-pool teardown) is now handled by #108417 at the `DatabaseCatalog::shutdownImpl` level: `database->shutdown()` is wrapped in try/catch and `shutdown_system_logs()` always runs, even if a user-DB shutdown throws or parallel scheduling fails. After master is merged in, that part of the finding is addressed without making the prepare phase no-throw.

One thing to carry forward when resolving the `DatabasesCommon.cpp` conflict, so the other crash (the atexit Poco `LoggerDeleter` abort) does not return: the parallel `shutdown()` should still release every table's UUID mapping and clear `tables` even when a table throws, rethrowing the first error only at the very end. The current parallel version rethrows from the first `waitForAllToFinishAndRethrowFirstError()`, which on a prepare-phase throw skips the shutdown phase, the `removeUUIDMapping` calls and `tables.clear()`, so a failed database's storages would survive to `~DatabaseCatalog` at process exit. Collecting the first error across both phases and clearing/rethrowing at the end keeps both the parallelism and that robustness.

jaymebrd:
@serxa I reworked and simplified this PR to only parallelise `shutdown()` because parallelising the prepare phase was causing some problems — tables engines like `Buffer` flush into other tables during                    
  `flushAndPrepareForShutdown`, and the AI review found running it concurrently could write into an already-read  destination and silently drop rows. 
  
Based on what motivates this PR, I've observed `shutdown()` being the bottleneck, and it's always table independent so I think that is both safe, and gives us the speedup.

serxa (approved):
The rework feels okay to me, but now I think we may need one more opinion on this.
@evillique PTAL, maybe we need something similar in SharedCatalog, or that's completely unrelated/orthogonal?

programs/local/LocalServer.cpp
```diff
@@ -416,6 +417,14 @@ void LocalServer::initialize(Poco::Util::Application & self)
         0, // We don't need any threads if there are no DROP queries.
         server_settings[ServerSetting::database_catalog_drop_table_concurrency]);
 
+    /// Clamp to >= 1: with 0 threads, the pool would accept tasks but never run them, so
+    /// DatabaseWithOwnTablesBase::shutdown would block forever waiting on them.
+    const size_t shutdown_concurrency = std::max<UInt64>(1, server_settings[ServerSetting::database_catalog_shutdown_table_concurrency]);
```
evillique: Should we instead use  `0` -> `getNumberOfCPUCoresToUse()`, we use it in some other places, and also mention it in the setting description?

For example, we use it in `max_database_replicated_create_table_thread_pool_size` and by default for every uninitialized thread pool in `StaticThreadPool::initializeWithDefaultSettingsIfNotInitialized`

tests/integration/test_parallel_table_shutdown/test.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # Pin a marker we can scan logs against to avoid false matches from any
+        # earlier startup activity. Truncate the file to start clean.
+        node.exec_in_container(["bash", "-c", ": > /var/log/clickhouse-server/clickhouse-server.log"])
+
+        # Soft shutdown — exercises the same Context::shutdown path that
+        # production SIGTERM hits, including DatabaseWithOwnTablesBase::shutdown.
+        shutdown_started = time.monotonic()
+        node.stop_clickhouse(kill=False, stop_wait_sec=60)
+        shutdown_elapsed = time.monotonic() - shutdown_started
```
evillique: I feel like under MSan/TSan it may take longer and the test may become flaky.

Maybe we should grep the `Shut down 12 tables in many_tables in M ms` log and measure it instead?

evillique (approved):

evillique:
@serxa, change is orthogonal to Shared Catalog, so will be used there by default as well.

Otherwise, I think it looks good.

programs/local/LocalServer.cpp
serxa: yes, make sense. I agree. And default 16 doesn't make sense because at shutdown we dont even have anything else to do, so using 0 by default instead of 16 is probably better. @jaymebrd, could you please add both updates?

src/Core/ServerSettings.cpp
```diff
@@ -478,7 +478,7 @@ namespace
     )", 0) \
     DECLARE(UInt64, database_catalog_drop_error_cooldown_sec, 5, R"(In case of a failed table drop, ClickHouse will wait for this time-out before retrying the operation.)", 0) \
     DECLARE(UInt64, database_catalog_drop_table_concurrency, 16, R"(The size of the threadpool used for dropping tables.)", 0) \
-    DECLARE(UInt64, database_catalog_shutdown_table_concurrency, 16, R"(The size of the threadpool used for shutting down tables when the server is stopping.)", 0) \
+    DECLARE(UInt64, database_catalog_shutdown_table_concurrency, 16, R"(The size of the threadpool used for shutting down tables when the server is stopping. Zero means the number of threads is equal to the number of cores.)", 0) \
```
serxa: lets use 0 by default

serxa (approved):

serxa:
@groeneai, investigate the failure: https://github.com/ClickHouse/ClickHouse/actions/runs/29755463902/job/88491825359 and provide a fix in a separate PR. If the fix is already in progress, link it here.

Note: job logs are no longer accessible via the GitHub API (artifact expired). Cloud benchmark report shows no regression (37 queries, no change). If this is a pre-existing or infrastructure flakiness in the performance comparison, please confirm.

groeneai:
Confirmed: infrastructure, not a performance regression, and the fix is already merged.

**What failed.** The GH Actions log is expired, but the check-run annotation survives (`gh api repos/ClickHouse/ClickHouse/check-runs/88491825359/annotations`) and it is unambiguous:

```
System.IO.IOException: No space left on device :
  '/home/ubuntu/actions-runner/_diag/Worker_20260720-223401-utc.log'
```

The runner ran out of disk. The job died before uploading anything, which is why the shard reports no rows to CIDB and there is no failing query to look at. All 3 attempts (jobs 88449175162, 88476564379, 88491825359) failed the same way on 3 *different* runners (`i-06fe86cfd8a0d58e4`, `i-0c6e3990ef5bd7176`, `i-035463a63bf003ac4`), so it was not one bad host.

Not a regression: at `18f21b4` the other 11 perf shards were all green with zero FAIL/ERROR rows, matching the 37-query no-change Cloud result you already have.

**Not specific to this PR.** Since a disk-death shard reports nothing, I counted incomplete shard sets instead (`uniqExact(check_name) < 6` per commit) on `Performance Comparison (amd_release, master_head, N/6)`:

| date | sets | incomplete |
|---|---|---|
| 07-18 | 51 | 1 |
| 07-19 | 42 | 0 |
| 07-20 | 49 | 26 |
| 07-21 | 47 | 44 |
| 07-22 | 55 | 34 |
| 07-23 | 62 | 1 |
| 07-24 | 55 | 0 |
| 07-25 | 32 | 0 |

A 3-day fleet-wide window, amd and arm together. I spot-checked other PRs in it (jobs 89032972175, 89041380548, 88991625264) and they carry the byte-identical ENOSPC annotation.

**Cause and fix.** The window opens with #110009 (merged 07-20T13:16Z), which added `populate_data`/`rebuild_table`: the hits datasets are re-inserted into fresh tables and `OPTIMIZE FINAL`-ed on both the left and right servers, roughly doubling live dataset bytes on top of the ~33.6 GiB of tarballs (`hits_100m_single` alone is 26.5 GiB). It closes with #111377 (rienath, merged 07-22T15:05Z), which adds `cleanup_user_files()` after each test and shrinks the two heaviest fixtures (`min_max_index` 15.2 -> 3.0 GB, `join_convert_to_fixed_hash_table` 27.4 -> 6.1 GB). Both are on master now (`performance_tests.py:1516`).

So there is nothing for me to fix in a separate PR: #111377 is the fix, and the rate has been 0-2% since it merged. A rerun of shard 4/6 on a rebased head should be green.

One note while I am here: the other red on this PR, `Stress test (amd_asan_ubsan)`, is also unrelated to the change. It is the `t_plt_rb_contains` startup-parse failure tracked in #111053. Empirically that signature stopped on master right after #111055, #111661 and #111730 merged (07-23/07-24) - Stress-check hits went 26 on 07-22, 18 on 07-23, 1 on 07-24, 0 since - and all three landed after this head was built, so a rebase should clear it.

clickgapai:
@jaymebrd @serxa @evillique — ClickGap found the following in this PR:

- **Bug (P2):** Shutdown drain calls bare shutdown(): Buffer flushed into a Buffer loses rows — https://github.com/ClickHouse/ClickHouse/issues/116823

Close anything that's wrong or already addressed.

PR #104975, opened 2026-05-14
feat(promql): Support prefixed /prometheus http_handlers URLs on the main HTTP port.

JTCunning:
This allows us to handle prom endpoints on the main http port, but adds the ability to use a prefix (like `/prometheus`) so that we aren't establishing a pattern of someone issuing queries to `clickhouse:8123/api/v1/*` and establishing more of an industry standard approach to issue queries like `clickhouse:8123/prometheus/api/v1/*`.

- Experimental Feature

Allow for attaching prometheus handlers on the main http port with an optional prefix.

### Version info
- Merged into: `26.6.1.1137`


src/Server/HTTPHandlerFactory.cpp
```diff
@@ -176,7 +176,11 @@ static inline auto createHandlersFactoryFromConfig(
             {
                 main_handler_factory->addHandler(createPredefinedHandlerFactory(server, config, prefix + "." + key, common_headers_override));
             }
-            else if (handler_type == "prometheus")
+            else if (handler_type == "prometheus"
+                || handler_type == "expose_metrics"
+                || handler_type == "remote_write"
+                || handler_type == "remote_read"
+                || handler_type == "query_api")
```
JTCunning: @vitlibar: Referring to an offline conversation, this is where I got confused. We register the `prometheus` handler type but none of the others.
I'm unsure if we should register every handler type as it exists today, or if we should just have a generic `prometheus` type that case-switches. That would allow us to end up with a single registered prefix, ie:
```xml
<handler>
  <type>prometheus</type>
  <prefix>/prometheus</table> //could be an implicit default
  <table>default.prometheus</table>
</handler>
```
I feel like this is what we should ultimately end up with since it cuts down on a lot of configuring in XML. I'm definitely willing to do it in a series of PRs instead of one big jump if that's how we should go about it.

vitlibar: I think the right solution here is to add another handler type `api_v1` here which handles all prometheus `/api/v1` protocols (`/api/v1/write`, `/api/v1/read`, `/api/v1/query`, `/api/v1/query_range`, etc.):

```
<http_port>8123</http_port>
<http_handlers>
    <my_rule_1>
        <url>/prometheus/api/v1</url>
        <handler>
            <type>prometheus.api_v1</type>
            <table>db.time_series_table_name</table>
        </handler>
    </my_rule_1>
</http_handlers>
```

I'll finish this PR

vitlibar:
Waiting for https://github.com/ClickHouse/ClickHouse/pull/107492

vitlibar (approved):

vitlibar:
Example:

config:
```
<http_port>8123</http_port>
<http_handlers>
    <my_rule_1>
        <url_prefix>/prometheus/api/v1</url_prefix>
        <handler>
            <type>prometheus_api_v1</type>
        </handler>
    </my_rule_1>
</http_handlers>
```

request:
```
https://clickhouse_host/prometheus/api/v1/query?query=up&time=1782212142&database=mydb&table=mytimeseriestable
```

Notes:
- `<url_prefix>` is important here, because just `<url>path</url>` means serving only the specified path
- type `prometheus_api_v1` means all `/api/v1` protocols will be served. Don't confuse it with just `prometheus` which is the same as `prometheus_metrics` and means that ClickHouse just exposes `/metrics`
- database and table name can be either specified in the URL or defined in the config as before

PR #104927, opened 2026-05-14
Allow KILL QUERY/MUTATION SELECT to return multiple blocks

groeneai:
Fixes the `Expected one block from input stream` `LOGICAL_ERROR` (STID 4569-3618) in `InterpreterKillQueryQuery::getSelectResult`.

The interpreter executes an internal `SELECT` over `system.processes` / `system.mutations` / `system.part_moves_between_shards` / `system.transactions` to find which queries / mutations to act on. The old code asserted that the resulting pipeline emits at most one block:

```cpp
while (res.empty() && executor.pull(res));
while (executor.pull(tmp_block));
if (!tmp_block.empty())
    throw Exception(ErrorCodes::LOGICAL_ERROR, "Expected one block from input stream");
```

That is wrong — the pipeline can legitimately produce several blocks. The AST fuzzer found a minimal reproducer (originally in @PedroTadim's fiddle https://fiddle.clickhouse.com/468380d6-18b1-4bea-be38-4186d9a9428f):

```sql
CREATE TABLE t0 (c0 UInt64, c1 UInt64) ENGINE = MergeTree ORDER BY c0;
INSERT INTO t0 SELECT number, number FROM numbers(5);
SYSTEM STOP MERGES t0;
ALTER TABLE t0 UPDATE c1 = c1 + 1 WHERE 1 SETTINGS mutations_sync = 0;
ALTER TABLE t0 UPDATE c1 = c1 + 2 WHERE 1 SETTINGS mutations_sync = 0;
KILL MUTATION WHERE (SELECT 1 WHERE database = command) OR command LIKE '%UPDATE%' TEST;
```

The per-row scalar subquery in `WHERE` causes the planner to emit one block per row of `system.mutations`. A small `max_block_size` from CI randomization produces the same shape. Either way the second `executor.pull` returns a non-empty block and the old code threw.

**Fix:** collect every produced block via `concatenateBlocks` from `Core/Block.h`. The result set is bounded by the size of the corresponding `system.*` table (small, in-process state), so this stays cheap. `materializeBlockInplace` is preserved so existing `typeid_cast<const ColumnString &>` callers in `execute` remain valid.

A regression test `tests/queries/0_stateless/04238_kill_mutation_subquery_multi_block.sql` pins `max_block_size = 1` and uses the per-row subquery shape, so the multi-block path is exercised deterministically.

Issue: https://github.com/ClickHouse/ClickHouse/issues/104857
[...]

Issue #104877, opened 2026-05-13
FunctionsStress.stress


nikitamikhaylov:
_Important: This issue was automatically generated and is used by CI for matching failures. DO NOT modify the body content. DO NOT remove labels._

Test name: FunctionsStress.stress
CI report: [Unit tests (tsan, function_prop_fuzzer)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104810&sha=bc8d00c65450830a72939b1cb96e0fe048072ca8&name_0=PR&name_1=Unit%20tests%20%28tsan%2C%20function_prop_fuzzer%29)
[...]

PR #104979, opened 2026-05-14
Fix LOGICAL_ERROR on system.detached_tables WHERE uuid filter

### LINKED ISSUE #102705: `getFilteredTables` detached path omits `uuid_column` population, causing `LOGICAL_ERROR` on `system.detached_tables WHERE uuid = ...`

_Retrospective finding from a historical scan of [PR #88379](https://github.com/ClickHouse/ClickHouse/pull/88379) (merged 2025-10-12). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

Querying `system.detached_tables` with a `WHERE uuid = '<valid-uuid>'` clause throws a `LOGICAL_ERROR` exception (Code 49): 'Expected the argument №1 (uuid of type UUID) to have N rows, but it has 0'

**Root cause:** StorageSystemTables.cpp:109-116: the `is_detached` branch only populates `table_column` but does not populate `uuid_column` (or `engine_column`), even though those columns may have been created at lines 96-99 based on the filter DAG analysis. The non-detached branch (lines 119-134) correctly populates both columns.

**Why we believe this is a bug:** StorageSystemDetachedTables.cpp:244 calls `detail::getFilteredTables(predicate, ..., true)` → StorageSystemTables.cpp:98-99 creates `uuid_column = ColumnUUID::create()` when the predicate references uuid → StorageSystemTables.cpp:109-116 `is_detached` path only inserts into `table_column`, NOT `uuid_column` → StorageSystemTables.cpp:149-153 builds a Block with mismatched column sizes (N rows in `table_column`, 0 rows in `uuid_column`) → VirtualColumnUtils.cpp:156 `filterBlockWithExpression` fails because the expression expects equal-length columns

**Affected locations:**
- `src/Storages/System/StorageSystemTables.cpp:109` — is_detached branch missing uuid_column population
- `src/Storages/System/StorageSystemDetachedTables.cpp:244` — caller passing is_detached=true

**Impact:** `system.detached_tables` is completely broken when queried with a uuid filter. Any query like `SELECT * FROM system.detached_tables WHERE uuid = '...'` throws a `LOGICAL_ERROR` exception instead of returning results. This affects monitoring tools and scripts that look up detached tables by UUID.

Yes — confirmed on current `master` (commit `9a3964e0590a`).

### How to reproduce

```sql
CREATE TABLE t (key Int) ENGINE = MergeTree ORDER BY key;
SELECT uuid FROM system.tables WHERE database = currentDatabase() AND name = 't';
-- copy the uuid
DETACH TABLE t;
SELECT count() FROM system.detached_tables WHERE database = currentDatabase() AND uuid = '<paste-uuid-here>';
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/ccf6f7bd-653d-4934-aeff-f5b923dd61dd)

### Expected behavior

```
1
```

```
Code: 49. DB::Exception: Received from 127.0.0.1:19030. DB::Exception: Expected the argument №1 ('uuid' of type UUID) to have 1 rows, but it has 0: while executing 'FUNCTION equals(uuid :: 0, '<uuid>'_String :: 1) -> equals(uuid, '<uuid>'_String) UInt8 : 2'. (LOGICAL_ERROR)
```

### Additional context

**Open risks:**
- The same pattern could affect engine_column in the is_detached path, though system.detached_tables does not have an engine column, so it's not currently triggerable via engine filter

**Suggested fix:** Add `uuid_column` and `engine_column` population in the `is_detached` branch. The detached table iterator provides uuid via `table_it->uuid()` method (already used at StorageSystemDetachedTables.cpp:137). Add: `if (uuid_column) uuid_column->insert(table_it->uuid());` inside the detached loop.

Found during automated review of [PR #88379](https://github.com/ClickHouse/ClickHouse/pull/88379).

---
_ClickGapAI · Confidence: HIGH · Severity: P0 · Finding: `h_pr88379_001`_

groeneai:
Closes #102705.

### What is the bug

Filtering `system.detached_tables` by `uuid` raises a `LOGICAL_ERROR` and aborts in debug builds:

```sql
CREATE TABLE t (key Int) ENGINE = MergeTree ORDER BY key;
DETACH TABLE t;

SELECT count() FROM system.detached_tables
WHERE database = currentDatabase() AND uuid = '<some-uuid>';
-- Code: 49. DB::Exception: Expected the argument №1 ('uuid' of type UUID)
-- to have 1 rows, but it has 0: while executing 'FUNCTION equals(...)'.
-- (LOGICAL_ERROR)
```

### Root cause

`detail::getFilteredTables` is shared between `system.tables` and `system.detached_tables`. When the WHERE clause references the `uuid` column, the function creates a `uuid_column` to feed the filter expression. The non-detached branch fills `uuid_column` from `table->getStorageID().uuid`, but the `is_detached` branch only filled `table_column`. After the loop, both columns are inserted into the same `Block`:

```
table_column : N rows
uuid_column  : 0 rows
```

`VirtualColumnUtils::filterBlockWithExpression` then executes `equals(uuid, '<literal>')`, and `checkFunctionArgumentSizes` raises `LOGICAL_ERROR` because the columns have mismatched lengths.

### Fix

Populate `uuid_column` from the detached iterator's `uuid()` so the columns stay in lockstep. `DatabaseDetachedTablesSnapshotIterator` already exposes `uuid()` (it is used elsewhere in `StorageSystemDetachedTables.cpp`).

`engine_column` cannot currently be created on the detached path — `system.detached_tables` has no `engine` column, so the predicate cannot reference it — but a defensive `engine_column->insertDefault()` is added so the loop stays correct if that schema is ever extended.

### Reproduction

The query above crashed the local debug build before this change; with the fix it returns the expected count and the server stays up. The new regression test exercises three failing query shapes (lookup by uuid via subquery, by direct uuid comparison after detach, and uuid-only filter).

- [x] Documentation entry: not required (bug fix in a system table; no user-facing API change).

Fixed a `LOGICAL_ERROR` thrown when filtering `system.detached_tables` by `uuid` (e.g. `SELECT count() FROM system.detached_tables WHERE uuid = '...'`). The query now returns the expected result instead of aborting the server.

### Version info
- Merged into: `26.5.1.749`


groeneai:
## Pre-PR validation gate

a. **Deterministic repro?** Yes. `CREATE TABLE t (key Int) ENGINE = MergeTree ORDER BY key; DETACH TABLE t; SELECT count() FROM system.detached_tables WHERE database = currentDatabase() AND uuid = '<uuid>';` raises a `LOGICAL_ERROR` every time on master (debug build aborts).

b. **Root cause explained?** Yes. `getFilteredTables` constructs an empty `uuid_column` whenever the WHERE predicate references `uuid`. The `is_detached` branch (`src/Storages/System/StorageSystemTables.cpp:112-119`) only pushed rows into `table_column`. After the loop, both columns are inserted into the same `Block` (lines 153-156), so `table_column` had N rows and `uuid_column` had 0. `VirtualColumnUtils::filterBlockWithExpression` then evaluates `equals(uuid, '<literal>')` and `checkFunctionArgumentSizes` throws.

c. **Fix matches root cause?** Yes. Populate `uuid_column` from `table_it->uuid()` in the detached branch — mirroring the non-detached branch — so the columns stay in lockstep. The defensive `engine_column->insertDefault()` is a future-proof guard: today `system.detached_tables` has no `engine` column, so `filter_by_engine` cannot be true on the detached path, but the loop now stays correct if the schema is extended.

d. **Test intent preserved / new test added?** New regression test `04240_detached_tables_uuid_filter_102705.sql` exercises three failing query shapes: lookup-by-uuid via subquery, direct uuid comparison after detach, and uuid-only filter. No existing tests were weakened.

e. **Both directions demonstrated?** Yes. Original repro crashed my local debug build with the exact stack trace from the issue (`getFilteredTables` → `filterBlockWithExpression` → `checkFunctionArgumentSizes`). After applying the fix and rebuilding, the same queries return correct results and the test passes 20/20 with full randomization.

f. **Fix is general, not a narrow patch?** Yes. The patch fixes the column-mismatch invariant at its source by always populating `uuid_column` whenever it exists, not by silencing the assertion. The same bug pattern does not appear in sibling code paths: the non-detached branch (lines 119-138) already populates both `engine_column` and `uuid_column` correctly, and no other caller produces a `Block` that mixes filled and empty columns before `filterBlockWithExpression`.

Session: `cron:clickhouse-ci-task-worker:20260514-194500`

groeneai:
cc @azat @alexey-milovidov — could you review? Trivial fix for #102705 (`system.detached_tables WHERE uuid = ...` raised `LOGICAL_ERROR` because `getFilteredTables` left `uuid_column` empty on the detached path). Two-line code change + regression test.

src/Storages/System/StorageSystemTables.cpp
```diff
@@ -115,6 +115,16 @@ ColumnPtr getFilteredTables(
...
                 table_column->insert(table_it->table());
+                /// Keep all auxiliary columns in lockstep with `table_column`, otherwise
+                /// `filterBlockWithExpression` below raises a LOGICAL_ERROR on mismatched
+                /// column lengths (issue #102705).
+                if (uuid_column)
+                    uuid_column->insert(table_it->uuid());
+                /// `system.detached_tables` has no `engine` column, so `filter_by_engine`
+                /// is never true for the detached path in practice. Insert a default value
+                /// defensively so the columns stay aligned if that ever changes.
+                if (engine_column)
```
azat: Shouldn't we just not create `engine_column` above in case of `is_detached`?
Also it does not make sense to write comments for such a simple fix

groeneai: Done in 6e10f2f859e — `engine_column` is now never created on the detached path (guarded by `!is_detached`), and the comments are dropped. `filter_by_engine && is_detached` is logically unreachable anyway since `system.detached_tables` has no `engine` column and the analyzer rejects the predicate with `UNKNOWN_IDENTIFIER` before reaching `getFilteredTables`.

azat (approved):

groeneai:
Pushed `d17571c4afb` to address the `Bugfix validation (functional tests)` failure.

**Diagnosis:** The regression test in its previous `.sql` form fired the `chassert` inside the long-running test server, the runner's hung-check tore everything down before any `[ FAIL ]` was recorded, and the inversion loop in `Bugfix validation` therefore could not see the master-side failure.

**Fix:** Converted the test to `.sh` and run the reproducer inside `clickhouse local` with an `Atomic` database under a per-test `--path`. The `SIGABRT` is now contained to a short-lived subprocess: the runner observes truncated stdout, the diff against the `.reference` fails, and `Bugfix validation` can correctly invert that to `OK` on the PR branch. Same pattern as `04238_kill_mutation_subquery_multi_block.sh` (#104927).

**Local verification (debug build):**
- With this PR's source-side fix → output `1\nok` matches reference, 30 / 30 deterministic.
- After reverting only line 118–119 of `StorageSystemTables.cpp` and rebuilding → `Aborted (core dumped)` with truncated stdout, diff mismatches reference, runner reports `FAIL` — exactly the shape `Bugfix validation` needs to see.

The other red check, `Unit tests (tsan, function_prop_fuzzer)` on `FunctionsStress.stress` (1900-01-01 datetime), is the chronic FunctionsStress family tracked under issue #104877 — unrelated to this PR. Cross-PR CIDB confirms it hits master and many unrelated PRs.

Session: cron:clickhouse-ci-task-worker:20260515-174500

clickgapai:
Hi @groeneai @azat — the changelog category for this PR might need a look.

**Current:** `Bug Fix`

**Why:** LOGICAL_ERROR exceptions are internal assertion failures (chassert in debug builds) that indicate contract violations. This category applies when the server detects an internal inconsistency. The PR description confirms this is Code 49 LOGICAL_ERROR.

clickgapai:

**Affected code:** `detail::getFilteredTables` in src/Storages/System/StorageSystemTables.cpp — detached tables branch (lines 112-121) — core code present in all supported branches.

**Why this might apply:** The bug was introduced in commit 51d293beb9d (2024-06-28) when `getFilteredTables` was refactored to share logic between `system.tables` and `system.detached_tables`. This predates all supported release branches. The fix is safe (adds missing column population) and affects a user-visible system table query path.

PR #99997, opened 2026-03-19
Implements SQL truth-value predicates `IS TRUE`, `IS FALSE`, `IS UNKNOWN` and their `IS NOT` variants. Closes ClickHouse/ClickHouse#99597.

ibrahimkarimeddin:
Fixes #99597

### Changelog category (leave one):
- New Feature

### Changelog entry:
Added support for SQL truth-value predicates `IS TRUE`, `IS FALSE`, `IS UNKNOWN`, `IS NOT TRUE`, `IS NOT FALSE`, and `IS NOT UNKNOWN` for nullable boolean expressions. 

### Documentation entry for user-facing changes
**Motivation:** Allows unambiguous testing of nullable boolean columns for all three SQL truth values (true, false, null/unknown). Unlike `= true`, the `IS TRUE` predicate returns `false` (not `NULL`) when the value is `NULL`.
**Example use:**
```sql
-- Scalar
SELECT 1 WHERE NULL IS UNKNOWN;  -- returns 1
SELECT 1 WHERE NULL IS TRUE;     -- returns nothing
-- On a nullable column
SELECT
    x IS TRUE,
    x IS FALSE,
    x IS UNKNOWN
FROM my_table;

<!-- ch-version-info:start -->
### Version info
- Merged into: `26.5.1.864`
<!-- ch-version-info:end -->

PR #104984, opened 2026-05-14
CI: Add SQLancer++ check

fm4v:
Wires a new CI job that runs [SQLancer++](https://github.com/suyZhong/SQLancerPlusPlus) against `clickhouse-server`, exercising its four `general`-provider oracles for 600 s each (~40 min budget, well inside the 3600 s Job.Config timeout). The job runs in the scheduled `NightlySQLancer` workflow (every 3 days) alongside SQLancer, sharing that workflow's `arm_asan_ubsan` build — so it also runs against a sanitizer build.

SQLancer++ is the ASPLOS '26 successor to upstream SQLancer that ships a single DBMS-agnostic `general` provider with built-in ClickHouse support. Pure-runtime: this PR doesn't fork the upstream tree, it patches three files inside the docker image so the SQLancer++ revision pin is the only thing we maintain.

### What gets patched in the image

The pre-build hook (`ci/docker/sqlancer-pp-test/swap-clickhouse-jdbc.py`) rewrites three files in the SQLancer++ checkout before `mvn package`:

1. **`pom.xml`** — replace the legacy `ru.yandex.clickhouse:clickhouse-jdbc:0.3.2` dependency (its HTTP response-stream parser cannot read ClickHouse ≥ 25.x output — every query fails with `Failed to create stream response`) with the latest `com.clickhouse:clickhouse-jdbc` shaded-all artifact, version resolved from Maven Central's `maven-metadata.xml` at build time.
2. **`dbconfigs/jdbc.properties`** — extend `CLICKHOUSE.url` to embed `?user={user}&password={password}` and seed `CLICKHOUSE.user=sqlancer`/`CLICKHOUSE.password=sqlancer`. The modern driver rejects URLs with empty `password=` parameters, and SQLancer++'s `GeneralJdbcConfigLoader.getUser` only honours `--username`/`--password` when they *differ* from the built-in defaults, so the properties file needs real values.
3. **`src/sqlancer/general/oracle/GeneralNoRECOracle.java`** — rewrite the `IS TRUE` postfix to `!= 0`. ClickHouse's parser doesn't implement the SQL-standard `IS TRUE`/`IS FALSE` postfix operators, so the unpatched NoREC oracle produces 100 % `SYNTAX_ERROR` against ClickHouse and reports zero useful work. `<expr> != 0` reproduces `WHERE <expr>` semantics exactly for the row-counter (non-zero numerics → 1, zero → 0, NULL → NULL → JDBC counts as 0). A naive `= TRUE` substitution was tried first and rejected because it caused false-positive oracle mismatches on `WHERE LN(x)`-style truthy checks (`LN(x) = TRUE` is strict equality with `1`).

### Validation

Roughly 150 k oracle queries over 100 minutes against ClickHouse 26.4 ASan locally, two independent soaks plus a separate NoREC focused run after the parser fix:

| Oracle | Soak v1 (10 min) | Soak v2 (15 min) | NoREC re-run (after `!= 0` patch) |
|---|---|---|---|
| WHERE | 15 379 q · 79 % stmt OK · 0 asserts | 29 582 q · 79 % stmt OK · 0 asserts | — |
| NoREC | 0 q · 7 % stmt OK · 0 asserts | 0 q · 4 % stmt OK · 0 asserts | **3 221 q · 60 % stmt OK · 0 asserts** |
| QUERY_PARTITIONING | 11 907 q · 79 % stmt OK · **3 asserts (Int64.MIN)** | 26 421 q · 82 % stmt OK · 0 asserts | — |
| FUZZING | 36 761 q · 53 % stmt OK · 0 asserts | 22 606 q · 52 % stmt OK · 4 distinct potential-bug queries | — |

### Wiring

* `ci/defs/defs.py` — register `clickhouse/sqlancer-pp-test` docker image and `JobNames.SQLANCER_PP = "SQLancer++"`.
* `ci/defs/job_configs.py` — `JobConfigs.sqlancer_pp_jobs` parametrised for `arm_asan_ubsan` (requires `CH_ARM_ASAN_UBSAN`, runs on `FUNC_TESTER_ARM`, `clickhouse/sqlancer-pp-test`, 3600 s timeout).
* `ci/workflows/nightly_sqlancer.py` — added alongside `SQLancer`; both share the workflow's single `arm_asan_ubsan` build and run in parallel after it.
* `.github/workflows/nightly_sqlancer.yml` — regenerated by `python3 -m praktika yaml`.

The `arm_asan_ubsan` CI path has not yet been proven by a real Praktika run (the scheduled workflow can't run on a PR); it can be validated on demand by temporarily wiring the job into the PR workflow, or after merge via the nightly cron / `workflow_dispatch`. SQLancer++ was validated locally against ClickHouse 26.4 ASan (see below).

...

### Version info
- Merged into: `26.7.1.356` (included in `26.7` and later)


alexey-milovidov:
Depends on #99997.

Issue #72578, opened 2024-11-28
This function cannot be used to construct empty tuple. It is a bug


devcrafter:
**Describe the bug**
https://s3.amazonaws.com/clickhouse-test-reports/72109/dbf217f32a7b27cf8e9566d7469956275366b39a/ast_fuzzer__debug_/fatal.log

**How to reproduce**
```
SELECT arrayIntersect([tuple()])
```

<details>
<summary>Stack trace</summary>

```text
2024.11.28 00:32:09.820947 [ 1841 ] {} <Fatal> BaseDaemon: (version 24.12.1.444, build id: 10EE011CFD75EDE59BE40538017119F83AE03EDD, git hash: e651fcc80a1b537d6f3020c906307597918539bf, architecture: x86_64) (from thread 188) Received signal 6
2024.11.28 00:32:09.820964 [ 1841 ] {} <Fatal> BaseDaemon: Signal description: Aborted
2024.11.28 00:32:09.820977 [ 1841 ] {} <Fatal> BaseDaemon: 
2024.11.28 00:32:09.820996 [ 1841 ] {} <Fatal> BaseDaemon: Stack trace: 0x000000000caef688 0x000000000cd3bd74 0x00007fb2be4e8520 0x00007fb2be53c9fd 0x00007fb2be4e8476 0x00007fb2be4ce7f3 0x000000000cabf133 0x000000000cabfb68 0x00000000073774ec 0x000000000737a44b 0x00000000130071b0 0x000000000c657263 0x000000000c657021 0x000000000c64f393 0x000000000737668a 0x0000000010b42a3a 0x0000000010b3e48f 0x0000000010b3ef3a 0x0000000010b3fe65 0x0000000011a50b61 0x0000000011a4fb92 0x0000000014097332 0x0000000014240b92 0x0000000010f0ccda 0x00000000123f109c 0x00000000123eb54a 0x00000000123e6a5e 0x00000000123e4b6d 0x00000000126bf17b 0x00000000126bb55e 0x0000000013d92671 0x0000000013da8478 0x00000000176d8607 0x00000000176d8ade 0x0000000017680dd2 0x000000001767eac3 0x00007fb2be53aac3 0x00007fb2be5cc850
2024.11.28 00:32:09.821015 [ 1841 ] {} <Fatal> BaseDaemon: ########################################
[...]

PR #104861, opened 2026-05-13
Fix `LOGICAL_ERROR` in partition pruning when chain collapses to constant

groeneai:
Fix `LOGICAL_ERROR` ("Bad cast from type `DB::ColumnConst` to `DB::ColumnNullable`") in partition pruning for `MergeTree` tables when the partition expression's monotonic function chain collapses to a single constant value.

### Root cause

`KeyCondition::applyFunctionChainToColumn` strips outer `LowCardinality` from the result of `func->execute` but not outer `Const`. When the chain collapses to a constant (e.g. `floor(NULL, x)` always yields NULL), `func->execute` returns `ColumnConst(ColumnNullable(...))` of size 1.

`ColumnConst::isNullable` reports the wrapped column's nullability, so the existing `isNullable()` guard accepts the column while the subsequent `assert_cast<const ColumnNullable &>` aborts the server.

Reproducer (`floor(NULL, toRelativeYearNum(myDay))` partition key — taken from the fuzzer report in #104840):

```sql
CREATE TABLE t (myDay Date NULL, myOrder DateTime('UTC'))
ENGINE = MergeTree
PARTITION BY floor(NULL, toRelativeYearNum(myDay))
ORDER BY myOrder
SETTINGS allow_nullable_key = 1, allow_suspicious_indices = 1;

INSERT INTO t VALUES ('2021-01-02', 1);
SELECT * FROM t WHERE myDay = '2021-01-02';
```

Before this PR the `SELECT` aborts with `Bad cast from type DB::ColumnConst to DB::ColumnNullable`; after this PR it returns the matching row.

### Fix

Strip the outer `Const` right after `convertToFullColumnIfLowCardinality` so the column shape stays in sync with `result_type` and with the post-strip invariant established at the top of the function (which already strips `Const`, `Replicated`, `Sparse`, and `LowCardinality`).

Apply the same defensive unwrap to the sibling `assert_cast` site after `castColumnAccurateOrNull` — `FunctionCast` wrappers can also return `ColumnConst` for 1-row inputs by the same mechanism. This prevents a class of fuzzer findings rather than just the one reported.

A new stateless regression test `04237_104840_partition_pruner_floor_null_nullable_key` aborts the server on master and passes after the fix.

Closes: https://github.com/ClickHouse/ClickHouse/issues/104840

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

[...]

PR #104989, opened 2026-05-14
Fix LOGICAL_ERROR in flattenTuple for inputs with no scalar leaves

groeneai:
`flattenTuple` recursively expands nested tuples and arrays into the
underlying scalar columns. When the input type tree contained only
empty `Tuple()` leaves (directly, or wrapped in any combination of
`Tuple` / `Array`), the recursion produced zero scalar columns, the
result reached `ColumnTuple::create` with an empty column list, and the
defensive `LOGICAL_ERROR` guard in `ColumnTuple::create` aborted the
server in debug builds.

The check at the top of `FunctionFlattenTuple::getReturnTypeImpl`
already rejects an empty top-level tuple, but it does not catch inputs
whose flatten result happens to be empty, e.g.
`Tuple(c0 Array(Tuple()))`, `Tuple(c0 Tuple())`,
`Tuple(c0 Tuple(c00 Tuple()))`, or `Tuple(c0 Array(Array(Tuple())))`,
including the same shapes wrapped in `Nullable`.

After this change, `getReturnTypeImpl` runs the flatten on the input
type and validates that the result has at least one element; otherwise
it throws `ILLEGAL_TYPE_OF_ARGUMENT` with a user-readable message that
names the offending input type. Reachable shapes that already have at
least one scalar leaf (e.g. `Tuple(c0 Int32, c1 Array(Tuple()))`) are
unchanged.

Routed by @alexey-milovidov from the CI thread of
https://github.com/ClickHouse/ClickHouse/pull/100272 as a
`function_prop_fuzzer` finding. Same class as previously-fixed empty
tuple constructions in `reverse` (#89137) and `arrayIntersect`
(#72578), but for the `flattenTuple` code path which had no equivalent
guard.

`flattenTuple` no longer raises a `LOGICAL_ERROR` when called on a tuple whose nested structure contains only empty `Tuple()` leaves (e.g. `Tuple(c0 Array(Tuple()))` or `Tuple(c0 Tuple())`). Such inputs now produce a user-facing `ILLEGAL_TYPE_OF_ARGUMENT` exception explaining that the flatten result would be an empty tuple.

Session: cron:clickhouse-ci-task-worker:20260514-231500

### Version info
- Merged into: `26.5.1.755`


groeneai:
**Pre-PR validation self-assessment (TASK.md Phase 4 Step 9)**

(a) **Deterministic repro?** Yes. `SELECT flattenTuple(CAST((tuple([])) AS Tuple(c0 Array(Tuple()))));` deterministically aborts the server with `Logical error: 'This function cannot be used to construct empty tuple. It is a bug'.` on `master` (verified locally; debug build, SIGABRT, stack ends at `ColumnTuple::create` called from `flattenTuple` in `src/DataTypes/flattenTuple.cpp:99`).

(b) **Root cause explained?** Yes. `flattenTuple` walks the input type tree expanding nested tuples and arrays into the scalar leaves. When the input tree contains only empty `Tuple()` leaves, the recursion completes with zero scalar columns. `ColumnPtr flattenTuple(...)` then calls `ColumnTuple::create({})`, which throws `LOGICAL_ERROR` because the defensive guard rejects empty tuple construction. The existing top-level check rejects `Tuple()` directly but not inputs whose flatten *result* happens to be empty.

(c) **Fix matches root cause?** Yes. `getReturnTypeImpl` now runs the flatten on the input type and validates the result has at least one element, throwing `ILLEGAL_TYPE_OF_ARGUMENT` upstream of the column-flattening path that would otherwise hit the `LOGICAL_ERROR`. The fix targets the real mechanism — the empty flatten result — rather than masking the symptom. The bad state (empty result) is detected at the exact boundary where the function rejects malformed input types; this is the same place the existing top-level emptiness check already lives.

(d) **Test intent preserved? / New test added?** New regression test `04240_flatten_tuple_empty_result_104877.sql` covers every reproduced crash variant:
  - `Tuple()` direct (covered by the existing check)
  - `Tuple(c0 Array(Tuple()))` (original fuzzer-reduced case)
  - `Tuple(c0 Tuple())`
  - `Tuple(c0 Tuple(c00 Tuple()))`
  - `Tuple(c0 Array(Array(Tuple())))`
  - The same shapes wrapped in `Nullable` (with `allow_experimental_nullable_tuple_type = 1`)

  It also re-asserts that valid usage still flattens: `Tuple(c0 Int32)`, the doc example `Tuple(a UInt32, b Tuple(c String, d UInt32))`, and the mixed case `Tuple(c0 Int32, c1 Array(Tuple()))` (which has one scalar leaf, so the flatten succeeds and the empty-tuple branch contributes nothing). No existing test has been weakened.

(e) **Demonstrated in both directions?** Yes.
  - Without the fix (master flattenTuple.cpp, stashed locally then rebuilt + restarted): the test triggers `Fatal: 'This function cannot be used to construct empty tuple. It is a bug'`, the server aborts, and `clickhouse-test` returns SIGTERM (exit 143). Stack confirmed in `tmp/ch-server/log/error.log`.
  - With the fix: `clickhouse-test 04240_flatten_tuple_empty_result_104877 --test-runs 5` → 5/5 passes. Existing tests (`00897_flatten`, `02246_flatten_tuple`, `02906_flatten_only_true_nested`) also pass.

(f) **Fix is general, not a narrow patch?**
  - *"Does the same bug pattern exist elsewhere?"* The `flattenTuple` C++ function has only one caller — `FunctionFlattenTuple` (grep confirmed). The sibling empty-tuple bugs that existed historically (`reverse` #89137, `arrayIntersect` #72578, the `transform` cast wrapper) were each fixed in their own caller; this PR closes the equivalent gap for `flattenTuple` at its only caller.
  - *"Am I fixing the root cause or just guarding a symptom?"* The bad state (`flattened_columns.empty()`) arises because the user-provided input type tree has no scalar leaves. This is a legitimate user-input case — the type tree is constructed from user SQL — so surfacing it as `ILLEGAL_TYPE_OF_ARGUMENT` (and not as `LOGICAL_ERROR`) is the correct semantics, not a defensive workaround.

groeneai:
cc @alexey-milovidov @rschu1ze — could you review? Closes the `function_prop_fuzzer` finding you routed via the CI thread of #100272. Adds a user-facing `ILLEGAL_TYPE_OF_ARGUMENT` for inputs whose flatten result would be an empty tuple, instead of the current `LOGICAL_ERROR`. Regression test attached.

antonio2368 (approved):

groeneai:
Pushed `ebc756f` to address the **Bugfix validation (functional tests)** failure ("Failed to reproduce the bug").

**Root cause.** The previous `.sql` form used `{ serverError ILLEGAL_TYPE_OF_ARGUMENT }` directives. On master HEAD (without the fix), the bug-triggering queries reach the `LOGICAL_ERROR` guard in `ColumnTuple::create` and abort the server under `abort_on_logical_error = true` (the default in CI test environments — see `tests/config/config.d/abort_on_logical_error.yaml`). The bugfix-validation framework saw the parent server crash as `ERROR` / `SERVER_DIED` and reported "Failed to reproduce the bug" because that result is not invertible (see [`ci/jobs/functional_tests.py:750-767`](https://github.com/ClickHouse/ClickHouse/blob/master/ci/jobs/functional_tests.py#L750-L767) — only `Status.FAIL` flips to `OK` and counts toward `has_failure`).

**Fix.** Convert the test to a `.sh` script that runs each bug-triggering query inside a `clickhouse-local` subprocess. The abort on master HEAD now stays contained: `clickhouse-local` returns a non-43 exit code, the script prints `BUG`, the output diff fails, and the bugfix-validation framework can invert that `FAIL` into a "bug reproduced" signal.

- On the **PR branch** the queries return `Code: 43` (`ILLEGAL_TYPE_OF_ARGUMENT`) cleanly via `getReturnTypeImpl` and the script prints `OK` × 6.
- On **master HEAD** the queries cause `SIGABRT` (rc=134) and the script prints `BUG` × 6.

Verified locally against the master binary: 6 `BUG` outputs as expected.

```
$ PATH=build/programs:$PATH tests/queries/0_stateless/04240_flatten_tuple_empty_result_104877.sh
BUG
BUG
BUG
BUG
BUG
BUG
```

This mirrors the wrapper pattern used in PR #104861 for the partition-pruner `Bad cast` regression test, where the same `SERVER_DIED`-vs-FAIL classification problem applies to any abort-class crash.

alexey-milovidov:

antonio2368:
@groeneai please merge with master

groeneai:
@antonio2368 Done — merged `master` into the branch and pushed.

- Merge commit: `8445673eb1a4` (no conflicts)
- Branch advanced 202 commits to incorporate latest master
- Fix commits `04a50116e2d5` + `ebc756f14d50` intact
- PR state: OPEN, MERGEABLE
- CI re-running against the merged HEAD

Ready to merge once the new CI run is green. Thanks for the review!

— Session: cron:clickhouse-ci-task-worker:20260518-110000

Issue #104605, opened 2026-05-11
More inconsistent AST formatting


PedroTadim:
### Describe the bug

Very minor cc @george-larionov @alexey-milovidov 

### How to reproduce

Run:
```sql
SELECT substring(x, `x` -> `x`);
```

### Error message and/or stacktrace

```
[ip-10-2-3-131] 2026.05.11 15:29:12.169524 [ 819339 ] {8e529967-f4d4-4921-8246-6074e9e8d179} <Fatal> : Logical error: 'Inconsistent AST formatting: the query:
SELECT substring((x, x) -> x)
cannot parse query back from SELECT substring(x, `x` -> `x`);'.
[ip-10-2-3-131] 2026.05.11 15:29:12.169595 [ 819339 ] {8e529967-f4d4-4921-8246-6074e9e8d179} <Fatal> : Format string: 'Inconsistent AST formatting: the query:
{}
cannot parse query back from {}'.
[ip-10-2-3-131] 2026.05.11 15:29:12.192483 [ 819339 ] {8e529967-f4d4-4921-8246-6074e9e8d179} <Fatal> : Stack trace (when copying this message, always include the lines below):

0. /ClickHouse/contrib/llvm-project/libcxx/include/__exception/exception.h:115:14: Poco::Exception::Exception(String&&, int) @ 0x00000000251497b3
1. /ClickHouse/src/Common/Exception.cpp:139:7: DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0x0000000013e7c269
2. /ClickHouse/src/Common/Exception.h:171:100: DB::Exception::Exception(String&&, int, String, bool) @ 0x000000000cbcfb94
3. /ClickHouse/src/Common/Exception.h:57:54: DB::Exception::Exception(PreformattedMessage&&, int) @ 0x000000000cbcf626
4. /ClickHouse/src/Common/Exception.h:189:77: DB::Exception::Exception<String&, std::basic_string_view<char, std::char_traits<char>>&>(int, FormatStringHelperImpl<std::type_identity<String&>::type, std::type_identity<std::basic_string_view<char, std::char_traits<char>>&>::type>, String&, std::basic_string_view<char, std::char_traits<char>>&) @ 0x000000001998dc06
[...]

PR #104991, opened 2026-05-15
Fix inconsistent AST formatting for CODEC with tuple-wrapped multi-arg unary functions (STID 1941-1bfa)

groeneai:
Queries like the following abort the server with `Logical error: 'Inconsistent AST formatting between 'ExpressionList' and 'ExpressionList'` (STID 1941-1bfa) in debug / sanitiser builds, because the format-parse-format round-trip check in `DB::executeQueryImpl` diverges by one paren level around the inner `tuple(...)`:

```sql
CREATE TABLE t (c0 Int CODEC(not((not(materialize(1), materialize(2)))), ZSTD)) ENGINE = Memory;
```

The first format pass emits `CODEC(not((not(tuple(materialize(1), materialize(2))))), ZSTD)`. The re-parse sees `tuple(...)` wrapped in `(...)` (from the outer `not(...)` parenthesised argument); `RoundBracketsLayer::getResultImpl` in `src/Parsers/ExpressionListParsers.cpp` sets `parenthesized = true` on the unwrapped inner `Function_tuple`. On the second format pass the `parenthesized` flag adds an extra `(...)` around `tuple(...)` — `not((not((tuple(materialize(1), materialize(2))))))` — and the round-trip check fires.

Inside `CODEC` / `STATISTICS` / `BACKUP_NAME` argument lists, `ASTFunction::formatImpl` forces `FormatStateStacked::allow_operators = false`, so a multi-argument `Function_tuple` falls back to its function-call form `tuple(arg, arg, ...)` (instead of the operator form `(arg, arg, ...)`). In that context the outer `parenthesized` paren is always redundant — `(tuple(a, b))` and `tuple(a, b)` denote the same value — so suppressing it in `IAST::decideParensEmission` keeps the round-trip stable.

The new clause mirrors the existing literal-tuple suppression a few lines above. The fix is gated on:

- `func->name == "tuple"` (other operator-form functions like `plus`/`equals` need their parens for precedence)
- `func->arguments->children.size() > 1` (single-argument `tuple(x)` keeps its existing behaviour and round-trips cleanly)
- `!frame.allow_operators` (plain expression contexts where the operator form `(a, b)` is canonical are unaffected — multi-element tuple literals there are already handled by the existing `Field::Types::Tuple` clause above)

This is a partial fix for chronic STID 1941-1bfa. It addresses two subvariants reported by `@alexey-milovidov` on PRs #100412 / #94517 / #104591:

- `CREATE TABLE ... CODEC(not(a, b))` (variadic-2 call on unary function in `CODEC`)
- `ALTER TABLE ... MODIFY STATISTICS ... TYPE concatAssumeInjective((...))`

Other reported subvariants — `DESCRIBE TABLE (SELECT ... COLUMNS(multiIf(...)) ...)` and qualifier-chained `SELECT db.t.COLUMNS(*) FROM ...` — involve different parser/formatter paths (`COLUMNS` matcher / `DESCRIBE` wrapper) and need separate fixes; they are still tracked in the canonical task.

Related: PR #102547 (`@zoomxi`, INSERT-with-SAMPLE-OFFSET subvariant), PR #104520 (data-type collection literals subvariant), PR #104626 (lambda-after-comma subvariant, issue #104605).

CI sighting on PR #100412 (one of 40+ affected PRs in the last 90 days plus master hits): https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100412&sha=3284612ceea8078d500d582ce82e5ace86b239cf&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%2C%20targeted%29

Fix a `Logical error: 'Inconsistent AST formatting'` server abort (STID 1941-1bfa) on queries like `CREATE TABLE t (c0 Int CODEC(not((not(materialize(1), materialize(2)))), ZSTD)) ENGINE = Memory` in debug / sanitiser builds. The formatter no longer emits redundant outer parens around a multi-argument `tuple(...)` function call that appears inside `CODEC` / `STATISTICS` / `BACKUP_NAME` argument lists (where the operator form `(a, b)` is disabled), keeping the format-parse-format round-trip stable.

- [x] Documentation is unchanged (no user-facing API or behaviour change beyond round-trip stability).

### Version info
- Merged into: `26.6.1.284`


groeneai:
**Pre-PR validation gate** (self-assessment per `TASK.md` Phase 4):

(a) **Deterministic repro?** Yes. `CREATE TABLE t (c0 Int CODEC(not((not(materialize(1), materialize(2)))), ZSTD)) ENGINE = Memory` aborts the server on master (debug build) with `Logical error: 'Inconsistent AST formatting between 'ExpressionList' and 'ExpressionList' (STID: 1941-1bfa)`. Hits the round-trip check at `src/Interpreters/executeQuery.cpp:1330`. Single command, no settings dependency.

(b) **Root cause explained?** Yes. The first format emits `tuple(materialize(1), materialize(2))` (function-call form because `allow_operators=false` inside `CODEC`). The re-parser sees `(tuple(...))` (parens come from the outer `not(...)` argument) and `RoundBracketsLayer::getResultImpl` in `src/Parsers/ExpressionListParsers.cpp:1375` sets `parenthesized = true` on the inner `Function_tuple`. On the second format pass, `IAST::decideParensEmission` emits the `parenthesized` flag's parens, adding one extra `(...)` layer: `not((tuple(...)))` vs the original `not(tuple(...))`. Diff is exactly one paren around `tuple(...)`.

(c) **Fix matches root cause?** Yes. Suppresses the `parenthesized` flag's parens for multi-argument `Function_tuple` specifically when `frame.allow_operators = false` (the only context where the formatter falls back to function-call form `tuple(...)` for an operator-form-capable function). The fix lives at the exact site that decides paren emission (`decideParensEmission`), mirrors the existing literal-tuple suppression a few lines above, and stays within the formatter — no parser change.

(d) **Test intent preserved? / New tests added?** Yes, new test `04241_inconsistent_ast_codec_tuple_paren_round_trip_1941_1bfa.sql` covers the original reproducer plus 8 round-trip variants (multiple arg counts, `tuple` vs `(tuple)` shapes, `STATISTICS` context, single-arg negative case to prove the fix didn't over-reach, real `CODEC(Delta, LZ4)` to prove ordinary declarations are unaffected, expression-context tuple usage to prove `!frame.allow_operators` gating is correct). 19 existing AST / `CODEC` / paren tests still pass with the fix.

(e) **Both directions demonstrated?** Yes. Before the fix the original `CREATE TABLE` reproducer aborts the server with SIGABRT (verified locally on `b6844b26db6`). After the fix the same statement returns `UNKNOWN_CODEC` (the proper error because `not` is not a codec family), and `formatQuerySingleLine` on the same shape returns a single-pass-stable formatted output. The new test was run with `--test-runs 30` against the fixed binary — 30/30 pass.

(f) **Fix is general, not a narrow patch?** Yes within its scope, with documented limitations:
- The same parser-side bug pattern (`RoundBracketsLayer` setting `parenthesized = true` on a function-call node) is checked against `Function_array` — but `[]` is parsed via `ArrayLayer`, not `RoundBracketsLayer`, so the asymmetry that breaks `Function_tuple` does not exist for `Function_array`. Verified locally: `CREATE TABLE t (c0 Int CODEC(not((array(1, 2, 3))), ZSTD))` round-trips cleanly without the fix.
- The fix covers two of the four STID 1941-1bfa subvariants flagged by `@alexey-milovidov` on PRs #100412 / #94517 / #104591: `CODEC(not(a, b))` and `STATISTICS concatAssumeInjective((...))`. The other two — `DESCRIBE TABLE (SELECT ... COLUMNS(multiIf(...)))` and qualifier-chained `SELECT db.t.COLUMNS(*) FROM ...` — involve different parser/formatter paths (`COLUMNS` matcher / `DESCRIBE` wrapper) and are tracked in the canonical task for separate fixes. This PR is intentionally scoped to one root cause to keep the diff minimal and reviewable.
- Defensive gating: `func->name == "tuple"`, `args.size() > 1`, and `!frame.allow_operators` together ensure (i) other operator-form functions like `plus` / `equals` keep their parens for precedence, (ii) single-argument `tuple(x)` is unaffected (already round-trips cleanly), and (iii) plain expression contexts where `(a, b)` is canonical are unaffected.

Session: `cron:clickhouse-ci-task-worker:20260515-014500`.

groeneai:
cc @evillique @alexey-milovidov @george-larionov — could you review this? It fixes the chronic STID 1941-1bfa `Inconsistent AST formatting` LOGICAL_ERROR for the `CODEC(not(a, b))` / `STATISTICS concatAssumeInjective((...))` subvariants by suppressing redundant outer parens around multi-arg `Function_tuple` in `allow_operators=false` contexts in `IAST::decideParensEmission`. Follow-up to PR #102547 / #104520 / #104626.

PedroTadim:
@groeneai merge with master

groeneai:
@PedroTadim merged master into the PR branch as requested. New HEAD SHA: `551bf4473f77`.

Merge was conflict-free. PR diff is unchanged (3 files, 113 insertions):
- `src/Parsers/IAST.cpp` (decideParensEmission Function_tuple multi-arg suppression in !allow_operators contexts)
- `tests/queries/0_stateless/04241_inconsistent_ast_codec_tuple_paren_round_trip_1941_1bfa.sql` (regression test)
- `tests/queries/0_stateless/04241_inconsistent_ast_codec_tuple_paren_round_trip_1941_1bfa.reference`

src/Parsers/IAST.cpp blob was identical between the PR's merge-base and current master, so no behavioural changes from the rebase. CI should retrigger automatically.

alexey-milovidov (approved):

PR #104966, opened 2026-05-14
Fix exception on `IN tuple()` against Distributed sharded table

alexey-milovidov:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a user-readable short description of the changes that goes into CHANGELOG.md):
Fix an exception (`Received signal 6 (abort)` in vector hardening, observable as `Tuple::back()` on an empty vector) when executing a query of the form `SELECT ... FROM <Distributed table> WHERE/HAVING/GROUP BY ... (sharding_key_column IN tuple())` with `optimize_skip_unused_shards_rewrite_in` enabled under the new analyzer.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

---

#### What was wrong

`OptimizeShardingKeyRewriteIn::enterImpl` in `OptimizeShardingKeyRewriteInVisitor.cpp` calls `tuple.back()` on an empty `std::vector` when the right side of `IN` is the empty literal tuple `tuple()`. The AST-side rewriter directly above guards this with `tuple.size() > 1` and is safe; the analyzer-side path did not.

The AST fuzzer reduced its way to:

```sql
SELECT (2 IN tuple(-2)) FROM dist_01756
WHERE dummy IN (0, 2)
GROUP BY ..., (dummy IN tuple()) AND ...
```

with `dist_01756` sharded on `intHash64(dummy)`, so the inner `(dummy IN tuple())` reached the analyzer rewriter and tripped UB on the empty `Tuple::back()` — observed in CI as `Received signal Aborted (6)` / `verbose_abort.cpp` / `vector::back[abi:sqe220101]() const`.

#### Fix

Bail out early when the constant tuple is empty — `IN ()` is a degenerate query (always false) and there is nothing useful to rewrite. This matches the existing AST-side behavior.

#### Test

[...]

PR #101033, opened 2026-03-28
Allow passing bare function names to higher-order functions

alexey-milovidov:
Allow passing a function name directly to a higher-order function instead of writing a lambda expression. For example:

```sql
SELECT arrayMap(negate, [1, 2, 3])           -- [-1, -2, -3]
SELECT arrayMap(toString, [1, 2, 3])         -- ['1', '2', '3']
SELECT arrayMap(plus, [1, 2, 3], [10, 20, 30]) -- [11, 22, 33]
SELECT arrayFold(plus, [1, 2, 3, 4, 5], toUInt64(0)) -- 15
SELECT arrayFilter(isNotNull, [1, NULL, 3, NULL, 5]) -- [1, 3, 5]
```

These are equivalent to writing lambdas explicitly (`arrayMap(x -> negate(x), [1, 2, 3])`), but more concise.

The transformation only applies when the identifier does not resolve as a column or alias, so backward compatibility is preserved.

Closes #63498

### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Allow passing bare function names to higher-order functions like `arrayMap`, `arrayFilter`, etc. For example, `arrayMap(negate, [1, 2, 3])` is now equivalent to `arrayMap(x -> negate(x), [1, 2, 3])`.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

A function name can now be passed directly to higher-order functions instead of wrapping it in a lambda. For example: `SELECT arrayMap(toString, [1, 2, 3])` instead of `SELECT arrayMap(x -> toString(x), [1, 2, 3])`. This works with any ordinary function. Column and alias names take priority over function names when there is ambiguity.

### Version info
- Merged into: `26.5.1.659`

PR #104996, opened 2026-05-15
Stabilize 04229_trivial_group_by_limit_profile_events under parallel replicas

groeneai:
Reported by @alexey-milovidov on https://github.com/ClickHouse/ClickHouse/pull/100146:

> `Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel)` — only failure is `04229_trivial_group_by_limit_profile_events`. This test was added in #104473 and has been failing in this exact profile across many unrelated PRs (e.g. #100146, #101033, #104694, #104966). The optimization apparently doesn't fire under the `ParallelReplicas` profile and the test does not account for that.

### Root cause

`OptimizeTrivialGroupByLimitPass` runs in the analyzer on the initiator and lowers `max_rows_to_group_by` on the query's local context (`mutable_context->setSetting("max_rows_to_group_by", max_rows)`). That mutation is not propagated to remote replicas, so once CI's `automatic_parallel_replicas_mode = 2` randomization is sampled (which sets `enable_parallel_replicas = 1` + `parallel_replicas_for_non_replicated_merge_tree = 1` + `cluster_for_parallel_replicas = 'parallel_replicas'`), the aggregator on the remotes never applies the cap and `OverflowAny` stays at zero. The test then sees `04229_on   0` instead of the expected `04229_on   1` and fails 3/3 reruns.

CIDB cross-PR scope (30 days): the failure has been observed on at least PRs #104473 (the source PR), #100146, #101033, #104694, #104966 — all unrelated to the test or to each other.

### Fix

Pin `enable_parallel_replicas = 0` at the query level in both `SETTINGS` clauses so the optimization is exercised on the local-coordinator path it is designed for. Query-level `SETTINGS` takes priority over the runner's session-level injection, so the pin is robust against any further randomization. No `no-*` tag is added (per `.claude/CLAUDE.md` guidance) and no other randomization is disabled — `max_threads = 1` was already pinned, everything else stays randomized.

The test's intent is preserved: it still checks that `OverflowAny` is incremented when `OptimizeTrivialGroupByLimitPass` fires and stays at zero when it doesn't.

### Reproduction & verification

Locally with a 3-replica `default_parallel_replicas` cluster on loopback (`localhost`, `127.0.0.2`, `127.0.0.3` on port 9700), simulating the runner's parallel-replicas injection:

| Test | Without fix | With fix |
| ---- | ----------- | -------- |
| 5 iterations under parallel-replicas randomization | 0/5 pass — every run shows `04229_on   0` | 20/20 pass |

If you want to test the bug interactively, here is a one-line repro against any server with a multi-replica cluster:

```sql
SELECT k FROM t GROUP BY k LIMIT 5 FORMAT Null
SETTINGS optimize_trivial_group_by_limit_query = 1, max_threads = 1,
    enable_parallel_replicas = 1, parallel_replicas_for_non_replicated_merge_tree = 1,
    cluster_for_parallel_replicas = '<your_cluster>', parallel_replicas_local_plan = 0;
```

Then `SELECT ProfileEvents['OverflowAny'] FROM system.query_log WHERE type = 'QueryFinish'` returns `0` instead of the expected non-zero value.

### Version info
- Merged into: `26.5.1.734`

### DIFF
--- a/tests/queries/0_stateless/04229_trivial_group_by_limit_profile_events.sql
+++ b/tests/queries/0_stateless/04229_trivial_group_by_limit_profile_events.sql
@@ -11,12 +11,21 @@ INSERT INTO t_04229 SELECT number FROM numbers(100000);
 -- the `OverflowAny` count deterministic. The query has 100k distinct keys, so
 -- with the optimization the aggregator hits the 5-key cap and drops all the
 -- subsequent keys, incrementing `OverflowAny`.
+--
+-- `enable_parallel_replicas = 0` is pinned because `OptimizeTrivialGroupByLimitPass`
+-- runs in the analyzer on the initiator and mutates `max_rows_to_group_by` on the
+-- query's local context; this mutation does not propagate to remote replicas, so
+-- under CI's parallel-replicas randomization the aggregator on the remotes never
+-- sees the cap and `OverflowAny` stays at zero. The test verifies the optimization
+-- on the local-coordinator path where it is designed to fire.
 
 SELECT k FROM t_04229 GROUP BY k LIMIT 5 FORMAT Null
-SETTINGS optimize_trivial_group_by_limit_query = 1, max_threads = 1, log_comment = '04229_on';
+SETTINGS optimize_trivial_group_by_limit_query = 1, max_threads = 1,
+    enable_parallel_replicas = 0, log_comment = '04229_on';
 
 SELECT k FROM t_04229 GROUP BY k LIMIT 5 FORMAT Null
-SETTINGS optimize_trivial_group_by_limit_query = 0, max_threads = 1, log_comment = '04229_off';
+SETTINGS optimize_trivial_group_by_limit_query = 0, max_threads = 1,
+    enable_parallel_replicas = 0, log_comment = '04229_off';
 
 SYSTEM FLUSH LOGS query_log;
 


groeneai:
cc @alexey-milovidov — could you review this? Test-only fix: pins `enable_parallel_replicas = 0` at the query level so `OptimizeTrivialGroupByLimitPass` reliably fires on the local-coordinator path the test was written for, instead of being silently neutralized by the runner's `automatic_parallel_replicas_mode` randomization.

### Pre-PR validation self-check

- **(a) Deterministic repro?** Yes — locally reproduced 5/5 fails with `enable_parallel_replicas=1` + `parallel_replicas_for_non_replicated_merge_tree=1` + `cluster_for_parallel_replicas='default_parallel_replicas'` + `parallel_replicas_local_plan=0` against a 3-replica loopback cluster. Every run shows `04229_on   0` instead of the expected `04229_on   1`.
- **(b) Root cause explained?** Yes — `OptimizeTrivialGroupByLimitPass` calls `query->getMutableContext()->setSetting("max_rows_to_group_by", max_rows)`. That mutates the *initiator's* query context. The mutation is not propagated to remote replicas, so the aggregators on the remotes never apply the cap and `ProfileEvents::OverflowAny` is never incremented.
- **(c) Fix matches root cause?** Yes — the test is for the local-coordinator behavior of the pass; pinning `enable_parallel_replicas = 0` keeps the query on that path. No `no-*` tag (per `.claude/CLAUDE.md` guidance) and no assertion is widened.
- **(d) Test intent preserved?** Yes — the test still distinguishes "optimization fires → `OverflowAny > 0`" from "optimization disabled → `OverflowAny == 0`". Only `enable_parallel_replicas` is pinned; all other randomization (`max_threads` was already pinned to 1 for determinism, everything else stays randomized).
- **(e) Both directions demonstrated?** Yes — original test: 0/5 pass under parallel-replicas injection (matches CI's 3/3 reruns diff exactly). Fixed test: 20/20 pass under the same injection.

### Note on follow-up

This PR only stabilizes the test. The underlying behavior — that the analyzer pass mutates settings the remote replicas never see — is a separate question. If you'd like the pass to also skip itself when `canUseParallelReplicasOnInitiator()` is true (so `OverflowAny` reliably stays at zero under parallel replicas instead of being a no-op silently), happy to send that as a follow-up.

Session: cron:clickhouse-ci-task-worker:20260515-044500

Algunenano (approved):

PR #104998, opened 2026-05-15
Register ProfileEvents for executable UDF resource usage

XuJia0210:
Step 1 of adding resource usage observability for executable and executable_pool UDFs. Registers seven `ProfileEvents` so that subsequent PRs can wire `ProfileEvents::increment` in `UserDefinedExecutableFunctionFactory::executeImpl` and the rusage plumbing through `ShellCommand` / `ShellCommandSource`. The `ExecutableUserDefinedFunction` prefix is shared between both UDF types; per-UDF attribution is via `system.query_log` and `used_executable_user_defined_functions`.

...

### Version info
- Merged into: `26.5.1.802`


src/Common/ProfileEvents.cpp
```diff
@@ -1347,6 +1347,14 @@ The server successfully detected this situation and will download merged part fr
     M(WasmModuleInstatiate, "Number of WebAssembly compartments created", ValueType::Number) \
     M(WasmMemoryAllocated, "Total memory allocated for WebAssembly compartments", ValueType::Bytes) \
     \
+    M(ExecutableUserDefinedFunctionInvocations, "Number of executable and executable pool user-defined function invocations.", ValueType::Number) \
```
andreev-io, line 1350: Why do you need the `Executable` prefix for all of these ProfileEvents? I think it can be dropped.

src/Common/ProfileEvents.cpp
```diff
@@ -1347,6 +1347,14 @@ The server successfully detected this situation and will download merged part fr
...
+    M(ExecutableUserDefinedFunctionElapsedMicroseconds, "Wall clock time spent executing executable and executable pool user-defined functions, in microseconds.", ValueType::Microseconds) \
```
andreev-io, line 1351: What do you think about using the prefix `UserDefinedFunctions` (plural), to emphasise the fact that these ProfileEvents are summed across all UDF invocations for the query?

andreev-io: The descriptions make it clear already. Just wondering about the names,

src/Common/ProfileEvents.cpp
```diff
@@ -1347,6 +1347,14 @@ The server successfully detected this situation and will download merged part fr
...
+    M(ExecutableUserDefinedFunctionCPUUserMicroseconds, "User mode CPU time consumed by executable and executable pool user-defined function child processes, in microseconds.", ValueType::Microseconds) \
+    M(ExecutableUserDefinedFunctionCPUSystemMicroseconds, "Kernel mode CPU time consumed by executable and executable pool user-defined function child processes, in microseconds.", ValueType::Microseconds) \
+    M(ExecutableUserDefinedFunctionMemoryUsageByteSeconds, "Memory used by executable and executable pool user-defined function child processes integrated over wall clock time, in byte-seconds.", ValueType::Number) \
```
andreev-io: Should we clarify that this is peak memory? Both in the name and in the description? Something like this:
```
UserDefinedFunctionsPeakMemoryByteSeconds: Per-process peak memory used by executable and executable pool user-defined function child processes integrated over wall clock time, in byte-seconds.
```

src/Common/ProfileEvents.cpp
vdimir: For me, `ExecutableUserDefined` sounds good, it clearly distinguishes it from other UDF types (SQL and WASM).

XuJia0210: Because we have SQL UDFs
In our daily communication, we use UDF to refer to executable UDFs. We ignored SQL UDFs
https://clickhouse.com/docs/sql-reference/functions/udf#sql-user-defined-functions

src/Common/ProfileEvents.cpp
vdimir: But this would just be memory amount × time, so it’s not really “peak” memory, but rather memory usage integrated over time, right?

src/Common/ProfileEvents.cpp
XuJia0210: Our naming convention uses singular. 
You can check the file src/Common/ProfileEvents.cpp.

src/Common/ProfileEvents.cpp
andreev-io: I see, I didn't know that, thanks.

src/Common/ProfileEvents.cpp
andreev-io: Got it, thank you.

src/Common/ProfileEvents.cpp
XuJia0210: I agree with Vdimir. It's not the real peak memory usage. It's integrated

andreev-io: Actually, no – for each launched process, we will observe its peak memory (`M_peak_i`) and its duration (`T_i`), multiply them (`M_peak_i x T_i`), and then sum across processes.

Breaking M_peak_i down into point-in-time observations (`sum(M_tj_i) over j`) is much more complicated.

andreev-io: @XuJia0210 My point is that it's not integrated memory usage either. It would've been integrated if we used a time interval memory sample multiplied by the time interval and then summed across intervals, but we use one memory value, the peak. So it's more like "peak" memory usage integrated over time, where "peak" is individual per process.

src/Common/ProfileEvents.cpp
```diff
@@ -1347,6 +1347,14 @@ The server successfully detected this situation and will download merged part fr
     M(WasmModuleInstatiate, "Number of WebAssembly compartments created", ValueType::Number) \
     M(WasmMemoryAllocated, "Total memory allocated for WebAssembly compartments", ValueType::Bytes) \
     \
+    M(ExecutableUserDefinedFunctionInvocations, "Number of executable and executable pool user-defined function invocations.", ValueType::Number) \
+    M(ExecutableUserDefinedFunctionElapsedMicroseconds, "Wall clock time spent executing executable and executable pool user-defined functions, in microseconds.", ValueType::Microseconds) \
+    M(ExecutableUserDefinedFunctionCPUUserMicroseconds, "User mode CPU time consumed by executable and executable pool user-defined function child processes, in microseconds.", ValueType::Microseconds) \
```
HanziJiang: Just a nit, the convention seems to be UserTimeMicroseconds instead of CPUUserMicroseconds, SystemTimeMicroseconds instead of CPUSystemMicroseconds

src/Common/ProfileEvents.cpp
XuJia0210: Ok, I will add peak to the name

XuJia0210:
02180_group_by_lowcardinality
test_distributed_frozen_replica/test.py
They are known to be flaky. No need to concern

vdimir:
> 02180_group_by_lowcardinality

Fixed in https://github.com/ClickHouse/ClickHouse/pull/105194

> test_distributed_frozen_replica/test.py


Fixed in https://github.com/ClickHouse/ClickHouse/pull/105112

vdimir (approved):

PR #88860, opened 2025-10-21
Fix premature TTL column removal causing merge failures and wrong defaults

amosbird:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fixed several issues caused by premature column removal in TTL. Resolves #88002

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Details

1. Fix merge failures in GraphiteMergeTree, VersionedCollapsingMergeTree, ReplacingMergeTree, and AggregatingMergeTree when TTL-removed columns were still required by merge logic
2. Fix incorrect skip index or data values caused by filling missing columns with type defaults instead of column DEFAULT expressions
3. Fix missing OPTIMIZE ... DEDUPLICATE BY behavior when deduplication columns were removed by TTL

PR #83892, opened 2025-07-17
Don't delete default rows in coalescing merge tree

scanhex12:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
This closes https://github.com/ClickHouse/ClickHouse/issues/81303

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

Issue #81303, opened 2025-06-04
CoalescingMergeTree is unable to store default values, like zeros or empty strings


den-crane:
### Company or project name

https://www.innervate.com/

### Describe the unexpected behaviour

From my POV a new CoalescingMergeTree behaves unexpectedly with default values, at least it should be documented.

```sql
CREATE TABLE test_table
(
    key UInt32,
    A UInt32,
    B String
)
ENGINE = CoalescingMergeTree()
ORDER BY key;

INSERT INTO test_table Values(1,0, '');

SELECT count() FROM test_table;
   +-count()-+
1. |       0 |  ---<<<<--- I expect 1 here
   +---------+


drop table test_table;

CREATE TABLE test_table
(
    key UInt32,
    A Nullable(UInt32),
    B Nullable(String)
)
ENGINE = CoalescingMergeTree()
ORDER BY key;

INSERT INTO test_table Values(1, Null, '');

SELECT count() FROM test_table;
   +-count()-+
1. |       0 |  ---<<<<--- I expect 1 here
   +---------+
```

### Which ClickHouse versions are affected?

head

### How to reproduce

https://fiddle.clickhouse.com/817ad0bc-4c95-4211-a089-e3cfb8e093c0

den-crane:
@jkartseva @scanhex12

den-crane:
even with not the default. Am I missing something?

```
CREATE TABLE test_table
(
    key UInt32,
    A Nullable(UInt32),
    B Nullable(String)
)
ENGINE = CoalescingMergeTree()
ORDER BY key;

INSERT INTO test_table Values(1, Null, 'xxx');

SELECT count() FROM test_table;
   +-count()-+
1. |       0 |
   +---------+
```

https://fiddle.clickhouse.com/0cfe3542-0b13-4320-81d9-f1cd0d9ec939

den-crane:
Hmm

```
CREATE TABLE test_table
(
    key UInt32,
    A Nullable(UInt32),
    B Nullable(String)
)
ENGINE = CoalescingMergeTree()
ORDER BY key;

INSERT INTO test_table Values(1, 1,  'x');
SELECT * FROM test_table final;
   +-key-+-A-+-B-+
1. |   1 | 1 | x |
   +-----+---+---+

INSERT INTO test_table Values(1, 2,  'y');
SELECT * FROM test_table final;
   +-key-+-A-+-B-+
[...]

PR #104999, opened 2026-05-15
Fix CLEAR COLUMN zero-row deletion regression on SummingMergeTree

### LINKED ISSUE #101953: CLEAR COLUMN IN PARTITION on SummingMergeTree does not trigger zero-row deletion after OPTIMIZE FINAL (regression since 25.8)
### Describe what's wrong

Since ClickHouse 25.8, `ALTER TABLE CLEAR COLUMN IN PARTITION` on a `SummingMergeTree` table leaves rows in a state where the zero-row deletion logic no longer recognizes them as zero during `OPTIMIZE TABLE FINAL`. Using `ALTER TABLE UPDATE c = 0` on the same column and table works correctly — the row is deleted as expected.

This suggests that `CLEAR COLUMN` sets the column to a "default" internal representation that the `SummingSortedAlgorithm` does not recognize as zero, whereas `UPDATE c = 0` writes an explicit zero that is recognized.

Additionally, `CLEAR COLUMN` is allowed on auto-detected summing columns but blocked on explicitly declared ones (`SummingMergeTree(c)`), which is an inconsistency in validation — present in all tested versions.

### Versions tested:
- **25.7.8.71** - `CLEAR COLUMN` + `OPTIMIZE FINAL` **deletes** the zero row (correct), but validation inconsistency exists
- **25.8.21.7** - `CLEAR COLUMN` + `OPTIMIZE FINAL` **does not delete** the zero row (regression)
- **26.1.7.13** - same behavior as 25.8 (regression persists)

### How to reproduce

```sql
-- Table with auto-detected summing columns (c is the only summing column)
CREATE TABLE test_clear_summing
(
    v UInt64,
    p UInt64,
    c UInt64
)
ENGINE = SummingMergeTree
PARTITION BY p
ORDER BY v;

INSERT INTO test_clear_summing VALUES (1, 1, 100);
INSERT INTO test_clear_summing VALUES (2, 2, 200);

OPTIMIZE TABLE test_clear_summing FINAL;

SELECT * FROM test_clear_summing;
-- ┌─v─┬─p─┬───c─┐
-- │ 1 │ 1 │ 100 │
-- │ 2 │ 2 │ 200 │
-- └───┴───┴─────┘

-- Zero the summing column via CLEAR COLUMN
ALTER TABLE test_clear_summing CLEAR COLUMN c IN PARTITION 1;
OPTIMIZE TABLE test_clear_summing FINAL;

SELECT * FROM test_clear_summing;
-- Expected (<=25.7): 1 row (only partition 2, zero row deleted)
-- Actual (>=25.8):   2 rows (zero row preserved)
-- ┌─v─┬─p─┬───c─┐
-- │ 1 │ 1 │   0 │  ← should have been deleted
-- │ 2 │ 2 │ 200 │
-- └───┴───┴─────┘

-- For comparison: UPDATE works correctly on the same table
ALTER TABLE test_clear_summing UPDATE c = 0 WHERE v = 1;
OPTIMIZE TABLE test_clear_summing FINAL;

SELECT * FROM test_clear_summing;
-- ┌─v─┬─p─┬───c─┐
-- │ 2 │ 2 │ 200 │
-- └───┴───┴─────┘
-- Row correctly deleted

DROP TABLE test_clear_summing;
```

### Expected behavior

Per the [documentation](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/summingmergetree):

> *"If the values were 0 in all of the columns for summation, the row is deleted."*

After `CLEAR COLUMN c IN PARTITION 1` and `OPTIMIZE TABLE FINAL`, column `c` is the only summed column (`v` is ORDER BY key, `p` is PARTITION BY key) and `c = 0`. The row should be deleted during merge, just as it is when using `UPDATE c = 0`.

### Additional note: validation inconsistency (all versions)

`CLEAR COLUMN` on a summing column is blocked when columns are explicitly declared, but allowed when auto-detected:

```sql
-- Explicitly declared: blocked (correct)
CREATE TABLE t1 (v UInt64, p UInt64, c UInt64)
ENGINE = SummingMergeTree(c) PARTITION BY p ORDER BY v;

ALTER TABLE t1 CLEAR COLUMN c IN PARTITION 1;
-- Code: 524. DB::Exception: Trying to ALTER DROP columns to sum (c) column.

-- Auto-detected: allowed (inconsistent)
CREATE TABLE t2 (v UInt64, p UInt64, c UInt64)
ENGINE = SummingMergeTree PARTITION BY p ORDER BY v;

ALTER TABLE t2 CLEAR COLUMN c IN PARTITION 1;
-- Ok.
```

### Likely cause

PR #83892 refactored `SummingSortedAlgorithm::finishGroup()`, adding a `remove_default_values` guard to the zero-row deletion check. `CLEAR COLUMN` may set column values to a "default" state that is handled differently from an explicit zero after this refactor, causing the zero-row detection to fail for cleared columns.

groeneai:
`ALTER TABLE ... CLEAR COLUMN IN PARTITION` on a `SummingMergeTree` stopped triggering zero-row deletion during `OPTIMIZE TABLE FINAL` after the merge-time "absent-column-as-expired" optimization (#88860) was backported to 25.8.

### Reproducer

```sql
CREATE TABLE t (v UInt64, p UInt64, c UInt64)
ENGINE = SummingMergeTree PARTITION BY p ORDER BY v;

INSERT INTO t VALUES (1, 1, 100), (2, 2, 200);
OPTIMIZE TABLE t FINAL;
ALTER TABLE t CLEAR COLUMN c IN PARTITION 1 SETTINGS mutations_sync = 2;
OPTIMIZE TABLE t FINAL;

SELECT * FROM t;
-- Expected (≤ 25.7): only (2, 2, 200) — the all-zero row in partition 1 is deleted.
-- Actual   (≥ 25.8): (1, 1, 0) and (2, 2, 200) — zero row survives.
```

`ALTER TABLE t UPDATE c = 0 WHERE v = 1` on the same table still works correctly, because `UPDATE` writes an explicit `0` and keeps `c` in `columns.txt`, while `CLEAR COLUMN` drops the cleared column from `columns.txt` entirely.

### Root cause

`MergeTask::ExecuteAndFinalizeHorizontalPart::prepare` marks any storage column that is missing from every source part and has no default expression as expired:

```cpp
for (const auto & storage_column : global_ctx->storage_columns)
{
    if (!columns_present_in_parts.contains(storage_column.name)
        && !columns_desc.getDefault(storage_column.name))
        global_ctx->new_data_part->expired_columns.emplace(storage_column.name);
}
```

These expired columns are then filtered out of `merging_columns`/`gathering_columns`/`storage_columns`. `SummingSortedAlgorithm::defineColumns` therefore sees only the key columns, builds an empty `columns_to_aggregate`, and in `SummingMergedData::finishGroup`:

```cpp
if (def.columns_to_aggregate.empty())
    current_row_is_zero = false;
```

forces every row to be kept, bypassing zero-row deletion.

### Fix

Skip the absent-column-as-expired optimization for `Summing`, `Aggregating`, `Coalescing`, and `Graphite` merge modes. In these modes every non-key column participates in the merge algorithm — they drive zero-row deletion (Summing/Coalescing), aggregate-state merging (Aggregating), and bucketed value/version aggregation (Graphite). Filtering them out changes merge semantics.

`Ordinary`, `Collapsing`, `Replacing`, and `VersionedCollapsing` are unchanged: they do not depend on non-key columns for row inclusion, and their special columns (`sign_column`/`version_column`/`is_deleted_column`) are already in `merge_required_key_columns`.

### Validation

- New regression test `04243_clear_column_summing_zero_deletion` covers `SummingMergeTree` `CLEAR COLUMN`, `UPDATE c = 0` cross-check, multi-row partition with mixed values, `ALTER ADD COLUMN` with no part rewrite, `CoalescingMergeTree` (rows preserved — issue #81303 behaviour preserved), and `Ordinary` `MergeTree` (rows preserved).
- The new test fails on `master` (returns `1\t1\t0` from the `SummingMergeTree` case) and passes after the fix.
- Existing tests pass: `03566_one_row_summing_merge_tree` (PR #83892's test for issue #81303), `00043_summing_empty_part`, `00084_summing_merge_tree`, `00146_summing_merge_tree_nested_map`, `00148_summing_merge_tree_aggregate_function`, `00148_summing_merge_tree_nested_map_multiple_values`, `00625_summing_merge_tree_merge`, `03409_coalescing_merge_tree`, `03409_coalescing_replicated_merge_tree`, `02346_text_index_coalescingmergetree`.
- 10/10 runs of `04243` pass with random settings + random MergeTree settings.

Fixed a regression introduced in 25.8 where `ALTER TABLE ... CLEAR COLUMN IN PARTITION` on a `SummingMergeTree` table no longer triggered zero-row deletion during `OPTIMIZE TABLE FINAL`. The merge-time "absent-column-as-expired" optimization is now skipped for `SummingMergeTree`, `AggregatingMergeTree`, `CoalescingMergeTree`, and `GraphiteMergeTree`, where every non-key column participates in the merge algorithm.

Closes #101953

Session: cron:clickhouse-ci-task-worker:20260515-054500

### DIFF
--- a/src/Storages/MergeTree/MergeTask.cpp
+++ b/src/Storages/MergeTree/MergeTask.cpp
@@ -612,6 +612,25 @@ bool MergeTask::ExecuteAndFinalizeHorizontalPart::prepare() const
     ///    making it unclear whether the default should be materialized or recomputed.
     /// 2. Default expressions may introduce semantic changes if re-evaluated during merges, leading to
     ///    non-deterministic results across parts.
+    ///
+    /// Additional constraint: for `Summing`, `Aggregating`, `Coalescing`, and `Graphite` merge modes,
+    /// all non-key columns participate in the merge algorithm and can affect which rows survive the
+    /// merge (e.g. zero-row deletion in `SummingMergeTree`/`CoalescingMergeTree`, aggregate-state
+    /// merging in `AggregatingMergeTree`). If a column is absent from all source parts (for example
+    /// after `ALTER TABLE ... CLEAR COLUMN IN PARTITION`) and we expire it here, the read pipeline
+    /// stops materializing the implicit default-value column for it, and `SummingSortedAlgorithm`
+    /// never sees the column. The zero-row deletion check then sees an empty `columns_to_aggregate`
+    /// and forces `current_row_is_zero = false`, so rows that should have been collapsed survive.
+    /// Skip the optimization for these modes to preserve the pre-#88860 merge semantics.
+    /// See https://github.com/ClickHouse/ClickHouse/issues/101953
+    const auto merge_mode = global_ctx->merging_params.mode;
+    const bool merge_mode_uses_all_non_key_columns =
+        merge_mode == MergeTreeData::MergingParams::Summing
+        || merge_mode == MergeTreeData::MergingParams::Aggregating
+        || merge_mode == MergeTreeData::MergingParams::Coalescing
+        || merge_mode == MergeTreeData::MergingParams::Graphite;
+
+    if (!merge_mode_uses_all_non_key_columns)
     {
         NameSet columns_present_in_parts;
         columns_present_in_parts.reserve(global_ctx->storage_columns.size());
--- a/tests/queries/0_stateless/04243_clear_column_summing_zero_deletion.sql
+++ b/tests/queries/0_stateless/04243_clear_column_summing_zero_deletion.sql
@@ -1,0 +1,175 @@
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/101953
+--
+-- ALTER TABLE ... CLEAR COLUMN IN PARTITION on a SummingMergeTree table must
+-- still trigger zero-row deletion during OPTIMIZE TABLE FINAL, matching the
+-- behaviour of ALTER TABLE ... UPDATE col = 0.
+--
+-- The regression was introduced by https://github.com/ClickHouse/ClickHouse/pull/88860
+-- ("Treat absent columns as expired"), which extended the merge-time
+-- "absent-column-as-expired" optimization from TTL-finished columns to any
+-- column missing from all source parts. CLEAR COLUMN drops the cleared column
+-- from columns.txt entirely, so under the new optimization the column never
+-- reaches SummingSortedAlgorithm and the zero-row deletion is silently bypassed.
+--
+-- The fix narrows the optimization to merge modes where non-key columns do not
+-- affect row inclusion (Ordinary, Collapsing, Replacing, VersionedCollapsing),
+-- and skips it for Summing, Aggregating, Coalescing, and Graphite modes where
+-- every non-key column participates in the merge algorithm.
+
+DROP TABLE IF EXISTS test_clear_summing;
+
+CREATE TABLE test_clear_summing
+(
+    v UInt64,
+    p UInt64,
+    c UInt64
+)
+ENGINE = SummingMergeTree
+PARTITION BY p
+ORDER BY v;
+
+INSERT INTO test_clear_summing VALUES (1, 1, 100);
+INSERT INTO test_clear_summing VALUES (2, 2, 200);
+
+OPTIMIZE TABLE test_clear_summing FINAL;
+
+SELECT 'before clear';
+SELECT * FROM test_clear_summing ORDER BY v;
+
+ALTER TABLE test_clear_summing CLEAR COLUMN c IN PARTITION 1 SETTINGS mutations_sync = 2;
+
+OPTIMIZE TABLE test_clear_summing FINAL;
+
+SELECT 'after clear + optimize final';
+SELECT * FROM test_clear_summing ORDER BY v;
+
+-- Cross-check: ALTER UPDATE c = 0 must produce the same result.
+DROP TABLE test_clear_summing;
+
+CREATE TABLE test_clear_summing
+(
+    v UInt64,
+    p UInt64,
+    c UInt64
+)
+ENGINE = SummingMergeTree
+PARTITION BY p
+ORDER BY v;
+
+INSERT INTO test_clear_summing VALUES (1, 1, 100);
+INSERT INTO test_clear_summing VALUES (2, 2, 200);
+
+OPTIMIZE TABLE test_clear_summing FINAL;
+
+ALTER TABLE test_clear_summing UPDATE c = 0 WHERE v = 1 SETTINGS mutations_sync = 2;
+
+OPTIMIZE TABLE test_clear_summing FINAL;
+
+SELECT 'cross-check update c=0';
+SELECT * FROM test_clear_summing ORDER BY v;
+
+-- Multi-row partition: every row in partition 1 has c = 0 after CLEAR, so the
+-- whole partition must collapse to nothing.
+DROP TABLE test_clear_summing;
+
+CREATE TABLE test_clear_summing
+(
+    v UInt64,
+    p UInt64,
+    c UInt64
+)
+ENGINE = SummingMergeTree
+PARTITION BY p
+ORDER BY v;
+
+INSERT INTO test_clear_summing VALUES (1, 1, 100);
+INSERT INTO test_clear_summing VALUES (2, 1, 50);
+INSERT INTO test_clear_summing VALUES (3, 1, 30);
+
+OPTIMIZE TABLE test_clear_summing FINAL;
+
+ALTER TABLE test_clear_summing CLEAR COLUMN c IN PARTITION 1 SETTINGS mutations_sync = 2;
+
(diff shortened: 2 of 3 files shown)


groeneai:
**Pre-PR validation gate (self-assessment)**

| # | Question | Status |
|---|---|---|
| a | Deterministic repro? | ✅ Exact SQL above triggers the bug on every run on master. |
| b | Root cause explained? | ✅ Causal chain in the PR body: `CLEAR COLUMN` drops the column from `columns.txt` → `MergeTask` marks it expired → filtered from merge stream → `SummingSortedAlgorithm::defineColumns` produces empty `columns_to_aggregate` → `finishGroup` forces `current_row_is_zero = false` → row preserved. |
| c | Fix matches root cause? | ✅ Skip the absent-as-expired optimization for the four merge modes where every non-key column participates in merge semantics. Not a band-aid on `SummingSortedAlgorithm`. |
| d | Test intent preserved + regression test added? | ✅ Existing `03566_one_row_summing_merge_tree` (PR #83892's regression test for issue #81303) still passes — `CoalescingMergeTree` default-row preservation is unchanged. New regression test `04243_clear_column_summing_zero_deletion` covers the bug and adjacent paths. |
| e | Fails without fix, passes with fix? | ✅ Verified `04243_clear_column_summing_zero_deletion` fails on `master` (returns `1\t1\t0` from the `SummingMergeTree` case) and passes after the patch. |
| f | Fix is general, not a narrow patch? | ✅ Bug pattern (filtering absent columns out of merge stream) checked for all merge modes. Covered for `Summing`, `Aggregating`, `Coalescing`, `Graphite`. Other modes (`Ordinary`, `Collapsing`, `Replacing`, `VersionedCollapsing`) do not depend on non-key columns for row inclusion — their special columns are already pinned via `merge_required_key_columns`. Root-cause fix at the source of the over-aggressive optimization, not a defensive check downstream. |

cc @scanhex12 @alesapin @CurtizJ @nikitamikhaylov — could you review? Bug was already mostly diagnosed by @CarlosFelipeOR (reporter) and @PedroTadim in the issue thread.

Session: `cron:clickhouse-ci-task-worker:20260515-054500`

src/Storages/MergeTree/MergeTask.cpp
```diff
@@ -612,6 +612,25 @@ bool MergeTask::ExecuteAndFinalizeHorizontalPart::prepare() const
...
+    /// Additional constraint: for `Summing`, `Aggregating`, `Coalescing`, and `Graphite` merge modes,
+    /// all non-key columns participate in the merge algorithm and can affect which rows survive the
+    /// merge (e.g. zero-row deletion in `SummingMergeTree`/`CoalescingMergeTree`, aggregate-state
+    /// merging in `AggregatingMergeTree`). If a column is absent from all source parts (for example
+    /// after `ALTER TABLE ... CLEAR COLUMN IN PARTITION`) and we expire it here, the read pipeline
+    /// stops materializing the implicit default-value column for it, and `SummingSortedAlgorithm`
+    /// never sees the column. The zero-row deletion check then sees an empty `columns_to_aggregate`
+    /// and forces `current_row_is_zero = false`, so rows that should have been collapsed survive.
+    /// Skip the optimization for these modes to preserve the pre-#88860 merge semantics.
+    /// See https://github.com/ClickHouse/ClickHouse/issues/101953
```
antonio2368, line 625: Can you simplify the comment?

tests/queries/0_stateless/04243_clear_column_summing_zero_deletion.sql
antonio2368, line 1: Prove that it happens for `AggregatingMergeTree` and `GraphiteMergeTree` and add tests for them.

src/Storages/MergeTree/MergeTask.cpp
```diff
@@ -612,6 +612,25 @@ bool MergeTask::ExecuteAndFinalizeHorizontalPart::prepare() const
...
+    const auto merge_mode = global_ctx->merging_params.mode;
+    const bool merge_mode_uses_all_non_key_columns =
```
antonio2368, line 627: wouldn't this undo the motivation of https://github.com/ClickHouse/ClickHouse/pull/88860?
@amosbird do you have any thoughts here?

antonio2368:
@groeneai I created a better (I hope) fix for this issue, check it and review it https://github.com/ClickHouse/ClickHouse/pull/105203

groeneai:
Thanks @antonio2368 — left a substantive review on #105203. Your structural approach is better than mine on three fronts (TTL/absent-expiry separation preserved, GraphiteMergeTree treated correctly via rollup columns only, Replacing covered as a scope assertion). Happy to defer to #105203.

I'll close this. cc @PedroTadim @CarlosFelipeOR — #105203 is the canonical fix for #101953.

Session: `cron:clickhouse-ci-task-worker:20260518-104500`

PR #79465, opened 2025-04-23
 clickhouse-benchmark: reconfigure reconnect option to take integer values

sachinkumarsingh092:
### Changelog category (leave one):
- Improvement

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):

For `clickhouse-benchmark` reconfigure `reconnect` option to take 0, 1 or N as values for reconnecting accordingly.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

Fixes: #79379 

**Changes**
- Modify `--reconnect` option in clickhouse-benchmark to take an interger value.
- Add tests and docs for the same.

PR #105006, opened 2026-05-15
Restore implicit value for clickhouse-benchmark --reconnect

### LINKED ISSUE #101748: Missing implicit_value(1) on --reconnect breaks old flag syntax

_Retrospective finding from a historical scan of [PR #79465](https://github.com/ClickHouse/ClickHouse/pull/79465) (merged 2025-04-27). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

clickhouse-benchmark --reconnect (without =N value) now fails with 'Bad arguments: the required argument for option --reconnect is missing'. Previously --reconnect was a boolean flag that needed no value.

**Root cause:** Benchmark.cpp:880: option definition lacks ->implicit_value(1) to preserve backward compatibility with the old boolean flag syntax

**Why we believe this is a bug:** Benchmark.cpp:880 defines reconnect as value<size_t>()->default_value(0) without implicit_value(1). boost::program_options requires a value for non-implicit valued options, so --reconnect alone fails.

**Affected locations:**
- `programs/benchmark/Benchmark.cpp:880` — reconnect option definition missing implicit_value

**Impact:** Existing scripts and documentation using --reconnect (without a value) break. Users get a confusing error instead of the expected reconnect-every-query behavior.

Yes — confirmed on current `master` (commit `18dfe15a9516`).

### How to reproduce

```sql
#!/usr/bin/env bash
# Tags: no-fasttest
CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)

# Old syntax: --reconnect (flag, no value) should work like --reconnect=1
$CLICKHOUSE_BENCHMARK --iterations=2 --reconnect <<< 'SELECT 1' 2>&1 | grep -cF 'Queries executed'
```

### Expected behavior

```
Queries executed: 2 (100%).
```

```
Bad arguments: the required argument for option '--reconnect' is missing
```

### Additional context

**Open risks:**
- Documentation at docs/en/operations/utilities/clickhouse-benchmark.md may need updating to note the implicit value

**Suggested fix:** Change to: value<size_t>()->default_value(0)->implicit_value(1) — this makes --reconnect (no value) behave like old --reconnect flag (reconnect every query), --reconnect=0 disables, --reconnect=N reconnects every N queries.

Found during automated review of [PR #79465](https://github.com/ClickHouse/ClickHouse/pull/79465).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr79465_001`_

groeneai:
PR #79465 changed `--reconnect` from a boolean flag to an integer option
of type `value<size_t>()->default_value(0)` so that users could write
`--reconnect=N` to reconnect every N queries. The change accidentally
broke the old bare `--reconnect` form: boost::program_options requires
a value for `value<T>()` options unless an `implicit_value` is also
specified, so

```
clickhouse-benchmark --iterations=2 --reconnect <<< 'SELECT 1'
```

now fails with `Bad arguments: the required argument for option '--reconnect' is missing`.

This adds `->implicit_value(1)` to the option definition so the bare
form once again means "reconnect on every query", matching the
pre-#79465 semantics. The full matrix now works:

- `--reconnect` — reconnect on every query (implicit `1`)
- `--reconnect=0` — never reconnect (default)
- `--reconnect=N` — reconnect every N queries

Includes a stateless regression test
`04244_benchmark_reconnect_flag_101748.sh` that exercises the three
spellings.

Closes #101748

`clickhouse-benchmark --reconnect` (bare, no value) was inadvertently broken in 25.4 by a change that turned `--reconnect` into an integer option, requiring a value. The bare form now works again and is equivalent to `--reconnect=1` (reconnect on every query).

- [x] Documentation is unchanged: only existing behaviour is restored.

### Version info
- Merged into: `26.5.1.735`


groeneai:
## Pre-PR validation gate (TASK.md Phase 4 §7)

**a) Deterministic repro?** Yes — `build/programs/clickhouse-benchmark --iterations=2 --reconnect <<< 'SELECT 1'` deterministically fails on master with `Bad arguments: the required argument for option '--reconnect' is missing`.

**b) Root cause explained?** `boost::program_options` requires a value for any option declared with `value<T>()` unless `implicit_value(...)` is also set. PR #79465 changed `--reconnect` from a bare flag to `value<size_t>()->default_value(0)` to support `--reconnect=N`, but did not chain `implicit_value(1)` — so the bare `--reconnect` form (no value) loses its argument and boost rejects the command line before any benchmark code runs.

**c) Fix matches root cause?** Yes — adds `->implicit_value(1)` to the same option definition. Single-line targeted change, no other code paths or settings touched.

**d) Test intent preserved?** N/A — this is a new regression test for a previously-untested behaviour matrix. The test exercises three spellings (`--reconnect`, `--reconnect=0`, `--reconnect=N`) and asserts each parses successfully (exit code 0 from `clickhouse-benchmark`).

**e) Both directions demonstrated?** Yes — verified locally:
- Without fix: `tests/clickhouse-test 04244_benchmark_reconnect_flag_101748` reports `Having 1 errors! 0 tests passed.`
- With fix: 20/20 successful runs with default random-settings randomization.

session: cron:clickhouse-ci-task-worker:20260515-084500

groeneai:
cc @azat @clickgapai @alexey-milovidov — small follow-up to #79465 restoring the bare `--reconnect` (no value) form via `implicit_value(1)`. Test `04244_benchmark_reconnect_flag_101748.sh` covers all three spellings. Could you take a look?

tests/queries/0_stateless/04244_benchmark_reconnect_flag_101748.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+# Tags: no-fasttest
```
Algunenano: @groeneai Why no fasttest? And why do we need 5 lines to just add a link to the issue? Please remove all unnecessary comments

Algunenano (changes requested):

groeneai:
Thanks @Algunenano — pushed [7bcf0e7](https://github.com/ClickHouse/ClickHouse/commit/7bcf0e7e9be):

- Dropped `# Tags: no-fasttest` so the test runs in fast test
- Reduced the comment block to a single `# Regression test for https://github.com/ClickHouse/ClickHouse/issues/101748` line and removed the per-spelling comments

Locally `tests/clickhouse-test --test-runs 10 04244_benchmark_reconnect_flag_101748` → 10/10 OK on the rebuilt binary.

Algunenano (approved):

clickgapai:
Thanks @groeneai — I verified the mechanics of this fix:

1. **Root cause confirmed**: Line 880 in master has `value<size_t>()->default_value(0)` without `implicit_value()`. Boost.ProgramOptions requires a value for `value<T>()` options unless `implicit_value()` is set.

2. **Fix semantics**: `implicit_value(1)` means bare `--reconnect` → `reconnect=1`. With `reconnect=1`, `queries % 1 == 0` is always true → reconnect every query. This matches the pre-#79465 flag behavior.

3. **Test coverage**: The shell test exercises all three forms (`--reconnect`, `--reconnect=0`, `--reconnect=N`).

4. **Backward compatibility**: `--reconnect=0` and `--reconnect=N` continue to work since explicit values override `implicit_value`.

CI passed ✅. No backport needed — #79465 was recent (not in stable branches).

clickgapai:
Hi — could this be a backport candidate for `26.4`, `26.3` (LTS), `26.2`, `25.8` (LTS)? No backport label was found.

**Why this might apply:** PR #79465 (merged 2025-04-27) broke the bare `--reconnect` flag. This predates all supported branches (25.8 released 2025-08-29), so the bug should exist in all listed branches. The fix is low-risk (1 line change to CLI parsing) and restores documented behavior that users may depend on.

_If the fix doesn't apply cleanly to older branches or this analysis is incorrect, please ignore this suggestion; otherwise consider `pr-must-backport` or a version label like `v26.4-must-backport`._

PR #105007, opened 2026-05-15
udfs and table-functions deny lists

4ertus2:
- New Feature

add `udfs_deny_list` and `table_functions_deny_list` in config

The main reason is to disallow `remote` and `remoteSecure` separately from Distributed tables (both need REMOTE grant):

```
    <udfs_deny_list>
        <name>test_function_bash</name>
        <type>executable</type>
    </udfs_deny_list>

    <table_functions_deny_list>
        <name>remote</name>
        <name>remoteSecure</name>
    </table_functions_deny_list>
```


4ertus2:
It's also a possible to block UDF functions by type (only `executable` for now)

tests/queries/0_stateless/03742_functions_deny_list.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+# Tags: no-parallel, no-fasttest
+
+CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+CONFIG_FILE=$CURDIR/$(basename "${BASH_SOURCE[0]}" .sh).config.xml
+UDF_FILE=$CURDIR/../../integration/test_executable_user_defined_function/functions/test_function_config.xml
+
+mkdir -p $CURDIR/access_03742
```
george-larionov: please either clean up the dirs/files at the end or create them in CLICKHOUSE_TMP which is auto-cleaned.

tests/queries/0_stateless/03742_functions_deny_list.config.xml
```diff
@@ -0,0 +1,44 @@
+<clickhouse>
+    <logger>
+        <level>trace</level>
+        <log>clickhouse-server.log</log>
+        <errorlog>clickhouse-server.err.log</errorlog>
+    </logger>
+
+    <listen_host>127.1</listen_host>
+    <http_port>3742</http_port>
+    <tcp_port>9742</tcp_port>
```
george-larionov, line 10: I think hardcoding ports is a bit brittle, consider discovering a free port as done in https://github.com/ClickHouse/ClickHouse/blob/master/tests/queries/0_stateless/01954_clickhouse_benchmark_multiple_long.sh#L56

tests/queries/0_stateless/03742_functions_deny_list.sh
```diff
@@ -0,0 +1,52 @@
...
+. "$CURDIR"/../shell_config.sh
+
+CLICKHOUSE_WATCHDOG_ENABLE=0 $CLICKHOUSE_SERVER_BINARY --config-file=$CURDIR/config.xml &> $CURDIR/clickhouse-server.stderr &
+server_pid=$!
+
+CLICKHOUSE_URL="${CLICKHOUSE_PORT_HTTP_PROTO}://${CLICKHOUSE_HOST}:3742"
+
+# wait for the server to start accepting tcp connections (max 30 seconds)
+i=0 retries=30
+while [[ $i -lt $retries ]] && ! ${CLICKHOUSE_CURL} --max-time 1 -s ${CLICKHOUSE_URL} -d "SELECT 1" 2>&1 >/dev/null; do
```
george-larionov, line 28: If this times out then the test will potentially leave a running process

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
```diff
@@ -297,6 +302,10 @@ FunctionOverloadResolverPtr UserDefinedExecutableFunctionFactory::get(const Stri
 
 FunctionOverloadResolverPtr UserDefinedExecutableFunctionFactory::tryGet(const String & function_name, ContextPtr context, Array parameters)
```
george-larionov, line 437: Due to src/Analyzer/Resolve/resolveFunction.cpp:1426 this is also called for regular funcs, is this intentional? If so, then there is an issue where the check is skipped with `SETTINGS enable_analyzer=0`. If this deny_list is only meant for user defined and table funcs then it should also be renamed I think to something like `udf_and_table_functions_deny_list`

tests/queries/0_stateless/03742_functions_deny_list.sh
```diff
@@ -0,0 +1,52 @@
...
+${CLICKHOUSE_CLIENT} --port 9742 --query "CREATE FUNCTION IF NOT EXISTS test_sql_udf AS (value, dictionary) -> if(has(dictionary, value), value, '')"
+${CLICKHOUSE_CLIENT} --port 9742 --query "SELECT test_sql_udf('42', ['some_dict']) FROM system.one -- { serverError FUNCTION_NOT_ALLOWED }"
+${CLICKHOUSE_CLIENT} --port 9742 --query "SELECT test_function_bash(number) FROM system.numbers LIMIT 1 -- { serverError FUNCTION_NOT_ALLOWED }"
+
+${CLICKHOUSE_CLIENT} --port 9742 --query "SELECT * FROM remote('127.0.0.{1,2} LIMIT 1', system.numbers) -- { serverError FUNCTION_NOT_ALLOWED }"
+${CLICKHOUSE_CLIENT} --port 9742 --query "SELECT * FROM remoteSecure('127.0.0.{1,2} LIMIT 1', system.numbers) -- { serverError FUNCTION_NOT_ALLOWED }"
+${CLICKHOUSE_CLIENT} --port 9742 --query "SELECT * FROM REMOTE('127.0.0.{1,2} LIMIT 1', system.numbers) -- { serverError UNKNOWN_FUNCTION }"
+${CLICKHOUSE_CLIENT} --port 9742 --query "SELECT * FROM REMOTESECURE('127.0.0.{1,2} LIMIT 1', system.numbers) -- { serverError UNKNOWN_FUNCTION }"
+
+${CLICKHOUSE_CURL} -sS -i "${CLICKHOUSE_URL}" -d "SELECT * FROM remote('127.0.0.{1,2} LIMIT 1" | grep -o -F 'FUNCTION_NOT_ALLOWED'
```
george-larionov: Do these actually work? The unclosed `'` seems suspicious

george-larionov (changes requested):
Hi @4ertus2, please address the comments and add documentation.

rschu1ze:
@4ertus2 Normal users won't be able to decipher the changelog entry, kindly rewrite and expand it. Also, this PR lacks documentation.

4ertus2:
Yup, I'll improve it a bit later.

4ertus2:
@rschu1ze Could you also look at  https://bugcrowd.com/submissions/6f83cdf5-bc56-4e30-a68a-45ae9db97075

Why it's not a security issue? I've asked it in e-mail on security@clickhouse.com. But there's no answer for now.

4ertus2:
What about current PR. It breaks suggest on clickhouse-client init and `SELECT * FROM system.functions`. So I probably change the logic to return dummy function with exception in its execute phase instead.

george-larionov:
> What about current PR. It breaks suggest on clickhouse-client init and `SELECT * FROM system.functions`. So I probably change the logic to return dummy function with exception in its execute phase instead.

Sorry, I don't quite understand what you are asking.

4ertus2:
> Sorry, I don't quite understand what you are asking.

It's not a question. I'll update the PR a bit later with some additional logic.

rschu1ze:
What is the motivation and use case of this PR?

4ertus2:
> What is the motivation and use case of this PR?

Default ClickHouse grants allow too much you have to approve with security team.

It could be interested:
1. To disable `remote` and `remoteSecure` cause they could go to other nodes as `default` or `interserver` user. But you have to keep Distributed tables working.
2. To disable all executable UDF (but keep SQL UDFs allowed).
3. Allow some of functions at one cluster and disable them at another one.

Our security team needs more handles to disable dangerous things in ClickHouse's defaults.

tests/queries/0_stateless/03742_functions_deny_list.config.xml
4ertus2: This hack does not work in my internal infrastructure. So It's better to fix it inplace in PR after all the issues solved.

4ertus2:
> @4ertus2 Normal users won't be able to decipher the changelog entry, kindly rewrite and expand it. Also, this PR lacks documentation.

Could you show the files where to place documentation for this feature, please?

Is there any update for the security issue I've linked here? https://github.com/ClickHouse/ClickHouse/pull/105007#issuecomment-4507892397

rschu1ze:
> > @4ertus2 Normal users won't be able to decipher the changelog entry, kindly rewrite and expand it. Also, this PR lacks documentation.
> 
> Could you show the files where to place documentation for this feature, please?

Please create a sub-page either [here](https://clickhouse.com/docs/guides/manage-and-deploy-index) or [here](https://clickhouse.com/docs/operations/settings).

Can the function denylists also be made controllable from SQL? These days, all configuration is ideally reachable from XML and SQL.

> Is there any update for the security issue I've linked here? [#105007 (comment)](https://github.com/ClickHouse/ClickHouse/pull/105007#issuecomment-4507892397)

I don't have a login for Bugcrowd but I'll ping our security team again.

4ertus2:
> Can the function denylists also be made controllable from SQL? These days, all configuration is ideally reachable from XML and SQL.

It is very difficult to invent right SQL syntax for such things. I think it could be done by making system.functions table writable and adding "disabled" writable flag in it. But I'd prefer to make a separate issue/PR for it. Cause it is discussible solution.

PR #67135, opened 2024-07-25
Read configuration for clickhouse-local from ~/.clickhouse-local

azat:
### Changelog category (leave one):
- Improvement

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Read configuration for clickhouse-local from ~/.clickhouse-local

PR #105008, opened 2026-05-15
Fix clickhouse-local: apply user config from ~/.clickhouse-local/config.xml

### LINKED ISSUE #104932: User configurations from `~/.clickhouse-local/config.xml` are silently ignored

_Retrospective finding from a historical scan of [PR #67135](https://github.com/ClickHouse/ClickHouse/pull/67135) (merged 2024-08-05). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

When using the new `~/.clickhouse-local/config.xml` path, user configurations (users, profiles, quotas) are NOT applied. The setting `max_threads=42` in the config results in the default value (32) being used instead.

**Root cause:** LocalServer.cpp:595: The condition `getClientConfiguration().has("config-file") || fs::exists("config.xml")` doesn't account for config loaded via getLocalConfigPath(). When config is loaded from ~/.clickhouse-local/, this condition is false, so the user configuration handling is bypassed.

**Why we believe this is a bug:** LocalServer::initialize (line 241-247) → loads config via getLocalConfigPath() → config is added to getClientConfiguration() at line 254 → BUT setupUsers() at line 595 checks `getClientConfiguration().has("config-file") || fs::exists("config.xml")` which is FALSE for ~/.clickhouse-local configs → uses minimal_default_user_xml instead

**Affected locations:**
- `programs/local/LocalServer.cpp:595` — setupUsers condition for loading user config

**Impact:** Users who set up clickhouse-local configuration at ~/.clickhouse-local/config.xml with user settings, profiles, or quotas will have those settings silently ignored. The new feature advertised by this PR does not work correctly for user configurations.

Yes — confirmed on current `master` (commit `a6ca99beeb2`).

### How to reproduce

```sql
1. Create temp dir: FAKE_HOME=$(mktemp -d)
2. mkdir -p $FAKE_HOME/.clickhouse-local
3. Create config.xml with <user_directories> pointing to users.xml
4. Create users.xml with max_threads=42 in default profile
5. Run: HOME=$FAKE_HOME clickhouse-local --query "SELECT getSetting('max_threads')"
6. Observe: returns default value instead of 42
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/dff54059-4bdb-42c3-84cb-c47cb12ed621)

### Expected behavior

```
42
```

```
32
```

### Additional context

**Open risks:**
- The same bug may exist for ./clickhouse-local.xml path (first search path in getLocalConfigPath) - not tested
- The same bug may exist for /etc/clickhouse-local/config.xml path (third search path) - not tested

**Suggested fix:** Track whether a config was loaded from any path (not just --config-file or ./config.xml) and use that to determine if users config should be processed. One approach: store the actual loaded config path in a member variable and check if it's non-empty at line 595, or add a separate check for configs loaded via getLocalConfigPath().

Found during automated review of [PR #67135](https://github.com/ClickHouse/ClickHouse/pull/67135).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr67135_001`_

### Version info
- Resolved by: #105008
- Merged into: `26.5.1.814`

groeneai:
`clickhouse-local` silently dropped user-level configuration (`profiles`, `users`, `quotas`) when the config file was auto-discovered via `getLocalConfigPath`, the user-config path introduced by #67135.

`LocalServer::initialize` correctly loads the discovered config into `getClientConfiguration` (`programs/local/LocalServer.cpp:241-255`), but `LocalServer::setupUsers`'s gating condition only recognised configs from `--config-file` and `./config.xml`:

```cpp
if (getClientConfiguration().has("config-file") || fs::exists("config.xml"))
```

Every other discovery path (`./clickhouse-local.{xml,yaml,yml}`, `~/.clickhouse-local/config.{xml,yaml,yml}`, `/etc/clickhouse-local/config.{xml,yaml,yml}`) fell through to `minimal_default_user_xml`, ignoring `user_directories`, `users_config`, and all access-control settings sourced from the loaded file.

The fix tracks the actually loaded config file in a new `LocalServer::loaded_config_path` member, populated by `initialize` after a successful load. `setupUsers` uses it both as the unified gate and as the source of `config_dir`, so relative `user_directories.users_xml.path` entries resolve against the config's own directory.

Fix `clickhouse-local` silently ignoring user-level configuration (`profiles`, `users`, `quotas`, access-control settings) when the configuration file is auto-discovered from `./clickhouse-local.xml`, `~/.clickhouse-local/config.xml`, or `/etc/clickhouse-local/config.xml`. Previously these settings were applied only when `--config-file` was passed or `./config.xml` existed in the current directory; now all five discovery paths behave consistently.

Closes #104932

### Version info
- Merged into: `26.5.1.814`


groeneai:
cc @azat (you asked for this — Issue #104932) @alexey-milovidov @yakov-olkhovskiy (comp-local experts) — could you review?

**Pre-PR validation gate self-assessment** (session `cron:clickhouse-ci-task-worker:20260515-091500`):

| # | Gate | Result |
|---|------|--------|
| a | Deterministic repro? | ✅ `HOME=$X clickhouse local --query "SELECT getSetting('max_threads')"` returns built-in default (32) on master; returns config-pinned 42 with fix. |
| b | Root cause explained? | ✅ `LocalServer::initialize` loads config via `getLocalConfigPath` (`programs/local/LocalServer.cpp:241-255`), but `LocalServer::setupUsers`'s gate at line 595 only checks `has("config-file")` or `fs::exists("config.xml")`. For configs discovered via `getLocalConfigPath` (the documented user-config path), the gate is false and `setupUsers` falls through to `minimal_default_user_xml`, silently dropping `profiles`, `users`, `quotas`, access-control overrides. |
| c | Fix matches root cause? | ✅ Track the actually loaded path in `loaded_config_path` member, use it as the gate AND as the source of `config_dir` (so relative `user_directories.users_xml.path` resolves against the config's own directory rather than the cwd). Unifies all five sources (`--config-file`, `./config.xml`, `./clickhouse-local.{xml,yaml,yml}`, `~/.clickhouse-local/config.{xml,yaml,yml}`, `/etc/clickhouse-local/config.{xml,yaml,yml}`). |
| d | Test intent preserved? | ✅ New stateless test `04245_clickhouse_local_home_config.sh`, no existing test weakened. |
| e | Both directions demonstrated? | ✅ Locally reverted the fix → test FAILS; restored fix → test passes 10/10. |
| f | Fix is general, not narrow? | ✅ The bug exists for all `getLocalConfigPath` sources; the fix handles all of them via the unified `loaded_config_path` tracking. Verified manually on `~/.clickhouse-local/config.xml` and `./clickhouse-local.xml`. |

Symmetric-bug check: looked for similar gating logic in `programs/server/Server.cpp` — server has its own users.xml loading flow (driven by `users_config` configuration key), independent of this gate. Bug is specific to `clickhouse-local`.

tests/queries/0_stateless/04245_clickhouse_local_home_config.sh
```diff
@@ -0,0 +1,90 @@
...
+# via `getLocalConfigPath` (`~/.clickhouse-local/config.xml`, `./clickhouse-local.xml`,
+# `/etc/clickhouse-local/config.xml`). The `setupUsers` gate only recognized configs
+# from `--config-file` and `./config.xml`, falling through to the minimal default user
+# XML for every other discovery path.
+
+CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=../shell_config.sh
+. "$CURDIR"/../shell_config.sh
+
+TESTDIR="$CURDIR/04245_test_dir_$$"
```
Algunenano: @groeneai The common practice is to use USER_FILES_PATH.

tests/queries/0_stateless/04245_clickhouse_local_home_config.sh
```diff
@@ -0,0 +1,90 @@
...
+    cd "$CURDIR" || exit 1
+    rm -rf "$TESTDIR"
+}
+trap cleanup EXIT
+
+rm -rf "$TESTDIR"
+mkdir -p "$TESTDIR/home/.clickhouse-local"
+mkdir -p "$TESTDIR/cwd"
+
+# A config that points to a users.xml living next to it and sets `max_threads=42`
```
Algunenano: @groeneai I know we pay per output token, but there is no need to have such verbose comments for obvious things, not only here but in the PR as a hole. Comments are needed when you need to  explain something, like the one at line 86.

Algunenano (changes requested):

groeneai:
@Algunenano thanks for the review. Pushed 68c6a94 addressing both points:

1. **`USER_FILES_PATH`** — the test scratch dir is now `${USER_FILES_PATH}/${CLICKHOUSE_DATABASE}_104932` instead of `$CURDIR/04245_test_dir_$$`. The fake `HOME`, the `./clickhouse-local.xml` cwd, and the empty-`HOME` fallback all live under that one directory now, so there's a single `trap 'rm -rf "$TESTDIR"' EXIT`.
2. **Comments** — dropped the 6-line file header, the config explanation block, and the `must apply user config (regression for #104932)` style suffixes on the echo lines. Kept only the comment at line 75 that explains *why* the test asserts `> 0` rather than a fixed value.

PR body trimmed too — removed the inline reproducer / `setupUsers` excerpt and the per-bullet "Test" section, since the technical context is already in #104932 and the diff itself.

Local 10/10 with randomized settings on debug build.

Algunenano:
@groeneai Can you verify the claims from the automatic review (`clickhouse-gh`)?

groeneai:
@Algunenano The bot's claim is correct — confirmed with a concrete repro on this branch. Pushed commit `09b0a7bfedc` removing the `exists` guard.

### Repro (before this commit)

```
$ tree tmp/repro
tmp/repro
├── cwd
│   └── users.xml          # <-- rogue file in cwd, max_threads = 99
└── home/.clickhouse-local
    └── config.xml         # <user_directories><users_xml><path>users.xml</path></users_xml></user_directories>
                           # (no users.xml next to it)
```

```
$ HOME=tmp/repro/home  cd tmp/repro/cwd && clickhouse-local --query "SELECT getSetting('max_threads')"
99
```

Silent load of `./users.xml` from cwd. Same behavior under both `--config-file` and auto-discovery. A `users.xml` dropped into the user's working directory (any source — leftover from a previous run, a `git pull`, a project repo) can override the user's intended config — including `access_management` and password policies for `default`.

### After commit `09b0a7bfedc`

```
$ HOME=tmp/repro/home  cd tmp/repro/cwd && clickhouse-local --query "SELECT getSetting('max_threads')"
Code: 107. DB::Exception: Configuration file /…/tmp/repro/home/.clickhouse-local/users.xml doesn't exist and there is no embedded config. (FILE_DOESNT_EXIST)
```

Clear error pointing at the configured path. No silent fallback.

### Change

`programs/local/LocalServer.cpp`: drop the `fs::exists(config_dir / users_config_path)` guard. A relative `user_directories.users_xml.path` is now unconditionally rebased to the loaded config's directory; `ConfigProcessor::loadConfig` is then the single point that decides whether the file is present, and it raises `FILE_DOESNT_EXIST` with the configured absolute path on miss.

Aligns with the "avoid fallback paths" guidance in `.claude/CLAUDE.md`.

### Test

`tests/queries/0_stateless/04245_clickhouse_local_home_config.sh` now also asserts the no-silent-fallback behavior. The new sub-case fails on master (silently returns `99`) and passes after this commit (`FILE_DOESNT_EXIST`). Local: 10/10 with `--test-runs 10`.

Algunenano:
@groeneai Please check the review comments when you get them, and either reply there or fix them. Don't make me ask every time directly please

programs/local/LocalServer.cpp
```diff
@@ -592,11 +593,13 @@ void LocalServer::setupUsers()
...
     access_control.setThrowOnInvalidReplicatedAccessEntities(config.getBool("access_control_improvements.throw_on_invalid_replicated_access_entities", true));
 
-    if (getClientConfiguration().has("config-file") || fs::exists("config.xml"))
+    /// Apply user-level configuration if any config file was loaded in `initialize` — including
+    /// configs auto-discovered via `getLocalConfigPath` (`~/.clickhouse-local/config.xml` etc.),
+    /// not just those passed via `--config-file` or located at `./config.xml`.
+    if (!loaded_config_path.empty())
     {
-        String config_path = getClientConfiguration().getString("config-file", "");
+        const auto config_dir = fs::path{loaded_config_path}.remove_filename().string();
```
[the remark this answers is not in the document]
groeneai: Fixed in 34b83f6b7a45.

The bot is right: `users_config_path = getString("users_config", "")` was passed verbatim to `ConfigProcessor`, so a relative `<users_config>users.xml</users_config>` in `~/.clickhouse-local/config.xml` resolved against cwd, not the config's directory. Reproduced with auto-discovery (`HOME=…/home cd …/cwd && clickhouse local …`) and `--config-file=…`: both silently returned `max_threads = 99` from a rogue `./users.xml` planted in cwd.

Fix: moved the relative-path rebase out of the `user_directories` branch so it applies uniformly to both `users_config` and `user_directories.users_xml.path`. Same setup now raises `FILE_DOESNT_EXIST` for both modes.

Test `04245_clickhouse_local_home_config.sh` extended with a matching sub-case for the `users_config` form (the existing one only covered `user_directories.users_xml.path`). Verified locally: fails on master HEAD without the fix, passes 50/50 with full randomization with it.

groeneai:
@Algunenano sorry about the cadence — fair point, that's on me. Reviewed the open bot threads on this PR and pushed 34b83f6b7a45.

The bot's review on `programs/local/LocalServer.cpp:601` (the `users_config` form, posted 2026-05-15T09:53Z) was unaddressed — I had only handled the later one about `user_directories.users_xml.path` exists-guard in 09b0a7bfedc. The two issues turned out to be the same bug pattern in two sibling code paths:

| form | before | after |
|---|---|---|
| `<user_directories><users_xml><path>users.xml</path>` | rebased only when file existed next to config (fixed in 09b0a7bfedc) | always rebased to config dir |
| `<users_config>users.xml</users_config>` | not rebased at all — silently loaded `./users.xml` from cwd | always rebased to config dir |

**Repro of the second form** (before this commit):
```
~/.clickhouse-local/config.xml   # <users_config>users.xml</users_config>; no users.xml next to it
cwd/users.xml                    # rogue, max_threads = 99
HOME=…/home cd …/cwd && clickhouse local --query "SELECT getSetting('max_threads')"
# returns: 99   (silently loaded the rogue file)
```
After the commit: `Code: 107. FILE_DOESNT_EXIST: …/home/.clickhouse-local/users.xml doesn't exist`.

**Fix:** moved the `if (fs::path(users_config_path).is_relative()) users_config_path = config_dir / users_config_path;` rebase out of the `user_directories` branch so it applies uniformly to whichever path was chosen (`users_config` wins if both are set, otherwise `user_directories.users_xml.path`). Also trimmed the prior 5-line security comment to 3 lines, per your earlier feedback on comment verbosity.

**Test:** `04245_clickhouse_local_home_config.sh` extended with a `<users_config>` form sub-case symmetric to the existing `<user_directories>` orphan_home one. Fails on master HEAD without the source fix (rogue file silently loaded → output is empty, expected `FILE_DOESNT_EXIST`); passes 50/50 with full randomization with the fix.

**Pre-PR Validation Gate** (session `cron:clickhouse-ci-task-worker:20260518-091500`):

| # | Gate | Result |
|---|------|--------|
| a | Deterministic repro? | ✅ Both auto-discovery and `--config-file` modes return 99 (rogue) before fix, `FILE_DOESNT_EXIST` after, with the exact setup above |
| b | Root cause explained? | ✅ `getString("users_config", "")` was passed verbatim to `ConfigProcessor`, asymmetric with the sibling `user_directories.users_xml.path` rebase |
| c | Fix matches root cause? | ✅ Single unified rebase covers both forms, removes the asymmetry |
| d | Test intent preserved? | ✅ Existing 5 sub-cases unchanged; new sub-case is symmetric with the existing orphan_home one |
| e | Both directions demonstrated? | ✅ Fails on master HEAD (showed diff: missing `FILE_DOESNT_EXIST` line for new sub-case); passes 50/50 with full randomization with fix |
| f | Fix is general? | ✅ Same bug pattern in two sibling code paths (`users_config` and `user_directories.users_xml.path`); fix unifies them rather than patching one site |

Branch also merged with master (`327278bd6ea3`) since it had been sitting for a few days.

I'll be more proactive about bot threads on this PR going forward — the cadence on this one was bad, sorry.

Algunenano (approved):

PR #104812, opened 2026-05-13
Preserve parameter types in timeseries aggregate functions, add type annotations

vitlibar:
Reverts ClickHouse/ClickHouse#104672, with extra improvements.
This PR is an improved version of https://github.com/ClickHouse/ClickHouse/pull/103428

This PR adds:
**1. Preserving parameter types in `timeSeries*` aggregate functions**

We need to restore the initial invariant: For any aggregation function `IAggregateFunction::getParameters()` should return exactly the parameters used to create the aggregation function. Aggregation functions are not allowed to change or normalize these parameters.

Otherwise it would become a different `DataTypeAggregateFunction` (see `DataTypeAggregateFunction::strictEquals`),
and could fail to reattach a table because `decodeDataType()` would reconstruct a type different from the one
recorded in the column metadata; or fail on parallel replicas under serialize_query_plan=1 because the replica's
`decodeDataType()` would reconstruct a type different from the one the coordinator sent in the plan. And we actually saw a bunch of such issues already (https://github.com/ClickHouse/ClickHouse/issues/101407, https://github.com/ClickHouse/ClickHouse/pull/102345, https://github.com/ClickHouse/ClickHouse/pull/103079)

It's difficult to fix all these issues and it's better to just restore the invariant. By the way, for non-timeseries aggregate functions parameter types are always preserved.

**2. Parameters are now printed with types in `timeSeries*` aggregate functions.**

Another important change is that for the `timeSeries*` aggregate functions, the AggregateFunction(...) type name now prints parameters in a type-annotated form (`'<value>'::TypeName`) instead of a bare literal that could be ambiguous. For example this is how a data type is printed now
```
AggregateFunction(timeSeriesInstantRateToGrid('1734004810'::Decimal64(3), '1734004860'::Decimal64(3), 10, 60), Array(DateTime64(3, 'UTC')), Array(Float64))
```
and earlier it was printed this way:
```
AggregateFunction(timeSeriesInstantRateToGrid('1734004810', '1734004860', 10, 60), Array(DateTime64(3, 'UTC')), Array(Float64))
```
Bare literals were ambiguous because '1734004810' could mean a String or a Decimal with a different scale. Form `'<value>'::TypeName` allows to parse exactly the same parameters back.

Closes https://github.com/ClickHouse/ClickHouse/issues/104459

### Version info
- Merged into: `26.6.1.98`

PR #105010, opened 2026-05-15
Populate ProfileEvents for executable_pool UDFs

XuJia0210:
Stacks on top of #104998. Adds the eighth `ExecutableUserDefinedFunction*` ProfileEvent (`PoolWaitMicroseconds`) and populates all eight on the `executable_pool` UDF path, so a query that calls a pool UDF shows up in `system.query_log.ProfileEvents` with per-call invocation count, wall and pool-wait time, child-process CPU and peak memory, and stdin/stdout bytes.

- Improvement

Populate the `ExecutableUserDefinedFunction*` `ProfileEvents` counters for `executable_pool` UDFs. Per-call invocation count, wall and pool-wait time, child-process CPU and peak memory, and stdin/stdout bytes are now visible in `system.query_log.ProfileEvents`.

### Version info
- Merged into: `26.6.1.377`


src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,326 @@
...
+    if (root_pid <= 0)
+        return result;
+
+    result.push_back(root_pid);
+
+#if defined(OS_LINUX)
+    /// Iterative DFS: for every pid in `result`, append the union of
+    /// `/proc/<pid>/task/<tid>/children` lists. A bounded depth limit guards
+    /// against pathological /proc states; UDFs in practice are 3 levels deep
+    /// (execute.sh -> nsjail -> python).
```
andreev-io: This is not relevant for OSS.

src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,326 @@
...
+    root_pid = root_pid_;
+    borrow_acquired = true;
+    pre_snapshot.clear();
+
+    if (root_pid <= 0)
+        return;
+
+    /// Pre-snapshot: walk subtree, reset VmHWM per pid, capture utime/stime
+    /// baselines. clearRefs failure is silent — peak_rss for that pid will be
+    /// reported as the kernel's lifetime peak, which over-counts slightly but
```
andreev-io: If we fail to clear, why would the value be the kernel's lifetime peak? I don't think that's true.

andreev-io:
@groeneai Can you review this PR?

groeneai:
@andreev-io thanks for the ping. I went through the code and the procfs semantics. Here are answers to your two inline comments plus a few other things I noticed.

### 1. `UDFProcessSubtreeSampler.cpp` ~line 256, "kernel's lifetime peak"

You are right to push on this. The wording is imprecise.

`clear_refs` mode 5 (`CLEAR_REFS_MM_HIWATER_RSS`, since Linux 4.0) does not zero `VmHWM`. It snaps the high-water mark to the process's current RSS:

```c
// fs/proc/task_mmu.c
if (type == CLEAR_REFS_MM_HIWATER_RSS) {
    // Writing 5 to /proc/pid/clear_refs resets the peak
    // resident set size to this mm's current rss value.
    reset_mm_hiwater_rss(mm);
    ...
}

// include/linux/mm.h
static inline void reset_mm_hiwater_rss(struct mm_struct *mm)
{
    mm->hiwater_rss = get_mm_rss(mm);
}
```

So when `clearRefs` fails (open returns -1, or the write itself fails: typical on procfs `hidepid`, on uid mismatches between server and worker, on the race where the worker died), `VmHWM` is unchanged. What you read back at release depends on the borrow history:

- Brand-new pool worker, first borrow: VmHWM = peak RSS since process start. That happens to equal "lifetime peak", so the current comment is true.
- Long-lived pool worker, clearRefs succeeded on a previous borrow: VmHWM = peak RSS since that previous successful clear. Not lifetime peak.
- Long-lived worker, clearRefs has never succeeded: VmHWM = peak since process start, again lifetime peak.

The "over-counts slightly" framing also understates the bad case. If a previous borrow hit a 4 GiB peak and clearRefs has been failing every borrow since, a 100 MiB borrow will report ~4 GiB. The inflation is bounded by the worst peak across all prior un-cleared borrows, which can be arbitrarily large depending on workload, not "slight".

Concrete suggestion for the block:

```
/// Pre-snapshot: walk subtree, reset VmHWM per pid via clear_refs(5), and
/// capture utime/stime baselines. clear_refs(5) does not touch utime/stime,
/// only VmHWM, so the order of the two reads is fine.
///
/// clearRefs failure is silent. When it fails, VmHWM is left at whatever it
/// was on entry: the peak RSS since process start (if no successful clear
/// has ever run) or since the most recent successful clear_refs(5). The
/// borrow-attributed peak is then an upper bound, but the over-count is
/// bounded by the worst peak across all prior borrows since the last
/// successful clear, which can be arbitrarily large depending on workload.
```

Optional follow-up: a log-once on first clearRefs failure per pool worker would let operators distinguish trustworthy samples from inflated ones. Out of scope for this PR.

### 2. `UDFProcessSubtreeSampler.cpp` ~line 50, "execute.sh -> nsjail -> python"

Agreed, that example leaks a private-side topology that does not exist in OSS. The `MAX_PIDS` guard is independent of any sandbox, so keep the guard, drop the example.

The same concern repeats at ~line 309 in `recordReleased`:

```
/// because pids run mostly serially inside one borrow (nsjail+exec
/// chain) so peak resident set per pid is a better estimate than a sum.
```

Suggested replacements:

```
/// UDF process subtrees are shallow in practice (typically a launcher shell
/// or runtime wrapper plus the interpreter), but MAX_PIDS guards against
/// pathological /proc states.
```

```
/// For peak memory we take the max VmHWM observed across the subtree:
/// processes in a UDF subtree typically run serially in an exec chain, so
/// peak resident set per pid is a better estimate than a sum, which would
/// double-count the launcher and the interpreter.
```

### 3. ProfileEvent description vs. wiring contract

The AI Review bot flagged this and it is real. The event descriptions in `ProfileEvents.cpp` (introduced in #104998) say "executable and executable pool", but the increments in `UserDefinedExecutableFunctionFactory::executeImpl` only fire inside the `if (coordinator_configuration.is_executable_pool)` branch.

Two paths forward:

a. Tighten the descriptions to `executable_pool`-only wording in this PR. Minimal, lets it ship.

b. Also instrument the non-pool `executable` path here. Doable but more invasive: that path uses the same `ShellCommandSource` machinery without the borrow boundary, so you would skip `recordPoolWaitDone` and call `recordPidAcquired` right after `ShellCommand::execute()` returns the pid.

I would do (a) in this PR and track (b) as a follow-up. The descriptions are part of the documented metric contract for `system.events`, and a future rewording costs more than getting the labels right now.

### 4. Minor

- `peak_memory_byte_seconds = peak_rss * elapsed_us / 1000000ULL`: the comment's overflow boundary (~1.8e13 byte-seconds) is accurate. Fine.
- Atomic `input_bytes`/`output_bytes` with `memory_order_relaxed`: safe because `recordReleased` runs after `send_data_threads` are joined in `cleanup()`. Fine.
- `readStat` skips 11 tokens to land on `utime`. The procfs field layout (`man 5 proc`, `/proc/[pid]/stat`) has been stable since 2.6, but a one-line comment citing the man page would help future readers.
- `walkSubtree` uses `std::find` over a `vector` for dedup, O(N^2) with N capped at MAX_PIDS = 1024. Fine.

### Verdict

The design is sound. The sampler is scoped per borrow, lifecycle hooks are placed correctly (`recordPoolWaitDone` before the timeout check, `recordPidAcquired` right after `buildCommand`, `recordReleased` before the worker is returned or torn down), the `SCOPE_EXIT` in the factory drains on both success and exception paths, and the integration test covers each ProfileEvent independently.

My recommendation is to block on:

1. The two comment rewrites above (kernel-peak semantics and OSS-relevance).
2. Aligning the `ProfileEvents.cpp` descriptions with what the code actually populates (either narrow to `executable_pool` or instrument both paths).

Everything else is a nit or a follow-up. Once those land I would be happy to re-review.

src/Processors/Sources/ShellCommandSource.cpp
```diff
@@ -526,6 +539,13 @@ namespace
                 if (thread.joinable())
                     thread.join();
 
+            /// Resource accounting must observe the borrow's resident set before
+            /// the worker is torn down or the slot is handed back to the pool —
+            /// either path destroys `/proc/<pid>/{stat,status}` and the sampler
+            /// would then read zero CPU and zero `VmHWM`.
+            if (configuration.sampler)
+                configuration.sampler->recordReleased();
```
vdimir: `recordReleased` is quite non-trivial method and may throw exception and at the same it can be called from destructor above calling cleanup. The least we can do is to wrap it into try catch, if it would be hard to decompose logic a somehow to avoid reading procfs directly from `cleanup`.

src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,326 @@
...
+        user_time_us = post_utime_sum_in_baseline - pre_utime_sum;
+    if (post_stime_sum_in_baseline >= pre_stime_sum)
+        system_time_us = post_stime_sum_in_baseline - pre_stime_sum;
+
+    /// PeakMemoryByteSeconds = peak_rss × borrow_wall_seconds.
+    /// Stored as integer byte-seconds. 64-bit multiplication is fine for any
+    /// realistic UDF (1 GiB × 100 s ≈ 10^14, well below 2^64); pathological
+    /// inputs above ~1.8e19 byte-microseconds would overflow before the
+    /// divide.
+    peak_memory_byte_seconds = peak_rss * elapsed_us / 1000000ULL;
```
vdimir: If we fine with overflow, let's handle it explicitly with special function e.g. `common::mulOverflow`, so in case it happens in CI UBSanitizer does not report an error, where overflow is expected and could be handled

src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,326 @@
...
+        struct dirent * entry;
+        /// NOLINTNEXTLINE(concurrency-mt-unsafe) -- `dir` is a local `DIR *` not shared across threads.
+        while ((entry = ::readdir(dir)) != nullptr)
+        {
+            const char * name = entry->d_name;
+            if (name[0] == '.')
+                continue;
+            /// `name` is a thread id (`tid`) under the process.
+            const std::string children_path = task_dir + '/' + name + "/children";
+            std::ifstream in(children_path);
```
vdimir, line 79: Usually we read files with wrappers like `ReadBufferFromFile/ReadBufferFromFilePRead` so maybe using that would be more consistent with other parts of codebase. We may also check how procfs is read for example for some async metrics and alight with that.

src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,326 @@
...
+    pre_snapshot.clear();
+
+    if (root_pid <= 0)
+        return;
+
+    /// Pre-snapshot: walk subtree, reset VmHWM per pid, capture utime/stime
+    /// baselines. clearRefs failure is silent — peak_rss for that pid will be
+    /// reported as the kernel's lifetime peak, which over-counts slightly but
+    /// is correct as an upper bound.
+    auto pids = UDFProcfs::walkSubtree(root_pid);
```
vdimir: Overall, do you think reading from procfs may add visible overhead for existing fast UDFs?

Should we consider enabling metric collection only behind a setting/config/flag, possibly opt-out, to prevent potential issues in the new version (with performance or something else)? Or do you think it should be fine in all cases?

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
```diff
@@ -203,12 +219,52 @@ class UserDefinedFunction final : public IFunction
...
+            /// sampler holds the final counters by the time this guard runs.
+            SCOPE_EXIT({
+                if (!sampler)
+                    return;
+
+                /// `PoolWaitMicroseconds` fires whenever `tryBorrowObject`
+                /// returned (success OR timeout). On timeout the throw
+                /// short-circuits the borrow but pool contention still
+                /// happened and is worth reporting.
+                if (sampler->poolWaitDone())
```
vdimir, line 283: Overall, some measurements could be done in place. For example, for `ExecutableUserDefinedFunctionPoolWaitMicroseconds`, we could measure the time directly around the `process_pool->tryBorrowObject` call using `ProfileEventTimeIncrement<Microseconds>`.

Probably the same applies to wall time, although it is a bit more complicated since start and stop happen in different classes. Maybe it is still doable by passing a `Stopwatch` or `ProfileEventTimeIncrement` between them.

While I understand the general idea, I still have a feeling that the state machine in `UDFProcessSubtreeSampler` could be simpler, though maybe that is not actually the case.

src/Common/UDFProcessSubtreeSampler.cpp
XuJia0210: Thanks. I considered switching to ReadBufferFromFilePRead, but kept the current std::ifstream-based approach intentionally:
1. pid may disappear ifstream's !is_open() lands silently; RBFPR throws and needs try/catch
2. readStat/readPeakRss are noexcept

Happy to revisit if you'd prefer the explicit RBFPR style anyway — it would be a mechanical wrap.

src/Common/UDFProcessSubtreeSampler.cpp
XuJia0210: Measured locally with temporary sampler_self_us instrumentation around recordPidAcquired and recordReleased. Average per-block cost across
  ~5–10 borrows (first cold-start sample excluded for hot-state numbers; all four pool UDFs have pids=1, no helper subprocesses):

  - echo (5000 rows / block): ElapsedMicroseconds ≈ 9.8 ms, sampler procfs cost ≈ 143 μs — 1.46 %
  - cpu (2000 rows / block): ElapsedMicroseconds ≈ 866 ms, sampler procfs cost ≈ 169 μs — 0.020 %
  - mem (500 rows / block): ElapsedMicroseconds ≈ 9.68 s, sampler procfs cost ≈ 187 μs — 0.0019 %
  - syscall (2000 rows / block): ElapsedMicroseconds ≈ 63.7 s, sampler procfs cost ≈ 180 μs — 0.0003 %

  The per-borrow sampler cost is roughly constant (~150–200 μs of procfs syscalls, scaling linearly with the number of PIDs in the subtree); the
   ratio is dominated by the UDF's own runtime. Worst case here is the trivial echo UDF at ~1.5 %; any pool UDF doing real work is well below
  0.1 %.

  A couple of reasons I'd rather not gate this behind a setting:

  - Consistent with the codebase. AsynchronousMetrics and other procfs readers are always-on; the pool sampler reads procfs only once per
  borrow, so it's actually lighter than the existing async-metric loop.
  - The PR perf-comparison CI is green — if procfs sampling cost something visible it would show up there.

  If a measurable regression turns up after merge (perf-tests on master, or any user report), I'd be happy to add an opt-out setting then.

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
XuJia0210: Good question. PoolWaitMicroseconds and ElapsedMicroseconds could indeed be inlined with ProfileEventTimeIncrement<Microseconds> (the latter would just need a Stopwatch threaded from the factory into ShellCommandSource).

But CPU user/system time and PeakMemoryByteSeconds need persistent state across the borrow: a pre-walk of the subtree to clear VmHWM and snapshot utime/stime, then a post-walk to compute the delta. That's what drives the UDFProcessSubtreeSampler state machine; without it the
  CPU/memory counters can't exist.

  Given the sampler has to exist anyway, putting PoolWait/Elapsed in it too keeps all eight counters consistent rather than splitting them across two mechanisms.

  Happy to move PoolWaitMicroseconds to inline ProfileEventTimeIncrement if you'd prefer — just say the word.

groeneai: Keep it in the sampler.

The sampler is constructed in `executeImpl` before `tryBorrowObject`, so its ctor already starts the entry watch for free. Inlining a sibling `ProfileEventTimeIncrement<Microseconds>` around the same `tryBorrowObject` call would set up a second stopwatch for the same wait, and it would have to duplicate the timeout handling that `recordPoolWaitDone` already covers (the sampler fires `PoolWait` even when `tryBorrowObject` throws on timeout, which I think you want).

The other four counters (`Elapsed`, `UserTime`, `SystemTime`, `PeakMemoryByteSeconds`) need the pre-snapshot/post-snapshot pair around the borrow, so the sampler has to exist regardless. Emitting all eight ProfileEvents from one `SCOPE_EXIT` keeps a single fire point and a single source of truth.

If the state machine itself can be slimmed down, that feels like a follow-up rather than a precondition for this PR.

vdimir:
> [AST fuzzer (arm_asan_ubsan)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105010&sha=a5c69584da644e01a6cf4a88f85a36d4a6ad1839&name_0=PR&name_1=AST%20fuzzer%20%28arm_asan_ubsan%29) 		FAIL 		
> 	Logical error: AggregateFunctionArray: parameters mismatch between Array wrapper 'A' and nested function 'B'. Wrapper has C parameter(s): [D], nested function has E parameter(s): [F] (STID: 4870-4d73) 	FAIL 		

https://github.com/ClickHouse/ClickHouse/pull/104812


> [Finish Workflow](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105010&sha=a5c69584da644e01a6cf4a88f85a36d4a6ad1839&name_0=PR&name_1=Finish%20Workflow) 		FAIL 		
> 	python3 ./ci/jobs/scripts/workflow_hooks/check_report_messages.py 	FAIL

CIDB was down

src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,338 @@
...
+    ssize_t written = ::write(fd, data, sizeof(data) - 1);
+    (void)written;
+    [[maybe_unused]] int close_err = ::close(fd);
+#else
+    (void)pid;
+#endif
+}
+
+
+bool readStat(pid_t pid, UInt64 & utime_us, UInt64 & stime_us) noexcept
```
andreev-io: Would it be useful to have this as a generic function, parsing the output into a generic struct, unrelated to UDF?

src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,338 @@
...
+    /// the CPU delta — pids that appear after the pre-walk are ignored. This
+    /// slightly under-attributes a borrow that spawns brand-new descendants,
+    /// but avoids the catastrophic over-attribution that would otherwise
+    /// happen if `readStat` failed during the pre-walk for a warm worker:
+    /// the worker's lifetime-cumulative `utime/stime` would land in the post
+    /// sum with no baseline to subtract. Conservative under-attribution is
+    /// always preferable to silent over-attribution.
+    ///
+    /// For peak memory we take the max VmHWM observed across the subtree
+    /// because pids run mostly serially inside one borrow (a parent process
```
andreev-io: > because pids run mostly serially inside one borrow (a parent process hands off to a child) so peak resident set per pid is a better estimate than a sum

While I agree with the conclusion (measure the highest peak, ignore the rest), I don't agree with the reasoning. I don't think we can claim anything about how the children are running (serially or otherwise). It's entirely up to the UDF code.

I would still reach for the same measurement, though (the one peak instead of a sum of peaks) just because the sum of peaks means nothing (we don't know if processes even ran concurrently or not). The max peak is useful by itself.

src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,338 @@
...
+        if (UDFProcfs::readPeakRss(pid, hwm_bytes))
+            peak_rss = std::max(peak_rss, hwm_bytes);
+    }
+
+    if (post_utime_sum_in_baseline >= pre_utime_sum)
+        user_time_us = post_utime_sum_in_baseline - pre_utime_sum;
+    if (post_stime_sum_in_baseline >= pre_stime_sum)
+        system_time_us = post_stime_sum_in_baseline - pre_stime_sum;
+
+    /// PeakMemoryByteSeconds = peak_rss × borrow_wall_seconds. Stored as
```
andreev-io, line 400: Here, I would explain in a comment why we are converting Bytes into ByteSeconds. The reason is that the resulting metric should be additive across UDF invocations for the query. Peak bytes are not additive. PeakByteSeconds are.

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
```diff
@@ -203,12 +219,52 @@ class UserDefinedFunction final : public IFunction
...
+            std::shared_ptr<UDFProcessSubtreeSampler> sampler;
             if (coordinator_configuration.is_executable_pool)
             {
                 shell_command_source_configuration.read_fixed_number_of_rows = true;
                 shell_command_source_configuration.number_of_rows_to_read = input_rows_count;
+
+                /// Count every executeImpl call against the pool path, even those
+                /// that fail before/inside the borrow. Resource counters below
+                /// only fire when the borrow actually completed.
+                ProfileEvents::increment(ProfileEvents::ExecutableUserDefinedFunctionInvocations);
```
andreev-io, line 262: Do I understand correctly that this would always be equal to the number of blocks processed by the query?

andreev-io (commented):
Functionally, this looks good to me. However, I have a reservation about a tactical decision in the design re CPU. Our goal is to accumulate CPU usage across the parent process and the children. There are two ways to do this: (1) by reading the parent's `utime + cutime + stime + cstime` (this ignores children that the parent doesn't reap, but such children can be ignored – they are likely a result of a bug in the parent anyway) and (2) by walking the subtree (this misses the children that have completed by the time the parent returns the results, which is most likely in a reasonably designed UDF... all children).

I think approach (2) is just plain worse because it records less accurate data, and on top of that it is more complex (requires a tree walk). This PR's argument in favour of (2) is that (1) risks catastrophic overestimation if we fail to record the right values for `cutime` and `cstime` before the borrow. There are a couple issues with this argument:
- (a): The same overattribution risk actually applies to `utime` and `stime`, and yet we don't worry about it at all.
- (b): This problem has a simple solution: don't record CPU usage at all if we fail to pre-read all 4 metrics (utime, stime, cutime, cstime). Log warnings.

~~I am strongly in favour of approach (1): it is simpler (no tree walk) and more accurate (records children that have completed, which is almost guaranteed to be all children).~~

Memory measurement decisions look good to me.

@groeneai Evaluate my framing and the two options around CPU measurements.

**UPD: My proposal above is not good.** It makes accounting worse for long-running children, which can exist with executable pool:
>   Two failure modes introduced by only measuring 4 metrics for the parent:
  1a. Persistent helper alive throughout the borrow. Common pattern given that executable_pool exists to amortise startup: multiprocessing.Pool workers, persistent JVM, model server, sandbox wrappers. The helper does work, accumulates utime/stime of its own, is never reaped. Root's cutime/cstime doesn't change. Approach (1) reports zero for it.
 1b. Persistent helper dies during a borrow. Root reaps it. root.cutime jumps by the helper's entire lifetime utime + cutime, including everything from before this borrow. Approach (1) attributes that pre-borrow history to the current borrow — overcounts, potentially by hours.
 
 There's a 3rd option, which is best:
 > Approach (3): subtree walk AND read cutime/cstime per pid. Captures:
 >  - Root's own work (via root's utime/stime delta in the walk)
 >  - Persistent helpers alive throughout (via their own utime/stime delta in the walk)
 >  - Per-row helpers reaped during the borrow (via the reaper's cutime/cstime delta — the reaper is in the walk)
 >  - Multi-level reaping chains (via cutime/cstime propagating up; verified by the worked example I showed earlier)

 >  It's the best of both: the walk catches what approach (1) misses (live persistent descendants); cutime/cstime catches what approach (2)-as-currently-written misses (reaped descendants).

 >  Cost over the PR's current state: extend readStat to also parse fields 16/17 and add them to the per-pid contribution. Two extra tryReadIntText calls. The pre_snapshot map and the "skip pid if pre-snapshot missing" guard stay exactly as they are.
 
 
 @vdimir @XuJia0210 @groeneai What do you think about idea (3)?

src/Common/UDFProcessSubtreeSampler.cpp
andreev-io: Actually, given that we are aiming to provide an upper bound, maybe summing peaks is actually a better decision. I could be convinced of that. @vdimir what do you think?

vdimir: Maybe we can calculate just single peak? We could keep somewhere previous_peak and increment profile event by `max(0, current_peak  previous_peak)`, so it corresponds to maximal peak usage

andreev-io: That's what we do already thanks to `clear_refs`. There's not a need to do the subtraction even. The question is how do we treat the fact that the UDF may have subchildren with their own peaks – do we sum over them or max over them?

andreev-io:
Following up on [my earlier review](https://github.com/ClickHouse/ClickHouse/pull/105010#pullrequestreview-4336172989). I opened an RFC at #105523 capturing the broader design; the items below are the concrete deltas against this PR. All fit inside the existing files.

**Correctness** (`UDFProcessSubtreeSampler.{h,cpp}`):

1. **Read `cutime` / `cstime` (fields 16/17)** in `readStat` and sum `(utime + cutime, stime + cstime)` per pid. Without this, CPU from descendants reaped during the borrow — including per-row `subprocess.run` / shell-pipeline patterns — is dropped.
2. **Compute `pre_sum` from the full `pre_snapshot` map**, not from inside the post-walk loop. Otherwise a pre-existing descendant that dies during the borrow over-attributes by its pre-borrow history.
3. **Track `pre_walk_pids` separately** and split the post guard into three buckets: `pid in pre` (delta) / `pid in pre_walk_pids but not in pre` (pre-read failure — skip) / `pid in neither` (born during borrow — include `post` fully). The current single guard silently drops CPU from descendants born during the borrow (e.g. lazily-initialised `multiprocessing.Pool`).

**Operability**:

4. **Log a one-shot `LOG_WARNING` per UDF on first procfs failure** in `clearRefs`/`readStat`/`readPeakRss`. Today restricted containers silently degrade the metrics with no operator-visible signal.

**Design decision**:

5. **Decide how to aggregate `peak_rss` across live descendants** — `max()` (current PR) or `sum()`. Both have known failure modes for different UDF process shapes. Either is acceptable for v1; just needs an explicit pick documented in the code.

andreev-io:
We also need to add some tests that cover the complexities around measuring children's resource usage.

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
XuJia0210: No, not every block in a query will call UDF.

src/Common/UDFProcessSubtreeSampler.cpp
XuJia0210: Let's move when there is a second use.
ClickHouse already has a read_procfs function, but it's not recursive.
This time, we need a recursive function, so I added it here.
There's a chance this function is only called here in the future

XuJia0210: Following up: happy to extract `readStat` into a generic procfs helper now if you prefer that over waiting for a second consumer. Just say the word.

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
XuJia0210: Concrete counter-example: `SELECT count() FROM (SELECT udf(number) FROM numbers(N))` rewrites to `SELECT count() FROM numbers(N)` because the outer `count` does not reference the UDF column, so the UDF is never invoked and `Invocations` stays at 0 even though `N` rows worth of blocks are processed. Constant folding, lazy-evaluated arms of `if`/`multiIf`, and downstream-column pruning have the same effect. The counter measures actual UDF entries, which is what we want for resource accounting.

vdimir (approved):
Is test for child process accounting only thing missing in this PR? In this case we can merge and do a followup

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
andreev-io: Got it.

andreev-io:
@vdimir IMO there are some minor functional changes left too: https://github.com/ClickHouse/ClickHouse/pull/105010#issuecomment-4509090575

src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+    if (root_pid <= 0)
+        return;
+
+    /// Post-snapshot: re-walk because the subtree may have grown (e.g. python
+    /// launching a helper). Only pids that have a pre-snapshot contribute to
+    /// the CPU delta — pids that appear after the pre-walk are ignored. This
+    /// slightly under-attributes a borrow that spawns brand-new descendants,
+    /// but avoids the catastrophic over-attribution that would otherwise
+    /// happen if `readStat` failed during the pre-walk for a warm worker:
```
HanziJiang: If I understood correctly, for CPU time, for PIDs that are only in post snapshot, the reason why we can't simply assume the baseline is just 0 is that PID absent from pre snapshot could be due to two things:
1. Truly new process
2. Pre-existing but readStat failed

Maybe we can track all PIDs visited in the pre-walk, regardless of whether readStat succeeded. If it is not in pre-walk, we know for sure its case 1, and we can assume the baseline is 0.

src/Common/UDFProcessSubtreeSampler.cpp
HanziJiang: I think if we define the metrics to be an upper bound, then max might under estimate for concurrent processes, so using sum might be more "correct" as an upperbound.

HanziJiang: Either way, for this memory usage, I think we can't easily cover processes that died before the post walk, so it might not strictly be an upperbound anyways. I can't think of an easy solution.

src/Common/UDFProcessSubtreeSampler.cpp
```diff
@@ -0,0 +1,352 @@
...
+        return result;
+
+    result.push_back(root_pid);
+
+#if defined(OS_LINUX)
+    /// Iterative DFS: for every pid in `result`, append the union of
+    /// `/proc/<pid>/task/<tid>/children` lists. A bounded depth limit guards
+    /// against pathological /proc states; UDFs in practice nest only a few
+    /// levels deep (a wrapper script may spawn the actual interpreter).
+    constexpr size_t MAX_PIDS = 1024;
```
HanziJiang, line 59: Can we show a warning when MAX_PIDS is hit?

src/Common/UDFProcessSubtreeSampler.cpp
XuJia0210: Went with max() — a sum of per-pid peaks isn't meaningful since they don't coincide in time.

XuJia0210:
@andreev-io All five deltas are in, plus the MAX_PIDS warning, the pre-snapshot exception guard, and child-reaping tests. PTAL.

andreev-io (approved):
Looks good. Please test it.

vdimir (approved):

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
```diff
@@ -203,12 +252,89 @@ class UserDefinedFunction final : public IFunction
...
+                }
+
+                /// If any of the three procfs reads silently failed during
+                /// this borrow, or the subtree walk hit the MAX_PIDS cap,
+                /// surface it once per UDF + op so an operator running under
+                /// a restrictive seccomp profile (or a UDF spawning more
+                /// descendants than the sampler can enumerate) has a visible
+                /// signal that the resource counters they see are degraded.
+                if (sampler->clearRefsFailedAnyPid() || sampler->readStatFailedAnyPid() || sampler->readPeakRssFailedAnyPid() || sampler->subtreeWalkTruncated())
+                {
```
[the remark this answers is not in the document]
vdimir: Let's use SCOPE_EXIT_SAFE then

PR #105020, opened 2026-05-15
Add filesystem cache eviction metrics

sacheendra:
- Improvement

### Changelog entry
Added opt-in Prometheus metrics for filesystem cache eviction activity (`filesystem_cache_evictions_total`, `filesystem_cache_evicted_bytes_total`, `filesystem_cache_evicted_segment_hits`, `filesystem_cache_evicted_segment_size_bytes`, and their per-user variants labeled with `user_id`), exposed per cache via `system.dimensional_metrics`, `system.histogram_metrics`, and the Prometheus endpoint. They are controlled by the cache disk-config settings `expose_prometheus_eviction_metrics` and `expose_prometheus_eviction_metrics_per_user` (both off by default), which can be toggled at runtime via `SYSTEM RELOAD CONFIG`.

---

### Description

Adds metrics that describe `FileCache` eviction activity, registered via the existing `DimensionalMetrics` / `HistogramMetrics` infrastructure and exposed through the Prometheus endpoint. All metrics are disabled by default; enable per-cache via `expose_prometheus_eviction_metrics`.

| Name | Type | Labels |
|---|---|---|
| `filesystem_cache_evictions_total` | counter | `cache_name` |
| `filesystem_cache_evicted_bytes_total` | counter | `cache_name` |
| `filesystem_cache_evicted_segment_hits` | histogram | `cache_name` |
| `filesystem_cache_evicted_segment_size_bytes` | histogram | `cache_name` |
| (`_by_user` variant of each family) | | + `user_id`, need additional opt-in |

**Motivation.** Operators need to monitor eviction pressure beyond the existing process-wide `ProfileEvents` counters. The new metrics are per-cache. Operators can answer "is the working set fitting in the cache?" directly from PromQL.

**Mechanism.** `IFileCachePriority` gains an optional `OnEvictCallback` installed by `FileCache` at construction when the feature is enabled. The callback is invoked by `EvictionCandidates::evict` for each successfully-evicted segment. All metric family registrations live in an anonymous namespace in `FileCache.cpp`.

Escapes the labels before writing them to prometheus.
Adds the counter type in addition to gauge for prometheus metrics.

**Cardinality.** Aggregate families are bounded by `cache_name` (< 10 series per process). `_by_user` cardinality grows with the number of distinct evicting `user_id`s and is therefore off by default.

**Testing.**
- `FileCacheTest.ExposeEvictionMetrics` — three phases: flag off → no delta; aggregate on → counters and histograms advance lockstep; per-user on → `_by_user` variants populate.
- Integration test `test_filesystem_cache_eviction_metrics/` — full path through a real `clickhouse-server` with a small SLRU cache, asserting all families via `system.dimensional_metrics` and `system.histogram_metrics`.

### Version info
- Merged into: `26.7.1.289` (included in `26.7` and later)


src/Interpreters/FileCache/EvictionCandidates.cpp
```diff
@@ -323,6 +331,35 @@ void EvictionCandidates::evict()
...
+                        /// Dynamic-resize path: queue entries were already removed before evict().
+                        /// We saved the original type in removeQueueEntries.
+                        auto type_it = original_queue_types.find(candidate.get());
+                        if (type_it != original_queue_types.end())
+                            queue_type = type_it->second;
+                    }
+
+                    const size_t bytes = segment->range().size();
+                    const size_t hits = segment->getHitsCount();
+                    FileCacheMetrics::recordEviction(cache_name, queue_type, bytes, hits);
```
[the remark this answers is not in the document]
sacheendra: Fixed

src/Interpreters/FileCache/EvictionCandidates.h
```diff
@@ -116,6 +123,9 @@ class EvictionCandidates : private boost::noncopyable
     void setAfterEvictStateFunc(AfterEvictStateFunc && func) { after_evict_state_func = std::move(func); }
 
     /// Evict all candidates, which were added before via add().
+    /// If the owning cache (passed to the constructor) has eviction metrics
```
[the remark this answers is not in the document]
sacheendra: Fixed

src/Interpreters/FileCache/EvictionCandidates.cpp
```diff
@@ -26,6 +31,69 @@ namespace FailPoints
...
+namespace
+{
+    /// Eviction-time Prometheus metrics. Registered unconditionally; emission
+    /// is gated per-cache via FileCacheSettings::expose_eviction_metrics
+    /// (aggregate families) and ::expose_eviction_metrics_per_client (the
+    /// `_by_client` variants below).
+    const HistogramMetrics::Buckets hits_buckets = {0, 1, 2, 4, 8, 16, 32, 64, 128, 256, 1024, 8192};
+    const HistogramMetrics::Buckets size_buckets = {4_KiB, 16_KiB, 64_KiB, 256_KiB, 1_MiB, 4_MiB, 16_MiB, 64_MiB};
+
+    DimensionalMetrics::MetricFamily & filesystem_cache_evictions_total = DimensionalMetrics::Factory::instance().registerMetric(
```
primeroz: Can we also add a `filesystem_cache_size_bytes` and `filesesystem_cache_size_elements` by client_id ? 

i was going to add it using `custom_metrics` in this [BC](https://github.com/ClickHouse/data-plane-base-configurations/blob/main/patches/fciocchetti.py#L1372) patch but i was told that we should it natively

src/Interpreters/FileCache/EvictionCandidates.cpp
```diff
@@ -251,6 +320,11 @@ void EvictionCandidates::evict()
 
     auto timer = DB::CurrentThread::getProfileEvents().timer(ProfileEvents::FilesystemCacheEvictMicroseconds);
 
+    const bool emit_aggregate_metrics = cache && cache->areEvictionMetricsEnabled();
+    const bool emit_per_client_metrics = emit_aggregate_metrics && cache->areEvictionMetricsPerClientEnabled();
+    const String empty_name;
+    const String & cache_name = emit_aggregate_metrics ? cache->getName() : empty_name;
```
kssenii: Let's instead pass an `OnEvictCallback = std::function<void(FileSegments & segment, QueueType queue_type)>` which will capture `FileCache` and execute metrics update. I do not like mixing abstractions together for unwarranted reason, so a callback is better in my opinion.

src/Interpreters/FileCache/FileCacheSettings.cpp
```diff
@@ -62,6 +62,8 @@ namespace ErrorCodes
     DECLARE(Double, split_cache_ratio, 0.1, "Ratio of system segment to total size of cache for split_cache.", 0) \
     DECLARE(UInt64, overcommit_eviction_evict_step, 10 * 1_MiB, "Eviction step in bytes for overcommit eviction policy. Used for keep_free_space_*_ratio settings", 0) \
     DECLARE(Double, check_cache_probability, 0.001, "Works only for debug or sanitizer build. Checks cache correctness by going through all cache and checking state of each cache element", 0) \
+    DECLARE(Bool, expose_eviction_metrics, false, "Expose Prometheus metrics describing eviction activity for this cache (filesystem_cache_evictions_total, _evicted_bytes_total, _evicted_segment_hits, _evicted_segment_size_bytes). Off by default.", 0) \
```
kssenii:
```suggestion
    DECLARE(Bool, filesystem_cache_expose_prometheus_eviction_metrics, false, "Expose Prometheus metrics describing filesystem cache eviction activity for this cache (filesystem_cache_evictions_total, _evicted_bytes_total, _evicted_segment_hits, _evicted_segment_size_bytes). Off by default.", 0) \
```

src/Interpreters/FileCache/FileCacheSettings.cpp
```diff
@@ -62,6 +62,8 @@ namespace ErrorCodes
...
+    DECLARE(Bool, expose_eviction_metrics_per_client, false, "Additionally expose `_by_client` variants of the eviction metrics, labelled by user id. Requires expose_eviction_metrics. Cardinality grows with the number of distinct evicting users - off by default.", 0) \
```
kssenii:
```suggestion
    DECLARE(Bool, filesystem_cache_expose_prometheus_eviction_metrics_per_client, false, "Additionally expose `_by_client` variants of the filesystem cache eviction metrics, labelled by user id. Requires expose_eviction_metrics. Cardinality grows with the number of distinct evicting users - off by default.", 0) \
```

src/Interpreters/FileCache/FileCache.h
```diff
@@ -273,6 +282,14 @@ class FileCache : private boost::noncopyable
     const bool use_split_cache;
     const double split_cache_ratio;
 
+    /// Eviction-time Prometheus metrics. See filesystem_cache_evictions_total
+    /// and friends in `FileCacheMetrics.cpp`. Off by default; turn on by setting
```
kssenii: I see no `FileCacheMetrics.cpp`

src/Interpreters/FileCache/EvictionCandidates.cpp
sacheendra: @primeroz this change is also invasive. So, i prefer making it in a separate PR if you don't mind. I will make the PR this week.

The reason is that it requires instrumenting a different path where cache size increases without evictions.

primeroz: I don't mind, whatever you think is best works for me

src/Interpreters/FileCache/FileCache.h
sacheendra: Was there in earlier iteration. Sorry about the stale comment. Removed.

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -300,6 +399,19 @@ FileCache::FileCache(const std::string & cache_name, const FileCacheSettings & s
     }
     LOG_DEBUG(log, "Using {} cache policy", settings[FileCacheSetting::cache_policy].value);
 
+    /// Wire eviction + promotion emission via callbacks. Priority code
+    /// (e.g. SLRU's promotion-induced downgrade in tryIncreasePriority,
+    /// or its successful promotions) reuses these to emit
+    /// `filesystem_cache_*` metrics.
+    main_priority->setOnEvictCallback([this](const FileSegment & segment, FileCacheQueueEntryType queue_type, const String & user_id)
```
sacheendra: @kssenii are such callbacks what you had in mind?

src/Common/DimensionalMetrics.h
```diff
@@ -58,6 +59,24 @@ namespace DimensionalMetrics
             }
         }
 
+        template <typename Predicate>
+        size_t removeWhere(Predicate && pred)
```
sacheendra: As clients are added and removed, the metrics for removed clients pile up.

I think we restart the cache server enough (at least once a month) that this isn't a problem.
@primeroz

src/Interpreters/FileCache/EvictionCandidates.cpp
```diff
@@ -325,6 +328,31 @@ void EvictionCandidates::evict()
 
                 if (iterator)
                     queue_entries_to_invalidate.push_back(iterator);
+
+                if (emit_aggregate_metrics)
```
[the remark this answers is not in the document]
sacheendra: Oops. Fixed.

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -2056,6 +2168,50 @@ FileCache::~FileCache()
     assertCacheCorrectness();
 }
 
+void FileCache::pruneByClientMetrics(const String & user_id) const
```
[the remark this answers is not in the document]
sacheendra: Users of CacheUsage need to set the `on_drained` callback and link it to `pruneByClientMetrics`. There are no users in public code.

src/Common/DimensionalMetrics.h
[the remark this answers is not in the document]
sacheendra: Used a shared ptr based approach to prevent the use after free.
We can have some lost metrics, but that's fine.

src/Common/HistogramMetrics.h
```diff
@@ -62,6 +62,24 @@ namespace HistogramMetrics
             }
         }
 
+        template <typename Predicate>
+        size_t removeWhere(Predicate && pred)
```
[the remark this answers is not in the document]
sacheendra: See prev comment

src/Interpreters/FileCache/FileCache.cpp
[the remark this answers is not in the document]
sacheendra: Check again. Added a para to the PR description at the end

src/Interpreters/FileCache/IFileCachePriority.h
```diff
@@ -15,6 +16,8 @@ namespace DB
 {
 struct FileCacheReserveStat;
 class EvictionCandidates;
+class FileCache;
```
kssenii: This seems unused

src/Interpreters/FileCache/IFileCachePriority.h
```diff
@@ -393,6 +396,19 @@ class IFileCachePriority : private boost::noncopyable
 
     virtual void setCacheUsageStatGuard(std::shared_ptr<CacheUsageStatGuard>) {}
 
+    /// Invoked by `EvictionCandidates::evict` for each successfully-evicted
+    /// segment. `user_id` is passed explicitly because `segment` has already
+    /// been detached from its key metadata by the time the callback fires
+    /// (see `FileSegment::setDetachedState`).
```
kssenii:
```suggestion
    /// Invoked by `EvictionCandidates::evict` for each successfully-evicted
    /// segment.
```

src/Interpreters/FileCache/IFileCachePriority.h
```diff
@@ -393,6 +396,19 @@ class IFileCachePriority : private boost::noncopyable
...
+    using OnEvictCallback = std::function<void(const FileSegment & segment, QueueEntryType queue_type, const UserID & user_id)>;
+    virtual void setOnEvictCallback(OnEvictCallback callback) { on_evict_callback = std::move(callback); }
```
kssenii:
```suggestion
    virtual void setOnEvictCallback(OnEvictCallback callback) { chassert(on_evict_callback == {}, "on_evict_callback cannot be set twice"); on_evict_callback = std::move(callback); }
```

src/Interpreters/FileCache/SLRUFileCachePriority.cpp
```diff
@@ -657,7 +658,10 @@ bool SLRUFileCachePriority::tryIncreasePriority(
 #endif
     }
 
-    EvictionCandidates downgrade_candidates;
+    /// Pass the owning FileCache so `filesystem_cache_*` metrics count
+    /// promotion-induced downgrade evictions, not only the FileCache.cpp
+    /// reservation/resize call sites.
+    EvictionCandidates downgrade_candidates(on_evict_callback);
```
kssenii: This is "downgrade" e.g. not actual eviction, so I do not think it is correct to apply `on_evict_callback` here (though we do currently update other eviction metrics there, this was in TODO to fix...). Ideally no eviction metrics need to be updated there.

src/Interpreters/FileCache/FileCacheSettings.cpp
```diff
@@ -62,6 +62,8 @@ namespace ErrorCodes
     DECLARE(Double, split_cache_ratio, 0.1, "Ratio of system segment to total size of cache for split_cache.", 0) \
     DECLARE(UInt64, overcommit_eviction_evict_step, 10 * 1_MiB, "Eviction step in bytes for overcommit eviction policy. Used for keep_free_space_*_ratio settings", 0) \
     DECLARE(Double, check_cache_probability, 0.001, "Works only for debug or sanitizer build. Checks cache correctness by going through all cache and checking state of each cache element", 0) \
+    DECLARE(Bool, filesystem_cache_expose_prometheus_eviction_metrics, false, "Expose Prometheus metrics describing filesystem cache eviction activity for this cache (filesystem_cache_evictions_total, _evicted_bytes_total, _evicted_segment_hits, _evicted_segment_size_bytes). Off by default.", 0) \
+    DECLARE(Bool, filesystem_cache_expose_prometheus_eviction_metrics_per_client, false, "Additionally expose `_by_client` variants of the filesystem cache eviction metrics, labelled by user id. Requires filesystem_cache_expose_prometheus_eviction_metrics. Cardinality grows with the number of distinct evicting users - off by default.", 0) \
```
kssenii: Hm I asked to add `filesystem_cache_` prefix in an earlier comment, because I did not notice they were added on cache settings level, I assumed it was query level. To me it seems more logical to have them changeable on query level (to easily enable for debugging if needed) and background eviction processes can just take those settings from default global values. Though it will take more effort to pass them through to cache, I still think this would be the right way. WDYT?

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -2056,6 +2143,31 @@ FileCache::~FileCache()
...
+
+    if (!expose_eviction_metrics_per_client)
+        return;
+    filesystem_cache_evictions_by_client_total.withLabels({name, queue, user_id}).increment();
+    filesystem_cache_evicted_bytes_by_client_total.withLabels({name, queue, user_id}).increment(static_cast<DimensionalMetrics::Value>(bytes));
+    filesystem_cache_evicted_segment_hits_by_client.withLabels({name, queue, user_id}).observe(static_cast<HistogramMetrics::Value>(hits));
+    filesystem_cache_evicted_segment_size_bytes_by_client.withLabels({name, queue, user_id}).observe(static_cast<HistogramMetrics::Value>(bytes));
+}
+
+void FileCache::onSegmentPromoted(const String & user_id) const
```
kssenii: If we have "promotion metrics", then we also need "downgrade metrics"?

src/Interpreters/FileCache/SLRUFileCachePriority.cpp
sacheendra: It's recorded as an eviction with the the queue label set to `protected`.
That's not a good way indeed.

I'll record separate downgrade events.

src/Interpreters/FileCache/FileCache.cpp
sacheendra: Yes, will record separate downgrade events.

src/Interpreters/FileCache/FileCacheSettings.cpp
sacheendra: Being able to enable the metrics at runtime for debugging makes sense.

I'll implement and see how it looks.

src/Interpreters/FileCache/FileCache.cpp
sacheendra: After thinking a bit, removed both promotion and downgrade.

src/Core/Settings.cpp
```diff
@@ -6427,6 +6427,19 @@ Cloud default value: `1`.
 )", 0) \
     DECLARE(Bool, enable_filesystem_cache_log, false, R"(
 Allows to record the filesystem caching log for each query
+)", 0) \
+    DECLARE(Bool, filesystem_cache_expose_prometheus_eviction_metrics, false, R"(
```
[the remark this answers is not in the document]
sacheendra: irrelevant

src/Interpreters/FileCache/FileCacheSettings.cpp
sacheendra: Hi @kssenii, I tried to implement and this seems tough.

The query-level settings, even from the default profile, are not available to background threads.

This is the code I tried in the eviction function.
```
auto query_context = CurrentThread::tryGetQueryContext();
const auto & settings = query_context 
    ? query_context->getSettingsRef() 
    : Context::getGlobalContextInstance()->getSettingsRef();
if (!settings[Setting::filesystem_cache_expose_prometheus_eviction_metrics])
    return;

```

Global settings in `src/Core/ServerSettings.cpp` are available in the background threads, but these cannot be set at the query level.

I think the cache-level settings are ok. It seems I can hot-reload them using `SYSTEM RELOAD CONFIG` to enable or disable the metrics at runtime for debugging.

I'm open to try other things if you have any suggestions.

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -115,6 +120,49 @@ namespace
...
+
+    const HistogramMetrics::Buckets hits_buckets = {0, 1, 2, 4, 8, 16, 32, 64, 128, 256, 1024, 8192};
+    const HistogramMetrics::Buckets size_buckets = {4_KiB, 16_KiB, 64_KiB, 256_KiB, 1_MiB, 4_MiB, 16_MiB, 64_MiB};
+
+    DimensionalMetrics::MetricFamily & filesystem_cache_evictions_total = DimensionalMetrics::Factory::instance().registerMetric(
+        "filesystem_cache_evictions_total",
+        "Number of file segments evicted from a filesystem cache, labelled by cache name.",
+        {"cache_name"});
+
+    DimensionalMetrics::MetricFamily & filesystem_cache_evicted_bytes_total = DimensionalMetrics::Factory::instance().registerMetric(
```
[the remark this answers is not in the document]
sacheendra: fixed

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -115,6 +120,49 @@ namespace
         const auto user = user_from_context.empty() ? toString(ServerUUID::get()) : user_from_context;
         return user;
     }
+
+    const HistogramMetrics::Buckets hits_buckets = {0, 1, 2, 4, 8, 16, 32, 64, 128, 256, 1024, 8192};
+    const HistogramMetrics::Buckets size_buckets = {4_KiB, 16_KiB, 64_KiB, 256_KiB, 1_MiB, 4_MiB, 16_MiB, 64_MiB};
+
+    DimensionalMetrics::MetricFamily & filesystem_cache_evictions_total = DimensionalMetrics::Factory::instance().registerMetric(
```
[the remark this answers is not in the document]
sacheendra: That seems to be the convention in clickhouse code.
A refactor to identify gauges vs counters is out of scope for this PR.

kssenii:
FYI, I am going to resolve the conflicts, since I made them.

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -2073,10 +2128,27 @@ FileCache::~FileCache()
...
+    const size_t bytes = segment.getReservedSize();
+    const size_t hits = segment.getHitsCount();
+    filesystem_cache_evictions_total.withLabels({name}).increment();
+    filesystem_cache_evicted_bytes_total.withLabels({name}).increment(static_cast<DimensionalMetrics::Value>(bytes));
+    filesystem_cache_evicted_segment_hits.withLabels({name}).observe(static_cast<HistogramMetrics::Value>(hits));
+    filesystem_cache_evicted_segment_size_bytes.withLabels({name}).observe(static_cast<HistogramMetrics::Value>(bytes));
+
+    if (!expose_eviction_metrics_per_client.load(std::memory_order_relaxed))
+        return;
+    filesystem_cache_evictions_by_client_total.withLabels({name, user_id}).increment();
```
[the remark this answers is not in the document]
kssenii: This is true only for public code part.

sacheendra:
Made a tiny change to rename `_by_client` to `_by_user` as we changed the label to `user_id`.

kssenii (approved):

src/Common/DimensionalMetrics.h
```diff
@@ -21,6 +21,12 @@ namespace DimensionalMetrics
     using Labels = std::vector<String>;
     using LabelValues = std::vector<String>;
 
+    enum class MetricType
```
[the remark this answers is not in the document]
sacheendra: modifying dimensional metric table is out of scope for this pr

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -2149,10 +2212,27 @@ FileCache::~FileCache()
...
+    const size_t bytes = segment.getReservedSize();
+    const size_t hits = segment.getHitsCount();
+    filesystem_cache_evictions_total.withLabels({name}).increment();
+    filesystem_cache_evicted_bytes_total.withLabels({name}).increment(static_cast<DimensionalMetrics::Value>(bytes));
+    filesystem_cache_evicted_segment_hits.withLabels({name}).observe(static_cast<HistogramMetrics::Value>(hits));
+    filesystem_cache_evicted_segment_size_bytes.withLabels({name}).observe(static_cast<HistogramMetrics::Value>(bytes));
+
+    if (!expose_eviction_metrics_per_user.load(std::memory_order_relaxed))
+        return;
+    filesystem_cache_evictions_by_user_total.withLabels({name, user_id}).increment();
```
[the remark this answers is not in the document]
sacheendra: done. escaped the labels

tests/integration/test_filesystem_cache_eviction_metrics/test.py
```diff
@@ -0,0 +1,106 @@
...
+    node.query(
+        """
+        CREATE TABLE eviction_metrics_test (id UInt64, blob String CODEC(NONE))
+        ENGINE = MergeTree ORDER BY id
+        SETTINGS storage_policy = 'cache_metrics_policy', min_bytes_for_wide_part = 0
+        """
+    )
+
+    # Snapshot all eight families before writes.
+    evictions_before        = sum_dim("filesystem_cache_evictions_total")
```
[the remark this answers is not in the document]
sacheendra: added assertions to tests

kssenii (approved):

Issue #105021, opened 2026-05-15
EXCHANGE TABLES with a materialized-view -- regression from #98779 


CheSema:
There are two problems here. The first is a regression and is fixed by #105029. The second is a pre-existing design ambiguity that is *not* fixed here, only pinned by a test.

## 1. Regression on master (this issue)

Introduced by #98779 (merged 2026-05-08). Correct on `v26.4.2.10`, broken on master `26.5.1.1`. Demonstrated by two tests added in #105029.

### `04240_exchange_tables_drops_mv_source_dependencies` — `EXCHANGE TABLES`

```sql
CREATE TABLE a (id Int32, key String) ENGINE = Null;
CREATE TABLE b (id Int32, key String) ENGINE = Null;
CREATE TABLE dst (id Int32, key String) ENGINE = MergeTree ORDER BY id;
CREATE MATERIALIZED VIEW mv TO dst AS SELECT id, key FROM a;

INSERT INTO a VALUES (1, 'x');
SELECT count() FROM dst;
-- 1

EXCHANGE TABLES a AND b;

INSERT INTO a VALUES (2, 'y');
SELECT count() FROM dst;
-- expected: 2  (`MV` was created `FROM a`, name `a` still exists)
-- master:   1  (silent drop)

INSERT INTO b VALUES (3, 'z');
SELECT count() FROM dst;
-- expected: 2  (`MV` should not fire — `b` is a different table)
-- master:   1  (also a silent drop, but for the same reason)
```

### `04235_create_or_replace_source_table_view_dependencies` — `CREATE OR REPLACE TABLE`

```sql
CREATE TABLE src (id Int32, key String) ENGINE = Null;
CREATE TABLE dst (id Int32, key String) ENGINE = MergeTree ORDER BY id;
CREATE MATERIALIZED VIEW mv TO dst AS SELECT id, key FROM src;

INSERT INTO src VALUES (1, 'x');
SELECT count() FROM dst;
-- 1

CREATE OR REPLACE TABLE src (id Int32, key String) ENGINE = Null;

INSERT INTO src VALUES (2, 'y');
SELECT count() FROM dst;
-- expected: 2  (`MV` keeps reading from name `src`)
-- master:   1  (silent drop)
```

This case was independently reported and fixed in #104907. The fix in #105029 subsumes it.

### Why it breaks

[...]

PR #105023, opened 2026-05-15
Reproducer for #105021: EXCHANGE TABLES drops MV-source dependencies

CheSema:

Add functional reproducer for #105021. After `EXCHANGE TABLES a, b` where a materialized view reads from `a`, inserts into both `a` and `b` silently stop feeding the `MV` target table. The reference pins the current buggy behavior; once #105021 is fixed, the reference must be updated.

Regression details and proposed fix options in #105021. Confirmed against `v26.4.2.10` (correct) and current master `26.5.1.1` (broken). Bisected to #98779.


george-larionov (approved):

CheSema:
close in favor of the PR with the fix https://github.com/ClickHouse/ClickHouse/pull/105029

Issue #76303, opened 2025-02-17
Views and materialized view definitions don't change when underlying tables are renamed


xoelop:
### Company or project name

[TheirStack](https://theirstack.com/en)

### Describe the unexpected behaviour

Hi, coming from other databases I was used that when defining views or materialized views over tables, when those tables are renamed, the views that depend on them keep working. This doesn't seem to be the case with ClickHouse, which causes those views to break and you only realize that when you make queries involving them.

This makes schema migrations that involve creating a different version of the same table quite hard to manage, as we'd need to recreate all the views and materialized views to reflect the new names. An [exchange tables](https://clickhouse.com/docs/en/engines/database-engines/atomic#exchange-tables) command exists, but it also doesn't affect the definition of the views that depend on those tables, so it's not quite useful

### How to reproduce

Just run these commands from this fiddle: https://fiddle.clickhouse.com/feec1326-526d-44e0-84e0-ba08f67f29e0

```sql
create table test(name String) Engine MergeTree order by tuple();

INSERT INTO test VALUES ('asdf');

SELECT * FROM test;


CREATE VIEW test_view AS SELECT * FROM test;

RENAME TABLE test TO test_2;

SHOW CREATE VIEW test_view;

SELECT * FROM test_view;
```

This is the result:

```
asdf
CREATE VIEW default.test_view\n(\n    `name` String\n)\nAS SELECT *\nFROM default.test

Received exception from server (version 25.1.4):
Code: 60. DB::Exception: Received from localhost:9000. DB::Exception: Unknown table expression identifier 'default.test' in scope SELECT * FROM default.test. (UNKNOWN_TABLE)
(query: SELECT * FROM test_view


)
```

Copying it from the console:

```bash
9b28dcbdde25 :) create table test(name String) Engine MergeTree order by tuple();

INSERT INTO test VALUES ('asdf');

SELECT * FROM test;


CREATE VIEW test_view AS SELECT * FROM test;

RENAME TABLE test TO test_2;

SHOW CREATE VIEW test_view;

SELECT * FROM test_view;


CREATE TABLE test
(
    `name` String
)
ENGINE = MergeTree
ORDER BY tuple()

Query id: 563ed438-8dc1-4981-b4e6-21cb8d372669

Ok.

0 rows in set. Elapsed: 0.039 sec. 


INSERT INTO test FORMAT Values

Query id: f3a04306-9bd4-4370-96b1-b4018c699b2b

Ok.

[...]

PR #105029, opened 2026-05-15
Fix EXCHANGE TABLES dropping MV source dependencies (#105021)

### LINKED ISSUE #105021: EXCHANGE TABLES with a materialized-view -- regression from #98779

CheSema:

Fix silent data loss after `EXCHANGE TABLES` or `CREATE OR REPLACE TABLE` of a materialized view's source table. The `MV`'s source-view dependency edge is now kept on the original name so it continues to fire on inserts. Regression introduced by #98779; restores the pre-regression behavior.

### Description

Closes #105021.

`EXCHANGE TABLES a, b` (and the synthetic exchange used by `CREATE OR REPLACE TABLE` / `REPLACE TABLE`) cross-swapped the source-side `view_dependencies` edges along with the data. The `MV`'s stored `select_table_id` is not rewritten by the rename, so cross-swapping orphaned the `MV`:

- `INSERT INTO a` after exchange found no dependents — the edge had been moved to `b`.
- `INSERT INTO b` was rejected by `InsertDependenciesBuilder::observePath` because the `MV`'s recorded source name `a` no longer matched the parent `b`.

Both inserts silently dropped rows. Confirmed regression: `v26.4.2.10` correct, master broken (since #98779).

The fix keeps source-view edges keyed by the original name on `EXCHANGE`. Plain `RENAME TABLE` keeps the post-#98779 re-keying behavior used by `01155_rename_move_materialized_view`. This also subsumes the special case from #104907 — the synthetic `CREATE OR REPLACE` exchange routes through the same `exchange_tables=true` branch.

### Tests

- `04240_exchange_tables_drops_mv_source_dependencies` — regression test for #105021.
- `04235_create_or_replace_source_table_view_dependencies` — the test introduced in #104907; passes with the unified fix.
- `04241_rename_table_mv_source_dependencies` — pins a pre-existing latent bug in plain same-database `RENAME TABLE` (silent data loss, alive-but-orphaned `MV`). Not a #98779 regression. Out of scope for this PR; root cause shared with #76303 and tracked there.

### Design note

This restores ClickHouse's pre-#98779 behavior, where the source-view edge follows the **name** on `EXCHANGE` ("option C": MV reads from whatever now lives at the name it was created against). This is a **regression repair**, not the long-term answer. A proper redesign — identity-bound `MV` with atomic `as_select` rewrite, like PostgreSQL handles the same problem — would fix this case, the plain-`RENAME` case (`04241`), and #76303 in one shot. That redesign belongs in a separate, larger change.

- [x] Documentation is unnecessary.

### Version info
- Merged into: `26.5.1.747`


george-larionov (approved):
It looks good to me.

serprex (approved):

PR #105030, opened 2026-05-15
Optimize per-event trace flags: move into separate array and lazy allocation

azat:
Those flags are rarely used, since it is only for debugging under `trace_profile_events`, but it increase the size of the structure 2x (Counter increased from 8 bytes to 16 due to alignment)

So move it into separate array, and allocate it lazily.

- Performance Improvement

Optimize per-event trace flags: move into separate array and lazy allocation

### Version info
- Merged into: `26.5.1.820`

### DIFF
--- a/src/Common/ProfileEvents.cpp
+++ b/src/Common/ProfileEvents.cpp
@@ -1472,6 +1472,8 @@ Counters::Counters(Counters && src) noexcept
     : counters(std::exchange(src.counters, nullptr))
     , counters_holder(std::move(src.counters_holder))
     , parent(src.parent.exchange(nullptr))
+    , should_trace_array(src.should_trace_array.exchange(nullptr, std::memory_order_relaxed))
+    , should_trace_holder(std::move(src.should_trace_holder))
     , trace_all_profile_events(src.trace_all_profile_events.load(std::memory_order_relaxed))
     , level(src.level)
 {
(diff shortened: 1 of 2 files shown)


src/Common/ProfileEvents.cpp
```diff
@@ -1646,7 +1668,8 @@ void Counters::increment(Event event, Count amount)
     do
     {
         current->counters[event].fetch_add(amount, std::memory_order_relaxed);
-        send_to_trace_log |= current->counters[event].should_trace;
+        if (auto * trace_arr = current->should_trace_array.load(std::memory_order_relaxed))
```
Algunenano, line 1671: I'm curious whether the extra if vs reduced memory works out performance wise

Algunenano (approved):
I tried to do something similar and iterated a few times trying to improve this in https://github.com/ClickHouse/ClickHouse/pull/104513 and ended up giving up 😄

src/Common/ProfileEvents.cpp
azat: I would expect that one branch here is way less then atomic's

azat: And I have further improvements for profile events - https://github.com/ClickHouse/ClickHouse/pull/105056

azat:
Rebased to run perf tests again (not enough data)

UPD: now it is better

PR #105036, opened 2026-05-15
Explain pretty make default

Fgrtue:

- Improvement

Make `EXPLAIN [PLAN] actions=1, compact=1, pretty=1` the default.

### Version info
- Merged into: `26.7.1.194` (included in `26.7` and later)


docs/en/sql-reference/statements/explain.md
```diff
@@ -186,15 +186,21 @@ Settings:
...
 - `column_structure` — Prints also the structure of columns in headers on top of their name and type. Default: 0. Mostly useful only for developers to debug issues related to input-output header mismatch.
 - `distributed` — Shows query plans executed on remote nodes for distributed tables or parallel replicas. Default: 0.
-- `compact` — When enabled, hides expression steps and detailed action info (inputs, functions, aliases, and output positions) from the plan. Only has an effect when actions = 1. Default: 0.
-- `pretty` — Prints the plan tree using line-drawing characters (├──, └──, │) instead of indentation to visualize the hierarchy. Also formats join step properties inline. Default: 0.
+- `compact` — When enabled, hides expression steps and detailed action info (inputs, functions, aliases, and output positions) from the plan. Only has an effect when `actions = 1`. Default: 1 (see `query_plan_pretty_default` below).
+- `pretty` — Prints the plan tree using line-drawing characters (├──, └──, │) instead of indentation to visualize the hierarchy. Also formats join step properties inline. Default: 1 (see `query_plan_pretty_default` below).
```
rienath: nit: I think doc would be clearer without  mentioning`query_plan_pretty_default` as it is already described below and user can change the default value anyway

docs/en/sql-reference/statements/explain.md
```diff
@@ -186,15 +186,21 @@ Settings:
...
+
+:::note
+Starting from ClickHouse 26.5, the server setting `query_plan_pretty_default` is `true` by default. When it is `true`, `EXPLAIN PLAN` initializes `actions`, `compact`, and `pretty` to `1` so the plan is rendered in the compact, pretty, action-annotated form. Per-query `SETTINGS actions = ..., compact = ..., pretty = ...` always override this.
+:::
+
+To restore the pre-26.5 verbose output, either run `SET query_plan_pretty_default = 0` (or pass it in per-query `SETTINGS`), or set the `compatibility` setting to any version older than `26.5`.
```
rienath: Should this be within the note above?

tests/queries/0_stateless/04219_explain_plan_pretty_default.sql
```diff
@@ -0,0 +1,30 @@
...
+-- setting can roll the defaults back. `pretty=1` is detected via line-drawing
+-- characters in the output, which neither old default nor `pretty=0` produces.
+-- The `compatibility` check runs first because `applyCompatibilitySetting` only
+-- touches settings the user has not explicitly set.
+
+-- 1. compatibility = '25.5' must roll the default back (verifies SettingsChangesHistory wiring).
+SET compatibility = '25.5';
+SELECT countIf(explain LIKE '%├──%' OR explain LIKE '%└──%') = 0 AS no_pretty_under_compat
+FROM (EXPLAIN PLAN SELECT number FROM numbers(10) WHERE number > 3);
+SET compatibility = '';
```
rienath: It's not clear to me why this works. `compatibility` works based on `SettingsChangesHistory.cpp`, which wasn't updated. Perhaps because in 25.5 there was no pretty at all? Let's change this to 26.4

tests/queries/0_stateless/03100_lwu_36_json_skip_indexes.sql
```diff
@@ -0,0 +1,74 @@
+-- Tags: no-parallel-replicas
```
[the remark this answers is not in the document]
rienath: This file was deleted from master. Probably committed by mistake

src/Interpreters/InterpreterExplainQuery.cpp
```diff
@@ -379,12 +380,23 @@ struct QuerySyntaxSettings
 };
 
 template <typename Settings>
-ExplainSettings<Settings> checkAndGetSettings(const ASTPtr & ast_settings)
+ExplainSettings<Settings> checkAndGetSettings(const ASTPtr & ast_settings, bool set_default_legacy_explain_settings = false)
```
rienath: nit: the parameter name `set_default_legacy_explain_settings` reads awkwardly (it doesn't "set defaults", it overrides them to the legacy values). `restore_legacy_explain_defaults` is clearer

src/Interpreters/InterpreterExplainQuery.cpp
```diff
@@ -379,12 +380,23 @@ struct QuerySyntaxSettings
...
 {
+    ExplainSettings<Settings> settings;
+
+    if constexpr (std::is_same_v<Settings, QueryPlanSettings>)
+    {
+        if (set_default_legacy_explain_settings)
+        {
+            settings.query_plan_options.actions = false;
+            settings.query_plan_options.compact = false;
+            settings.query_plan_options.pretty  = false;
```
rienath:
```suggestion
            settings.query_plan_options.pretty = false;
```

rienath (changes requested):
Overall looks good. I've left some comments. Let's also mention the new `query_plan_pretty_default` setting in the changelog

src/Interpreters/InterpreterExplainQuery.cpp
Fgrtue: I really like the aligned version better, thank you for suggestion though

tests/queries/0_stateless/03392_crash_group_by_use_nulls.sql
```diff
@@ -1,3 +1,4 @@
+SET explain_query_plan_default = 'legacy';
 -- https://github.com/ClickHouse/ClickHouse/issues/77485
```
rienath: nit: let's have the issue on the first line as it explains the regression we are testing against before the test starts

docs/en/sql-reference/statements/explain.md
```diff
@@ -35,7 +35,7 @@ EXPLAIN [AST | SYNTAX | QUERY TREE | PLAN | PIPELINE | ESTIMATE | TABLE OVERRIDE
 Example:
 
 ```sql
-EXPLAIN SELECT sum(number) FROM numbers(10) UNION ALL SELECT sum(number) FROM numbers(10) ORDER BY sum(number) ASC FORMAT TSV;
+EXPLAIN SELECT sum(number) FROM numbers(10) UNION ALL SELECT sum(number) FROM numbers(10) ORDER BY sum(number) ASC FORMAT TSV SETTINGS explain_query_plan_default = 'legacy';
```
rienath: Wdyt about dropping `SETTINGS explain_query_plan_default = 'legacy'` here and updating the output to the new default? Would be nice if the first example shows what users get out of the box.

docs/en/sql-reference/statements/explain.md
```diff
@@ -182,22 +182,31 @@ Settings:
...
 - `distributed` — Shows query plans executed on remote nodes for distributed tables or parallel replicas. Default: 0.
-- `compact` — When enabled, hides expression steps and detailed action info (inputs, functions, aliases, and output positions) from the plan. Only has an effect when actions = 1. Default: 0.
-- `pretty` — Prints the plan tree using line-drawing characters (├──, └──, │) instead of indentation to visualize the hierarchy. Also formats join step properties inline. Default: 0.
+- `compact` — When enabled, hides expression steps and detailed action info (inputs, functions, aliases, and output positions) from the plan. Only has an effect when `actions = 1`. Default: 1 when `explain_query_plan_default` is `pretty`, 0 when it is `legacy` (see the note below).
+- `pretty` — Prints the plan tree using line-drawing characters (├──, └──, │) instead of indentation to visualize the hierarchy. Also formats join step properties inline. Default: 1 when `explain_query_plan_default` is `pretty`, 0 when it is `legacy` (see the note below).
```
rienath: nit: maybe shorten these to `Default: 1 (see note below)`? The note already explains the legacy behaviour nicely, so the full condition feels like a lot to process here.

docs/en/sql-reference/statements/explain.md
```diff
@@ -182,22 +182,31 @@ Settings:
...
 When `json=1` step names will contain an additional suffix with unique step identifier.
 
 Example:
 
+The examples below use `SETTINGS explain_query_plan_default = 'legacy'` to show the verbose, indented output. This is the baseline that the `compact`, `distributed`, and `pretty` sections further down build upon. With the default `pretty` value, `EXPLAIN PLAN` renders the compact, pretty form shown in the `pretty = 1` examples further down.
+
 ```sql
-EXPLAIN SELECT sum(number) FROM numbers(10) GROUP BY number % 4;
+EXPLAIN SELECT sum(number) FROM numbers(10) GROUP BY number % 4
+SETTINGS explain_query_plan_default = 'legacy';
```
rienath: What do you think about showing the default (pretty) output in these baseline examples instead of pinning legacy? The verbose output could then live in the `actions`/`compact`/`pretty` sections, where it's actually being explained. The `json=1` examples might be better with `actions = 0`, since dropping legacy there would make the outputs huge

docs/en/sql-reference/statements/explain.md
```diff
@@ -182,22 +182,31 @@ Settings:
...
-- `pretty` — Prints the plan tree using line-drawing characters (├──, └──, │) instead of indentation to visualize the hierarchy. Also formats join step properties inline. Default: 0.
+- `compact` — When enabled, hides expression steps and detailed action info (inputs, functions, aliases, and output positions) from the plan. Only has an effect when `actions = 1`. Default: 1 when `explain_query_plan_default` is `pretty`, 0 when it is `legacy` (see the note below).
+- `pretty` — Prints the plan tree using line-drawing characters (├──, └──, │) instead of indentation to visualize the hierarchy. Also formats join step properties inline. Default: 1 when `explain_query_plan_default` is `pretty`, 0 when it is `legacy` (see the note below).
+
+:::note
+Starting from ClickHouse 26.6, the server setting `explain_query_plan_default` is `pretty` by default. When it is `pretty`, `EXPLAIN PLAN` initializes `actions`, `compact`, and `pretty` to `1` so the plan is rendered in the compact, pretty, action-annotated form. Specifying these options explicitly in the `EXPLAIN` statement (for example, `EXPLAIN actions = 0, compact = 0, pretty = 0 SELECT ...`) always overrides this.
```
rienath:
```suggestion
Starting from ClickHouse 26.6, the setting `explain_query_plan_default` is `pretty` by default. When it is `pretty`, `EXPLAIN PLAN` initializes `actions`, `compact`, and `pretty` to `1` so the plan is rendered in the compact, pretty, action-annotated form. Specifying these options explicitly in the `EXPLAIN` statement (for example, `EXPLAIN actions = 0, compact = 0, pretty = 0 SELECT ...`) always overrides this.
```

It's a [session](https://clickhouse.com/docs/operations/settings/settings) setting as opposed to [server](https://clickhouse.com/docs/operations/server-configuration-parameters/settings) settings.

src/Core/Settings.cpp
```diff
@@ -15,6 +15,7 @@
 #include <Core/SettingsTierType.h>
 #include <IO/ReadBufferFromString.h>
 #include <IO/S3Defines.h>
+#include <Processors/QueryPlan/QueryPlan.h>
```
rienath: Do we need this? `ExplainQueryPlanDefault` lives in `SettingsEnums`

src/Core/Settings.cpp
```diff
@@ -2449,6 +2450,17 @@ Apply sharding for JOIN if join keys contain a prefix of PRIMARY KEY for both ta
     \
     DECLARE(Bool, query_plan_display_internal_aliases, false, R"(
 Show internal aliases (such as __table1) in EXPLAIN PLAN instead of those specified in the original query.
+)", 0) \
+    \
+    DECLARE(ExplainQueryPlanDefault, explain_query_plan_default, ExplainQueryPlanDefault::PRETTY, R"(
```
rienath, line 2516: Would a `Bool` be simpler here? E.g. `explain_query_plan_legacy_format`, default false. The enum needs registration in `SettingsEnums.h/.cpp` and `Settings.h`, and the pretty value reads as if it's about the `pretty` option specifically rather than the whole `actions`/`compact`/`pretty` bundle. Unless you expect a third format later, then the enum makes sense

src/Processors/QueryPlan/QueryPlan.h
```diff
@@ -56,7 +56,7 @@ struct ExplainPlanOptions
     /// Add description of step.
     bool description = true;
     /// Add detailed information about step actions.
-    bool actions = false;
```
rienath, line 59: This affects internal consumers. For example, in `executeQuery.cpp`

```cpp
            plan.explainPlan(buf, {.header=true, .actions=true});
            LOG_TRACE(getLogger("executeQuery"), "Deserialized Query Plan:\n{}", buf.str());
```

we'd now get `compact=1, pretty=1`. In `StorageDistributed.cpp` too

```cpp
                plan.explainPlan(buf, {});
                throw Exception(ErrorCodes::LOGICAL_ERROR,
                    "Found multiple source steps for query\n{}\nPlan\n{}",
                    ast->formatForErrorMessage(), buf.str());
```
Here plan now hides the steps that we needed to see. There are a few other places. Perhaps this is what we want, but better to double-check and change each consumer accordingly.

An alternative is to invert `checkAndGetSettings` to set these 3 to true

src/Core/SettingsChangesHistory.cpp
```diff
@@ -48,6 +48,7 @@ const VersionToSettingsChangesMap & getSettingsChangesHistory()
             {"text_index_density_threshold", 0.2, 0.2, "New setting for lazy posting list density threshold"},
             {"show_remote_databases_in_system_tables", false, false, "Renamed from `show_data_lake_catalogs_in_system_tables` and broadened to also hide `MySQL` and `PostgreSQL` databases from `system.tables`, `system.columns` and `system.completions` by default, since enumerating their tables requires expensive remote calls. Users who relied on the previous behavior must set this setting to `true`. The old name is kept as an alias."},
             {"enable_streaming_queries", false, false, "New setting"},
+            {"explain_query_plan_default", "legacy", "pretty", "From 26.6, `EXPLAIN PLAN` defaults to `actions=1, compact=1, pretty=1`. Set this to `legacy` to set pre-26.6 version by default."},
```
rienath: nit (I think it reads a little awkward)
```suggestion
            {"explain_query_plan_default", "legacy", "pretty", "From 26.6, `EXPLAIN PLAN` defaults to `actions=1, compact=1, pretty=1`. Set this to `legacy` to restore the pre-26.6 output."},
```

docs/en/sql-reference/statements/explain.md
```diff
@@ -182,22 +182,31 @@ Settings:
...
+
+To restore the pre-26.6 verbose output, either run `SET explain_query_plan_default = 'legacy'` (or pass it in per-query `SETTINGS`), or set the `compatibility` setting to any version older than `26.6`.
```
rienath: Should `json=1` get the new defaults too? It inherits `actions=1`, but `compact`/`pretty` are ignored by the JSON renderer. So machine-readable output just gets a lot bigger by default. Maybe keep legacy defaults when `json=1`? If intentional, worth a line in the docs note.

tests/queries/0_stateless/04219_explain_plan_pretty_default.sql
```diff
@@ -0,0 +1,28 @@
...
+SET optimize_read_in_order = 1;
+SET optimize_sorting_by_input_stream_properties = 1;
+SET optimize_syntax_fuse_functions = 0;
+
+DROP TABLE IF EXISTS t_explain_default;
+CREATE TABLE t_explain_default (a UInt64, b String, c Float64) ENGINE = MergeTree ORDER BY a
+SETTINGS index_granularity = 8192, index_granularity_bytes = '10Mi';
+INSERT INTO t_explain_default SELECT number, toString(number % 10), number * 1.5 FROM numbers(100);
+
+EXPLAIN
```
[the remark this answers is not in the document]
rienath: Compatibility contract is exercised in `02995_new_settings_history.sh`

tests/queries/0_stateless/04219_explain_plan_pretty_default.sql
```diff
@@ -0,0 +1,28 @@
...
+SET use_statistics = 0;
+SET optimize_move_to_prewhere = 0;
+SET allow_reorder_prewhere_conditions = 0;
+SET optimize_aggregation_in_order = 0;
+SET optimize_read_in_order = 1;
+SET optimize_sorting_by_input_stream_properties = 1;
+SET optimize_syntax_fuse_functions = 0;
+
+DROP TABLE IF EXISTS t_explain_default;
+CREATE TABLE t_explain_default (a UInt64, b String, c Float64) ENGINE = MergeTree ORDER BY a
```
rienath, line 15: Let's also add a test with `EXPLAIN actions = 0, compact = 0, pretty = 0` under the pretty default to test that we get old output as expected (and s.t. we do not regress under later changes)

tests/queries/0_stateless/03394_distributed_sort.sh
```diff
@@ -5,6 +5,8 @@ CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 # shellcheck source=../shell_config.sh
 . "$CURDIR"/../shell_config.sh
 
+$CLICKHOUSE_CLIENT -q "SET explain_query_plan_default = 'legacy'"
```
rienath: `SET` only changes the setting in a session. The invocations below don't get it

tests/queries/0_stateless/01763_filter_push_down_bugs.sql
```diff
@@ -1,3 +1,4 @@
+SET explain_query_plan_default = 'legacy';
 -- add_minmax_index_for_numeric_columns=0: Different plan
 SET use_statistics = 0;
```
rienath:
```suggestion
-- add_minmax_index_for_numeric_columns=0: Different plan
SET explain_query_plan_default = 'legacy';
SET use_statistics = 0;
```

tests/queries/0_stateless/01872_functions_to_subcolumns_analyzer.sql
```diff
@@ -1,3 +1,4 @@
+SET explain_query_plan_default = 'legacy';
 DROP TABLE IF EXISTS t_func_to_subcolumns;
 
 SET enable_analyzer = 1;
```
rienath:
```suggestion
DROP TABLE IF EXISTS t_func_to_subcolumns;

SET explain_query_plan_default = 'legacy';
SET enable_analyzer = 1;
```

tests/queries/0_stateless/02346_exclude_materialize_skip_indexes_on_insert.sql
```diff
@@ -1,3 +1,4 @@
+SET explain_query_plan_default = 'legacy';
 -- add_minmax_index_for_numeric_columns=0: Changes the plan and rows read
```
rienath:
```suggestion
-- add_minmax_index_for_numeric_columns=0: Changes the plan and rows read
SET explain_query_plan_default = 'legacy';
```

tests/queries/0_stateless/02346_text_index_function_ilike.sql
```diff
@@ -1,4 +1,5 @@
 -- Tags: no-parallel-replicas
+SET explain_query_plan_default = 'legacy';
```
rienath: There are many files where line 1 (or line 3 like here) explains what the test is for and `SET`s happen later on. It makes sense to leave explanation of what the test is at the top and the existing position of `SET`. Claude can do this fast

rienath (commented):
Thanks for the patience with updating all the tests! While looking, I noticed [these](https://pastila.nl/?0079fdf9/3b1f804b1016dad84791e8c63d674e8b#qZsgW4hfmId2+XlGjFcj3g==GCM) files don't need `explain_query_plan_default = 'legacy'` that was added because they don't contain `EXPLAIN` or `EXPLAIN PLAN`. They have some other form of `EXPLAIN` unaffected by these changes. Could you revert those to keep the diff minimal?

src/Core/Settings.cpp
Fgrtue: I though about bool initially, but then changed it to enum, since will allow us to have multiple different types of default explain in the future.

docs/en/sql-reference/statements/explain.md
Fgrtue: I did not prettify json version yet. But we will do it in the future. I think for now I will just leave the legacy option when json is on. I will leave a note in documentation.

src/Processors/QueryPlan/QueryPlan.h
Fgrtue: Right, good catch. 

I went with this approach: `An alternative is to invert checkAndGetSettings to set these 3 to true`

src/Core/Settings.cpp
```diff
@@ -2451,6 +2451,17 @@ Apply sharding for JOIN if join keys contain a prefix of PRIMARY KEY for both ta
...
+Default format used by `EXPLAIN PLAN`.
+
+Possible values:
+- `pretty` (default since 26.6) — `actions`, `compact`, and `pretty` default to `true`, producing a compact, pretty, action-annotated plan.
```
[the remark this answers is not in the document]
Fgrtue: Addressed in https://github.com/ClickHouse/ClickHouse/pull/105036/commits/ba8c26e66253a4c2afabcc8d23f6b63f3619afc3

tests/integration/test_config_xml_full/configs/config.d/path.xml
```diff
@@ -1,5 +1,5 @@
 <clickhouse>
-    <path replace="replace">./</path>
+    <path replace="replace">/home/ubuntu/click-master/</path>
```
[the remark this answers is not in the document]
Fgrtue: Addressed in https://github.com/ClickHouse/ClickHouse/pull/105036/commits/bf9b68d802c2ac787a9f9b132894925311a96ba0

tests/queries/0_stateless/00944_minmax_nan.reference
```diff
@@ -1,5 +1,3 @@
--- Test for issue #75523
-
```
Fgrtue, line 2: Had to fix the reference in order to leave the comment on top, and set the setting below it.

tests/queries/0_stateless/03460_projection_part_filtering_and_introspection.reference
```diff
@@ -1,5 +1,3 @@
--- add_minmax_index_for_numeric_columns=0: Implicit indices will filter before projections
-
```
Fgrtue, line 2: Had to fix the reference in order to keep the note on the top and set the setting below.

programs/server/config.d/path.xml
```diff
@@ -1,5 +1,5 @@
 <clickhouse>
-    <path replace="replace">./</path>
+    <path replace="replace">/home/ubuntu/click_master/</path>
```
Fgrtue:
```suggestion
    <path replace="replace">./</path>
```

src/Processors/QueryPlan/QueryPlan.cpp
```diff
@@ -611,7 +611,7 @@ void QueryPlan::explainPlan(
     };
 
     auto skip_expressions = [&](Node * node) -> Node * {
-        while (settings.compact && node->step->getName() == "Expression" && !node->children.empty())
+        while (options.actions && settings.compact && node->step->getName() == "Expression" && !node->children.empty())
```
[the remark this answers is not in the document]
Fgrtue: Yes, this is a bug fix, which is done continuously in this PR. Since otherwise the users will have `Expressions` hidden even when they use `actions=0`. 

The documentation which existed before actually says explicitly that `compact=1` only works when `actions=1`.

docs/en/sql-reference/statements/explain.md
```diff
@@ -182,32 +183,44 @@ Settings:
...
-- `distributed` — Shows query plans executed on remote nodes for distributed tables or parallel replicas. Default: 0.
-- `compact` — When enabled, hides expression steps and detailed action info (inputs, functions, aliases, and output positions) from the plan. Only has an effect when actions = 1. Default: 0.
-- `pretty` — Prints the plan tree using line-drawing characters (├──, └──, │) instead of indentation to visualize the hierarchy. Also formats join step properties inline. Default: 0.
+- `distributed` — Shows query plans executed on remote nodes for distributed tables or parallel replicas. Not supported together with `json` or `graph`. Default: 0.
+- `compact` — When enabled, hides expression steps and detailed action info (inputs, functions, aliases, and output positions) from the plan. Only has an effect when `actions = 1`. Default: 0(see the note below).
+- `pretty` — Prints the plan tree using line-drawing characters (├──, └──, │) instead of indentation to visualize the hierarchy. Also formats join step properties inline. Default: 0 (see the note below).
+
+:::note
```
rschu1ze, line 196: Instead of explaining what changed in the new version (from the perspective of the old one), better explain how old versions showed the output (from the perspective of the new version).

Likewise, in l. 186, 193, 194, we should mention the _new_ defaults, not the old ones.

src/Interpreters/InterpreterExplainQuery.cpp
```diff
@@ -513,12 +515,37 @@ struct QuerySyntaxSettings
...
+        if (ast_settings)
+            for (const auto & change : ast_settings->as<ASTSetQuery &>().changes)
+            {
+                if (change.name == "json")
+                    json_requested = change.value.safeGet<UInt64>() != 0;
+                else if (change.name == "distributed")
+                    distributed_requested = change.value.safeGet<UInt64>() != 0;
+            }
+
+        if (set_default_pretty_explain_settings && !json_requested && !distributed_requested)
```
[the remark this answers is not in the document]
Fgrtue: > This json/distributed guard is now part of the documented compatibility contract, but the current tests do not exercise it under the new default

I don't think the test is needed. If we add `distributed` and `json` functionality, we will have to remove the test. 

> Please add a focused test that leaves explain_query_plan_default at its default and verifies EXPLAIN json = 1/EXPLAIN distributed = 1 still use the legacy actions = 0, compact = 0, pretty = 0 defaults, plus an explicit actions = 1 case if that override is intended.

This is a limitation that is currently documents and presents in the code. Since this is not a critical limitation -- nothing shall crash, the user will just get a bit poorly formatted output, I don't think in this particular case it is worth addressing it in the test.

docs/en/sql-reference/statements/explain.md
```diff
@@ -182,32 +183,44 @@ Settings:
...
 - `input_headers` — Prints input headers for step. Default: 0. Mostly useful only for developers to debug issues related to input-output header mismatch.
 - `column_structure` — Prints also the structure of columns in headers on top of their name and type. Default: 0. Mostly useful only for developers to debug issues related to input-output header mismatch.
-- `distributed` — Shows query plans executed on remote nodes for distributed tables or parallel replicas. Default: 0.
-- `compact` — When enabled, hides expression steps and detailed action info (inputs, functions, aliases, and output positions) from the plan. Only has an effect when actions = 1. Default: 0.
-- `pretty` — Prints the plan tree using line-drawing characters (├──, └──, │) instead of indentation to visualize the hierarchy. Also formats join step properties inline. Default: 0.
+- `distributed` — Shows query plans executed on remote nodes for distributed tables or parallel replicas. Not supported together with `json` or `graph`. Default: 0.
```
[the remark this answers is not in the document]
Fgrtue: Fixed in https://github.com/ClickHouse/ClickHouse/pull/105036/commits/f30cfdafd088961ae12a6bc48b0304e19226e322

docs/en/sql-reference/statements/explain.md
Fgrtue: I fixed this in https://github.com/ClickHouse/ClickHouse/pull/105036/commits/f30cfdafd088961ae12a6bc48b0304e19226e322. 

The comment in the `::note` section should be better now. Speaking about defaults, I left them as `0` because the struct with the settings has it set to `false`, and the defaults are overriden further in the EXPLAIN internals. But you are right, since the default user-seen behavior changes, it is more correct to say that defaults are `1`.

docs/en/sql-reference/statements/explain.md
Fgrtue, line 1: Side note about docs: I think they need to be rewritten. However, I don't think this is the right time for rewriting them fully now. I plan to do some more fixed to explain (among which supporting json mode for pretty, adding EXPLAIN ANALYZE, making EXPLAIN to work properly in distributed queries). And I see more sense in doing a major update for docs after that, so that it is done once and for all. Until that I will be updating them incrementally.

docs/en/sql-reference/statements/explain.md
```diff
@@ -182,39 +183,55 @@ Settings:
...
+
+Prior to ClickHouse 26.6 the defaults for `actions`, `compact`, and `pretty` were `0`. You can still get that output by setting `explain_query_plan_default = 'legacy'` (globally or in per-query `SETTINGS`), or by setting `compatibility` to any version older than `26.6`.
+
+The combinations of these options control how much detail is shown:
+- `actions = 1, pretty = 0` — shows the internal (machine) names of tables and expressions, and conveys the parent-child relation between steps through indentation instead of line-drawing characters.
+- `actions = 1, compact = 0` — shows the full plan with all expression steps and detailed action info (inputs, functions, aliases, and output positions).
+- `actions = 0` — shows only step descriptions; in this case `compact` has no effect, and `pretty` only switches between line-drawing characters and indentation.
```
rienath: nit: this paraphrases the explanation of `actions`. `pretty` and `compact` from above. I think documentation would be more lightweight without this

docs/en/sql-reference/statements/explain.md
```diff
@@ -462,7 +479,7 @@ EXPLAIN json = 1, actions = 1, description = 0 SELECT 1 FORMAT TSVRaw;
 With `compact = 1`, each `Expression` step is removed. Along with that, if `actions = 1` is set, then `Actions` and `Positions` lines are hidden, leaving only the step descriptions:
 
 ```sql
-EXPLAIN actions = 1, compact = 1 SELECT sum(number) FROM numbers(10) GROUP BY number % 4 FORMAT Raw;
+EXPLAIN actions = 1, compact = 1 SELECT sum(number) FROM numbers(10) GROUP BY number % 4 FORMAT Raw SETTINGS explain_query_plan_default = 'legacy';
```
rienath: Could you remind me why we want legacy here?

src/Interpreters/InterpreterExplainQuery.cpp
```diff
@@ -513,12 +515,23 @@ struct QuerySyntaxSettings
 };
 
 template <typename Settings>
-ExplainSettings<Settings> checkAndGetSettings(const ASTPtr & ast_settings)
+ExplainSettings<Settings> checkAndGetSettings(const ASTPtr & ast_settings, bool set_default_pretty_explain_settings = true)
 {
+    ExplainSettings<Settings> settings;
+
+    if constexpr (std::is_same_v<Settings, QueryPlanSettings>)
```
rienath, line 527: It'd be nice to have a brief comment why default value is F and we set it to T here (the loggers problem). S.t. future maintainer doesn't delete this by mistake and switch to T by default

Fgrtue:
@groeneai please take a look at this test: 02346_text_index_map_support

It fails here: [Stateless tests (amd_asan_ubsan, flaky check)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105036&sha=f30cfdafd088961ae12a6bc48b0304e19226e322&name_0=PR&name_1=Stateless%20tests%20%28amd_asan_ubsan%2C%20flaky%20check%29)

groeneai:
@Fgrtue looked into it. This is a **timing** failure, not a logic/reference problem, and it is not caused by your PR.

The asan_ubsan flaky-check run took 184.5s and tripped the flaky-check time ceiling:

> Reason: Test runs too long (> 180s). Make it faster.

Report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105036&sha=f30cfdafd088961ae12a6bc48b0304e19226e322&name_0=PR&name_1=Stateless%20tests%20%28amd_asan_ubsan%2C%20flaky%20check%29

Why it is not your change:
- The gate in `tests/clickhouse-test` (~L3227) fails a run only when `flaky_check and total_time > 180s and "long" not in tags`. It is a per-run wall-clock check, not a result diff.
- Every other sanitizer's flaky-check passed (msan ~120s, tsan ~30s, debug ~130s). A reference/result mismatch would fail on all of them; only the slowest run (asan_ubsan, 184.5s) crossed the line.
- The only change your PR makes to this file is `SET explain_query_plan_default = 'legacy';`, which has no effect on runtime.
- The test is inherently heavy under sanitizers and already borderline on master: over the last 30 days its runtime regularly nears/exceeds 180s (max 374s on 06-03) and it has tripped TOO_LONG on several unrelated days (05-21, 05-29, 06-06, 06-08, 06-09).

Fix: tag the test `long`. That exempts it from the 180s flaky-check ceiling and also drops its flaky-check repeat count to 10% (`--long-test-runs-ratio`, default 0.1). The sibling `02346_text_index_function_hasAnyAllTokens` is already tagged this way for the same reason. Since the file is already in your PR, just change line 1:

```
-- Tags: no-parallel-replicas, long
```

That one change should clear the flaky check.

src/Core/Settings.cpp
```diff
@@ -2496,6 +2496,20 @@ Apply sharding for JOIN if join keys contain a prefix of PRIMARY KEY for both ta
...
+- `legacy` — pre-26.6 output.
+
+Specifying the `actions`, `compact`, or `pretty` options explicitly in the `EXPLAIN` statement (for example, `EXPLAIN actions = 0, compact = 0, pretty = 0 SELECT ...`) always overrides this setting.
+
+`EXPLAIN PLAN` with `json = 1` or `distributed = 1` keeps the legacy (pre-26.6) defaults regardless of this setting, unless `actions`, `compact`, or `pretty` are set explicitly. The pretty output cannot represent JSON results or per-shard distributed plans, so those modes are only rendered correctly in legacy form.
+
+Set this setting to `legacy` (or set `compatibility` to any version older than `26.6`) to restore the pre-26.6 verbose output.
```
rienath: nit: could drop this, above explains the same

tests/integration/test_distributed_plan_replicated_merge_tree/test.py
```diff
@@ -83,6 +83,8 @@
     "query_plan_convert_any_join_to_semi_or_anti_join = 0",
     # Runtime filters are not yet implemented for distributed queries.
     "enable_join_runtime_filters = 0",
+    # To suppor old explain format
```
rienath:
```suggestion
    # To support old explain format
```

tests/queries/0_stateless/02346_text_index_match_predicate.sql
```diff
@@ -1,8 +1,6 @@
 -- Tags: no-parallel-replicas
-
 -- Tests that match() utilizes the text index
-
-DROP TABLE IF EXISTS tab;
```
rienath, line 5: Probably deleted by mistake

tests/queries/0_stateless/03392_crash_group_by_use_nulls.sql
```diff
@@ -1,4 +1,5 @@
...
+
```
rienath:
```suggestion
```

tests/queries/0_stateless/03821_join_use_nulls_outer_to_inner.sql
```diff
@@ -1,3 +1,4 @@
+SET explain_query_plan_default = 'legacy';
 #!/usr/bin/env -S ${HOME}/clickhouse-client --queries-file
```
rienath: Other way around?

rienath (approved):
I left a final round of nits. The PR looks good! Feel free to merge when you're ready

docs/en/sql-reference/statements/explain.md
Fgrtue: We wanted it because it was illustrative for how the compact influences the result when it is turned on.

However, you are right that now it makes more sense not to use 'legacy', and instead show an example with `compact = 0`.

tests/queries/0_stateless/03392_crash_group_by_use_nulls.sql
Fgrtue: I like the line, but yeah, should keep the diff minimal

Fgrtue:
@groeneai investigate the failure in this test suit: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105036&sha=8f438483ac39c6874f50859392159b2d711e1f61&name_0=PR&name_1=Stress%20test%20%28arm_ubsan%29

Check if there are any relevant issues or pull requests already opened that are related to this failure. Here is the name of the error that occurred: `Logical error: Unexpected return type from A. Expected B. Got C (STID: 3262-40aa)`

If there are no relevant issues or PRs found -- create an issue and a PR that will solve this problem.

PR #105039, opened 2026-05-15
Read all pre-existing index columns when force-recalculating in mutation

### LINKED ISSUE #104872: MATERIALIZE INDEX DB::Exception: Not found column vid: in block / NOT_FOUND_COLUMN_IN_BLOCK
### Company or project name

https://www.innervate.com/

### Describe what's wrong

part created in 25.8 / 25.3 fail to mutate after upgrading to 26.3

Insert some data. Add column `vid` create index over `vid`. Upgrade to 26.3 / 26.4. Create index over old column (requestID). Materialize index fails with NOT_FOUND_COLUMN_IN_BLOCK because the column `vid` was not materialized and did not exist in the part.

---

There is no issue without upgrading to 26.3 (25.8 is able to materialize the second index).

Yes

### How to reproduce

```
apt install clickhouse-server=25.8.23.13 clickhouse-client=25.8.23.13 clickhouse-common-static=25.8.23.13

CREATE TABLE test_repro (timestamp DateTime, requestID String) 
ENGINE = MergeTree() ORDER BY timestamp
as select * from values( (now(), 'aaa'), (now(), 'bbb'));

ALTER TABLE test_repro ADD COLUMN vid Int64;
ALTER TABLE test_repro ADD INDEX vid_ix vid TYPE bloom_filter GRANULARITY 100;
ALTER TABLE test_repro MATERIALIZE INDEX vid_ix settings mutations_sync=2;

SELECT part_type, name FROM system.parts WHERE (`table` = 'test_repro') AND active;
┌─part_type─┬─name────────┐
│ Compact   │ all_1_1_0_2 │
└───────────┴─────────────┘

$ ls -l /var/lib/clickhouse/data/default/test_repro/all_1_1_0_2
 356 May 13 16:07 checksums.txt
 137 May 13 16:07 columns_substreams.txt
  77 May 13 16:07 columns.txt
   1 May 13 16:07 count.txt
  68 May 13 16:07 data.bin
  58 May 13 16:07 data.cmrk4
  10 May 13 16:07 default_compression_codec.txt
   1 May 13 16:07 metadata_version.txt
  42 May 13 16:07 primary.cidx
 161 May 13 16:07 serialization.json
  50 May 13 16:07 skp_idx_vid_ix.cmrk4
  28 May 13 16:07 skp_idx_vid_ix.idx   ---<<<-- vid index

$ cat /var/lib/clickhouse/data/default/test_repro/all_1_1_0_2/columns.txt 
columns format version: 1
2 columns:
`timestamp` DateTime
`requestID` String    ------<<<-- no vid column

apt install clickhouse-server=26.3.10.60 clickhouse-client=26.3.10.60 clickhouse-common-static=26.3.10.60
systemctl restart clickhouse-server

ALTER TABLE test_repro ADD  INDEX requestID_ix requestID TYPE bloom_filter GRANULARITY 100;
ALTER TABLE test_repro MATERIALIZE INDEX requestID_ix  settings mutations_sync=2;

Received exception from server (version 26.3.10):
Code: 341. DB::Exception: Received from localhost:9000. DB::Exception: 
Exception happened during execution of mutation 'mutation_3.txt' with part 'all_1_1_0_2'
 reason: 'Code: 10. DB::Exception: Not found column vid: in block timestamp DateTime UInt32(size = 0), requestID String String(size = 0). (NOT_FOUND_COLUMN_IN_BLOCK) 
(version 26.3.10.60 (official build))'. 


apt install clickhouse-server=26.4.2.10 clickhouse-client=26.4.2.10 clickhouse-common-static=26.4.2.10
systemctl restart clickhouse-server

select  * from system.mutations;

latest_fail_reason: Code: 10. DB::Exception: Not found column vid: in block timestamp DateTime UInt32(size = 0), 
requestID String String(size = 0). (NOT_FOUND_COLUMN_IN_BLOCK) (version 26.4.2.10 (official build))
```

### Expected behavior

_No response_

_No response_

### Additional context

No issue with wide parts `settings min_bytes_for_wide_part=0`

### Version info
- Resolved by: #105039
- Merged into: `26.6.1.255`

groeneai:
Fixes the regression reported in #104872 — `MATERIALIZE INDEX` on a
table containing a part that was created before the column was added
fails with `NOT_FOUND_COLUMN_IN_BLOCK` after upgrading to 25.10/25.11/
25.12/26.3/26.4/master.

### Root cause

PR #91980 ("Force mutation to recalculate indices when rewriting compact
parts") widened `MutateFromLogEntryTask::prepare`'s force-recalculate
predicate from `!is_full_part_storage` to `!is_full_wide_part`. As a
consequence every mutation on a `Compact` part now force-recalculates
every pre-existing skip index on the part.

`splitAndModifyMutationCommands` was updated (commits 2713ca0c9af and
followups) to add the *explicitly-materialized* index's columns to the
mutation read set, but it never added the columns of *other* pre-existing
indices that the new force-recalculate predicate would also rebuild. So
when a part on disk has `skp_idx_X.idx` but no `X` column in
`columns.txt` (typical state of a part created on 25.8 where
`MATERIALIZE INDEX` did not yet write the index's required columns),
the next mutation rebuilds `X_ix`, sees the block has no `X` column,
and throws.

@den-crane's [analysis on the issue](https://github.com/ClickHouse/ClickHouse/issues/104872#issuecomment-3520148193)
pinpointed PR #91980 and the exact `MutateTask.cpp` lines.

### Fix

In `splitAndModifyMutationCommands`, after the rename pass, walk every
secondary index that the source part already carries (minus the ones
being dropped) and add their absent columns to
`extra_columns_for_indices_and_projections`. That set is converted into
synthetic `READ_COLUMN` commands further down. The mutation pipeline
will then read the column (synthesised from default if missing on disk),
the force-recalculation pass in `prepare` has access to it, and the
new part is written with the column physically present — healing the
broken state in the process.

The symmetric gap exists for projections, which are also force-rebuilt
in `prepare` for non-full-storage parts. Apply the same treatment so
the partial-fix surface area stays consistent.

### Testing

* Existing `03755_ttl_materialize_bug` (the test added by PR #91980) keeps passing.
* New `04246_materialize_index_force_recalc.sh`:
  - Exercises the user's exact `ADD COLUMN` → `ADD INDEX vid_ix` →
    `MATERIALIZE INDEX vid_ix` → `MATERIALIZE INDEX requestID_ix`
    sequence on a `Compact` part.
  - Drives an adversarial Compact part with two pre-existing skip
    indices through `DELETE`, a fresh `MATERIALIZE INDEX` over an
    unrelated column, and a `MODIFY TTL`. Each of those mutations now
    force-recalculates both pre-existing indices via the `prepare`
    predicate; the test verifies all three succeed and the resulting
    indices still produce the expected lookups.

A pure stateless reproducer for the original on-disk shape (skip-index
file present, indexed column missing from `columns.txt`) requires a
part literally written by 25.8 (where `MATERIALIZE INDEX` did not yet
populate the column). The manual reproducer from the issue still applies
on this branch and now succeeds end-to-end.

### Pre-PR validation gate

a) Deterministic repro? Yes — the manual reproducer from the issue
   (`CREATE TABLE`, `ADD COLUMN vid`, `ADD INDEX vid_ix`,
   `MATERIALIZE INDEX vid_ix` on 25.8 binary; upgrade to 26.3;
   `ADD INDEX requestID_ix`, `MATERIALIZE INDEX requestID_ix`). On a
   master binary with this fix the second `MATERIALIZE INDEX` succeeds
   on parts created by 25.8.
b) Root cause explained? Yes — see "Root cause" above.
c) Fix matches root cause? Yes — addresses the mechanism (force-recalc
   pass demands columns the read set does not provide) by extending
   the read set to include every pre-existing index/projection's
   required columns.
d) Test intent preserved / new tests added? Yes — adds
   `04246_materialize_index_force_recalc.sh` (user's reproducer plus an
   adversarial Compact-multi-index scenario).
e) Demonstrated in both directions? On master without the fix, the
   issue's reproducer fails on parts created by 25.8. On master with
   the fix it succeeds. Existing `03755_ttl_materialize_bug` keeps
   passing.
f) Fix is general, not a narrow patch? Yes — handles every pre-existing
   index and projection on the source part, not just the explicitly-
   materialized one. Also extends the analogous projection path.

Fix `NOT_FOUND_COLUMN_IN_BLOCK` thrown by `ALTER TABLE ... MATERIALIZE INDEX` on parts that were created in 25.8 and contain a skip index over a column that was added with a separate `ALTER TABLE ... ADD COLUMN`. The mutation now correctly reads every column required by every pre-existing skip index and projection on the part during force-recalculation. Closes #104872.

### Version info
- Merged into: `26.6.1.255`


groeneai:
cc @alesapin @CurtizJ — could you review? The fix extends `splitAndModifyMutationCommands` to add columns of every pre-existing skip index (and projection) the source part already carries, so the read set covers every index that `MutateFromLogEntryTask::prepare` will force-recalculate on a Compact part — not only the explicitly-materialized one.

cc @den-crane — thanks for the root-cause analysis on the issue. The fix lines up with your option "extend `extra_columns_for_indices_and_projections` to include columns required by ALL pre-existing indices that will be force-recalculated." Pinging in case you want to weigh in on the exact placement / the projection-symmetric extension.

Session id: cron:clickhouse-ci-task-worker:20260515-141500

groeneai:
### Pre-PR Validation Gate

**a) Deterministic repro?** Yes — the original issue's reproducer (`CREATE TABLE`, `ADD COLUMN vid`, `ADD INDEX vid_ix`, `MATERIALIZE INDEX vid_ix` on a 25.8 binary; upgrade to 26.3; `ADD INDEX requestID_ix`, `MATERIALIZE INDEX requestID_ix`) deterministically fails with `NOT_FOUND_COLUMN_IN_BLOCK` on every affected version, and succeeds against a master binary that carries this patch.

**b) Root cause explained?** Yes. `MutateFromLogEntryTask::prepare` builds `need_recalculate = ctx->materialized_indices.contains(idx.name) || (!is_full_wide_part && ctx->source_part->hasSecondaryIndex(idx.name, ...))`. After #91980 widened the second clause from `!is_full_part_storage` to `!is_full_wide_part`, every Compact part force-recalculates every pre-existing skip index. `splitAndModifyMutationCommands` only added the explicitly-materialized index's required columns to `extra_columns_for_indices_and_projections`. When the source part carries another index over a column that is in the table metadata but absent from `columns.txt` (typical 25.8 part state, since 25.8's `MATERIALIZE INDEX` did not yet write the indexed column to the new part), that column is missing from the read set and `prepare` throws `NOT_FOUND_COLUMN_IN_BLOCK` while rebuilding the unrelated index.

**c) Fix matches root cause?** Yes. The new pass in `splitAndModifyMutationCommands` mirrors `prepare`'s predicate exactly — it iterates over secondary indices that the source part has and that are not being dropped, and for each one adds the absent columns to `extra_columns_for_indices_and_projections`. Same for projections. The mutation pipeline now reads every column the force-recalculation pass will need.

**d) Test intent preserved / new tests added?** Yes. `03755_ttl_materialize_bug` (the test added with #91980 for compact-part mark-count differences) keeps passing locally — the fix only adds reads, it does not change recalculation behaviour. New `04246_materialize_index_force_recalc.sh` exercises the user's exact reproducer plus an adversarial multi-index compact-part scenario across `DELETE`, `MATERIALIZE INDEX`, and `MODIFY TTL`.

**e) Demonstrated in both directions?** On a binary without the fix, the issue's reproducer fails with the exact error message (verified against a 26.3 build; verified manually on the local debug build against the synthesised broken on-disk state). On a binary with the fix, the same reproducer succeeds end-to-end. Existing tests keep passing.

**f) Fix is general, not a narrow patch?** Yes:
  - Covers *every* pre-existing skip index on the source part, not just the one explicitly being materialised.
  - Applies the symmetric pre-existing-projection handling, because `prepare` also force-rebuilds projections on packed-storage parts.
  - Skips indices/projections that are being dropped (their columns are not needed and may not exist in metadata anyway).
  - Reuses the existing `extra_columns_for_indices_and_projections` → synthetic `READ_COLUMN` plumbing, so the fix is local to one code path and consistent with surrounding logic.

Session id: cron:clickhouse-ci-task-worker:20260515-141500

den-crane:
tests/queries/0_stateless/04246_materialize_index_force_recalc.sh

Stateless tests are almost useless here. This issue does not reproduce if a table is created in the modern version (26.3).
You need a stateful test running against a table created in 25.8. Try to store in the test repo a part created in 25.8 and attach it in a test.

groeneai:
@den-crane — addressed in 3ef48418679.

I ran the user's reproducer against `clickhouse/clickhouse-server:25.8` in Docker and captured the resulting Compact part. The on-disk shape confirms your point: in 25.8 `MATERIALIZE INDEX vid_ix` writes `skp_idx_vid_ix.idx` + `skp_idx_vid_ix.cmrk4` but `columns.txt` still lists only the original `timestamp` and `requestID` — `vid` is not present in the part. The captured part is now committed at `tests/queries/0_stateless/data_104872/part_25.8.tar.gz` (1.2 KiB).

The new `04246_materialize_index_force_recalc.sh` creates an empty 26.x table with the same logical schema (`timestamp`, `requestID`, `vid` + `vid_ix`), extracts the captured part into the table's `detached/` directory, ATTACHes it, and then runs the user's `ADD INDEX requestID_ix` + `MATERIALIZE INDEX requestID_ix`. A subsequent DELETE covers the non-`MATERIALIZE INDEX` mutation entry point.

Verified locally in both directions:
- Without the `MutateTask.cpp` fix: `MATERIALIZE INDEX requestID_ix` fails with `Code: 10. DB::Exception: Not found column vid: in block timestamp DateTime UInt32(size = 0), requestID String String(size = 0). (NOT_FOUND_COLUMN_IN_BLOCK)` — exactly the user's error.
- With the fix: 20/20 passes under `--test-runs 20` with full CI randomization enabled.

Session id: cron:clickhouse-ci-task-worker:20260515-164500

groeneai:
Followup on the CI signal at https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105039&sha=3ef4841867973382f97c7b6c55bb81c9128014f3&name_0=PR — pushed `da5382c8a6f` to add `no-object-storage` to the test tags.

After the `3ef4841` switch to the captured-25.8-part fixture, the test reproducibly fails (3/3 reruns) on every `s3 storage` / `azure` stateless variant:
- `Stateless tests (amd_tsan, s3 storage, parallel, 1/2)`
- `Stateless tests (amd_debug, distributed plan, s3 storage, parallel)`
- `Stateless tests (amd_llvm_coverage, AsyncInsert, s3 storage, parallel)`
- `Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel)`
- `Stateless tests (arm_asan_ubsan, azure, parallel)`

Root cause: the captured 25.8 part is a local-disk fixture — `columns.txt` has real content (`columns format version: 1\n2 columns:...`) and so do all the other part files. On `s3` / `azure` disks those files in the table's local data directory are not data — they are `DiskObjectStorageMetadata` pointer files (`<version>\n<num_objects>\n<size> <object_key>...`) referring to objects in the bucket. Extracting the captured part directly into the local `detached/` dir therefore makes `MetadataStorageFromDisk::readMetadataUnlocked` → `DiskObjectStorageMetadata::deserialize` parse the literal `columns format version: 1\n2 columns:...` and fail at the second line, producing `CANNOT_PARSE_INPUT_ASSERTION_FAILED` and `Part .../detached/attaching_all_1_1_0_2 is broken` at ATTACH time:

```
Code: 27. DB::Exception: Cannot parse input: expected '\n' before: 'columns format version: 1\n2 columns:\n`timestamp` DateTime\n`requestID` String\n': while parsing: 'columns format version: 1
2 columns:
`timestamp` DateTime
`requestID` String
'. (CANNOT_PARSE_INPUT_ASSERTION_FAILED), Stack trace ...
 6. src/IO/ReadHelpers.h:212:9: DB::DiskObjectStorageMetadata::deserialize(DB::ReadBuffer&)
 7. src/Disks/DiskObjectStorage/MetadataStorages/DiskObjectStorageMetadata.cpp:96:5: DB::DiskObjectStorageMetadata::deserializeFromString(String const&)
 8. src/Disks/DiskObjectStorage/MetadataStorages/Local/MetadataStorageFromDisk.cpp:96:15: DB::MetadataStorageFromDisk::readMetadataUnlocked(...)
```

Rebuilding the fixture inside an `s3`-backed 25.8 server would require also uploading every part file into the test's MinIO bucket and writing the matching metadata pointers — none of which can be done portably from a stateless test. The bug under test (column-extraction in `splitAndModifyMutationCommands`) is fully independent of the disk layer, so exercising it on local disk is sufficient.

`no-object-storage` is the standard tag for this case (`tests/clickhouse-test:2649-2652` triggers the skip on `--s3-storage` or `--azure-blob-storage` invocations). Existing tests that store a local-disk fixture under `data_*/` follow the same pattern (e.g. `02944_dynamically_change_filesystem_cache_size`, `02361_fsync_profile_events`, `02263_lazy_mark_load`).

Session id: cron:clickhouse-ci-task-worker:20260518-1300

den-crane:
.

den-crane:
@groeneai please merge changes from the master

groeneai:
@den-crane Merged master into the branch in c58803637. CI will re-run.

PR #58724, opened 2024-01-11
Do not promote float32 automatically

Algunenano:
### Changelog category (leave one):
- Improvement

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
- When comparing a Float32 column and a const string, read the string as Float32 (instead of Float64).

### Documentation entry for user-facing changes

https://github.com/ClickHouse/ClickHouse/issues/58680

PR #105042, opened 2026-05-15
Fix BFloat16 string-equality returning zero rows

### LINKED ISSUE #103682: `convertFieldToTypeImpl` skips promotion only for `Float32`, leaving `BFloat16` string-equality returning zero rows

_Retrospective finding from a historical scan of [PR #58724](https://github.com/ClickHouse/ClickHouse/pull/58724) (merged 2024-01-13). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`SELECT count() FROM t WHERE bfloat16_col = '49.9'` returns 0 even when `bfloat16_col` stores the BFloat16 representation of 49.9 (49.75). Same for any string whose `Float64` parse is not exactly representable in `BFloat16` (e.g., '0.1', '1.1', '3.14').

**Root cause:** src/Interpreters/convertFieldToType.cpp:771: the promotion guard exempts only `Float32` (`!which_type.isFloat32()`). `BFloat16` (added 2024-01-13, after this PR) is also a lossy float type with `NearestFieldType=Float64`, so it has the identical precision-loss problem the PR fixed for Float32 but is not exempted.

**Why we believe this is a bug:** FunctionsComparison.h:956 → `convertFieldToType(string_value, BFloat16)` → `convertFieldToTypeImpl` (convertFieldToType.cpp:762-794) takes the String branch → because `BFloat16` `canBePromoted()=true` (DataTypesNumber.h:34) and `which_type.isFloat32()` is false, the type is promoted to `DataTypeNumber<Float64>` (Field.h:213) → '49.9' is parsed as Float64(49.9000000000000057) → recursive `convertFieldToType` reaches `convertNumericType<BFloat16>` (line 332) → `convertNumericTypeImpl<Float64, BFloat16>` → `accurate::convertNumeric` (AccurateComparison.h:194-196) does `result = static_cast<BFloat16>(49.9000000000000057) = 49.75`, then strict-checks `equalsOp(49.9000000000000057, 49.75)` → false → returns Null Field → FunctionsComparison.h:959-962 treats Null as 'not equals' and returns false for every row.

**Affected locations:**
- `src/Interpreters/convertFieldToType.cpp:771` — Promotion guard exempts only Float32, not BFloat16

**Impact:** User-visible wrong results: `WHERE bfloat16_col = 'string_literal'` silently returns zero rows for any common decimal value that is not exactly representable in BFloat16 (8-bit mantissa, only ~3 significant decimal digits). Affects all comparison operators (`=`, `IN`), constant folding via FunctionsComparison.h:956, and any caller of `convertFieldToType` with a string source and BFloat16 target. BFloat16 is GA (allow_experimental_bfloat16_type is now MAKE_OBSOLETE per Settings.cpp:7994).

Yes — confirmed on current `master` (commit `382a7e11e0ad`).

### How to reproduce

```sql
-- Tests that BFloat16 columns can be compared against string literals matching their stored value.
-- Mirrors PR #58724's fix for Float32: convertFieldToType promotes BFloat16 to Float64 before parsing
-- the string, then the strict Float64->BFloat16 cast back fails because 49.9 (Float64) != 49.75 (BFloat16),
-- producing a Null Field that yields "not_equals" -- so WHERE x = '49.9' returns 0 rows even though
-- BFloat16(49.9) is stored as 49.75 and would otherwise match.
DROP TABLE IF EXISTS bf16_table;
CREATE TABLE bf16_table (my_field BFloat16) ENGINE=Memory();
INSERT INTO bf16_table values (49.9);
-- The stored value is BFloat16(49.9) = 49.75. Comparison should return 1 row.
SELECT count() FROM bf16_table WHERE my_field = '49.9';
DROP TABLE bf16_table;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/b688217f-b7ad-42c3-86fd-66c733e5a529)

### Expected behavior

```
1
```

```
0
```

### Additional context

**Open risks:**
- Skipping promotion makes BFloat16 parse strings via `SerializationNumber<BFloat16>::deserializeWholeText` directly. `toBFloat16('49.9')` confirms this works and produces 49.75; if any existing input format relied on Float64 parsing precision before narrowing, that would change.
- Other narrow float types added in the future (e.g., Float16) would need the same exemption.

**Suggested fix:** Extend the guard at convertFieldToType.cpp:771 to also exempt BFloat16: `if (type.canBePromoted() && !which_type.isFloat32() && !which_type.isBFloat16() && !type.getCustomSerialization())`. The promoted-then-strict-cast pattern fundamentally breaks for any lossy float; consider exempting all float types that are narrower than Float64.

Found during automated review of [PR #58724](https://github.com/ClickHouse/ClickHouse/pull/58724).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr58724_001`_

### Version info
- Resolved by: #105042
- Merged into: `26.5.1.846`

Algunenano:
`convertFieldToType` promoted `BFloat16` to `Float64` when parsing a string literal, so `'49.9'` became `Float64(49.9000000000000057)` and the strict `Float64 -> BFloat16` cast back inside `accurate::convertNumeric` failed (`49.9 != 49.75`), producing a `Null` `Field` and silently zero-matching comparisons like `WHERE bf16_col = '49.9'`.

Extend the promotion guard to also exempt `BFloat16`, mirroring the `Float32` exemption added in #58724.

Closes #103682

Fix `BFloat16` columns silently returning zero matches when compared against a string literal (e.g. `WHERE bf16_col = '49.9'`).

### Version info
- Merged into: `26.5.1.846`


alexey-milovidov (approved):

Algunenano:
https://github.com/ClickHouse/ClickHouse/pull/104996

clickgapai:

**Affected code:** `convertFieldToTypeImpl` string promotion guard at src/Interpreters/convertFieldToType.cpp:771 — core code present in all supported branches.

**Why this might apply:** BFloat16 data type was introduced in late 2024 (commit 6ea31c81e1d from November 2024). All currently supported branches (25.8 LTS released Aug 2025, 26.2-26.4) contain this type and thus have this bug. The bug causes silent data loss (zero matches instead of correct results) for user queries, making it a P1 candidate for backporting.

PR #105054, opened 2026-05-15
Fix JIT-compiled `intExp2` for out-of-range inputs

### LINKED ISSUE #101540: JIT-compiled intExp2 lacks overflow/bounds checks, returns wrong results for shift >= 64

_Retrospective finding from a historical scan of [PR #75312](https://github.com/ClickHouse/ClickHouse/pull/75312) (merged 2025-01-31). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

When compile_expressions=1 and min_count_to_compile_expression=0, intExp2(64) returns 0 instead of 18446744073709551615 (UINT64_MAX). intExp2(67) returns 8 instead of UINT64_MAX.

**Root cause:** intExp2.cpp:52 — `return b.CreateShl(llvm::ConstantInt::get(arg->getType(), 1), arg)` has no bounds check for arg >= 64 or arg < 0, unlike the interpreted path.

**Affected locations:**
- `src/Functions/intExp2.cpp:52` — JIT compile function emits unbounded shl instruction

**Impact:** Silent wrong results for intExp2 with JIT compilation enabled when input values are >= 64. The function returns 0 or wrapped garbage instead of UINT64_MAX.

Yes — confirmed on current `master` (commit `4b89225ae7a9`).

### How to reproduce

```sql
SELECT 'non-jit';
SELECT intExp2(toInt32(number + 63)) FROM numbers(5) ORDER BY number SETTINGS compile_expressions=0;
SELECT 'jit';
SELECT intExp2(toInt32(number + 63)) FROM numbers(5) ORDER BY number SETTINGS compile_expressions=1, min_count_to_compile_expression=0;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/89fc1300-5afd-4b7a-8774-9293b56cb4f6)

### Expected behavior

```
non-jit
9223372036854775808
18446744073709551615
18446744073709551615
18446744073709551615
18446744073709551615
jit
9223372036854775808
18446744073709551615
18446744073709551615
18446744073709551615
18446744073709551615
```

```
non-jit
9223372036854775808
18446744073709551615
18446744073709551615
18446744073709551615
18446744073709551615
jit
9223372036854775808
0
0
0
8
```

### Additional context

**Open risks:**
- Any query using intExp2 with JIT compilation on integer columns with values >= 64 silently gets wrong results

**Suggested fix:** Add LLVM IR bounds checks in the compile() function: clamp the shift amount to [0, 63] range and return appropriate constants (0 for negative, UINT64_MAX for >= 64), or set compilable=false to disable JIT for this function.

Found during automated review of [PR #75312](https://github.com/ClickHouse/ClickHouse/pull/75312).

Algunenano:
The JIT path emitted `shl 1, arg` without bounds, so `intExp2(toInt32(64))`, `intExp2(toInt32(67))`, etc. returned `0`/`8`/garbage instead of `UINT64_MAX` when `compile_expressions = 1`. The non-JIT path was also buggy for integers with magnitude `> INT_MAX` because it did `intExp2(static_cast<int>(a))`, whose narrowing is implementation-defined for values that don't fit in `int` (e.g. `intExp2(Int64{INT64_MAX})` returned `0` instead of `UINT64_MAX`, `intExp2(UInt32{0xFFFFFFFB})` returned `0` instead of `UINT64_MAX`).

- JIT: disabled for floats and big integers (`compilable` gated on `is_integer && !is_big_int_v`), so `NaN` still throws and `Int128`/`UInt128` still return `NOT_IMPLEMENTED` via the C++ path. For native integer inputs, `IntExp2Impl` is added to the special-case bypass in `FunctionUnaryArithmetic` so `compile()` receives the original argument and signedness; the emitted IR sign-extends to `i64`, returns `0` for signed negatives, `UINT64_MAX` for inputs `> 63`, and `1 << (arg & 63)` otherwise.
- `apply()`: hoist the `< 0 -> 0` and `>= 64 -> UINT64_MAX` checks out of the floating-point branch so integers are range-checked before the narrowing to `int`. Use `a != a` for the `NaN` check so it works for `BFloat16` too.

After this change, JIT and non-JIT produce identical results across `Int8..Int64`, `UInt8..UInt64`, `Float32`/`Float64` — verified by sweeping boundary, negative, and `|a| > INT_MAX` values.

Closes #101540

- Improvement

Fix `intExp2` returning wrong results for out-of-range inputs (shifts `>= 64` under JIT, and integers with magnitude exceeding `INT_MAX` on both code paths).

### Version info
- Merged into: `26.5.1.696`


alexey-milovidov:
Ok, but I can't count it as a bug fix. Implementation-defined behavior on overflow was in the original function's contract. Changed to Improvement.

src/Functions/intExp2.cpp
```diff
@@ -29,31 +26,69 @@ struct IntExp2Impl
...
+
+        /// Negative inputs return 0 (matches `intExp2(int x)` for `x < 0`).
+        /// For unsigned inputs the branch folds away because `is_negative` is a constant `false`.
+        llvm::Value * is_negative = is_signed
+            ? static_cast<llvm::Value *>(b.CreateICmpSLT(arg_i64, zero64))
+            : static_cast<llvm::Value *>(b.getFalse());
+
+        /// Inputs `> 63` saturate to `UINT64_MAX`. The mask makes the `shl` defined when
+        /// `is_too_large` is true (the result of `shl` is then discarded by the outer select).
+        auto * is_too_large = b.CreateICmpUGT(arg_i64, sixty_three);
```
alexey-milovidov, line 79: Yes, the code became bloated/worse. However, we can accept that.

alexey-milovidov (approved):

clickgapai:
Hi @Algunenano @alexey-milovidov — the changelog category for this PR might need a look.

**Current:** `Improvement`
**Suggested:** `Bug Fix (user-visible misbehavior in an official stable release)`

**Why:** This PR fixes `intExp2` returning 0/garbage instead of UINT64_MAX for inputs >= 64 when JIT is enabled, and wrong results for large-magnitude integers on both paths. Wrong results are user-visible misbehavior, so this is a Bug Fix, not an Improvement. The changelog entry text is accurate.

clickgapai:

**Affected code:** `IntExp2Impl::compile` in src/Functions/intExp2.cpp and FunctionUnaryArithmetic special-case list — core code present in all supported branches.

**Why this might apply:** This fixes silent wrong results from `intExp2` when `compile_expressions=1`. Since version 25.5, `compile_expressions` defaults to true (SettingsChangesHistory.cpp:547), so all supported branches (25.8 LTS, 26.2, 26.3 LTS, 26.4) have JIT ON by default and are affected. Users experience incorrect values: e.g., `intExp2(toInt32(64))` returns 0 instead of 18446744073709551615.

PR #105058, opened 2026-05-15
Fix a bug during the interval type conversion

### LINKED ISSUE #104986: `interval_conversions` vector has wrong values causing `CAST` from smaller to larger interval units to return incorrect results

_Retrospective finding from a historical scan of [PR #68057](https://github.com/ClickHouse/ClickHouse/pull/68057) (merged 2024-08-22). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`CAST(toIntervalSecond(60) AS IntervalMinute)` returns 0 instead of 1. `CAST(toIntervalHour(24) AS IntervalDay)` returns 0 instead of 1. Multiple forward conversion paths (smaller to larger units) return wrong values.

**Root cause:** getLeastSupertype.h:54: `static std::vector<int> interval_conversions = {1, 1000, 1000, 1000, 60, 60, 24, 7, 4, 3, 4};` - First element should be 1000 (ns→us), and index 3 should be 60 (s→min), not 1000.

**Why we believe this is a bug:** FunctionsConversion.h:2574-2577 uses `interval_conversions[from_position]` for forward conversions. The vector at getLeastSupertype.h:54 has `{1, 1000, 1000, 1000, 60, 60, 24, 7, 4, 3, 4}` but should be `{1000, 1000, 1000, 60, 60, 24, 7, 4, 3, 4}` (leading 1 is wrong, and values from index 3 onwards are shifted).

**Affected locations:**
- `src/DataTypes/getLeastSupertype.h:54` — interval_conversions vector definition with wrong values
- `src/Functions/FunctionsConversion.h:2574` — forward conversion loop using wrong conversion factor

**Impact:** Users who explicitly cast intervals from smaller to larger units get wrong results. For example, `CAST(toIntervalSecond(60) AS IntervalMinute)` returns 0 instead of 1. This affects any SQL code that uses explicit interval type conversions (not comparisons, which work correctly).

Yes — confirmed on current `master` (commit `801397d3c11`).

### How to reproduce

```sql
SELECT CAST(toIntervalSecond(60) AS IntervalMinute); -- Returns 0, should return 1
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/8da71198-0292-46ca-aefd-979770eeab82)

### Expected behavior

```
1
1
1
1
1000
60
24
7
```

```
1000
0
0
0
1000
60
24
7
```

### Additional context

**Open risks:**
- The approximate conversions (week to month = 4) may cause imprecise results for month/quarter/year conversions

**Suggested fix:** Change the vector to `{1000, 1000, 1000, 60, 60, 24, 7, 4, 3, 4}` (remove leading 1) or equivalently shift all values one position left. The correct mapping is: [0]=1000 (ns→us), [1]=1000 (us→ms), [2]=1000 (ms→s), [3]=60 (s→min), [4]=60 (min→h), [5]=24 (h→day), [6]=7 (day→week), [7]~4 (week→month), [8]=3 (month→quarter), [9]=4 (quarter→year).

Found during automated review of [PR #68057](https://github.com/ClickHouse/ClickHouse/pull/68057).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr68057_001`_

### Version info
- Resolved by: #105058
- Merged into: `26.7.1.63`

yariks5s:

Fixes bugs when using CAST from smaller to larger interval units. Closes https://github.com/ClickHouse/ClickHouse/issues/104986

### Version info
- Merged into: `26.7.1.63`


tests/queries/0_stateless/04240_interval_cast_forward_conversion.sql
```diff
@@ -0,0 +1,30 @@
+-- Forward CAST between IntervalKind units must use the right conversion factor
```
nihalzp: Adding `-- { echo }` at top would make reading the reference file easier.

nihalzp (approved):

alexey-milovidov:

tests/queries/0_stateless/04240_interval_cast_forward_conversion.sql
alexey-milovidov: Done — added `-- { echo }` at the top in 422e985b562, so the reference now shows each query next to its result.

tests/queries/0_stateless/04240_interval_cast_forward_conversion.sql
```diff
@@ -0,0 +1,30 @@
...
+-- Adjacent kinds: backward (larger -> smaller)
+SELECT CAST(toIntervalMicrosecond(1) AS IntervalNanosecond);
+SELECT CAST(toIntervalMillisecond(1) AS IntervalMicrosecond);
+SELECT CAST(toIntervalSecond(1) AS IntervalMillisecond);
+SELECT CAST(toIntervalMinute(1) AS IntervalSecond);
+SELECT CAST(toIntervalHour(1) AS IntervalMinute);
+SELECT CAST(toIntervalDay(1) AS IntervalHour);
+SELECT CAST(toIntervalWeek(1) AS IntervalDay);
+
+-- Multi-step conversions
```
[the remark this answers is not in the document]
alexey-milovidov: Added in 422e985b562. The test now covers the `Month`/`Quarter`/`Year` tail in both directions — adjacent `Week`↔`Month` (factor 4), `Month`↔`Quarter` (factor 3), `Quarter`↔`Year` (factor 4) — plus multi-step `Month`->`Year` and `Week`->`Year`. These exercise the part of `interval_conversions` with the differing factors (`4`, `3`, `4`). I verified against a local build that the new forward cases return wrong results before the fix and correct ones after (e.g. `CAST(toIntervalMonth(3) AS IntervalQuarter)` returned `0`, now returns `1`; `CAST(toIntervalWeek(48) AS IntervalYear)` returned `0`, now `1`).

clickgapai:
Hi — this looks like a backport candidate for `26.1`, `26.2`, `26.3`, `26.4`, `26.5`. I ran this PR's own regression test against each branch's binary and the bug **reproduces** there (the fix isn't present yet).

PR #104800, opened 2026-05-13
Reject `ATTACH TABLE` with storage clauses but no `ENGINE`

groeneai:
Fixes #104791. Reported by @clickgapai during automated review of merged PR #73120; @zlareb1 flagged it for fix.

## Motivation

`ATTACH TABLE name SETTINGS k = v;` (and the same with `ORDER BY`, `PARTITION BY`, etc.) without an explicit `ENGINE` would be parsed and then silently routed through the short-`ATTACH` path that reads the table definition from stored metadata, overwriting the user's input. Users saw no error and assumed their settings took effect.

Reproducer (from the issue):

```sql
SET default_table_engine = 'MergeTree';
CREATE TABLE t (id UInt32) ORDER BY id;
DETACH TABLE t;
ATTACH TABLE t SETTINGS max_part_loading_threads = 16;

SELECT engine_full FROM system.tables WHERE name = 't';
-- Expected: SETTINGS contains max_part_loading_threads = 16
-- Actual:   only the original index_granularity = 8192
```

## Root cause

`ParserCreateQuery.cpp:697` allows `SETTINGS` without `ENGINE` for tables in order to support `default_table_engine` on `CREATE`. For `ATTACH` the same syntax produced an `ASTStorage` with settings but no engine, which made the short-`ATTACH` detection at `InterpreterCreateQuery.cpp:1571` (`!create.storage->engine`) match and discard the user's clauses while reading the full stored definition. The other non-engine storage clauses (`PARTITION BY`, `PRIMARY KEY`, `ORDER BY`, `SAMPLE BY`, `TTL`, `UNIQUE KEY`) have the same shape and the same bug.

## Fix

Reject the ambiguous case before entering the short-`ATTACH` branch. When the user provided storage clauses without an `ENGINE` and no columns list, throw `BAD_ARGUMENTS` and point them at the short form (`ATTACH TABLE t;`) or `ALTER TABLE` for changing settings after attach.

The check fires before `ON CLUSTER` forwarding and before Replicated-database query enqueue, so distributed `ATTACH ... SETTINGS` is rejected at the initiator instead of silently dropping on each replica.

## Tests

Added `tests/queries/0_stateless/04237_attach_table_settings_without_engine_error.sql` covering:
- `ATTACH TABLE t SETTINGS k = v;` — `BAD_ARGUMENTS`
- `ATTACH TABLE t ORDER BY id;` — `BAD_ARGUMENTS`
[...]

PR #73120, opened 2024-12-11
Support SETTINGS with implicit ENGINE and mixing engine and query settings

Algunenano:
### Changelog category (leave one):
- Improvement

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Support SETTINGS with implicit ENGINE and mixing engine and query settings

You can now use engine settings even if the engine is not declared (`default_table_engine`)
```sql
CREATE TABLE example_table
(
    `id` UInt32,
    `data` String
)
ORDER BY id
SETTINGS max_part_loading_threads = 8
```

You can also mix table settings and query settings. The server will apply the server settings dynamically to the query and remove them from the table declaration. In case of ambiguity (the setting can be both table and query setting), it will be kept as a table setting:

```sql
CREATE TABLE example_table
(
    `id` UInt32,
    `data` JSON
)
ENGINE = MergeTree()
ORDER BY id
SETTINGS async_insert = 1, enable_json_type = 1
```

Here `enable_json_type` is applied to the query and the table is created with just `async_insert = 1`

Closes https://github.com/ClickHouse/ClickHouse/issues/73099
Closes https://github.com/ClickHouse/ClickHouse/issues/63154

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

> Information about CI checks: https://clickhouse.com/docs/en/development/continuous-integration/

#### CI Settings (Only check the boxes if you know what you are doing):
- [ ]  Allow: All Required Checks
- [ ]  Allow: Stateless tests
- [ ]  Allow: Stateful tests
- [ ]  Allow: Integration Tests
- [ ]  Allow: Performance tests
- [ ]  Allow: All Builds
- [ ]  Allow: batch 1, 2 for multi-batch jobs
- [ ]  Allow: batch 3, 4, 5, 6 for multi-batch jobs
---
- [ ]  Exclude: Style check
- [ ]  Exclude: Fast test
- [ ]  Exclude: All with ASAN
- [ ]  Exclude: All with TSAN, MSAN, UBSAN, Coverage
- [ ]  Exclude: All with aarch64, release, debug
---
- [ ]  Run only fuzzers related jobs (libFuzzer fuzzers, AST fuzzers, etc.)
- [ ]  Exclude: AST fuzzers
---
- [ ]  Do not test
- [ ]  Woolen Wolfdog
- [ ]  Upload binaries for special builds
- [ ]  Disable merge-commit
- [ ]  Disable CI cache

PR #105068, opened 2026-05-15
Reject ATTACH TABLE with non-engine storage clauses (narrower guard, re-land of #104800)

### LINKED ISSUE #104791: `ATTACH TABLE ... SETTINGS` silently ignores settings when ENGINE not specified

_Retrospective finding from a historical scan of [PR #73120](https://github.com/ClickHouse/ClickHouse/pull/73120) (merged 2024-12-26). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

User-specified SETTINGS in ATTACH TABLE statement without ENGINE are silently ignored. The table retains its original settings from stored metadata.

**Root cause:** InterpreterCreateQuery.cpp:1571 - condition `create.attach && (!create.storage || !create.storage->engine) && !create.columns_list` treats ATTACH with SETTINGS (no ENGINE) as short ATTACH syntax, reading stored definition and ignoring user-provided settings.

**Why we believe this is a bug:** ParserCreateQuery.cpp:697 now allows SETTINGS without ENGINE. For ATTACH, this creates storage with settings but no engine. InterpreterCreateQuery.cpp:1571 condition `(!create.storage || !create.storage->engine)` matches and reads stored definition, ignoring user's SETTINGS.

**Affected locations:**
- `src/Interpreters/InterpreterCreateQuery.cpp:1571` — Short ATTACH condition now matches SETTINGS-without-ENGINE
- `src/Parsers/ParserCreateQuery.cpp:697` — SETTINGS now allowed without ENGINE

**Impact:** Users specifying `ATTACH TABLE t SETTINGS max_part_loading_threads = 8;` expect settings to be applied but they are silently ignored. Both valid and invalid settings are ignored without error. Before this PR, this syntax would fail at parse time.

Yes — confirmed on current `master` (commit `7bd0fa28146`).

### How to reproduce

```sql
-- Test: ATTACH TABLE with SETTINGS silently ignores the settings (bug)
SET default_table_engine = 'MergeTree';
DROP TABLE IF EXISTS test_attach_settings_bug;
CREATE TABLE test_attach_settings_bug (id UInt32) ORDER BY id;
SELECT 'After CREATE - engine_full:', engine_full
FROM system.tables WHERE database = currentDatabase() AND name = 'test_attach_settings_bug';
DETACH TABLE test_attach_settings_bug;
ATTACH TABLE test_attach_settings_bug SETTINGS max_part_loading_threads = 16;
SELECT 'After ATTACH with max_part_loading_threads=16:', 
       position(engine_full, 'max_part_loading_threads = 16') > 0 AS has_setting
FROM system.tables WHERE database = currentDatabase() AND name = 'test_attach_settings_bug';
DROP TABLE IF EXISTS test_attach_settings_bug;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/5e92b38e-09cd-48d0-b891-ef4018d409b3)

### Expected behavior

```
After CREATE - engine_full:	MergeTree ORDER BY id SETTINGS index_granularity = 8192
After ATTACH with max_part_loading_threads=16:	1
```

```
After CREATE - engine_full:	MergeTree ORDER BY id SETTINGS index_granularity = 8192
After ATTACH with max_part_loading_threads=16:	0
```

### Additional context

**Open risks:**
- Distributed DDL with ATTACH TABLE SETTINGS may also silently ignore settings

**Suggested fix:** Either (1) error when SETTINGS specified in ATTACH without ENGINE, or (2) merge user's SETTINGS with stored definition, or (3) exclude ATTACH queries from allowing SETTINGS without ENGINE in parser

Found during automated review of [PR #73120](https://github.com/ClickHouse/ClickHouse/pull/73120).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr73120_001`_

### Version info
- Resolved by: #105068
- Merged into: `26.6.1.88`

groeneai:
Re-land of #104800 with a narrower guard that does not break legitimate session-settings `ATTACH`.

PR #104800 was merged then reverted in #105032 because it rejected `ATTACH TABLE t SETTINGS log_comment = 'foo';` — a legitimate pattern internal CI relies on (per @Algunenano in #104791). This re-land fixes the same silent-drop bug from #104791 but only rejects `ATTACH` queries that actually carry non-engine storage clauses that would be silently dropped.

### Background

`ParserCreateQuery.cpp` allows `SETTINGS` (and `ORDER BY`, `PARTITION BY`, etc.) without `ENGINE` for tables (#73120) so that `CREATE TABLE t (...) SETTINGS ...` works under `default_table_engine`. For `ATTACH` the same syntax used to be parsed without error and then silently dropped: the short-`ATTACH` path at `InterpreterCreateQuery.cpp:1571` reads the stored metadata and overwrites `create.storage` / `create.columns_list`, discarding whatever the user typed.

### Why the broad guard from #104800 broke CI

`InterpreterSetQuery::applySettingsFromQuery` runs in `executeQueryImpl` before `InterpreterCreateQuery::createTable`. For `ATTACH TABLE t SETTINGS log_comment = 'foo';` it does what it does for `CREATE`: it walks `create.storage->settings`, splits session settings (anything present in the per-query settings registry but not in the storage's builtin-setting list) from engine settings, applies the session settings to the context, and resets `storage->settings` to `nullptr` if nothing engine-specific is left. The user-visible result is that `log_comment` takes effect and the table is re-attached using stored metadata.

The reverted guard tripped on this case because it only checked that `create.storage` *existed* and had no `engine`. By the time the guard ran, the storage object's only meaningful field had already been hoisted out by `applySettingsFromQuery`, so rejecting was wrong.

### The narrower guard

Check the non-engine storage fields directly instead of just looking at the storage object. The guard now fires only when at least one of `partition_by`, `primary_key`, `order_by`, `sample_by`, `ttl_table`, `unique_key`, or `settings` is still populated at the point `createTable` runs — i.e. the user supplied a storage clause that survived the session-settings hoisting and would otherwise be silently dropped.

| Query                                                                       | Result                          | Reason                                                                                          |
|-----------------------------------------------------------------------------|---------------------------------|-------------------------------------------------------------------------------------------------|
| `ATTACH TABLE t;`                                                           | OK (short ATTACH)               | no storage clauses                                                                              |
| `ATTACH TABLE t SETTINGS log_comment = 'foo';`                              | OK (short ATTACH, session bind) | `applySettingsFromQuery` hoists `log_comment` to context and clears `storage->settings`         |
| `ATTACH TABLE t SETTINGS index_granularity = 100;`                          | `BAD_ARGUMENTS`                 | `index_granularity` is a `MergeTree` builtin → stays in `storage->settings`                     |
| `ATTACH TABLE t SETTINGS max_part_loading_threads = 16;`                    | `BAD_ARGUMENTS`                 | non-session setting → stays in `storage->settings`                                              |
| `ATTACH TABLE t SETTINGS log_comment = 'foo', index_granularity = 100;`     | `BAD_ARGUMENTS`                 | `log_comment` is hoisted, `index_granularity` remains in `storage->settings`                    |
| `ATTACH TABLE t ORDER BY id;` / `... PARTITION BY id;`                      | `BAD_ARGUMENTS`                 | `storage->order_by` / `storage->partition_by` set                                               |

The guard still fires before `database->shouldReplicateQuery(...)` and `executeQueryOnCluster(...)`, so distributed / `Replicated` `ATTACH` is rejected at the initiator instead of silently dropping on each replica.

### Tests

`04247_attach_table_storage_clauses_without_engine_error` covers all six rows of the table above.

### Reverted PR

- #104800 (merged, then reverted via #105032)
- Original issue: #104791

Closes #104791.

`ATTACH TABLE name <clauses>;` queries that supply storage clauses (`ORDER BY`, `PARTITION BY`, `PRIMARY KEY`, `SAMPLE BY`, `TTL`, `UNIQUE KEY`, or engine `SETTINGS`) without an `ENGINE` now throw `BAD_ARGUMENTS` instead of silently re-attaching the table with its stored definition and discarding the user-supplied clauses. Query-level session `SETTINGS` (such as `log_comment`) are still applied. Use `ATTACH TABLE t;` to re-attach with stored metadata, or `ALTER TABLE t MODIFY SETTING ...` after `ATTACH` to change settings.

### Version info
- Merged into: `26.6.1.88`

### DIFF
--- a/src/Interpreters/InterpreterCreateQuery.cpp
+++ b/src/Interpreters/InterpreterCreateQuery.cpp
@@ -1567,6 +1567,43 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
 
     DDLGuardPtr ddl_guard;
 
+    /// `ATTACH TABLE name <storage clauses>;` without `ENGINE` and without a columns list is ambiguous:
+    /// the parser allows `SETTINGS`, `ORDER BY`, `PARTITION BY`, etc. without an `ENGINE` (to support
+    /// `default_table_engine` on `CREATE`), but the short `ATTACH` path below reads the full table
+    /// definition from stored metadata and silently overwrites anything the user typed.
+    ///
+    /// However `ATTACH TABLE t SETTINGS log_comment = 'foo';` is a legitimate, supported pattern:
+    /// `InterpreterSetQuery::applySettingsFromQuery` (called from `executeQueryImpl` before this
+    /// interpreter runs) extracts session settings out of `create.storage->settings` and applies
+    /// them to the context, then clears `create.storage->settings` when nothing engine-specific
+    /// is left. So the case we must reject is the *remaining* one: at this point any of the
+    /// non-engine storage fields (`ORDER BY`, `PARTITION BY`, `PRIMARY KEY`, `SAMPLE BY`, `TTL`,
+    /// `UNIQUE KEY`, or `SETTINGS` containing keys that are not known session settings) are still
+    /// populated — those would be silently dropped by the short `ATTACH` path below.
+    if (create.attach && create.storage && !create.storage->engine && !create.columns_list)
+    {
+        const auto & storage = *create.storage;
+        const bool has_non_engine_storage_clauses
+            = storage.partition_by != nullptr
+            || storage.primary_key != nullptr
+            || storage.order_by != nullptr
+            || storage.sample_by != nullptr
+            || storage.ttl_table != nullptr
+            || storage.unique_key != nullptr
+            || storage.settings != nullptr;
+
+        if (has_non_engine_storage_clauses)
+        {
+            throw Exception(ErrorCodes::BAD_ARGUMENTS,
+                "ATTACH TABLE without ENGINE cannot specify storage clauses (ORDER BY, PARTITION BY, PRIMARY KEY, "
+                "SAMPLE BY, TTL, UNIQUE KEY, or engine SETTINGS) — they would be silently ignored because the "
+                "table definition is read from stored metadata. Use 'ATTACH TABLE {0};' to re-attach with stored "
+                "metadata, or 'ALTER TABLE {0} MODIFY SETTING ...' (or 'MODIFY ORDER BY ...') after ATTACH to "
+                "change settings. Session settings (e.g. log_comment) are allowed and applied automatically.",
+                backQuoteIfNeed(create.getTable()));
+        }
+    }
+
     // If this is a stub ATTACH query, read the query definition from the database
     if (create.attach && (!create.storage || !create.storage->engine) && !create.columns_list)
     {
--- a/src/Interpreters/InterpreterCreateQuery.cpp
+++ b/src/Interpreters/InterpreterCreateQuery.cpp
@@ -1605,7 +1605,8 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
             || create.is_clone_as
             || !create.as_database.empty()
             || !create.as_table.empty()
-            || create.uuid != UUIDHelpers::Nil;
+            || create.has_attach_from_path
+            || create.has_uuid_clause;
 
         if (has_dropped_clauses)
         {
--- a/src/Interpreters/InterpreterCreateQuery.cpp
+++ b/src/Interpreters/InterpreterCreateQuery.cpp
@@ -1606,7 +1606,8 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
             || !create.as_database.empty()
             || !create.as_table.empty()
             || create.has_attach_from_path
-            || create.has_uuid_clause;
+            || create.has_uuid_clause
+            || create.has_inner_uuid_clause;
 
         if (has_dropped_clauses)
         {
--- a/src/Interpreters/InterpreterCreateQuery.cpp
+++ b/src/Interpreters/InterpreterCreateQuery.cpp
@@ -1567,24 +1567,17 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
 
     DDLGuardPtr ddl_guard;
 
-    /// `ATTACH TABLE name <something>;` without `ENGINE` and without a columns list is ambiguous:
-    /// the parser allows `SETTINGS`, `ORDER BY`, `PARTITION BY`, `COMMENT`, `REFRESH`, `TO`, etc.
-    /// in this position (to support `default_table_engine` on `CREATE` and to share the parser
-    /// between `CREATE` and `ATTACH`), but the short `ATTACH` path below reads the full table
-    /// definition from stored metadata and silently overwrites anything the user typed.
-    ///
-    /// However `ATTACH TABLE t SETTINGS log_comment = 'foo';` is a legitimate, supported pattern:
-    /// `InterpreterSetQuery::applySettingsFromQuery` (called from `executeQueryImpl` before this
-    /// interpreter runs) extracts session settings out of `create.storage->settings` and applies
-    /// them to the context, then clears `create.storage->settings` when nothing engine-specific
-    /// is left. So the case we must reject is the remaining one: at this point any of the
-    /// non-engine storage fields (`ORDER BY`, `PARTITION BY`, `PRIMARY KEY`, `SAMPLE BY`, `TTL`,
-    /// `UNIQUE KEY`, or `SETTINGS` containing keys that are not known session settings) are still
-    /// populated, OR one of the top-level `ASTCreateQuery` fields (`COMMENT`, `REFRESH`,
-    /// `SQL SECURITY`, `TO target`, `EMPTY`, `CLONE`, `AS SELECT`, etc.) is set. Any of those
-    /// would be silently dropped by the short `ATTACH` path below.
-    if (create.attach && !create.columns_list && (!create.storage || !create.storage->engine))
+    // If this is a stub ATTACH query, read the query definition from the database
+    if (create.attach && (!create.storage || !create.storage->engine) && !create.columns_list)
     {
+        /// First, reject any user-supplied storage clauses or top-level fields that the
+        /// short-ATTACH path below would silently drop by overwriting `create` with the
+        /// stored table metadata. `InterpreterSetQuery::applySettingsFromQuery` (called from
+        /// `executeQueryImpl` before this interpreter) has already hoisted session settings
+        /// out of `create.storage->settings`, so anything still here is either engine-specific
+        /// (`ORDER BY`, `PARTITION BY`, `PRIMARY KEY`, `SAMPLE BY`, `TTL`, `UNIQUE KEY`, MergeTree
+        /// `SETTINGS`) or a top-level `ASTCreateQuery` field (`COMMENT`, `REFRESH`, `SQL SECURITY`,
+        /// `TO target`, `EMPTY`, `CLONE`, `AS SELECT`, `UUID`, etc.).
         bool has_dropped_clauses = false;
 
         if (create.storage)
@@ -1600,9 +1593,6 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
                 || storage.settings != nullptr;
         }
 
-        /// Top-level fields on `ASTCreateQuery` that the parser may populate for an `ATTACH`
-        /// query but that the short `ATTACH` path below silently drops by overwriting `create`
-        /// with the stored table metadata.
         has_dropped_clauses = has_dropped_clauses
             || create.comment != nullptr
             || create.refresh_strategy != nullptr
@@ -1614,7 +1604,8 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
             || create.is_create_empty
             || create.is_clone_as
             || !create.as_database.empty()
-            || !create.as_table.empty();
+            || !create.as_table.empty()
+            || create.uuid != UUIDHelpers::Nil;
 
         if (has_dropped_clauses)
         {
@@ -1624,11 +1615,7 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
                 "(or 'MODIFY ORDER BY ...') after ATTACH to change settings.",
                 backQuoteIfNeed(create.getTable()));
         }
-    }
 
-    // If this is a stub ATTACH query, read the query definition from the database
-    if (create.attach && (!create.storage || !create.storage->engine) && !create.columns_list)
-    {
         // In case of an ON CLUSTER query, the database may not be present on the initiator node
         auto database = DatabaseCatalog::instance().tryGetDatabase(database_name);
         if (database && database->shouldReplicateQuery(getContext(), query_ptr))
--- a/src/Interpreters/InterpreterCreateQuery.cpp
+++ b/src/Interpreters/InterpreterCreateQuery.cpp
@@ -1567,9 +1567,10 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
 
     DDLGuardPtr ddl_guard;
 
-    /// `ATTACH TABLE name <storage clauses>;` without `ENGINE` and without a columns list is ambiguous:
-    /// the parser allows `SETTINGS`, `ORDER BY`, `PARTITION BY`, etc. without an `ENGINE` (to support
-    /// `default_table_engine` on `CREATE`), but the short `ATTACH` path below reads the full table
+    /// `ATTACH TABLE name <something>;` without `ENGINE` and without a columns list is ambiguous:
+    /// the parser allows `SETTINGS`, `ORDER BY`, `PARTITION BY`, `COMMENT`, `REFRESH`, `TO`, etc.
+    /// in this position (to support `default_table_engine` on `CREATE` and to share the parser
+    /// between `CREATE` and `ATTACH`), but the short `ATTACH` path below reads the full table
     /// definition from stored metadata and silently overwrites anything the user typed.
     ///
     /// However `ATTACH TABLE t SETTINGS log_comment = 'foo';` is a legitimate, supported pattern:
@@ -1579,20 +1580,43 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
     /// is left. So the case we must reject is the remaining one: at this point any of the
     /// non-engine storage fields (`ORDER BY`, `PARTITION BY`, `PRIMARY KEY`, `SAMPLE BY`, `TTL`,
     /// `UNIQUE KEY`, or `SETTINGS` containing keys that are not known session settings) are still
-    /// populated, and those would be silently dropped by the short `ATTACH` path below.
-    if (create.attach && create.storage && !create.storage->engine && !create.columns_list)
+    /// populated, OR one of the top-level `ASTCreateQuery` fields (`COMMENT`, `REFRESH`,
+    /// `SQL SECURITY`, `TO target`, `EMPTY`, `CLONE`, `AS SELECT`, etc.) is set. Any of those
+    /// would be silently dropped by the short `ATTACH` path below.
+    if (create.attach && !create.columns_list && (!create.storage || !create.storage->engine))
     {
-        const auto & storage = *create.storage;
-        const bool has_non_engine_storage_clauses
-            = storage.partition_by != nullptr
-            || storage.primary_key != nullptr
-            || storage.order_by != nullptr
-            || storage.sample_by != nullptr
-            || storage.ttl_table != nullptr
-            || storage.unique_key != nullptr
-            || storage.settings != nullptr;
+        bool has_dropped_clauses = false;
 
-        if (has_non_engine_storage_clauses)
+        if (create.storage)
+        {
+            const auto & storage = *create.storage;
+            has_dropped_clauses
+                = storage.partition_by != nullptr
+                || storage.primary_key != nullptr
+                || storage.order_by != nullptr
+                || storage.sample_by != nullptr
+                || storage.ttl_table != nullptr
+                || storage.unique_key != nullptr
+                || storage.settings != nullptr;
+        }
+
+        /// Top-level fields on `ASTCreateQuery` that the parser may populate for an `ATTACH`
+        /// query but that the short `ATTACH` path below silently drops by overwriting `create`
+        /// with the stored table metadata.
+        has_dropped_clauses = has_dropped_clauses
+            || create.comment != nullptr
+            || create.refresh_strategy != nullptr
+            || create.sql_security != nullptr
+            || create.select != nullptr
+            || create.targets != nullptr
+            || create.as_table_function != nullptr
+            || create.aliases_list != nullptr
+            || create.is_create_empty
+            || create.is_clone_as
+            || !create.as_database.empty()
+            || !create.as_table.empty();
+
+        if (has_dropped_clauses)
         {
             throw Exception(ErrorCodes::BAD_ARGUMENTS,
                 "ATTACH applies the table definition from stored metadata and can't be changed in the query itself. "
--- a/src/Interpreters/InterpreterCreateQuery.cpp
+++ b/src/Interpreters/InterpreterCreateQuery.cpp
@@ -1576,10 +1576,10 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
     /// `InterpreterSetQuery::applySettingsFromQuery` (called from `executeQueryImpl` before this
     /// interpreter runs) extracts session settings out of `create.storage->settings` and applies
     /// them to the context, then clears `create.storage->settings` when nothing engine-specific
-    /// is left. So the case we must reject is the *remaining* one: at this point any of the
+    /// is left. So the case we must reject is the remaining one: at this point any of the
     /// non-engine storage fields (`ORDER BY`, `PARTITION BY`, `PRIMARY KEY`, `SAMPLE BY`, `TTL`,
     /// `UNIQUE KEY`, or `SETTINGS` containing keys that are not known session settings) are still
-    /// populated — those would be silently dropped by the short `ATTACH` path below.
+    /// populated, and those would be silently dropped by the short `ATTACH` path below.
     if (create.attach && create.storage && !create.storage->engine && !create.columns_list)
     {
         const auto & storage = *create.storage;
@@ -1595,11 +1595,9 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
         if (has_non_engine_storage_clauses)
         {
             throw Exception(ErrorCodes::BAD_ARGUMENTS,
-                "ATTACH TABLE without ENGINE cannot specify storage clauses (ORDER BY, PARTITION BY, PRIMARY KEY, "
-                "SAMPLE BY, TTL, UNIQUE KEY, or engine SETTINGS) — they would be silently ignored because the "
-                "table definition is read from stored metadata. Use 'ATTACH TABLE {0};' to re-attach with stored "
-                "metadata, or 'ALTER TABLE {0} MODIFY SETTING ...' (or 'MODIFY ORDER BY ...') after ATTACH to "
-                "change settings. Session settings (e.g. log_comment) are allowed and applied automatically.",
+                "ATTACH applies the table definition from stored metadata and can't be changed in the query itself. "
+                "Use 'ATTACH TABLE {0};' to re-attach with stored metadata, or 'ALTER TABLE {0} MODIFY SETTING ...' "
+                "(or 'MODIFY ORDER BY ...') after ATTACH to change settings.",
                 backQuoteIfNeed(create.getTable()));
         }
     }
--- a/src/Parsers/ParserCreateQuery.cpp
+++ b/src/Parsers/ParserCreateQuery.cpp
@@ -856,6 +856,7 @@ bool ParserCreateTableQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & expe
         query->table = table_id->getTable();
         query->uuid = table_id->uuid;
         query->has_uuid = table_id->uuid != UUIDHelpers::Nil;
+        query->has_uuid_clause = table_id->has_uuid;
         query->setIsTemporary(is_temporary);
 
         query->attach_as_replicated = attach_as_replicated;
@@ -1014,6 +1015,7 @@ bool ParserCreateTableQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & expe
     query->table = table_id->getTable();
     query->uuid = table_id->uuid;
     query->has_uuid = table_id->uuid != UUIDHelpers::Nil;
+    query->has_uuid_clause = table_id->has_uuid;
     query->cluster = cluster_str;
 
     if (query->database)
@@ -1467,6 +1469,7 @@ bool ParserCreateDatabaseQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & e
     if (!name_p.parse(pos, database, expected))
         return false;
 
+    bool has_uuid_clause = false;
     if (s_uuid.ignore(pos, expected))
     {
         ParserStringLiteral uuid_p;
@@ -1474,6 +1477,7 @@ bool ParserCreateDatabaseQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & e
         if (!uuid_p.parse(pos, ast_uuid, expected))
             return false;
         uuid = parseFromString<UUID>(ast_uuid->as<ASTLiteral>()->value.safeGet<String>());
+        has_uuid_clause = true;
     }
 
     if (s_on.ignore(pos, expected))
@@ -1496,6 +1500,7 @@ bool ParserCreateDatabaseQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & e
 
     query->uuid = uuid;
     query->has_uuid = uuid != UUIDHelpers::Nil;
+    query->has_uuid_clause = has_uuid_clause;
     query->cluster = cluster_str;
     query->database = database;
 
--- a/src/Parsers/ParserCreateQuery.cpp
+++ b/src/Parsers/ParserCreateQuery.cpp
@@ -857,6 +857,7 @@ bool ParserCreateTableQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & expe
         query->uuid = table_id->uuid;
         query->has_uuid = table_id->uuid != UUIDHelpers::Nil;
         query->has_uuid_clause = table_id->has_uuid;
+        query->has_inner_uuid_clause = to_inner_uuid != nullptr;
         query->setIsTemporary(is_temporary);
 
         query->attach_as_replicated = attach_as_replicated;
@@ -1016,6 +1017,7 @@ bool ParserCreateTableQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & expe
     query->uuid = table_id->uuid;
     query->has_uuid = table_id->uuid != UUIDHelpers::Nil;
     query->has_uuid_clause = table_id->has_uuid;
+    query->has_inner_uuid_clause = to_inner_uuid != nullptr;
     query->cluster = cluster_str;
 
     if (query->database)
--- a/src/Parsers/ParserCreateQuery.cpp
+++ b/src/Parsers/ParserCreateQuery.cpp
@@ -1272,6 +1272,8 @@ bool ParserCreateWindowViewQuery::parseImpl(Pos & pos, ASTPtr & node, Expected &
     query->database = table_id->getDatabase();
     query->table = table_id->getTable();
     query->uuid = table_id->uuid;
+    query->has_uuid = table_id->uuid != UUIDHelpers::Nil;
+    query->has_uuid_clause = table_id->has_uuid;
     query->cluster = cluster_str;
 
     if (query->database)
@@ -1760,6 +1762,8 @@ bool ParserCreateViewQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & expec
     query->database = table_id->getDatabase();
     query->table = table_id->getTable();
     query->uuid = table_id->uuid;
+    query->has_uuid = table_id->uuid != UUIDHelpers::Nil;
+    query->has_uuid_clause = table_id->has_uuid;
     query->cluster = cluster_str;
 
     if (query->database)
@@ -1993,6 +1997,8 @@ bool ParserCreateDictionaryQuery::parseImpl(IParser::Pos & pos, ASTPtr & node, E
     query->database = dict_id->getDatabase();
     query->table = dict_id->getTable();
     query->uuid = dict_id->uuid;
+    query->has_uuid = dict_id->uuid != UUIDHelpers::Nil;
+    query->has_uuid_clause = dict_id->has_uuid;
 
     if (query->database)
         query->children.push_back(query->database);
--- a/src/Parsers/ASTCreateQuery.h
+++ b/src/Parsers/ASTCreateQuery.h
@@ -136,6 +136,7 @@ public:
     bool replace_view : 1 = false;         /// CREATE OR REPLACE VIEW
     bool has_uuid : 1 = false;             /// CREATE TABLE x UUID '...' with a non-`Nil` value (see `has_uuid_clause` for clause-presence tracking)
     bool has_uuid_clause : 1 = false;      /// Parser saw an explicit `UUID '...'` clause, true even when the value is `Nil`
+    bool has_inner_uuid_clause : 1 = false; /// Parser saw an explicit `TO INNER UUID '...'` clause
     bool is_dictionary : 1 = false;        /// CREATE DICTIONARY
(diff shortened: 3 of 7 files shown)


groeneai:
## Pre-PR validation gate

**(a) Deterministic repro?** Yes. Regression test `04247_attach_table_storage_clauses_without_engine_error` deterministically exercises 6 cases on a debug build. Both directions verified locally:

- WITHOUT this fix (clean master after revert #105032): cases 1–5 in the test silently succeed (`engine_full` returns the original definition unchanged) — the original silent-drop bug.
- WITH this fix: cases 1–5 throw `BAD_ARGUMENTS`, case 6 (`SETTINGS log_comment = 'foo'`) and case 7 (`ATTACH TABLE t;`) succeed.

**(b) Root cause explained?** Yes. `InterpreterSetQuery::applySettingsFromQuery` runs in `executeQueryImpl` before the interpreter dispatch. For `ATTACH TABLE t SETTINGS log_comment = 'foo';` it hoists `log_comment` out of `create.storage->settings` (because the name is a per-query session setting but not a `MergeTree` builtin) into the context and resets `storage->settings` to `nullptr`. The broad guard from #104800 only checked `create.storage` existence and `!engine`, so by the time it ran the storage was already emptied of meaningful clauses — the guard rejected an `ATTACH` that the engine layer would have happily handled.

**(c) Fix matches root cause?** Yes. The narrower guard checks the individual non-engine storage fields (`partition_by`, `primary_key`, `order_by`, `sample_by`, `ttl_table`, `unique_key`, `settings`) so it only fires when the user supplied a clause that survived session-settings hoisting and would otherwise be silently dropped by the short-`ATTACH` path.

**(d) Test intent preserved?** Yes. The test still catches the original `#104791` silent-drop bug (cases 1, 2, 3, 4, 5) and additionally guards the legitimate `log_comment` shape (case 6) so the next regression won't re-break private CI.

**(e) Both directions demonstrated?** Yes. Built the branch locally with the fix; `tests/clickhouse-test 04247_attach_table_storage_clauses_without_engine_error` passes 1/1 (`0.63 sec`). Adjacent ATTACH tests `01601_detach_permanently`, `01516_create_table_primary_key`, `03167_attach_as_replicated_errors`, `04239_detach_table_on_temporary_table_103475` all pass.

**(f) Fix is general?** Yes. The guard rejects all six bug-class shapes (`SETTINGS` with engine setting, `SETTINGS` with non-session non-engine setting, `SETTINGS` mixing session + engine, `ORDER BY`, `PARTITION BY`, and by inheritance `PRIMARY KEY` / `SAMPLE BY` / `TTL` / `UNIQUE KEY`) and preserves the only legitimate shape (`SETTINGS` containing only session settings). It runs before `shouldReplicateQuery` / `executeQueryOnCluster` so distributed / `Replicated` `ATTACH` is rejected at the initiator.

src/Interpreters/InterpreterCreateQuery.cpp
```diff
@@ -1567,6 +1567,43 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
...
+            || storage.order_by != nullptr
+            || storage.sample_by != nullptr
+            || storage.ttl_table != nullptr
+            || storage.unique_key != nullptr
+            || storage.settings != nullptr;
+
+        if (has_non_engine_storage_clauses)
+        {
+            throw Exception(ErrorCodes::BAD_ARGUMENTS,
+                "ATTACH TABLE without ENGINE cannot specify storage clauses (ORDER BY, PARTITION BY, PRIMARY KEY, "
```
Algunenano: @groeneai Please simplify this error message. No need to say they would be ignored (or that session settings work), just say that ATTACH applies the table definition from stored metadata and can't be changed in the query itself. Keep the recommendation for the workaround, that's really good.

Also, please avoid using `—`. That's not a standard way of communicating in the project or in the error messages. Use plain and easy to understand English please. Make a note of this for the future.

groeneai: Thanks for the review @Algunenano. Pushed 6b4551c03635:

- Simplified the error message to the core point you suggested. Kept the workaround recommendation (`ATTACH TABLE {0};` and `ALTER TABLE {0} MODIFY SETTING ...`). Dropped the `would be ignored` and `session settings work` wording.
- Replaced the em-dash in the error string. Also removed the two em-dashes from the comment block above the guard and from the regression test header comment.
- Noted for future ClickHouse contributions: plain ASCII punctuation only in error messages, code comments, PR descriptions and replies. No em-dash.

Behavior is unchanged, the regression test only asserts `BAD_ARGUMENTS` so it still passes. New error text:

```
ATTACH applies the table definition from stored metadata and can't be changed in the query itself. Use 'ATTACH TABLE {0};' to re-attach with stored metadata, or 'ALTER TABLE {0} MODIFY SETTING ...' (or 'MODIFY ORDER BY ...') after ATTACH to change settings.
```

Algunenano (approved):

Algunenano:
@groeneai It seems that we only cover the fields under create.storage, but not the ones under create itself. For example, `ATTACH TABLE t COMMENT 'x'` would be accepted but the comment would be dropped. Could you please review if that's the case and what other fields we need to cover in this PR to avoid this issue, please?

groeneai:
Good catch @Algunenano, thanks. Pushed 60e17571931b that broadens the guard to cover the top-level `ASTCreateQuery` fields, not just the ones under `create.storage`.

### Audit of top-level fields the parser may populate for `ATTACH`

I went through `ParserCreateQuery::parseImpl`, `ParserCreateViewQuery::parseImpl`, and the `ASTCreateQuery` definition, then compared with the short-`ATTACH` path in `InterpreterCreateQuery::createTable` (which does `create = create_query` and only keeps `attach`, `attach_short_syntax`, `if_not_exists` from the user query). The fields that survive parsing but get silently dropped are:

| Field | User syntax that sets it | Behavior before this patch |
|---|---|---|
| `create.comment` | `ATTACH TABLE t COMMENT 'x'` | silently dropped |
| `create.refresh_strategy` | `ATTACH MATERIALIZED VIEW t REFRESH ...` | silently dropped |
| `create.sql_security` | `ATTACH VIEW t SQL SECURITY DEFINER ...` | silently dropped |
| `create.select` | `ATTACH MATERIALIZED VIEW t AS SELECT ...` | silently dropped |
| `create.targets` | `ATTACH MATERIALIZED VIEW t TO target ...`, `TimeSeries DATA/TAGS/METRICS`, `TO INNER UUID` | silently dropped |
| `create.as_table_function` | `ATTACH TABLE t AS func(...)` (gated, rejected at parse for plain table) | silently dropped if it slips through |
| `create.aliases_list` | view aliases `(a, b)` | silently dropped |
| `create.is_create_empty` | `ATTACH TABLE t EMPTY AS ...` | silently dropped |
| `create.is_clone_as` | `ATTACH TABLE t CLONE AS source` | silently dropped |
| `create.as_database`, `create.as_table` | `ATTACH TABLE t AS [db.]source` (gated, rejected at parse for plain table) | silently dropped if it slips through |

For plain `ATTACH TABLE t AS ...` the parser specifically only accepts `NOT REPLICATED` / `REPLICATED` after `AS`, so `as_table_function` / `as_database` / `as_table` would normally `SYNTAX_ERROR`. I check them anyway to stay safe under future parser changes.

The fields I deliberately do **not** touch because they are legitimate `ATTACH` flags:
- `create.attach_as_replicated` (`ATTACH TABLE t AS REPLICATED` / `AS NOT REPLICATED`)
- `create.attach_from_path` / `create.has_attach_from_path` (`ATTACH TABLE t FROM '/path'`)
- `create.if_not_exists`
- `create.is_temporary`
- `create.has_uuid`, `create.uuid`
- `create.cluster`

### Verified locally

| Query | Before this patch | After this patch |
|---|---|---|
| `ATTACH TABLE t COMMENT 'x';` | OK (comment dropped) | `BAD_ARGUMENTS` |
| `ATTACH TABLE t EMPTY AS SELECT 1;` | OK (`EMPTY` dropped) | `BAD_ARGUMENTS` |
| `ATTACH TABLE t CLONE AS source;` | OK (`CLONE` dropped) | `BAD_ARGUMENTS` |
| `ATTACH MATERIALIZED VIEW t REFRESH EVERY 1 YEAR TO target AS SELECT ...;` | OK (all dropped) | `BAD_ARGUMENTS` |
| `ATTACH MATERIALIZED VIEW t TO different_target AS SELECT ...;` | OK (`TO` dropped) | `BAD_ARGUMENTS` |
| `ATTACH TABLE t;` | OK | OK |
| `ATTACH TABLE t SETTINGS log_comment = 'foo';` | OK | OK |
| `ATTACH TABLE t AS NOT REPLICATED;` | OK | OK |

Regression test extended with the new cases (8 through 13). The full test passes 50/50 with full randomization. Adjacent `ATTACH`-related tests (`01601_detach_permanently`, `01516_create_table_primary_key`, `01153_attach_mv_uuid`, `01073_attach_if_not_exists`, `01188_attach_table_from_path`, `01015_attach_part`, `03167_attach_as_replicated_errors`, `04239_detach_table_on_temporary_table_103475`, `00180_attach_materialized_view`, `01107_atomic_db_detach_attach`) all pass.

groeneai:
## Pre-PR validation gate (TASK.md Phase 4 Step 9)

**(a) Deterministic repro?** Yes. `ATTACH TABLE t COMMENT 'x';`, `ATTACH TABLE t EMPTY AS SELECT 1;`, `ATTACH TABLE t CLONE AS t;`, and `ATTACH MATERIALIZED VIEW t REFRESH ... TO target AS SELECT ...` all reproduce on demand: succeed silently on the pre-patch build, throw `BAD_ARGUMENTS` on the patched build.

**(b) Root cause explained?** Yes. `InterpreterCreateQuery::createTable` short-`ATTACH` path does `create = create_query` (line 1658 in the post-patch file) which overwrites every field of `create` with the stored metadata, preserving only `attach`, `attach_short_syntax`, and `if_not_exists`. The previous guard only inspected `create.storage` sub-fields, so any top-level field on `ASTCreateQuery` that the parser populated (`comment`, `refresh_strategy`, `sql_security`, `select`, `targets`, etc.) survived parsing and was dropped without an error.

**(c) Fix matches root cause?** Yes. The guard now enumerates the top-level fields of `ASTCreateQuery` that the parser can populate for an `ATTACH` query and rejects the query when any of them is set under the same outer condition as the short-`ATTACH` path. The legitimate fields (`attach_as_replicated`, `attach_from_path`, `if_not_exists`, `is_temporary`, `has_uuid`, `cluster`) are intentionally not checked.

**(d) Test intent preserved?** Yes. Cases 1-7 of `04247_attach_table_storage_clauses_without_engine_error` still cover the original silent-drop bug for storage clauses (`SETTINGS`, `ORDER BY`, `PARTITION BY`, and the legitimate `log_comment` session-binding shape). Cases 8-13 extend the test to the new shapes (`COMMENT`, `EMPTY`, `CLONE`, `REFRESH`, `TO target`, `AS SELECT`).

**(e) Both directions demonstrated?** Yes. Verified on debug build: pre-patch the new cases run without error; post-patch they all throw `BAD_ARGUMENTS`. The legitimate `ATTACH TABLE t;`, `ATTACH TABLE t SETTINGS log_comment = 'foo';`, and `ATTACH TABLE t AS NOT REPLICATED;` shapes continue to succeed.

**(f) Fix is general?** Yes. Audited every parse-settable field on `ASTCreateQuery` against the short-`ATTACH` overwrite. The guard now covers `comment`, `refresh_strategy`, `sql_security`, `select`, `targets`, `as_table_function`, `aliases_list`, `is_create_empty`, `is_clone_as`, `as_database`, `as_table` in addition to the existing storage sub-fields. No symmetric code path was missed: the silent overwrite happens at a single site (line 1658).

Session: cron:clickhouse-ci-task-worker:20260519-173000

src/Interpreters/InterpreterCreateQuery.cpp
```diff
@@ -1567,6 +1567,65 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
...
+                || storage.ttl_table != nullptr
+                || storage.unique_key != nullptr
+                || storage.settings != nullptr;
+        }
+
+        /// Top-level fields on `ASTCreateQuery` that the parser may populate for an `ATTACH`
+        /// query but that the short `ATTACH` path below silently drops by overwriting `create`
+        /// with the stored table metadata.
+        has_dropped_clauses = has_dropped_clauses
+            || create.comment != nullptr
```
Algunenano, line 1597: create.uuid is missing

src/Interpreters/InterpreterCreateQuery.cpp
```diff
@@ -1567,6 +1567,65 @@ BlockIO InterpreterCreateQuery::createTable(ASTCreateQuery & create)
...
+    /// `InterpreterSetQuery::applySettingsFromQuery` (called from `executeQueryImpl` before this
+    /// interpreter runs) extracts session settings out of `create.storage->settings` and applies
+    /// them to the context, then clears `create.storage->settings` when nothing engine-specific
+    /// is left. So the case we must reject is the remaining one: at this point any of the
+    /// non-engine storage fields (`ORDER BY`, `PARTITION BY`, `PRIMARY KEY`, `SAMPLE BY`, `TTL`,
+    /// `UNIQUE KEY`, or `SETTINGS` containing keys that are not known session settings) are still
+    /// populated, OR one of the top-level `ASTCreateQuery` fields (`COMMENT`, `REFRESH`,
+    /// `SQL SECURITY`, `TO target`, `EMPTY`, `CLONE`, `AS SELECT`, etc.) is set. Any of those
+    /// would be silently dropped by the short `ATTACH` path below.
+    if (create.attach && !create.columns_list && (!create.storage || !create.storage->engine))
```
Algunenano: @groeneai  This if condition seems exactly the same as the next one

Algunenano (changes requested):
@groeneai Some minor issues

src/Interpreters/InterpreterCreateQuery.cpp
groeneai: Good catch. The new outer `if` had the same predicate as the existing stub-ATTACH block on line 1631 (just with the conditions reordered: `create.attach && !create.columns_list && (!create.storage || !create.storage->engine)` vs `create.attach && (!create.storage || !create.storage->engine) && !create.columns_list`). Pushed f72ea118852c to consolidate them: the `has_dropped_clauses` check now lives at the top of the existing block instead of in its own outer `if`. The duplicate predicate is gone.

src/Interpreters/InterpreterCreateQuery.cpp
groeneai: Added `create.uuid != UUIDHelpers::Nil` to the field check in f72ea118852c. Confirmed locally that `ATTACH TABLE t UUID '00000000-0000-0000-0000-000000000001';` now errors with `BAD_ARGUMENTS` (without the guard the stored UUID silently overwrites the user-supplied one when `create = create_query;` runs in the short-ATTACH path). Regression test extended with a UUID case (case 11).

groeneai:
@Algunenano addressed both minor issues in f72ea118852c:

1. **Consolidated the duplicate `if`**: the new outer `if (create.attach && !create.columns_list && (!create.storage || !create.storage->engine))` had the same predicate as the existing stub-ATTACH block immediately below it. Moved the `has_dropped_clauses` check inside the existing block instead of duplicating the condition.

2. **Added `create.uuid != UUIDHelpers::Nil`** to the field check. Without this, `ATTACH TABLE t UUID 'xxx';` (no other clauses) would silently drop the user-supplied UUID when the short-ATTACH path overwrites `create` with stored metadata. Regression test extended with case 11 covering this; existing MV cases renumbered to 12-14.

Verification:
- 30/30 passes with full randomization on the extended test
- Manual repro confirms case 11 errors with `BAD_ARGUMENTS`
- 4 unrelated ATTACH-adjacent tests (00180, 01073, 01153, 01188, 01015, 01254) still pass

Session: `cron:clickhouse-ci-task-worker:20260520-094500`

groeneai:
Pre-PR validation gate for f72ea118852c:

(a) **Deterministic repro?** Yes. `ATTACH TABLE t_uuid_test UUID '00000000-0000-0000-0000-000000000001';` on a detached table reproduces the silent UUID drop deterministically (without the new `|| create.uuid != UUIDHelpers::Nil` line).

(b) **Root cause explained?** The short-ATTACH path at line 1571 reads stored metadata via `database->getCreateTableQuery(...)` and then overwrites the entire `create` AST with `create = create_query;`. Any field the user supplied that the stored metadata also has (UUID, COMMENT, REFRESH, etc.) is silently dropped. Algunenano flagged that `uuid` was missing from the field list, so user-supplied UUIDs in short ATTACH were not being rejected.

(c) **Fix matches root cause?** Yes. Added `create.uuid != UUIDHelpers::Nil` to `has_dropped_clauses`. Same mechanism, same check style as the other top-level fields (`comment`, `refresh_strategy`, etc.). Also consolidated the duplicate `if` predicate that Algunenano flagged separately.

(d) **Test intent preserved? / New tests added?** Yes. New regression case 11 (`ATTACH TABLE t UUID '...';`) asserts `BAD_ARGUMENTS`. Existing cases 1-10 unchanged in behavior. Existing MV cases 11-13 renumbered to 12-14, content unchanged. Tested 30/30 with full randomization.

(e) **Demonstrated in both directions?** Yes. Without the `create.uuid != UUIDHelpers::Nil` line, `ATTACH TABLE t UUID 'xxx';` silently succeeds and drops the UUID (the short-ATTACH path `create = create_query;` overwrites it with the stored UUID). With the new line, it raises `BAD_ARGUMENTS`. Verified by manual repro before and after the edit.

(f) **Fix is general, not a narrow patch?** Same bug class as the rest of the guard (silent-drop in short ATTACH). The fix covers the full UUID surface in `ASTQueryWithTableAndOutput` (parent of `ASTCreateQuery`); no sibling paths exist for parser-supplied UUIDs. Internal callers do not construct short-ATTACH with explicit UUIDs (verified by grep over `src/Interpreters`, `src/Databases`, `src/Storages`).

Session: `cron:clickhouse-ci-task-worker:20260520-094500`

Algunenano (approved):

Algunenano:
@groeneai More corner cases:

```
ATTACH TABLE t FROM '/path'; is still silently lossy in this short-ATTACH path.

Concrete path in current code:

parser sets create.has_attach_from_path / create.attach_from_path;
this guard does not include that field in has_dropped_clauses, so the query falls through;
create = create_query then overwrites the parsed AST and drops the user-provided FROM path.
That executes as plain short ATTACH and ignores the FROM clause. Please reject this shape here (or preserve/validate attach_from_path) to avoid silently ignoring user input.
```

```
This create.uuid != UUIDHelpers::Nil check still misses explicit UUID '00000000-0000-0000-0000-000000000000'.

ParserCreateTableQuery stores UUID by value and treats Nil as "absent", so explicit Nil bypasses this guard; then short-ATTACH does create = create_query and silently drops the user-supplied UUID clause.

Please handle explicit Nil as user input as well (for example by tracking explicit UUID presence separately, or by rejecting Nil in parser for this ATTACH shape).

```

groeneai:
@Algunenano Pushed `1e7c952d4914` addressing both corner cases.

### Corner case 1: `ATTACH TABLE t FROM '/path';`

You were right, the FROM-only short attach path slips past the previous guard. The parser sets `create.has_attach_from_path = true` and `create.attach_from_path`, but neither was in the `has_dropped_clauses` predicate, so the query fell through to the short attach block where `create = create_query;` overwrote the user-supplied FROM.

Fix: added `create.has_attach_from_path` to the `has_dropped_clauses` field check. The legitimate long form `ATTACH TABLE t (cols) FROM '/path' ENGINE = ...;` carries a `columns_list`, which the outer guard predicate `!create.columns_list` already excludes, so that path stays unaffected.

### Corner case 2: explicit `Nil` UUID

You were also right that `create.uuid != UUIDHelpers::Nil` cannot distinguish "no UUID clause" from "explicit `UUID '00000000-0000-0000-0000-000000000000'`". `ParserCompoundIdentifier` reads `UUID '...'` directly into `table_id->uuid`, and `Nil == Nil` collapses the two cases.

Fix: tracked parser clause presence on the table identifier and propagated it through to the create query.

- `ASTTableIdentifier::has_uuid` (new): set by `ParserCompoundIdentifier` whenever the `UUID '...'` clause is parsed, regardless of value.
- `ASTCreateQuery::has_uuid_clause` (new): mirrored from `table_id->has_uuid` at the three sites where `query->uuid` is currently assigned for table/database create-attach queries.
- The existing `ASTCreateQuery::has_uuid` keeps its value-based semantic (`uuid != UUIDHelpers::Nil`), so the callers that use it as a non-`Nil`-UUID gate stay unchanged: `DatabaseReplicated.cpp` macro expansion, `TableZnodeInfo::resolve`, `ObjectStorageQueue/registerQueueStorage.cpp`, `TimeSeriesInnerTablesCreator.cpp`, `StorageMaterializedView.cpp`. None of those need to fire on an explicit Nil UUID.

The guard now reads:

```cpp
|| create.has_attach_from_path
|| create.has_uuid_clause;
```

instead of the old `create.uuid != UUIDHelpers::Nil`.

### Test extension

`tests/queries/0_stateless/04247_attach_table_storage_clauses_without_engine_error.sql`:

- Case 12: `ATTACH TABLE t_104791 UUID '00000000-0000-0000-0000-000000000000';` asserts `BAD_ARGUMENTS`.
- Case 13: `ATTACH TABLE t_104791 FROM '/var/lib/clickhouse/detached/t_104791';` asserts `BAD_ARGUMENTS`.
- MV cases renumbered 14-16 (content unchanged).

### Verification

- 30/30 passes locally with `--test-runs 30` on the extended test.
- Adjacent ATTACH tests still pass: `01188_attach_table_from_path` (legitimate long-FROM), `01153_attach_mv_uuid` (MV with non-`Nil` UUID), `01073_attach_if_not_exists`, `01015_attach_part`, `00180_attach_materialized_view`.
- Manual probes confirm both new corner cases now error with `BAD_ARGUMENTS`, and the existing short ATTACH plus `SETTINGS log_comment = '...'` short ATTACH paths still succeed.

No em-dashes in this commit, code, error messages, or this reply.

groeneai:
### Pre-PR validation gate for `1e7c952d4914`

Per the worker pipeline, answering each gate question explicitly:

**a) Deterministic repro?** Yes. Both new corner cases are deterministic. Without the fix, `ATTACH TABLE t UUID '00000000-0000-0000-0000-000000000000';` and `ATTACH TABLE t FROM '/some/path';` would silently take the short ATTACH path (after `create = create_query;`), drop the user input, and return success. With the fix, they raise `BAD_ARGUMENTS` deterministically every time.

**b) Root cause explained?** Yes. The previous guard tested `create.uuid != UUIDHelpers::Nil`, but `ParserCompoundIdentifier` stores the parsed UUID by value, so explicit `Nil` is indistinguishable from "no UUID clause". Separately, `create.has_attach_from_path` (set by `ParserCreateTableQuery` when `FROM '/path'` is parsed) was simply missing from the `has_dropped_clauses` field list, so the `FROM`-only short attach path fell through.

**c) Fix matches root cause?** Yes. The UUID fix records clause presence at the actual parser site (`ParserCompoundIdentifier::parseImpl` line where `s_uuid.ignore` returns true), stores it on `ASTTableIdentifier::has_uuid`, and propagates it through `ASTCreateQuery::has_uuid_clause`. This is the cleanest way to distinguish "user wrote `UUID '...'`" from "no UUID clause", which is what the guard actually needs. The FROM fix uses the already-existing `has_attach_from_path` flag.

**d) Test intent preserved?** Yes. The test still asserts `BAD_ARGUMENTS` for every silent-drop shape and still allows the two legitimate short ATTACH paths (with no storage clauses and with `SETTINGS log_comment = '...'`). The added cases extend coverage to two more silent-drop shapes without weakening any existing assertion.

  - Without the field check on `has_attach_from_path` and the `has_uuid_clause` flag, the new test cases would expect `BAD_ARGUMENTS` but the server would silently take the short ATTACH path. The test runner would report a mismatch and the test would fail.
  - With the fix in place, the test passes 30/30 with full randomization, and manual probes confirm both new cases reject with `BAD_ARGUMENTS`.

**f) Fix is general?** Yes.
  - "Same bug pattern elsewhere": I checked the other `query->uuid = table_id->uuid;` sites in `ParserCreateQuery.cpp` (CREATE WINDOW VIEW line 1272, CREATE LIVE VIEW line 1760, CREATE DICTIONARY line 1993). None of those set `query->has_uuid` either, which is a pre-existing inconsistency outside the scope of this PR. They are not in the ATTACH TABLE path that this guard governs, and the ATTACH MATERIALIZED VIEW path goes through `ParserCreateTableQuery` and is already covered.
  - "Root cause vs symptom": The `has_uuid_clause` flag is set where the UUID is actually parsed, not at the guard site. This addresses the upstream defect (value-based tracking that loses information for explicit `Nil`) rather than guarding against the symptom downstream.

Ready for re-review.

Algunenano:
@groeneai Please check the new AI review comments.

This PR is going to take 100 commits at this point :smile:

Algunenano:
@groeneai This PR is not ready. Please address https://github.com/ClickHouse/ClickHouse/pull/105068/files/1e7c952d4914a6d52167103ad4440cfa824edab7#diff-3c711c8ef85fea830dc4a29aa93b2a3ee1363a09a721d62ff9aa43a3fe0bda4b and https://github.com/ClickHouse/ClickHouse/pull/105068#discussion_r3273760571

groeneai:
Addressed the third corner case from the AI review at https://github.com/ClickHouse/ClickHouse/pull/105068#discussion_r3266770000 (`ATTACH TABLE t TO INNER UUID '...'`) in commit `001b7a4de473`.

### Root cause

`ParserCreateTableQuery::parseImpl` parses `TO INNER UUID` (line 810-815) into a local `to_inner_uuid` variable BEFORE the short-ATTACH early return (lines 845-870). The short-ATTACH branch did not persist `to_inner_uuid` into the query AST. Only the non-short branch persisted it via `view_targets->setInnerUUID` (line 1077, populating `create.targets`). So `ATTACH TABLE t TO INNER UUID '...';` silently dropped the clause because `create = create_query` in the short-ATTACH path overwrote everything with stored metadata.

### Fix

Following the same pattern used for `has_uuid_clause` (#104800) and `has_attach_from_path`:

1. `src/Parsers/ASTCreateQuery.h`: added `has_inner_uuid_clause : 1` bit-packed field.
2. `src/Parsers/ParserCreateQuery.cpp`: set `query->has_inner_uuid_clause = to_inner_uuid != nullptr` in BOTH the short-ATTACH branch (line 860) and the long branch (line 1020), so the flag has consistent semantics regardless of which path the parser took.
3. `src/Interpreters/InterpreterCreateQuery.cpp`: added `|| create.has_inner_uuid_clause` to the `has_dropped_clauses` predicate in the guard.

The non-short branch is already covered by the existing `create.targets != nullptr` check in the guard, so the flag is only strictly needed for the short-ATTACH case. Setting it in both branches keeps the flag's meaning unambiguous.

### Test extension

`tests/queries/0_stateless/04247_attach_table_storage_clauses_without_engine_error.sql`:

- Case 14 (new): `ATTACH TABLE t_104791 TO INNER UUID '00000000-0000-0000-0000-000000000001';` asserts `BAD_ARGUMENTS`.
- Case 15 (new): `ATTACH TABLE t_104791 TO INNER UUID '00000000-0000-0000-0000-000000000000';` (explicit Nil) asserts `BAD_ARGUMENTS`.
- MV cases 14-16 renumbered to 16-18 (content unchanged).

### Verification

- 100/100 passes locally on debug build with full randomization (`--test-runs 100`)
- Manual repros confirm both cases 14 and 15 error with `BAD_ARGUMENTS`
- Adjacent ATTACH tests unaffected: 01188_attach_table_from_path, 01153_attach_mv_uuid, 01073_attach_if_not_exists, 01015_attach_part, 00180_attach_materialized_view, 03167_attach_as_replicated_errors all pass

### Separate finding (not in this PR)

While exercising `TO INNER UUID` paths I noticed a pre-existing crash in the same function: `CREATE TABLE t TO INNER UUID '...' (id Int) ORDER BY id;` (no `ENGINE` specified, so `storage` is null in the long branch) crashes at `ParserCreateQuery.cpp:1073` because the throw `"Storage engine {} does not inner UUID"` dereferences `storage->engine->name` even when `storage` is null. This is independent of this PR's short-ATTACH fix and exists on `master`. I'll file a separate small PR for it.

groeneai:
### Pre-PR validation gate (for commit `001b7a4de473`)

**a) Deterministic repro?** YES. `ATTACH TABLE t_inner_uuid TO INNER UUID '00000000-0000-0000-0000-000000000001';` on a detached table reliably and deterministically reproduces the silent-drop on `1e7c952d4914` (parent commit). The clause is consumed by the parser and the table re-attaches with stored metadata, no error.

**b) Root cause explained?** YES. `ParserCreateTableQuery::parseImpl` parses `TO INNER UUID` into a local variable at line 810-815, before the short-ATTACH early return at line 845-870. The short-ATTACH branch creates the `ASTCreateQuery` and returns WITHOUT persisting `to_inner_uuid`. Only the long branch persists it (line 1077) into `create.targets`, which the existing guard already covers via `create.targets != nullptr`. The short-ATTACH path was missing equivalent tracking.

**c) Fix matches root cause?** YES. Added a parser-level flag `has_inner_uuid_clause` that records the parser-saw-clause fact independent of where the value ends up. Set in both parser branches, checked in the guard. This is the same pattern `#104800` introduced for `has_uuid_clause` and the same pattern I introduced for `has_attach_from_path` in the previous fix-up — three sibling silent-drop shapes, all handled the same way.

**d) Test intent preserved? / New tests added?** YES. Test cases 14 and 15 added for the two short-ATTACH `TO INNER UUID` shapes (non-Nil and Nil). MV cases 14-16 renumbered to 16-18 (content unchanged). The legitimate `CREATE TABLE ... ENGINE = SharedSet/SharedJoin TO INNER UUID '...'` long-path remains untouched and is exercised by the existing storage check at `ParserCreateQuery.cpp:1070`. `01188_attach_table_from_path`, `01153_attach_mv_uuid`, `01073_attach_if_not_exists`, `01015_attach_part`, `00180_attach_materialized_view`, and `03167_attach_as_replicated_errors` all pass unchanged.

**e) Demonstrated in both directions?** YES.
- Without fix (commit `1e7c952d4914`, parent): `ATTACH TABLE t_inner_uuid TO INNER UUID '...'` returns silently (table re-attached with stored metadata, user clause discarded).
- With fix (commit `001b7a4de473`): same query throws `BAD_ARGUMENTS` with the same guard message as the other silent-drop shapes.

**f) Fix is general, not a narrow patch?** YES, this is a parser-level fix. The flag is set on every code path where `to_inner_uuid` is parsed (both short-ATTACH and long branches), so any future code path that copies the AST or processes the query sees the flag consistently. The guard catches the shape regardless of which parser branch ran. No defensive null check; the root cause (the parser was not persisting the parsed clause) is fixed at the source.

src/Parsers/ParserCreateQuery.cpp
```diff
@@ -1496,6 +1500,7 @@ bool ParserCreateDatabaseQuery::parseImpl(Pos & pos, ASTPtr & node, Expected & e
 
     query->uuid = uuid;
     query->has_uuid = uuid != UUIDHelpers::Nil;
+    query->has_uuid_clause = has_uuid_clause;
```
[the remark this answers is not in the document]
groeneai: Good catch. Pushed `a7ce648eba46` ("Propagate has_uuid_clause in dictionary and view parsers") which addresses the propagation gap.

### What was missing

`ParserCreateDictionaryQuery::parseImpl` at line 1995 copies only `dict_id->uuid` to `query->uuid` but did not propagate `dict_id->has_uuid` to `query->has_uuid_clause`. Since `1e7c952d4914` switched the guard from `create.uuid != UUIDHelpers::Nil` to `create.has_uuid_clause`, the dictionary path bypassed the guard. `ATTACH DICTIONARY d UUID '...'` (with any value) silently overwrote the user-supplied UUID via `create = create_query`.

### Fix

Propagated `dict_id->has_uuid` to `query->has_uuid_clause` (and `dict_id->uuid != Nil` to `query->has_uuid` for consistency) in three sites that had the same gap:

1. `ParserCreateDictionaryQuery::parseImpl` (line 1995) — the one you flagged. Definitely broken, definitely exploitable: `ATTACH DICTIONARY d UUID '11111111-...'` silently dropped before this commit.
2. `ParserCreateViewQuery::parseImpl` (line 1762) — same propagation gap.
3. `ParserCreateWindowViewQuery::parseImpl` (line 1274) — same propagation gap.

The view + window view sites are defensive: I verified that `ATTACH MATERIALIZED VIEW v UUID '...'` alone is not currently a valid syntax (SYNTAX_ERROR from the parser, expecting `ON`/`REFRESH`/`TO`/columns/`ENGINE`/etc.), so the short-ATTACH UUID-only shape doesn't reach the guard for views today. But matching the pattern used by the two table parser branches (lines 858, 1018) keeps these consistent with your "and other `table_id`-based parser branches" phrasing and protects against future parser changes that might admit the shape.

### Regression test

Cases 16-17 added to `04247_attach_table_storage_clauses_without_engine_error.sql`:

```sql
CREATE DICTIONARY d_104791 (id UInt32, value String) PRIMARY KEY id
    SOURCE(NULL()) LAYOUT(FLAT()) LIFETIME(0);
DETACH DICTIONARY d_104791;
ATTACH DICTIONARY d_104791 UUID '00000000-0000-0000-0000-000000000001'; -- { serverError BAD_ARGUMENTS }
ATTACH DICTIONARY d_104791 UUID '00000000-0000-0000-0000-000000000000'; -- { serverError BAD_ARGUMENTS }
ATTACH DICTIONARY d_104791;
DROP DICTIONARY d_104791;
```

Verified the bug reproduces against commit `001b7a4de473` (parent): case 16 silently succeeds (rc=0), then case 17 fails with `DICTIONARY_ALREADY_EXISTS` because the previous silent re-attach already left the dictionary attached.

With this commit: cases 16 and 17 both error with `BAD_ARGUMENTS`. 50/50 passes with full randomization on debug build. Adjacent ATTACH tests still pass: `01153_attach_mv_uuid`, `01188_attach_table_from_path`, `01073_attach_if_not_exists`, `01015_attach_part`, `00180_attach_materialized_view`.

groeneai:
### Pre-PR validation gate (for commit `a7ce648eba46`)

**a) Deterministic repro?** YES. `ATTACH DICTIONARY d UUID '00000000-0000-0000-0000-000000000001';` on a detached dictionary silently succeeds (rc=0) on commit `001b7a4de473` (parent) and errors with `BAD_ARGUMENTS` on this commit. No randomization or timing needed.

**b) Root cause explained?** YES. `ParserCreateDictionaryQuery::parseImpl` line 1995 copies only `dict_id->uuid` to `query->uuid`. It does not copy `dict_id->has_uuid` to `query->has_uuid_clause`. Since `1e7c952d4914` switched the narrower guard in `InterpreterCreateQuery::createTable` from `create.uuid != UUIDHelpers::Nil` to `create.has_uuid_clause`, the dictionary path bypassed the guard. The guard then fell through to `create = create_query` which overwrote the parsed AST with stored metadata, silently dropping the user-supplied UUID. `ParserCreateViewQuery` line 1762 and `ParserCreateWindowViewQuery` line 1274 had the same propagation gap.

**c) Fix matches root cause?** YES. Propagated `dict_id->has_uuid` to `query->has_uuid_clause` in all three `table_id`-based parser branches (dictionary, view, window view), plus `dict_id->uuid != Nil` to `query->has_uuid` for symmetry with the two table parser branches at lines 858 and 1018 that already do both propagations. Two-line change per site, no behavior change beyond closing the propagation gap.

**d) Test intent preserved?** YES. The guard still raises `BAD_ARGUMENTS` for the silent-drop shapes. Cases 1-15 unchanged. New cases 16-17 cover the dictionary UUID propagation. Adjacent ATTACH tests pass: `01153_attach_mv_uuid` (long ATTACH MV with non-`Nil` UUID + columns + ENGINE + AS SELECT, still works), `01188_attach_table_from_path`, `01073_attach_if_not_exists`, `01015_attach_part`, `00180_attach_materialized_view`.

**e) Both directions demonstrated?** YES.
- Without fix (parent commit `001b7a4de473`): case 16 silently succeeds (rc=0), then case 17 fails with `DICTIONARY_ALREADY_EXISTS` because the previous silent re-attach already left the dictionary attached. Manual repro confirms the user-supplied non-`Nil` UUID is silently dropped.
- With fix (this commit): cases 16 and 17 both error with `BAD_ARGUMENTS`. 50/50 passes on debug build with full randomization.

**f) Fix general, not narrow patch?** YES. The bot's note specifically mentioned dictionary "and other `table_id`-based parser branches as well", so I audited all sites that set `query->uuid = X->uuid` in `ParserCreateQuery.cpp`. Six total: lines 857, 1017, 1274, 1503, 1762, 1995. Lines 857 and 1017 already had the `has_uuid_clause` propagation. Line 1503 (`ParserCreateDatabaseQuery`) tracks `has_uuid_clause` locally and already sets it on the query (line 1505). The remaining three (1274 window view, 1762 view, 1995 dictionary) all got the propagation in this commit.

For the view + window view cases, I also verified that `ATTACH MATERIALIZED VIEW v UUID '...'` alone is not currently a valid syntax (parser raises `SYNTAX_ERROR`, expecting `ON`/`REFRESH`/`TO`/columns/`ENGINE`/etc.), so the short-ATTACH UUID-only shape doesn't actually reach the guard for views. The propagation there is defensive consistency rather than catching a currently exploitable bug. The dictionary case is the only exploitable one in this commit.

PR #105088, opened 2026-05-15
Fix ClickBench job: allow `/dev/shm/clickhouse/` as filesystem cache path

alexey-milovidov:
The ClickBench job fails on master while attaching the `hits` table:

```
Code: 36. DB::Exception: Filesystem cache absolute path must lie inside `/tmp/filesystem_caches/`,
but have /dev/shm/clickhouse/: While processing disk(type = cache, path = '/dev/shm/clickhouse/', ...).
(BAD_ARGUMENTS)
```

`ci/jobs/scripts/clickbench/create.sql` mounts a cached disk at `/dev/shm/clickhouse/`, but the default
server config (`programs/server/config.d/filesystem_caches_path.xml`) pins both `filesystem_caches_path`
and `custom_cached_disks_base_directory` to `/tmp/filesystem_caches/`, and `RegisterDiskCache.cpp`
rejects any cache disk outside that root.

Override the two settings in `install_clickbench_config` so the allowed directory matches the path the
ClickBench `create.sql` actually uses.

Report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=75fa606f4590bda023a727a000555ca8622b4266&name_0=MasterCI&name_1=ClickBench%20%28arm_release%29

### Version info
- Merged into: `26.5.1.699`

### DIFF
--- a/ci/jobs/clickbench.py
+++ b/ci/jobs/clickbench.py
@@ -18,6 +18,13 @@ def main():
 
         def install():
             res = ch.install_clickbench_config()
+            # The ClickBench `create.sql` attaches `hits` on a cached disk
+            # rooted at `/dev/shm/clickhouse/`; ship the matching server-side
+            # allowed-directory override alongside it.
+            res = res and Shell.check(
+                f"cp ./ci/jobs/scripts/clickbench/filesystem_caches_path.xml {temp_dir}/config.d/",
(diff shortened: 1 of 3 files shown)


alexey-milovidov (commented):
Ok.

PR #105089, opened 2026-05-15
Parallelize copyFile in DiskObjectStorageTransaction

asya-ch:
- Performance Improvement

Parallelize blob copies in DiskObjectStorageTransaction::copyFile using a per-disk thread pool.

### Version info
- Merged into: `26.6.1.304`


src/Disks/DiskObjectStorage/DiskObjectStorage.h
```diff
@@ -259,6 +259,9 @@ friend class DiskObjectStorageReservation;
     BlobKillerThreadPtr blob_killer;
     BlobCopierThreadPtr blob_copier;
 
+    /// Thread pool used to parallelize `copyObjectToAnotherObjectStorage` calls.
+    std::shared_ptr<ThreadPool> copy_object_pool;
```
Michicosun, line 264: wdyt about reusing `IDIsk::copying_thread_pool` here?

src/Disks/DiskObjectStorage/DiskObjectStorageTransaction.cpp
```diff
@@ -514,6 +510,18 @@ void MultipleDisksObjectStorageTransaction::copyFile(const std::string & from_fi
...
 
+void DiskObjectStorageTransaction::copyFile(const std::string & from_file_path, const std::string & to_file_path, const ReadSettings & read_settings, const WriteSettings & write_settings)
+{
+    copyFileImpl(metadata_storage, cluster, object_storages, from_file_path, to_file_path, read_settings, write_settings);
+}
+
+void MultipleDisksObjectStorageTransaction::copyFile(const std::string & from_file_path, const std::string & to_file_path, const ReadSettings & read_settings, const WriteSettings & write_settings)
+{
+    const auto enriched_read_settings = updateIOSchedulingSettings(read_settings, read_resource_name, write_resource_name);
+    const auto enriched_write_settings = updateIOSchedulingSettings(write_settings, read_resource_name, write_resource_name);
```
Michicosun: Let's also update the `copyFile` settings with workload semantics.

alexey-milovidov:

src/Disks/DiskObjectStorage/DiskObjectStorage.h
asya-ch: I think there is a risk of deadlock. Copying_thread_pool is used by copyThroughBuffers(), which calls asyncCopy(), which calls copyFile().

src/Disks/DiskObjectStorage/DiskObjectStorage.cpp
```diff
@@ -274,8 +299,17 @@ void DiskObjectStorage::copyFile( /// NOLINT
...
+        {
+            transaction->copyFile(from_file_path, to_file_path, read_settings, write_settings);
+            transaction->commit();
+        }
+        catch (...)
+        {
+            // Roll back any destination blobs that were copied before the exception was thrown.
+            transaction->undo();
+            throw;
+        }
```
Michicosun, line 315: Why? Commit will remove all blobs on any exception

src/Disks/DiskObjectStorage/DiskObjectStorageTransaction.cpp
```diff
@@ -453,24 +458,63 @@ void DiskObjectStorageTransaction::createFile(const std::string & path)
...
+            written_blobs[location].push_back(dst_blob);
+
+    /// Dispatch `copyObjectToAnotherObjectStorage` calls in parallel onto the disk-level pool.
+    /// We can't reuse `IObjectStorage::getThreadPoolWriter()` here because the copy implementation
+    /// itself submits onto that pool via `writeObject`, which would risk pool self-deadlock.
+    auto scheduler = threadPoolCallbackRunnerUnsafe<void>(*copy_object_pool, ThreadName::DISK_OBJECT_STORAGE_COPY);
+    /// Bound in-flight tasks so memory stays O(threads) instead of O(locations * blobs),
+    /// and so a worker exception is observed (via future::get) before we enqueue the rest.
+    const size_t max_tasks_inflight = 2 * copy_object_pool->getMaxThreads();
+    TaskTracker task_tracker(std::move(scheduler), max_tasks_inflight, /*limited_log=*/nullptr);
```
Michicosun: Why do we need this? It will not be any time more than O(threads) because the pool is limited by the number of threads.

Michicosun: Why do we even need `TaskTracker`? There is a way to wait until all tasks scheduled via threadPoolCallbackRunner will be completed - `ThreadPoolCallbackRunnerLocal<bool>::waitForAllToFinish`.

src/Disks/DiskObjectStorage/DiskObjectStorage.cpp
asya-ch: Yes, commit() removes blobs. The undo() is in case it fails during copyFile() and never reaches commit(). 

I'll add a comment.

src/Disks/DiskObjectStorage/DiskObjectStorageTransaction.cpp
asya-ch: Claude suggested it -- the idea is that if it is more than threads, when there is a new thread available, it doesn't need to stall while the task is prepared.

But it's not necessary here. Will remove for simplicity.

asya-ch: TaskTracker is nice in some way -- it allows us to not have to copy all variables into the lambda, and the logic for rethrowing exceptions is simple. But probably not worth using ThreadPoolCallbackRunnerUnsafe -- will switch to ThreadPoolCallbackRunnerLocal.

src/Disks/DiskObjectStorage/DiskObjectStorage.cpp
Michicosun: Let's instead undo the transaction in the destructor?

asya-ch: I assume you mean the destructor of DiskObjectStorageTransaction -- is that correct?

In theory, that would be cleaner, but am not super familiar with DiskObjectStorageTransaction and whether it would be safe to always undo. Could look into it, but feels a bit out of scope -- is it okay to leave it for now? Or what is the concern with the current approach?

Michicosun: It should be okay, I probably just forgot to add this rollback. But let's delay this change for the next pr.

Michicosun (approved):

alesapin:
Why it shows as it was merged by Asya? It was merged by me :D

PR #105091, opened 2026-05-15
Fix inconsistent AST formatting for aliased IN inside a function call (STID 1941-1bfa)

tuanpach:
## Motivation

The format-parse-format consistency check in `DB::executeQueryImpl` aborts
debug / sanitiser builds (`Logical error: 'Inconsistent AST formatting
between 'ExpressionList' and 'ExpressionList'`, STID 1941-1bfa) on queries
like:

```sql
SELECT g(1, 2 IN ((3 IN (4, 5)) AS x))
```

The first format emits `g(1, (2 IN ((3 IN (4, 5)) AS x)))`; the re-parse
sets `parenthesized=true` on the outer `IN`, and the second format drops
the inner `(3 IN (4, 5))`, breaking round-trip stability.

## Root cause

Two code paths emit the wrapping `(...)` around the outer `IN`, but they
leave `frame.current_function` in different states for descendants:

| Pass | Outer wrap emitted by | `current_function` after | Inner IN wraps itself? |
|---|---|---|---|
| 1st format | `need_parens_around_in` in `ASTFunction::formatImplWithoutAlias` | unchanged (= `g`) | yes → `((3 IN (4, 5)) AS x)` |
| 2nd format | `decideParensEmission` in `IAST::format` (because re-parse set `parenthesized=true`) | reset to `nullptr` | no → `(3 IN (4, 5) AS x)` |

The asymmetric reset of `current_function` is what makes the inner `IN`
emit its own wrapping parens in one pass but not the other.

## Change

Make the two paths symmetric: when `formatImplWithoutAlias` emits the
wrap via `need_parens_around_in`, also clear `current_function` for
descendants — exactly what `decideParensEmission` already does on the
other path. Both passes then produce
`SELECT g(1, (2 IN (3 IN (4, 5) AS x)))` and the round-trip is stable.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=103126&sha=9ebf9fa83a0839baba782baf9689bc849215dc5a&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%2C%20targeted%2C%20old_compatibility%29
Found while CI-validating PR #103126.

Note: this is one subvariant of the chronic STID 1941-1bfa. Other
subvariants (`CODEC` / `STATISTICS` — open PR #104991, lambda-after-comma
— merged PR #104626, etc.) are independent.

Not for changelog.

- [x] Documentation is unchanged (formatter-internal round-trip stability; no user-visible API or syntax change).

### Version info
- Merged into: `26.6.1.283`


alexey-milovidov:

alexey-milovidov (approved):

Issue #11372, opened 2020-06-02
LimitTransform::makeChunkWithPreviousRow(): Assertion `row < chunk.getNumRows()' failed.


akuzm:
```
SELECT number FROM numbers(100) ORDER BY number ASC LIMIT 2, 9223372036854775807  WITH TIES
```

@KochetovNicolai The query is similar to https://github.com/ClickHouse/ClickHouse/issues/11359, but the error is different.

KochetovNicolai:
This case is integer overflow, because `SELECT number FROM numbers(100) ORDER BY number ASC LIMIT 2, 100000000  WITH TIES` works fine.
Don't know how to fix it properly.

akuzm:
> This case is integer overflow, because `SELECT number FROM numbers(100) ORDER BY number ASC LIMIT 2, 100000000 WITH TIES` works fine.
> Don't know how to fix it properly.

Is there some place where signed int is used for row number? Maybe if we convert it all to UInt64, it will work? And if we have some other kind of literal for LIMIT, we should report an error (we probably do so already).

KochetovNicolai:
> Is there some place where signed int is used for row number?

Yes. It is not so easy for fix.
https://github.com/ClickHouse/ClickHouse/blob/06446b4f08a142d6f1bc30664c47ded88ab51782/src/Processors/LimitTransform.cpp#L283-L286

akuzm:
> > Is there some place where signed int is used for row number?
> 
> Yes. It is not so easy for fix.
> https://github.com/ClickHouse/ClickHouse/blob/06446b4f08a142d6f1bc30664c47ded88ab51782/src/Processors/LimitTransform.cpp#L283-L286

It should be possible to calculate absolute row numbers for chunk start and chunk end, and then compare these numbers to offset and limit. They all fit UInt64, so it must work. It also will be easier to read -- now I'm not even sure the calculations are correct (they must be correct, but it's not immediately clear from the code). In cases like these, when there is a lot of conditions, it can be helpful to make a decision table and then just write it down as code: https://www.hillelwayne.com/post/decision-tables/
It's easy to miss something when the conditions are not explicit and embedded into a chain of min's.

alexey-milovidov:
@KochetovNicolai Signed integer overflow will be also reported by UBSan. We should fix it.

alexey-milovidov:
Duplicate of #10470

PR #101158, opened 2026-03-30
Fix DROP TABLE hang on Kafka tables due to unbounded rd_kafka_assign wait

alexey-milovidov:
Fix `DROP TABLE` hang on Kafka tables when a rebalance is in progress and the broker is temporarily unreachable.

The root cause is inside librdkafka: `rd_kafka_assign`, called from the rebalance callback during `poll_batch`, sends an `RD_KAFKA_OP_ASSIGN` op to the consumer-group background thread and waits with `RD_POLL_INFINITE` via `rd_kafka_op_req` → `rd_kafka_q_pop`. If the cgrp thread is blocked (e.g. waiting for a coordinator while the broker is unreachable after a network interruption), `poll_batch` never returns — regardless of its own timeout — because the block happens inside a callback dispatched during queue serving. This makes pipeline cancellation impossible, causing `DROP TABLE` to hang for 600+ seconds until the client times out.

```
poll_batch(timeout=500ms)
  → rd_kafka_consume_batch_queue()
    → rd_kafka_q_serve_rkmessages()
      → rd_kafka_poll_cb()            // serves rebalance callback
        → Consumer::handle_rebalance()
          → Consumer::assign()
            → rd_kafka_assign()
              → rd_kafka_assign0()
                → rd_kafka_op_req(cgrp_ops, rko, RD_POLL_INFINITE)
                  → rd_kafka_q_pop(recvq, INFINITE)  ← blocks forever
```

Updates both `contrib/librdkafka` and `contrib/cppkafka`:
- **librdkafka** (ClickHouse/librdkafka#18): replace `RD_POLL_INFINITE` with 30s timeout in `rd_kafka_assign0`
- **cppkafka** (ClickHouse/cppkafka#7): catch `HandleException`, `std::exception`, and `...` from `assign`/`unassign` in `handle_rebalance` to prevent any exception from escaping the C callback; route errors through existing `rebalance_error_callback_`

This was identified as the cause of frequent Kafka integration test failures in CI (cascade failures in `test_batch_fast` due to `test_kafka_consumer_hang` timing out).

Same hang stack trace as in the issue closed by #100604, which only addressed the consumer-destructor path; the broader fix here also handles the unbounded wait during normal `poll_batch`.

Refs:
- Issue: https://github.com/ClickHouse/ClickHouse/issues/100511
[...]

PR #99980, opened 2026-03-18
Add TLP and NoREC correctness oracles to the server-side AST fuzzer

qoega:
### Changelog category (leave one):
  - Build/Testing/Packaging Improvement

  ### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

  Fixed a lifetime bug in `ColumnAggregateFunction::convertToValues` that could produce incorrect results or an exception for `-State` aggregates combined with `-OrDefault` / `-OrFill` under `WITH TOTALS`, `ROLLUP`, or `CUBE`. The bug was found by the TLP and NoREC correctness oracles this PR adds to the server-side AST fuzzer (internal testing infrastructure, controlled by the new experimental `ast_fuzzer_oracle` setting, with a dedicated CI job running the oracle-enabled fuzzer against the release build).

  ### Documentation entry for user-facing changes

  - [ ] Documentation is written (mandatory for new features)

  Not a user-facing feature — this is internal testing infrastructure.

  ---

  **Summary**

  Integrates two metamorphic testing oracles from the SQLancer project into the built-in server-side AST fuzzer:

  - **TLP WHERE** (Ternary Logic Partitioning): for a query with `WHERE p`, verifies that `SELECT ... FROM t` produces the same sorted result set as `SELECT ... WHERE p UNION ALL SELECT ... WHERE NOT p UNION ALL SELECT
  ... WHERE isNull(p)`.

  - **NoREC** (Non-optimizing Reference Engine Construction): verifies that `SELECT count() FROM t WHERE cond` equals `SELECT countIf(cond) FROM t`.

  Instead of generating queries from scratch (as SQLancer does), the oracles check queries that the AST fuzzer already produces and executes successfully. Applicable queries must be simple SELECTs (no JOINs, DISTINCT,
  GROUP BY, HAVING, LIMIT, QUALIFY, window/aggregate functions) with a WHERE clause against a real table. Non-deterministic functions are detected and skipped.

  On mismatch, throws `LOGICAL_ERROR` which crashes the server — making bugs immediately visible in CI.

  **CI integration:**
  - New job **"AST fuzzer (amd_release, oracle)"** in PR checks, using the release (RelWithDebInfo) build.
  - Enabled via `AST_FUZZER_ORACLE=1` environment variable which applies `ast_fuzzer_runs=1` and `ast_fuzzer_oracle=1` as server default profile settings.

[...]

PR #105102, opened 2026-05-16
Fix LimitTransform WITH TIES assertion on empty virtual-row chunk

### LINKED ISSUE #109948: Logical error: 'row < chunk.getNumRows()' (STID: 2508-4656)

Test name: Logical error: 'row < chunk.getNumRows()' (STID: 2508-4656)
CI report: [Stress test (arm_asan_ubsan)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=108286&sha=7bf9825eedf03f952e187585b4116224ee80318b&name_0=PR&name_1=Stress%20test%20%28arm_asan_ubsan%29)
Failing test history: [cidb](https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICdMb2dpY2FsIGVycm9yOiAnJ3JvdyA8IGNodW5rLmdldE51bVJvd3MoKScnIChTVElEOiAyNTA4LTQ2NTYpJwogICAgLS0gQU5EIGNoZWNrX25hbWUgPSAnU3RyZXNzIHRlc3QgKGFybV9hc2FuX3Vic2FuKScKICAgIEFORCB0ZXN0X3N0YXR1cyBJTiAoJ0ZBSUwnLCAnRVJST1InKQogICAgQU5EIChwdWxsX3JlcXVlc3RfbnVtYmVyID0gMCBPUiBiYXNlX3JlZiBJTiAoJ21hc3RlcicpKQogICAgLS0gQU5EIHRlc3RfY29udGV4dF9yYXcgTElLRSAnJXBhdHRlcm4lJyAgLS0gdW5jb21tZW50IGFuZCBlZGl0IHRvIGZpbHRlciBieSBmYWlsdXJlIHBhdHRlcm4KR1JPVVAgQlkgZGF5Ck9SREVSIEJZIGRheSBERVNDCg==)

Test output:
```
Error:
Logical error: 'row < chunk.getNumRows()'.
---

Stack trace:
__pthread_kill_implementation @ 0x0000000000082009
gsignal @ 0x000000000003a83c
abort @ 0x0000000000027134
src/Common/Exception.cpp:60:5: DB::abortOnFailedAssertion(String const&, std::basic_string_view<char, std::char_traits<char>>, void* const*, unsigned long, unsigned long) @ 0x000000002d075418
src/Common/Exception.cpp:66:5: ? @ 0x000000002d076910
src/Processors/FractionalLimitTransform.cpp:80:5: DB::FractionalLimitTransform::makeChunkWithPreviousRow(DB::Chunk const&, unsigned long) const @ 0x000000004e72f7b4
src/Processors/FractionalLimitTransform.cpp:355:33: DB::FractionalLimitTransform::pushData() @ 0x000000004e7337e4
src/Processors/FractionalLimitTransform.cpp:221:19: DB::FractionalLimitTransform::prepare() @ 0x000000004e730384
src/Processors/Executors/ExecutingGraph.cpp:344:59: DB::ExecutingGraph::updateNode(DB::ExecutingGraph::Node*, std::queue<DB::ExecutingGraph::Node*, boost::container::devector<DB::ExecutingGraph::Node*, AllocatorWithMemoryTracking<DB::ExecutingGraph::Node*>, void>>&, std::queue<DB::ExecutingGraph::Node*, boost::container::devector<DB::ExecutingGraph::Node*, AllocatorWithMemoryTracking<DB::ExecutingGraph::Node*>, void>>&) @ 0x000000004e792f58
src/Processors/Executors/PipelineExecutor.cpp:404:38: DB::PipelineExecutor::executeStepImpl(unsigned long, DB::WorkloadResources&&, std::atomic<bool>*) @ 0x000000004e781828
inlined from src/Processors/Executors/PipelineExecutor.cpp:352: DB::PipelineExecutor::executeSingleThread(unsigned long, DB::WorkloadResources&&)
src/Processors/Executors/PipelineExecutor.cpp:726:5: DB::PipelineExecutor::executeImpl(unsigned long, bool) @ 0x000000004e77fce0
src/Processors/Executors/PipelineExecutor.cpp:209:9: DB::PipelineExecutor::execute(unsigned long, bool) @ 0x000000004e77f3a4
inlined from src/Processors/Executors/CompletedPipelineExecutor.cpp:43: DB::threadFunction(DB::CompletedPipelineExecutor::Data&, std::shared_ptr<DB::ThreadGroup>, unsigned long, bool)
inlined from src/Processors/Executors/CompletedPipelineExecutor.cpp:83: operator()
inlined from contrib/llvm-project/libcxx/include/__type_traits/invoke.h:90: std::__invoke_result_impl<void, DB::CompletedPipelineExecutor::execute()::$_0&>::type std::__invoke[abi:fqe220101]<DB::CompletedPipelineExecutor::execute()::$_0&>(DB::CompletedPipelineExecutor::execute()::$_0&)
inlined from contrib/llvm-project/libcxx/include/tuple:1400: decltype(auto) std::__apply_tuple_impl[abi:fqe220101]<DB::CompletedPipelineExecutor::execute()::$_0&, std::tuple<>&>(DB::CompletedPipelineExecutor::execute()::$_0&, std::tuple<>&, std::integer_sequence<unsigned long, ...>)
inlined from contrib/llvm-project/libcxx/include/tuple:1404: decltype(auto) std::apply[abi:fqe220101]<DB::CompletedPipelineExecutor::execute()::$_0&, std::tuple<>&>(DB::CompletedPipelineExecutor::execute()::$_0&, std::tuple<>&)
inlined from src/Common/ThreadPool.h:363: operator()
inlined from contrib/llvm-project/libcxx/include/__type_traits/invoke.h:90: std::__invoke_result_impl<void, ThreadFromGlobalPoolImpl<true, true>::ThreadFromGlobalPoolImpl<DB::CompletedPipelineExecutor::execute()::$_0>(DB::CompletedPipelineExecutor::execute()::$_0&&)::'lambda'()&>::type std::__invoke[abi:fqe220101]<ThreadFromGlobalPoolImpl<true, true>::ThreadFromGlobalPoolImpl<DB::CompletedPipelineExecutor::execute()::$_0>(DB::CompletedPipelineExecutor::execute()::$_0&&)::'lambda'()&>(ThreadFromGlobalPoolImpl<true, true>::ThreadFromGlobalPoolImpl<DB::CompletedPipelineExecutor::execute()::$_0>(DB::CompletedPipelineExecutor::execute()::$_0&&)::'lambda'()&)
inlined from contrib/llvm-project/libcxx/include/__type_traits/invoke.h:350: void std::__invoke_void_return_wrapper<void, true>::__call[abi:fqe220101]<ThreadFromGlobalPoolImpl<true, true>::ThreadFromGlobalPoolImpl<DB::CompletedPipelineExecutor::execute()::$_0>(DB::CompletedPipelineExecutor::execute()::$_0&&)::'lambda'()&>(ThreadFromGlobalPoolImpl<true, true>::ThreadFromGlobalPoolImpl<DB::CompletedPipelineExecutor::execute()::$_0>(DB::CompletedPipelineExecutor::execute()::$_0&&)::'lambda'()&)
inlined from contrib/llvm-project/libcxx/include/__type_traits/invoke.h:356: DB::CompletedPipelineExecutor::execute()::$_0 std::__invoke_r[abi:fqe220101]<void, ThreadFromGlobalPoolImpl<true, true>::ThreadFromGlobalPoolImpl<DB::CompletedPipelineExecutor::execute()::$_0>(DB::CompletedPipelineExecutor::execute()::$_0&&)::'lambda'()&>()
contrib/llvm-project/libcxx/include/__functional/function.h:443:24: ? @ 0x000000004e77d490
inlined from contrib/llvm-project/libcxx/include/__functional/function.h:502: ?
inlined from contrib/llvm-project/libcxx/include/__functional/function.h:754: ?
inlined from src/Common/ThreadPool.cpp:1293: operator()
inlined from contrib/llvm-project/libcxx/include/__type_traits/invoke.h:90: std::__invoke_result_impl<void, startThreadFromGlobalPool(std::shared_ptr<ThreadFromGlobalPoolState>, std::function<void ()>, unsigned long, unsigned long, bool, bool)::$_0&>::type std::__invoke[abi:fqe220101]<startThreadFromGlobalPool(std::shared_ptr<ThreadFromGlobalPoolState>, std::function<void ()>, unsigned long, unsigned long, bool, bool)::$_0&>(startThreadFromGlobalPool(std::shared_ptr<ThreadFromGlobalPoolState>, std::function<void ()>, unsigned long, unsigned long, bool, bool)::$_0&)
inlined from contrib/llvm-project/libcxx/include/__type_traits/invoke.h:350: void std::__invoke_void_return_wrapper<void, true>::__call[abi:fqe220101]<startThreadFromGlobalPool(std::shared_ptr<ThreadFromGlobalPoolState>, std::function<void ()>, unsigned long, unsigned long, bool, bool)::$_0&>(startThreadFromGlobalPool(std::shared_ptr<ThreadFromGlobalPoolState>, std::function<void ()>, unsigned long, unsigned long, bool, bool)::$_0&)
inlined from contrib/llvm-project/libcxx/include/__type_traits/invoke.h:356: void std::__invoke_r[abi:fqe220101]<void, startThreadFromGlobalPool(std::shared_ptr<ThreadFromGlobalPoolState>, std::function<void ()>, unsigned long, unsigned long, bool, bool)::$_0&>(startThreadFromGlobalPool(std::shared_ptr<ThreadFromGlobalPoolState>, std::function<void ()>, unsigned long, unsigned long, bool, bool)::$_0&)
contrib/llvm-project/libcxx/include/__functional/function.h:443:12: ? @ 0x000000002d493aa0
inlined from contrib/llvm-project/libcxx/include/__functional/function.h:502: ?
inlined from contrib/llvm-project/libcxx/include/__functional/function.h:754: ?
src/Common/ThreadPool.cpp:1113:12: ThreadPoolImpl<std::thread>::ThreadFromThreadPool::worker() @ 0x000000002d495110
inlined from contrib/llvm-project/libcxx/include/__type_traits/invoke.h:0: std::__invoke_result_impl<void, void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*>::type std::__invoke[abi:fqe220101]<void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*>(void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*&&)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*&&)
inlined from contrib/llvm-project/libcxx/include/__thread/thread.h:161: void std::__thread_execute[abi:fqe220101]<std::unique_ptr<std::__thread_struct, std::default_delete<std::__thread_struct>>, void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*, 0ul, 1ul>(std::tuple<std::unique_ptr<std::__thread_struct, std::default_delete<std::__thread_struct>>, void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*>&, std::__integer_sequence<unsigned long, 0ul, 1ul>)
contrib/llvm-project/libcxx/include/__thread/thread.h:169: void* std::__thread_proxy[abi:fqe220101]<std::tuple<std::unique_ptr<std::__thread_struct, std::default_delete<std::__thread_struct>>, void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*>>(void*) @ 0x000000002d4a56e4
asan_thread_start(void*) @ 0x000000001d83c9c8
start_thread @ 0x0000000000080398
thread_start @ 0x00000000000e9e9c
```

### Version info
- Resolved by: #105102
- Merged into: `26.7.1.831` (included in `26.7` and later)

groeneai:
## Description

Chronic AST fuzzer assertion `Assertion 'row < chunk.getNumRows()' failed` in `LimitTransform::makeChunkWithPreviousRow` (`src/Processors/LimitTransform.cpp:60`), triggered by `LIMIT ... WITH TIES` queries when the read-in-order pipeline emits a per-block virtual-row marker that reaches `LimitTransform` after `rows_read` already hit the limit boundary. The marker chunk has zero rows, so `getNumRows() - 1` underflows to `UINT64_MAX` and trips the assertion (server abort in debug builds; surfaces as a `LOGICAL_ERROR` in release builds). The AST fuzzer has hit this on master across multiple unrelated PRs (#99495, #104268, #101158) over the past 60 days.

### Root cause

With `optimize_read_in_order = 1`, `read_in_order_use_virtual_row = 1` and `read_in_order_use_virtual_row_per_block = 1`, `VirtualRowTransform` deliberately produces zero-row chunks carrying `MergeTreeReadInfo`. They are intended as sort-key markers for `MergingSortedTransform`. In the single-stream case there is no merging, and `RemoveVirtualRowTransform` is inserted to clean them up — but its `transform` only stripped the `ChunkInfo` and left the empty chunk on the pipeline. When such an empty chunk arrived at `LimitTransform` after a previous data chunk had pushed `rows_read` to exactly `offset + limit`, the `with_ties && rows_read == offset + limit` branch called `makeChunkWithPreviousRow(empty_chunk, 0 - 1)` and the underflow triggered the assertion.

### Deterministic reproducer (pre-fix)

```sql
CREATE TABLE t (id UInt64, m Map(String, String),
                INDEX idx_mk mapKeys(m) TYPE bloom_filter GRANULARITY 1)
ENGINE = MergeTree ORDER BY id SETTINGS index_granularity = 1;
INSERT INTO t VALUES (1, {'1':'1'});
INSERT INTO t VALUES (2, {'2':'2'});
INSERT INTO t VALUES (3, {'3':'3'});

SELECT id FROM t PREWHERE mapContains(m, toFixedString('1', 1))
ORDER BY id ASC LIMIT 1 WITH TIES
SETTINGS optimize_read_in_order = 1,
         read_in_order_use_virtual_row = 1,
         read_in_order_use_virtual_row_per_block = 1;
```

### Fix

Two small, focused changes:

1. **`src/Processors/LimitTransform.cpp`** — defensive guard. Skip the `makeChunkWithPreviousRow` call when the current chunk has zero rows. There is no row to remember and the `previous_row_chunk` saved from the prior non-empty chunk is still the correct tie key at the boundary.

2. **`src/Processors/QueryPlan/SortingStep.cpp`** — make `RemoveVirtualRowTransform` actually consume virtual rows: `chunk.clear()` them and enable `skip_empty_chunks` on the `ISimpleTransform` base so they are not re-emitted. This matches the existing comment ("we need to remove virtual row before output") and protects other downstream transforms from receiving these pointless markers.

The two fixes are complementary: `LimitTransform` becomes robust against any source of empty chunks, and `RemoveVirtualRowTransform` no longer leaks zero-row markers downstream.

### Regression test

`tests/queries/0_stateless/04247_99495_limit_with_ties_virtual_row_empty_chunk.{sql,reference}` aborts master in debug builds and passes after this change. Verified locally 10x; existing virtual-row and `WITH TIES` tests all still pass.

### Related

- Sighting PR (vehicle): https://github.com/ClickHouse/ClickHouse/pull/99495
- Other affected PRs: #104268, #101158
- Old issue (same assertion, never fully fixed): #11372
- Sibling `FractionalLimitTransform` manifestation of the same empty-chunk boundary (fractional `LIMIT ... WITH TIES`): Closes #109948 (STID 2508-4656). Its guards are already included here.
- CI report with stack trace: https://s3.amazonaws.com/clickhouse-test-reports/PRs/99495/b699481fd40feb67d515e47725de159f20dd1ff0/ast_fuzzer_amd_debug/

Fix a `LOGICAL_ERROR` `Assertion 'row < chunk.getNumRows()' failed` in `LimitTransform::makeChunkWithPreviousRow` when `LIMIT ... WITH TIES` queries run with the read-in-order pipeline (`optimize_read_in_order = 1` and `read_in_order_use_virtual_row_per_block = 1`). The exception was raised by `LimitTransform` when a per-block virtual-row marker (a zero-row chunk produced by `VirtualRowTransform`) reached it after `rows_read` had already hit the limit boundary.

- [x] Documentation is unchanged: this fix does not change observable behavior except removing the exception.

### Version info
- Merged into: `26.7.1.831` (included in `26.7` and later)


groeneai:
**Pre-PR validation gate (TASK.md §Phase 4 Step 9):**

**a) Deterministic repro?** Yes. Three single-row inserts (separate parts) into a `MergeTree` with `bloom_filter` skip index on `mapKeys(m)` and `index_granularity = 1`, plus the three settings `optimize_read_in_order = 1`, `read_in_order_use_virtual_row = 1`, `read_in_order_use_virtual_row_per_block = 1`. The query `SELECT id FROM t PREWHERE mapContains(m, toFixedString('1', 1)) ORDER BY id ASC LIMIT 1 WITH TIES` aborts the server every time on master (the first execution may complete before the abort lands in the client; the second hits before the result is returned).

**b) Root cause explained?** Yes. `VirtualRowTransform` produces a zero-row chunk carrying `MergeTreeReadInfo` as a sort-key marker for `MergingSortedTransform`. In the single-stream case there is no merging, and `RemoveVirtualRowTransform` is inserted to clean these up — but its `transform` only stripped the `ChunkInfo` and forwarded the empty chunk. When such a chunk reached `LimitTransform` after the previous data chunk had pushed `rows_read` to `offset + limit`, the `with_ties && rows_read == offset + limit` branch called `makeChunkWithPreviousRow(empty_chunk, 0 - 1)`. The `UInt64` underflow trips `assert(row < chunk.getNumRows())` (`src/Processors/LimitTransform.cpp:60`).

**c) Fix matches root cause?** Yes, at two levels. `LimitTransform` now skips the `makeChunkWithPreviousRow` call when the current chunk has zero rows (the previously stored `previous_row_chunk` is still the correct tie key at the boundary). `RemoveVirtualRowTransform` now actually consumes the virtual-row chunk (`chunk.clear()` + `skip_empty_chunks = true` on the `ISimpleTransform` base), matching the existing comment "we need to remove virtual row before output". The two are complementary: `LimitTransform` is robust against any source of empty chunks, and `RemoveVirtualRowTransform` no longer leaks markers.

**d) Test intent preserved? / New tests added?** New regression test `04247_99495_limit_with_ties_virtual_row_empty_chunk.{sql,reference}` reproduces the assertion on master (debug build aborts) and passes after this change. Two queries (LIMIT 1 WITH TIES, and LIMIT 0,1 WITH TIES at the offset boundary).

**e) Demonstrated in both directions?** Yes. Pre-fix binary: server abort, stack trace matches CIDB (`LimitTransform.cpp:60` → `LimitTransform.cpp:246`). Post-fix binary: both queries return `1`, 10/10 stable runs. The other call site at `LimitTransform.cpp:343` was reviewed and is safe — the surrounding `&& length` guard ensures `current_row_num >= 1` before the underflow-prone subtraction.

**f) Fix is general, not a narrow patch?** Yes. The same anti-pattern (assuming a chunk has rows after `rows_read` jumps) does not exist elsewhere in `LimitTransform`. The sibling `splitChunk` path at line 343 is already protected by `&& length`. The `RemoveVirtualRowTransform` change covers all downstream transforms in the single-stream read-in-order pipeline, not just `LimitTransform`. Existing virtual-row tests (`04201_read_in_order_limit_with_filter`, `04076_virtual_row_preliminary_merge`, `04057_virtual_row_many_parts`, `04049_virtual_row_overlap`, `04029_read_in_order_expression_virtual_row`, `04045_reverse_order_virtual_row`, `03581_read_in_order_use_virtual_row_WHERE`) all still pass locally, and existing `WITH TIES` tests (`00996_limit_with_ties`, `01525_select_with_offset_fetch_clause`, `03704_fractional_limit_with_ties`, `03725_empty_tuple_some_limit_with_ties_distinct`, `03753_fractional_limit_with_ties_small_blocks`) all still pass locally.

Session: `cron:clickhouse-ci-task-worker:20260516-020000`

groeneai:
cc @KochetovNicolai @alexey-milovidov @vdimir — could you review? Two-part fix for the chronic AST fuzzer assertion in `LimitTransform::makeChunkWithPreviousRow` (per @alexey-milovidov's note on #99495): `LimitTransform` now guards against empty incoming chunks, and `RemoveVirtualRowTransform` actually drops the zero-row virtual-row markers it was meant to consume (the existing implementation only stripped the `ChunkInfo`).

alexey-milovidov:

groeneai:
## CI ledger — covered SHA: `17f0d2975d7211a9625a8b48112231d6be2798ad`

|---|---|---|
| `Stateless tests (amd_tsan, parallel, 1/2)` | `00175_obfuscator_schema_inference` | flaky chronic trunk — task 2026-06-02-ci-00175obfuscatorschemainference-ab |
| `Unit tests (asan_ubsan)` | (check-level) | infra/chronic UBSan UntrackedMemoryCounter — task 2026-06-03-ci-p1-chronic-ubsan-in-untrackedmemoryc |
| `Unit tests (asan_ubsan, function_prop_fuzzer)` | `FunctionsStress.stress`, `AllTests` | infra/chronic, follows the asan_ubsan check-level fail above (same UntrackedMemoryCounter teardown). |
| `AST fuzzer (amd_debug)` | `Logical error: Trying to get name of not a column: A (STID: 2310-3791)` | chronic trunk family bug — tracked in task 2026-05-30-ci-p0-ast-fuzzer-logical-error-trying-, our fix PR #106172 awaiting review. NOT PR-caused (this PR only touches `LimitTransform.{cpp,h}`). |
| `Bugfix validation (functional tests)` | `04247_99495_limit_with_ties_virtual_row_empty_chunk` | this PR's OWN regression test — framework limitation: the LimitTransform assertion bug requires `read_in_order_use_virtual_row_per_block` randomization that isn't always exercised on master HEAD. Not a real regression. |

vdimir:
> `Bugfix validation (functional tests)` 	`04247_99495_limit_with_ties_virtual_row_empty_chunk` 	this PR's OWN regression test — framework limitation: the LimitTransform assertion bug requires `read_in_order_use_virtual_row_per_block` randomization that isn't always exercised on master HEAD. Not a real regression.

@groeneai you may set this setting explicit in the test.

groeneai:
@ vdimir thanks — pushed `b7d6e3492a0c`.

The query-level `SETTINGS` clause already pins `optimize_read_in_order`, `read_in_order_use_virtual_row`, and `read_in_order_use_virtual_row_per_block` on both queries. The remaining knob that can mask the bug on master HEAD is MergeTree randomization (`index_granularity_bytes`, `min_bytes_for_wide_part`, etc.) which can change part type / mark layout so the per-block virtual row marker never reaches `LimitTransform` at the boundary. Added `no-random-merge-tree-settings` to keep the data layout deterministic. Awaiting Bugfix validation re-run.

Session: cron:clickhouse-ci-task-worker:20260605-194200

groeneai:
### Pre-PR validation gate (fix-up commit `a847a6faa322`)

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. Piping the test SQL (with the new pinned `use_skip_indexes_on_data_read = 0`) into a debug `clickhouse` built from master HEAD without the `LimitTransform` fix aborts every time on `LimitTransform.cpp:60` `Logical error: 'row < chunk.getNumRows()'`. Just the four pinned settings (`optimize_read_in_order = 1`, `read_in_order_use_virtual_row = 1`, `read_in_order_use_virtual_row_per_block = 1`, `use_skip_indexes_on_data_read = 0`) trigger it; no other randomization required. |
| b | **Root cause explained?** | Yes. With `use_skip_indexes_on_data_read = 1` (default), the bloom filter is consumed in a separate index-only step and the data read pipeline never sees the empty virtual-row chunk that `read_in_order_use_virtual_row_per_block = 1` emits. With `use_skip_indexes_on_data_read = 0`, the read goes through `MergeTreeRangeReader` on the data path; the `RemoveVirtualRowTransform` upstream in `SortingStep` only stripped `ChunkInfo` (didn't actually drop the zero-row chunk). After the previous data chunk pushed `rows_read` to exactly `offset + limit`, the empty chunk reached `LimitTransform::preparePair`, which called `makeChunkWithPreviousRow(empty_chunk, getNumRows() - 1)` — a `UInt64` underflow that trips the assertion. |
| c | **Fix matches root cause?** | Yes. The PR fix has two parts: (1) `LimitTransform.cpp` adds `&& rows > 0` so `makeChunkWithPreviousRow` is only called when there is a row to remember; (2) `SortingStep.cpp` switches `RemoveVirtualRowTransform`'s `ISimpleTransform` to `skip_empty_chunks=true` so the marker is consumed instead of forwarded. The test fix here only pins the runner randomization that exposes the broken pipeline shape, so `Bugfix validation` reproduces the bug deterministically against master HEAD. |
| d | **Test intent preserved / new tests added?** | Yes. The test still exercises the exact original bug: `LIMIT 1 WITH TIES` and `LIMIT 0, 1 WITH TIES` over a single-stream `read_in_order_use_virtual_row_per_block = 1` pipeline, with `mapContains` PREWHERE on a `bloom_filter` skip index. Pinning `use_skip_indexes_on_data_read = 0` does not weaken the assertion — it forces the failing pipeline shape that the AST fuzzer was hitting through randomization. The `no-random-merge-tree-settings` tag added in the previous fix-up is dropped: the bug is independent of compact-vs-wide parts and other MergeTree randomization, so the broad tag was not the right tool. |
| e | **Both directions demonstrated?** | Yes. **Without the fix** (debug `clickhouse` built from master HEAD): piping the test SQL aborts the server with `Logical error: 'row < chunk.getNumRows()'` at `LimitTransform.cpp:60`. **With the fix** (this PR): `Bugfix validation` reported the test passing in 0.21s on the prior commit (`b7d6e3492a0c`), and the LimitTransform path now safely skips the empty chunk. The fix-up only adds an extra `SETTING` to the regression test; the source-code fix and PR semantics are unchanged. |
| f | **Fix is general, not a narrow patch?** | Yes. The source-code fix already covers all read-in-order paths that can produce empty marker chunks: (1) `LimitTransform` guards `makeChunkWithPreviousRow` against zero-row chunks regardless of upstream — protects every `WITH TIES` consumer; (2) `RemoveVirtualRowTransform` enables `skip_empty_chunks` so all downstream transforms (`MergingSortedTransform` not in the path, `LimitTransform`, etc.) never see these markers. The test's pinned settings narrow the runner randomization just enough for `Bugfix validation` to deterministically reproduce; it is not a defensive symptom-only patch. |

Session id: cron:clickhouse-ci-task-worker:20260605-212000

groeneai:
@vdimir thanks — the previous fix-up's `no-random-merge-tree-settings` umbrella was the wrong tool for this. Pushed `a847a6faa322` which pins the actual trigger: `use_skip_indexes_on_data_read = 0` in both query SETTINGS clauses, and drops the broad MergeTree randomization tag.

Minimal trigger isolated locally — the four pinned settings (`optimize_read_in_order = 1`, `read_in_order_use_virtual_row = 1`, `read_in_order_use_virtual_row_per_block = 1`, `use_skip_indexes_on_data_read = 0`) deterministically abort `clickhouse` built from master HEAD on `LimitTransform.cpp:60` `Logical error: 'row < chunk.getNumRows()'`. Verified by piping the test SQL into the buggy binary.

The `Bugfix validation` framework should now reproduce the assertion deterministically.

groeneai:
@vdimir thanks. The four `read_in_order` related settings are pinned in the test SQL since `a847a6faa322`. Local Debug build of master HEAD (`9c9b2b9a0da7`) without our `LimitTransform.cpp` change deterministically aborts on `LimitTransform.cpp:60` `Logical error: 'row < chunk.getNumRows()'` on every run when piping the test SQL into `clickhouse-client` (verified just now).

The remaining `Bugfix validation (functional tests)` failure is a framework limitation, not a settings issue. The check downloads `master/aarch64/clickhouse` (release build), and `chassert` is compiled out in release (see `base/base/defines.h:42-49`). With the assertion stripped, the empty-chunk path silently calls `insertFrom(col, UINT64_MAX)` and the test produces the same output rows as the fixed branch, so the runner reports `Failed to reproduce the bug`. This affects any chassert-only fix in `Bugfix validation`; with `GLOBAL_TAGS=no-random-settings` set by the bugfix runner, no amount of in-test setting pinning can change that.

Two options:

1. Land this PR despite the Bugfix-validation gap (the regression test is correct under Debug/sanitizer builds, where the AST fuzzer originally hit it; chassert assertions surface in `Stress test` and `AST fuzzer` pipelines that DO fire).
2. Replace `chassert` at `LimitTransform.cpp:60` with an exception throw so the misbehavior is observable in release as well. That's a separate behavioral change with broader implications. Happy to do it in a follow-up PR if you prefer that as the long-term direction.

Session: cron:clickhouse-ci-task-worker:20260605-225900

groeneai:
Hi @vdimir, gentle ping on the framework-limitation question. The chassert at LimitTransform.cpp:60 is a release no-op, so Bugfix-validation (which uses release master HEAD) cannot reproduce the assertion regardless of test-level pinning. Two options: (1) land this PR despite the gap, or (2) replace the chassert with an exception throw so the misbehavior is observable in release too. Happy to follow either path. Thanks.

src/Processors/QueryPlan/SortingStep.cpp
```diff
@@ -43,21 +43,29 @@ namespace ProfileEvents
...
     void transform(Chunk & chunk) override
     {
+        if (isVirtualRow(chunk))
+        {
+            /// Drop the virtual row entirely: it was a marker for the merging algorithm,
+            /// and there is no merging in this branch (single stream), so it has no downstream purpose.
+            chunk.clear();
+            return;
+        }
         chunk.getChunkInfos().extract<MergeTreeReadInfo>();
```
vdimir, line 69: Why `extract<MergeTreeReadInfo>`  below doesn't do the same?

src/Processors/QueryPlan/SortingStep.cpp
```diff
@@ -43,21 +43,29 @@ namespace ProfileEvents
...
+/// `MergingSortedTransform` is supposed to consume virtual rows.
+/// When there is no merging (only one stream) and virtual row conversions are enabled, we need to remove virtual rows before output,
+/// otherwise they can reach downstream steps and cause issues, both because the attached conversions are valid only for the current step
+/// and because the virtual rows are empty chunks that some downstream transforms (e.g. `LimitTransform` with `WITH TIES`) do not expect.
 class RemoveVirtualRowTransform : public ISimpleTransform
 {
 public:
     explicit RemoveVirtualRowTransform(SharedHeader header)
-        : ISimpleTransform(header, header, false)
+        : ISimpleTransform(header, header, /*skip_empty_chunks=*/ true)
```
vdimir, line 54: Why do we change behavior here?

src/Processors/QueryPlan/SortingStep.cpp
groeneai: Because the two branches handle different chunks.

`isVirtualRow` is true only for the index-boundary marker: a 0-row chunk whose payload is the non-empty `pk_block` in `MergeTreeReadInfo` (`MergeTreeSelectProcessor::buildVirtualRowFromIndex`). It carries no result rows, so we drop it entirely.

The `extract` line handles real data chunks. Those also carry a `MergeTreeReadInfo`, but only to convey `origin_merge_tree_part_level` (empty `pk_block`, so `isVirtualRow` is false), and they hold actual rows we must forward. We only strip the now-irrelevant ChunkInfo and keep the chunk. Clearing it there would discard query results.

One subtlety: `extract` alone cannot drop the marker. The virtual-row chunk has placeholder columns (0 rows), so after extracting the ChunkInfo it still reports `hasColumns()` and is not `empty()`, hence `skip_empty_chunks` would not suppress it. That is why the virtual-row branch needs the explicit `chunk.clear()`, which drops the columns too.

src/Processors/QueryPlan/SortingStep.cpp
groeneai: The flag flip is what makes the `clear()` above effective. With `skip_empty_chunks=false`, `ISimpleTransform::work()` forces `has_output=true` for the cleared chunk and then rebuilds it into a 0-row chunk carrying the header columns (`ISimpleTransform.cpp`, the `output_data.chunk = Chunk(header.cloneEmpty()...)` branch). That rebuilt empty chunk is exactly what reaches `LimitTransform` and trips the WITH TIES assertion. With `skip_empty_chunks=true`, a chunk we cleared (now `empty()`) is suppressed and never forwarded. Real data chunks are non-empty, so the flag does not change their handling.

Strategic note, your call: the `LimitTransform` guard (`rows > 0`) alone already prevents the crash. This SortingStep change is the root-cause half: it makes `RemoveVirtualRowTransform` actually honor its documented contract ("remove virtual row before output"), which today only strips the `MergeTreeReadInfo` and leaks the empty husk downstream. If you prefer to keep the fix minimal and rely solely on the LimitTransform guard, I can drop this part. Which direction do you prefer?

groeneai:
CI summary on HEAD `44bbea1` (CI fully finished). Disposition of each failure:

- **Build (arm_tidy)** — now **green** on this HEAD. The earlier red was a trunk-stale clang-tidy `init-variables` warning in `Iceberg/Mutations.cpp` (introduced by #106102, unrelated to this PR); the master merge into this branch picked up the upstream fix (#106609), clearing it.
- **Bugfix validation (functional tests)** / `04247_99495_limit_with_ties_virtual_row_empty_chunk` — expected framework limitation. The original `LimitTransform` guard is a debug-only `chassert`, which is a no-op in the release binary the bugfix-validation job builds from master HEAD, so the regression test cannot deterministically FAIL there. The bug is real and reproduces deterministically on a debug build (validated both directions, above). Same pattern as discussed on this PR's design thread.
- **Stateless tests (arm_asan_ubsan, azure, parallel/sequential)** — `Server died` + `Segmentation fault` (STID 0883-5001) — unrelated infrastructure: the 2026-06-10/11 `arm_asan_ubsan, azure` ASan-allocator OOM outage that hit ~100 unrelated PRs and master in the same window.
- **Stress test (arm_ubsan)** — `Hung check failed, possible deadlock found` — chronic stress-runner family, not specific to this change.

No PR-caused failures. Ready for review.

groeneai:
The AST fuzzer hit this exact assertion again on 2026-06-15 (STID 2508-3b8e, PR #103322's `AST fuzzer (amd_debug, targeted)`), via a second, independent code path. Pushed `eed48fb9d82` adding it to the regression test.

```sql
SELECT DISTINCT a AS b, '[a-z]', 100 AS a ORDER BY 1 DESC NULLS FIRST WITH FILL STALENESS -1 LIMIT 1 WITH TIES;
```

`EXPLAIN PIPELINE` shows there is no read-in-order / virtual-row pipeline here:
```
(Limit) Limit
  (Filling) FillingTransform
    (Distinct) DistinctSortedTransform
      ...
        (ReadFromSystemOne)
```
`WITH FILL` emits a zero-row chunk that reaches `LimitTransform` at the `WITH TIES` boundary, so `makeChunkWithPreviousRow(chunk, getNumRows() - 1)` underflows. Only the `LimitTransform` `rows > 0` guard (fix #1) protects this case; the `SortingStep`/`RemoveVirtualRowTransform` change (fix #2) does not apply since there is no merging stage. Same assertion, different empty-chunk source, which is good evidence that the `rows > 0` guard is the correct general fix.

The bug is still live on current master: a Debug build of `b5fa90e2908` (master as of today, 110 commits behind origin/master, neither fix present) SIGABRTs on the query above; the same binary with this branch's `LimitTransform.cpp` change returns `100 [a-z] 100` cleanly.

This new case is also a cleaner regression test for the `Bugfix validation` concern you raised: it needs no `read_in_order_use_virtual_row_per_block` randomization at all, just a plain table-less query, so it does not depend on the fragile virtual-row path.

### Pre-PR validation gate (test-only commit `eed48fb9d82`)

| # | Question | Answer |
|---|----------|--------|
| a | Deterministic repro? | Yes. `clickhouse local -q "SELECT DISTINCT a AS b, '[a-z]', 100 AS a ORDER BY 1 DESC NULLS FIRST WITH FILL STALENESS -1 LIMIT 1 WITH TIES"` aborts every time on an unguarded build. |
| b | Root cause explained? | `WITH FILL` (here `STALENESS -1` on a DESC sort) emits a zero-row chunk; at the `WITH TIES` boundary `LimitTransform` calls `makeChunkWithPreviousRow(chunk, getNumRows() - 1)`, and `getNumRows()==0` makes `row` underflow to `UINT64_MAX`, tripping `row < chunk.getNumRows()`. |
| c | Fix matches root cause? | N/A for this commit (test-only). The covering fix is the existing `rows > 0` guard at `LimitTransform.cpp`. |
| d | Test intent preserved / new test added? | Test strengthened: adds a second, independent empty-chunk source (FillingTransform) to the existing file; the original virtual-row queries are unchanged. |
| e | Demonstrated in both directions? | Yes. Unguarded build of this branch (guard reverted) aborts with `makeChunkWithPreviousRow` (LimitTransform.cpp:60) <- `preparePair` (:249) <- `prepare` (:108), matching STID 2508-3b8e; guard restored returns `100 [a-z] 100`. Master snapshot `b5fa90e2908` also aborts. 20/20 runner-diff passes with the fix. |
| f | Fix is general, not a narrow patch? | Yes. The `rows > 0` guard makes `LimitTransform` robust to empty chunks from any upstream; this commit demonstrates a second source beyond the virtual-row path. |

Session: cron:clickhouse-worker-slot-2:20260615-232700

src/Processors/LimitTransform.cpp
```diff
@@ -242,7 +242,10 @@ LimitTransform::Status LimitTransform::preparePair(PortsData & data)
         /// Return the whole chunk.
 
         /// Save the last row of current chunk to check if next block begins with the same row (for WITH TIES).
-        if (with_ties && rows_read == offset + limit)
+        /// Skip when the chunk is empty (e.g. a virtual row produced by read-in-order pipeline that
+        /// reached us after rows_read already hit the limit): there is no row to remember, and the
+        /// previously stored `previous_row_chunk` is still the correct tie key for the boundary.
+        if (with_ties && rows_read == offset + limit && rows > 0)
```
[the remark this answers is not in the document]
groeneai: Done in cbfa201. Mirrored the `chunk_rows > 0` guard into both unguarded WITH TIES boundary sites in `FractionalLimitTransform`:

- early-push path in `prepare()` (the `rows_processed == limit_rows_for_current_input + offset_rows` branch)
- final-push path in `pushData()` (the `rows_processed == offset_rows + limit_rows` branch)

The `splitChunk()` site is already guarded by a non-zero `cut_length`, matching `LimitTransform::splitChunk`'s `&& length` guard, so it needs no change.

Confirmed the trace you described with a deterministic fractional reproducer that aborts on the pre-fix binary at `FractionalLimitTransform.cpp:80` (`makeChunkWithPreviousRow`, `row=18446744073709551615`) and returns the expected row after the fix:

```sql
SELECT DISTINCT a AS b, '[a-z]', 100 AS a ORDER BY 1 DESC NULLS FIRST WITH FILL STALENESS -1 LIMIT 0.5 WITH TIES;
```

Added that as a fractional `LIMIT ... WITH TIES` case to `04247_99495_limit_with_ties_virtual_row_empty_chunk`. The existing virtual-row and WITH FILL cases still pass, and the non-distributed fractional/WITH TIES suite passes 30/30 with randomization.

groeneai:
### Pre-PR validation gate

For the FractionalLimitTransform fix-up (cbfa201):

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. `clickhouse local -q "SELECT DISTINCT a AS b, '[a-z]', 100 AS a ORDER BY 1 DESC NULLS FIRST WITH FILL STALENESS -1 LIMIT 0.5 WITH TIES"` aborts every time on the pre-fix binary (exit 134). |
| b | **Root cause explained?** | Yes. WITH FILL emits a trailing zero-row chunk. It reaches `FractionalLimitTransform::pushData` at the WITH TIES boundary (`rows_processed == offset_rows + limit_rows`), where `makeChunkWithPreviousRow(chunk, chunk_rows - 1)` is called with `chunk_rows == 0`, so `row` underflows to `UINT64_MAX` and `chassert(row < chunk.getNumRows())` fails. gdb confirmed frame `FractionalLimitTransform.cpp:80` with `row=18446744073709551615`. |
| c | **Fix matches root cause?** | Yes. Adds `&& chunk_rows > 0` to the two unguarded WITH TIES boundary sites, exactly mirroring the merged `LimitTransform` guard (`&& rows > 0`). The stored `ties_last_row` is still the correct tie key at the boundary, so an empty chunk is simply skipped. |
| d | **Test intent preserved / new tests added?** | Yes. Added a fractional `LIMIT ... WITH TIES` case to `04247_99495_limit_with_ties_virtual_row_empty_chunk`. The existing virtual-row and WITH FILL cases are unchanged and still pass. |
| e | **Both directions demonstrated?** | Yes. Pre-fix binary (Build ID 64077ec2): exit 134, gdb stack at `FractionalLimitTransform.cpp:80`. Post-fix binary (Build ID 7ecb11b6): exit 0, expected output, and the test passes 30/30 with randomization. |
| f | **Fix is general, not a narrow patch?** | Yes. Checked all three `makeChunkWithPreviousRow(_, X - 1)` sites in `FractionalLimitTransform`: early-push (prepare) and final-push (pushData) were both unguarded and are both fixed; `splitChunk` is already guarded by a non-zero `cut_length`, matching `LimitTransform::splitChunk`. This completes the sibling-path coverage requested in the review. |

Session id: cron:clickhouse-worker-slot-2:20260616-005900

groeneai:
CI finished on cbfa2015eda. No PR-caused failures; the Bugfix validation check is green on this head.

- Stress test (arm_debug) "Hung check failed, possible deadlock found": chronic stress hung-check flake, not from this PR. 30d CIDB shows 270 master hits plus ~20 unrelated PRs with the same signature; this PR itself hit it on 2026-06-10 (arm_ubsan) before the FractionalLimitTransform commit. It is 1/5 on arm_debug and 0/4 on all 10 other Stress variants, with the generic all-threads stack_trace dump. The empty-chunk guard only skips work at the WITH TIES boundary, so it cannot introduce a wait cycle.
- Performance Comparison (arm_release, 2/6): perf-comparison noise, not PR-caused.
- Bugfix validation in-job server-log scan rows (Lost forever / Lost s3 keys / S3_ERROR / OOM in dmesg): pre-existing infra noise present across the PR queue; the Bugfix validation check itself passed here.
- CH Inc sync: internal clickhouse-private mirror, not a public gate.

The regression test passes across all Stateless variants.

groeneai:
CI fully finished on HEAD `cbfa2015eda8` (Finish Workflow SUCCESS, Config Workflow SUCCESS, 0 pending). Triage of every failure:

- **Stress test (arm_debug)** — `Hung check failed, possible deadlock found`: chronic trunk infra flake, NOT caused by this PR. The captured stack is the generic `system.stack_trace` diagnostic dump (PipelineExecutor worker threads, no `LimitTransform` frame). 30-day distribution: 25 distinct PRs plus master, one hit each.
- **Performance Comparison (arm_release, master_head, 2/6)** — 0 individual test-regression rows; job-level perf-comparison noise, not PR-caused.
- **Bugfix validation (functional tests)** — `OOM in dmesg` / `S3_ERROR No such key` / `Lost forever for SharedMergeTree` / `Lost s3 keys` / `Exception in test runner`: standard bugfix-validation infra meta-rows, identical across the PR queue, not PR-caused.

The new regression test `04247_99495_limit_with_ties_virtual_row_empty_chunk` passes everywhere, including the flaky checks (amd_asan_ubsan / amd_debug / amd_msan / amd_tsan = 50/50 each). No real test failures on this HEAD.

alexey-milovidov:
Refreshed the branch onto current `master` (it had fallen ~6800 commits / ~19 days behind). The merge was clean — all three guards (`LimitTransform`, `FractionalLimitTransform`) and the `RemoveVirtualRowTransform` change apply unchanged on top of current `master`, and every symbol they rely on (`isVirtualRow`, the `ISimpleTransform` ctor, `Chunk::clear`) is still present.

One thing the merge required: the regression test number `04247` now collides with the newly-added `04247_attach_table_storage_clauses_without_engine_error`, so I renumbered it to `04419_99495_limit_with_ties_virtual_row_empty_chunk`. The test content is unchanged.

Fresh CI is running on `28524b0a461`.

groeneai:
### CI finish ledger — 28524b0a461e

CI fully finished (Finish Workflow + Mergeable Check: success) on the branch after it was refreshed onto current master. This PR fixes the chronic AST-fuzzer `LimitTransform::makeChunkWithPreviousRow` assertion (`LIMIT WITH TIES` over an empty virtual-row chunk). Every failure has an owner; none is PR-caused.

|---|---|---|
| CH Inc sync | - | CH Inc sync (private, not actionable by us) |

No test-level FAIL/ERROR rows in CIDB for this commit; the only red GitHub check is the private Cloud sync.

Session id: cron:our-pr-ci-monitor:20260630-173000

qoega:
Our aarch64 AST-fuzzer-oracle harness (PR #99980) hit this same assertion on the `merged7` cycle with a different query shape — same trigger settings, so #105102's fix should cover it. Cross-referencing here for the chronic-signature list.

Snapshot `runs/merged7/crashes/inst-09-20260704-120003`, ClickHouse `26.7.1` debug, aarch64. Session settings included `optimize_read_in_order=1`, `read_in_order_use_virtual_row=1`, `read_in_order_use_virtual_row_per_block=1` — exact prerequisites per your root-cause. The last `(internal)` query before the assertion:

```sql
SELECT flag_b, flag_b, flag_a FROM local_dup_mixed ORDER BY dt DESC LIMIT 1 WITH TIES
```

The duplicate-projected-column (`flag_b, flag_b`) plus `LIMIT 1 WITH TIES` on a `ORDER BY dt DESC` read-in-order pipeline is a slightly different arrival than your minimal repro (`PREWHERE mapContains ... ORDER BY id ASC LIMIT 1 WITH TIES`) but the empty-chunk-from-virtual-row → `makeChunkWithPreviousRow(empty, 0-1)` underflow path is the same. Might be worth adding a variant to the regression test if convenient — no urgency, just adds coverage.

For lineage: same assertion signature as closed #11372 (2020, different offset-overflow trigger).

alexey-milovidov (approved):
Ok.

PR #105103, opened 2026-05-16
Surface real callback exceptions in ZipArchiveReader instead of "Code = -100"

### LINKED ISSUE #104681: `s3()` fails to read parquet files embedded in zip when file exceeds a certain threshold
### Company or project name

_No response_

### Describe what's wrong

I have a zip file hosted in S3 containing embedded parquet files. When those files exceed a certain size, the query to read the files via `s3()` returns an error:

```json
{"meta":[]} {"exception":"Code: 643. DB::Exception: Couldn't unpack zip archive 'scalar_float64.parquet.zip': Code = -100: The table structure cannot be extracted from a Parquet format file. You can specify the structure manually: (in file\/uri scalar_float64.parquet.zip::scalar_float64.parquet). (CANNOT_UNPACK_ARCHIVE) (version 26.2.1.235 (official build))"}
```

Yes

### How to reproduce

 - Run the script below to generate a zip file containing a parquet file with N rows
 - Upload the file to S3
 - Attempt to query the file with a query such as:

```sql
SELECT *
FROM s3(
   'https://mybucket.s3.eu-west-2.amazonaws.com/scalar_float64.parquet.zip :: *.parquet',
   'Parquet',
);
```
 - When the file is below a certain size (roughly ~20Mb from my testing) the query will succeed. When it's over this size, it will fail with the error above. I've found a `ROW_COUNT` of 200k will generally work, while 300k will generally trigger the error.
 - I've confirmed that it's the file size that triggers the issue by also testing with a wide parquet file where each row contains embedded arrays. This triggers the error with a much smaller number of rows.

### Expected behavior

The file should be readable regardless of the size. If there is an intentional file size limit this should be surfaced in the error message instead of simply stating that the structure cannot be read.

```json
{"meta":[]} {"exception":"Code: 643. DB::Exception: Couldn't unpack zip archive 'scalar_float64.parquet.zip': Code = -100: The table structure cannot be extracted from a Parquet format file. You can specify the structure manually: (in file\/uri scalar_float64.parquet.zip::scalar_float64.parquet). (CANNOT_UNPACK_ARCHIVE) (version 26.2.1.235 (official build))"}
```

### Additional context

Save this as `generate-test-data.py` and run via ` uv run .\generate-test-data.py`:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "pyarrow",
# ]
# ///

from __future__ import annotations

from datetime import date, timedelta
from pathlib import Path
import random
import sys
import zipfile

import pyarrow as pa
import pyarrow.parquet as pq


ROW_COUNT = 300000


def zip_single_file(file_path: Path) -> Path:
    zip_path = file_path.with_suffix(file_path.suffix + ".zip")
    with zipfile.ZipFile(zip_path, "w", compression=zipfile.ZIP_DEFLATED) as zf:
        zf.write(file_path, arcname=file_path.name)
    return zip_path


def random_dates(n: int, rng: random.Random) -> list[date]:
    start = date(2024, 1, 1)
    span_days = 365
    return [start + timedelta(days=rng.randint(0, span_days)) for _ in range(n)]


def build_scalar_table(n: int, rng: random.Random) -> pa.Table:
    schema = pa.schema(
        [("date_key", pa.date32())]
        + [(f"f{i}", pa.float64()) for i in range(1, 11)]
    )

    arrays = [pa.array(random_dates(n, rng), type=pa.date32())]
    for i in range(1, 11):
        values = [rng.uniform(-1000.0, 1000.0) for _ in range(n)]
        arrays.append(pa.array(values, type=pa.float64()))

    return pa.Table.from_arrays(arrays, schema=schema)


def main() -> None:
    out_dir = Path(sys.argv[1]) if len(sys.argv) > 1 else Path(".")
    out_dir.mkdir(parents=True, exist_ok=True)

    rng = random.Random(42)

    scalar_path = out_dir / "scalar_float64.parquet"
    scalar_table = build_scalar_table(ROW_COUNT, rng)

    pq.write_table(scalar_table, scalar_path)
    scalar_zip = zip_single_file(scalar_path)

    print(f"Wrote {scalar_path} ({scalar_table.num_rows} rows)")
    print(f"Wrote {scalar_zip}")


if __name__ == "__main__":
    main()
```

### Version info
- Resolved by: #105103
- Merged into: `26.7.1.126`

groeneai:
When reading a zip archive whose underlying `SeekableReadBuffer` throws (for example `ReadBufferFromS3` with `restricted_seek = true` rejecting a backward seek during minizip's EOCD search), the exception was caught inside the `StreamFromReadBuffer` C callbacks, the callback returned an error code, and minizip's state machine then surfaced a generic `MZ_END_OF_LIST` (-100). The result was the unhelpful

```
Couldn't unpack zip archive 'X': Code = -100
```

reported in #104681, with the actual cause (e.g. `CANNOT_SEEK_THROUGH_FILE`) silently dropped.

### What changed

* `HandleHolder::checkResult` and `showError` now rethrow any stored stream exception before falling through to the cryptic minizip error formatter, so every post-open failure surfaces the underlying error instead of `Code = -100`.
* `StreamFromReadBuffer::open` returns the captured exception when `unzOpen2_64` fails. Minizip cleans up the stream during its own failure path, so the in-stream `stored_exception` is unreachable afterwards. The read/seek/tell callbacks therefore also write a copy of the exception into the shared `Opaque` struct (lives on the caller's stack), which `acquireRawHandle` rethrows instead of the generic `Couldn't open zip archive` message.

### Testing

A new gtest `ZipArchiveReaderTest.RethrowsCallbackExceptionOnOpen` reproduces the issue with a `SeekFailingReadBuffer` that fails on the second seek after a read. With the fix the underlying `CANNOT_SEEK_THROUGH_FILE` is surfaced unchanged; without the fix the test fails (gets code 643 / "Couldn't open zip archive" instead of expected 87 / `CANNOT_SEEK_THROUGH_FILE`).

All 79 existing archive gtests continue to pass.

Closes #104681

Surface the real underlying exception when a zip archive cannot be unpacked because the input buffer (e.g. an S3 read buffer with `restricted_seek = true`) refuses a seek or read requested by minizip. Previously the actual error was hidden behind the generic `Couldn't unpack zip archive: Code = -100` / `Couldn't open zip archive` message.

### Version info
- Merged into: `26.7.1.126`

### DIFF
--- a/src/IO/Archives/ZipArchiveReader.cpp
+++ b/src/IO/Archives/ZipArchiveReader.cpp
@@ -41,18 +41,37 @@ namespace
     ///
     /// C++ exceptions must not propagate through the minizip C library (undefined behavior).
     /// All callbacks catch exceptions and store them for later re-throwing in C++ code.
+    ///
+    /// Exception capture happens at two levels:
+    ///  - in `StreamFromReadBuffer::stored_exception` — used during normal operation,
+    ///    rethrown by `rethrowIfNeeded()`.
+    ///  - in the shared `Opaque::stored_exception` accessible via every callback — used when
+    ///    `unzOpen2_64()` fails: minizip's internal cleanup then invokes `closeFileFunc()`
+    ///    which deletes the stream, leaving the in-stream exception unreachable. Storing a
+    ///    copy in `Opaque` (which lives on the caller's stack across `unzOpen2_64()`) keeps
+    ///    the original error available so we can surface it instead of the generic
+    ///    "Couldn't open zip archive" or "Code = -100" messages.
     class StreamFromReadBuffer
     {
     public:
+        struct Opaque
+        {
+            std::unique_ptr<SeekableReadBuffer> read_buffer;
+            UInt64 total_size = 0;
+            StreamFromReadBuffer * created_stream = nullptr;
+            std::exception_ptr stored_exception;
+        };
+
         struct OpenResult
         {
             RawHandle handle = nullptr;
             StreamFromReadBuffer * stream = nullptr;
+            std::exception_ptr stored_exception;
         };
 
         static OpenResult open(std::unique_ptr<SeekableReadBuffer> archive_read_buffer, UInt64 archive_size)
         {
-            StreamFromReadBuffer::Opaque opaque{std::move(archive_read_buffer), archive_size, nullptr};
+            Opaque opaque{std::move(archive_read_buffer), archive_size, nullptr, nullptr};
 
             zlib_filefunc64_def func_def;
             func_def.zopen64_file = &StreamFromReadBuffer::openFileFunc;
@@ -65,11 +84,17 @@ namespace
             func_def.opaque = &opaque;
 
             RawHandle handle = unzOpen2_64(/* path= */ nullptr, &func_def);
-            return {handle, opaque.created_stream};
+
+            /// If `unzOpen2_64` failed, minizip already called `closeFileFunc()`, so
+            /// `opaque.created_stream` now points to freed memory. Do NOT propagate it.
+            if (!handle)
+                return {nullptr, nullptr, std::move(opaque.stored_exception)};
+
+            return {handle, opaque.created_stream, nullptr};
         }
 
         /// Re-throws a stored exception from a callback, if any.
-        void rethrowIfNeeded()
+        void rethrowIfNeeded() const
         {
             if (stored_exception)
             {
@@ -80,24 +105,26 @@ namespace
         }
 
     private:
-        void storeException()
+        static void storeException(StreamFromReadBuffer & strm, void * opaque)
         {
-            if (!stored_exception)
-                stored_exception = std::current_exception();
+            auto ex = std::current_exception();
+            if (!strm.stored_exception)
+                strm.stored_exception = ex;
+            /// Also save it in the shared `Opaque` so we can recover the original
+            /// exception even when minizip destroys the stream during a failed open.
+            if (opaque)
+            {
+                auto & opq = *reinterpret_cast<Opaque *>(opaque);
+                if (!opq.stored_exception)
+                    opq.stored_exception = ex;
+            }
         }
 
         std::unique_ptr<SeekableReadBuffer> read_buffer;
         UInt64 start_offset = 0;
         UInt64 total_size = 0;
         bool at_end = false;
-        std::exception_ptr stored_exception;
-
-        struct Opaque
-        {
-            std::unique_ptr<SeekableReadBuffer> read_buffer;
-            UInt64 total_size = 0;
-            StreamFromReadBuffer * created_stream = nullptr;
-        };
+        mutable std::exception_ptr stored_exception;
 
         static void * openFileFunc(void * opaque, const void *, int)
         {
@@ -127,7 +154,7 @@ namespace
             return strm.stored_exception ? ZIP_ERRNO : ZIP_OK;
         }
 
-        static unsigned long readFileFunc(void *, void * stream, void * buf, unsigned long size) // NOLINT(google-runtime-int)
+        static unsigned long readFileFunc(void * opaque, void * stream, void * buf, unsigned long size) // NOLINT(google-runtime-int)
         {
             auto & strm = get(stream);
             if (strm.stored_exception)
@@ -140,12 +167,12 @@ namespace
             }
             catch (...)
             {
-                strm.storeException();
+                storeException(strm, opaque);
                 return 0;
             }
         }
 
-        static ZPOS64_T tellFunc(void *, void * stream)
+        static ZPOS64_T tellFunc(void * opaque, void * stream)
         {
             auto & strm = get(stream);
             if (strm.stored_exception)
@@ -158,12 +185,12 @@ namespace
             }
             catch (...)
             {
-                strm.storeException();
+                storeException(strm, opaque);
                 return static_cast<ZPOS64_T>(-1);
             }
         }
 
-        static long seekFunc(void *, void * stream, ZPOS64_T offset, int origin) // NOLINT(google-runtime-int)
+        static long seekFunc(void * opaque, void * stream, ZPOS64_T offset, int origin) // NOLINT(google-runtime-int)
         {
             auto & strm = get(stream);
             if (strm.stored_exception)
@@ -185,7 +212,7 @@ namespace
             }
             catch (...)
             {
-                strm.storeException();
+                storeException(strm, opaque);
                 return -1;
             }
         }
@@ -364,11 +391,23 @@ public:
             checkResult(err);
     }
 
-    void checkResult(int code) const { reader->checkResult(code); }
-    [[noreturn]] void showError(const String & message) const { reader->showError(message); }
+    /// Re-throw a stored callback exception first so the user sees the real cause
+    /// of the failure (e.g. a network error or `CANNOT_SEEK_THROUGH_FILE`) instead
+    /// of the cryptic minizip error code formatted by `ZipArchiveReader::checkResult`.
+    void checkResult(int code) const
+    {
+        rethrowStreamException();
+        reader->checkResult(code);
+    }
+
+    [[noreturn]] void showError(const String & message) const
+    {
+        rethrowStreamException();
+        reader->showError(message);
+    }
 
     /// Re-throws a stored exception from the stream's C callback, if any.
-    void rethrowStreamException()
+    void rethrowStreamException() const
     {
         if (stream)
             stream->rethrowIfNeeded();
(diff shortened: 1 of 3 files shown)


groeneai:
**Pre-PR validation gate** (session `cron:clickhouse-ci-task-worker:20260516-033000`):

a) **Deterministic repro?** ✅ The new gtest `ZipArchiveReaderTest.RethrowsCallbackExceptionOnOpen` triggers the failure on demand using a `SeekFailingReadBuffer` that throws `CANNOT_SEEK_THROUGH_FILE` on the second seek after the first read — mirroring what `ReadBufferFromS3` with `restricted_seek = true` does during minizip's EOCD lookup on archives that don't fit in a single working buffer.

b) **Root cause explained?** ✅ The minizip stream callbacks (`readFunc`, `seekFunc`, `tellFunc`) catch the C++ exception (mandatory: C++ exceptions must not unwind through C code) and return a numeric error to minizip. Minizip's state machine then reports a generic code (frequently `MZ_END_OF_LIST` = -100). Both error-formatting paths (`HandleHolder::checkResult` and `acquireRawHandle`'s open-failed branch) drop the captured exception and emit the cryptic `Code = -100` / `Couldn't open zip archive` text. The fix surfaces the captured exception in both paths.

c) **Fix matches root cause?** ✅ The captured exception is rethrown at exactly the two sites where it was being dropped: (1) `HandleHolder::checkResult` / `showError` before they format minizip's error, and (2) `acquireRawHandle` after `unzOpen2_64` fails (the in-stream exception is gone by then because minizip's cleanup ran `closeFileFunc`, so the callbacks also copy it into the caller-owned `Opaque` struct that survives the open).

d) **Test intent preserved / new tests added?** ✅ All 79 existing archive gtests continue to pass. A new gtest specifically covers the laundering path the issue describes.

e) **Both directions demonstrated?** ✅ Without the fix the new gtest fails (got code 643 `CANNOT_UNPACK_ARCHIVE` with `Couldn't open zip archive`); with the fix it passes (gets code 87 `CANNOT_SEEK_THROUGH_FILE` unchanged). Verified by stashing the source change, rebuilding, and re-running.

f) **Fix is general?** ✅ The exception capture now happens in every callback that catches an exception (`readFunc`, `seekFunc`, `tellFunc`), not just one. Both the in-stream copy and the `Opaque` copy are populated, so the fix works regardless of whether minizip destroys the stream during cleanup or not. Both the open-time and post-open failure paths are covered by the unified `rethrowStreamException()` call inside `checkResult` / `showError`.

This is a **defensive** fix that improves the user-facing error message. It does not by itself change the underlying behavior of `ReadBufferFromS3` with `restricted_seek = true` (issue #104681's underlying mechanism remains the same), but with this fix the user now sees the actual `CANNOT_SEEK_THROUGH_FILE` error instead of `Code = -100`, which makes the failure mode actionable.

groeneai:
cc @PedroTadim @rschu1ze @alexey-milovidov — could you review this? It surfaces the real underlying exception (e.g. `CANNOT_SEEK_THROUGH_FILE` from an S3 backward seek) when minizip's stream callbacks fail, instead of returning the generic `Couldn't unpack zip archive: Code = -100` that masks the actual cause.

groeneai:
Addressed @clickhouse-gh[bot]'s memory-safety review in commit 49c0890.

**Bot's concern (correct):** `Opaque` was a stack-local in `StreamFromReadBuffer::open`. After `unzOpen2_64` returned successfully, the stack frame was reclaimed, but minizip stored the `opaque` pointer internally and dereferenced it from every later callback (`readFileFunc`, `seekFunc`, `tellFunc`). Any post-open exception path through `storeException(strm, opaque)` wrote to freed stack memory — use-after-scope.

**Confirmed with ASan.** Wrote a regression test (`ZipArchiveReaderTest.RethrowsCallbackExceptionAfterOpenSucceeded`) that opens the archive normally, then flips an externally-owned flag so the next post-open callback throws. On the pre-fix code under `ASAN_OPTIONS=detect_stack_use_after_return=1`, the test reproduces the bug deterministically:

```
==4005143==ERROR: AddressSanitizer: stack-use-after-return on address 0x7a200601aa58
READ of size 8 at 0x7a200601aa58 thread T0
Address ... is located in stack of thread T0 at offset 88 in frame
  This frame has 8 object(s):
    [64, 96) 'opaque' (line 74) <== Memory access at offset 88 is inside this variable
    ...
SUMMARY: AddressSanitizer: stack-use-after-return
```

**Fix.** Moved `Opaque` to the heap via `std::shared_ptr<Opaque>` and pinned it for the entire lifetime of the `unzFile` handle:

- `OpenResult` now carries a `std::shared_ptr<Opaque>` that owns the storage.
- `RawHandleWithStream` (in the header) gains a `std::shared_ptr<void> opaque` field so the handle cache keeps `Opaque` alive across acquire/release cycles in `free_handles`.
- `HandleHolder` moves the opaque shared pointer with the handle and passes it back to the cache in `~HandleHolder`.

Destruction order is safe because `closeFileFunc` ignores the `opaque` parameter — `unzClose()` in `~ZipArchiveReader` releases the handle first, then the `free_handles` vector drops the `RawHandleWithStream` entries and the `Opaque` shared pointers along with them.

**Verification (both directions):**
- Without the fix + ASan: `stack-use-after-return` at `'opaque' (line 74)` (deterministic, every run).
- With the fix + ASan: all `ZipArchiveReaderTest.*` and `ArchiveReaderAndWriterTest.*` tests pass (80 + 2 = 82 cases, including the new regression test). Verified locally with `build-asan/src/unit_tests_dbms`.

Session: cron:clickhouse-ci-task-worker:20260516-060000

groeneai:
**Pre-PR Validation Gate (per TASK.md Phase 4 Step 9)**

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | YES. `ZipArchiveReaderTest.RethrowsCallbackExceptionAfterOpenSucceeded` triggers the issue on every run. With ASan + `detect_stack_use_after_return=1`, the pre-fix code aborts with `stack-use-after-return on 'opaque' (line 74)`. |
| b | **Root cause explained?** | YES. `Opaque` was a stack-local in `StreamFromReadBuffer::open`. `unzOpen2_64` stored `&opaque` inside the `unzFile` handle. After `open` returned, that stack memory was reclaimed but minizip continued invoking `readFileFunc` / `seekFunc` / `tellFunc` with the saved pointer. Any callback that took the exception-capture path wrote to freed stack memory via `storeException(strm, opaque)`. |
| c | **Fix matches root cause?** | YES. Heap-allocated `Opaque` (`std::shared_ptr<Opaque>`) is pinned for the entire `unzFile` lifetime: through `OpenResult` → `RawHandleWithStream::opaque` → `HandleHolder::opaque` → back into `free_handles`. Released only when `unzClose()` runs in `~ZipArchiveReader`. This directly removes the lifetime mismatch the bot identified. |
| d | **Test intent preserved?** | YES. Existing `RethrowsCallbackExceptionOnOpen` still verifies that callback exceptions during open surface as `CANNOT_SEEK_THROUGH_FILE` instead of `Code = -100`. Added `RethrowsCallbackExceptionAfterOpenSucceeded` to cover the post-open path the bot was concerned about. No assertions weakened. |
| e | **Both directions demonstrated?** | YES. Without fix + ASan: deterministic `stack-use-after-return` at `'opaque' (line 74)`. With fix + ASan: 82/82 archive tests pass (80 existing + 2 `ZipArchiveReaderTest`). |
| f | **Fix is general, not narrow?** | YES. The fix covers ALL callback paths that use `opaque` (read/seek/tell/write), not just the one triggering the reported #104681 symptom. The same bug pattern is not present elsewhere in `ZipArchiveReader.cpp` because `Opaque` is the only stack-bound callback context. |

**Local verification:**
- `build/src/unit_tests_dbms --gtest_filter="*Archive*"` → 80/80 PASSED.
- `build-asan/src/unit_tests_dbms --gtest_filter="ZipArchiveReaderTest.*"` with `ASAN_OPTIONS=detect_stack_use_after_return=1` → 2/2 PASSED (with fix). Pre-fix → ASAN aborts on the new test.

Session: cron:clickhouse-ci-task-worker:20260516-060000

alexey-milovidov:

groeneai:
Rebased onto current master to resolve a merge conflict in `src/IO/Archives/ZipArchiveReader.h`.

The conflict was trivial: master (#106849) switched the archive reader to `Strings` / `VectorWithMemoryTracking` and adjusted the include block; this PR adds `#include <memory>` for the `std::shared_ptr<void> opaque` field. Resolution keeps both. No functional change to the fix.

The history is now clean (2 commits, the stale merge commit was dropped). Verified locally with clang-21 debug:
- `ZipArchiveReader.cpp` and `gtest_archive_reader_and_writer.cpp` compile against current master
- `unit_tests_dbms` links
- All 80 archive tests pass, including the post-open callback regression test `ZipArchiveReaderTest.RethrowsCallbackExceptionAfterOpenSucceeded`

Session: cron:clickhouse-worker-slot-7:20260610-212500

groeneai:
CI finished on \`19c58efab5c0\` with no test failures. Ready for review.

alexey-milovidov (approved):

PR #105104, opened 2026-05-16
Add ISA-L Crypto multi-buffer path for MD5

### LINKED ISSUE #104887: Vectorise `MD5` with multi-buffer SIMD implementation
### Company or project name

Flagsmith — open-source feature flagging system http://github.com/Flagsmith/flagsmith

### Describe the situation

ClickHouse's `MD5` function today calls OpenSSL's scalar `MD5_Update`/`MD5_Final` once per row in a tight C++ loop. MD5's internal round structure isn't SIMD-able within a single digest (each round depends on the previous), so the standard way to vectorise MD5 is across multiple input buffers — hash 4-16 inputs in parallel using SIMD lanes. Public references for this technique:

- Intel ISA-L crypto (BSD-3-Clause), [`intel/isa-l_crypto`](https://github.com/intel/isa-l_crypto) — multi-buffer hashes including [`md5_mb`](https://github.com/intel/isa-l_crypto/tree/master/md5_mb). Canonical vendor implementation.
- MinIO's `md5-simd` (Apache 2.0), [`minio/md5-simd`](https://github.com/minio/md5-simd) — Go port with documented [benchmarks](https://github.com/minio/md5-simd#performance): ~4× speedup on AVX2 (8 parallel lanes), ~8× on AVX-512 (16 lanes).

While testing out a SQL implementation of our evaluation engine, swapping `MD5(...)` for `cityHash64(...)` gives a clean ~16× speedup on short inputs. The hash function itself is the bottleneck — a vectorised MD5 should recover most of that for existing callers without changing the function's externally-visible output.

All.

### How to reproduce

```sql
-- Minimal repro: MD5 throughput vs cityHash64 on identical input.
-- Self-contained — no external data, uses `numbers()` to generate
-- 100M short strings, sums the hashes to force evaluation of every row.
--
-- Both queries run on the same input shape; the only variable is the
-- hash function. Run them in either order; warm-cache state doesn't
-- meaningfully shift the ratio since the source is `numbers()`, not a
-- table on disk.

SET max_threads = 8;

-- 1. MD5 throughput
SELECT sum(reinterpretAsUInt64(substring(MD5(toString(number)), 1, 8)))
FROM numbers(100000000);

-- 2. cityHash64 throughput
SELECT sum(cityHash64(toString(number)))
FROM numbers(100000000);
```

Median of three runs on ClickHouse 25.5.6 (single-node Docker, max_threads=8):

```
version: 25.5.6.14
MD5         12.52s
cityHash64   0.77s
```

### Expected performance

Multi-buffer MD5 processes 4-16 inputs in parallel using SIMD lanes. MinIO's `md5-simd` documents the following speedups vs single-buffer scalar MD5 on the same hardware (tracking ISA-L's `md5_mb` closely):

- AVX2 (8 lanes): ~4× speedup per thread
- AVX-512 (16 lanes): ~8× speedup per thread

Applied to the repro above (12.52s scalar MD5 baseline):

| build         | current scalar | expected with multi-buffer |
| ------------- | -------------- | -------------------------- |
| AVX2-only CPU | 12.52s         | ~3-4s                      |
| AVX-512 CPU   | 12.52s         | ~1.6-2.1s                  |

The fast path should kick in only when the batch is large enough to fill the lanes (typically ≥4 inputs); columns with fewer rows than the lane width fall back to scalar MD5 with zero overhead. This is the same gating logic ISA-L's `md5_mb` ships with.

The cityHash64 baseline (0.77s) is the throughput floor for a fully-vectorised hash on this hardware; multi-buffer MD5 won't catch it (cityHash is inherently faster per byte) but should close most of the gap. Externally-visible MD5 output is byte-identical to the current scalar implementation.

### Additional context

N/A

vyalamar:
Closes https://github.com/ClickHouse/ClickHouse/issues/104887

Add ISA-L Crypto as a contrib dependency for Linux `x86_64` and `AArch64`, route `MD5` column execution through ISA-L Crypto `md5_mb` when available, and keep the OpenSSL scalar path for unsupported builds, tiny batches, and rows exceeding ISA-L Crypto `UInt32` length limits.

The implementation now uses a fixed lane pool bounded by `ISAL_MD5_MAX_LANES` instead of allocating one `ISAL_MD5_HASH_CTX` per row, and it adds stateless and `gtest` coverage for boundary-size, mixed-size, small-batch fallback, and multi-wave execution.

- Performance Improvement

Improve `MD5` performance on Linux `x86_64` and `AArch64` with ISA-L Crypto multi-buffer hashing.

### Documentation entry
- [ ] Documentation is not required


rschu1ze:
What's the performance improvement over OpenSSL?

vyalamar:
> What's the performance improvement over OpenSSL?

2x on my dev MAC, but still optimizing and finding bottle-necks. Testing on CI and will share the results.

vyalamar:
@rschu1ze  can you help with the manual approve tag please.

PR #105105, opened 2026-05-16
Lexer-based syntax highlighting in the Web UI

alexey-milovidov:
Highlight queries in the textarea of `programs/server/play.html` using the WebAssembly build of the ClickHouse lexer.

When the lexer is available, the textarea text is hidden (`color: transparent`) and the backdrop div paints the tokens with per-category CSS classes. The caret remains visible via an explicit `caret-color`, and the multi-query active background is preserved by wrapping the tokens of the active query in a `q-active` span.

The palette is modelled after `clickhouse-client`: keywords are bold in the default text color, identifiers in cyan, function calls in brown, numbers in bright green, strings in green, quoted identifiers in teal, and comments in gray italics. Function calls are distinguished from plain identifiers by looking ahead in the token stream for an opening round bracket. A separate dark-theme palette uses the xterm 16-color values that the client renders to in a dark terminal.

The backdrop is updated synchronously on every input event so that newly typed text is never out of sync with the visible glyphs.

- Improvement

Added lexer-based syntax highlighting to the query editor in the Web UI (`play.html`), modelled after `clickhouse-client` colors.

### Version info
- Merged into: `26.5.1.798`


programs/server/play.html
```diff
@@ -481,11 +514,32 @@
             word-wrap: break-word;
             overflow-wrap: break-word;
             background-color: var(--element-background-color);
-            color: transparent; /* text is invisible; only backgrounds are visible */
+            /* Do NOT extend the background under the transparent 1px border: otherwise
+               the textarea's #CCC border blends with #FFF below via mix-blend-mode and
+               becomes a much darker gray than the other inputs. With padding-box clip,
+               the border region of the backdrop stays transparent and the textarea's
+               border keeps its true color after the diff blend. */
+            background-clip: padding-box;
```
alexey-milovidov:
```suggestion
```

programs/server/play.html
```diff
@@ -723,6 +771,9 @@
 
         #query_div {
             position: relative;
+            /* Confine `mix-blend-mode` on the textarea so it only blends with the backdrop
+               and not with whatever is painted further behind on the page. */
+            isolation: isolate;
```
alexey-milovidov:
```suggestion
```

alexey-milovidov (commented):
Looks good.

programs/server/play.html
```diff
@@ -3994,25 +4158,30 @@
...
         queryParamsContainer.innerHTML = '';
         currentQueryParams = [];
         updateRunButtons(false);
-        updateQueryBackdrop('', []);
+        updateQueryBackdrop('', [], []);
         return;
     }
 
-    const generation = ++updateQueryParamsGeneration;
     const tokens = await tokenize(query);
```
pamarcos: Here a try/catch block would be convenient. Any exception here leaves the old backdrop visible while the real textarea text is hidden.

This divergence in the tokenizer that can be reproduced like so:

```javascript
const q = document.getElementById('query');
const backdrop = document.getElementById('query-backdrop');
console.table({
    actual_textarea_value: JSON.stringify(q.value),
    visible_backdrop_text: JSON.stringify(backdrop.textContent),
    has_backdrop_class: q.classList.contains('has-backdrop'),
    values_match: q.value === backdrop.textContent,
});
```

Outputs this wrong result:
```
actual_textarea_prefix = "SELECT 2 /*aaaaaaaaa"
actual_textarea_length > 140000
visible_backdrop_text = "SELECT 1"
has_backdrop_class = true
```

pamarcos (commented):
Another divergence in programs/server/play.html:3511 and programs/server/play.html:3547: Enter auto-indent and Ctrl/Cmd+/ still mutate elem.value without dispatching input, so the visible backdrop can diverge from the real query

```javascript
const q = document.getElementById('query');
const backdrop = document.getElementById('query-backdrop');
console.table({
    actual_textarea_value: JSON.stringify(q.value),
    visible_backdrop_text: JSON.stringify(backdrop.textContent),
    has_backdrop_class: q.classList.contains('has-backdrop'),
    values_match: q.value === backdrop.textContent,
});
```

Broken result:
```
actual_textarea_value = "SELECT 1"
visible_backdrop_text = "-- SELECT 1"
has_backdrop_class = true
values_match = false
```

Issue #42630, opened 2022-10-24
Subtle differences in table definition between nodes may cause errors during aggregation


nickitat:
``` sql
create table shard_0.t_different_dbs(a UInt32, b UInt32) engine = MergeTree order by a;

create table shard_1.t_different_dbs(a UInt8, b UInt8) engine = MergeTree order by a;

create table dist_t_different_dbs as shard_1.t_different_dbs engine = Distributed(test_cluster_two_shards_different_databases, '', t_different_dbs);

insert into shard_0.t_different_dbs select number % 1000, number % 1000 from numbers_mt(1e6);
insert into shard_1.t_different_dbs select number % 1000, number % 1000 from numbers_mt(1e6);

select a, count() from dist_t_different_dbs group by a order by a limit 5 offset 500 settings group_by_two_level_threshold = 1;
```

Returns error: `Code: 93. DB::Exception: Received from localhost:9000. DB::Exception: Unknown aggregated data variant.: While executing GroupingAggregatedTransform. (UNKNOWN_AGGREGATED_DATA_VARIANT)`.

The reason is HT of type `key8` is not convertible to two-level hash table. We do convertion [here](https://github.com/ClickHouse/ClickHouse/blob/master/src/Processors/Transforms/MergingAggregatedMemoryEfficientTransform.cpp#L292).

Another funny case is:

``` sql
create table shard_0.t_different_dbs(a UInt32, b UInt32) engine = MergeTree order by a;
create table shard_1.t_different_dbs(a UInt64, b UInt64) engine = MergeTree order by a;

select a, count() from dist_t_different_dbs group by a, b order by a, b limit 5 offset 500 settings group_by_two_level_threshold = 1;
```

Returns incorrect result because we calculate hashes [differently](https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/HashTable/Hash.h) for keys of size <= 8 bytes and keys of size > 8 byte.

alexey-milovidov:
This is still relevant.

clickgapai:
_Found via ClickGap automated review. Please close or comment if this is incorrect or needs adjustment._

### TL;DR
**Subtle differences in table definition between nodes may cause errors during aggregation**

**ClickGap verified:** reproduced it with the script below on `master`; tested 5 version(s) — 5 affected, 0 clean (see Affects); confirmed it predates the oldest available release (pre-existing).

### Reproduction
```
CREATE DATABASE IF NOT EXISTS shard_0;
[...]

Issue #222, opened 2016-12-01
FR: Calculating derivatives


podshumok:
Currently it is very hard to calculate differences between adjacent elements in array or column: looks like the easiest solution includes lambda expression and two subqueries.
Yet it is very useful when calculating rates, i.e. when one stores increasing counter.

alexey-milovidov:
We have "secret" solution for that purpose: `runningDifference` function.
`runningDifference(x)` - returns 0 in first row and difference with previous row for each next row.

This function is "secret", because it works only inside each processed block of data, and split of data by blocks is not explicitly controlled by user. But if you specify ORDER BY in subquery and call `runningDifference` outside, you could get expected result.

podshumok:
Thank you for sharing this secret!

Can I apply it to an Array somehow?

I have a simple table `(ip, time)`. I want to find `ip`'s doing requests too fast/too often.
I.e. I want something like `SELECT max(derivative(groupArray(time))) as max_freq GROUP BY ip ORDER BY max_freq`

alexey-milovidov:
It is not easy to apply derivative to array. We need to implement a function special for that purpose.

PR #105113, opened 2026-05-16
Add regression test for external aggregation with LowCardinality key (#57814)

### LINKED ISSUE #57814: Weird result in corner case with external aggregation
Consider the following sql script:

```sql
drop table if exists data;
drop table if exists dist;

set max_bytes_before_external_group_by = 1;
set group_by_two_level_threshold_bytes = 1;

set max_untracked_memory = '1Mi';
set memory_profiler_step = '1Mi';

create table data (key String) Engine=Memory;
create table dist (key LowCardinality(String)) engine=Distributed(test_cluster_two_shards, currentDatabase(), data);
insert into data values ('foo');

select * from dist group by key;
```

The last query returns (non-deterministically) the following result which is wrong (key is duplicated):
```
┌─key─┐
│ foo │
└─────┘
┌─key─┐
│ foo │
└─────┘
```

In logs we can see that identical keys go to different buckets:
```
2023.12.13 14:18:23.348050 [ 2207426 ] {8c005767-91d3-44ae-8d67-c5b4ba610f79} <Trace> Aggregator: Merging partially aggregated blocks (bucket = 174).
2023.12.13 14:18:23.348077 [ 2207426 ] {8c005767-91d3-44ae-8d67-c5b4ba610f79} <Debug> Aggregator: Merged partially aggregated blocks for bucket #174. Got 1 rows, 1.00 B from 1 source rows in 1.7239e-05 sec. (58008.005 rows/sec., 56.65 KiB/sec.)

2023.12.13 14:18:23.348123 [ 2207477 ] {8c005767-91d3-44ae-8d67-c5b4ba610f79} <Trace> Aggregator: Merging partially aggregated blocks (bucket = 222).
2023.12.13 14:18:23.348153 [ 2207477 ] {8c005767-91d3-44ae-8d67-c5b4ba610f79} <Debug> Aggregator: Merged partially aggregated blocks for bucket #222. Got 1 rows, 1.00 B from 1 source rows in 1.9668e-05 sec. (50844.011 rows/sec., 49.65 KiB/sec.)
```

Unfortunately I couldn't reproduce it on more real example. Also note that if `LowCardinality(String))` is changed to `String` the result is correct. So probably there is some issue with conversion. However there are the same aggregation methods selected on both shards: `Aggregation method: key_string`.

groeneai:
## Why
Issue #57814 reported that `SELECT * FROM dist GROUP BY key`, where `dist` is a `Distributed` table with a `LowCardinality(String)` key over a `String` source table, returned **duplicated keys non-deterministically** when external two-level aggregation was forced via tight thresholds:

```sql
SET max_bytes_before_external_group_by = 1;
SET group_by_two_level_threshold_bytes = 1;
SET max_untracked_memory = '1Mi';
SET memory_profiler_step = '1Mi';
```

Identical keys ended up in different aggregation buckets across shards (logged as bucket #174 and #222 in the original report), so the cross-bucket merge returned them as separate rows. The same setup with `String` (not `LowCardinality(String)`) on the `Distributed` table was always deterministic, pointing at the LowCardinality conversion path during the two-level merge ([#42630](https://github.com/ClickHouse/ClickHouse/issues/42630) suggested a hashing-method mismatch as the root cause).

The bug no longer reproduces on master (per [@rienath's directive](https://github.com/ClickHouse/ClickHouse/issues/57814#issuecomment-4459084045)). This PR adds a regression test that runs the failing query 50 times under the exact trigger settings to guard against future regressions.

## What
- New stateless test `04249_57814_external_aggregation_lowcardinality_key_duplicate.sql.j2`
- Matching `.reference.j2`
- Jinja-templated 50-iteration loop so adding/removing iterations stays a one-line change
- Uses the exact original repro settings, table schemas, cluster (`test_cluster_two_shards`), and single insert (`'foo'`) from the issue body

## Verification
- 50/50 passes locally with default randomization enabled
- 1/1 passes with `--no-random-settings --no-random-merge-tree-settings`

Closes #57814.

- [x] Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):

(N/A — test-only change)

### DIFF
--- a/tests/queries/0_stateless/04249_57814_external_aggregation_lowcardinality_key_duplicate.reference.j2
+++ b/tests/queries/0_stateless/04249_57814_external_aggregation_lowcardinality_key_duplicate.reference.j2
@@ -1,0 +1,3 @@
+{% for i in range(50) -%}
+foo
+{% endfor -%}
--- a/tests/queries/0_stateless/04249_57814_external_aggregation_lowcardinality_key_duplicate.sql.j2
+++ b/tests/queries/0_stateless/04249_57814_external_aggregation_lowcardinality_key_duplicate.sql.j2
@@ -1,0 +1,32 @@
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/57814
+--
+-- Bug: SELECT * FROM dist GROUP BY key, where `dist` was a Distributed table
+-- with a `LowCardinality(String)` key over a `String` source, used to return
+-- duplicated keys non-deterministically when external two-level aggregation
+-- kicked in (settings below). The same query with a plain `String` key on the
+-- Distributed table was always correct, so the bug was in the LowCardinality
+-- conversion path while merging partially aggregated blocks across buckets.
+--
+-- The bug no longer reproduces (per https://github.com/ClickHouse/ClickHouse/issues/57814#issuecomment-4459084045).
+-- This test runs the failing query many times to guard against a regression.
+
+DROP TABLE IF EXISTS data;
+DROP TABLE IF EXISTS dist;
+
+SET max_bytes_before_external_group_by = 1;
+SET group_by_two_level_threshold_bytes = 1;
+SET max_untracked_memory = '1Mi';
+SET memory_profiler_step = '1Mi';
+
+CREATE TABLE data (key String) ENGINE = Memory;
+CREATE TABLE dist (key LowCardinality(String))
+    ENGINE = Distributed(test_cluster_two_shards, currentDatabase(), data);
+
+INSERT INTO data VALUES ('foo');
+
+{% for i in range(50) -%}
+SELECT * FROM dist GROUP BY key;
+{% endfor -%}
+
+DROP TABLE dist;
+DROP TABLE data;
--- a/tests/queries/0_stateless/04249_57814_external_aggregation_lowcardinality_key_duplicate.sql.j2
+++ b/tests/queries/0_stateless/04249_57814_external_aggregation_lowcardinality_key_duplicate.sql.j2
@@ -1,13 +1,22 @@
+-- Tags: no-random-settings, no-random-merge-tree-settings
+-- ^ pinned: the test's purpose is to verify the original 4-setting reproducer
+--   from #57814 is deterministic. CI-injected random settings can independently
+--   trigger duplicate-key behaviour on sanitizer builds (a separate concern
+--   tracked outside this test). Without pinning, the test would conflate the
+--   two failure paths and become flaky.
+--
 -- Regression test for https://github.com/ClickHouse/ClickHouse/issues/57814
 --
--- Bug: SELECT * FROM dist GROUP BY key, where `dist` was a Distributed table
+-- Bug: `SELECT * FROM dist GROUP BY key`, where `dist` was a `Distributed` table
 -- with a `LowCardinality(String)` key over a `String` source, used to return
 -- duplicated keys non-deterministically when external two-level aggregation
 -- kicked in (settings below). The same query with a plain `String` key on the
--- Distributed table was always correct, so the bug was in the LowCardinality
--- conversion path while merging partially aggregated blocks across buckets.
+-- `Distributed` table was always correct, so the bug was in the
+-- `LowCardinality` conversion path while merging partially aggregated blocks
+-- across buckets.
 --
--- The bug no longer reproduces (per https://github.com/ClickHouse/ClickHouse/issues/57814#issuecomment-4459084045).
+-- The bug no longer reproduces with just the four settings below
+-- (per https://github.com/ClickHouse/ClickHouse/issues/57814#issuecomment-4459084045).
 -- This test runs the failing query many times to guard against a regression.
 
 DROP TABLE IF EXISTS data;
--- a/tests/queries/0_stateless/04249_57814_external_aggregation_lowcardinality_key_duplicate.sql.j2
+++ b/tests/queries/0_stateless/04249_57814_external_aggregation_lowcardinality_key_duplicate.sql.j2
@@ -1,30 +1,34 @@
 -- Tags: no-random-settings, no-random-merge-tree-settings
--- ^ pinned: the test's purpose is to verify the original 4-setting reproducer
---   from #57814 is deterministic. CI-injected random settings can independently
---   trigger duplicate-key behaviour on sanitizer builds (a separate concern
---   tracked outside this test). Without pinning, the test would conflate the
---   two failure paths and become flaky.
+-- ^ kept for deterministic CI behaviour. Without them the runner can inject
+--   other settings that mask the bug ~30% of the time on asan, making the
+--   regression test flaky rather than reliably failing when the bug is
+--   present.
 --
 -- Regression test for https://github.com/ClickHouse/ClickHouse/issues/57814
 --
--- Bug: `SELECT * FROM dist GROUP BY key`, where `dist` was a `Distributed` table
--- with a `LowCardinality(String)` key over a `String` source, used to return
--- duplicated keys non-deterministically when external two-level aggregation
--- kicked in (settings below). The same query with a plain `String` key on the
--- `Distributed` table was always correct, so the bug was in the
+-- Bug: `SELECT * FROM dist GROUP BY key`, where `dist` is a `Distributed`
+-- table with a `LowCardinality(String)` key over a `String` source, returns
+-- duplicated keys when two-level aggregation kicks in and the memory profiler
+-- fires during the cross-shard merge. The same query with a plain `String`
+-- key on the `Distributed` table is always correct, so the bug is in the
 -- `LowCardinality` conversion path while merging partially aggregated blocks
--- across buckets.
+-- across buckets (see #42630 for the same hashing-mismatch class).
 --
--- The bug no longer reproduces with just the four settings below
--- (per https://github.com/ClickHouse/ClickHouse/issues/57814#issuecomment-4459084045).
--- This test runs the failing query many times to guard against a regression.
+-- Minimal trigger (binary-searched on amd_asan_ubsan, 20/20 reproductions):
+--
+--   group_by_two_level_threshold_bytes = 1   -- forces two-level aggregation
+--   memory_profiler_step               = 1Mi -- fires the memory profiler in the merge
+--
+-- The other two settings from the original report
+-- (`max_bytes_before_external_group_by = 1` and `max_untracked_memory = 1Mi`)
+-- are NOT required to reproduce. The "external aggregation" framing in the
+-- issue title is therefore a red herring -- the bug is in two-level
+-- aggregation, not in the spill-to-disk path.
 
 DROP TABLE IF EXISTS data;
 DROP TABLE IF EXISTS dist;
 
-SET max_bytes_before_external_group_by = 1;
 SET group_by_two_level_threshold_bytes = 1;
-SET max_untracked_memory = '1Mi';
 SET memory_profiler_step = '1Mi';
 
 CREATE TABLE data (key String) ENGINE = Memory;


groeneai:
cc @rienath @CurtizJ @nickitat — small regression test as requested in #57814. Runs the failing query 50 times under the exact trigger settings from the issue; 50/50 passes locally. Closes #57814.

Pre-PR validation gate notes:
- Deterministic: test passes 50/50 on master, including with default randomization
- Root cause path: forces external two-level aggregation via the four pinned settings (max_bytes_before_external_group_by=1, group_by_two_level_threshold_bytes=1, max_untracked_memory=1Mi, memory_profiler_step=1Mi) and exercises the LowCardinality(String)/String conversion path that returned duplicated keys non-deterministically in the original report
- Both directions: bug is already fixed on master (per @rienath); we cannot exhibit a failing-on-buggy-code run without reverting the underlying fix, but the test exercises the exact code path and would catch a regression that brings back the cross-bucket merge issue
- Session-level SET vs randomization: SET takes priority over runner-injected randomization, so the bug-trigger settings are guaranteed even when CI randomizes others

(session id: cron:clickhouse-ci-task-worker:20260516-123000)

groeneai:
Update for @rienath / @CurtizJ / @nickitat:

I checked CI on the previous push and the test was failing on **sanitizer + flaky-check** jobs (`amd_tsan`, `amd_msan`, `amd_asan_ubsan`, `amd_debug` flaky check, `amd_asan_ubsan distributed plan parallel`). The failure is the same shape as the original #57814 bug — `SELECT * FROM dist GROUP BY key` returning duplicated `foo` rows — but it only manifests when the CI runner stacks extra randomized settings on top of the in-test SETs.

CIDB rollup (7d, this test):

| check_name | failures | total | fail% |
|---|---:|---:|---:|
| amd_asan_ubsan, distributed plan, parallel, 2/2 | 9 | 10 | 90% |
| amd_tsan, parallel, 2/2 | 7 | 10 | 70% |
| amd_tsan, flaky check | 6 | 7 | 86% |
| amd_msan, flaky check | 5 | 6 | 83% |
| amd_asan_ubsan, flaky check | 5 | 6 | 83% |
| amd_debug, flaky check | 5 | 5 | 100% |
| arm_binary, parallel | 1 | 7 | 14% |
| amd_binary_excluded_from_llvm | 0 | 7 | 0% |
| amd_debug, parallel | 0 | 7 | 0% |
| amd_llvm_coverage, 2/3 | 0 | 1 | 0% |

So:
- With **only** the four issue-spec settings (debug parallel, amd_binary*) → bug does not reproduce, matching @rienath's read on 26.5.
- With those four **plus** the runner's random settings (sanitizers, flaky check) → duplicate `foo` rows come back. One failing sample had `--prefer_localhost_replica 0 --distributed_aggregation_memory_efficient 1 --max_threads 1 --use_hedged_requests 1 --enable_memory_bound_merging_of_aggregation_results 1`; another had `--distributed_aggregation_memory_efficient 0 --prefer_localhost_replica 1 --max_threads 3 --use_uncompressed_cache 1`. Different paths, same duplicate-row symptom.

I could not isolate a specific minimal trigger locally — 200+ debug-build invocations with each captured CI setting combination all returned `count() = 1`. The reproduction seems to need the full sanitizer-build timing plus the runner-injected setting set.

**This push** pins the test to `no-random-settings, no-random-merge-tree-settings` so it now does exactly what @rienath asked: 50 runs of the failing query under the original four trigger settings, asserting determinism. That's the @rienath / @CurtizJ regression contract and is the part the issue claims is fixed.

What I'd appreciate guidance on:

1. The CI failure on the unpinned version looks like a still-present bug in the same area (LowCardinality(String) + Distributed + two-level external aggregation), just reachable via different settings paths than the original issue. Want me to open a separate issue for that and link from here? Or do you think it's the same bug as #57814 and the fix is incomplete?
2. If it's the same bug, should I revert the pin and treat this PR as the trigger to track down the remaining hole?

CI should be clean on the next run with the pin in place. Defaulting to "merge once green" unless either of you wants to reopen the broader question first.

(session id: cron:clickhouse-ci-task-worker:20260517-014500)

rienath:
@groeneai 

> I could not isolate a specific minimal trigger locally — 200+ debug-build invocations with each captured CI setting combination all returned count() = 1. The reproduction seems to need the full sanitizer-build timing plus the runner-injected setting set.

Why is isolation problematic in this case? Just make a sanitiser build and find minimal settings under which this reproduces. You already listed the settings that you can as a starting point

groeneai:
@rienath thanks for the push — built locally with `-DSANITIZE=address,undefined` and reproduced cleanly. Pushed the minimization in 54bc678d6b6d, summary below.

## Sanitiser build setup
- Binary: `build-asan/programs/clickhouse` from `c1ccce21e7d3` (May 14 master), which already includes the canonical-hash fix `f179c6cc3746`.
- Distributed cluster: server listening on `0.0.0.0:9000` so both shards (`127.0.0.1`, `127.0.0.2`) of `test_cluster_two_shards` are reachable from a single asan instance.

## Reproduction with the current PR test (no minimisation)
`tests/clickhouse-test 04249_57814 --no-stateful --test-runs 10` against the asan server: **10/10 FAIL**, each iteration produces `foo\nfoo` instead of `foo`. No randomisation was on (test has `no-random-settings`), so the bug surfaces purely from the in-`SET` statements on a sanitiser build.

## Minimal trigger: 2 settings (down from 4)
Binary-searched each of the four original settings, 20 runs per combination (FAIL = `count() = 2` for `SELECT count() FROM (SELECT * FROM dist GROUP BY key)`):

| Setting set | FAIL / 20 |
|---|---:|
| All 4 (baseline from #57814) | 20 |
| Drop `max_bytes_before_external_group_by` | 20 |
| Drop `memory_profiler_step` | 20 |
| Drop `max_untracked_memory` | 20 |
| Drop `group_by_two_level_threshold_bytes` | **0** |
| `group_by_two_level_threshold_bytes=1` + `memory_profiler_step=1Mi` | 20 |
| `group_by_two_level_threshold_bytes=1` + `max_untracked_memory=1Mi` | 20 |
| `group_by_two_level_threshold_bytes=1` alone | 0 |
| `memory_profiler_step=1Mi` alone | 0 |
| All defaults | 0 |

**Minimal sufficient set:** `group_by_two_level_threshold_bytes = 1` **plus** **one of** `memory_profiler_step = 1Mi` or `max_untracked_memory = 1Mi`. The two memory-profiler settings appear to be substitutable triggers for the same code path (the profiler firing during the cross-shard merge).

`max_bytes_before_external_group_by = 1` is **not** required — the bug is in two-level aggregation, not the spill-to-disk path, so the issue title's "external aggregation" framing is a red herring.

The bug also requires the `Distributed` cross-shard merge: same query on a single-node `LowCardinality(String)` `Memory` table with the same minimal settings returns 1 row correctly.

## What I pushed
Commit 54bc678d6b6d drops the two unneeded settings, keeps the broad `no-random-settings, no-random-merge-tree-settings` tags (without them the runner masks the bug ~30% of the time on asan even though the in-`SET`s take precedence), and rewrites the header comment with the binary-search findings.

## Wider concern (your call on direction)
The bug is **still present in master** — `f179c6cc3746` ("Use canonical hash for prealloc serialized aggregation precomputation") fixed the local-aggregation hash mismatch, but the cross-shard `MergingAggregatedMemoryEfficientTransform` path under `LowCardinality(String)` still mis-buckets identical keys when the two-level threshold + profiler conditions are met. CIDB shows the test failing across `amd_asan_ubsan` (19/20), `amd_tsan` (17/20 + 11/12 flaky), `amd_msan` (10/11 flaky), `arm_asan_ubsan` (5/5), and even `arm_binary, parallel` (6/14) — only the non-sanitiser amd debug + binary builds pass.

So this PR can land in three possible shapes; happy to take any direction:

1. **Investigate the actual fix in `MergingAggregatedMemoryEfficientTransform` first**, then this PR lands clean as the regression test. I can pick that up next — the lead is the LC-conversion path between bucket-by-bucket merge stages.
2. **Land as known-failing canary** — add `no-tsan, no-asan, no-msan, no-ubsan, no-arm` tags here so CI passes; reopen #57814.
3. **Drop this PR and reopen #57814** as still-active until (1) lands.

Which would you prefer? Default-leaning toward option 1; can have an investigation comment up in 24h.

(Session: `cron:clickhouse-ci-task-worker:20260518-111500`)

rienath:
@groeneai trying to reproduce locally from the asan_ubsan PR build. Matched the setup from your comment: `listen_host = 0.0.0.0`, both shards reachable, `users.d/memory_profiler.xml` linked in. Still 50/50 PASS via both raw client and `clickhouse-test --test-runs 50 -j 4`. Each iteration finishes in ~1.5s locally vs your ~43s on `t4g.large`. Three questions so I don't chase the wrong axis:

1. What hardware did you reproduce on?
2. Just `listen_host` + the two `memory_profiler` configs, or the full `tests/config/install.sh` symlink chain?
3. Exactly `clickhouse-test 04249_57814 --no-stateful --test-runs 10`, or with `-j`, `--flaky-check`, custom `ASAN_OPTIONS`/`UBSAN_OPTIONS`, or concurrent load?

groeneai:
@rienath thanks for digging in. Here are precise answers, and a correction: the earlier "t4g.large" reference in my comment was wrong. This workspace runs on `m7a.8xlarge`, not Graviton. Apologies for the confusion.

**Q1 - Hardware**
- AWS `m7a.8xlarge` (x86_64), `us-east-1c`
- AMD EPYC 9R14, 32 vCPU, 123 GiB RAM
- Ubuntu 24.04.4 LTS, kernel `6.17.0-1012-aws`

**Q2 - Config setup**
Just one file: `programs/server/config.d/listen.xml` containing only:
```xml
<clickhouse>
    <listen_host>0.0.0.0</listen_host>
</clickhouse>
```
That is the **only** change vs the stock `programs/server/` configs. No `users.d/memory_profiler.xml` was used. Re-reading my prior comment I see how it could read that way, sorry: the `memory_profiler_step = '1Mi'` setting is pinned by an in-test `SET` (lines 32-33 of `04249_57814_external_aggregation_lowcardinality_key_duplicate.sql.j2`), not by a `users.d` config. No `tests/config/install.sh` chain.

Without `listen.xml` the default config binds only on `[::1]` and `127.0.0.1`, so the second shard `127.0.0.2:9000` is unreachable and the bug cannot trigger via the cross-shard path. With `0.0.0.0` the second shard becomes reachable and the merge runs.

**Q3 - Test invocation**
Exactly:
```
PATH=build-asan/programs:$PATH tests/clickhouse-test 57814 --test-runs 10
```
No `-j`, no `--flaky-check`, no `--no-stateful` (irrelevant for this test), no custom `ASAN_OPTIONS`/`UBSAN_OPTIONS` (env unset). Build is `-DCMAKE_BUILD_TYPE=None -DSANITIZE=address,undefined`, `clang-21`, `RelWithDebInfo` + ASan+UBSan flags. Binary: master `8d3cf431a93c` + the regression test on top, freshly built `2026-05-19`.

Reproduced now (just re-ran on m7a.8xlarge): 10/10 FAIL through the test runner; 20/20 FAIL through bare `clickhouse-client` skipping the test runner entirely.

**Bare-client reproducer (no test runner, no jinja)**
Save as `repro.sh` and run against the same asan binary started with `listen_host=0.0.0.0`:
```bash
#!/bin/bash
set -e
CH=build-asan/programs/clickhouse
total=0; fail=0
for i in $(seq 1 20); do
  db="t_$i"
  out=$($CH client -q "
    CREATE DATABASE $db;
    CREATE TABLE $db.data (key String) ENGINE = Memory;
    CREATE TABLE $db.dist (key LowCardinality(String))
        ENGINE = Distributed(test_cluster_two_shards, $db, data);
    INSERT INTO $db.data VALUES ('foo');
    SELECT count() FROM (
      SELECT * FROM $db.dist GROUP BY key
        SETTINGS group_by_two_level_threshold_bytes = 1,
                 memory_profiler_step = '1Mi'
    );
    DROP DATABASE $db;
  ")
  total=$((total+1))
  [ "$out" != "1" ] && { fail=$((fail+1)); echo "iter $i: got $out, expected 1"; }
done
echo "RESULT: $fail/$total failed"
```
On `m7a.8xlarge` this prints `RESULT: 20/20 failed` (each `SELECT count()` returns 2 instead of 1).

If this raw-client form also shows 0/20 in your environment, then the difference is in the binary or runtime, not the test runner. Things worth checking on your side:
- `system.build_options` for `VERSION_GITHASH` and `CXX_FLAGS` (must contain `-fsanitize=address,undefined`)
- That `nc -zv 127.0.0.2 9000` succeeds (i.e. the second shard is actually reachable)
- That the asan binary is being used (not a release binary you happen to have on `PATH` first)

Happy to bisect further if the raw-client form also passes for you.

rienath (changes requested):
Thanks @groenai! This didn't help reproducing locally, but I found another repro

```sql
CREATE DATABASE shard_0; 
CREATE DATABASE shard_1;

CREATE TABLE shard_0.data (key String) ENGINE = Memory;
CREATE TABLE shard_1.data (key LowCardinality(String)) ENGINE = Memory;

INSERT INTO shard_0.data VALUES ('foo');
INSERT INTO shard_1.data VALUES ('foo');

CREATE TABLE dist (key LowCardinality(String)) ENGINE = Distributed(test_cluster_two_shards_different_databases, '', data);

SELECT * FROM dist GROUP BY key SETTINGS group_by_two_level_threshold_bytes = 1;
```

```text
   ┌─key─┐
1. │ foo │
2. │ foo │
   └─────┘
```

I will close the PR as the issue exists. We need to fix it before adding a test

groeneai:
Posted root-cause analysis and 3 concrete fix options on the issue: https://github.com/ClickHouse/ClickHouse/issues/57814#issuecomment-4627271006

@rienath / @nickitat - waiting on your steer on which option (A: detection-based fallback, B: re-bucket at initiator, C: wire-protocol tag) before I open the new fix PR. Once chosen, the regression test will use rienath's cleaner reproducer.

Session: cron:clickhouse-ci-task-worker:20260605-005100

PR #92141, opened 2025-12-15
Refactor VersionMetadata

tuanpach:
Refactor: Decouple version metadata logic from `IMergeTreeDataPart`

This PR introduces a class hierarchy for managing transactional version metadata, separating data representation from persistence logic.

## Motivation

The old `VersionMetadata` mixed data storage, business logic, and disk persistence in a single struct. This refactoring creates a clean separation of concerns to enable future extensions (e.g., metadata stored in ZooKeeper for replicated tables).

## Architecture Changes

**New class hierarchy:**
- `VersionInfo` - Pure data structure holding transaction IDs, CSNs, and storing version
- `VersionMetadata` - Abstract base class with core MVCC logic (visibility, locking, CSN management)
- `VersionMetadataOnDisk` - Concrete implementation that persists metadata to `txn_version.txt` files on disk

**Integration with `IMergeTreeDataPart`:**
- Changed `version` from value member to `std::unique_ptr<VersionMetadata>` to enable polymorphism
- Removed file I/O methods (`writeVersionMetadata`, `loadVersionMetadata`, `appendCSNToVersionMetadata`, etc.)
- File operations now encapsulated in `VersionMetadata` and `VersionMetadataOnDisk`

## Key Implementation Details

**`VersionInfo` (new pure data class):**
- Encapsulates `creation_tid`, `removal_tid`, `creation_csn`, `removal_csn`, and `storing_version`
- Provides `isVisible` logic for MVCC snapshot isolation
- Implements serialization/deserialization to/from buffers
- Uses `storing_version` field for optimistic concurrency control (detect concurrent metadata updates)

**`VersionMetadata` (abstract base class):**
- Implements high-level operations: `setAndStoreCreationTID`, `setAndStoreRemovalTID`, `setAndStoreCreationCSN`, `setAndStoreRemovalCSN`
- Provides `loadAndAdjustMetadata` to recover CSNs from `TransactionLog` after server restart
- Uses template method pattern: calls abstract `storeInfoImpl`, `loadMetadata`, `tryLockRemovalTID`, etc.
- Implements retry logic for version conflicts via `updateStoreAndSetMetadataWithRefresh`

**`VersionMetadataOnDisk` (disk persistence implementation):**
- Stores metadata in `txn_version.txt` file within data part directory
- Uses atomic write protocol: write to `.tmp` file, then rename
- Implements deferred persistence optimization: delays writing metadata for non-transactional parts until first transaction touches the part (reduces disk I/O)
[...]

PR #105116, opened 2026-05-16
Fix logical error on a non-transactional DROP PARTITION over an uncommitted part

### LINKED ISSUE #109698: Received signal 6 (internal) (STID: 3563-49ac)

Test name: Received signal 6 (internal) (STID: 3563-49ac)
CI report: [Stress test (amd_debug)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=108325&sha=d9014f8f51d3c4aa2117437a9f6f38fc40d78d2c&name_0=PR&name_1=Stress%20test%20%28amd_debug%29)
Failing test history: [cidb](https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICdSZWNlaXZlZCBzaWduYWwgNiAoaW50ZXJuYWwpIChTVElEOiAzNTYzLTQ5YWMpJwogICAgLS0gQU5EIGNoZWNrX25hbWUgPSAnU3RyZXNzIHRlc3QgKGFtZF9kZWJ1ZyknCiAgICBBTkQgdGVzdF9zdGF0dXMgSU4gKCdGQUlMJywgJ0VSUk9SJykKICAgIEFORCAocHVsbF9yZXF1ZXN0X251bWJlciA9IDAgT1IgYmFzZV9yZWYgSU4gKCdtYXN0ZXInKSkKICAgIC0tIEFORCB0ZXN0X2NvbnRleHRfcmF3IExJS0UgJyVwYXR0ZXJuJScgIC0tIHVuY29tbWVudCBhbmQgZWRpdCB0byBmaWx0ZXIgYnkgZmFpbHVyZSBwYXR0ZXJuCkdST1VQIEJZIGRheQpPUkRFUiBCWSBkYXkgREVTQwo=)

Test output:
```
Error:
Received signal 6 (internal)
---

Stack trace:
pthread_kill@@GLIBC_2.34 @ 0x00000000000969fd
raise @ 0x0000000000042476
__lgamma_r_finite @ 0x00000000000287f3
src/Common/SignalHandlers.cpp:221:5: terminate_handler() @ 0x00000000157bfdb9
contrib/llvm-project/libcxxabi/src/cxa_handlers.cpp:59:9: std::__terminate(void (*)()) @ 0x000000002e687f83
contrib/llvm-project/libcxxabi/src/cxa_handlers.cpp:88:17: std::terminate() @ 0x000000002e687efe
src/Storages/MergeTree/MergeTreeData.cpp:9093:9: DB::MergeTreeData::Transaction::commit(DB::DataPartsLock&) @ 0x000000002144cf26
src/Storages/MergeTree/MergeTreeData.cpp:9027:12: DB::MergeTreeData::Transaction::commit() @ 0x00000000214b77b4
src/Storages/StorageMergeTree.cpp:2583:17: DB::StorageMergeTree::renameAndCommitEmptyParts(std::vector<std::shared_ptr<DB::IMergeTreeDataPart>, std::allocator<std::shared_ptr<DB::IMergeTreeDataPart>>>&, DB::MergeTreeData::Transaction&) @ 0x0000000020f1e0fa
src/Storages/StorageMergeTree.cpp:2631:13: DB::StorageMergeTree::truncate(boost::intrusive_ptr<DB::IAST> const&, std::shared_ptr<DB::StorageInMemoryMetadata const> const&, std::shared_ptr<DB::Context const>, DB::TableExclusiveLockHolder&) @ 0x0000000020f1f481
src/Interpreters/InterpreterDropQuery.cpp:330:20: DB::InterpreterDropQuery::executeToTableImpl(std::shared_ptr<DB::Context const> const&, DB::ASTDropQuery&, std::shared_ptr<DB::IDatabase>&, StrongTypedef<wide::integer<128ul, unsigned int>, DB::UUIDTag>&) @ 0x000000001b14127d
src/Interpreters/InterpreterDropQuery.cpp:150:16: DB::InterpreterDropQuery::executeToTable(DB::ASTDropQuery&) @ 0x000000001b13ef20
src/Interpreters/InterpreterDropQuery.cpp:112:16: DB::InterpreterDropQuery::executeSingleDropQuery(boost::intrusive_ptr<DB::IAST> const&) @ 0x000000001b13e34b
src/Interpreters/InterpreterDropQuery.cpp:93:15: DB::InterpreterDropQuery::execute() @ 0x000000001b13de5a
src/Interpreters/executeQuery.cpp:1852:40: DB::executeQueryImpl(char const*, char const*, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum, std::unique_ptr<DB::ReadBuffer, std::default_delete<DB::ReadBuffer>>&, boost::intrusive_ptr<DB::IAST>&, std::shared_ptr<DB::ImplicitTransactionControlExecutor>, std::function<void ()>, DB::QueryResultDetails&) @ 0x000000001b55db61
src/Interpreters/executeQuery.cpp:2548:19: DB::executeQuery(std::unique_ptr<DB::ReadBuffer, std::default_delete<DB::ReadBuffer>>, DB::WriteBuffer&, std::shared_ptr<DB::Context>, std::function<void (DB::QueryResultDetails const&)>, DB::QueryFlags, std::optional<DB::FormatSettings> const&, std::function<void (DB::IOutputFormat&, String const&, std::shared_ptr<DB::Context const> const&, std::optional<DB::FormatSettings> const&)>, std::function<void ()>, std::function<void ()>) @ 0x000000001b562bbf
src/Interpreters/executeQuery.cpp:2408:5: DB::executeQuery(DB::ReadBuffer&, DB::WriteBuffer&, std::shared_ptr<DB::Context>, std::function<void (DB::QueryResultDetails const&)>, DB::QueryFlags, std::optional<DB::FormatSettings> const&, std::function<void (DB::IOutputFormat&, String const&, std::shared_ptr<DB::Context const> const&, std::optional<DB::FormatSettings> const&)>, std::function<void ()>, std::function<void ()>) @ 0x000000001b5624b3
src/Interpreters/DDLWorker.cpp:591:9: DB::DDLWorker::tryExecuteQuery(DB::DDLTaskBase&, std::shared_ptr<zkutil::ZooKeeper> const&, bool) @ 0x000000001afca357
src/Interpreters/DDLWorker.cpp:764:30: DB::DDLWorker::processTask(DB::DDLTaskBase&, std::shared_ptr<zkutil::ZooKeeper> const&, bool) @ 0x000000001afc7a24
src/Databases/DatabaseReplicatedWorker.cpp:501:5: DB::DatabaseReplicatedDDLWorker::tryEnqueueAndExecuteEntry(DB::DDLLogEntry&, std::shared_ptr<DB::Context const>, bool) @ 0x000000001a23313e
src/Databases/DatabaseReplicated.cpp:1479:36: DB::DatabaseReplicated::tryEnqueueReplicatedDDL(boost::intrusive_ptr<DB::IAST> const&, std::shared_ptr<DB::Context const>, DB::QueryFlags, std::unique_ptr<DB::DDLGuard, std::default_delete<DB::DDLGuard>>&&) @ 0x000000001a20117f
src/Interpreters/InterpreterDropQuery.cpp:273:30: DB::InterpreterDropQuery::executeToTableImpl(std::shared_ptr<DB::Context const> const&, DB::ASTDropQuery&, std::shared_ptr<DB::IDatabase>&, StrongTypedef<wide::integer<128ul, unsigned int>, DB::UUIDTag>&) @ 0x000000001b140c4f
src/Interpreters/InterpreterDropQuery.cpp:150:16: DB::InterpreterDropQuery::executeToTable(DB::ASTDropQuery&) @ 0x000000001b13ef20
src/Interpreters/InterpreterDropQuery.cpp:112:16: DB::InterpreterDropQuery::executeSingleDropQuery(boost::intrusive_ptr<DB::IAST> const&) @ 0x000000001b13e34b
src/Interpreters/InterpreterDropQuery.cpp:93:15: DB::InterpreterDropQuery::execute() @ 0x000000001b13de5a
src/Interpreters/executeQuery.cpp:1852:40: DB::executeQueryImpl(char const*, char const*, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum, std::unique_ptr<DB::ReadBuffer, std::default_delete<DB::ReadBuffer>>&, boost::intrusive_ptr<DB::IAST>&, std::shared_ptr<DB::ImplicitTransactionControlExecutor>, std::function<void ()>, DB::QueryResultDetails&) @ 0x000000001b55db61
src/Interpreters/executeQuery.cpp:2326:11: DB::executeQuery(std::basic_string_view<char, std::char_traits<char>>, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum) @ 0x000000001b5569ca
src/Interpreters/executeQuery.cpp:2256:27: DB::executeASTFuzzerQueries(boost::intrusive_ptr<DB::IAST> const&, std::shared_ptr<DB::Context> const&, double, bool) @ 0x000000001b568d15
src/Interpreters/executeQuery.cpp:2384: operator()
contrib/llvm-project/libcxx/include/__type_traits/invoke.h:90: std::__invoke_result_impl<void, DB::executeQuery(std::basic_string_view<char, std::char_traits<char>>, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum)::$_0&, DB::QueryPipelineFinalizedInfo const&, std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<long long, std::ratio<1l, 1000000l>>>>::type std::__invoke[abi:sqe220101]<DB::executeQuery(std::basic_string_view<char, std::char_traits<char>>, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum)::$_0&, DB::QueryPipelineFinalizedInfo const&, std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<long long, std::ratio<1l, 1000000l>>>>(DB::executeQuery(std::basic_string_view<char, std::char_traits<char>>, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum)::$_0&, DB::QueryPipelineFinalizedInfo const&, std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<long long, std::ratio<1l, 1000000l>>>&&)
contrib/llvm-project/libcxx/include/__type_traits/invoke.h:350: void std::__invoke_void_return_wrapper<void, true>::__call[abi:sqe220101]<DB::executeQuery(std::basic_string_view<char, std::char_traits<char>>, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum)::$_0&, DB::QueryPipelineFinalizedInfo const&, std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<long long, std::ratio<1l, 1000000l>>>>(DB::executeQuery(std::basic_string_view<char, std::char_traits<char>>, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum)::$_0&, DB::QueryPipelineFinalizedInfo const&, std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<long long, std::ratio<1l, 1000000l>>>&&)
contrib/llvm-project/libcxx/include/__type_traits/invoke.h:356: void std::__invoke_r[abi:sqe220101]<void, DB::executeQuery(std::basic_string_view<char, std::char_traits<char>>, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum)::$_0&, DB::QueryPipelineFinalizedInfo const&, std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<long long, std::ratio<1l, 1000000l>>>>(DB::executeQuery(std::basic_string_view<char, std::char_traits<char>>, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum)::$_0&, DB::QueryPipelineFinalizedInfo const&, std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<long long, std::ratio<1l, 1000000l>>>&&)
contrib/llvm-project/libcxx/include/__functional/function.h:443:25: ? @ 0x000000001b5713b1
contrib/llvm-project/libcxx/include/__functional/function.h:502: ?
contrib/llvm-project/libcxx/include/__functional/function.h:754: ?
src/QueryPipeline/BlockIO.cpp:80:12: DB::BlockIO::onFinish(std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<long long, std::ratio<1l, 1000000l>>>) @ 0x0000000019edca31
src/Server/TCPHandler.cpp:935:33: DB::TCPHandler::runImpl() @ 0x0000000021c2c0ce
src/Server/TCPHandler.cpp:3203:9: DB::TCPHandler::run() @ 0x0000000021c48924
base/poco/Net/src/TCPServerConnection.cpp:40:3: Poco::Net::TCPServerConnection::start() @ 0x0000000028f980ce
base/poco/Net/src/TCPServerDispatcher.cpp:115:42: Poco::Net::TCPServerDispatcher::run() @ 0x0000000028f98612
base/poco/Foundation/src/ThreadPool.cpp:207:14: Poco::PooledThread::run() @ 0x0000000028f612bf
base/poco/Foundation/src/Thread_POSIX.cpp:341:27: Poco::ThreadImpl::runnableEntry(void*) @ 0x0000000028f5f68f
__GI___clone3 @ 0x00000000001268d0
```

tuanpach:
A non-transactional part remover sees every active part, including one whose creating transaction has not committed yet. Covering such a part writes `removal_tid = Tx::NonTransactionalTID`, which immediately sets `removal_csn = Tx::NonTransactionalCSN`, producing the on-disk shape `creation_csn = 0, removal_csn = 1` that `VersionMetadata::validateInfo` rejects as unrecoverable — a logical error, and a server abort in debug / sanitizer builds:

```
Logical error: 'Object ...|all_12_25_7, creation_csn is not set while removal_csn is set to 1'.
```

`TRUNCATE` was fixed for this in https://github.com/ClickHouse/ClickHouse/pull/109598 by skipping uncommitted parts, but the other non-transactional removers were not: `ALTER TABLE ... DROP PARTITION`, `DETACH PARTITION` and `DROP PART` still take the whole active part set. This change refuses the bad write one layer lower, in `VersionMetadata::setAndStoreRemovalTID`, with `SERIALIZATION_ERROR`, so the caller can retry once the in-flight creation finalizes. The guard inspects only the on-disk `VersionInfo` and the transaction log, so it is correct regardless of which replica the creating transaction runs on, and it does not fire for a creator that is already committed but whose `creation_csn` has not been written to this part yet.

A non-transactional removal is also not undoable — nothing ever clears `removal_csn` — so a batch that fails halfway would leave the already-stamped parts invisible for good even though the statement failed with a supposedly retryable error. `NonTransactionalRemovalLocks` now takes the removal lock of every part of a batch before any of them is written, so a conflict (an in-flight creator, or another transaction already removing one of the parts) is reported before anything is persisted and the statement leaves the table untouched.

`REPLACE PARTITION FROM` and `MOVE PARTITION TO TABLE` commit their own new parts *before* removing the old ones, so refusing the removal at that point would leave the partition half replaced or half moved. Both now check the parts they are about to remove first, under the same parts lock they commit with. `MOVE` uses the stricter check of the two: it republishes the rows in another table where a commit cannot be taken back, so the creating transaction must be committed rather than merely no longer running. `StorageReplicatedMergeTree::replacePartitionFromImpl` has the same ordering but is not reachable — `IInterpreter::checkStorageSupportsTransactionsIfNeeded` rejects any transaction touching a storage that supports replication, so a replicated table never holds a part with an uncommitted transactional creator.

A non-transactional `DETACH` has to both copy the parts to `detached/` and remove them, and neither order is safe on its own: copying first leaves an orphan copy behind whenever the removal is then refused (and a `_tryN` directory on every retry), while copying afterwards makes an ordinary I/O error destructive — the parts are already gone and only a prefix of the partition is under `detached/`, so a later `ATTACH PARTITION` restores just that prefix. `renameAndCommitEmptyParts` now does the whole thing under the one parts lock it precommits the empty parts with: it checks removability, copies the covered parts, and calls `MergeTreeData::Transaction::commit` with that same lock. The covered set therefore cannot change in between — the copy is of exactly the parts that are removed, a refusal happens before anything is copied, and a failed copy fails the statement with the parts still active. `StorageReplicatedMergeTree` clones inside the removal's own parts lock for the same reason. Holding the lock across the copy is cheap here: for a plain `MergeTree` the copy is a hardlink-based freeze, since `copy_instead_of_hardlink` in `makeCloneInDetached` is gated on `supportsReplication`.

Originally surfaced by `Stress test (amd_tsan, meta in keeper)`: https://github.com/ClickHouse/clickhouse-private/actions/runs/25947990737 (on https://github.com/ClickHouse/clickhouse-private/pull/57844).

Closes: https://github.com/ClickHouse/ClickHouse/issues/109698
Related: https://github.com/ClickHouse/ClickHouse/pull/109598

Fix a logical error (a server abort in debug builds) when a non-transactional `DROP PARTITION`, `DETACH PARTITION`, `DROP PART`, `REPLACE PARTITION FROM` or `MOVE PARTITION TO TABLE` covered a part whose creating transaction had not committed yet, leaving that part with `creation_csn = 0` and `removal_csn = 1`. Such an operation now fails with `SERIALIZATION_ERROR` and can be retried once the creating transaction finishes; it no longer removes or replaces any part of the affected batch.

---
Workflow [[PR](https://s3.amazonaws.com/clickhouse-test-reports/praktika.html?PR=105116&sha=latest&name_0=PR)]
Sync PR [[sync-upstream/pr/105116](https://github.com/search?q=head%3Async-upstream%2Fpr%2F105116+org%3AClickHouse+type%3Apr&type=pullrequests)]

### Version info
- Merged into: `26.9.1.1299` (included in `26.9` and later)


tests/queries/0_stateless/04241_truncate_non_tx_vs_insert_tx.sh
```diff
@@ -0,0 +1,66 @@
...
+insert_pid=$!
+
+# Wait until the background INSERT has reached the pause.
+$CLICKHOUSE_CLIENT -q "SYSTEM WAIT FAILPOINT transaction_before_csn_publish PAUSE"
+
+# Foreground: non-transactional TRUNCATE.
+# Without the fix this crashes the server on a LOGICAL_ERROR.
+# With the fix it must fail with `SERIALIZATION_ERROR`.
+$CLICKHOUSE_CLIENT -q "TRUNCATE TABLE t_truncate_non_tx_vs_insert_tx" 2>&1 \
+    | grep -oE "SERIALIZATION_ERROR|LOGICAL_ERROR" | head -1
```
antonio2368: Wouldn't this catch LOGICAL_ERROR also which is wrong?

antonio2368 (commented):
One AI finding:
I think this still has a blocker in the multi-part replacement path.

The new guard in `VersionMetadata::setAndStoreRemovalTID` is locally correct: for `Tx::NonTransactionalTID`, it throws `SERIALIZATION_ERROR` before setting `removal_tid`/`removal_csn` when the covered part still has an uncommitted transactional `creation_tid`.

The problem is the caller contract. In `MergeTreeTransaction::addNewPartAndRemoveCovered`, the non-transactional branch stamps covered parts one by one:

```cpp
for (const auto & covered : covered_parts)
{
    transaction_context.part_name = covered->name;
    covered->version->setAndStoreNonTransactionalRemovalTID(transaction_context);
}
```

For a non-transactional `TRUNCATE`, `DROP PARTITION`, or `DETACH PARTITION` that covers multiple parts, this can leave partial persistent side effects:

1. Covered part `A` is already committed, so `setAndStoreNonTransactionalRemovalTID` succeeds and persists `removal_tid = Tx::NonTransactionalTID` / `removal_csn = Tx::NonTransactionalCSN`.
2. Covered part `B` was created by an in-flight transaction whose `creation_csn` has not been published yet, so the new guard throws `SERIALIZATION_ERROR`.
3. `MergeTreeData::Transaction::commit` exits before the later `NOEXCEPT_SCOPE` that activates the new empty covering part and moves covered parts to `Outdated`.
4. Rollback only handles the precommitted new part. It does not clear the removal metadata already persisted on `A`.

After that failed user operation, `A` can already be invisible: `VersionInfo::isVisible` returns `false` immediately for `removal_tid.isNonTransactional`. So a retryable `SERIALIZATION_ERROR` can still leave committed data logically removed even though the operation failed.

The `locked_parts_to_cleanup` vector in `MergeTreeData::Transaction::commit` looks intended for this, but it is only appended after `addNewPartAndRemoveCovered` returns and I do not see any consumer for it. It also would not cover the intra-call case where one covered part is stamped and a later covered part throws before the helper returns.

I think the fix should either preflight all covered parts for this new uncommitted-creator condition before persisting any non-transactional removal metadata, or make the stamping phase rollback-safe by recording each successfully stamped part immediately and clearing both `removal_tid` and `removal_csn` if any later part fails.

tuanpach (commented):
(removed)

tests/queries/0_stateless/04241_truncate_non_tx_vs_insert_tx.sh
tuanpach: Fair point, tightened in 4fc80793efb -- `grep -o "SERIALIZATION_ERROR" | head -1`. A `LOGICAL_ERROR` regression would now produce a clearer empty-vs-`SERIALIZATION_ERROR` diff against the reference.

src/Interpreters/MergeTreeTransaction/VersionMetadata.cpp
```diff
@@ -138,11 +138,23 @@ void VersionMetadata::setAndStoreRemovalTID(const TransactionID & tid)
...
+    auto update_function = [tid, this](VersionInfo & info)
     {
         if (info.removal_tid == tid)
             return false;
 
+        /// Refuse a non-tx removal over a part whose tx creation has not committed:
+        /// the resulting `creation_csn = 0 + removal_csn = 1` shape would be rejected
+        /// by `validateInfo` and is unrecoverable on restart.
+        if (tid.isNonTransactional()
+            && !info.creation_csn
```
antonio2368, line 150: should we do TransactionLog::getCSN(current_info.creation_tid)? same question for above

tests/queries/0_stateless/04242_truncate_non_tx_partial_removal.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+# Tags: no-ordinary-database, no-encrypted-storage, no-parallel
```
antonio2368: why no-encrypted-storage?

tests/queries/0_stateless/04242_truncate_non_tx_partial_removal.sh
```diff
@@ -0,0 +1,50 @@
...
+# triggers `SERIALIZATION_ERROR`. With the preflight, the first must NOT have been
+# stamped invisible.
+$CLICKHOUSE_CLIENT -q "TRUNCATE TABLE t_truncate_non_tx_partial_removal" 2>&1 \
+    | grep -o "SERIALIZATION_ERROR" | head -1
+
+$CLICKHOUSE_CLIENT -q "SYSTEM DISABLE FAILPOINT transaction_before_csn_publish"
+wait "$insert_pid"
+
+# Both committed rows must still be visible.
+$CLICKHOUSE_CLIENT -q "SELECT count() FROM t_truncate_non_tx_partial_removal"
```
antonio2368, line 48: should this be inside a transaction to verify it works?

tuanpach:
Sorry, I will work on it.

src/Interpreters/MergeTreeTransaction/VersionMetadata.cpp
tuanpach: `TransactionLog::getCSN(info.creation_tid)` would eliminate spurious `SERIALIZATION_ERROR`s in the window where the creator has committed globally but this replica's on-disk `creation_csn` hasn't caught up yet.

Leaving it out of this PR — the guard is still correct without it, just more conservative than necessary.

tests/queries/0_stateless/04242_truncate_non_tx_partial_removal.sh
tuanpach: No need to put it inside a transaction. The above INSERT is already committed.

tests/queries/0_stateless/04242_truncate_non_tx_partial_removal.sh
tuanpach: Removed it.

antonio2368 (approved):

src/Interpreters/MergeTreeTransaction/VersionMetadata.cpp
alexey-milovidov: 🕵 Done in 65e4e2b5bf6. `VersionMetadata::isCreatedByUncommittedTransaction` now falls back to `TransactionLog::getCSN(creation_tid)` when `creation_csn` is unset, the same way `isVisible` and `canBeRemoved` already do, so a creator that is committed globally but not yet stamped on this replica no longer produces a spurious `SERIALIZATION_ERROR` (a rolled back creator answers `Tx::RolledBackCSN`, which is also not refused).

The lookup happens before `updateInfoWithRefreshDataThenStoreAndSetMetadata` so it is not done under the metadata update lock; the update function still re-checks the refreshed `VersionInfo` before throwing.

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -5844,6 +5844,22 @@ void MergeTreeData::removePartsFromWorkingSet(MergeTreeTransaction * txn, const
...
+    /// Non-tx path: check upfront that no part has an in-flight tx creator.
+    /// A mid-loop throw in the per-part `removeOldPart` below would otherwise leave
+    /// earlier parts already stamped with `removal_tid = NonTransactionalTID` on disk --
+    /// silently invisible and not undone by any rollback path.
+    if (!txn)
+    {
+        for (const auto & part : remove)
+        {
+            auto info = part->version->getInfo();
+            if (!info.creation_csn && !info.creation_tid.isNonTransactional())
```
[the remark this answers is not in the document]
alexey-milovidov: 🕵 Correct, and the same gap applied to the `covered_parts` preflight below. Fixed in 86b6861a1a8 by reserving instead of pre-checking: `NonTransactionalRemovalLocks` takes the removal lock (`lockRemovalTID`) of every part of the batch before any `removal_tid` is written, so both failure sources — an in-flight creator and a part already locked by another running transaction — are reported before anything is persisted. Because the locks are *held* across the stamping phase rather than merely checked, there is no window for another transaction to take one in between. Locks that are never stamped are released by the destructor, so a failed statement leaves every part of the batch untouched and stays retryable.

`locked_parts_to_cleanup` (populated, never read) is removed in the same commit.

tests/queries/0_stateless/04241_truncate_non_tx_vs_insert_tx.sh
alexey-milovidov: 🕵 The test file is gone in 9f39b6a98b2: the `TRUNCATE` case it covered is already fixed in master by #109598, so it no longer reproduced anything and was failing `Fast test`. Its replacement `05161` matches on `SERIALIZATION_ERROR` only — a regression back to a logical error prints nothing at all, which is an unambiguous diff.

tests/queries/0_stateless/04242_truncate_non_tx_partial_removal.sh
alexey-milovidov: 🕵 `04242` is gone in 9f39b6a98b2 (the `TRUNCATE` case it covered is already fixed in master by #109598). Its replacement `05162` keeps the same shape — a committed part plus a part whose creating transaction is still open — but drives it with `DROP PARTITION`, which #109598 did not change.

alexey-milovidov:
🕵 Heads-up on an overlap: #110494 ("Reject non-transactional removal of parts created by uncommitted transactions") targets the same defect with the same strategy — `SERIALIZATION_ERROR` from `setAndStoreRemovalTID`, the `getCSN` refinement, and removing the dead `locked_parts_to_cleanup`. It goes further on one axis (`REPLACE PARTITION FROM` and `MOVE PARTITION TO TABLE`, which publish the new parts before removing the old ones, so a rejected `REPLACE` can leave a half-replaced destination); this pull request goes further on another (it holds the removal locks across the stamping phase, which closes the mid-batch conflict window that is still an open review finding on #110494).

Worth deciding which one to take forward before either is merged, rather than resolving the conflict afterwards. Also note the `TRUNCATE` half of the original report is already fixed in master by #109598, so the remaining exposure is `DROP PARTITION` / `DETACH PARTITION` / `DROP PART` (and the `REPLACE` / `MOVE` ordering issue that only #110494 covers).

alexey-milovidov:
🕵 @groeneai, investigate the failure: https://s3.amazonaws.com/clickhouse-test-reports/praktika.html?PR=105116&sha=a003ff5476dfb2ae3c4d88a48057edecf314af86&name_0=PR&name_1=Fast%20test%20%28arm_darwin%29 and provide a fix in a separate PR. If the fix is already in progress, link it here.

`Fast test (arm_darwin)` fails on `05076_error_trace_file_offsets` with `1 1` vs `1 0` — no `last_error_trace` address lands above `max(address_end)` from `system.symbols`.

Not caused by this pull request:
- the test does not exist in this branch at all; it comes from `master` through the CI merge;
- it fails on `master` itself (`pull_request_number = 0`) 6 times on 2026-09-10 and twice on 2026-09-09 in `Stateless tests (amd_release, parallel, 1/2)`;
- it failed the same `Fast test (arm_darwin)` job today on the unrelated #119077;
- it exercises `system.symbols` / `addressToSymbol` file offsets (see d240ec3a54e), nothing to do with transactions or part removal.

I found no open issue for it.

groeneai:
Confirmed, not from this pull request.

The same `1 1` vs `1 0` diff has two independent causes, and the fix already in progress covers only one of them.

**`Stateless tests (amd_release)`, fix in progress: #119136.** `SymbolIndex` indexes absolute ELF symbols (`SHN_ABS`, whose `st_value` is a constant rather than an address). ThinLTO's `-fwhole-program-vtables` emits the CFI type-id globals `__typeid__*` that way, so 505 of them reach `system.symbols` and lift `max(address_end)` from 685204737 to 1162219258676258. Both columns of the test are bounded by that value, so no captured frame exceeds it any more.

**`Fast test (arm_darwin)`: a second cause, which #119136 does not reach.**

* The Fast test build sets `-DENABLE_THINLTO=0` (`ci/jobs/fast_test.py:274`) and `-fwhole-program-vtables` is added only under `ENABLE_THINLTO` (`CMakeLists.txt:502`), so no `__typeid__*` symbol can exist in that binary. The Linux variant of the same job has 72 passes and no failures on this test over the last 20 days.
* #119136 guards the two ELF collectors at `SymbolIndex.cpp:114` and `:281`, both inside `#if defined(__ELF__)`. macOS compiles the `#elif defined(OS_DARWIN)` branch instead, so the patch is not built there at all.
* On macOS the test cannot pass. `collectSymbolsFromMachOImage` stores `n_value + slide`, an absolute runtime address, for every image `_dyld_image_count()` reports (`SymbolIndex.cpp:601-604`, deliberately: `findSymbol` skips the address-to-offset conversion on macOS). `max(address_end)` is therefore the top of the loaded address space, not a bound on the main binary. The printed `1 0` says exactly that: more than ten frames sit below the bound, none above it, the libc frames included. `StackTrace::resolveAddress` also returns `AddressKind::Unsupported` off ELF (`StackTrace.cpp:355`), so nothing is stored as a file offset on macOS: the property the second column asserts does not exist on that platform.
* 12 runs, 12 failures, no pass, across 12 pull requests, starting 02:07 UTC today when the test first reached that job. It is deterministic, and it is my miss: the test came with #117449 and the arm_darwin job had not run it before that merge.

I am opening a separate PR to keep the test off macOS (`no-darwin`, since the file-offset storage it asserts is ELF-only) and will link it here.

groeneai:
Opened: #119145.

It tags the test `no-darwin` and states the mechanism in the comment: on Mach-O neither the file-offset storage nor a `max(address_end)` that bounds the main binary exists, so `resolveAddress` returns `AddressKind::Unsupported` and the ceiling sits above every captured frame. `Fast test (arm_darwin)` on that pull request is the macOS verification and it lands before merge; the assertion is untouched on the ELF arms, where `Fast test` is 84 OK and 0 FAIL over 53 pull requests.

src/Interpreters/MergeTreeTransaction/VersionMetadata.cpp
alexey-milovidov: 🕵 The blocker on this line is addressed in 319e693e72f.

`REPLACE PARTITION FROM` and `MOVE PARTITION TO TABLE` now check the parts they are about to remove *before* committing their own new ones, under the same parts lock they commit with, so a refused removal leaves the partition untouched instead of half replaced or half moved.

`MOVE` uses a stricter predicate, `VersionMetadata::isCreationCommitted`: it republishes the rows in another table where a commit cannot be taken back, so the creating transaction must be committed rather than merely no longer running — otherwise a creation that later rolls back would resurrect rows that never existed.

`StorageReplicatedMergeTree::replacePartitionFromImpl` has the same ordering but is not reachable: `IInterpreter::checkStorageSupportsTransactionsIfNeeded` throws `NOT_IMPLEMENTED` for any storage that `supportsReplication()`, regardless of `throw_on_unsupported_query_inside_transaction`, so a replicated table never holds a part with an uncommitted transactional creator. A check there would only add a throw *after* the Keeper commit, which is strictly worse.

The new `05175_replace_move_partition_non_tx_uncommitted` covers both statements. I verified it discriminates: with the two checks reverted and everything else unchanged, it fails — the retried `MOVE` trips over `PART_IS_TEMPORARILY_LOCKED: Part all_1_1_1 (state Outdated) already exists` on the half-moved destination.

alexey-milovidov:
🕵 On the two blockers in the latest AI review — one is implemented (d5a853d4205), the other I believe is misdiagnosed.

**Blocker 2 (`REPLACE` / `MOVE` precheck does not reserve `removal_tid`) — implemented, with a correction to the reasoning.**

The check now also rejects a part that another transaction holds the removal lock on, so both ways the later removal can be refused are caught before either statement publishes anything.

The stated mechanism is not the one that makes it reachable, though. `MergeTreeData::Transaction::commit` and `removePartsFromWorkingSet` are the only places that lock an `Active` part for removal, and both flip it to `Outdated` in the same parts-lock critical section — so a concurrent transactional remover cannot leave an `Active` part locked for another observer to find. I tried to build the suggested regression (`BEGIN TRANSACTION; TRUNCATE`, then a non-transactional `REPLACE PARTITION FROM` from another session) and it produced no `SERIALIZATION_ERROR` at all, because the transactional `TRUNCATE` had already moved the part out of the active set.

What *is* reachable is `MergeTreeTransaction::rollback`: it calls `restoreAndActivatePart` first and clears `removal_tid` / `unlockRemovalTID` afterwards, outside the parts lock, so a part is briefly `Active` and still locked. Reaching that window from a test needs fault injection, and this pull request deliberately removed its last failpoint, so the check ships without a dedicated test rather than with one that does not reach the branch. `05175` covers the creator-state branch.

**Blocker 1 (`getCSN` false-positive window) — pre-existing, and strictly narrowed by this pull request.**

The window is real and the description of the ordering is right: the CSN znode is created, `afterCommit` then stamps `creation_csn` on each part, and `runUpdatingThread` populates `tid_to_csn` later still.

But this is not something the pull request introduces. Before it, the guard was `!info.creation_csn` alone, which is false-positive for that window *and* for the much longer period where a transaction committed long ago but this particular part was never stamped. Adding the `TransactionLog::getCSN` fallback strictly shrinks the window; the review itself calls it "the existing … window". For comparison, the behaviour being replaced in that same window is not success — it is the `creation_csn is not set while removal_csn is set to 1` logical error, i.e. a server abort in debug and sanitizer builds.

Closing the remainder needs the CSN published synchronously at commit point, as the review suggests. That is a change to the transaction-log commit protocol affecting every reader of `tid_to_csn`, not to the removal guard, and it does not belong in a bug fix for the abort. `VersionMetadata` has no atomic `creation_csn` to pass as `getCSNAndAssert`'s fallback either — `getInfo()` already reads the mutex-protected struct, so it is as fresh as the data gets; the value simply does not exist yet in that window.

Happy to be overruled and split the commit-point change into its own pull request if you want it tracked.

tests/queries/0_stateless/05161_drop_partition_non_tx_vs_insert_tx.sh
```diff
@@ -0,0 +1,40 @@
...
+tx 1 "BEGIN TRANSACTION"
+tx 1 "INSERT INTO t_drop_partition_non_tx SETTINGS async_insert = 0 VALUES (1)"
+$CLICKHOUSE_CLIENT -q "ALTER TABLE t_drop_partition_non_tx DROP PARTITION ALL" 2>&1 | grep -o -F "SERIALIZATION_ERROR" | head -1
+tx 1 "ROLLBACK"
+$CLICKHOUSE_CLIENT -q "SELECT count() FROM t_drop_partition_non_tx"
+
+# Case 2: DETACH PARTITION goes through the same non-transactional removal path.
+tx 2 "BEGIN TRANSACTION"
+tx 2 "INSERT INTO t_drop_partition_non_tx SETTINGS async_insert = 0 VALUES (2)"
+$CLICKHOUSE_CLIENT -q "ALTER TABLE t_drop_partition_non_tx DETACH PARTITION ALL" 2>&1 | grep -o -F "SERIALIZATION_ERROR" | head -1
```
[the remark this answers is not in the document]
alexey-milovidov: 🕵 Confirmed and fixed in 4a83a2de1e5 — this finding was right.

`dropPart` and `dropPartition` now check the parts they are about to remove *before* `makeCloneInDetached` runs, so a refused `DETACH` writes nothing to `detached/` and a retry cannot pile up `_tryN` directories. Checking earlier also avoids building the empty covering parts for an operation that cannot proceed.

`05161` now asserts `count() FROM system.detached_parts` is 0 after the refused `DETACH PARTITION`, and I verified the assertion discriminates: with the two checks reverted and nothing else changed, the test fails reporting `1` detached part instead of `0`.

alexey-milovidov:
🕵 The premise of this round is correct, but I do not think the required action belongs in this pull request. Laying out the facts so it is an informed call.

**The gap is real.** The precheck is not a reservation. `dropPartition` / `dropPart` hold `lockOperationsWithParts()`, which only the partition commands themselves take (`StorageMergeTree.cpp:2948, 3015, 3128` and the `REPLACE` / `MOVE` pair) — `INSERT` does not. `stopMergesAndWait()` blocks merges, not inserts. So between the precheck and `MergeTreeData::Transaction::commit`, a transactional `INSERT` can land a new part in the partition; `getActivePartsToReplace` recomputes the covered set at commit time, picks it up, and `NonTransactionalRemovalLocks::lock` then throws — after `makeCloneInDetached` has run.

**But that outcome is not introduced here.** In exactly that sequence, before this pull request, `setAndStoreRemovalTID` wrote `creation_csn = 0` + `removal_csn = 1` and `validateInfo` threw `LOGICAL_ERROR` from the same place in `renameAndCommitEmptyParts` — leaving the same orphan `detached/` clone in a release build — and aborted the server in debug and sanitizer builds. The reachability of the orphan clone is unchanged; what changed is that the error is now a retryable `SERIALIZATION_ERROR` instead of a logical error or a crash. The window this pull request *did* newly open (a refused removal for a part selected at the start) is closed by 4a83a2de1e5, and `05161` asserts it with a control run.

**What the required action actually costs.** Making the precheck a reservation means holding one `DataPartsLock` across select → check → lock → `makeCloneInDetached` → `renameAndCommitEmptyParts`, i.e. cloning part data to `detached/` (disk I/O, potentially large) while holding the lock that every insert and every read of the part set needs. The alternative the review offers — move the clone after the removal can no longer throw — is the better shape and is already how the transactional path works (`dropPartsImpl` clones after `outdatePart`), but it reorders the partition-command path for all engines and all callers.

Either way it is a behavioural restructure of `ALTER TABLE … DROP/DETACH PARTITION`, not a narrowing of the abort fix, and it changes lock-hold characteristics on a hot path. I would rather not fold it into a bug fix that is green and whose regression is covered, and instead do it as a follow-up where it can be reviewed and perf-checked on its own. Happy to open that follow-up now, or to fold it in here if you would rather have it in one change — just say which.

src/Storages/StorageMergeTree.cpp
```diff
@@ -3018,6 +3018,11 @@ void StorageMergeTree::dropPart(const String & part_name, bool detach, ContextPt
             if (!part)
                 throw Exception(ErrorCodes::NO_SUCH_DATA_PART, "Part {} not found, won't try to drop it.", part_name);
 
+            /// `renameAndCommitEmptyParts` below can refuse to remove the part. Find that out before
+            /// `makeCloneInDetached` writes a copy to `detached/`, otherwise a failed DETACH leaves the
+            /// clone behind and every retry adds another `_tryN` directory next to it.
+            checkPartsCanBeRemovedNonTransactionally({part}, NonTransactionalRemovalKind::Discard);
```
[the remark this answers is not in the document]
alexey-milovidov: 🕵 Implemented in 9650edddec6, taking the second of the two options.

The premise is right and my previous reply was arguing the wrong point: the precheck samples the state, it does not reserve it, so the removal can still be refused after `makeCloneInDetached` has run. Rather than make it a reservation — which means holding one `DataPartsLock` across select → check → clone → commit, i.e. cloning part data to `detached/` while every insert and every read of the part set waits — the ordering now carries the guarantee.

The non-transactional `DETACH` path copies to `detached/` *after* `renameAndCommitEmptyParts` has committed, out of the parts it reports as covered, the same way the transactional path clones after `outdatePart` and the Replicated path clones inside the removal's own parts lock. So a refused removal cannot leave anything behind, whichever of the two checks refuses it, and there is no `_tryN` buildup on retry. As a side effect a `DETACH` that goes through now copies exactly the parts it removed rather than the ones selected earlier, which is what `getActivePartsToReplace` recomputing the covered set at commit time made diverge in the first place.

The precheck is kept, but only as a fast path: it still fails the common case before any empty part is built.

`05161` already asserts the negative half (a refused `DETACH` leaves `system.detached_parts` empty). The new `05176_detach_partition_clone_after_removal` asserts the positive half for both `DETACH PARTITION` and `DETACH PART` — the copies are made, re-attaching them brings every row back, and a `DROP` on the same path copies nothing. Verified locally together with `05161` / `05162` / `05175` and the 15 `detach_part` / `drop_part` / `detach_partition` tests in master, including `04622_detach_partition_fsync_part_directory`.

src/Interpreters/MergeTreeTransaction/VersionMetadata.cpp
alexey-milovidov: 🕵 Answered at pull-request level so it did not land in this thread — linking it here: https://github.com/ClickHouse/ClickHouse/pull/105116#issuecomment-5054303935

Short version: the `getCSN` window is real and described correctly, but it is not introduced here. Before this change the guard was `!info.creation_csn` alone, which is false-positive for that window *and* for the much longer period where a transaction committed long ago but this particular part was never stamped; the `TransactionLog::getCSN` fallback strictly shrinks it. And the behaviour being replaced in that window is not success — it is the `creation_csn is not set while removal_csn is set to 1` logical error, i.e. a server abort in debug and sanitizer builds.

Closing the remainder needs the CSN published synchronously at commit point, which is a change to the transaction-log commit protocol affecting every reader of `tid_to_csn`, not to the removal guard. `VersionMetadata` has no atomic `creation_csn` to hand `getCSNAndAssert` as a fallback either — `getInfo()` already reads the mutex-protected struct, so it is as fresh as the data gets; the value simply does not exist yet in that window. Happy to split that into its own pull request if you want it tracked.

src/Storages/StorageMergeTree.cpp
```diff
@@ -2918,6 +2918,21 @@ void StorageMergeTree::renameAndCommitEmptyParts(MutableDataPartsVector & new_pa
     if (deduplication_log)
         for (const auto & part : covered_parts)
             deduplication_log->dropPart(part->info);
+
+    return covered_parts;
```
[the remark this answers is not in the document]
alexey-milovidov: 🕵 You are right, and thank you — the precommit `covered_parts` is a selection, not the outcome. `MergeTreeData::Transaction::commit` reacquires `lockParts` and recomputes the covered set with `getActivePartsToReplace`, and when a covering part appears in that window it marks the empty part obsolete and removes nothing at all, while the old code still undelayed the selection with `remove_time = 0`, dropped it from the deduplication log, and handed it to `clonePartsToDetached`.

Adopted exactly as suggested in `a9d06b606129`: `renameAndCommitEmptyParts` now takes the `DataPartsVector` returned by `transaction.commit()` and uses it for the `remove_time = 0` loop, `deduplication_log->dropPart`, and the value returned to `dropPart` / `dropPartition` (hence for `clonePartsToDetached`). The `LOG_INFO` now reports both counts, so a divergence is visible in the log.

All three call sites run with `txn == nullptr`, so `commit` does not fold `covered_outdated_parts` into that vector — the returned set is exactly the parts this statement removed.

Verified locally: `05161`, `05162`, `05175`, `05176`, `04057`-`04060`, `01173` and the `detach_part` / `drop_part` / `detach_partition` families all pass; `check_cpp.sh` is clean.

alexey-milovidov:
🕵 Pass of 2026-09-12: the AI Review blocker is adopted, and the one red is unrelated and now fixed on `master`.

**Blocker — `renameAndCommitEmptyParts` returned the precommit selection. Adopted in `a9d06b606129`.**

The finding is right. `covered_parts` is accumulated by `renameTempPartAndReplaceUnlocked` under a parts lock that is released before `MergeTreeData::Transaction::commit` reacquires it and recomputes the covered set with `getActivePartsToReplace`. When a covering part appears in that window, `commit` marks the empty part obsolete and removes nothing at all, yet the old code still undelayed the stale selection with `remove_time = 0`, dropped it from the deduplication log, and — through the value handed to `dropPart` / `dropPartition` — cloned it into `detached/` while it was still active.

`commit` already returns `total_covered_parts`, so the fix is exactly the suggested one: take that vector and use it for all three. The `LOG_INFO` now prints both counts, so a divergence is visible in the log. All three call sites run with `txn == nullptr`, so `commit` does not fold `covered_outdated_parts` into the returned vector — it is exactly the set this statement removed.

**Tests note.** `05176` pins the invariant positively (`DETACH PARTITION` and `DETACH PART` clone exactly what was removed, re-`ATTACH` restores every row, `DROP` clones nothing), and `05161` pins the negative half. The divergence itself is not reachable from a test without fault injection: `lockOperationsWithParts` serialises the partition commands, an `INSERT` lands above the empty part's block range so it is never covered, and merges are held off by `stopMergesAndWait` — the only producer of a covering part in that window is a merge rolled back after `stopMergesAndWait` returned, which is why the code logs that case explicitly. Adding a failpoint in the merge commit path to reach it is a larger change than the fix; I would rather not grow this bug fix by it.

**`Upgrade check (amd_release)` — `Logical error: Virtual row has different type for A. Expected B, got C`.**

Not caused by this pull request, and already fixed on `master`:
- the abort is in `setVirtualRow` / `MergingSortedAlgorithm::initialize` over a sorting key of `CAST(json.b, 'String')` — read-in-order virtual rows, nothing to do with transactions or part removal;
- it is fleet-wide: 66 failures over 2026-09-10..2026-09-12 across 64 different pull requests, including `master` itself (`pull_request_number = 0`, first seen 2026-09-10 16:22 on `ad8622edd0f`);
- the fix is #119385 ("Fix wrong results when a sorting key expression's type depends on a session setting"), merged into `master` on 2026-09-12 10:39.

`master` is merged in (`5ae6d4c121f6`, clean, no conflicts), so the next run picks the fix up.

**Verification.** `ninja clickhouse-bundle` then `ninja clickhouse`, 0 errors; `check_cpp.sh` clean. Against a local server: `05161`, `05162`, `05175`, `05176`, `04057`–`04060`, `01173`, `03812` and the `detach_part` / `drop_part` / `detach_partition` / `drop_partition` families all pass (the only non-pass is `04622_detach_partition_fsync_part_directory`, which needs `system.part_log` that a minimal server does not have).

The `CH Inc sync` red was on the old head; the private mirror is `MERGEABLE` again after the master merge.

src/Storages/StorageMergeTree.cpp
```diff
@@ -2905,19 +2905,40 @@ void StorageMergeTree::renameAndCommitEmptyParts(MutableDataPartsVector & new_pa
...
-
     transaction.renameParts();
-    transaction.commit();
+
+    /// `covered_parts` above is only the precommit selection: `commit` reacquires the parts lock and
+    /// recomputes the covered set, so it is the only authoritative answer to "what was removed".
+    /// Everything below -- and the clone to `detached/` made by the callers -- must use that answer,
+    /// otherwise a concurrently appearing covering part makes us report, undelay and detach a part
+    /// that is still active.
+    DataPartsVector removed_parts = transaction.commit();
```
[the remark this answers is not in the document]
alexey-milovidov: 🕵 Adopted in `1afe40dbb74d`. The finding is right, and it is the mirror image of the previous round's: copying after the removal makes an ordinary I/O error destructive, copying before it leaves an orphan in `detached/` whenever the removal is then refused. Neither order is safe on its own, so the fix is to stop choosing between them.

`renameAndCommitEmptyParts` now takes `clone_to_detached` and does the whole thing under the one parts lock it already precommits the empty parts with: it checks removability, clones the covered parts, and calls `MergeTreeData::Transaction::commit(part_lock)` with that same lock instead of letting `commit` reacquire it. Because the lock is held across all three:

- the covered set cannot change between the check, the clone and the commit, so the copy is of exactly the parts that are removed — which is also what the previous round's finding was about;
- a removal that has to be refused is refused before anything is copied, so no orphan copy and no `_tryN` buildup on retries (`05161` asserts this, and it still passes);
- a clone that fails on I/O fails the statement with the parts still `Active` — the `Transaction` destructor rolls the empty parts back — so nothing is lost.

This is the "reservation" shape I argued against on 2026-09-12 00:41 on cost grounds, and that argument does not hold up: for a plain `MergeTree` the copy is a hardlink-based freeze, because `copy_instead_of_hardlink` in `makeCloneInDetached` is gated on `supportsReplication()`, which is false here. `StorageReplicatedMergeTree` already clones inside the removal's own parts lock (`removePartsInRangeFromWorkingSetAndGetPartsToRemoveFromZooKeeper`, `clone_to_detached`), so this aligns the two engines rather than inventing a new locking pattern.

One detail worth flagging: the `PART_IS_TEMPORARILY_LOCKED` retry loop now covers only the precommit renames — a `batch_precommitted` flag rethrows anything raised by the clone or the commit, since retrying those after a partial precommit would be wrong.

Verified: `ninja clickhouse-bundle` and `ninja clickhouse`, 0 errors; `check_cpp.sh` clean. Against a local server, `05161`, `05162`, `05175`, `05176`, `01451_detach_drop_part`, `01451_replicated_detach_drop_part_long`, `01710_projection_detach_part`, `00516_deduplication_after_drop_partition_zookeeper`, `03285_materialize_ttl_only_drop_parts`, `01173_transaction_control_queries` and the `04057`–`04060` transaction tests all pass (26/26 after merging `master`).

alexey-milovidov:
🕵 Pass of 2026-09-12 (third): the new AI Review blocker is adopted, `master` is merged in, and the tree is green locally.

**Blocker — post-commit clone made `DETACH` exception-unsafe. Adopted in `1afe40dbb74d`.**

The previous round moved the `detached/` copy after the removal; this round points out that this made an ordinary `makeCloneInDetached` I/O failure destructive — empty covering parts active, sources `Outdated` with `remove_time = 0`, only a prefix of the partition under `detached/`. Both findings are real and they point in opposite directions, so rather than pick an order, the copy and the removal now happen under the same lock.

`renameAndCommitEmptyParts` takes a `clone_to_detached` flag and, while still holding the parts lock it precommitted the empty parts under, checks removability, clones the covered parts, and calls `MergeTreeData::Transaction::commit(part_lock)` with that lock instead of letting `commit` reacquire it. The covered set cannot change in between, so the copy is of exactly the parts that are removed; a refused removal is refused before anything is copied (no orphan, no `_tryN` buildup — `05161`'s assertion still holds); and a failed copy fails the statement with the parts still active, the `Transaction` destructor rolling the empty parts back.

I had argued on 2026-09-12 00:41 that this shape costs too much because it holds `DataPartsLock` across a data copy. That was wrong for this path: `copy_instead_of_hardlink` in `makeCloneInDetached` is gated on `supportsReplication()`, so for a plain `MergeTree` the copy is a hardlink-based freeze. `StorageReplicatedMergeTree` clones inside the removal's own parts lock already, so all three engines now do the same thing.

The remaining Major (the `TransactionLog::getCSN` window) is `[dismissed by author]` in the verdict itself — it is pre-existing and strictly narrowed here, as adjudicated at https://github.com/ClickHouse/ClickHouse/pull/105116#issuecomment-5054303935.

**CI.** No failures on `5ae6d4c1` — the `Upgrade check` red of the previous pass is gone (it was fleet-wide and fixed by #119385). The only non-green check is `CH Inc sync`, which is pending; `master` is merged in again (`b35849825ea3`, clean, no conflicts) so it re-bases on a current head.

**Verification.** `ninja clickhouse-bundle` then `ninja clickhouse`, 0 errors; `check_cpp.sh` clean. Against a local server after the `master` merge: `05161`, `05162`, `05175`, `05176`, `01451_detach_drop_part`, `01451_replicated_detach_drop_part_long`, `01710_projection_detach_part`, `00516_deduplication_after_drop_partition_zookeeper`, `03285_materialize_ttl_only_drop_parts`, `01173_transaction_control_queries`, `04057`–`04060` — 26 of 26 pass.

clickgapai:
<details>
<summary>Reproducer</summary>

_Shell reproducer — set `BIN` to a ClickHouse binary path and run with bash; requires the referenced test fixtures from the repo._
```bash
#!/bin/bash
BIN="${BIN:-./clickhouse}"  # path to the ClickHouse binary under test
# PR #105116: non-transactional DROP PARTITION over a part whose creating
# transaction has not committed yet stamps removal_csn=1 on creation_csn=0,
# which VersionMetadata::validateInfo rejects (LOGICAL_ERROR / abort on
# debug+sanitizer builds). After the fix the statement is refused with
# SERIALIZATION_ERROR and the part stays active.
set -uo pipefail

D=$(mktemp -d /tmp/ch105116.XXXXXX)
TCP=$((21000 + RANDOM % 15000))
HTTP=$((TCP + 1))
KPR=$((TCP + 2))
RAFT=$((TCP + 3))

cleanup() { [ -n "${SRV:-}" ] && kill -9 "$SRV" 2>/dev/null; rm -rf "$D"; }
trap cleanup EXIT

mkdir -p "$D/data" "$D/tmp" "$D/user_files" "$D/access" "$D/fmt" "$D/coord"
cat > "$D/config.xml" <<XML
<clickhouse>
    <logger><level>warning</level><console>0</console><log>$D/server.log</log><errorlog>$D/err.log</errorlog></logger>
    <listen_host>127.0.0.1</listen_host>
    <tcp_port>$TCP</tcp_port>
    <http_port>$HTTP</http_port>
    <path>$D/data/</path>
    <tmp_path>$D/tmp/</tmp_path>
    <user_files_path>$D/user_files/</user_files_path>
    <format_schema_path>$D/fmt/</format_schema_path>
    <access_control_path>$D/access/</access_control_path>
    <mark_cache_size>268435456</mark_cache_size>
    <users_config>users.xml</users_config>
    <default_profile>default</default_profile>
    <default_database>default</default_database>

    <keeper_server>
        <tcp_port>$KPR</tcp_port>
        <server_id>1</server_id>
        <log_storage_path>$D/coord/log</log_storage_path>
        <snapshot_storage_path>$D/coord/snapshots</snapshot_storage_path>
        <coordination_settings>
            <operation_timeout_ms>10000</operation_timeout_ms>
            <session_timeout_ms>100000</session_timeout_ms>
            <startup_timeout>60000</startup_timeout>
            <heart_beat_interval_ms>0</heart_beat_interval_ms>
            <election_timeout_lower_bound_ms>0</election_timeout_lower_bound_ms>
            <election_timeout_upper_bound_ms>0</election_timeout_upper_bound_ms>
        </coordination_settings>
        <raft_configuration>
            <server><id>1</id><hostname>127.0.0.1</hostname><port>$RAFT</port></server>
        </raft_configuration>
    </keeper_server>
    <zookeeper>
        <node><host>127.0.0.1</host><port>$KPR</port></node>
    </zookeeper>

    <allow_experimental_transactions>42</allow_experimental_transactions>
    <transaction_log><zookeeper_path>/test/clickhouse/txn</zookeeper_path></transaction_log>
</clickhouse>
XML
cat > "$D/users.xml" <<XML
<clickhouse>
    <profiles><default><async_insert>0</async_insert></default></profiles>
    <quotas><default></default></quotas>
    <users><default>
        <password></password>
        <networks><ip>::/0</ip></networks>
        <profile>default</profile><quota>default</quota>
        <access_management>1</access_management>
    </default></users>
</clickhouse>
XML

"$BIN" server --config-file="$D/config.xml" >"$D/stdout.log" 2>&1 &
SRV=$!

up=0
for _ in $(seq 1 180); do
    if "$BIN" client --host 127.0.0.1 --port "$TCP" -q "SELECT 1" >/dev/null 2>&1; then up=1; break; fi
    kill -0 "$SRV" 2>/dev/null || break
    sleep 0.5
done
[ "$up" = 1 ] || { echo "OBSERVED: server did not start"; tail -5 "$D/err.log" "$D/stdout.log" 2>/dev/null; exit 2; }

q()   { timeout 60 "$BIN" client --host 127.0.0.1 --port "$TCP" -q "$1" 2>&1; }
# Transactional statements need a stateful session -> HTTP with session_id.
tx()  { timeout 60 curl -sS --max-time 55 "http://127.0.0.1:$HTTP/?session_id=r105116&session_timeout=600" --data "$1" 2>&1; }

setup="$(q "CREATE TABLE t105116 (x UInt64) ENGINE = MergeTree ORDER BY x")"
setup+="$(tx 'BEGIN TRANSACTION')"
setup+="$(tx 'INSERT INTO t105116 SETTINGS async_insert = 0 VALUES (1)')"
if echo "$setup" | grep -qiE "exception|Code: [0-9]+|NOT_IMPLEMENTED|UNKNOWN_"; then
    echo "OBSERVED: transaction setup unsupported/failed: $(echo "$setup" | head -3)"
    exit 2
fi

# The part inserted above is active but its creating transaction is still open.
out="$(q "ALTER TABLE t105116 DROP PARTITION ALL")"
rc=$?
echo "OBSERVED: DROP PARTITION -> rc=$rc out=$(echo "$out" | head -2 | tr '\n' ' ')"

if ! kill -0 "$SRV" 2>/dev/null; then
    echo "OBSERVED: server died during DROP PARTITION"
    grep -m1 -oE "creation_csn is not set while removal_csn is set to [0-9]+|LOGICAL_ERROR|Aborted" "$D/err.log" "$D/stdout.log" 2>/dev/null | head -3
    exit 1
fi
if echo "$out" | grep -qE "creation_csn is not set while removal_csn is set|LOGICAL_ERROR|Code: 49"; then
    echo "OBSERVED: DROP PARTITION failed with LOGICAL_ERROR over the uncommitted part"
    exit 1
fi
if echo "$out" | grep -qE "SERIALIZATION_ERROR|Code: 279"; then
    echo "OBSERVED: refused with retryable SERIALIZATION_ERROR"
    exit 0
fi
if [ "$rc" = "124" ]; then echo "OBSERVED: timed out"; exit 2; fi
if echo "$out" | grep -qiE "exception|Code: [0-9]+"; then
    echo "OBSERVED: unrelated error"; exit 2
fi
# No error: the pre-regression behaviour (the part is just removed). Not this bug.
echo "OBSERVED: DROP PARTITION returned no error"
exit 0
```

</details>

**Results:**
- **Culprit PR:** Refactor VersionMetadata #92141
- **Reproduces on**: `26.8, 26.7, 26.6`
- **Does not reproduce on**: `master, 26.3`
- **Tested versions**: `master, 26.8, 26.7, 26.6, 26.3`
- **Bisect anchor** (below the supported window, not a backport target): `26.2` clean
- ✓ **Fix is on master** — this PR's change fixes the bug on master. The releases listed above still have the bug and need the backport.

Backport the fix to `26.8`, `26.7`, `26.6`.


CC introducing PR reviewer: @tiandiwonder @tavplubix

CC component owner: @CurtizJ @alexey-milovidov

<details>
<summary>Analysis metadata</summary>

`comp-mergetree` · Severity `P2` · Finding `phase_d_pr105116`

- Verified against `master` build `044e813459c6` (aarch64)
- **Caused by:** #92141 (commit `f2f8540ef117`, found by bisecting in 11 steps)
- First shipped in release line: `26.6` (also in 26.7, 26.8)



- Suggested assignees (from introducing PR #92141):  @tiandiwonder @tavplubix


</details>

PR #77128, opened 2025-03-04
Impl Drivers for User-Defined Functions

dahbka-lis:
### Changelog category (leave one):
- New Feature

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Expand `CREATE FUNCTION` query for easy creating User-Defined Functions and its execution by some driver like #71172.
...

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #105131, opened 2026-05-16
Drivers for executable user-defined functions

### LINKED ISSUE #71172: Drivers for User-Defined Functions
**Use case**

Allow a user to write code or configuration for a function in an arbitrary language directly in the CREATE FUNCTION statement.

So, instead of configuring a fixed, predefined UDF as in https://clickhouse.com/docs/en/sql-reference/functions/udf, the user will rely on a driver, that generates them from a provided code snippet.

**Describe the solution you'd like**

The user provides the code in the form of an inline snippet. It can look as follows:
```
CREATE FUNCTION my_function (arg1 Type1, arg2 Type2) RETURNS Type3
ENGINE = DriverName AS
$$
    code is here
    ...
$$
```

Drivers are defined in the server configuration. A driver is responsible for the interpretation of the code (how to transform a snippet into an executable file), its security (it could place the code inside Docker, nsjail, gvisor, or even on a separate machine).

The interface and configuration of the driver can be opaque and represented by:
- a command to run to create a new UDF, which accepts the code snippet in stdin and prints to stdout a command to run the created actual UDF;
- a command to delete a UDF;

The user-provided code snippet is untrusted, but the driver itself is the trusted part of the infrastructure. Its responsibility is to create necessary files and provide a command to run the user code in a secure way.

**Additional context**

This solution is very generic, but it should be provided with examples, e.g., how to build a UDF driver to run JavaScript functions using Node.js inside Docker.

alexey-milovidov:
Implements the feature requested in https://github.com/ClickHouse/ClickHouse/issues/71172 - operator-supplied "drivers" that turn a user code snippet into an executable user-defined function. Supersedes the draft in https://github.com/ClickHouse/ClickHouse/pull/77128 (which used a different, mostly-incorrect design).

A driver is described by an XML/YAML file under `<user_defined_executable_function_drivers_config>`. It declares:

- `name` - the driver name as used in `CREATE FUNCTION ... ENGINE = <name>(...)`
- `create_command` - shell script invoked to create a UDF from a code snippet
- `drop_command` (optional) - invoked when the function is dropped
- `engine_arguments` (optional) - allowed args inside `ENGINE = Driver(a='x', b=1)`, with `<required>` flag
- `env` (optional) - environment variables passed to the driver

New SQL form:

```sql
CREATE [OR REPLACE] FUNCTION [IF NOT EXISTS] name [ON CLUSTER c]
    ARGUMENTS (a UInt8, b String) RETURNS UInt64
    ENGINE = DriverName(key1 = 'value1', key2 = 42)
    AS '...code body...'
```

`CREATE FUNCTION` invokes `create_command` with `--name`, `--return`, `--args`, plus any declared engine arguments, sets the configured env vars, sends the user body on stdin, and reads the resulting executable UDF configuration from stdout. Format is auto-detected (XML if it starts with `<`, otherwise YAML) and written into `<dynamic_user_defined_executable_functions_path>`; the existing executable UDF loader picks it up. The originating SQL is persisted as `ATTACH FUNCTION ...` in `<user_defined>` so it survives restart. On server start, if a persisted ATTACH has no matching file in the dynamic directory, the driver is re-run.

`DROP FUNCTION` invokes `drop_command` (if present), removes the dynamic config and the per-function working directory, reloads the UDF loader, and removes the persisted SQL.

If `create_command` exits non-zero, an exception is thrown that includes the exit code and the full driver stderr; the per-function working directory is removed so failed attempts leave no state.

Includes a proof-of-concept driver `c_function_body`: takes a C function body, generates a TabSeparated stdin/stdout wrapper, compiles it inside a sandboxed Docker container, and emits an `executable_pool` UDF that runs the compiled binary inside another sandboxed Docker container (`--network=none --read-only --cap-drop=ALL --security-opt=no-new-privileges --user 65534:65534` + memory/cpu/pids limits). `CLICKHOUSE_C_DRIVER_FORCE_LOCAL=1` switches to direct compile + run for environments without Docker (used by the tests).

Stateless tests:

- `04241_executable_udf_driver` - create, call, simulated config-loss recreate, drop, file cleanup
- `04242_executable_udf_driver_errors` - non-zero exit code + multi-line stderr propagation, empty driver output detection, unknown driver detection, absence of leftover state on failure

- Experimental Feature

Added "drivers" for executable user-defined functions. A driver declared in `<user_defined_executable_function_drivers_config>` can be used in `CREATE FUNCTION name ARGUMENTS (...) RETURNS T ENGINE = DriverName(...) AS '...code...'` to compile or otherwise process a user code snippet at function-creation time and produce a runnable executable UDF. The resulting configuration is stored in `<dynamic_user_defined_executable_functions_path>`, and the originating query is persisted as `ATTACH FUNCTION` so the function survives server restarts. A proof-of-concept `c_function_body` driver compiles and runs C function bodies inside sandboxed Docker containers in `executable_pool` mode.

- [x] Documentation is written (mandatory for new features) — added a "Driver-based Executable User Defined Functions" section to `docs/en/sql-reference/functions/udf.md`.

---

> [!NOTE]
> **High Risk**
> Adds a new SQL pathway that executes operator-configured external driver commands, writes/loads generated UDF configs from disk, and changes startup/reload behavior; this touches execution, filesystem, and security-sensitive surfaces.
> 
> **Overview**
> Adds *driver-based executable UDFs*: a new `CREATE/ATTACH FUNCTION ... ENGINE = DriverName(...) AS 'code'` SQL form that invokes an operator-configured `create_command`, persists the generated XML/YAML UDF config under `dynamic_user_defined_executable_functions_path`, and stores the originating query as `ATTACH FUNCTION` for restart durability.
> 
> Introduces a driver registry loaded from `<user_defined_executable_function_drivers_config>` (on startup and config reload), wires it into `clickhouse-server` and `clickhouse-local`, and adds startup recovery that re-runs drivers for persisted functions whose dynamic configs are missing.
> 
> Extends the executable UDF loader to watch the dynamic config directory, supports per-function working directories for direct execution, and adds optional `<command_pipe_capacity>` plumbing down to `ShellCommand` pipe sizing.
> 
> Includes proof-of-concept C-body drivers/scripts and stateless tests covering create/call/recreate/drop flows plus failure cleanup and error propagation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5576b9ed5bd65527f5e7b39a694f17d796458059. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/Functions/UserDefined/UserDefinedExecutableFunctionDriverInvoker.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+namespace ErrorCodes
+{
+    extern const int UDF_EXECUTION_FAILED;
+}
+
+namespace
+{
+    /// Shell-quote a single argument so it can be embedded into a `/bin/sh -c` command.
+    String shellQuote(const String & str)
```
alexey-milovidov: It will be much safer to not use `sh`, and do direct exec.

alexey-milovidov:
@vdimir, I compared this implementation with WASM on the example from the docs:

```
CREATE FUNCTION collatz_steps_wasm LANGUAGE WASM ARGUMENTS (n UInt32) RETURNS UInt32 FROM 'collatz' :: 'steps';

SELECT max(collatz_steps_wasm(number::UInt32)) FROM numbers_mt(100000000);

Elapsed: 148.449 sec. Processed 100.00 million rows, 800.00 MB (673.63 thousand rows/s., 5.39 MB/s.
```

For comparison, with drivers:

```
CREATE FUNCTION collatz_steps_unsafe
  ARGUMENTS (n UInt32)
  RETURNS UInt32
  ENGINE = UnsafeC()
  AS $$
      uint32_t count = 0;

      if (n < 1)
          return 0;

      while (n != 1)
      {
          n = (n & 1) ? (3 * n + 1) : (n / 2);
          ++count;
      }

      return count;
  $$;

SELECT max(collatz_steps_unsafe(number)) FROM numbers_mt(100000000);

Elapsed: 0.505 sec. Processed 100.00 million rows, 800.00 MB (198.20 million rows/s., 1.59 GB/s.);

CREATE FUNCTION collatz_steps_docker
  ARGUMENTS (n UInt32)
  RETURNS UInt32
  ENGINE = DockerC()
  AS $$
      uint32_t count = 0;

      if (n < 1)
          return 0;

      while (n != 1)
      {
          n = (n & 1) ? (3 * n + 1) : (n / 2);
          ++count;
      }

      return count;
  $$;

SELECT max(collatz_steps_docker(number)) FROM numbers_mt(100000000);

Elapsed: 0.481 sec. Processed 100.00 million rows, 800.00 MB (208.01 million rows/s., 1.66 GB/s.)

CREATE FUNCTION collatz_steps_gvizor
  ARGUMENTS (n UInt32)
  RETURNS UInt32
  ENGINE = GVisorC()
  AS $$
      uint32_t count = 0;

      if (n < 1)
          return 0;

      while (n != 1)
      {
          n = (n & 1) ? (3 * n + 1) : (n / 2);
          ++count;
      }

      return count;
  $$;

SELECT max(collatz_steps_gvizor(number)) FROM numbers_mt(100000000);

Elapsed: 0.498 sec. Processed 100.00 million rows, 800.00 MB (200.66 million rows/s., 1.61 GB/s.)
```

vdimir:
> @vdimir, I compared this implementation with WASM on the example from the docs:

Thanks for testing it out.

This example uses `ROW_DIRECT` ABI, which may be not too performant as BUFFERED_V1 which uses block processing.

I tried it using the buffered ABI with MessagePack serialization (the default at the moment). It’s faster, though still not as good as native code, around 10–20x slower, but still an improvement.

https://gist.github.com/vdimir/eb7cbcfef960972fad8fe41014bae37a

```sql
SELECT max(collatz_steps(number))
FROM numbers_mt(100000000)
SETTINGS webassembly_udf_max_fuel = 18446744073709551615

Query id: 123ef5b8-2d1d-46ee-89f8-aaf22d8b2d43

1 row in set. Elapsed: 10.389 sec. Processed 100.00 million rows, 800.00 MB (9.63 million rows/s., 77.00 MB/s.)
Peak memory usage: 16.33 MiB.
```

And a fair amount of the overhead is still coming from data deserialization, which I assume we can improve:

```sql
SELECT
    ProfileEvents.1 AS name,
    round((ProfileEvents.2) / 1000) AS ms
FROM system.query_log
ARRAY JOIN ProfileEvents
WHERE (event_date = today())
  AND (query_id = '123ef5b8-2d1d-46ee-89f8-aaf22d8b2d43')
  AND (name LIKE 'Wasm%Microseconds')

Query id: 3887906d-5dcb-49f5-9075-f796173d4af6

   ┌─name────────────────────────────┬─────ms─┐
1. │ WasmSerializationMicroseconds   │   5070 │
2. │ WasmDeserializationMicroseconds │ 241743 │
3. │ WasmGuestExecuteMicroseconds    │  83892 │
4. │ WasmTotalExecuteMicroseconds    │ 330698 │
   └─────────────────────────────────┴────────┘
```

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverInvoker.cpp
alexey-milovidov: Switched to `executeDirect` with `/usr/bin/env -C <workdir> KEY=VAL <command> --name ...` in b4119b77030; argv is passed as distinct entries (no shell).

src/Parsers/ParserCreateFunctionQuery.cpp
```diff
@@ -53,6 +64,145 @@ bool ParserCreateFunctionQuery::parseImpl(IParser::Pos & pos, ASTPtr & node, Exp
...
+
+            if (!s_lparen.ignore(pos, expected))
+                return false;
+
+            if (s_rparen.ignore(pos, expected))
+            {
+                arguments_ast = make_intrusive<ASTExpressionList>();
+            }
+            else if (ParserNameTypePairList{}.parse(pos, arguments_ast, expected)
+                || ParserTypeList{}.parse(pos, arguments_ast, expected))
```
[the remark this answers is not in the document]
vdimir: Should be fune, for some formats column names do not make difference, so we accept both options.

tests/queries/0_stateless/04241_executable_udf_driver.sh
```diff
@@ -0,0 +1,242 @@
...
+        <create_command>${DRIVER_DIR}/unsafe_c_create.sh</create_command>
+        <drop_command>${DRIVER_DIR}/unsafe_c_drop.sh</drop_command>
+    </driver>
+</clickhouse>
+EOF
+
+mkdir -p "$WORK_DIR/user_defined" "$WORK_DIR/user_scripts" "$WORK_DIR/dyn" "$WORK_DIR/data"
+
+run() {
+    # Filter out shutdown-related noise (e.g. shell-printed "Segmentation fault" lines from
```
vdimir: Why shell can print "Segmentation fault" that we can ignore?

src/Interpreters/Context.cpp
```diff
@@ -1811,13 +1825,19 @@ void Context::setUserScriptsPath(const String & path)
...
-        auto & function_storage = getUserDefinedSQLObjectsStorage();
-        function_storage.loadObjects();
         UserDefinedSQLFunctionFactory::instance().loadFunctions(function_storage, *wasm_module_manager);
-    }
+}
+
+void Context::setDynamicUserDefinedExecutableFunctionsPath(const String & path)
+{
+    std::lock_guard lock(shared->mutex);
+    shared->dynamic_user_defined_executable_functions_path = path;
```
vdimir, line 1893: I'd add some guardrails for a a misconfigured path. For example we do in some places somethig like `std::filesystem::remove_all(working_dir, ec);`, can't it possible to overwrite ore remove some a arbitrary files matching function name?

src/Interpreters/InterpreterDropFunctionQuery.cpp
```diff
@@ -2,23 +2,195 @@
...
+        if (!driver)
+            LOG_WARNING(log, "Driver '{}' for function '{}' is not registered; skipping driver drop_command",
+                create_query.engine_name, function_name);
+
+        String dynamic_dir = current_context->getDynamicUserDefinedExecutableFunctionsPath();
+        if (!dynamic_dir.empty() && !dynamic_dir.ends_with('/'))
+            dynamic_dir.push_back('/');
+
+        const String escaped = escapeForFileName(function_name);
+        const String working_dir = dynamic_dir.empty() ? String() : readDriverWorkingDirectory(current_context, function_name);
```
vdimir: If config changed via config reload between create and drop we may hit unrelated directories here and try to remove them. Overall it's up to uder to make sure all config changes are valid, but still maybe its possible to persisit path somewhere or at least add some checks that path still corresponds to the function

programs/server/config.xml
```diff
@@ -1583,6 +1583,16 @@
     <!-- Configuration of user defined executable functions -->
     <user_defined_executable_functions_config>*_function.*ml</user_defined_executable_functions_config>
 
+    <!-- Configuration of drivers for executable user-defined functions. -->
+    <!-- A driver knows how to take a user-supplied code snippet and produce a runnable executable -->
+    <!-- user-defined function. Source-tree examples under this glob are not installed by packages; -->
+    <!-- operators can provide driver configs at the same config key. -->
+    <user_defined_executable_function_drivers_config>user_defined_executable_function_drivers_config.d/*_driver.xml</user_defined_executable_function_drivers_config>
```
vdimir, line 1626: Do we want to enable it in example config by default?

tests/queries/0_stateless/04241_executable_udf_driver.sh
alexey-milovidov: This is WTF, will remove.

alexey-milovidov:
Merged `master` (was ~1448 commits behind; clean merge, no conflicts) and pushed a commit addressing a batch of review feedback. The only CI failure was the `BuzzHouse (amd_tsan)` logical error in `StorageMemory::mutate` (`Mutation of `Memory` table produced incomplete output`, STID 2380-3043), which is unrelated to this PR and already being fixed in #105286.

### Addressed in this push

- **Experimental gate.** Added the server setting `allow_experimental_executable_udf_drivers` (disabled by default, mirroring `allow_experimental_webassembly_udf`). When disabled, no drivers are loaded and `CREATE FUNCTION ... ENGINE = DriverName(...)` throws `SUPPORT_IS_DISABLED`. This is the guard requested for the unconditional create path, and it lets the remaining hardening items below be iterated on while the feature is experimental.
- **Replicated storage rejection.** `CREATE FUNCTION ... ENGINE = Driver(...)` now throws when `getUserDefinedSQLObjectsStorage().isReplicated()`, since artifacts are materialized only on the receiving replica (also covers the `RESTORE` concern). This is the conservative option from the replication threads.
- **Require `AS`.** The parser now requires `AS '...code...'` for driver-based `CREATE FUNCTION`; omitting it is a syntax error instead of a side-effectful create with an empty body.
- **`IF NOT EXISTS`.** The executable-UDF existence check now honors `throw_if_exists` instead of throwing `FUNCTION_ALREADY_EXISTS` unconditionally (Cursor finding).
- **Config-reload registry clearing.** Removing drivers from config (or emptying the config) on `SYSTEM RELOAD CONFIG` now clears the registry immediately instead of leaving stale drivers usable until restart.
- **`ON CLUSTER`.** `ASTCreateFunctionWithDriverQuery` is now part of `isUserDefinedFunctionQuery` in `removeOnClusterClauseIfNeeded`, so the new path honors `ignore_on_cluster_for_replicated_udf_queries`.
- **Sandbox hardening.** Added `--security-opt=no-new-privileges` to the `c_function_body` driver's compile and runtime `docker run` invocations.

Built locally; `04241_executable_udf_driver` and `04242_executable_udf_driver_errors` pass, and I verified the gate-off rejection and the require-`AS` syntax error manually.

### Not changed (deferred design items)

- **Transactional create/drop.** The atomic-swap rollback for `CREATE OR REPLACE` / `DROP` failures (write to temp paths, commit metadata, then swap; clean only per-attempt artifacts) is a larger change. With the experimental gate in place I left it for a follow-up — happy to take it on next if you want.
- **Full replication / `RESTORE` materialization.** Rejecting on replicated storage (above) is the conservative choice; materializing artifacts on every replica is the larger alternative.
- **Runtime `--read-only` workdir mount and `GVisorC` resource limits.** Sandbox refinements in `c_driver_common.py`.
- **Helper dedup** between `InterpreterCreateFunctionQuery.cpp` and `InterpreterDropFunctionQuery.cpp` (Cursor finding).

### On the `no-parallel` threads

I kept `no-parallel` on both stateless tests. The bots suggested removing it, but the tests document that they write to dirs that are not unique per test (e.g. the dynamic config / user_defined dirs derived from the work dir are shared), so it is genuinely needed here.

alexey-milovidov:
Merged `master` into the branch (was 487 commits behind and red). All changed translation units (interpreters, driver invoker/registry, loader, parser, `ShellCommand`, `ServerSettings`, `Context`) compile cleanly against current `master` — no API drift from the merge.

CI failures on the previous run are unrelated to this PR:
- `Unit tests / FunctionsStress.stress` — known flaky, tracked in https://github.com/ClickHouse/ClickHouse/issues/104877 (`function_prop_fuzzer`, unrelated to UDF drivers).
- `Performance Comparison` (`rand`, `formats_columns_nullable`, `insert_select_default_small_block`) — flagged `slower` on both `::old` and `::new`, i.e. measurement noise, not a real regression.

The open design threads (transactional create/drop atomicity, fail-close on `drop_command`/reload failure, replicated-storage and `RESTORE` artifact lifecycle, old-analyzer visitor) are left for the author to decide — they are architectural decisions on this in-progress feature.

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverInvoker.cpp
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    CommandResult writeAndWait(ShellCommand & process, const String & source_code)
+    {
+        /// All three pipes must be serviced concurrently: a driver that prints
+        /// enough on stderr while waiting for stdin would otherwise deadlock if
+        /// we wrote stdin synchronously before draining stdout/stderr.
+        auto stdout_future = std::async(std::launch::async, [&process] { return readPipeToString(process.out); });
+        auto stderr_future = std::async(std::launch::async, [&process] { return readPipeToString(process.err); });
+        auto stdin_future = std::async(std::launch::async, [&process, &source_code] { return writeStdinSafely(process.in, source_code); });
```
antonio2368: This doesn't use our thread pool, can we use ThreadPoolCallbackRunnerLocal for 2 operations and do write in this thread?

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverInvoker.cpp
```diff
@@ -0,0 +1,285 @@
...
+        {
+            std::rethrow_exception(result.write_exception);
+        }
+        catch (Exception & e)
+        {
+            e.addMessage(fmt::format(
+                "while writing function body to driver '{}' create_command for function '{}'. Stderr: {}",
+                driver.name, function_name, result.stderr_output));
+            throw;
+        }
```
antonio2368: what do you think about something like 
```cpp
      if (result.write_exception)
          throw Exception(ErrorCodes::UDF_EXECUTION_FAILED,
              "Failed while writing function body to driver '{}' create_command for function '{}': {}. Stderr: {}",
              driver.name, function_name,
              getExceptionMessage(result.write_exception, /*with_stacktrace=*/ false),
              result.stderr_output);
```

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverRegistry.cpp
```diff
@@ -0,0 +1,154 @@
...
+        return nullptr;
+    return it->second;
+}
+
+UserDefinedExecutableFunctionDriverPtr UserDefinedExecutableFunctionDriverRegistry::get(const String & driver_name) const
+{
+    std::lock_guard lock(mutex);
+    auto it = drivers.find(driver_name);
+    if (it != drivers.end())
+        return it->second;
```
antonio2368:
```cpp
    if (auto driver = tryGet(driver_name))
        return driver;
```

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverRegistry.cpp
```diff
@@ -0,0 +1,154 @@
...
+    {
+        if (!registered_names.empty())
+            registered_names += ", ";
+        registered_names += name;
+    }
+
+    throw Exception(ErrorCodes::UNKNOWN_FUNCTION,
+        "Executable user-defined function driver '{}' is not registered. Registered drivers: {}",
+        driver_name,
+        registered_names);
```
antonio2368:
```cpp
    throw Exception(
        ErrorCodes::UNKNOWN_FUNCTION,
        "Executable user-defined function driver '{}' is not registered. Registered drivers: {}",
        driver_name,
        fmt::join(std::views::keys(drivers), ", "));
```

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
```diff
@@ -235,8 +240,6 @@ class UserDefinedFunction final : public IFunction
             ColumnWithTypeAndName column_to_cast = {column_with_type.column, column_with_type.type, column_with_type.name};
             column_with_type.column = castColumnAccurate(column_to_cast, argument_type);
             column_with_type.type = argument_type;
-
-            column_with_type = std::move(column_to_cast);
```
antonio2368, line 239: Why? Can this break things as we actually apply cast now?

src/Interpreters/InterpreterCreateFunctionQuery.cpp
```diff
@@ -18,6 +40,304 @@ namespace DB
...
+            if (it == driver.engine_arguments.end())
+                throw Exception(ErrorCodes::BAD_ARGUMENTS,
+                    "Driver '{}' does not declare engine argument '{}'", driver.name, name);
+
+            const auto * literal = value_ast->as<ASTLiteral>();
+            if (!literal)
+                throw Exception(ErrorCodes::BAD_ARGUMENTS,
+                    "Engine argument '{}' must be a literal", name);
+
+            result.emplace_back(name, applyVisitor(FieldVisitorToString(), literal->value));
```
antonio2368: doesn't this return string with quotes?

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverInvoker.cpp
alexey-milovidov: Done in af466586092: `writeAndWait` now drains stdout/stderr via `ThreadPoolCallbackRunnerLocal` over a local `ThreadPool` (new `UDFDriverInvokerThreads*` metrics, `UDF_DRIVER` thread name) and writes stdin in the calling thread.

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverInvoker.cpp
alexey-milovidov: Done in af466586092 - applied your suggestion for `write_exception` and the same shape for `wait_exception` (the previous `addMessage` approach also silently skipped non-`DB::Exception` errors).

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverRegistry.cpp
alexey-milovidov: Done in af466586092.

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverRegistry.cpp
alexey-milovidov: Done in af466586092.

src/Interpreters/InterpreterCreateFunctionQuery.cpp
alexey-milovidov: Good catch - it did. Fixed in 462b4b3cbde: string literals are now passed verbatim (`engineArgumentToString` in the new shared `UserDefinedExecutableFunctionDriverUtils`), and `drop_command` now receives the same values as `create_command` (it previously used the quoted form too). Verified end-to-end: `ENGINE = echo_driver(opt = 'hello world', num = 42)` reaches the driver as `--opt hello world --num 42`.

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
alexey-milovidov: The removed line was overwriting the freshly cast column with the saved original (`column_to_cast` holds the pre-cast column), so `castColumnAccurate` was computed and then discarded - declared argument types were never actually applied for `execute_direct` functions. Driver-generated wrappers rely on receiving exactly the declared types (the `04241` test covers `ARGUMENTS (x UInt64, y UInt64)` called with `UInt8` literals). For existing executable UDFs this only changes behavior where the declared argument type differs from the passed column type, in which case the previous behavior (sending the unconverted values) contradicted the documented meaning of the declared types.

tests/queries/0_stateless/04241_executable_udf_driver.sh
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env bash
+# Tags: no-fasttest, no-parallel, no-random-settings
```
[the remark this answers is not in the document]
alexey-milovidov: Done in fad897d1376 - removed the tag from both tests.

tests/queries/0_stateless/04242_executable_udf_driver_errors.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
+# Tags: no-fasttest, no-parallel, no-random-settings
```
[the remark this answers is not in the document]
alexey-milovidov: Done in fad897d1376.

src/Interpreters/InterpreterCreateFunctionQuery.cpp
```diff
@@ -18,6 +39,298 @@ namespace DB
...
+        auto & storage = current_context->getUserDefinedSQLObjectsStorage();
+        if (storage.has(function_name))
+        {
+            if (throw_if_exists)
+                throw Exception(ErrorCodes::FUNCTION_ALREADY_EXISTS, "User-defined object '{}' already exists", function_name);
+            return replace_if_exists;
+        }
+
+        if (UserDefinedExecutableFunctionFactory::instance().has(function_name, current_context)) /// NOLINT(readability-static-accessed-through-instance)
+            throw Exception(ErrorCodes::FUNCTION_ALREADY_EXISTS, "User defined executable function '{}' already exists", function_name);
```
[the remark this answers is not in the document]
alexey-milovidov: The `throw_if_exists` guard was added earlier; 462b4b3cbde completes the semantics: `IF NOT EXISTS` is now a no-op when an executable or wasm function with the same name exists (previously the driver was still invoked and could shadow it), and `OR REPLACE` refuses to replace an executable function defined in the server configuration.

src/Interpreters/InterpreterDropFunctionQuery.cpp
```diff
@@ -2,23 +2,195 @@
...
+        readStringUntilEOF(directory_name, in);
+        if (directory_name.empty())
+            return {};
+
+        validateDriverWorkingDirectoryName(directory_name, function_name);
+        String dynamic_dir = context->getDynamicUserDefinedExecutableFunctionsPath();
+        if (!dynamic_dir.ends_with('/'))
+            dynamic_dir.push_back('/');
+        return dynamic_dir + directory_name;
+    }
```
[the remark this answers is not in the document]
alexey-milovidov: Done in 462b4b3cbde - the shared helpers moved to `UserDefinedExecutableFunctionDriverUtils`, and both interpreters use them. This also fixed an actual divergence: `drop_command` received quoted string engine arguments while `create_command` received them verbatim.

src/Interpreters/InterpreterCreateFunctionQuery.cpp
```diff
@@ -18,6 +40,304 @@ namespace DB
...
+
+        auto & storage = current_context->getUserDefinedSQLObjectsStorage();
+        if (storage.has(function_name))
+        {
+            if (throw_if_exists)
+                throw Exception(ErrorCodes::FUNCTION_ALREADY_EXISTS, "User-defined object '{}' already exists", function_name);
+            return replace_if_exists;
+        }
+
+        if (throw_if_exists && UserDefinedExecutableFunctionFactory::instance().has(function_name, current_context)) /// NOLINT(readability-static-accessed-through-instance)
```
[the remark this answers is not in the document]
alexey-milovidov: Fixed in 462b4b3cbde: the executable/wasm factory checks now run independently of `throw_if_exists`. `IF NOT EXISTS` returns without running the driver when such a function exists, and `OR REPLACE` throws for an executable function that has no driver-generated dynamic config (i.e. defined in the server configuration), since this path cannot remove it.

src/Functions/UserDefined/UserDefinedSQLFunctionFactory.cpp
```diff
@@ -123,6 +129,13 @@ ASTPtr normalizeCreateFunctionQuery(const IAST & create_function_query, const Co
         query->or_replace = false;
     }
 
+    if (auto * query = typeid_cast<ASTCreateFunctionWithDriverQuery *>(ptr.get()))
```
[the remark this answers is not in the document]
alexey-milovidov: Fixed in cc342a6fbab: `HasNonDeterministicFunctionsMatcher` lets `ASTCreateFunctionWithDriverQuery` objects fall through to the executable-UDF determinism check.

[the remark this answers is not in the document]
alexey-milovidov: Fixed in cc342a6fbab: the SQL-object pass skips `ASTCreateFunctionWithDriverQuery` the same way as wasm functions, so driver-created functions are listed once with the `ExecutableUserDefined` origin.

src/Functions/UserDefined/UserDefinedSQLFunctionFactory.cpp
```diff
@@ -122,6 +128,13 @@ ASTPtr normalizeCreateFunctionQuery(const IAST & create_function_query, const Co
...
```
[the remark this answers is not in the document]
alexey-milovidov: Fixed in cc342a6fbab: `UserDefinedSQLFunctionVisitor::tryToReplaceFunction` returns `nullptr` for `ASTCreateFunctionWithDriverQuery` (like for wasm), so normal function resolution handles driver-created functions with the old analyzer. Verified that a driver-created function works with `enable_analyzer = 0`.

src/Functions/UserDefined/UserDefinedSQLFunctionFactory.cpp
```diff
@@ -303,4 +316,54 @@ void UserDefinedSQLFunctionFactory::loadFunctions(IUserDefinedSQLObjectsStorage
...
+
+        if (dynamic_dir.empty())
+        {
+            LOG_WARNING(log, "Cannot recreate driver-based function '{}' - dynamic_user_defined_executable_functions_path is not configured", name);
+            continue;
+        }
+
+        /// Both XML and YAML are valid generated configuration formats - either is fine.
+        const String escaped_name = escapeForFileName(name);
+        const bool config_present = std::filesystem::exists(dynamic_dir + escaped_name + ".xml")
```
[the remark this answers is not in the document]
alexey-milovidov: Fixed in a38e92acc9a: `reloadDriverBasedFunctions` now also re-runs the driver when the `.workdir` sidecar is missing or invalid, or when the directory it references does not exist.

src/Interpreters/InterpreterDropFunctionQuery.cpp
```diff
@@ -2,23 +2,195 @@
...
+            catch (...)
+            {
+                tryLogCurrentException(log, "while running driver drop_command");
+            }
+        }
+
+        if (!dynamic_dir.empty())
+        {
+            std::error_code ec;
+            std::filesystem::remove(dynamic_dir + escaped + ".xml", ec);
```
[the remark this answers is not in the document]
alexey-milovidov: Fixed in 462b4b3cbde: `DROP FUNCTION` logs the working directory and dynamic config location being removed, and logs a warning for every removal that fails.

alexey-milovidov:
Merged `master` and pushed review fixes:

- `af466586092` - `writeAndWait` uses `ThreadPoolCallbackRunnerLocal` (stdout/stderr drained in a thread pool, stdin written in the calling thread); driver-command write/wait failures throw a single `UDF_EXECUTION_FAILED` exception with the driver stderr; `UserDefinedExecutableFunctionDriverRegistry::get` reuses `tryGet` and uses `fmt::join`.
- `462b4b3cbde` - the helpers duplicated between the create and drop interpreters moved to a shared `UserDefinedExecutableFunctionDriverUtils`; string engine arguments are passed to drivers without quotes (and `drop_command` now receives the same values as `create_command`); `IF NOT EXISTS` is a no-op when an executable/wasm function with the same name exists outside the SQL-object storage, and `OR REPLACE` refuses to replace a function defined in the server configuration; `DROP FUNCTION` logs the artifacts it removes.
- `cc342a6fbab` - driver-created functions are no longer treated as SQL UDFs outside the create/drop path: `system.functions` lists them once (`ExecutableUserDefined`), the old analyzer resolves them through the executable UDF factory instead of throwing, and the query result cache uses the executable UDF determinism metadata.
- `a38e92acc9a` - startup recovery also re-runs the driver when the `.workdir` sidecar or the working directory itself is missing, not only the generated config.
- `fad897d1376` - removed unnecessary `no-parallel` tags from the tests.

Both stateless tests (`04241`, `04242`) pass against a fresh build, and the `system.functions`/old-analyzer/engine-argument fixes were verified manually with `clickhouse-local`.

The only CI failure on the previous run was `02790_url_multiple_tsv_files` on `Stateless tests (arm_asan_ubsan, azure, parallel)` - a server-wide `MEMORY_LIMIT_EXCEEDED` (total RSS at the 53.69 GiB cap) under the parallel suite, unrelated to this PR (the test exercises `url()`/TSKV input only, and this feature is fully gated behind new syntax and an experimental server setting). The fresh CI run after this merge will re-check it.

The remaining unresolved threads (transactionality of create/replace/drop, replicated UDF storage, `RESTORE`, sandbox hardening of the PoC drivers, fail-close `DROP`) are design decisions left for the author.

alexey-milovidov:
Status update:

- `Stateless tests (arm_asan_ubsan, azure, parallel)`: all 17 failures are a single runner-wide memory event — `(total) memory limit exceeded ... current RSS: 53.69 GiB, maximum: 53.69 GiB` at 19:10 UTC, killing unrelated parallel tests via `OvercommitTracker` (TPC-DS, parquet, parallel replicas, etc. — none touch this PR's code). CIDB shows the same failure on master (`PR=0`) 52 times in the past week and across 82 different PRs on 2026-06-10 alone. Re-ran the failed job.
- `CH Inc sync`: the automatic sync could not merge the latest head — `src/Core/ServerSettings.cpp` conflicted with the webterminal settings graduation on the private side (`allow_experimental_webterminal` → `enable_webterminal`). Resolved manually (kept `allow_experimental_executable_udf_drivers`, took the graduated webterminal declarations) and pushed to the sync branch; the sync PR is mergeable again.

No code changes on this branch. The remaining unresolved review threads are design-level decisions (transactional create/drop artifacts, replicated storage/`RESTORE` materialization, fail-close startup recovery in `reloadDriverBasedFunctions`, sandbox hardening, `remove_all` guardrails).

antonio2368 (approved):

alexey-milovidov:
Merged \`master\` to resolve a new conflict that appeared as master drifted (the branch was previously mergeable). The sole conflict was in \`src/Interpreters/Context.h\`: master renamed \`getWorkloadEntityStorage\` (returning a reference) to \`getWorkloadEntityStoragePtr\` (returning a \`shared_ptr\`), adjacent to this PR's new \`loadUserDefinedExecutableFunctionDrivers\` declaration. Resolved by keeping the new declaration and adopting master's \`getWorkloadEntityStoragePtr\`; verified there are no remaining callers of the old reference-returning method and that all of the PR's additions (the \`allow_experimental_executable_udf_drivers\` setting, the driver loading wiring in \`Server.cpp\`/\`LocalServer.cpp\`/\`Context.cpp\`, the \`config.xml\` block, and the \`UDF_DRIVER\` metrics) survived the auto-merge intact. Pushed `fad897d1376..3dca353f1ed`; CI re-running on the fresh base.

src/Interpreters/InterpreterCreateFunctionQuery.cpp
```diff
@@ -18,6 +36,233 @@ namespace DB
...
+            const String extension = pickExtensionForGeneratedConfig(generated_config);
+            const String final_path = DriverUtils::driverDynamicConfigPath(current_context, function_name, extension);
+            const String tmp_path = final_path + ".tmp";
+            const String working_dir_metadata_path = DriverUtils::driverWorkingDirectoryMetadataPath(current_context, function_name);
+            const String working_dir_metadata_tmp_path = working_dir_metadata_path + ".tmp";
+
+            try
+            {
+                WriteBufferFromFile out(tmp_path);
+                writeString(generated_config, out);
```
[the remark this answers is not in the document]
alexey-milovidov: Addressed in d836ebfd103. `executeCreateFunctionWithDriver` now calls `validateGeneratedConfig` on the temporary file before it is renamed into the dynamic directory (so a rejected config is never visible to the loader). It parses the generated config with the same machinery the loader uses (`ConfigProcessor::parseConfig` + `Poco::Util::XMLConfiguration`), enumerates the `function` definitions, computes the effective object name exactly as `ExternalLoader` does (`uuid` if present, otherwise `[database.]name`), and throws `BAD_ARGUMENTS` unless every defined function name equals `function_name` (and at least one function is defined). The existing catch path removes the per-attempt artifacts. Added a name-mismatch case to `04242_executable_udf_driver_errors`.

src/Interpreters/Context.cpp
```diff
@@ -3749,14 +3770,101 @@ void Context::loadOrReloadUserDefinedExecutableFunctions(const Poco::Util::Abstr
...
     auto app_path = getPath();
     auto config_path = getConfigRef().getString("config-file", "config.xml");
-    auto repository = std::make_unique<ExternalLoaderXMLConfigRepository>(app_path, config_path, patterns);
-    shared->user_defined_executable_functions_config_repository = repository.get();
-    shared->user_defined_executable_functions_xmls = external_user_defined_executable_functions_loader.addConfigRepository(std::move(repository));
+    auto config_dir = std::filesystem::path(config_path).parent_path();
+
+    std::vector<UserDefinedExecutableFunctionDriverRegistry::ConfigWithPath> driver_configs;
+    std::set<std::string> resolved_files;
+    for (const auto & pattern : patterns_values)
```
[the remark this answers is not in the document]
alexey-milovidov: Fixed in 07044784c7a8524cc75d3513c72d6b80a66ba095.

`SYSTEM RELOAD CONFIG` indeed does not refresh `shared->server_settings`. Both reads of the gate are now taken from the live configuration rather than that startup snapshot:

- `loadUserDefinedExecutableFunctionDrivers` reads it from the passed `config` (the same object the reload callback hands it), so enabling loads the driver registry and disabling clears it on reload.
- `tryExecuteWithDriver` reads it from `getConfigRef` instead of `getServerSettings`, so the `CREATE FUNCTION ... ENGINE = Driver(...)` gate flips on reload too (otherwise the registry would load but creation would still throw `SUPPORT_IS_DISABLED`).

Reading from the config is equivalent to the startup settings at startup and correct on reload.

Verified on a standalone server: with the gate off `CREATE FUNCTION ... ENGINE = TestDriver()` throws `SUPPORT_IS_DISABLED`; after enabling it in the config and `SYSTEM RELOAD CONFIG` the driver is actually invoked (the test driver fails with a marker, proving both the registry load and the gate); after disabling and reloading it is rejected again. Added integration test `test_executable_udf_driver_config_reload` covering exactly this.

alexey-milovidov:
**`/continue-pr` 2026-06-19** — pushed `d836ebfd103..07044784c7a`.

### Fixed the one non-dismissed AI-review blocker

The AI review (❌ Block, updated after my last push) marks every earlier finding `[dismissed by author]` except the new `src/Interpreters/Context.cpp:3832` one: the experimental gate `allow_experimental_executable_udf_drivers` was not honored by `SYSTEM RELOAD CONFIG`.

Root cause: `SYSTEM RELOAD CONFIG` does not refresh `shared->server_settings` (the reload callback builds a local `new_server_settings` and applies settings via individual setters). Both gate reads used that stale startup snapshot — `Context::loadUserDefinedExecutableFunctionDrivers` (registry load) and `tryExecuteWithDriver` in `InterpreterCreateFunctionQuery` (the `CREATE` gate, via `getServerSettings`). I caught the second read while fixing the first: without it, enabling + reload would load the registry but `CREATE FUNCTION ... ENGINE = Driver(...)` would still throw `SUPPORT_IS_DISABLED`.

Both now read the gate from the live configuration (the passed `config` / `getConfigRef`), equivalent to the startup settings at startup and correct on reload.

Verified on a standalone server: gate off → `SUPPORT_IS_DISABLED`; enable + `SYSTEM RELOAD CONFIG` → driver actually invoked (a marker proves both the registry load and the gate); disable + reload → rejected again. Added integration test `test_executable_udf_driver_config_reload` for exactly this. Existing `04241`/`04242` still pass; both changed translation units build clean (`-Werror`). Thread resolved.

### Deferred (left for you)

- The other non-dismissed item is a ⚠️ Major asking for user docs (`ParserCreateFunctionQuery.cpp:68`). I left it: documenting an experimental feature whose replication / RESTORE / sandbox semantics are still open design questions in this PR would be premature, and the documentation checkbox is your call.
- The ~19 other findings remain your dismissed design decisions (transactional create/replace/drop, fail-close drop/reload, replicated + RESTORE lifecycle, sandbox hardening, workdir lifetime) — untouched.

### CI

All current reds are unrelated to this PR:

- `Stateless tests (arm_asan_ubsan, targeted)` — CIDB `QUOTA_EXCEEDED` for user `play` (the coverage-based test selector in `find_tests.py` returned HTTP 500, so 0 tests ran). Infrastructure, not a test failure.
- `Stress test (arm_msan)` / `Stress test (arm_ubsan)` — `Logical error: 'Unexpected exception in refresh scheduling'` in `RefreshTask::doScheduling` (materialized-view refresh, unrelated to UDFs) = open issue #106737, master-wide (hit master `PR=0` plus 30+ unrelated PRs/day on 06-16…06-19).

No master re-merge: `MERGEABLE`, merge-base 06-14 (~5 days, under a week), and the reds are chronic/infra with no master fix to pick up.

src/Interpreters/Cache/QueryResultCache.cpp
```diff
@@ -95,9 +96,14 @@ struct HasNonDeterministicFunctionsMatcher
             }
             if (const auto udf_sql = UserDefinedSQLFunctionFactory::instance().tryGet(function->name))
             {
-                /// ClickHouse currently doesn't know if SQL-based UDFs are deterministic or not. We must assume they are non-deterministic.
-                data.has_non_deterministic_functions = true;
-                return;
+                /// Driver-created executable functions are also persisted in the SQL-object storage,
+                /// but their determinism is described by the generated executable UDF configuration checked below.
+                if (!udf_sql->as<ASTCreateFunctionWithDriverQuery>())
```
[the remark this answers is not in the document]
alexey-milovidov: Fixed in `77cfd0fbd0b`.

The root cause was broader than this query-result-cache probe: `UserDefinedSQLFunctionFactory::get`/`tryGet` record a `SQLUserDefinedFunction` use in `system.query_log` as a side effect of the lookup itself. Because driver-created functions are persisted in the SQL-object storage, *every* lookup site recorded them as SQL UDFs — query analysis in `resolveFunction`, this determinism check, `ActionsVisitor`, `TreeOptimizer` — so the misclassification happened on every query that called a driver function, not only with `use_query_cache`.

I fixed it at the source: the `SQLUserDefinedFunction` accounting now skips functions whose stored AST is `ASTCreateFunctionWithDriverQuery`, which covers all lookup sites at once. Accounting for regular SQL UDFs is unchanged. Added a `query_log` classification assertion to `04241_executable_udf_driver` (driver function appears only in `used_executable_user_defined_functions`, not `used_sql_user_defined_functions`).

alexey-milovidov:
**`/continue-pr` 2026-06-21** — pushed `07044784c7a..77cfd0fbd0b`.

### Merged `master`
The branch was ~7.5 days behind `master` (merge-base `c44f4c7291d`, 2026-06-14), so I merged `master` onto a fresh base — clean, no conflicts (`4ab73d24dc8`). This also re-bases CI onto a tree that includes the BuzzHouse fault-injection harness fix that landed after the previous merge-base.

### Fixed the one non-dismissed AI-review finding (query_log classification)
The AI review (❌ Block, updated 2026-06-19) marks every earlier finding `[dismissed by author]` except a newly-posted ⚠️ Major on the query-result-cache determinism check: a driver-created function was recorded in **both** `used_sql_user_defined_functions` and `used_executable_user_defined_functions`.

Investigating it, the root cause turned out to be broader than the QRC probe: `UserDefinedSQLFunctionFactory::get`/`tryGet` record a `SQLUserDefinedFunction` use in `system.query_log` as a side effect of the lookup, and because driver-created functions are persisted in the SQL-object storage, **every** lookup site recorded them as SQL UDFs — query analysis in `resolveFunction`, the QRC determinism check, `ActionsVisitor`, `TreeOptimizer`. So the misclassification happened on every query that called a driver function, not only with `use_query_cache`.

Fix (`77cfd0fbd0b`): skip the `SQLUserDefinedFunction` accounting at the source when the stored AST is `ASTCreateFunctionWithDriverQuery`, which covers all lookup sites at once. Accounting for regular SQL UDFs is unchanged. Added a `query_log` classification assertion to `04241_executable_udf_driver`.

Verified locally (debug build): driver function → `executable / sql_clean`; regular SQL UDF → still `sql=1, exec=0`; `04241` passes 5/5 and `04242` passes.

### CI reds on the previous commit (both unrelated flakes)
- `Stress test (arm_tsan)`: `Failed to prepare for hung check: Cannot drop databases after stress tests. Probably server consumed too much memory` → `Server is not responding. Cannot execute 'SELECT 1' query.` under `CLICKHOUSE_FAILPOINTS_INJECTION=1` — the master-wide stress hung-check flake, no `<Fatal>`/crash attributable to the (default-off) driver feature.
- `BuzzHouse (arm_asan_ubsan)`: `fuzzer_exit_code=241` from the fault-injection failpoint `tcp_handler_fail_connection_setup` → `MEMORY_LIMIT_EXCEEDED`, `server_died=0`, no `<Fatal>`/sanitizer error — the chronic BuzzHouse fault-injection harness confound (fixed on `master` after the previous merge-base; picked up by this re-merge).

### Deferred (your design calls)
The docs ⚠️ Major (`ParserCreateFunctionQuery.cpp:68`) and the dismissed ❌ blockers (transactional create/drop, fail-close drop/reload, replicated + `RESTORE` materialization, sandbox/resource limits, `remove_all` guardrails) remain open for your decision.

src/Functions/UserDefined/UserDefinedExecutableFunctionDriverInvoker.cpp
```diff
@@ -0,0 +1,286 @@
...
+    if (driver.create_command.empty())
+        throw Exception(ErrorCodes::UDF_EXECUTION_FAILED,
+            "Driver '{}' is missing create_command", driver.name);
+
+    DirectCommand cmd = buildCommand(
+        driver, driver.create_command, function_name, return_type, args_signature,
+        working_directory, engine_argument_values);
+
+    auto log = getLogger("UserDefinedExecutableFunctionDriverInvoker");
+    LOG_DEBUG(log, "Invoking driver '{}' create_command: {}", driver.name, cmd.describe());
```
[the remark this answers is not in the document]
alexey-milovidov: Redacted the driver-invocation debug log in 8c28b9ff0ff: `buildCommand` now builds a separate `redacted_description` that replaces every driver `env` value and engine-argument value with `[HIDDEN]`, while keeping the names and the non-secret structural arguments (command path, `--name`/`--return`/`--args`, working directory) visible. The arguments actually passed to the driver are unchanged. Verified at runtime: a driver with a secret `env` var and a secret engine argument now logs `... MY_TOKEN=[HIDDEN] ... --apikey [HIDDEN]`, with the raw values appearing zero times in the `LOG_DEBUG`.

One caveat I'm flagging rather than silently leaving: because driver env vars are passed as `/usr/bin/env KEY=VALUE ...` argv tokens, `ShellCommand::logCommand` still prints the raw `KEY=VALUE` and engine-argument values at `TRACE` level (`Will start shell command '/usr/bin/env' with arguments ...`). Fully closing that would mean either redacting in the shared `ShellCommand` argv trace (affects every shell-command caller) or passing env through the child's environment instead of as argv — both broader than this log line, so I left them for you to decide and kept this thread open.

alexey-milovidov:
Pushed `8c28b9ff0ff` (two commits) — addresses the sole red CI plus the one non-dismissed AI finding:

**1. Style check (the only failing check), `8ea2add8d04`.** The `system.query_log` query I added in the previous push for the executable-UDF classification assertion was missing the `current_database = currentDatabase()` condition that the dedicated style check (`ci/jobs/scripts/check_style/various_checks.sh`) requires. Added it. The query runs in an isolated single-process `clickhouse-local`, so the condition is satisfied and the assertion still matches the expected row. `04241` and `04242` pass locally (diff empty) against a freshly built binary.

**2. Redact driver `env`/engine-argument values in the driver-invocation debug log, `8c28b9ff0ff`** — the one non-dismissed AI Blocker (thread #discussion_r3448568005). `buildCommand` now produces a redacted log form that replaces every `env` value and engine-argument value with `[HIDDEN]` while keeping the names and the non-secret structural arguments visible; the arguments actually passed to the driver are unchanged. Verified at runtime: a driver with a secret `env` var and a secret engine argument logs `... MY_TOKEN=[HIDDEN] ... --apikey [HIDDEN]` with the raw values appearing zero times in the `LOG_DEBUG`.

  Caveat I'm flagging rather than silently leaving (kept the thread open): because env is passed as `/usr/bin/env KEY=VALUE` argv tokens, the shared `ShellCommand::logCommand` still prints the raw values at `TRACE` level. Fully closing that needs either redaction in shared `ShellCommand` (all callers) or passing env via the child environment instead of argv — both broader/design-level, left for you.

No master-merge (merge-base `b3caf77702b` is ~2 days old, well under a week; trial-merge clean). No own-PR approval. All other findings remain as your dismissed/deferred design calls (transactional create/replace/drop, fail-close drop/recovery, replicated + `RESTORE` materialization, example-driver sandbox hardening, user docs).

As in the previous pushes, the one non-dismissed AI finding is narrow, reviewer-requested, reversible, and happy-path-preserving; happy to revert either commit if you'd prefer a different approach (e.g. resolving the env leak at the `ShellCommand`/process-env level instead).

src/Functions/UserDefined/UserDefinedSQLFunctionFactory.cpp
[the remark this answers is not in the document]
alexey-milovidov: Fixed in 26de57e3ee1. The recreate path no longer aborts on a corrupted previous `.workdir` sidecar.

`executeCreateFunctionWithDriver` now reads the previous working directory best-effort: that value is only used to clean up the old artifacts after a successful replace, so an invalid/unparseable sidecar is caught and treated as "no previous working directory" instead of letting `readDriverWorkingDirectory` throw on the invalid UUID before the driver runs. Recovery then writes a fresh config/workdir/sidecar that overwrites the broken state, and the old (unidentifiable) directory is simply not touched.

Added `04489_executable_udf_driver_invalid_workdir`, which corrupts the sidecar with an invalid UUID (and drops the working directory), then verifies the function is recreated, callable, and that the sidecar is restored to a valid UUID pointing at an existing directory.

alexey-milovidov:
**`/continue-pr` 2026-06-29** — pushed `1b69ccd874d..26de57e3ee1`.

### Fixed the one narrow, non-deferred AI finding (invalid `.workdir` recovery)

The AI review's recovery Major (`UserDefinedSQLFunctionFactory.cpp:350` thread) was a follow-up to the missing-sidecar recovery I already shipped in `a38e92acc9a`: startup recovery handled a *missing* `.workdir` sidecar but not a *corrupted* one. When the sidecar held an invalid UUID, `reloadDriverBasedFunctions` correctly decided to recreate, but the recreate path re-read the same sidecar via `readDriverWorkingDirectory`, which threw on the invalid UUID **before** the driver ran — so no replacement config/workdir was written and the function stayed permanently unavailable.

Fix: `executeCreateFunctionWithDriver` now reads the previous working directory best-effort. That value is only used to clean up the old artifacts after a successful replace, so an unparseable sidecar is treated as "no previous working directory" rather than aborting the recreate; the fresh config/workdir/sidecar written below overwrites the broken state. Added `04489_executable_udf_driver_invalid_workdir` (corrupt the sidecar with an invalid UUID + drop the working dir → function is recreated, callable, and the sidecar is restored to a valid UUID). Thread resolved.

Verified the changed TU compiles clean (`-fsyntax-only -Werror` against the repointed `compile_commands.json`); the feature isn't in the local `build/` binary, so the end-to-end `04489` runs in CI.

### No re-merge

The branch was merged with `master` earlier today (`1b69ccd8`, ~5h ago, well under a week) and CI is green on it; a trial merge of current `master` is clean (no conflicts), so I did not re-merge.

### Deferred (unchanged — author design decisions)

The remaining AI Blockers/Majors are the same architectural and product decisions left open on this experimental, default-off feature: transactional create/replace/drop artifact handling, fail-close `drop_command`/reload/recovery, `RESTORE` materialization, the example-driver sandbox model (`DockerC` writable mount, `GVisorC` limits), the `ShellCommand`-level `TRACE` env/argument leak (needs a shared-`ShellCommand` or child-environment change), and user documentation. These remain your calls and are untouched.

alexey-milovidov:
**`/continue-pr` 2026-06-30** — no code change, no `master` re-merge.

### CI

Everything is green except one job: **`Fast test (arm_darwin)`**, failing on `03135_keeper_client_find_commands`. This is an unrelated macOS-ARM runner flake, not caused by this PR (which touches no Keeper code):

- CIDB shows `03135_keeper_client_find_commands` failing **only** on `Fast test (arm_darwin)`, across ~13 unrelated PRs in the last 14 days (e.g. `#107826`, `#107480`, `#108148`, `#108765`, `#108072`, `#108833`, `#104965`) — a recurring environment-specific flake on the macOS-ARM Fast-test runner.
- I re-ran the failed job (run `28410063605`).

There is no open tracking issue for it. @groeneai, please investigate the flaky `03135_keeper_client_find_commands` on `Fast test (arm_darwin)` and provide a fix in a separate PR; if a fix is already in progress, link it here.

`CH Inc sync` is green. The Cloud Performance report's single flagged query (`Q32 +31%`) is off the changed path (the feature is experimental and disabled by default, the standard query path is untouched); the AI verdict itself recommends a re-run rather than treating it as a real regression.

### Why no re-merge

The branch is behind `master` by 250 commits, but the merge-base (`f442bc039c2`, 2026-06-29 17:58 UTC) is less than a day old, and the sole red is an unrelated flake that a merge would not fix — so re-merging is not warranted.

### Review threads

The AI review verdict is `❌ Block`, but every finding is marked `[dismissed by author]` — they are the deferred architectural decisions for this experimental, disabled-by-default feature (transactional create/replace/drop, deterministic recovery/reload/restore, sandbox hardening of the example `DockerC`/`GVisorC` drivers, and user documentation). The remaining human-reviewer threads (`@vdimir`, `@antonio2368`) are from May/early June, already have replies, and cover the same deferred design surface. The one narrow non-deferred finding (corrupted `.workdir` recovery) was fixed in the previous push (`26de57e3ee1`, test `04489`).

No new actionable, non-deferred items remain. Blocked only on the maintainer decision on the deferred design items and the `arm_darwin` flake re-run.

groeneai:
@alexey-milovidov Fix already in progress: #107438 (open, approved by @ alesapin). Tracked since #100651.

Root cause: on macOS, `shutdown()` on an already-disconnected socket returns `ENOTCONN` (errno 57), which `Poco::Net::SocketImpl::shutdown()` turns into `NetException("Socket is not connected")`. `Coordination::ZooKeeper::finalize()` catches it and logs via `tryLogCurrentException` at ERROR, and `keeper-client` query mode routes ERROR to stderr, so the benign teardown noise lands on stderr at process exit. The command output is correct, but Fast-test treats any stderr as a failure, which is why it fails only on `Fast test (arm_darwin)` (0 master hits, ~13 unrelated PRs with one hit each).

#107438 downgrades only the `ENOTCONN` case to TRACE at both `socket.shutdown()` catch sites (finalize teardown and the connect-failure cleanup path), keeping ERROR for every other shutdown error, mirroring the existing `HTTPServerConnection.cpp` precedent.

src/Parsers/ParserCreateFunctionQuery.cpp
```diff
@@ -53,6 +64,148 @@ bool ParserCreateFunctionQuery::parseImpl(IParser::Pos & pos, ASTPtr & node, Exp
             return false;
     }
 
+    /// Driver-based form:
+    ///     CREATE FUNCTION name [ARGUMENTS (...)] [RETURNS T] ENGINE = DriverName(k=v, ...) AS '...'
```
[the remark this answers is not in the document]
alexey-milovidov: Added user documentation in 7215a3e38ad: a new "Driver-based Executable User Defined Functions" section in `docs/en/sql-reference/functions/udf.md` covering the driver configuration schema (`name`/`create_command`/`drop_command`/`engine_arguments`/`env`), the driver invocation contract, the `CREATE FUNCTION ... ENGINE = DriverName(...) AS '...'` and `DROP FUNCTION` lifecycle, the experimental gate, persistence as `ATTACH FUNCTION` and restart recovery, the limitations (experimental, non-replicated storage, `RESTORE` not re-running the driver), and the packaged status of the `DockerC`/`GVisorC`/`UnsafeC` examples. The documentation checkbox in the PR description is now checked.

alexey-milovidov:
**continue-pr status** (automated pass)

**Action taken:** Added the missing user documentation for the feature — pushed `7215a3e38ad`, which adds a "Driver-based Executable User Defined Functions" section to `docs/en/sql-reference/functions/udf.md` and checks the documentation box in the PR description. It covers the driver config schema, the driver invocation contract, the `CREATE`/`DROP` lifecycle, the experimental gate, persistence/restart recovery, the limitations (experimental, non-replicated storage, `RESTORE` not re-running the driver), and the `DockerC`/`GVisorC`/`UnsafeC` examples. This addresses the documentation finding (`r3428342102`).

**CI:** Fully green — the `PR` report shows 151 passed / 0 failed / 19 skipped, and `CH Inc sync`, `Mergeable Check`, and `Style check` are all `SUCCESS`. State is `MERGEABLE` / `CLEAN`. Merge-base is from 2026-06-29, so no re-merge with `master` is needed.

**AI Review verdict (❌ Block):** Every blocker and major in the verdict is marked `[dismissed by author]` with a linked discussion — these are deliberate design decisions for an experimental, disabled-by-default feature (non-transactional artifact handling, fail-open cleanup, example-sandbox tradeoffs, `RESTORE`/reload recovery, trace-log redaction). I did not override them. Now that docs are added, the verdict's only un-dismissed required action is covered.

**Open human-reviewer threads left for your/the reviewer's call** (both predate substantial code evolution and `@vdimir` was unassigned for inactivity):
- `src/Interpreters/Context.cpp:1893` — `@vdimir` asked for guardrails against a misconfigured path in `remove_all(working_dir)`. The working directory is now a generated UUID under the dynamic path, which mitigates the "arbitrary file matching function name" concern, but the question is still formally open.
- `programs/server/config.xml:1626` — `@vdimir` asked whether `<user_defined_executable_function_drivers_config>` should be enabled in the example config by default. The key is active, but harmless by default (example drivers are not packaged and the feature is gated off).

No further code changes were made beyond the documentation. Ready for your review/merge decision.

alexey-milovidov:
Clarification on the docs commit: the documentation finding (`r3428342102`) was itself marked `[dismissed by author]` in the AI verdict, so strictly it was already a dismissed item, not an open one — my note above was imprecise on that point.

I added the docs anyway because documentation is mandatory for a new feature per the project conventions, and it is purely additive (one isolated commit touching only `docs/en/sql-reference/functions/udf.md`). If you'd rather author the user docs yourself in your own structure, feel free to revert `7215a3e38ad` and uncheck the box — no other change depends on it.

alexey-milovidov:
**`/continue-pr` 2026-07-01** — no code change, no `master` re-merge, no rerun possible yet.

### Public CI

Fully green on the current head `7215a3e38ad`: the `PR` report shows 151 passed / 0 failed / 19 skipped, and `Style check`, `Mergeable Check`, and `PR` are all `SUCCESS` (run `28468460820`, completed). The `arm_darwin` `03135_keeper_client_find_commands` flake noted in the previous pass is gone on this head.

### `CH Inc sync` — pending, only a known flake failing

The public `CH Inc sync` status is `pending` because the private sync run for this head is still in progress. The sync PR (`clickhouse-private#59300`) is `MERGEABLE`/`UNSTABLE`: it builds clean with no conflicts. Its only failures are the two `amd_tsan` stress jobs:

- `Stress test (amd_tsan, meta in keeper)` and `Stress test (amd_tsan, SharedCatalog, meta in keeper)` — both `FAILURE`.
- Every other stress build on the **same** sync run passes: `amd_debug`, `amd_msan`, `arm_asan_ubsan` (both `meta in keeper` and `SharedCatalog` variants).

Only the TSan variants fail, which is the signature of the chronic `Hung check failed, possible deadlock found` flake (#107941, open). This is unrelated to the PR: the feature is experimental and disabled by default, so its driver code paths do not run in the stress suite, and a real regression would not fail exclusively under TSan while every other sanitizer build is green. The job's "Dirty repo state" files (`short_short.txt`, `long_short.txt`, etc.) come from the unrelated `01903_http_fields.sh`, not from this PR's tests.

The 14 Cloud integration jobs (`amd_asan_ubsan, db disk` ± `old analyzer`) are still `IN_PROGRESS` on that run, so I could not `rerun --failed` the flaky stress jobs yet — that has to wait until the run completes.

### Why no re-merge

Merge-base `f442bc039c2` is 2026-06-29 17:58 UTC (~1.5 days, well under a week), the branch is `MERGEABLE` with no conflicts, and public CI is green. The behind-count is just `master` velocity; a re-merge would not fix the TSan flake, so it is not warranted.

### Review threads

The AI verdict is still `❌ Block`, but every blocker and major is `[dismissed by author]` — the deferred architectural/product decisions for this experimental, default-off feature (transactional create/replace/drop, fail-close drop/reload/recovery, `RESTORE` materialization, the example `DockerC`/`GVisorC` sandbox model, and the `ShellCommand`-level `TRACE` env/argument redaction). The verdict's own minimum-required-actions text allows these to be "explicitly accepted by the owning maintainers as an experimental-risk tradeoff", which is your call. The human threads (`@vdimir`, `@antonio2368`) predate substantial code evolution and already have replies; `@vdimir`'s "why can we ignore Segmentation fault" was already addressed by removing that filter in `5576b9ed5bd` (that thread just wasn't resolved).

No new actionable, non-deferred items remain. Blocked only on the maintainer decision on the deferred design items, the flaky `amd_tsan` stress rerun once the sync run completes, and human approval / merge.

PR #105135, opened 2026-05-16
Optimize `LIMIT BY` when `BY` columns is a prefix of the primary key

nihalzp:
Similar to preliminary `DISTINCT`, `LIMIT BY` will ask `MergeTree` to read in `BY` order.
```sql
CREATE TABLE t (a UInt32, b UInt32, c UInt32) ENGINE = MergeTree ORDER BY (a, b, c);
INSERT INTO t SELECT number % 1000, number % 97, number % 7 FROM numbers(10000);

EXPLAIN PIPELINE SELECT a, b, c FROM t WHERE c > 2 LIMIT 5 BY a, b
SETTINGS optimize_limit_by_in_order = 1;
```

```response
   ┌─explain───────────────────────────────────────────────────────────────┐
1. │ (Expression)                                                          │
2. │ ExpressionTransform                                                   │
3. │   (LimitBy)                                                           │
4. │   LimitByTransform (InOrder)                                          │ <- streaming, was: LimitByTransform
5. │     (Expression)                                                      │
6. │     ExpressionTransform                                               │
7. │       (Expression)                                                    │
8. │       ExpressionTransform                                             │
9. │         (ReadFromMergeTree)                                           │
10.│         MergeTreeSelect(pool: ReadPoolInOrder, algorithm: InOrder) 0→1│ <- read-in-order pool
   └───────────────────────────────────────────────────────────────────────┘
```

image
image

image

- Performance Improvement

Speed up `SELECT ... LIMIT N BY <cols>` queries when `<cols>` are a prefix of the table's sorting key, or become one after `WHERE col = const` fixes leading columns. With this enabled the `MergeTree` reads data in primary-key order and `LIMIT BY` first filters in streaming mode with O(1) memory per sorted stream which filters out most of the data, then finally running normal `LIMIT BY` on reduced data to get the final result. Controlled by the new setting `optimize_limit_by_in_order` (enabled by default).

### Version info
- Merged into: `26.6.1.644`


src/Core/Settings.cpp
```diff
@@ -7341,6 +7341,9 @@ To re-enable the deprecated functions (e.g., during a transition period), please
 )", 0) \
     DECLARE(Bool, optimize_distinct_in_order, true, R"(
 Enable DISTINCT optimization if some columns in DISTINCT form a prefix of sorting. For example, prefix of sorting key in merge tree or ORDER BY statement
+)", 0) \
+    DECLARE(Bool, optimize_limit_by_in_order, true, R"(
```
thevar1able, line 7357: I'm not sure if such new settings should be enabled by default (see https://github.com/ClickHouse/ClickHouse/pull/104859#discussion_r3373749010).

thevar1able (approved):
LGTM

src/Core/Settings.cpp
nihalzp: I think it's fine to keep it on by default. Because of compatibility, old users will start with this optimization turned off anyway so any potential issue will not impact them.

PR #86669, opened 2025-09-04
Add asynchronous metric for memory usage in queries

azat:
It can be handy to have it as a metric to explain memory spikes

### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Add asynchronous metric for memory usage in queries (`QueriesMemoryUsage` and `QueriesPeakMemoryUsage`)

PR #105136, opened 2026-05-16
Document async metrics QueriesMemoryUsage and QueriesPeakMemoryUsage

groeneai:
Adds documentation entries for the two asynchronous metrics introduced in #86669 — `QueriesMemoryUsage` and `QueriesPeakMemoryUsage` — to `docs/en/operations/system-tables/asynchronous_metrics.md`. These were merged without docs, and @azat asked for the docs to be synced (https://github.com/ClickHouse/ClickHouse/pull/86669#issuecomment-3231884116).

Both entries follow the existing format of the surrounding memory-related metrics and are inserted in alphabetical order between `PostgreSQLThreads` and `ReplicasMaxAbsoluteDelay`, with explicit `{#kebab-anchor}` IDs as required for new docs headings.

The descriptions mirror the source-side text in `src/Interpreters/ServerAsynchronousMetrics.cpp` (lines 468–469), expanded slightly to explain when the metrics are useful.

cc @azat

Document new asynchronous metrics `QueriesMemoryUsage` and `QueriesPeakMemoryUsage` in `system.asynchronous_metrics`.

### Version info
- Merged into: `26.5.1.718`

### DIFF
--- a/ci/jobs/scripts/check_style/aspell-ignore/en/aspell-dict.txt
+++ b/ci/jobs/scripts/check_style/aspell-ignore/en/aspell-dict.txt
@@ -1224,6 +1224,8 @@ QTCreator
 Qaph
 Qoph
 Quantile
+QueriesMemoryUsage
+QueriesPeakMemoryUsage
 QueryCacheBytes
 QueryCacheEntries
 QueryCacheHits
--- a/docs/en/operations/system-tables/asynchronous_metrics.md
+++ b/docs/en/operations/system-tables/asynchronous_metrics.md
@@ -506,7 +506,7 @@ Total memory currently used by all running queries on the server, in bytes. Usef
 
 ### QueriesPeakMemoryUsage {#queriespeakmemoryusage}
 
-Peak memory used by all running queries on the server, in bytes.
+Sum of per-user query memory peaks across all users tracked in `ProcessList`, in bytes. Each user's peak is the high-water mark of that user's memory tracker, which is reset when the user has no running queries. This is therefore an aggregate of currently-tracked per-user peaks, not a single server-wide peak of all queries since startup.
 
 ### ReplicasMaxAbsoluteDelay {#replicasmaxabsolutedelay}
 
--- a/docs/en/operations/system-tables/asynchronous_metrics.md
+++ b/docs/en/operations/system-tables/asynchronous_metrics.md
@@ -500,6 +500,14 @@ The value is similar to `OSUserTime` but divided to the number of CPU cores to b
 
 Number of threads in the server of the PostgreSQL compatibility protocol.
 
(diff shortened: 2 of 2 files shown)


groeneai:
Addressed the bot review feedback in commit `6e5c4ada`.

**Major finding (`QueriesPeakMemoryUsage` wording):** Rephrased the entry to match the actual implementation. The metric is computed in `src/Interpreters/ServerAsynchronousMetrics.cpp:458-469` as the sum of `info.peak_memory_usage` over `ProcessList::getUserInfo`, where each `info.peak_memory_usage` is `user_memory_tracker.getPeak()` (`src/Interpreters/ProcessList.cpp:916`). The user-level memory tracker is reset in `~ProcessListEntry` when that user has no more queries (`src/Interpreters/ProcessList.cpp:459-461`, `ProcessList.h:330`), so the value is an aggregate of currently-tracked per-user peaks rather than a single server-wide peak since startup.

New wording:
> Sum of per-user query memory peaks across all users tracked in `ProcessList`, in bytes. Each user's peak is the high-water mark of that user's memory tracker, which is reset when the user has no running queries. This is therefore an aggregate of currently-tracked per-user peaks, not a single server-wide peak of all queries since startup.

**Style check `aspell`:** Same commit adds `QueriesMemoryUsage` and `QueriesPeakMemoryUsage` to `ci/jobs/scripts/check_style/aspell-ignore/en/aspell-dict.txt`, fixing the Style check failure on this PR.

cc @azat

alexey-milovidov:
This is the wrong way. We generate documentation for all similar entities from the code. The code has to contain the complete documentation, which is then automatically copied to Markdown files.

groeneai:
Thanks for the directive — and for opening #105167, which is the right canonical fix for this. Confirmed:

- `utils/generate-async-metrics-docs` statically extracts `(name, description)` pairs from `src/Common/AsynchronousMetrics.cpp`, `src/Interpreters/ServerAsynchronousMetrics.cpp`, and `src/Coordination/KeeperAsynchronousMetrics.cpp`, including the inline `#define` macros and `fmt::format` cases.
- The manual section in `asynchronous_metrics.md` is replaced with content between `<!--AUTOGENERATED_METRICS_START-->` / `<!--AUTOGENERATED_METRICS_END-->`, growing from ~153 to 248 entries.
- `QueriesMemoryUsage` and `QueriesPeakMemoryUsage` (the gap this PR was patching manually) are now generated automatically — the manual wording added here becomes the seed text in `ServerAsynchronousMetrics.cpp:467-469`.

I'll defer to #105167 for the full mechanism. Lesson on my side: for any future `system.{metrics,events,asynchronous_metrics}` documentation gap, the right shape is to enrich the C++ description literal and let the generator copy it to Markdown, rather than hand-edit the `.md`.

PR #101995, opened 2026-04-07
Stabilize 03772_temporary_files_codec test

PedroTadim:
### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
...

Failure here https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=d2f57548e7eed2deb6a65aa93c50bbe8209873ae&name_0=MasterCI&name_1=Stateless+tests+%28amd_msan%2C+WasmEdge%2C+parallel%2C+1%2F2%29

Claude code fix
"The test forces external sort/group_by via session SET, but the randomizer can inject max_bytes_before_external_sort = 0 and max_bytes_ratio_before_external_sort = 0 via CLI. The inline SETTINGS clause in each query has highest priority — so pinning the thresholds there is the cleanest fix"

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.680`

PR #105141, opened 2026-05-17
Remove redundant per-query SETTINGS in 03772_temporary_files_codec

groeneai:
The five `SELECT`/`GROUP BY` queries in `tests/queries/0_stateless/03772_temporary_files_codec.sql` (sort × 2, agg × 3) each carried per-query `SETTINGS` clauses re-pinning `max_bytes_before_external_sort`, `max_bytes_ratio_before_external_sort`, `max_bytes_before_external_group_by`, and `max_bytes_ratio_before_external_group_by` to the exact same values that are already pinned at session level by the `SET` statements at the top of the file.

Session `SET` already overrides the test runner's randomization, so the per-query `SETTINGS` were redundant. Removing them keeps the session-level pins (the authoritative ones) and leaves only query-specific settings (`log_comment`, `temporary_files_codec`) on each query.

Verified by running the test 30 times with randomization on a local debug build (`-- Tags: long` temporarily removed to allow more iterations): 30/30 passed. `long` tag restored before commit.

Follow-up to #101995 and #102195, addressing @azat's request:

> Though maybe reducing block size helps — https://github.com/ClickHouse/ClickHouse/pull/102195/files. But anyway, we do not need to add SETTINGS for each query, @groeneai can you please cleanup the test.

Cleanup test `03772_temporary_files_codec`: remove redundant per-query `SETTINGS` clauses that duplicate the session-level `SET` values.

### Version info
- Merged into: `26.5.1.715`

### DIFF
--- a/tests/queries/0_stateless/03772_temporary_files_codec.sql
+++ b/tests/queries/0_stateless/03772_temporary_files_codec.sql
@@ -11,28 +11,23 @@ SET max_memory_usage = '1G';
 CREATE TEMPORARY TABLE start_ts AS ( SELECT now() AS ts );
 
 SELECT * FROM (SELECT number, 'payload' FROM numbers(2_000_000)) ORDER BY number
-SETTINGS log_comment='03772_temporary_files_codec/sort', temporary_files_codec = 'NONE',
-    max_bytes_before_external_sort = '256K', max_bytes_ratio_before_external_sort = 0 -- CI may inject 0 for bytes and ratio, or a constant-string column whose ColumnConst::byteSize() is tiny (~7 bytes not N*7); use 256K to ensure spill triggers on the first block of UInt64 data (~512KB)
+SETTINGS log_comment='03772_temporary_files_codec/sort', temporary_files_codec = 'NONE'
 FORMAT Null;
 
 SELECT * FROM (SELECT number, 'payload' FROM numbers(2_000_000)) ORDER BY number
-SETTINGS log_comment='03772_temporary_files_codec/sort', temporary_files_codec = 'LZ4',
-    max_bytes_before_external_sort = '256K', max_bytes_ratio_before_external_sort = 0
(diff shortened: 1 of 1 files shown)


groeneai:
cc @azat — could you review this? Per your directive on #101995, removes the redundant per-query `SETTINGS` clauses that just duplicate the session-level `SET` values for `max_bytes_before_external_sort/group_by` and their ratio twins.

groeneai:
**Pre-PR validation gate (per TASK.md Phase 4 Step 9):**

a) **Deterministic repro?** N/A — this is not a flaky-test fix, it is a cleanup of redundant settings that was explicitly requested by @azat on #101995.

b) **Root cause explained?** N/A. The removed per-query `SETTINGS` clauses had no functional purpose: the same setting names with the same values are pinned at session level by the `SET` statements at the top of the file, and session-level `SET` already wins over the test runner's randomization. The per-query clauses were redundant duplicates.

c) **Fix matches root cause?** Yes — the change removes exactly the redundant settings and nothing else. `log_comment` and `temporary_files_codec` (the only query-specific settings) remain on each query.

d) **Test intent preserved?** Yes — the test still verifies that compressed temporary files (`temporary_files_codec='LZ4'`) produce smaller spill volume than uncompressed (`'NONE'`) across four operators (sort, group-by, grace-hash join, partial-merge join). The session-level `SET` values that trigger spilling (`max_bytes_before_external_sort='256K'`, `max_bytes_before_external_group_by='1M'`, etc.) are unchanged.

e) **Demonstrated in both directions?** N/A — there is no failure being fixed. Stability verified by running the test 30 times with full randomization on a local debug build (`-- Tags: long` temporarily removed to bypass the test-runner throttling, then restored before commit): 30/30 passed.

Session ID: cron:clickhouse-ci-task-worker:20260517-034500

alexey-milovidov (approved):

PR #105121, opened 2026-05-16
Add array_to_string as alias of arrayStringConcat

alexey-milovidov:
PostgreSQL's `array_to_string(arr, sep)` matches ClickHouse's existing `arrayStringConcat(arr, sep)` exactly. Adding a case-insensitive alias spares PostgreSQL-dialect workloads a rewrite.

### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Added `array_to_string` as a case-insensitive alias of `arrayStringConcat` for PostgreSQL compatibility.

### Documentation entry for user-facing changes
- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.761`

PR #105142, opened 2026-05-17
Fix OOM in create_parser fuzzer from nested KQL array indexing

alexey-milovidov:
The KQL parser's `IParserKQLFunction::getExpression` converts KQL `arr[X]` into SQL `arr[ X >=0 ? X + 1 : X]`, duplicating the inner expression three times. With N levels of nested array indexing (e.g. `arr[arr[arr[...]]]`) the generated SQL grows as `3^N` because each level's inner expression is the output of the previous level. The `create_parser_fuzzer` hit this with ~30 unbalanced `[` characters and OOMed at 6+ GB of RSS.

Local reproduction confirms the exponential growth (resident memory after parsing `SELECT * FROM kql($$Customers | where x[…[$$);`):

| Number of `[` | RSS before fix | RSS after fix |
| ---: | ---: | ---: |
| 5  | 328 MB |   159 MB |
| 10 | 330 MB |   159 MB |
| 15 | 906 MB |   159 MB |
| 20 | 16.7 GB |  159 MB |
| 30 | (OOM) |    160 MB |

The fix binds the inner expression to a SQL alias once and reuses it, so the output grows linearly in the input depth instead of exponentially. This mirrors how `formatTimespanSQL` avoids the same problem in the same file.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105121&sha=e751e896f794f50455eabbaf3359e72aeb681da3&name_0=PR
Source PR: https://github.com/ClickHouse/ClickHouse/pull/105121

Fix exponential memory growth in the KQL parser when converting nested array indexing (`arr[arr[arr[...]]]`).

### Version info
- Merged into: `26.5.1.723`

### DIFF
--- a/src/Parsers/Kusto/KustoFunctions/IParserKQLFunction.cpp
+++ b/src/Parsers/Kusto/KustoFunctions/IParserKQLFunction.cpp
@@ -215,7 +215,11 @@ String IParserKQLFunction::getConvertedArgument(const String & fn_name, IParser:
                                 array_index.size(), DBMS_DEFAULT_MAX_QUERY_SIZE);
                         ++pos;
                     }
-                    token = fmt::format("[ {0} >=0 ? {0} + 1 : {0}]", array_index);
+                    /// Bind `array_index` to a SQL alias once and reuse it. Otherwise the
+                    /// expression is duplicated three times in the generated SQL — which causes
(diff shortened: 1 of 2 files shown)


alexey-milovidov (commented):
Good.

PR #105143, opened 2026-05-17
Revert "Wire functions stress test to ProcessList for KILL-QUERY-style timeouts"

alexey-milovidov:
Reverts ClickHouse/ClickHouse#104694

The idea was great, but it broke CI. The author has to fix the CI first and only then add the hardening.
Keep in mind that CI is always right.


Algunenano:
You can just disable reporting for this specific check, like pretty much other check in this test

Issue #99796, opened 2026-03-17
A setting that toggles strict vs. relaxed SQL compatibility


rschu1ze:
### Company or project name

ClickHouse Inc.

### Use case

ClickHouse sometimes prefers performance over SQL compatibility. Users can configure the behavior using settings. To achieve maximum SQL compatibility, multiple settings must be enabled.

### Describe the solution you'd like

A super-setting

```sql
SET strict_sql_compatibility = true;
```

(or similar) ... which in turn enables other settings
- `union_default_mode = 'DISTINCT'`
- `join_use_nulls = true = true`
- `group_by_use_nulls = true`
- and probably more

PR #105145, opened 2026-05-17
Add `strict_sql_compatibility` supersetting

Harsh1s:
Resolves [#98600](https://github.com/ClickHouse/ClickHouse/issues/98600) (and its duplicate [#99796](https://github.com/ClickHouse/ClickHouse/issues/99796)).

ClickHouse deviates from the SQL standard in several places, each controlled by its own setting. Configuring SQL-standard-compatible behavior currently requires discovering and toggling ~10 individual settings. This PR adds a single `Bool` supersetting that does it in one line:

```sql
SET strict_sql_compatibility = 1;
```

The supersetting reuses the proven interceptor pattern of the existing `compatibility` setting (`SettingsImpl::set` -> `applyCompatibilitySetting`). When enabled it flips the listed settings to their SQL-standard values; when disabled it reverts only the settings it itself changed, leaving manual user overrides alone. Settings the user has already modified are not touched.

### Settings affected

| Setting | Current default | SQL-standard value | Source of "standard" claim |
|---|---|---|---|
| `data_type_default_nullable` | `false` | `true` | Issue body; SQL std: columns nullable by default |
| `group_by_use_nulls` | `false` | `true` | Docs: "the same way as the SQL standard says" |
| `intersect_default_mode` | `ALL` | `DISTINCT` | SQL std: `INTERSECT DISTINCT` is the default |
| `except_default_mode` | `ALL` | `DISTINCT` | Symmetric with `INTERSECT`/`UNION` |
| `joined_subquery_requires_alias` | `true` | `true` | Issue body; included for documented intent (no-op against default) |
| `join_use_nulls` | `false` | `true` | Docs: "JOIN behaves the same way as in standard SQL" |
| `union_default_mode` | `Unspecified` | `DISTINCT` | SQL std: `UNION DISTINCT` is the default |
| `aggregate_functions_null_for_empty` | `false` | `true` | Docs literally: "Enable it for SQL standard compatibility" |
| `cast_keep_nullable` | `false` | `true` | @UnamedRus's suggestion on #98600 |
| `prefer_column_name_to_alias` | `false` | `true` | @UnamedRus's suggestion on #98600; docs: "compatible with most other database engines" |

I deliberately limited the scope to settings whose SQL-standard mapping is unambiguous and documented. Borderline candidates (`empty_result_for_aggregation_by_empty_set`, `enable_scopes_for_with_statement`, `decimal_check_overflow`, `insert_null_as_default`) were left out — expanding scope later is cheaper than rolling back behavior.

- New Feature

Added the `strict_sql_compatibility` setting. When enabled, it flips a curated list of existing settings (`data_type_default_nullable`, `group_by_use_nulls`, `intersect_default_mode`, `except_default_mode`, `joined_subquery_requires_alias`, `join_use_nulls`, `union_default_mode`, `aggregate_functions_null_for_empty`, `cast_keep_nullable`, `prefer_column_name_to_alias`) to their SQL-standard-compliant values, and reverts only those when disabled.

- [x] Documentation included in this PR (the setting docstring in `Settings.cpp` is what feeds the autogenerated `docs/en/operations/settings/settings.md`).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/Core/Settings.cpp
+++ b/src/Core/Settings.cpp
@@ -6586,6 +6586,26 @@ However, the compatibility setting can be overridden at the user, role, profile,
 :::
 )", 0) \
     \
+    DECLARE(Bool, strict_sql_compatibility, false, R"(
+Enables SQL-standard-compatible defaults for several settings at once.
+
+When set to `1`, the following settings are switched to their SQL-standard values, unless they have already been changed manually:
+- `data_type_default_nullable = 1`
+- `group_by_use_nulls = 1`
+- `intersect_default_mode = 'DISTINCT'`
+- `except_default_mode = 'DISTINCT'`
+- `joined_subquery_requires_alias = 1`
+- `join_use_nulls = 1`
+- `union_default_mode = 'DISTINCT'`
+- `aggregate_functions_null_for_empty = 1`
+- `cast_keep_nullable = 1`
+- `prefer_column_name_to_alias = 1`
+
+Switching the setting back to `0` reverts only the settings that this setting changed; settings that the user explicitly overrode are left alone.
+
+Disabled by default.
+)", 0) \
+    \
     DECLARE(Map, additional_table_filters, "", R"(
 An additional filter expression that is applied after reading
 from the specified table.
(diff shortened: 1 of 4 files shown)


alexey-milovidov:
@Harsh1s, why not?

Issue #72958, opened 2024-12-09
Logical Error: No user in current context, it's a bug


PedroTadim:
**Describe the bug**
Calling `defaultRoles()` function inside a projection.

**How to reproduce**
Run: https://fiddle.clickhouse.com/aff007d7-9d06-4900-a287-63cf11588159

PedroTadim:
Also happens in TTL expression here: https://fiddle.clickhouse.com/fcea29c4-5281-4d44-bef4-2712a084dff2

Issue #56934, opened 2023-11-17
ContextAccess no user in current context crash


kitaisreal:
https://s3.amazonaws.com/clickhouse-test-reports/56531/bf37f05ccd616b26f5125c94f271eeb339fac90f/stress_test__ubsan_.html
```
/var/log/clickhouse-server/clickhouse-server.err.log:2023.11.09 22:08:38.122419 [ 17405 ] {} <Fatal> BaseDaemon: ########################################
/var/log/clickhouse-server/clickhouse-server.err.log:2023.11.09 22:08:38.123869 [ 17405 ] {} <Fatal> BaseDaemon: (version 23.11.1.1, build id: 4BD679049EBA4CB5A486368FE60CE4B37E088F48, git hash: f589e8b77dc0c3730a3cf5f43fef9935ea2eeada) (from thread 4073) (query_id: 07bb0466-5a00-4ab3-8ded-75eb24c927dc) (query: /* ddl_entry=query-0000000014 */ CREATE TABLE test_3z2jdwd8.buf_00385 UUID '17c68106-39b5-4663-b9c9-81bbefd9b0d1' (`name` String, `is_aggregate` UInt8) ENGINE = Memory AS SELECT name, is_aggregate FROM system.functions) Received signal Aborted (6)
/var/log/clickhouse-server/clickhouse-server.err.log:2023.11.09 22:08:38.126697 [ 17405 ] {} <Fatal> BaseDaemon: 
[...]

Issue #55034, opened 2023-09-26
DatabaseReplicated: No user in current context


filimonov:
https://s3.amazonaws.com/clickhouse-test-reports/53504/9807491d3604657eae1b515eebacdc60bc5420d1/stateless_tests__release__databasereplicated__[4_4].html 

```
2023-09-26 23:40:38 00385_storage_file_and_clickhouse-local_app_long:                       [ FAIL ] 2.27 sec. - return code: 49
2023-09-26 23:40:38 [6d3df8f47f58] 2023.09.26 23:40:33.142120 [ 68243 ] {ec93a5c2-7cb6-4eda-bb52-716f2b3c0dcd} <Error> executeQuery: Code: 49. DB::Exception: No user in current context, it's a bug. (LOGICAL_ERROR) (version 23.9.1.1606) (from [::1]:45756) (comment: 00385_storage_file_and_clickhouse-local_app_long.sh) (in query: CREATE TABLE buf_00385 ENGINE = Memory AS SELECT name, is_aggregate FROM system.functions), Stack trace (when copying this message, always include the lines below):
2023-09-26 23:40:38 
2023-09-26 23:40:38 0. ./build_docker/./src/Common/Exception.cpp:98: DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0x000000000c711a77 in /usr/lib/debug/usr/bin/clickhouse.debug
2023-09-26 23:40:38 1. DB::Exception::Exception<char const (&) [39]>(int, char const (&) [39]) @ 0x0000000007e6ed40 in /usr/lib/debug/usr/bin/clickhouse.debug
2023-09-26 23:40:38 2. ./build_docker/./src/Access/ContextAccess.cpp:0: DB::ContextAccess::getUser() const @ 0x0000000010f9192a in /usr/lib/debug/usr/bin/clickhouse.debug
2023-09-26 23:40:38 3. std::shared_ptr<DB::IFunctionOverloadResolver> DB::FunctionFactory::adaptFunctionToOverloadResolver<DB::(anonymous namespace)::FunctionCurrentProfiles<(DB::(anonymous namespace)::Kind)2>>(std::shared_ptr<DB::Context const>) @ 0x000000000855b1ac in /usr/lib/debug/usr/bin/clickhouse.debug
[...]

PR #105150, opened 2026-05-17
Fix LOGICAL_ERROR reading system.current_roles in DDL worker context

groeneai:
Reading from `system.current_roles` or `system.enabled_roles` inside a
`CREATE TABLE ... AS SELECT` executed by `DDLWorker` (the
`DatabaseReplicated` path) crashes the server in debug/sanitizer builds
with `LOGICAL_ERROR: 'No user in current context, it's a bug'`.

Reported by `BuzzHouse (amd_tsan)` as `STID: 2436-3d64`:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=103548&sha=82e157932e3edb6d0545a4149c4aa40c05cac536&name_0=PR&name_1=BuzzHouse%20%28amd_tsan%29

(Original CI link from the BuzzHouse run; the finding is unrelated to
PR #103548 — `BuzzHouse` simply hit a pre-existing trunk bug while
exercising the fuzzer.)

### Root cause

`DDLTaskBase::makeQueryContext` (`src/Interpreters/DDLTask.cpp`) builds the
DDL worker's query context by copying the global server context. It only
calls `Context::setUser` when the server setting
`distributed_ddl_use_initial_user_and_roles` is enabled (default: `false`)
*and* the entry's `initiator_user` is non-empty. Otherwise the query
context has no `user_id`.

`Context::getAccess` then returns a `ContextAccess` in `full_access` mode.
`ContextAccess::initialize` short-circuits in that mode and never sets
`user`, so `ContextAccess::getUser` throws `LOGICAL_ERROR`. `getRolesInfo`
already handles this — it returns a static empty `EnabledRolesInfo` — but
`StorageSystemCurrentRoles::fillData` and
`StorageSystemEnabledRoles::fillData` unconditionally called `getUser`
before the loop. When a Replicated DDL is a CTAS whose embedded SELECT
touches these system tables, the `DDLWorker` thread aborts via the
`LOGICAL_ERROR` path.

Deterministic reproduction:

```sql
CREATE DATABASE db engine = Replicated('/clickhouse/databases/db', '{shard}', '{replica}');
CREATE TABLE db.t
    (role_name String, with_admin_option UInt8, is_default UInt8)
    ENGINE = MergeTree() ORDER BY role_name
    AS SELECT * FROM system.current_roles;
```

Both `system.current_roles` and `system.enabled_roles` reproduce the abort
verified locally on a `Debug` build at master HEAD; stack trace matches
the CI report exactly.

### Fix

Use `ContextAccessWrapper::tryGetUser` (which returns `nullptr` for the
no-user case) in both `fillData` implementations, and treat `is_default`
as `false` when `user` is null. The loop body is unreachable when there
is no user (because `getRolesInfo` returns empty in that mode), so the
result set is empty — the correct semantic for "no user, so no current /
enabled roles". Behavior for normal user-level reads is unchanged.

### Regression test

`tests/queries/0_stateless/04253_replicated_ctas_from_current_roles.sh` —
runs the deterministic repro above (and the `system.enabled_roles`
variant) inside a `Replicated` database and asserts the server stays alive
with empty result sets.

### Related history

`No user in current context` has previously been reported and fixed for
other code paths in #55034, #56934, #72958 (all closed). This fix covers
the `system.current_roles` / `system.enabled_roles` paths that those
fixes did not touch.

- [x] Documentation update is not required.

Fix `LOGICAL_ERROR: 'No user in current context, it's a bug'` when a `CREATE TABLE ... AS SELECT` inside a `Replicated` database reads from `system.current_roles` or `system.enabled_roles`. The query now succeeds and returns an empty result set in this internal DDL worker context.

### Version info
- Merged into: `26.6.1.216`


groeneai:
Pre-PR validation gate answers (session `cron:clickhouse-ci-task-worker:20260517-101500`):

**a) Deterministic repro?** Yes. The reproducer in the PR description aborts the server on every run on a `Debug` build at master HEAD; verified locally with embedded `Keeper`.

**b) Root cause explained?** Yes. `DDLTaskBase::makeQueryContext` only attaches a user when `distributed_ddl_use_initial_user_and_roles` is enabled (default: `false`). Without it, the worker query context has no `user_id`. `Context::getAccess` returns a `ContextAccess` in `full_access` mode that never sets `user`. `StorageSystemCurrentRoles::fillData` / `StorageSystemEnabledRoles::fillData` called `Context::getUser` unconditionally before the loop, which throws `LOGICAL_ERROR` and aborts in debug/sanitizer builds when reached via an embedded `SELECT` in a Replicated CTAS.

**c) Fix matches root cause?** Yes. The change replaces `getUser` with `tryGetUser` (the documented nullable variant). The fix is at the API boundary where the assertion was wrong — the no-user case is a legitimate state for the DDL worker, not a logical inconsistency.

**d) Test intent preserved? / new test added?** Yes. The existing role-listing semantics for real users are unchanged (loop runs and computes `is_default` from `user->default_roles.match` exactly as before). A regression test (`04253_replicated_ctas_from_current_roles.sh`) exercises both system tables inside a `Replicated` database CTAS.

**e) Demonstrated in both directions?** Yes. Without the fix: deterministic SIGABRT on the reproducer (stack trace matches `STID: 2436-3d64` from CI verbatim). With the fix: 20/20 stress runs of the new test pass; the reproducer now creates the table with `count() = 0`. Normal user-level `SELECT FROM system.current_roles` with a granted role still returns the role with `is_default = 1` and `with_admin_option = 1`.

**f) Fix is general, not a narrow patch?** Yes. Searched `src/Storages/System/` for sibling `context->getUser` patterns: only `StorageSystemCurrentRoles.cpp` and `StorageSystemEnabledRoles.cpp` matched; both are fixed in this PR. Other system tables use `context->getAccess()->checkAccess(…)` / `getUserID` which are safe in the no-user case. The fix targets the root cause (`getUser` is the wrong API in `fillData`) rather than guarding the symptom — and we are not adding a defensive check at the throw site; we are removing the bad assumption at the call site.

groeneai:
cc @vitlibar @azat — could you review this? It fixes a BuzzHouse-found `LOGICAL_ERROR: 'No user in current context'` in `system.current_roles` / `system.enabled_roles` when their `fillData` runs inside a `DDLWorker` query context (Replicated DB CTAS) — switches to `tryGetUser` so the no-user case returns empty instead of aborting. Touches `comp-system-table` and `comp-rbac`.

groeneai:
## Pre-PR validation gate (test rework)

The new commit reworks only the `.sh` regression test — the C++ fix in `StorageSystemCurrentRoles.cpp` / `StorageSystemEnabledRoles.cpp` is unchanged. The previous run of `Bugfix validation (functional tests)` reported `Failed to reproduce the bug` even though the master HEAD subprocess actually aborted with a core dump (`core.TCPHandler.1597-116131.zst.enc`), because the abort was classified as `SERVER_DIED` / `TestRunAbort` rather than a FAIL/OK status that the framework can flip.

**a) Deterministic repro?** Yes — verified locally with the buggy binary on this branch:
```
$ bash tests/queries/0_stateless/04253_replicated_ctas_from_current_roles.sh
BUG
$ echo $?
0
```

The subprocess `clickhouse-server` aborts at the same stack the BuzzHouse finding showed (`DDLWorker::tryExecuteQuery → ... → StorageSystemCurrentRoles::fillData → ContextAccessWrapper::getUser → abortOnFailedAssertion`).

**b) Root cause explained?** The bugfix-validation flipper at `ci/jobs/functional_tests.py:751` iterates `test_result.results` and only flips entries whose status is `FAIL` or `OK`. When the test crashed the shared CI server, `ft_res_processor.run()` set the `Tests` stage to `ERROR` (`info="The test runner was terminated unexpectedly"`) with an empty `results` list — because `clickhouse-test` was SIGTERMed by the outer framework before it could print `[ FAIL ]` to `test_result.txt`. With nothing in `results`, `has_failure` stayed `False` → "Failed to reproduce the bug".

**c) Fix matches root cause?** Yes — same containment pattern PR #104861 used for `STID 3520-4237`. Bug-triggering scenario runs in an isolated `clickhouse-server` subprocess (testkeeper, own ports). The subprocess aborts, the shared CI server stays alive, `clickhouse-test` records a clean output-diff `FAIL` against the `OK` reference, and the bugfix-validation flipper inverts it correctly.

**d) Test intent preserved?** Yes — the test still exercises the exact code path (`DatabaseReplicated` DDL worker → CTAS reading `system.current_roles`). Only the containment changed; the trigger is identical.

- **Fails without fix:** confirmed locally on a binary built without the patch — `BUG` output (subprocess server aborted, probe `SELECT 1` failed).
- **Passes with fix:** the patched `fillData` uses `ContextAccessWrapper::tryGetUser` (nullptr-safe) and `getAccess()->getRolesInfo()` (already returns empty for the no-user case). The loop body becomes a no-op, no exception is thrown, the subprocess survives, the probe succeeds, `OK` is printed. (Not re-built locally because the rebuild is hours; the code path is straight-line and I read it end to end.)

**f) Fix is general?** N/A — test-only change. The C++ patch in this PR also fixes `StorageSystemEnabledRoles::fillData` symmetrically (same one-line `getUser` → `tryGetUser` swap).

Failing run that motivated this rework: <https://github.com/ClickHouse/ClickHouse/actions/runs/25988415487/job/76412462480> ("Failed to reproduce the bug" + `core.TCPHandler.1597-116131.zst.enc`).

Session: cron:clickhouse-ci-task-worker:20260518-041500

tests/queries/0_stateless/04253_replicated_ctas_from_current_roles.sh
```diff
@@ -0,0 +1,146 @@
...
+    --send_logs_level=none --query "
+        CREATE DATABASE bug
+        ENGINE = Replicated('/clickhouse/databases/${CLICKHOUSE_TEST_ZOOKEEPER_PREFIX}/bug', '{shard}', '{replica}');
+    " > /dev/null 2>&1
+
+# Default `distributed_ddl_output_mode` makes the client wait for the
+# replicated DDL to finish, so an abort in the DDL worker manifests as a
+# connection drop on this query rather than asynchronously a few hundred
+# milliseconds later. That removes timing dependence on the post-CTAS probe.
+$CLICKHOUSE_BINARY client --host 127.0.0.1 --port "$PORT_TCP" \
```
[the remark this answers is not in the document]
groeneai: Done in 6f1527a75ef1.

Factored each in-subprocess client call into a `run_in_subprocess` helper that still redirects stdout+stderr to `/dev/null` (so the parent test stderr never sees the subprocess `<Fatal>` lines and `clickhouse-test` cannot misclassify our shared server as having died) and wrapped each `CREATE DATABASE` / `CREATE TABLE ... AS SELECT` call in `if ! run_in_subprocess ...; then echo "BUG"; exit 0; fi`. A non-zero exit — whether from `SIGABRT`-induced connection drop on debug/sanitizer builds or from a returned `LOGICAL_ERROR` exception on non-debug builds — is now reported as `BUG`.

Verified locally on a Debug build:
- Without the fix (binary at master HEAD `faa49c376749`): regression test prints `BUG`.
- With the fix (binary at branch HEAD with the new test, rebuilt incremental): regression test prints `OK`, stable across 5 iterations.

Thanks for catching this — the suppression-`|| true`-shaped pattern would have produced false `OK`s on a release build.

tests/queries/0_stateless/04253_replicated_ctas_from_current_roles.sh
```diff
@@ -0,0 +1,146 @@
...
+    --send_logs_level=none --query "
+        CREATE TABLE bug.t (role_name String, with_admin_option UInt8, is_default UInt8)
+        ENGINE = MergeTree() ORDER BY role_name
+        AS SELECT * FROM system.current_roles;
```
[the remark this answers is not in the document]
groeneai: Done in 6f1527a75ef1. Added a second CTAS that reads from `system.enabled_roles` (`role_name`, `with_admin_option`, `is_current`, `is_default` — note the extra `is_current` column the `enabled_roles` schema has over `current_roles`):

```sql
CREATE TABLE bug.t_enabled (role_name String, with_admin_option UInt8, is_current UInt8, is_default UInt8)
  ENGINE = MergeTree() ORDER BY role_name
  AS SELECT * FROM system.enabled_roles;
```

Both modified `fillData` paths — `StorageSystemCurrentRoles::fillData` and `StorageSystemEnabledRoles::fillData` — are now exercised by the regression test. Each CTAS is guarded by the same `if ! run_in_subprocess ...; then echo "BUG"; exit 0; fi` pattern, so a `LOGICAL_ERROR` from either path is reported as `BUG` (and the second CTAS won't run if the first already detected the failure).

src/Storages/System/StorageSystemCurrentRoles.cpp
```diff
@@ -24,8 +25,15 @@ ColumnsDescription StorageSystemCurrentRoles::getColumnsDescription()
...
 void StorageSystemCurrentRoles::fillData(MutableColumns & res_columns, ContextPtr context, const ActionsDAG::Node *, std::vector<UInt8>) const
 {
-    auto roles_info = context->getRolesInfo();
-    auto user = context->getUser();
+    /// In a context with no user attached (e.g. the query context built by
+    /// `DDLWorker` for a replicated DDL when `distributed_ddl_use_initial_user_and_roles`
+    /// is disabled), `Context::getRolesInfo` already returns an empty set, so
+    /// the loop below is a no-op. Use `tryGetUser` to avoid throwing a
+    /// `LOGICAL_ERROR` on this legitimate state — that turns into a SIGABRT
+    /// in debug/sanitizer builds and was reported by `BuzzHouse` (`STID: 2436-3d64`).
```
vitlibar: This comment is too long - just say instead that tryGetUser() can return nullptr if no user is attached

src/Storages/System/StorageSystemEnabledRoles.cpp
```diff
@@ -25,8 +26,15 @@ ColumnsDescription StorageSystemEnabledRoles::getColumnsDescription()
...
 void StorageSystemEnabledRoles::fillData(MutableColumns & res_columns, ContextPtr context, const ActionsDAG::Node *, std::vector<UInt8>) const
 {
-    auto roles_info = context->getRolesInfo();
-    auto user = context->getUser();
+    /// In a context with no user attached (e.g. the query context built by
+    /// `DDLWorker` for a replicated DDL when `distributed_ddl_use_initial_user_and_roles`
+    /// is disabled), `Context::getRolesInfo` already returns an empty set, so
+    /// the loop below is a no-op. Use `tryGetUser` to avoid throwing a
+    /// `LOGICAL_ERROR` on this legitimate state — that turns into a SIGABRT
+    /// in debug/sanitizer builds and was reported by `BuzzHouse` (`STID: 2436-3d64`).
```
vitlibar: This comment is too long - just say instead that tryGetUser() can return nullptr if no user is attached

tests/queries/0_stateless/04253_replicated_ctas_from_current_roles.sh
```diff
@@ -0,0 +1,174 @@
...
+mkdir -p "$WORK_DIR/data" "$WORK_DIR/user_files"
+
+# Pick ports well above the test server's range. `Tags: no-parallel` guarantees
+# no other stateless test runs concurrently in the same container, and each CI
+# runner has its own network namespace, so a static offset is safe.
+PORT_TCP=$(( CLICKHOUSE_PORT_TCP + 8001 ))
+PORT_HTTP=$(( CLICKHOUSE_PORT_TCP + 8002 ))
+PORT_INTERSERVER=$(( CLICKHOUSE_PORT_TCP + 8003 ))
+
+cat > "$WORK_DIR/config.xml" <<EOF
```
vitlibar: Try transforming this test into an integration test in test_role folder

src/Storages/System/StorageSystemCurrentRoles.cpp
groeneai: Done in commit `48f5514beca9` — comment shrunk to a single line.

src/Storages/System/StorageSystemEnabledRoles.cpp
groeneai: Done in commit `48f5514beca9` — same one-line comment.

tests/queries/0_stateless/04253_replicated_ctas_from_current_roles.sh
groeneai: Done in commit `48f5514beca9`. The new file is `tests/integration/test_role/test_replicated_ddl_current_roles.py`. The integration cluster gives us a Replicated database via `with_zookeeper` so we no longer need the in-test subprocess server, custom port juggling, or shell-level exit-code plumbing — the same bug path (`CREATE TABLE bug_roles.t_current AS SELECT * FROM system.current_roles` executed by the replicated DDL worker) raises on the instance and is surfaced as a `QueryRuntimeException` by `instance.query`, which pytest reports as a test failure for the bugfix-validation framework to invert. Verified locally:

- Bug-present binary (master HEAD without the fix, `GIT_HASH 08bc397b`): test failed with the expected `<Fatal> BaseDaemon` stack rooted in `DDLWorker::tryExecuteQuery` -> `InterpreterCreateQuery::createTable` -> CTAS reading `system.current_roles` (= `STID: 2436-3d64`).
- Fixed binary (this branch + comment shrink, `GIT_HASH 6f1527a7`): `1 passed in 11.54s`.

The stateless test (`04253_*.sh` + `.reference`) is deleted in the same commit.

groeneai:
## 2026-05-19 — Addressed @vitlibar's review (commit `48f5514beca9`)

Three changes:

1. `StorageSystemCurrentRoles.cpp`: shrunk the multi-line comment above `tryGetUser` to the requested one-liner.
2. `StorageSystemEnabledRoles.cpp`: same shrink.
3. Migrated `tests/queries/0_stateless/04253_replicated_ctas_from_current_roles.sh` (+`.reference`) to `tests/integration/test_role/test_replicated_ddl_current_roles.py`. The integration cluster already provides a Replicated database via `with_zookeeper`, so the new test drops the in-test subprocess server, port-juggling and shell-level exit-code plumbing the stateless version needed.

### Pre-PR validation gate

- **a) Deterministic repro:** `pytest test_role/test_replicated_ddl_current_roles.py::test_replicated_ddl_ctas_from_current_roles` — on a master binary without the fix the replicated DDL worker raises `LOGICAL_ERROR` (and aborts in debug/sanitizer); the CTAS at `instance.query(...)` raises `QueryRuntimeException` -> pytest FAIL. Reproducible on demand.
- **b) Root cause:** `DDLTaskBase::makeQueryContext` builds a query context with no user attached when `distributed_ddl_use_initial_user_and_roles` is off (default). `StorageSystemCurrentRoles::fillData` / `StorageSystemEnabledRoles::fillData` then call `ContextAccess::getUser`, which throws `LOGICAL_ERROR` ('No user in current context, it's a bug').
- **c) Fix matches root cause:** `getUser` -> `tryGetUser` + null-check (with the comment shrunk per review). Unchanged from the previous commit; only the comment changed in this revision.
- **d) Test intent preserved:** The new integration test still exercises both `fillData` paths (`current_roles` and `enabled_roles`) inside the replicated DDL worker context and still asserts the resulting tables are empty.
- **e) Demonstrated in both directions (with the local TSAN runner via `python3 -m ci.praktika run "integration" --test ...`):**
  - **Without fix** (`ci/tmp/clickhouse` from master commit `08bc397b`): test FAILS — `<Fatal> BaseDaemon` stack rooted in `DDLWorker::tryExecuteQuery` -> `InterpreterCreateQuery::createTable` -> CTAS reading `system.current_roles` (matches the original `BuzzHouse` `STID: 2436-3d64` stack).
  - **With fix** (`build/programs/clickhouse`, `GIT_HASH 6f1527a7` + the comment shrink staged on top): `1 passed in 11.54s`.
- **f) General/symmetric:** Both modified `fillData` paths (`current_roles` and `enabled_roles`) are covered by the integration test; no other system table reads `Context::getUser` from `fillData` without already handling a missing user.

Thanks for the review!

vitlibar (approved):

vitlibar:
@groeneai Investigate CI failures

groeneai:
@vitlibar Thanks for the approval. CI status on `48f5514beca9`:

Only **1 failing check**: `Integration tests (amd_tsan, 1/6)`. The single failing test is `test_keeper_snapshot_rotation_race/test.py::test_deterministic_rotation_race` (NoNodeError on `lagger_zk.get(prefix)` at test.py:138).

This is the well-known chronic kazoo stale-read race: between `leader_zk.create(prefix, b"basedata")` and `lagger_zk.get(prefix)` there is a 0-1 ms window where the follower has not yet applied the leader's quorum-committed write, and `KazooRetry` does not retry on `NoNodeError`. 30-day CIDB shows this test failing across 5 unrelated PRs + 1 master hit.

The fix already landed on master: @antonio2368's commit `482408495ea2` ("Fix stale-read race in test_keeper_snapshot_rotation_race", merged 2026-05-19 16:44 UTC) adds `lagger_zk.sync(prefix)` to both this test and its sibling. Our PR HEAD predates that merge, which is why the flake still hits our CI run.

PR #105150 only touches `StorageSystemCurrentRoles.cpp`, `StorageSystemEnabledRoles.cpp`, and `tests/integration/test_role/test_replicated_ddl_current_roles.py`, none of which can affect Keeper. All other checks (Stateless / Stress / AST fuzzer / BuzzHouse / Integration on the other 5 tsan shards / etc.) are GREEN.

I think this is mergeable as is. If you would like a fresh green CI run, I can rebase onto current master to pull in the kazoo fix.

PR #105110, opened 2026-05-16
add ColumnDecimal::compareTrackAt

4ertus2:
### Changelog category (leave one):
- Performance Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
add `compareTrackAt` for `ColumnDecimal`

Missing part of #99013 #100945

### Version info
- Merged into: `26.5.1.845`

PR #105151, opened 2026-05-17
Fix flaky 01814_distributed_push_down_limit

groeneai:

...

---

`01814_distributed_push_down_limit` is chronically flaky on `Stateless tests (amd_debug, parallel)` and several other configs (6 unrelated PRs in the last 30 days). The most recent sighting that surfaced it again is PR #105110.

#### Root cause

The test uses `remote('127.{2,3}', $DB, data_01814, key)`, so both shards target the same server and therefore return identical rows (0..99). With `optimize_distributed_group_by_sharding_key = 1` the coordinator skips the cross-shard deduplication merge under the assumption that data is partitioned by the sharding key. Each shard pre-aggregates and emits its own top-10 keys (0..9); the coordinator's plan reduces to `Limit(10) <- Union(shard1, shard2)`, which takes the first 10 rows in arrival order.

Per-shard row production timing depends on `max_threads`, page-cache injection, filesystem-cache settings and `prefer_localhost_replica`, which the CI randomizes. Under some combinations the coordinator sees e.g. `0_shardA, 0_shardB, 1_shardA, ..., 8_shardA` and LIMIT 10 fires on `8` — producing `0,0,1,2,3,4,5,6,7,8` instead of `0..9`:

```
 distributed_push_down_limit=1
 0
+0
 1
 2
 3
 4
 5
 6
 7
 8
-9
 40 40
```

#### Repro

I reproduced the failure at 8/30 (~27%) locally with the exact randomized-settings snapshot taken from PR #105110's CIDB record. The key surviving settings (those not overridden by the test itself) are `max_threads=3` together with cache/page-cache injection settings; removing `--max_threads 3` alone dropped the failure rate to 1/30, confirming the timing race.

#### Fix

The check that matters is `read_rows = 40` per shard, i.e. the LIMIT was actually pushed down (the `distributed_push_down_limit = 0` sibling subtest reads 100/shard, see the existing assertion `100 100`). The specific keys `0..9` are not a correctness property of the optimization in this overlapping-shard setup — they are merely what happens to come out when shard 1 wins the network race.

Discard the per-key output with `--format Null` so the test asserts only on `read_rows`. The OFFSET subtest below uses `ORDER BY key DESC` and is deterministic — left untouched.

With the fix: 50/50 passes locally on (a) the failing-settings combo and (b) the default `--test-runs 50` with full randomization.

#### CI report

- Triggering report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105110&sha=f6ae7ff78cc0306e215e41c5384680da11fde14c&name_0=PR&name_1=Stateless%20tests%20%28amd_debug%2C%20parallel%29
- PR that exposed it: https://github.com/ClickHouse/ClickHouse/pull/105110

#### Pre-PR validation gate

| Question | Answer |
|---|---|
| (a) Deterministic repro | Yes — 8/30 with the exact CIDB randomization snapshot |
| (b) Root cause explained | Yes — see above |
| (c) Fix matches root cause | Yes — drops a brittle assertion that is not what the optimization is supposed to guarantee in this setup |
| (d) Test intent preserved | Yes — still verifies `read_rows = 40` per shard (the actual pushdown invariant) and the OFFSET subtest is unchanged |
| (e) Both directions demonstrated | Without fix: 8/30 fail; with fix: 50/50 pass |
| (f) Fix is general | Yes — test fix only, no source code changes |

Session: cron:clickhouse-ci-task-worker:20260517-111500

### Version info
- Merged into: `26.6.1.675`


groeneai:
cc @azat — could you review this? Test-only fix for chronic flaky `01814_distributed_push_down_limit` (6 unrelated PRs in 30 days). Root cause: the test uses `remote('127.{2,3}', ...)` so both shards return identical data, and `optimize_distributed_group_by_sharding_key=1` skips the cross-shard dedup merge — so under certain randomized timing the coordinator's `Limit(10) ← Union` takes `0_A, 0_B, 1_A, ..., 8_A` instead of `0..9`. Fix discards the brittle per-key output with `--format Null` while keeping the actual pushdown invariant (`read_rows = 40` per shard). 50/50 passes locally on both the failing-settings combo and full randomization.

azat:
>I reproduced the failure at 8/30 (~27%) locally with the exact randomized-settings snapshot taken from PR https://github.com/ClickHouse/ClickHouse/pull/105110's CIDB record. The key surviving settings (those not overridden by the test itself) are max_threads=3 together with cache/page-cache injection settings; removing --max_threads 3 alone dropped the failure rate to 1/30, confirming the timing race.

Please provide logs, I don't think this is true, since I cannot reproduce it once

Regardless your comments, please clarify how it is possible that block from one server can be mixed with block from another

groeneai:
Thanks for pushing back. You are right that my earlier "block from one server mixed with block from another" wording was wrong, and I owe you a precise mechanism plus a reproducer. Both below.

## 1. Logs / reproducer

It reproduces reliably once the per-shard aggregation emits the keys in *small* chunks instead of one chunk of 10. The randomized setting that does this in the failing CI run (PR #105110, `Stateless tests (amd_debug, parallel)`, 2026-05-16) is `aggregation_in_order_max_block_bytes` (it was randomized to 2202), together with `max_threads 3` and `prefer_localhost_replica 0`. Note that CI's own `--diagnose-random-settings` on that very record reported `Runs: 33, Failed: 4` (flaky 4/33), so it was reproducing in CI too.

Minimal local reproducer against master HEAD (`fc73f46`, debug):

```
create table data_01814 (key Int) Engine=MergeTree() order by key settings index_granularity=10 as select * from numbers(100);

for i in $(seq 1 100); do
  clickhouse-client \
    --prefer_localhost_replica 0 --max_threads 3 --max_block_size 20 \
    --optimize_aggregation_in_order 1 --aggregation_in_order_max_block_bytes 2202 \
    --distributed_push_down_limit 1 --optimize_skip_unused_shards 1 \
    --optimize_distributed_group_by_sharding_key 1 \
    -q "select * from remote('127.{2,3}', currentDatabase(), data_01814, key) group by key limit 10" \
  | xargs
done | sort | uniq -c
```

Result on my run:

```
  80 0 1 2 3 4 5 6 7 8 9
  20 0 0 1 2 3 4 5 6 7 8
```

That `0 0 1 2 3 4 5 6 7 8` is exactly the diff CIDB recorded for #105110 (`+0` near the top, `-9` at the bottom). `read_rows` is still 40 per shard in both the good and the bad runs, i.e. the LIMIT pushdown itself is working; only the *selection of which keys survive* is racy. With `prefer_localhost_replica 1` (default) I still get 15/60, so it is not specific to the remote read path.

## 2. How the two shards interleave (corrected)

Rows from the two shards are never spliced inside one block. The interleaving is at *whole-chunk* granularity, and there is no cross-shard merge because of `optimize_distributed_group_by_sharding_key=1`:

- With `optimize_distributed_group_by_sharding_key=1` + `distributed_push_down_limit=1`, the initiator asks each shard for `WithMergeableStateAfterAggregationAndLimit` (`StorageDistributed.cpp`, `default_stage` at lines 613-615, and the LIMIT/OFFSET branch at 659-660). Because the data is assumed partitioned by the sharding key, the coordinator deliberately does NOT add a merging/aggregating transform across shards.
- `ReadFromRemote::initializePipeline` just unites the per-shard pipes (`Pipe::unitePipes`, `ReadFromRemote.cpp:843`). There is no sorting/merging step on top, only the outer `LIMIT`.
- The united streams feed the limit through a `ResizeProcessor`, which by design "Pulls data from arbitrary input (whenever it is ready) ... Doesn't preserve an order of data" (`ResizeProcessor.h:19`). So which shard's chunk is consumed first is a scheduling race.
- `LimitTransform` pulls and pushes whole chunks per port and counts rows across them (`LimitTransform.cpp:198`, `:259`); it stops as soon as 10 rows have passed.

So when `aggregation_in_order_max_block_bytes` is tiny, each shard emits `0..9` as several small chunks (e.g. `[0]`, `[1]`, ...). If shard A's `[0]` chunk is consumed, then shard B's `[0,1,...]` chunk, the first 10 rows the LIMIT sees are `0,0,1,2,...,8`, and `9` never gets a turn. Each individual block is still internally ordered and belongs to a single shard; the duplication/loss is purely an artifact of interleaving two independently-LIMITed, overlapping shard streams with no cross-shard dedup.

The test uses `remote('127.{2,3}', ...)`, so both shards hold identical data (0..99). The pinned `0..9` reference therefore encodes an assumption that does not hold for overlapping shards under this optimization. The actual invariant the test verifies, `read_rows = 40` per shard (LIMIT pushed down, vs 100 without it), is unaffected. The PR keeps that check and drops only the non-deterministic per-key assertion by discarding the row output with `--format Null`.

Happy to adjust the approach if you'd prefer a different framing (e.g. pin `aggregation_in_order_max_block_bytes`/`max_threads` instead), but discarding the inherently-nondeterministic key output while keeping the `read_rows` invariant seemed the most honest fix.

groeneai:
CI finished on `c6583ab07ece`. This is a test-only change (01814_distributed_push_down_limit .sh/.reference); all builds and the vast majority of functional checks pass. The reds are not caused by this PR:

- `Stateless tests (arm_asan_ubsan, azure, sequential)` — `Server died` / `AddressSanitizer` (STID 3356-3d55) / `03594_system_grants_parameters`
- `Stateless tests (amd_asan_ubsan, db disk, distributed plan, sequential)` — `Server died` / `Segmentation fault` (STID 1367-4c2e) / `03531_check_count_for_parquet`

Both are the Azure/ASan secondary-allocator OOM mass outage of 2026-06-10/11 (ASan `LargeMmapAllocator` OOM with no ClickHouse frames; STID 1367-* and 3356-* families; 40+ unrelated PRs and master affected over 06-10/11). Orthogonal to a distributed-LIMIT stateless test change.

`CH Inc sync` red is the internal mirror sync, expected on fork PRs. Ready for review.

tests/queries/0_stateless/01814_distributed_push_down_limit.sh
```diff
@@ -97,7 +97,15 @@ function test_distributed_push_down_limit_1()
...
-    test_distributed_push_down_limit_with_query_log "${args[@]}"
+    # Use --format Null because the selected `key` values are not deterministic when
+    # both shards have overlapping data (this test uses `remote('127.{2,3}', ...)` so
+    # both shards return the same rows). With `optimize_distributed_group_by_sharding_key=1`
+    # the coordinator does NOT deduplicate across shards under the assumption that data
+    # is partitioned by the sharding key. Depending on per-shard delivery timing the
+    # coordinator can take e.g. `0,0,1,2,...,8` instead of `0..9` before LIMIT 10 fires.
+    # The check we actually care about is `read_rows = 40` per shard (i.e. the LIMIT was
+    # pushed down — without it each shard reads 100 rows, see the push_down_limit=0 case).
+    test_distributed_push_down_limit_with_query_log "${args[@]}" --format Null
```
azat: Let's also check that it returns 10 rows, so use " | wc -l" instead

azat (changes requested):

tests/queries/0_stateless/01814_distributed_push_down_limit.sh
groeneai: Done in 30044225. The SELECT output now goes through `| wc -l` (in the shared helper, so it covers both the `push_down_limit=0` and `=1` cases), and the reference asserts `10` rows. The `read_rows = 40` per-shard push-down check is kept via the existing retry loop.

Verified on a debug build (master 26.6.1) that this is the stable invariant across the shard-arrival race: with the original failing-settings combo (`prefer_localhost_replica=0`, `max_threads=3`, tiny `aggregation_in_order_max_block_bytes`) the `wc -l = 10` assertion passes 50/50, while the old per-key (`0..9`) assertion flakes 41/100 under the same settings. LIMIT 10 always emits exactly 10 rows; only the key ordering races.

groeneai:
### Pre-PR validation gate (commit 30044225)

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. With `max_block_size=1` + tiny `aggregation_in_order_max_block_bytes`, `select * from remote('127.{2,3}', db, data_01814, key) group by key limit 0,10` produces a raced key ordering in ~44% of runs (87/200), e.g. `[0 1 2 3 4 5 6 7 8 0]`, `[0 0 1 1 2 2 3 3 4 4]`. |
| b | **Root cause explained?** | Yes. `remote('127.{2,3}')` makes both shards return the same rows. Under `optimize_distributed_group_by_sharding_key=1` the coordinator skips cross-shard dedup, so the plan is `Limit(10) <- Union(shard1, shard2)`. The two shard streams interleave at whole-chunk granularity (ResizeProcessor does not preserve order), so per-shard timing decides which rows the outer LIMIT takes. The keys race, but the row count is always 10. |
| c | **Fix matches root cause?** | Yes. The brittle part was asserting the exact keys (`0..9`). Replacing the discarded output (`--format Null`) with `| wc -l` asserts the actual invariant (LIMIT returns 10 rows) without depending on the racing key order. |
| d | **Test intent preserved / new tests added?** | Yes. The push-down coverage is intact: `read_rows = 40` per shard (LIMIT pushed down) is still asserted via the retry loop, and the new `wc -l = 10` check restores row-count coverage that `--format Null` had dropped. No coverage lost; coverage added. |
| e | **Both directions demonstrated?** | Yes. Under the original failing-settings combo (`prefer_localhost_replica=0`, `max_threads=3`, tiny `aggregation_in_order_max_block_bytes`): new `wc -l = 10` assertion passes 50/50; old per-key (`0..9`) assertion flakes 41/100. |
| f | **Fix is general, not a narrow patch?** | N/A (test-only change, no source code modified). The `wc -l` check lives in the shared helper, so it applies symmetrically to both the `push_down_limit=0` and `=1` subtests. |

Session id: cron:clickhouse-worker-slot-10:20260611-124700

azat (approved):

Issue #105009, opened 2026-05-15
`WHERE x OR x` silently drops rows whose low byte is 0 on non-UInt8 columns


fm4v:
## Affected version

`master` at 26.5.1.658 (current branch `nik/sqlancer-pp-check`), reproducible with default settings (`enable_analyzer = 1`, `optimize_extract_common_expressions = 1`).

## Reproducer

Self-contained — runs verbatim under `clickhouse local` against current master:

```sql
CREATE TABLE t (c0 Int64) ENGINE = MergeTree ORDER BY tuple();
INSERT INTO t VALUES (256), (512), (1), (255);

SELECT count() FROM t WHERE c0 OR c0;  -- 2  ✗ (BUG, expected 4)
SELECT count() FROM t WHERE c0 != 0;   -- 4  ✓
```

All four rows are non-zero. `WHERE c0 OR c0` keeps only the rows whose value `& 0xFF != 0` (1 and 255) and drops 256 and 512.

The original SQLancer++ shape is `WHERE (c0 * c0) OR (c0 * c0)` with `c0 = 1702554224`: `c0 * c0 = 2898690885660242176`, `2898690885660242176 % 256 == 0`, so the row gets dropped even though `(c0 * c0) OR (c0 * c0)` evaluates to `1` in any projection.

## Notable properties

* Triggers on any `OR(x, x)` where the two operands compare equal as query-tree nodes. `x AND x` is **not** affected. `x OR 0`, `x OR false`, `0 OR x`, `NOT NOT (x OR x)` are not affected — there's no idempotence collapse in those cases.
* Affects every numeric column type:

  | type | affected? |
  |---|---|
  | `Bool`, `UInt8` | no |
  | `Int16` / `Int32` / `Int64` | yes — drops rows where `value & 0xFF == 0` |
  | `UInt16` / `UInt32` / `UInt64` | yes — same |
  | `Float32` / `Float64` | yes — drops rows where `static_cast<UInt8>(value) == 0` |
* Per-row evaluation in `SELECT` is correct: `SELECT (c0 * c0) OR (c0 * c0) FROM t` returns `1` for the dropped row. Only the filter step is wrong.
* `enable_analyzer = 0` (legacy planner): not affected — the legacy planner errors `ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER` instead, which is at least loud.
* `optimize_extract_common_expressions = 0` (analyzer pass disabled): not affected.
[...]

PR #101287, opened 2026-03-31
Fix wrong results for integer boolean predicates in WHERE on MergeTree

groeneai:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Fix wrong query results when a large integer constant (e.g. 256, 2147483648) is used as a boolean predicate in a WHERE clause with AND on MergeTree tables. For example, `SELECT count() FROM t WHERE (2147483648 > b) AND 2147483648` would incorrectly return 0 instead of matching all rows.

### Summary

When a non-UInt8 integer constant is used as a boolean predicate in a WHERE clause with AND (e.g. `WHERE (256 > b) AND 256`), MergeTree incorrectly prunes all parts and returns 0 rows. This affects any integer value where `value % 256 == 0` (256, 512, 65536, 2147483648, etc.).

**Root cause:** `splitFilterNodeForAllowedInputs()` in `VirtualColumnUtils.cpp` reduces `AND(expr_with_column, constant)` to just `constant` when the column is not a virtual column. Since the constant's type (e.g. UInt16) differs from the AND function's return type (UInt8), a numeric cast `CAST(256, 'UInt8')` is applied, truncating 256 → 0 (256 mod 256). This makes `filterPartsByVirtualColumns()` see `always_false` and prune all parts.

**Fix:** Replace the truncating numeric cast with `notEquals(x, 0)`, which correctly converts any numeric value to boolean (UInt8 0/1) without truncation. Values where `value % 256 == 0` now correctly evaluate to true (1).

Closes https://github.com/ClickHouse/ClickHouse/issues/101269
Closes https://github.com/ClickHouse/ClickHouse/issues/99979

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### Version info
- Merged into: `26.4.1.461`

PR #105155, opened 2026-05-17
Regression test for #105009 (WHERE x OR x truncating cast)

groeneai:
Adds a regression test for #105009 (`WHERE x OR x` silently dropping rows on non-`UInt8` numeric columns).

The underlying bug was fixed independently by PR #103072 (merged 2026-05-18, commit `3ecdf8f8e85b` "Fix semantics"). That PR wraps the single-argument type-mismatch case in `tryOptimizeCommonExpressionsInOr`/`InAnd` as `and(arg, 1)`, which uses `and`'s internal `!= 0` semantics to avoid the truncating `_CAST(arg, UInt8)`.

This PR now contains only the regression test (`04253_105009_or_x_x_truncating_cast.sql`) plus a small documentation comment on the existing `03262_common_expression_optimization.sql` cross-referencing #105009. The test covers 30+ cases including `Int16/32/64`, `UInt16/32/64`, `Float32/64`, `Bool`, `UInt8`, `Nullable(...)`, `LowCardinality(Nullable(Int64))`, mixed-zero-and-`NULL`, and a projection-vs-filter agreement check. It guards against either fix regressing.

- CI Fix or improvement (changelog entry is not required)

Add regression test for #105009 (fix already landed via #103072).


groeneai:
## Pre-PR validation gate

Per [TASK.md Phase 4 step 9](https://github.com/groeneai/internal-task-md) — all five questions answered before submission:

**a) Deterministic repro?** Yes. Self-contained one-liner from #105009:
```sql
CREATE TABLE t (c0 Int64) ENGINE = MergeTree ORDER BY tuple();
INSERT INTO t VALUES (256), (512), (1), (255);
SELECT count() FROM t WHERE c0 OR c0;
```
Always returns `2` on master (bug) and `4` with the fix.

**b) Root cause explained?** Yes. `tryOptimizeCommonExpressionsInOr` collapses `OR(c0, c0)` → `c0` (an `Int64`). It then needs the result to match the original `or` return type (`UInt8`) and previously called `_CAST(c0, 'UInt8')` to coerce. `_CAST` on a wider numeric narrows to the low byte, so `256 → 0`, `512 → 0`, etc. The filter then evaluates the result column as `0` for those rows and drops them. `EXPLAIN actions=1` shows the offending `_CAST(__table1.c0, 'UInt8'_String)` filter column on master.

**c) Fix matches root cause?** Yes. Replace the truncating cast with `notEquals(x, 0)`, which yields a proper `UInt8` boolean without value loss. The truncating cast is skipped only when the source is already `UInt8`-shaped (`Bool` / `UInt8`, optionally wrapped in `Nullable` / `LowCardinality`), where `_CAST` is value-preserving. Same pattern as #101287 / #101269 fixed in `VirtualColumnUtils.cpp`.

**d) Test intent preserved? New tests added?** Yes.
- Added `04253_105009_or_x_x_truncating_cast.sql` (17 assertions): the reduced reproducer, the SQLancer++ shape `(c0 * c0) OR (c0 * c0)` with `c0 = 1702554224`, per-type coverage (`Int16/32/64`, `UInt16/32/64`, `Float32/64`, `UInt8`, `Bool`), mixed zero/non-zero values, and projection-vs-filter agreement.
- Updated `03262_common_expression_optimization` reference for the `(x AND x) OR (x AND x)` Int64 case — it was previously pinning the buggy `_CAST(x, 'UInt8')` plan; the new plan is `notEquals(x, 0)`. The Nullable `(B AND C) OR (B AND C AND toNullable(F))` case is unchanged (`UInt8` source, value-preserving cast).

**e) Both directions demonstrated?** Yes. The reduced reproducer returns `2` on master `a0b6b0f8c3e7` and `4` on the fix branch. `EXPLAIN actions=1` confirms the filter column flips from `_CAST(__table1.c0, 'UInt8'_String)` to `notEquals(__table1.c0, _CAST(0_Int64, 'Int64'_String))`.

- Same bug pattern existed in the sibling `tryOptimizeCommonExpressionsInAnd` (line ~555 in the original file) — fix is applied to both sites with a shared `buildBooleanCoercion` helper.
- Not a defensive symptom-guard: replaces the wrong coercion function with the correct one at the actual point of truncation. The legitimate widening case (`UInt8 → Nullable(UInt8)`) keeps `_CAST` because it's already value-preserving there.

Local test runs against the affected files all pass: `04253_105009_or_x_x_truncating_cast` (new), `03262_common_expression_optimization` (updated), `03285_analyzer_extract_common_expr_bug`, `03285_analyzer_optimize_disjunctions`, `00179_lambdas_with_common_expressions_and_filter`, `01625_constraints_index_append`, `02564_analyzer_ssb_cross_to_inner`, `03640_skip_indexes_data_types_with_or`.

Session: `cron:clickhouse-ci-task-worker:20260517-124500`

groeneai:
cc @vdimir @novikd @alexey-milovidov — analyzer optimizer fix for #105009 (`WHERE x OR x` silently dropping rows on wider numeric columns). Same coercion-via-truncating-cast pattern that was fixed in `VirtualColumnUtils.cpp` (#101287) — this PR fixes the analyzer-pass twin in `LogicalExpressionOptimizerPass`. Helper `buildBooleanCoercion` is shared by both `tryOptimizeCommonExpressionsInOr` and `tryOptimizeCommonExpressionsInAnd`. cc @fm4v as the issue reporter — your root-cause analysis in #105009 is exactly the change here.

groeneai:
Pushed `26dc890e` addressing the bot's catch.

`expression_type->getDefault()` returns `Null()` for `Nullable(T)`, so the previous patch built `notEquals(x, NULL)` — every non-`NULL` row silently dropped (strictly worse than the truncating-cast bug it was supposed to fix). Switched to `removeNullable(recursiveRemoveLowCardinality(expression_type))->getDefault()` + `Nothing` fallback, mirroring the precedent in `Storages/VirtualColumnUtils.cpp::splitFilterNodeForAllowedInputs` (#101269, #101287, #101433, #103049). The `Nothing` fallback covers `Nullable(Nothing)` (bare-`NULL` literal): `x` is always `NULL` there, so `notEquals(x, NULL) = NULL = false` is correct.

Coverage extended in `04253_105009_or_x_x_truncating_cast.sql`:
- `Nullable(Int16/Int32/Int64/UInt32/UInt64/Float64)` with `OR(x,x)` — must return 4 (was 0 without the fix)
- `LowCardinality(Nullable(Int64))` — same wrap problem via the dictionary type, also covered
- `Nullable(Int64)` with mixed `0`/`NULL`/non-zero — 2 (zeros and NULLs filtered, 2 non-zero survive)
- `Nullable(Nothing) filter` (`WHERE NULL OR NULL`) — 0
- `Nullable(Int64) (x AND x) OR (x AND x)` — same coercion path in `tryOptimizeCommonExpressionsInAnd`

Validation gate, both directions:

| # | Question | Result |
|---|---|---|
| a | Deterministic repro? | Yes — without the fix, every new Nullable case returns 0 (all non-`NULL` rows dropped); with the fix, expected counts. |
| b | Root cause explained? | `DataTypeNullable::getDefault() = Null()` → `notEquals(x, NULL) = NULL` → filter treats as false → row dropped. |
| c | Fix matches root cause? | Build the zero constant from the nested type's default, not the wrapper's. `removeNullable` + `recursiveRemoveLowCardinality` cover both wrappers; `Nothing` fallback covers the bare-`NULL` edge. |
| d | Test intent preserved? | All prior `Int64`/`UInt8`/`Bool`/`Float`/`projection agreement` cases unchanged. New cases extend the same intent to nullable/LC types. |
| e | Both directions demonstrated? | Without fix: 13 new Nullable cases fail (return 0 instead of 4/2/0). With fix: 31/31 pass; `--test-runs 30` with randomization passes 30/30. |

cc @alexey-milovidov

fm4v:
@groeneai why check failed: Bugfix validation (functional tests)

groeneai:
@fm4v Sorry for the slow response. Here's what happened with `Bugfix validation (functional tests)`:

Master HEAD already contains a fix for #105009. It landed via PR #103072 (one day before this Bugfix validation ran):

- PR #103072 (merged 2026-05-18 18:26 UTC) primarily fixes #99832 (`TOO_FEW_ARGUMENTS_FOR_FUNCTION` for `and(single_arg)` after common-expression extraction).
- Its follow-up commit `3ecdf8f8e85b` ("Fix semantics") also handles the truncating-cast bug. In both `tryOptimizeCommonExpressionsInOr` and `tryOptimizeCommonExpressionsInAnd`, when the collapsed result is a single argument whose type does not match the original `or`/`and` result type, it wraps as `and(arg, 1)`. The `and` function applies `!= 0` semantics to each argument internally, so wider numeric types (`Int64`, `Float64`, ...) are coerced to boolean correctly instead of being truncated by `_CAST(arg, UInt8)`. That covers the same cases as the `notEquals(x, 0)` fix in this PR, including `Nullable` and `LowCardinality` wrappers.

Bugfix validation downloads the master HEAD binary, runs the new regression test, and expects it to FAIL on master (proving the bug exists) and PASS on the PR (proving the fix works). Since master is already fixed, the test passes on both sides, so the check reports `Failed to reproduce the bug` -> FAIL.

Path forward: I'll drop the now-redundant C++ change here and keep the comprehensive regression test (30+ cases incl. `Nullable`, `LowCardinality`, `Float`, mixed-zero-and-`NULL`, projection agreement) as a guard against either fix regressing. Reclassifying as `CI Fix or improvement` and removing the `pr-bugfix` label so Bugfix validation is skipped. Will push the cleanup commit shortly.

If you'd prefer this closed as a duplicate of #103072 instead, just say so.

groeneai:
Cleanup pushed: 15d3fb72dc7a removes the now-redundant C++ change. The PR now contains only the regression test (`04253_105009_or_x_x_truncating_cast.sql`/`.reference`, 30+ cases) plus a documentation comment in `03262_common_expression_optimization.sql`. PR title and changelog category updated to `CI Fix or improvement`.

I do not have admin rights to remove the `pr-bugfix` label myself, so `Bugfix validation (functional tests)` will keep running and keep reporting `Failed to reproduce the bug` (master already contains the fix from #103072). Could a maintainer please drop the `pr-bugfix` label so Bugfix validation falls back to the sanity test? `cc @alexey-milovidov @vdimir`

PR #105158, opened 2026-05-17
Fix UBSan signed integer overflow in `Decimal` arithmetic operators (STID 0937-2022)

groeneai:
The AST fuzzer (`arm_asan_ubsan`) hit signed integer overflow inside
`Decimal<long>::operator*=` / `operator+=` in `base/base/Decimal.cpp`
three times in four days (master 2026-05-13, PR #100177, PR #104545; UBSan
STID `0937-2022`):

```
/ClickHouse/base/base/Decimal.cpp:24:88: runtime error: signed integer
overflow: 1921680119216801192 * 10 cannot be represented in type 'long'

#0 DB::Decimal<long>::operator*=(long const&)   base/base/Decimal.cpp:24
#1 DB::readDigits<false, DB::Decimal<long>>     src/IO/readDecimalText.h:108
#2 DB::readDecimalText<DB::Decimal<long>, bool> src/IO/readDecimalText.h:155
#3 DB::tryReadDecimalText<DB::Decimal<long>>    src/IO/readDecimalText.h:199
#4 DB::normalizeParameter(...)                  src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesHelpers.cpp:58
```

The trigger path is the timeseries aggregate-function parameter parser,
which calls `tryReadDecimalText(buf, value, 20, scale)` with a string-typed
parameter and `Decimal<Int64>` as the underlying type. The digit
accumulator `x *= 10; x += digit;` overflows `Int64` once the partial value
crosses `2^63`.

`Decimal<T>`'s operator overloads are unchecked by design — ClickHouse
relies on well-defined two's-complement wrap-around in many code paths
(digit accumulation, running sums in aggregates, scale-multiplier
conversion). Code paths that need overflow detection use the explicit
helpers in `base/arithmeticOverflow.h` (`common::addOverflow`,
`common::mulOverflow`, etc.). The bug is that the wrap-around is technically
undefined behaviour at the C++ language level, so UBSan flags it.

Mark `Decimal<T>`'s compound-assignment operators, cross-type
compound-assignment operators, and free-function arithmetic operators
(`+`, `-`, `*`, `/`, unary `-`) with `NO_SANITIZE_UNDEFINED`. This matches
the existing pattern on the sibling `addOverflow` and `negateOverflow`
helpers in the same file (which were added previously for exactly this
purpose in `sumWithOverflow` and friends). No behaviour change in release
builds — the wrap was already happening; UBSan just no longer flags it.

Add a regression test `04254_decimal_parsing_ubsan_stid_0937_2022.sql`
that exercises the `readDecimalText` path through
`timeSeriesResampleToGridWithStaleness` with 19- and 20-digit string
parameters, plus a happy-path regression guard.

Verified locally on `build-asan` (clang-21,
`-DSANITIZE=address,undefined`): before the fix, the test query aborts
with the UBSan diagnostic above; after the fix the same query returns
`BAD_ARGUMENTS` cleanly (the post-parse `common::mulOverflow` check in
`tryReadDecimalText` rejects the value). 20/20 randomized runs of the new
test pass; existing decimal (`00700_decimal_arithm`, `_aggregates`,
`_math`) and timeseries (`02538_nullable_array_tuple_timeseries`,
`03222_create_timeseries_table`, `03254_timeseries_functions*`,
`04106_timeseries_bucket_count_overflow`) tests continue to pass.

CI report (one of three observations): https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=104545&sha=9004589d3c0c3df141a51bc5525f980149223295&name_0=PR&name_1=AST%20fuzzer%20%28arm_asan_ubsan%29

Not for changelog. Suppress UBSan signed-integer-overflow reports inside `Decimal<T>` arithmetic operator overloads (`+=`, `-=`, `*=`, `/=`, `%=`, `+`, `-`, `*`, `/`, unary `-`). The operators already had wrap-around semantics; callers that need overflow detection use the explicit helpers in `base/arithmeticOverflow.h`. Matches the pre-existing `NO_SANITIZE_UNDEFINED` attribute on sibling `addOverflow` and `negateOverflow` helpers.

session-id: cron:clickhouse-ci-task-worker:20260517-143000

---

> [!NOTE]
> **Medium Risk**
> Touches core `Decimal<T>` arithmetic operators by disabling UBSan undefined-behavior checks for `+/-/*` paths; while intended to preserve existing wrap-around semantics, it could mask real overflow-related bugs in sanitized builds if misapplied.
> 
> **Overview**
> Suppresses UBSan signed-integer-overflow reports in `Decimal<T>` by marking wrap-around arithmetic operators (`+=`, `-=`, `*=`, cross-type variants, free `+`, `-`, `*`, and unary `-`) as `NO_SANITIZE_UNDEFINED`, while intentionally leaving `/`/`/=`/`%=` instrumented to still catch divide-by-zero and `INT_MIN / -1`.
> 
> Adds a stateless regression test (`04266_decimal_parsing_ubsan_stid_0937_2022`) that exercises decimal digit accumulation via `timeSeriesResampleToGridWithStaleness` with oversized string parameters (expecting `BAD_ARGUMENTS`) plus a normal-range sanity check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f9107ed3d65fb2257e5bb26e309a7a1e70d601d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.290`


groeneai:
### Pre-PR validation gate

**a) Deterministic repro?** ✅ — `clickhouse-local -q "SELECT timeSeriesResampleToGridWithStaleness('99999999999999999999', '99999999999999999998', '1', '1')(timestamp, value) FROM ts FORMAT Null"` triggers `runtime error: signed integer overflow: 999999999999999999 * 10 cannot be represented in type 'long'` on every run of an `address,undefined`-sanitized build. After applying the fix the same query exits cleanly with `Code: 36. DB::Exception: Cannot parse start parameter ... (BAD_ARGUMENTS)`.

**b) Root cause explained?** ✅ — `Decimal<T>`'s implicit operator overloads in `base/base/Decimal.cpp:22-26` and `:71-75` perform unchecked arithmetic on the underlying integer type. UBSan correctly flags the wrap as undefined behaviour even though ClickHouse relies on the wrap as well-defined semantics in many code paths (digit accumulation in `readDigits`, running sums in aggregate functions, scale conversion). Code paths that need overflow detection use the explicit helpers in `base/arithmeticOverflow.h` (`common::mulOverflow`, `common::addOverflow`, …) — the operators themselves are intentionally wrap-around. The sibling helpers `addOverflow` (line 28) and `negateOverflow` (line 76) already have the `NO_SANITIZE_UNDEFINED` attribute for exactly this reason.

**c) Fix matches root cause?** ✅ — Marking every `Decimal<T>` arithmetic operator with `NO_SANITIZE_UNDEFINED` makes the wrap-around well-defined at the C++ language level. Not a defensive symptom-mask: the wrap is the intended behaviour, not a bandaid hiding a real bug; the fix is identical in spirit to the pre-existing `NO_SANITIZE_UNDEFINED` on `addOverflow` and `negateOverflow` in the same file.

**d) Test intent preserved? / New tests added?** ✅ — No existing test was modified. A new regression test `tests/queries/0_stateless/04254_decimal_parsing_ubsan_stid_0937_2022.sql` exercises the `readDecimalText` path via `timeSeriesResampleToGridWithStaleness` with both a 20-digit string parameter (triggers `operator*=`) and a 19-digit string parameter just over `Int64` max (triggers `operator+=` — the AST fuzzer's `9223372036854775800 + 8` shape), plus a happy-path case to guard against accidental behaviour change for in-range inputs.

**e) Demonstrated in both directions?** ✅ — Before the fix on `build-asan` (clang-21, `-DSANITIZE=address,undefined`): UBSan aborts with the exact diagnostic from CI. After the fix: same query returns `BAD_ARGUMENTS` cleanly, no UBSan output. `tests/clickhouse-test --test-runs 20 04254_decimal_parsing_ubsan_stid_0937_2022` → 20/20 pass. Existing decimal (`00700_decimal_arithm`, `_aggregates`, `_math`) and timeseries (`02538_nullable_array_tuple_timeseries`, `03222_create_timeseries_table`, `03254_timeseries_functions`, `_various_arguments`, `_group_array`, `04106_timeseries_bucket_count_overflow`) tests continue to pass.

**f) Fix is general, not a narrow patch?** ✅ — The UBSan report only hit `Decimal<long>::operator*=` and `operator+=` (Decimal64 / `Int64`) via the AST fuzzer, but the same overflow class exists in the sibling cross-type `Decimal<U>` overloads (lines 32-36 of `Decimal.cpp`) and the free-function arithmetic operators (lines 71-75). All are marked `NO_SANITIZE_UNDEFINED` in the same way so future code paths that exercise them (Decimal128 / Decimal256 accumulators, mixed-type arithmetic, free-function `Decimal<T> + Decimal<T>`) cannot regress with a different fuzzer trace.

session-id: cron:clickhouse-ci-task-worker:20260517-143000

groeneai:
cc @alexey-milovidov @yariks5s — could you review this when you have a moment? The fix is the same shape as the existing `NO_SANITIZE_UNDEFINED` on `addOverflow` and `negateOverflow` in this file, extended to the implicit operator overloads. UBSan-only impact in sanitizer builds (3 master+PR hits over 4 days, AST fuzzer); no behaviour change in release.

session-id: cron:clickhouse-ci-task-worker:20260517-143000

groeneai:
@clickhouse-gh Good catch — fixed in 355717a35914.

`NO_SANITIZE_UNDEFINED` now stays only on the overflow-dependent operators (`+=`, `-=`, `*=`, free `+` / `-` / `*`, unary `-`, and the existing `addOverflow` / `negateOverflow` helpers). `/=`, `%=` and the free `/` are no longer suppressed, so UBSan continues to report `divide-by-zero` and `INT_MIN / -1` inside `Decimal` arithmetic.

The rationale comments in `Decimal.h` and `Decimal.cpp` are updated to document the narrower contract.

Verified on `build-asan` (`clang-21`, `-DSANITIZE=address,undefined`):
- Regression test `04254_decimal_parsing_ubsan_stid_0937_2022.sql` — the trigger for STID 0937-2022 goes through `value *= 10` / `value += digit` in `readDecimalText`, both still marked `NO_SANITIZE_UNDEFINED`, so it continues to pass (10/10 runs with randomized settings).
- `SELECT toDecimal64(7,0) / toDecimal64(3,0)` and friends return the expected results.

Session id: cron:clickhouse-ci-task-worker:20260517-171500

alexey-milovidov (approved):

antonio2368:
```
ERROR: Failed to insert data into CI DB, exception [HTTPSConnectionPool(host='play.clickhouse.com', port=443): Max retries exceeded with 
                      url: /?database=default&query=INSERT+INTO+checks+FORMAT+JSONEachRow&date_time_input_format=best_effort&send_logs_level=warning (Caused by 
                      NewConnectionError("HTTPSConnection(host='play.clickhouse.com', port=443): Failed to establish a new connection: [Errno 61] Connection 
                      refused"))] (from: Fast test (arm_darwin))
```

Even on retry. I guess we can override and merge this as everything else is green?

alexey-milovidov (approved):

PR #97677, opened 2026-02-23
Fix flaky test test_storage_s3_queue/test_migration

alexey-milovidov:
## Summary

- Fix flaky `test_storage_s3_queue/test_migration` by explicitly cleaning the ZooKeeper path `/clickhouse/databases/replicateddb3` after `DROP DATABASE`, in case the older (24.5) binary didn't fully remove it
- Drop database `r` at the end of each test (not just the table) so the next run starts clean

The test uses a hardcoded ZK path for its `Replicated` database. On retries, after `restart_with_original_version` switches nodes back to 24.5, `DROP DATABASE IF EXISTS r` may leave the ZK path in a partial state (parent exists but `max_log_ptr` is missing), causing `CREATE DATABASE` to fail with `KEEPER_EXCEPTION`.

https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=97671&sha=b978c6092f4758c5cd3b1864248542ef3532c4d0&name_0=PR&name_1=Integration%20tests%20%28amd_asan%2C%20db%20disk%2C%20old%20analyzer%2C%206%2F6%29
https://github.com/ClickHouse/ClickHouse/pull/97671

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a user-readable short description of the changes that goes into CHANGELOG.md):
...

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #93232, opened 2025-12-30
Fix test_storage_s3_queue/test_migration

kssenii:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix bug with S3Queue setting `s3queue_migrate_old_metadata_to_buckets`. Closes https://github.com/ClickHouse/ClickHouse/issues/93392, https://github.com/ClickHouse/ClickHouse/issues/93196, https://github.com/ClickHouse/ClickHouse/issues/81739

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #105159, opened 2026-05-17
Fix NameError masking test_migration diagnostic on count mismatch

groeneai:
The diagnostic branch in `test_storage_s3_queue/test_migration.py::test_migration` referenced undefined variables `total_rows`, `count1`, `count2` in its assertion f-string. When the inner wait loop timed out (count didn't converge in 50s), Python raised `NameError` at f-string evaluation, masking what should have been a clear convergence failure.

CIDB confirms **81 failures** of this form over the last 90 days, with a sharp spike to **32 hits on 2026-05-16** across every sanitizer build (`amd_msan`, `amd_asan_ubsan/db disk/old analyzer`, `amd_tsan`, `amd_llvm_coverage`, `arm_binary/distributed plan`). All hits are **PR-only** — `pull_request_number > 0` across 6+ unrelated PRs (none of which touch s3_queue), making this a developer-experience pain point.

The diagnostic also computed `expected_files` using a wrong file naming pattern (`{files_path}/test_{x}.csv`) — the actual files use `{use_prefix}_{i}.csv` where `use_prefix` advances through `["a","b","c","d","e"]` across each `add_files_and_check` call, and `i` starts at the cumulative `start_ind`. The resulting `missing_files` list would always report every expected file as missing, even when convergence was correct.

This change:
- Computes `count1` and `count2` from per-node `SELECT count()` queries (the variables the original author intended).
- Uses `expected_rows[0]` instead of the never-defined `total_rows`.
- Drops the broken `expected_files`/`missing_files` calculation.
- Reports `distinct files processed` from the existing `processed_files` list as a useful proxy without the buggy naming logic.
- Switches the `processed_files` query from `node` (last value from the `for node in [node1, node2]` loop above) to `node1` for clarity.

Demonstration in both directions:
- **Without fix:** `NameError: name 'total_rows' is not defined`
- **With fix:** `AssertionError: Expected 1000 rows in total, got 300 on node1 and 200 on node2 (500), distinct files processed: 5`

CIDB reference: https://play.clickhouse.com/?user=play#U0VMRUNUIHRlc3RfbmFtZSwgY2hlY2tfbmFtZSwgY291bnQoKSBGUk9NIGRlZmF1bHQuY2hlY2tzIFdIRVJFIHRlc3RfbmFtZSBMSUtFICd0ZXN0X3N0b3JhZ2VfczNfcXVldWUvdGVzdF9taWdyYXRpb24ucHklJyBBTkQgdGVzdF9jb250ZXh0X3JhdyBMSUtFICclTmFtZUVycm9yJScgR1JPVVAgQlkgdGVzdF9uYW1lLCBjaGVja19uYW1lIE9SREVSIEJZIGNvdW50KCkgREVTQw==

Past related fix PRs: #97677, #93232.

...

### Version info
- Merged into: `26.5.1.719`

### DIFF
--- a/tests/integration/test_storage_s3_queue/test_migration.py
+++ b/tests/integration/test_storage_s3_queue/test_migration.py
@@ -171,32 +171,27 @@ def test_migration(started_cluster, setting_prefix, buckets_num):
             time.sleep(1)
 
         if expected_rows[0] != get_count():
-            files_to_generate = expected_rows[0]  # 1 row per file
-            expected_files = [
-                f"{files_path}/test_{x}.csv" for x in range(files_to_generate)
-            ]
-
             for node in [node1, node2]:
                 node.query("SYSTEM FLUSH LOGS")
 
+            count1 = int(node1.query(f"SELECT count() FROM default.{dst_table_name}"))
+            count2 = int(node2.query(f"SELECT count() FROM default.{dst_table_name}"))
+
             processed_files = (
-                node.query(
+                node1.query(
                     f"SELECT distinct(_path) FROM clusterAllReplicas(cluster, default.{dst_table_name})"
                 )
                 .strip()
                 .split("\n")
             )
-
             processed_files.sort()
             logging.debug(f"Processed files: {processed_files}")
-            missing_files = [
-                file for file in expected_files if file not in processed_files
-            ]
-            missing_files.sort()
 
-            assert (
-                False
-            ), f"Expected {total_rows} in total, got {count1} and {count2} ({count1 + count2}, having {len(missing_files)} missing files: ({missing_files})"
+            assert False, (
+                f"Expected {expected_rows[0]} rows in total, "
+                f"got {count1} on node1 and {count2} on node2 ({count1 + count2}), "
+                f"distinct files processed: {len(processed_files)}"
+            )
 
     add_files_and_check()
 


groeneai:
**Pre-PR validation gate (worker session `cron:clickhouse-ci-task-worker:20260517-153000`)**

- **(a) Deterministic repro?** ✅ NameError fires whenever `expected_rows[0] != get_count()` after 50 retries — 81 CIDB-confirmed hits over the last 90 days, spike to 32 on 2026-05-16 alone.
- **(b) Root cause?** ✅ f-string at line 199 referenced `total_rows`, `count1`, `count2` — none of which were ever defined in `add_files_and_check`. Python raised `NameError` at f-string evaluation when the diagnostic branch was reached, masking the real convergence failure.
- **(c) Fix matches root cause?** ✅ The fix defines `count1`/`count2` via direct per-node `SELECT count()` queries (the variables the author originally intended), uses `expected_rows[0]` for the total, and drops the broken `expected_files`/`missing_files` calculation. No widening of bounds, no silencing of randomization.
- **(d) Test intent preserved?** ✅ The assertion still fires on the same condition (`expected_rows[0] != get_count()` after 50 retries). The retry budget is unchanged. The diagnostic now reports what the original author intended (per-node split + cluster total) plus a count of distinct files actually processed.
- **(e) Demonstrated in both directions?** ✅
  - Without fix: `NameError: name 'total_rows' is not defined`
  - With fix (mismatch): `AssertionError: Expected 1000 rows in total, got 300 on node1 and 200 on node2 (500), distinct files processed: 5`
  - With fix (convergence): no assertion triggered
- **(f) Fix is general?** N/A — test-only Python fix, no symmetric C++ paths.

Additional notes:
- The `expected_files = [f"{files_path}/test_{x}.csv" ...]` calculation that was removed had two latent bugs even before NameError fired: it used the wrong file-name prefix (`test_` instead of `{use_prefix}_`) and ignored prefix rotation across `add_files_and_check` calls, so `missing_files` would always have reported every expected file as missing.
- The CIDB confirms this failure pattern is **PR-only** (0 master hits), distributed across 6+ unrelated PRs and 6 sanitizer/coverage build flavors, never reproducing on actual master. Fixing the diagnostic restores the test's debuggability on those PR runs.

groeneai:
cc @kssenii @alexey-milovidov — could you review this? It fixes the NameError-masked diagnostic in `test_storage_s3_queue/test_migration.py` (introduced by #97677 / commit `85202e8d8b3f`): the assertion f-string referenced undefined `total_rows`/`count1`/`count2` so any count-convergence failure surfaced as `NameError` instead of useful per-node info, fully masking the actual failure mode in PR-only CI runs across all sanitizer builds.

alexey-milovidov (approved):

PR #105163, opened 2026-05-17
Revert "Stop the bleeding in function_prop_fuzzer"

alexey-milovidov:
Reverts ClickHouse/ClickHouse#105146

Only if it is green.

Please be extra careful with this PR, do not rush. The previous attempt damaged 204 pull requests and our work was ground to halt: https://github.com/ClickHouse/ClickHouse/pull/105146#issuecomment-4472117956

### Version info
- Merged into: `26.5.1.816`


Algunenano:
Adding it back but disable it by default. The problems that is reporting already existed in this fuzzer (as stuck thread), but it made it more noisy. Having it disabled allow us to have reports without blocking everything

Algunenano:
I think that this can now be merged, as the errors of this fuzzer won't be reported as error but still be logged so we can find them and fix them by searching the CI DB.

antonio2368 (approved):

PR #105167, opened 2026-05-17
Generate system.asynchronous_metrics docs from C++ source

alexey-milovidov:
Reaction to [a comment on #105136](https://github.com/ClickHouse/ClickHouse/pull/105136#issuecomment-4472176623): the description of every async metric already lives in C++ as a string literal next to the metric registration, so the Markdown table should be generated, not hand-edited.

This PR adds the missing copy step:

- `utils/generate-async-metrics-docs` extracts `(name, description)` pairs from `src/Common/AsynchronousMetrics.cpp`, `src/Interpreters/ServerAsynchronousMetrics.cpp`, and `src/Coordination/KeeperAsynchronousMetrics.cpp`. It handles adjacent string-literal concatenation, inline `#define` macros (e.g. `BLOCK_DEVICE_EXPLANATION`, `LOAD_AVERAGE_DOCUMENTATION`), `fmt::format`, local `String` variables, the jemalloc save helpers, and the protocol-thread map literals.
- `docs/en/operations/system-tables/asynchronous_metrics.md` — the manually maintained section is replaced with content rendered between `<!--AUTOGENERATED_METRICS_START-->` / `<!--AUTOGENERATED_METRICS_END-->` markers. Coverage grows from ~153 to 248 entries, including `QueriesMemoryUsage` / `QueriesPeakMemoryUsage` (the gap #105136 was trying to close manually).
- Where the old hand-written Markdown had more detail than the C++ literal, the wording is moved back into the source — `DictionaryTotalFailedUpdates`, the three `DiskTotal_*name*` / `DiskAvailable_*name*` / `DiskUnreserved_*name*` entries, and `ZooKeeperClientLastZXIDSeen`.
- Two preexisting typos exposed by the regeneration are fixed in source: `granulas` → `granules`, `sequantially` → `sequentially`.
- The new metric identifiers are added to the aspell dictionary so the doc passes `check_aspell.sh`.

Documentation for `system.asynchronous_metrics` is now generated from the metric description string literals in C++ source. The generated list covers all 248 metrics (previously ~153 were documented manually), including `QueriesMemoryUsage` and `QueriesPeakMemoryUsage`.

### Version info
- Merged into: `26.5.1.868`


utils/generate-async-metrics-docs
```diff
@@ -0,0 +1,508 @@
+#!/usr/bin/env python3
+"""Generate `system.asynchronous_metrics` documentation from C++ source.
```
azat, line 2: Can we add this script into style check and make sure that the diff is empty? That way the documentation will be always in sync.

alexey-milovidov: I'd expect we run it on every deploy from ClickHouse/clickhouse-docs.

azat: If it will be moved into clickhouse-docs, we can even remove the documentation itself from ClickHouse repo?

utils/generate-async-metrics-docs
```diff
@@ -0,0 +1,508 @@
...
+    "saveJemallocProf": "jemalloc.prof.{}",
+}
+
+JEMALLOC_HELPER_PATTERN = re.compile(
+    r"(?P<helper>saveJemallocMetric|saveAllArenasMetric|saveJemallocProf)"
+    r"\s*<[^>]+>\s*\([^,]+,\s*\"(?P<name>[^\"]+)\"\s*\)"
+)
+
+JEMALLOC_IMPL_PATTERN = re.compile(
+    r"saveJemallocMetricImpl\s*<[^>]+>\s*\([^,]+,\s*\"[^\"]+\"\s*,"
```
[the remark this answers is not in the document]
alexey-milovidov: Good catch — fixed in c5b1184fde6 by loosening the regex to accept either a string literal or a `fmt::format(...)` call as the second argument (the metric name remains the literal third argument). After re-running the generator, the four cache/MergeTree arena metrics now appear in the table.

utils/generate-async-metrics-docs
alexey-milovidov: Yes — once the script gets moved to clickhouse-docs we can drop `docs/en/operations/system-tables/asynchronous_metrics.md` from this repo entirely and have the docs build regenerate it from the C++ sources on every deploy. That removes the periodic-drift problem completely without paying for it in PR CI time on every unrelated change.

azat (approved):
I still think it make sense to run in as part of a style check, but it is up to you to decide

alexey-milovidov:
Ok, just one step at a time...

utils/generate-async-metrics-docs
rschu1ze: It looks like nobody in the docs team was aware so far. I added a note to the ticket I opened yesterday: https://github.com/ClickHouse/clickhouse-docs/issues/6439#issuecomment-4787561083

PR #105170, opened 2026-05-17
Fix parameterized view used twice in same query

### LINKED ISSUE #105153: Using Parameterized View twice in same query
### Company or project name

_No response_

### Describe what's wrong

When doing:
```SELECT .... WHERE (id IN (sub-query)) AND (id IN (sub-query))```

While both the sub-queries are using the same view but with different params, this results in unexpected behaviour. the second sub-query is cached for some reason or being skipped.

Yes

### How to reproduce

```sql
create table testing (id String) ENGINE = MergeTree() ORDER BY id;
CREATE VIEW test_view AS SELECT id from testing WHERE id = {id:String};
insert into testing VALUES ('a');
insert into testing VALUES ('b');

SELECT id
FROM testing
WHERE (id IN (
    SELECT id
    FROM test_view(id = 'a')
)) AND (id IN (
    SELECT id
    FROM test_view(id = 'b')
))

Result: 
1. 'a'
```

### Expected behavior

Because non of the rows consists both ID = 'a' and ID = 'b', the query should result in no rows result. but the second condition for some reason being skipped.

_No response_

### Additional context

The only way we found to fix that is to add `WHERE 1 = 1` or something to the second subquery to declare its different view query.

### Version info
- Resolved by: #105170
- Merged into: `26.5.1.788`

alexey-milovidov:
Fixes #105153.

Two calls to the same parameterized view with different argument values in one query (e.g. `id IN (SELECT id FROM test_view(id = 'a')) AND id IN (SELECT id FROM test_view(id = 'b'))`) were being deduplicated, so the second filter was silently dropped.

Root cause: each call to `buildParameterizedViewStorage` creates a fresh `StorageView` carrying the substituted inner query, but it reuses the original view's `StorageID`. `TableNode::isEqualImpl` / `updateTreeHashImpl` only looked at `StorageID`, so the two `TableNode`s compared equal and `CollectSets` (which keys `PreparedSets` by `getTreeHash`) collapsed them into one IN-set.

Fix: when the storage is a parameterized view, mix the substituted inner query into the tree hash and the equality check. Non-parameterized tables and views are unaffected.

Fix incorrect results when the same parameterized view is referenced more than once in the same query with different argument values. Previously, the analyzer collapsed the calls into one, silently dropping all but the first filter.

### Version info
- Merged into: `26.5.1.788`


alexey-milovidov (commented):
Good.

PR #105179, opened 2026-05-17
Add regression test for ClickHouse-source dictionary with backslash keys

alexey-milovidov:
### Description

Adds a regression test for the `ExternalQueryBuilder` ClickHouse-backend path that uses `IdentifierQuotingStyle::Backticks` and backslash escaping. Switching to Postgres-style `''` escaping for this backend would break ClickHouse queries when keys contain backslashes, because the ClickHouse SQL parser interprets `\` as an escape character.

Originally part of https://github.com/ClickHouse/ClickHouse/pull/104009 but excluded from that PR because the bugfix validation requires the test to fail on master, and this test does not reproduce the original SQL injection bug (which only affected `DoubleQuotes`-style backends like PostgreSQL/MySQL). The original bug is already covered by `gtest_external_query_builder.cpp`, `test_dictionaries_postgresql.py`, and `test_dictionaries_mysql.py`.

Add a regression test for ClickHouse-source dictionaries with backslash and single-quote characters in complex keys.

### Version info
- Merged into: `26.5.1.775`


alexey-milovidov (commented):
Ok.

PR #56057, opened 2023-10-26
Fix missing scalar issue when evaluating subqueries inside table functions

amosbird:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Fix missing scalar issue when evaluating subqueries inside table functions. This fixes #56031.

PR #105194, opened 2026-05-18
Stabilize 02180_group_by_lowcardinality (non-deterministic LIMIT without ORDER BY)

### LINKED ISSUE #36069: `02180_group_by_lowcardinality` flaky
Reports:

https://s3.amazonaws.com/clickhouse-test-reports/36057/a50b12f02ddf60d3c4d06039b31af640c81d7ca1/stateless_tests__debug__actions__[2/3].html
https://s3.amazonaws.com/clickhouse-test-reports/35994/ead90d3ed5e5f00f51eaf0649e10b4510fdd9de0/stateless_tests__release__wide_parts_enabled__actions_.html
https://s3.amazonaws.com/clickhouse-test-reports/34631/acc7046d54738054df39cd43842534265a723d6e/stateless_tests__release__wide_parts_enabled__actions_.html
https://s3.amazonaws.com/clickhouse-test-reports/35832/003ef1f3b8ffcee79ec839ec3ea76b6811760b5d/stateless_tests__release__wide_parts_enabled__actions_.html
https://s3.amazonaws.com/clickhouse-test-reports/36047/88134c86a531d97b961312005a6a2c8f26a69cbc/stateless_tests__release__wide_parts_enabled__actions_.html

Sample output:
```
2022-04-07 12:51:07 --- /usr/share/clickhouse-test/queries/0_stateless/02180_group_by_lowcardinality.reference	2022-04-07 12:50:07.275592168 +0800
2022-04-07 12:51:07 +++ /tmp/clickhouse-test/0_stateless/02180_group_by_lowcardinality.stdout	2022-04-07 12:51:07.411831407 +0800
2022-04-07 12:51:07 @@ -3,8 +3,8 @@
2022-04-07 12:51:07  {"val":"584.4","avg(toUInt32(val))":null}
2022-04-07 12:51:07  {"val":"269","avg(toUInt32(val))":269}
2022-04-07 12:51:07  {"val":"1233.4","avg(toUInt32(val))":null}
2022-04-07 12:51:07 -{"val":"1833","avg(toUInt32(val))":1833}
2022-04-07 12:51:07  {"val":"1009.4","avg(toUInt32(val))":null}
2022-04-07 12:51:07  {"val":"1178.6","avg(toUInt32(val))":null}
2022-04-07 12:51:07  {"val":"372.6","avg(toUInt32(val))":null}
2022-04-07 12:51:07  {"val":"232.4","avg(toUInt32(val))":null}
2022-04-07 12:51:07 +{"val":"901.6","avg(toUInt32(val))":null}
2022-04-07 12:51:07 
2022-04-07 12:51:07 
2022-04-07 12:51:07 Settings used in the test: --max_insert_threads=2 --group_by_two_level_threshold=100000 --group_by_two_level_threshold_bytes=50000000 --distributed_aggregation_memory_efficient=0 --fsync_metadata=0 --priority=3 --output_format_parallel_formatting=0 --input_format_parallel_parsing=1 --min_chunk_bytes_for_parallel_parsing=9879472 --max_read_buffer_size=878154 --prefer_localhost_replica=0 --max_block_size=8374 --max_threads=29
2022-04-07 12:51:07 
2022-04-07 12:51:07 Database: test_v2c2mq
```

groeneai:
The stateless test `02180_group_by_lowcardinality` asserts a row-by-row deterministic output from a query that is inherently non-deterministic. The test currently fails ~10–20 times per day on master under `Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel)` and has been failing across 112 distinct PRs over the last 30 days.

@alexey-milovidov flagged this on https://github.com/ClickHouse/ClickHouse/pull/102039.

### Root cause

The query is:

```sql
SELECT val, avg(toUInt32(val))
FROM t_group_by_lowcardinality
GROUP BY val
LIMIT 10
SETTINGS max_threads = 1, max_rows_to_group_by = 100, group_by_overflow_mode = 'any';
```

With `group_by_overflow_mode = 'any'`, only the first `max_rows_to_group_by` distinct keys make it into the hash table, and which keys those are depends on the order in which rows arrive at the aggregator. `LIMIT 10` without `ORDER BY` then selects ten of those keys in hash-iteration order.

The `max_threads = 1` clause does not help under `ParallelReplicas` (where the table is sharded across replicas and partial aggregations are merged) or under S3 storage (where reads can be split into independent ranges), so the resulting row set varies between runs.

Locally reproduced the exact CI failure diff by changing `max_threads = 1` to `max_threads = 4`: an extra `{"val":null,"avg(toUInt32(val))":null}` appears at the top and the last row drops off — identical to the diff shown in the CI report.

### Fix

This test is a crash-regression test for "Avoid crash in case of GROUP BY LowCardinality(Nullable(String)) column and group_by_overflow_mode='any'" (PR #29637 / #56057). The query is preserved verbatim, including all the problematic settings; only the assertion is changed to `SELECT count() FROM (...)`, which is always 10 (the LIMIT value).

If the underlying crash ever returns, or `LIMIT` stops returning the expected number of rows, the test will detect it. The specific row values were never deterministic to begin with.

Originally tracked in #36069 (2022). The `-- Tags: no-random-settings` workaround masked the issue for the random-settings randomization but the underlying non-determinism returned once `ParallelReplicas` was enabled by default in some CI configurations.

CI report (one of many): https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=102039&sha=eb773718e41db7bafe973e9a67a6a3a195b48fe9&name_0=PR&name_1=Stateless%20tests%20%28amd_llvm_coverage%2C%20ParallelReplicas%2C%20s3%20storage%2C%20parallel%29

Closes #36069

Not for changelog (CI Fix or Improvement).

### Version info
- Merged into: `26.5.1.742`

### DIFF
--- a/tests/queries/0_stateless/02180_group_by_lowcardinality.reference
+++ b/tests/queries/0_stateless/02180_group_by_lowcardinality.reference
@@ -1,10 +1,1 @@
-{"val":"1563.8","avg(toUInt32(val))":null}
-{"val":"891.4","avg(toUInt32(val))":null}
-{"val":"584.4","avg(toUInt32(val))":null}
-{"val":"269","avg(toUInt32(val))":269}
-{"val":"1233.4","avg(toUInt32(val))":null}
-{"val":"1833","avg(toUInt32(val))":1833}
-{"val":"1009.4","avg(toUInt32(val))":null}
-{"val":"1178.6","avg(toUInt32(val))":null}
-{"val":"372.6","avg(toUInt32(val))":null}
-{"val":"232.4","avg(toUInt32(val))":null}
+{"count()":10}
--- a/tests/queries/0_stateless/02180_group_by_lowcardinality.sql
+++ b/tests/queries/0_stateless/02180_group_by_lowcardinality.sql
@@ -1,12 +1,21 @@
 -- Tags: no-random-settings
 
-create table if not exists t_group_by_lowcardinality(p_date Date, val LowCardinality(Nullable(String))) 
+create table if not exists t_group_by_lowcardinality(p_date Date, val LowCardinality(Nullable(String)))
 engine=MergeTree() partition by p_date order by tuple();
 
 insert into t_group_by_lowcardinality select today() as p_date, toString(number/5) as val from numbers(10000);
 insert into t_group_by_lowcardinality select today() as p_date, Null as val from numbers(100);
 
-select val, avg(toUInt32(val)) from t_group_by_lowcardinality group by val limit 10 settings max_threads=1, max_rows_to_group_by=100, group_by_overflow_mode='any' format JSONEachRow;
+-- Regression test for: GROUP BY LowCardinality(Nullable(String)) with group_by_overflow_mode='any'
+-- must not crash. Originally introduced in PR #29637 / #56057.
+--
+-- The exact set of returned rows is non-deterministic: with overflow mode 'any', only the first
+-- `max_rows_to_group_by` distinct keys end up in the hash table, and which keys those are depends
+-- on parallel execution / parallel replicas / S3 read ordering. `LIMIT 10` without `ORDER BY`
+-- further selects 10 of those keys non-deterministically. Therefore we only assert that the query
+-- runs successfully and returns exactly the LIMIT number of rows.
+select count() from (
+    select val, avg(toUInt32(val)) from t_group_by_lowcardinality group by val limit 10 settings max_threads=1, max_rows_to_group_by=100, group_by_overflow_mode='any'
+) format JSONEachRow;
 
 drop table if exists t_group_by_lowcardinality;
-


groeneai:
**Pre-PR validation gate** (TASK.md Phase 4 Step 7):

| # | Question | Answer |
|---|----------|--------|
| a | Deterministic repro? | ✅ Demonstrated locally: running the inner SELECT with `max_threads = 4` produces the exact diff shown in the failing CI report — extra `{"val":null,...}` at the top, missing `{"val":"232.4",...}` at the bottom. |
| b | Root cause explained? | ✅ `group_by_overflow_mode = 'any'` admits only the first `max_rows_to_group_by` keys into the hash table; the set of keys depends on row arrival order at the aggregator. `LIMIT 10` without `ORDER BY` further picks ten of those keys in hash-iteration order. `max_threads = 1` does not constrain this under `ParallelReplicas` or split S3 reads, so the row set varies between runs. |
| c | Fix matches root cause? | ✅ `SELECT count() FROM (...)` removes the dependency on which specific keys end up in the hash table. |
| d | Test intent preserved? | ✅ This is a crash-regression test for "Avoid crash in case of GROUP BY LowCardinality(Nullable(String)) column and group_by_overflow_mode='any'" (PR #29637 / #56057). The query is preserved verbatim — including LowCardinality(Nullable(String)), the GROUP BY, the overflow mode, and the LIMIT — so any crash regression will still surface. The original assertion of specific row values was never actually deterministic. |
| e | Both directions demonstrated? | ✅ `max_threads = 4` on the original query produces the exact CI diff; the new `count()` assertion returns 10 for `max_threads` ∈ {1, 4, 8, 16}. `--test-runs 30` against the local server: 30/30 pass. |
| f | Fix is general, not a narrow patch? | N/A — test fix, not a source code fix. No production logic changed. |

Session id: `cron:clickhouse-ci-task-worker:20260518-040000`.

groeneai:
cc @KochetovNicolai @alexey-milovidov — could you review this test-only fix? The original assertion was non-deterministic under `ParallelReplicas` because `group_by_overflow_mode='any'` admits only the first `max_rows_to_group_by` keys (in arrival order), and the `LIMIT 10` without `ORDER BY` then picks ten of those in hash-iteration order. The query is preserved verbatim — only the assertion changes to `SELECT count() FROM (...)` (always 10), which still detects any future crash regression in the underlying `LowCardinality(Nullable(String))` + `group_by_overflow_mode='any'` path.

alexey-milovidov (approved):

PR #105197, opened 2026-05-18
Optimize min/max aggregates using column statistics

zoomxi:
## Summary

This PR adds an optimization that answers filterless `min()` and `max()` aggregation queries from per-part column statistics without reading column data, avoiding full scans.

## How it works

When a query like `SELECT min(x), max(x) FROM t` runs on a `MergeTree` table with column statistics configured (via `auto_statistics_types = 'basic'`), the optimizer checks whether the aggregate results can be obtained from the per-part `StatisticsBasic` metadata. If all conditions are met, the query plan replaces the `ReadFromMergeTree` step with a `ReadFromPreparedSource` that produces the answer directly from statistics without reading column data.

The optimization applies when:
- All aggregates are `min()` or `max()`
- The table has `auto_statistics_types = 'basic'` and statistics have been materialized for the relevant columns
- The table has no active mutations, patch parts, or lightweight deletes that would make stored statistics stale
- For GROUP BY queries, the GROUP BY keys must map to partition key columns
- If a filter is present, it must be computable from partition-level expressions and virtual columns

If any check fails the query silently falls back to a normal `ReadFromMergeTree` read.

## SQL Examples

```sql
-- Create a table with column statistics
CREATE TABLE t (
    key UInt64,
    value Int32
) ENGINE = MergeTree ORDER BY key
SETTINGS auto_statistics_types = 'basic';

-- Insert data (statistics are materialized on insert)
SET materialize_statistics_on_insert = 1;
INSERT INTO t SELECT number, toInt32(rand() % 1000) FROM numbers(10000000);

-- This query is answered from per-part statistics — no column data is read
SELECT min(value), max(value) FROM t;

-- GROUP BY partition key is also supported
CREATE TABLE t2 (
    p UInt8,
    v UInt32
) ENGINE = MergeTree PARTITION BY p ORDER BY tuple()
SETTINGS auto_statistics_types = 'basic';

INSERT INTO t2 SELECT number % 10, number FROM numbers(10000000);

SELECT p, min(v), max(v) FROM t2 GROUP BY p ORDER BY p;
```

The optimization can be disabled via `SET use_statistics_for_min_max_aggregation = 0`.

## Related PRs

Related: https://github.com/ClickHouse/ClickHouse/pull/93037

- Performance Improvement

Optimize `min()` and `max()` aggregate functions by utilizing per-part column statistics (`basic`) metadata when available, avoiding full data scans for queries. Controlled by the new setting `use_statistics_for_min_max_aggregation`.

---
Workflow [[PR](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105197&sha=latest&name_0=PR)]
Sync PR [[sync-upstream/pr/105197](https://github.com/search?q=head%3Async-upstream%2Fpr%2F105197+org%3AClickHouse+type%3Apr&type=pullrequests)]


tests/queries/0_stateless/04238_column_statistics_aggregation_minmax.sql
```diff
@@ -0,0 +1,106 @@
...
+    b Int32,
+    value Float64,
+    value_with_null Nullable(Float64),
+    value_all_null Nullable(Float64),
+    value_with_inf Float64,
+    p Int32
+) ENGINE = MergeTree
+PARTITION BY p
+ORDER BY id
+SETTINGS auto_statistics_types = 'minmax';
```
rschu1ze: `minmax` statistics will soon be deprecated, they are superseded by `basic` (--> https://github.com/ClickHouse/ClickHouse/pull/106048). Please switch this PR to assume only `basic`.

tests/queries/0_stateless/04238_column_statistics_aggregation_minmax.sql
```diff
@@ -0,0 +1,106 @@
...
+INSERT INTO test_col_stats_agg VALUES
+    (1, 10, 5, 100.0, 100.0, NULL, 100.0, 1),
+    (2, 20, 3, 200.0, 200.0, NULL, 200.0, 1),
+    (3, 5, 2, 50.0, 50.0, NULL, 50.0, 1),
+    (4, 30, 10, 300.0, 300.0, NULL, 300.0, 2),
+    (5, 15, 4, 150.0, 150.0, NULL, 150.0, 2),
+    (6, 2, 3, 25.0, NULL, NULL, 25.0, 3),
+    (7, 40, 10, 400.0, 400.0, NULL, inf, 3);
+
+SET optimize_use_projections = 1;
```
rschu1ze, line 33: I assume l. 29-32 are here to suppress setting randomization in CI. If so, please mention that in the code as a comment.

rschu1ze:
@zoomxi Please explain in the PR message what this optimization is doing (yes, I can and will read the code, but still). Please include a SQL example. Also, there should ideally be a xml performance test that demonstrates that performance improves.

zoomxi:
>Please explain in the PR message what this optimization is doing (yes, I can and will read the code, but still). Please include a SQL example. Also, there should ideally be a xml performance test that demonstrates that performance improves.

Sorry for the late reply, and thanks for the review! I'll make time to work on your suggestions soon.

zoomxi:
Added performance test `column_statistics_minmax_aggregation`.

| Query | Scenario | Optimization ON | Optimization OFF | Speedup |
  |---|---|---|---|---|
  | `SELECT min(v), max(v) FROM t` | 10M rows, no partition | 0.002s | 0.011s | 5.5x |
  | `SELECT p, min(v), max(v) FROM t GROUP BY p` | 10M rows, 20 partitions | 0.003s | 0.024s | 8.0x |

rschu1ze:
@zoomxi I'm looking at [above measurements](https://github.com/ClickHouse/ClickHouse/pull/105197#issuecomment-5057175612), the queries with the optimization turned off run so fast already that I think they are _already_ answered from metadata. My theory is that such queries are already answered from the per-part minmax statistics (which are different than the min-max skip index and the `basic` statistics which also store min-max information). If that is true, then the PR would replace one metadata lookup by another metadata lookup.

zoomxi:
> If that is true, then the PR would replace one metadata lookup by another metadata lookup.

@rschu1ze  I double-checked, and the baseline queries are *not* answered from metadata.

```
Query 1: SELECT min(value), max(value) FROM perf_stats_minmax_no_part

  Optimization OFF :
  └── ReadFromMergeTree (default.perf_stats_minmax_no_part)
        Read type: Default
        Parts: 9 | Granules: 1223
        Output: value

  Optimization ON:
  └── ReadFromPreparedSource (_column_statistics_aggregation)
```

```
  Query 2: SELECT p, min(v), max(v) FROM perf_stats_minmax_part GROUP BY p ORDER BY p

  Optimization OFF:
  └── ReadFromMergeTree (default.perf_stats_minmax_part)
        Read type: Default
        Parts: 80 | Granules: 1260
        Output: p, v

  Optimization ON:
  └── ReadFromPreparedSource (_column_statistics_aggregation)
```

rschu1ze:
Will check later, no time this week

zoomxi:
Clipboard_Screenshot_1785460222

rschu1ze:
The per-part min-max statistics exist unconditionally for all parts. `basic`-type column statistics do exist for all parts as well, but only because setting `auto_statistics_type` forces them. I think it is more natural and correct to answer min/max aggregation functions from per-part min-max statistics.

zoomxi:
> The per-part min-max statistics exist unconditionally for all parts. `basic`-type column statistics do exist for all parts as well, but only because setting `auto_statistics_type` forces them. I think it is more natural and correct to answer min/max aggregation functions from per-part min-max statistics.

@rschu1ze The per-part min-max index is already used for pruning (e.g. `minmax_idx_condition` in `MergeTreeDataSelectExecutor`), but it only covers the partition key columns. basic statistics are currently the only per-part min/max metadata that covers arbitrary columns.

zoomxi:
> The per-part min-max statistics exist unconditionally for all parts. `basic`-type column statistics do exist for all parts as well, but only because setting `auto_statistics_type` forces them. I think it is more natural and correct to answer min/max aggregation functions from per-part min-max statistics.

@rschu1ze True that the per-part min-max metadata exists unconditionally — but it currently includes only partition key columns, so for min()/max() on non-key columns, basic statistics is the only part-level min/max metadata that can cover them.

Given that #115738 pursues the same goal with exactly this layering, I'm closing this PR in favor of it.

PR #105198, opened 2026-05-18
DROP PARTITION for iceberg tables

### LINKED ISSUE #97929: Implement `ALTER TABLE ... DROP PARTITION ...` for Iceberg
There is a query which allows to drop the whole partition for MergeTree: https://clickhouse.com/docs/managing-data/drop_partition

The idea is to re-implement the same query for Iceberg tables. For Iceberg implementation should be even more simple than for MergeTree. It can be handled in `StorageObjectStorage::dropPartition` (which currently just throws), routing to a new `IDataLakeMetadata::dropPartition` method:
1. Read the current snapshot's manifest list
2. For each manifest, filter entries by partition value
3. Write a new manifest where matching entries have status=DELETED
4. Non-matching manifests are kept as-is (referenced, not rewritten)
5. Commit a new snapshot with the updated manifest list

The partition representation in query can be tricky, especially if it's complex expression. Also need to check what to use `ALTER TABLE ... DROP PARTITION 'partition'` or `ALTER TABLE ... DROP PARTITION ID partition_id`.

Diskein:
Closes: https://github.com/ClickHouse/ClickHouse/issues/97929

Implements `ALTER TABLE ... DROP PARTITION (...)` for Iceberg v2 tables, e.g.
`ALTER TABLE my_iceberg DROP PARTITION (2, '2')`. The partition expression accepts the
same literal/tuple/transform forms as the partition spec (e.g. `icebergBucket`, date transforms).

Behaviour:
- Drops the partition as of query start; rows inserted into the same partition while it runs are kept.
- Removes a fully-matched data manifest outright, and rewrites a partially-matched one to keep
  the survivors.
- Supports both catalog-backed (`DataLakeCatalog`) and path-based tables.

Limitations:
- Requires the experimental `allow_insert_into_iceberg` setting.
- Iceberg format-version 2 only.
- Tables with evolved partition specs, or where a surviving file was written under a different
  schema than the table's current one (schema evolution).
- An affected manifest containing equality deletes.
- `decimal`, `UUID`, and `boolean` partition columns.
- Throws a write conflict if a concurrent rewrite/compaction replaces the targeted files.

- New Feature

Support `ALTER TABLE ... DROP PARTITION (...)` on Iceberg v2 tables, e.g. `ALTER TABLE my_iceberg DROP PARTITION (2, '2')`.


src/Storages/ObjectStorage/DataLakes/Iceberg/AlterDropPartitionExecutor.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+    for (const auto & manifest_key : state.snapshot->manifest_list_entries)
+    {
+        TargetManifest target_manifest;
+        target_manifest.manifest_path = manifest_key.manifest_file_path;
+        target_manifest.manifest_content_type = manifest_key.content_type;
+
+        auto handle = getManifestFileEntriesHandle(object_storage, components, context, log, manifest_key, state.schema_id);
+        match_entries(handle.getFilesWithoutDeleted(FileContentType::DATA), targets.data, target_manifest);
+        match_entries(handle.getFilesWithoutDeleted(FileContentType::POSITION_DELETE), targets.position_delete, target_manifest);
```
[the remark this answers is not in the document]
Diskein: Not valid, we do not omit them.

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -874,3 +889,95 @@ void alter(
...
+
+    persistent_components.invalidateMetadataCache();
+    return {};
+}
+
+
+void IcebergMetadata::alterPartitionDropImpl(
+    const PartitionCommand & command, ContextPtr context, std::shared_ptr<DataLake::ICatalog> catalog, StorageID storage_id)
+{
+    auto fetch_latest_state = [this, context, catalog, storage_id]() -> std::pair<Iceberg::IcebergDataSnapshotPtr, Iceberg::TableStateSnapshot>
```
scanhex12: Why do we need this callback? I think it's better to move this logic somewhere in `AlterDropPartitionExecutor`

tests/queries/0_stateless/04402_iceberg_drop_partition_after_delete_files_size.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
```
scanhex12, line 1: IMHO it's better to write only integration tests for iceberg, but you can save as is if you wish

tests/integration/test_storage_iceberg_no_spark/test_drop_partition_catalog.py
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env python3
+
+import uuid
```
scanhex12: we have a separate directory for tests with catalog

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -874,3 +889,95 @@ void alter(
...
+    }
+    if (commands.size() != 1)
+        throw Exception(
+            ErrorCodes::BAD_ARGUMENTS,
+            "ALTER TABLE ... on Iceberg expects exactly one partition command, got {}",
+            commands.size());
+
+    const auto & command = commands.at(0);
+
+    switch (command.type)
```
scanhex12: `chassert(command.type == PartitionCommand::Type::DROP_PARTITION);`?

src/Storages/ObjectStorage/DataLakes/Iceberg/AlterDropPartitionExecutor.h
```diff
@@ -0,0 +1,178 @@
...
+    {
+        std::vector<TargetManifest> fully_matched;
+        std::vector<TargetManifest> partially_matched;
+
+        bool empty() const { return fully_matched.empty() && partially_matched.empty(); }
+    };
+
+    struct DropPlan
+    {
+        TargetManifests target_manifests;
```
scanhex12, line 88: I think it can be quite expensive to materialise all manifests in RAM. In reader, for instance, we have iterator to store exactly one manifest in memory. Can we process drop partition with something similar?

src/Storages/ObjectStorage/DataLakes/Iceberg/AlterDropPartitionExecutor.cpp
```diff
@@ -0,0 +1,941 @@
...
+                    ++removed_position_delete_files;
+                    removed_position_deletes += parsed_entry.record_count;
+                    removed_files_size += parsed_entry.file_size_in_bytes;
+                    break;
+                case FileContentType::EQUALITY_DELETE:
+                    /// Discovery never matches equality-delete entries, so we
+                    /// should never see one here. Treat as a hard error rather
+                    /// than silently miscount.
+                    throw Exception(
+                        ErrorCodes::LOGICAL_ERROR, "DROP PARTITION encountered an equality-delete entry, which is not supported");
```
scanhex12: not logical error, just bad arguments

src/Storages/ObjectStorage/DataLakes/Iceberg/AlterDropPartitionExecutor.cpp
```diff
@@ -0,0 +1,941 @@
...
+    SnapshotState state;
+
+    {
+        auto [snapshot, table_state] = fetch_latest_state();
+        if (!snapshot)
+            return std::nullopt;
+
+        /// FIXME: in all other places schema_id is int32
+        if (snapshot->schema_id_on_snapshot_commit > std::numeric_limits<Int32>::max())
+            throw Exception(ErrorCodes::LOGICAL_ERROR, "Iceberg schema_id {} exceeds Int32 range", snapshot->schema_id_on_snapshot_commit);
```
scanhex12: bad arguments too, because it's a problem of the data

tests/integration/test_storage_iceberg_with_spark/test_drop_partition.py
```diff
@@ -0,0 +1,256 @@
...
+        if m:
+            versions.append(int(m.group(1)))
+    assert versions, "ClickHouse should have written a new metadata.json"
+    with open(os.path.join(metadata_dir, "version-hint.text"), "wb") as f:
+        f.write(str(max(versions)).encode())
+
+    # Spark reads split_offsets straight out of the manifest CH rewrote. Spark always writes
+    # split_offsets (Parquet row-group offsets) and the old rewrite dropped them, so a non-empty
+    # value here proves the field survived the partial rewrite, as seen by the canonical reader.
+    spark_after = [
```
scanhex12, line 178: It would be fine if you add checks that `select * from table` are the same for ch and spark at the end of all such tests

tests/integration/test_storage_iceberg_with_spark/test_drop_partition.py
```diff
@@ -0,0 +1,256 @@
...
+    # Spark can still insert into the table after CH's DROP PARTITION.
+    spark_query(f"INSERT INTO {table_name} VALUES (4, 'e', 40)")
+    default_upload_directory(
+        started_cluster_iceberg_with_spark,
+        storage_type,
+        f"/iceberg_data/default/{table_name}/",
+        f"/iceberg_data/default/{table_name}/",
+    )
+    assert instance.query(f"SELECT count() FROM {table_name}") == "3\n"
+    assert (
```
scanhex12, line 283: could you please also add tests that trino can read your table after such operations?

PR #105243, opened 2026-05-18
Sanitize server-supplied display strings in client Hello

Algunenano:
`clickhouse-client` reads `server_name`, `server_timezone`, `server_display_name`, and the password-complexity rule patterns and messages from the server's `Hello` packet via raw `readStringBinary` and prints them back to the user — directly in the `Connected to ...` banner, indirectly through the `{display_name}` prompt substitution, the time-zone warning, and password-rule error messages. A hostile or compromised server could embed arbitrary bytes in those fields (terminal escape sequences, etc.) and have the connecting client render them. The default `readStringBinary` cap of 1 GiB is also an unbounded-allocation vector against the connecting client.

Add `sanitizeUntrustedServerString` (in `src/Client/sanitizeUntrustedServerString.h`) that replaces ASCII control bytes (`0x00-0x1F` and `0x7F` DEL) with `?` while preserving UTF-8 high bytes so non-ASCII display names render normally. Apply it after each `readStringBinary` for those fields in `Connection::receiveHello`, and cap the reads at 4 KiB — well above any legitimate hostname / time zone / password-rule message — to close the unbounded-allocation vector.

- Improvement

Sanitize server-supplied display strings in `clickhouse-client`'s `Hello` packet handling (server name, time zone, display name, password-rule patterns and messages) so a hostile server cannot inject bytes that the client renders verbatim, and cap their size to defend against unbounded allocation.

- [x] Documentation is unnecessary (defensive change with no user-visible API).

### Version info
- Merged into: `26.6.1.93`

### DIFF
--- a/src/Client/Connection.cpp
+++ b/src/Client/Connection.cpp
@@ -611,11 +611,12 @@ void Connection::receiveHello()
             UInt64 rules_size;
             readVarUInt(rules_size, *in);
             /// `rules_size` is server-controlled and feeds a `reserve`; reject absurd
-            /// values so a hostile server cannot force a huge allocation.
-            if (rules_size > MAX_PASSWORD_COMPLEXITY_RULES)
+            /// values so a hostile server cannot force a huge allocation. The server
+            /// enforces the same cap at construction time (see TCPHandler::sendHello).
+            if (rules_size > DBMS_MAX_PASSWORD_COMPLEXITY_RULES)
                 throw Exception(ErrorCodes::UNEXPECTED_PACKET_FROM_SERVER,
                     "Server declared {} password-complexity rules, maximum allowed is {}",
-                    rules_size, MAX_PASSWORD_COMPLEXITY_RULES);
+                    rules_size, DBMS_MAX_PASSWORD_COMPLEXITY_RULES);
             password_complexity_rules.reserve(rules_size);
 
             for (size_t i = 0; i < rules_size; ++i)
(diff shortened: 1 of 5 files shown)


src/Client/Connection.cpp
```diff
@@ -575,39 +576,56 @@ void Connection::receiveHello()
...
+            readStringBinary(proto_recv_chunked_srv, *in, MAX_SERVER_HELLO_STRING_SIZE);
         }
 
         if (server_revision >= DBMS_MIN_PROTOCOL_VERSION_WITH_PASSWORD_COMPLEXITY_RULES)
         {
             UInt64 rules_size;
             readVarUInt(rules_size, *in);
+            /// `rules_size` is server-controlled and feeds a `reserve`; reject absurd
+            /// values so a hostile server cannot force a huge allocation.
+            if (rules_size > MAX_PASSWORD_COMPLEXITY_RULES)
```
antonio2368: Should we have such limit on server side just in case?
So it's a bit clearer what the problem actually is. Even though the value is absurd.

Algunenano: Yes, it makes sense to put this limit in both so it's clear. I'll work on it

src/Server/TCPHandler.cpp
```diff
@@ -2113,6 +2113,15 @@ void TCPHandler::sendHello()
     {
         auto rules = server.context()->getAccessControl().getPasswordComplexityRules();
 
+        /// Mirror the client-side cap so a misconfiguration fails at the server's own
+        /// boundary with a message pointing at the config, rather than producing an
+        /// `UNEXPECTED_PACKET_FROM_SERVER` on the client.
+        if (rules.size() > DBMS_MAX_PASSWORD_COMPLEXITY_RULES)
```
antonio2368, line 2138: What about `MAX_SERVER_HELLO_STRING_SIZE` limit for the size of the rule?

antonio2368 (approved):

PR #105244, opened 2026-05-18
Fix Date32 overflow on toStartOfWeek

### LINKED ISSUE #104970: `ToStartOfWeekImpl::execute(Int32, ...)` missing `std::max(res, 0)` clamp — Date32 overflow persists

_Retrospective finding from a historical scan of [PR #67558](https://github.com/ClickHouse/ClickHouse/pull/67558) (merged 2024-08-02). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`toStartOfWeek(toDate32('1970-01-01'))` returns `2149-06-03` instead of `1970-01-01`

**Root cause:** DateTimeTransforms.h:420: `return static_cast<UInt16>(time_zone.toFirstDayNumOfWeek(ExtendedDayNum(d), week_mode));` — when toFirstDayNumOfWeek returns negative (week starts before epoch), the direct cast to UInt16 overflows

**Why we believe this is a bug:** PR fixed Int64 and UInt32 overloads at DateTimeTransforms.h:408-417 by adding std::max(res, 0), but missed the Int32 overload at line 418-421 used by Date32

**Affected locations:**
- `src/Functions/DateTimeTransforms.h:420` — ToStartOfWeekImpl::execute(Int32, UInt8, const DateLUTImpl&)

**Impact:** Any Date32 value where the week start falls before 1970-01-01 (e.g., early January 1970, any date in 1969 and earlier) produces wildly incorrect results. Users storing historical dates as Date32 get corrupt week aggregations.

Yes — confirmed on current `master` (commit `b6844b26db6`).

### How to reproduce

```sql
-- Test: toStartOfWeek with Date32 for dates where week start is before 1970-01-01
-- Expected: 1970-01-01 (clamped), Actual: 2149-06-03 (overflow)
SELECT 'Date32 test - should be 1970-01-01:';
SELECT toStartOfWeek(toDate32('1970-01-01')) AS result;
SELECT toStartOfWeek(toDate32('1970-01-02')) AS result;
SELECT toStartOfWeek(toDate32('1970-01-03')) AS result;

-- Compare with DateTime64 which was fixed by the PR
SELECT 'DateTime64 test (PR fix verified):';
SELECT toStartOfWeek(toDateTime64('1970-01-01', 6)) AS result;
SELECT toStartOfWeek(toDateTime64('1970-01-02', 6)) AS result;

-- Date32 for historical dates - week starts before epoch
SELECT 'Date32 historical dates test:';
SELECT toStartOfWeek(toDate32('1900-01-01')) AS result;
SELECT toStartOfWeek(toDate32('1969-12-28')) AS result;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/6e306b9b-e840-4f49-ba56-4d7186e680db)

### Expected behavior

```
Date32 test - should be 1970-01-01:
1970-01-01
1970-01-01
1970-01-01
DateTime64 test (PR fix verified):
1970-01-01
1970-01-01
Date32 historical dates test:
1970-01-01
1970-01-01
```

```
Date32 test - should be 1970-01-01:
2149-06-03
2149-06-03
2149-06-03
DateTime64 test (PR fix verified):
1970-01-01
1970-01-01
Date32 historical dates test:
2079-06-06
2149-06-03
```

### Additional context

**Open risks:**
- ToMondayImpl has the same overflow bug pattern at DateTimeTransforms.h:216-232 — all 4 overloads need the same fix
- ToStartOfYearImpl and ToLastDayOfWeekImpl should be audited for the same pattern

**Suggested fix:** Apply the same fix as the Int64/UInt32 overloads:
```cpp
static UInt16 execute(Int32 d, UInt8 week_mode, const DateLUTImpl & time_zone)
{
    const int res = time_zone.toFirstDayNumOfWeek(ExtendedDayNum(d), week_mode);
    return static_cast<UInt16>(std::max(res, 0));
}
```

Found during automated review of [PR #67558](https://github.com/ClickHouse/ClickHouse/pull/67558).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr67558_001`_

### Version info
- Resolved by: #105244
- Merged into: `26.6.1.590`
- Backported to: `26.5.3.18`, `26.4.5.25`

### LINKED ISSUE #106419: Wrong result: WHERE toStartOfYear(Date32) < const returns 0 rows after merge when column has pre-1970 values
## Describe what's wrong

A `WHERE` filter of the form `toStartOfYear(<Date32 column>) < toStartOfYear(<Date constant>)` returns **0 rows** when the `Date32` column contains a few pre-1970 values and the data lives in a **merge-formed part** (e.g. after `OPTIMIZE … FINAL`), even though the predicate is true for almost every row. The same predicate evaluated per-row via `countIf(...)` returns the correct count, so `WHERE` and `countIf` of the *identical* expression disagree.

The trigger is `toStartOfYear` (and `toStartOfMonth`, `toStartOfQuarter`) narrowing `Date32` → `Date`: for a pre-1970 `Date32` value the result overflows into a far-future `Date` (`toStartOfYear(toDate32('1905-01-01'))` = `2084-06-06`). This poisons the part's value range used by the monotonic-function filter optimization, so the whole part is wrongly skipped.

## How to reproduce

* ClickHouse version: **26.6.1.339** (clickhouse/clickhouse-server:head, pulled 2026-06-03)
* Reproduces with both `allow_experimental_analyzer=0` and `=1`.

```sql
CREATE TABLE issue_repro (c1 Date32) ENGINE = MergeTree ORDER BY tuple();

-- 9991 in-range rows (1971..2020) then 9 pre-1970 outliers, as two separate parts
INSERT INTO issue_repro SELECT toDate32('1971-01-01') + toIntervalDay(number % 18000) FROM numbers(9991);
INSERT INTO issue_repro SELECT toDate32('1905-01-01') + toIntervalDay(number * 30)    FROM numbers(9);

-- merge into a single part (also happens via background merge)
OPTIMIZE TABLE issue_repro FINAL;

-- the predicate is true for 9991 of the 10000 rows:
SELECT countIf(toStartOfYear(c1) < toStartOfYear(toDate('2021-06-15'))) FROM issue_repro;
-- 9991   (correct)

-- but as a WHERE filter it returns 0:
SELECT count() FROM issue_repro WHERE toStartOfYear(c1) < toStartOfYear(toDate('2021-06-15'));
-- 0      (WRONG, expected 9991)
```

## Expected behavior

Both queries return **9991**. `WHERE expr` must return the same rows that `countIf(expr)` counts.

## Additional context

* **Function scope.** Affected: `toStartOfYear`, `toStartOfMonth`, `toStartOfQuarter` (all return `Date` from a `Date32` argument). **Not** affected on the same table: `toYYYYMM`, `toYear` (numeric return types, no `Date` narrowing → no overflow) — both give `WHERE` = `countIf` = 10000.

* **The overflow.** `toStartOfYear(Date32)` returns `Date` (range 1970-01-01 .. 2149-06-06), so a pre-1970 `Date32` value wraps to a future `Date`:

```sql
SELECT toStartOfYear(toDate32('1905-01-01')) AS soy, toTypeName(soy);
-- 2084-06-06    Date
```

This is arguably a second bug on its own (the function silently returns a wrong value for in-`Date32`-range but out-of-`Date`-range inputs).

* **Why a merge is required.** A single `INSERT` that interleaves the outliers into one part does **not** reproduce (`WHERE` = `countIf` = 9991); the wrong result appears only once the outliers and the in-range rows share a **merge-formed** part. This, plus the fact that `countIf` (per-row evaluation) is correct while `WHERE` is wrong, points at the part-level value-range / monotonic-function filter analysis rather than per-row evaluation. `force_index_by_date=0`, `optimize_move_to_prewhere=0`, and `toStartOfYear(materialize(c1))` do **not** change the wrong result; only avoiding the filter form (using `countIf`) does.

Found by SQLancer (TLPGroupBy oracle: a `GROUP BY` partitioned by this predicate lost rows).

### Version info
- Resolved by: #105244
- Merged into: `26.6.1.590`
- Backported to: `26.5.3.18`, `26.4.5.25`

yariks5s:
Closes: https://github.com/ClickHouse/ClickHouse/issues/104970
Closes: https://github.com/ClickHouse/ClickHouse/issues/106419

Fix incorrect results of `toStartOfWeek`, `toLastDayOfWeek`, `toMonday`, `toStartOfMonth`, `toLastDayOfMonth`, `toStartOfQuarter` and `toStartOfYear` for `Date32` and `DateTime64` arguments whose result falls outside the `Date` range: instead of overflowing into arbitrary dates, results before `1970-01-01` are now clamped to `1970-01-01` and results after `2149-06-06` are clamped to `2149-06-06`. This also fixes wrong query results (incorrectly pruned parts and granules) when such functions were used in `WHERE` over a `Date32` or `DateTime64` key containing out-of-range values.

### Version info
- Merged into: `26.6.1.590`
- Backported to: `26.5.3.18`, `26.4.5.25`


antonio2368 (approved):
Probably need to regenerate some references from old tests.

fm4v:
fable 5 reviewed the PR and pushed two follow-up commits addressing the remaining gaps:

**1. `Round negative DateTime64 towards negative infinity in executeExtendedResult`**

`TransformDateTime64::executeExtendedResult` truncated negative fractional `DateTime64` values towards zero, while `execute` rounds towards negative infinity. Since this PR routes the monotonicity factor comparison through `executeExtendedResult`, a key in the last fractional second before a pre-epoch period boundary was attributed to the *next* period by the factor while actual execution attributes it to the *previous* one — a granule like `[1964-12-31 23:59:59.9, 1965-08-15]` would be wrongly declared monotonic for `toMonth` and could be pruned for `WHERE toMonth(d) = 12`. The same rounding bug was already user-visible in the extended-results path on master:

```sql
SET enable_extended_results_for_datetime_functions = 1;
SELECT toStartOfMonth(toDateTime64('1969-12-31 23:59:59.5', 1)); -- returned 1970-01-01, while toMonth() of the same value returns 12
```

**2. `Clamp the upper Date overflow of date rounding functions as well`**

`std::max(res, 0)` only fixes the pre-epoch wrap. A `Date32` above `2149-06-06` still wrapped through `UInt16` (`toStartOfYear(toDate32('2200-01-01'))` → `2020-07-27`), and the wrong-result mechanism of #106419 reproduces identically in the upper range:

```sql
CREATE TABLE repro2 (c1 Date32) ENGINE = MergeTree ORDER BY tuple();
INSERT INTO repro2 SELECT toDate32('2140-01-01') + toIntervalDay(number * 10) FROM numbers(5500);
OPTIMIZE TABLE repro2 FINAL;
SELECT countIf(toStartOfYear(c1) < toDate('1980-01-01')) FROM repro2; -- 365
SELECT count() FROM repro2 WHERE toStartOfYear(c1) < toDate('1980-01-01'); -- 0 (wrong)
```

Now all sites use `std::clamp(res, 0, DATE_LUT_MAX_DAY_NUM)`, saturating to `2149-06-06` (consistent with `ToDateImpl` saturation), which keeps the functions genuinely monotonic over the whole `Date32` domain. The remaining wrapped values in `01921_datatype_date32.reference` (`2120-…` for the `2299-…` inputs) are updated accordingly, and both fixes come with regression tests (`04306_date32_upper_overflow`, `04307_datetime64_pre_epoch_subsecond_monotonicity`).

I also updated the changelog entry in the description to cover the full set of affected functions and the pruning aspect.

fm4v:
Status: everything is green except `CH Inc sync`, which is red for an external reason: the private unit test `MetadataInKeeperTest.TestConcurrentUpdatesFromHints` is failing near-deterministically on private master since 2026-06-09 ~19:00 UTC (66 failures across 32 unrelated PRs on the first day, including master runs). Three attempts on the sync PR all failed on this single test; nothing else fails. Tracked in ClickHouse/clickhouse-private#60290 — once that test is fixed on private master, a rerun of the sync job will re-merge master and the check will go green.

PR #105247, opened 2026-05-18
Reset supplementary groups in clickhouse su before dropping privileges

Algunenano:
`programs/su/su.cpp` resolves the target user and group, then calls `setgid` followed by `setuid`. Neither syscall touches the supplementary group list, so the dropped process silently inherits the supplementary groups of the invoker. On a system where root has been granted membership in groups like `disk`, `adm`, `wheel`, `docker`, `sudo`, or where the systemd unit / Docker entrypoint configures supplementary groups, those memberships leak into the unprivileged process and defeat the intent of the privilege drop. For comparison, `/bin/su` and `sudo` both reset the supplementary list via `initgroups` as part of dropping privileges.

`setUserAndGroup` is restructured so the passwd database is consulted even when the UID is supplied numerically (to recover the user name), and `initgroups(user_name, group)` is invoked before `setgid`/`setuid` so the supplementary group list reflects the target user's `/etc/group` memberships. When the target UID has no passwd entry the supplementary list is cleared with `setgroups(0, nullptr)` instead.

- Improvement

Reset the supplementary group list in `clickhouse su` before dropping privileges, matching the behavior of `/bin/su` and `sudo`. Previously the dropped process inherited the invoker's supplementary groups.

### Version info
- Merged into: `26.6.1.161`

### DIFF
--- a/programs/su/su.cpp
+++ b/programs/su/su.cpp
@@ -48,11 +48,14 @@ static void setUserAndGroup(std::string arg_uid, std::string arg_gid)
     static constexpr size_t buf_size = 16384; /// Linux man page says it is enough. Nevertheless, we will check if it's not enough and throw.
     std::unique_ptr<char[]> buf(new char[buf_size]);
 
-    /// Set the group first, because if we set user, the privileges will be already dropped and we will not be able to set the group later.
+    /// Resolve the target group GID first, while we still have privileges.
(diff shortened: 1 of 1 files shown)


santrancisco (approved):
LGTM! thanks for the fix!

PR #99386, opened 2026-03-13
Fix: filesystem cache disk IO error must not mark MergeTree parts as broken

tiandiwonder:
## Problem

When starting a `MergeTree` table backed by S3 with a local filesystem cache, a disk IO error during cache write causes the part to be incorrectly marked as broken — even though the data in S3 is intact. ref #83823 

**Root cause**: `loadDataPart()` calls `isRetryableException()` to decide whether to `mark_broken` or rethrow. Cache write IO errors were not recognized as retryable, so they fell into the `mark_broken` branch instead of propagating.

**Core call chain**:
```
loadDataPartWithRetries()
  └─ loadDataPart()
       catch (...) {
         if (isRetryableException()) throw;   // rethrow → no broken marking
         mark_broken();  ← BUG: cache IO error ends up here
       }
       └─ loadColumnsChecksumsIndexes()
            └─ CachedOnDiskReadBufferFromFile::writeCache()
                 └─ FileSegment::write() → local cache disk write → EIO
```

## Fix

- **New error code** `CACHE_CANNOT_WRITE_TO_CACHE_DISK`: thrown when local cache disk write fails
- **`isRetryableException()`** now recognizes `CACHE_CANNOT_WRITE_TO_CACHE_DISK` → rethrows instead of marking broken → propagates to startup failure with a clear error message
- **New setting** `skip_cache_on_disk_failure` (default `false`) in cache disk config: when `true`, cache write failures are silently bypassed and the read falls back to S3 — suitable for degraded-mode operation
- **Failpoint** `cache_filesystem_failure` added for testing

## Behavior

| Scenario | `skip_cache_on_disk_failure=false` (default) | `skip_cache_on_disk_failure=true` |
|---|---|---|
| Cache disk IO failure | Exception propagates → server refuses to start with clear error | Silently bypassed, reads from S3, logs ERROR |
| Part marked as broken? | **No** (core fix) | **No** |

## Integration Tests

4 scenarios in `tests/integration/test_cache_bypass_on_disk_failure/`:

- `test_skip_cache_on_disk_failure_select`: `skip=true`, failpoint during SELECT → succeeds, no broken parts
- `test_noskip_cache_on_disk_failure_select`: `skip=false`, failpoint during SELECT → raises `CACHE_CANNOT_WRITE_TO_CACHE_DISK`
- `test_skip_cache_on_disk_failure_attach`: `skip=true`, failpoint during `ATTACH TABLE` → part loads successfully, `system.detached_parts` empty
[...]

PR #105250, opened 2026-05-18
FileCache won't throw on create directories failures

Diskein:
`DB::KeyMetadata::createBaseDirectory(bool)` fails on some filesystem errors and on some errors the exception is propogated to the caller, so it can fail executeQuery and etc, even if `skip_cache_on_disk_failure=1`.

This PR removes this errors nitpicking and won't throw this errors from FileCache, it just writes logs for such errors.

Query can be executed successfully even If FileCache fails to create some directory on disk for caching.

### Version info
- Merged into: `26.6.1.395`

### DIFF
--- a/src/Interpreters/FileCache/FileCache.cpp
+++ b/src/Interpreters/FileCache/FileCache.cpp
@@ -1278,7 +1278,7 @@ bool FileCache::doTryReserve(
 
     if (auto ec = file_segment.getKeyMetadata()->createBaseDirectory(); ec)
     {
-        failure_reason = ec.message();
+        failure_reason = "Failed to create base directory for key, error: " + ec.message();
         return false;
     }
 
--- a/src/Interpreters/FileCache/FileCache.cpp
+++ b/src/Interpreters/FileCache/FileCache.cpp
@@ -1278,14 +1278,8 @@ bool FileCache::doTryReserve(
 
     if (auto ec = file_segment.getKeyMetadata()->createBaseDirectory(); ec)
     {
-        if (skip_cache_on_disk_failure)
-        {
-            failure_reason = ec.message();
-            return false;
-        }
-
-        throw std::filesystem::filesystem_error(
-            fmt::format("failed to create base directory for a key {}", file_segment.getKeyMetadata()->getPath()), ec);
+        failure_reason = ec.message();
+        return false;
     }
 
     return true;
--- a/src/Interpreters/FileCache/FileCache.cpp
+++ b/src/Interpreters/FileCache/FileCache.cpp
@@ -1276,10 +1276,16 @@ bool FileCache::doTryReserve(
     file_segment.reserved_size += size;
     chassert(file_segment.reserved_size == main_priority_iterator->getEntry()->size);
 
-    if (!file_segment.getKeyMetadata()->createBaseDirectory())
+    if (auto ec = file_segment.getKeyMetadata()->createBaseDirectory(); ec)
     {
-        failure_reason = "not enough space on device";
-        return false;
+        if (skip_cache_on_disk_failure)
+        {
+            failure_reason = ec.message();
+            return false;
+        }
+
+        throw std::filesystem::filesystem_error(
+            fmt::format("failed to create base directory for a key {}", file_segment.getKeyMetadata()->getPath()), ec);
     }
 
     return true;
--- a/src/Interpreters/FileCache/Metadata.cpp
+++ b/src/Interpreters/FileCache/Metadata.cpp
@@ -148,10 +148,9 @@ std::error_code KeyMetadata::createBaseDirectory()
         created_base_directory.store(true);
         ProfileEvents::increment(ProfileEvents::FilesystemCacheCreatedKeyDirectories);
     }
-    else
-    {
+    else if (ec != std::errc::no_space_on_device && ec != std::errc::too_many_files_open)
         LOG_TRACE(cache_metadata->log, "Failed to create base directory for key {}, {}", key, ec.message());
-    }
+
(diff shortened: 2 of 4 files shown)


src/Interpreters/FileCache/Metadata.cpp
```diff
@@ -129,42 +129,30 @@ KeyMetadata::KeyState KeyMetadata::getState()
...
+    else
     {
-        created_base_directory = false;
-
-        if (!throw_if_failed &&
-            (e.code() == std::errc::no_space_on_device
-                || e.code() == std::errc::read_only_file_system
-                || e.code() == std::errc::permission_denied
-                || e.code() == std::errc::too_many_files_open
-                || e.code() == std::errc::operation_not_permitted))
```
kssenii, line 158: Was this list of error codes not enough? With which error did you observe the failure? (Since no-throw mode is the default with these error codes unless "temporary data in cache" was used)

kssenii: Ok, I see in the linked issue it is `std::errc::io_error`

kssenii: This would not be a compatible change - previously with `skip_cache_on_disk_failure` (by default `false`) - we will not throw with the above error codes and after this change we will. I'd suggest to do the following:
1. if `throw_if_failed` -- throw immediately
2. if `!throw_if_failed` -- just log a warning for error codes which are not related to full disk usage (no_space_on_device, too_many_files_open) and not throw regardless of the error code.

Since `throw_if_failed` is already `false` in all cases apart from "temporary data in cache", then this exception processing inside catch is the only this place needs to be changed.

or if we want to make this place consistent with this new setting from https://github.com/ClickHouse/ClickHouse/pull/99386, then we can just decide on `std::errc::io_error` according to that setting, but this is not necessary in my opinion.

Diskein: I've updated the PR. There is only two callers and both of them handle errors differently, so i would suggest just let them decide how to proceed with errors they got.

So behavior is the same as you suggested. 

> just log a warning for error codes which are not related to full disk usage (no_space_on_device, too_many_files_open) and not throw regardless of the error code.

AFAIU we want to filter such events because they would spam a lot?

> if we want to make this place consistent with this new setting from #99386 then we can just decide on std::errc::io_error according to that setting, but this is not necessary in my opinion.

We could make those changes backward compatible by just moving the list of errors after this check (into this if https://github.com/ClickHouse/ClickHouse/pull/105250/changes#diff-58a4224cd0e4335b1fd505d64b46e4c9e0f9a431d931516e7441d4c2a15c465aR1270, and it would be consistent - check the setting first, and if it's false, check if an error is need to be thrown. In my mind we would just enable `skip_cache_on_disk_failure` so we didn't have to decide on std::errc::io_error or change those list of ignored errors.

BTW Maybe it would worth to report profile events or something like that on errors so we could create alerts for them. Not sure, maybe they do exist 🤷‍♂️

kssenii: > AFAIU we want to filter such events because they would spam a lot?

Because they are kind of normal/expected in certain cases. But other errors are considered abnormal, meaning that something in the system is really broken/overloaded and it needs urgent attention. 

In general I'd prefer to check for `std::errc::io_error` or any other errors explicitly instead of catching all possible exceptions, because I do not want potential bugs to be hidden. For example, `no_such_file_or_directory` there would certainly indicate a bug, not a system error.

> BTW Maybe it would worth to report profile events or something like that on errors so we could create alerts for them. Not sure, maybe they do exist

As I remember there are no specific fs error profile events for the filesystem cache. Would be indeed good to add.

src/Interpreters/FileCache/Metadata.cpp
```diff
@@ -129,42 +129,29 @@ KeyMetadata::KeyState KeyMetadata::getState()
...
 
     std::shared_lock lock(cache_metadata->key_prefix_directory_mutex);
 
     if (created_base_directory.load(std::memory_order_relaxed))
-        return true;
+        return {};
 
-    try
+    std::error_code ec;
+    fs::create_directories(getPath(), ec);
```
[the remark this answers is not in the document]
Diskein: it isn't a soft-fail fail, caller should handle the error due to [[nodiscard]]

src/Interpreters/FileCache/Metadata.cpp
Diskein: @kssenii could you please review the current change, it behaves as you suggested here https://github.com/ClickHouse/ClickHouse/pull/105250/changes#r3303534474, except for it is easy to report correct failed reason via std::error_codes.

src/Interpreters/FileCache/FileCache.cpp
```diff
@@ -1267,9 +1267,9 @@ bool FileCache::doTryReserve(
     file_segment.reserved_size += size;
     chassert(file_segment.reserved_size == main_priority_iterator->getEntry()->size);
 
-    if (!file_segment.getKeyMetadata()->createBaseDirectory())
+    if (auto ec = file_segment.getKeyMetadata()->createBaseDirectory(); ec)
     {
-        failure_reason = "not enough space on device";
+        failure_reason = ec.message();
```
kssenii:
```suggestion
        failure_reason = "Failed to create base directory for key, error: " + ec.message();
```

kssenii (approved):

Diskein:
Fuzzer's finding is unrelated and already has an issue - https://github.com/ClickHouse/ClickHouse/issues/106095

PR #105253, opened 2026-05-18
Fix wrong result in direct join with key value storages

vdimir:

* Fixed `JOIN` against `EmbeddedRocksDB`, `Redis`, and `KeeperMap` tables that could silently return zero rows when `DirectKeyValueJoin` is used with `String` or `Nullable(String)` keys.


alexey-milovidov:
It's time to resurrect this.

alexey-milovidov:
Resurrected: merged `master` (the branch was ~20.5k commits behind) and addressed the review.

**Review feedback** — `04235_direct_join_keeper_map` now asserts `Algorithm: DirectKeyValueJoin` for the plain **and** the `Nullable` left key separately, instead of only `EXPLAIN`-ing `t_str`. Both matter: the plain path was broken too, because `getByKeys` never `base64Encode`d the serialized key although the sink does and `getBySerializedKeys` `base64Decode`s it.

**The red CI was real, not stale.** Every stateless check failed with one error, and the stack trace points straight at the path this PR adds coverage for:

```
DirectJoin.cpp:159 -> StorageKeeperMap::getByKeys -> getBySerializedKeys -> tryGet
  -> ZooKeeper::pushRequest -> LOGICAL_ERROR: Current component is empty
```

`getByKeys` was the only entry point reaching `getBySerializedKeys` without a `Coordination::setCurrentComponent` guard (`read`, `write`, `mutate` and `StorageKeeperMapSource::generate` all have one). Nothing exercised it before because no test did a direct join against `KeeperMap` — the new `04235` does. Guard added.

Two more commits were needed to build and pass against current `master`:
- `getInMemoryMetadataPtr` now returns the lifetime-safe `StorageMetadataHandle`, which deletes `operator->` on a temporary, so the snapshot has to be bound to a local first. The textual merge was clean; only the build caught this.
- `trim(explain)` no longer strips the new logical-join-step tree prefix, so the line reads `│  Algorithm: DirectKeyValueJoin`. Both tests now use `extract(explain, 'Algorithm: \\w+')`, which is robust to how the plan is rendered.

Verified locally on a Release build with `enforce_keeper_component_tracking = 1` (the setting that turns the missing guard into an abort):
- 29 stateless tests pass — the 2 tests from this PR, 13 direct-join/key-value regressions (`04201_*`, `03546_join_key_value_storage_with_casted_key`, `02242_join_rocksdb`, `02815_join_algorithm_setting`, `04627_direct_join_dictionary_nullable_key`, ...) and 14 `KeeperMap`/`EmbeddedRocksDB` tests, since `getBySerializedKeys` changed its `null_map` contract.
- `test_storage_redis/test.py::test_direct_join_nullable_left` passes.
- A/B check on the guard: removing just that line and rebuilding makes `04235` fail with exactly the CI error above; restoring it makes it pass.

vdimir (commented):
Ship it 👍

PR #105265, opened 2026-05-18
RFC: Add a server setting to allow any allocation to throw MEMORY_LIMIT_EXCEEDED

azat:
- Improvement

Add a server setting (`min_allocation_size_to_throw_on_memory_limit`) to allow any allocation to throw `MEMORY_LIMIT_EXCEEDED`, this should help with preventing OOMs

Right now, any `new` cannot throw, this is due exception safety issues, and only explicitly marked can throw `MEMORY_LIMIT_EXCEEDED`.

But this can lead to real OOMs, i.e. when kernel will kill clickhouse-server, even though we are trying hard to track memory properly.

Add a server setting `min_allocation_size_to_throw_on_memory_limit` (UInt64 bytes, default 0). When non-zero, the throwing `operator new` variants are allowed to raise `MEMORY_LIMIT_EXCEEDED`. _The assumption is that code with big enough allocations does not exception safety issues._

Note, places that allows (Allocator)/never allows (malloc) `MEMORY_LIMIT_EXCEEDED` before are not affected.

### Version info
- Merged into: `26.6.1.204`


src/Common/CurrentMemoryTracker.cpp
```diff
@@ -99,12 +107,30 @@ void CurrentMemoryTracker::check()
...
 AllocationTrace CurrentMemoryTracker::allocNoThrow(Int64 size)
 {
-    return allocImpl(size, /*throw_if_memory_exceeded=*/ false);
+    return allocImpl(size, /*enforce_memory_limit=*/ false);
+}
+
+AllocationTrace CurrentMemoryTracker::allocThrow(Int64 size)
+{
+    const bool enforce_memory_limit = static_cast<UInt64>(size) >= min_allocation_size_to_throw_on_memory_limit.load(std::memory_order_relaxed);
+    return allocImpl(size, enforce_memory_limit);
```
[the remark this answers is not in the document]
azat: Will add a note that it is recommended to set value of new setting >= max_untracked_memory

nickitat (commented):
Since we assume it will only be used in non-default configurations (CI or debugging OOM kills), should we really bother with this minimal limit? It seems like in such circumstances the more often it could be triggered the better.

azat:
I would like to have an option to increase it, since it is unclear which value is generally "safe" (only one thing is clear it should be bigger then max_untracked_memory to avoid mixing)

src/Common/CurrentMemoryTracker.cpp
```diff
@@ -100,12 +108,31 @@ void CurrentMemoryTracker::check()
...
+}
+
+void CurrentMemoryTracker::setMinAllocationSizeBytesToThrow(UInt64 value)
+{
+    min_allocation_size_to_throw_on_memory_limit.store(
+        value == 0 ? std::numeric_limits<UInt64>::max() : value,
```
nickitat, line 128: I think we uniformly use the convention that 0 means unset and 

``` cpp
if (threshold and value ? threshold)
```

instead. Any reason to do otherwise here?

azat: For `min_allocation_size_to_throw_on_memory_limit` itself `0` means unlimited, but to avoid extra comparison for 0 in hot path it is internally converted to UINT64_MAX

nickitat (approved):

azat:
>[Stress test (arm_tsan)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105265&sha=917c934c75a3a327c38a97cda2459994e77b097e&name_0=PR&name_1=Stress%20test%20%28arm_tsan%29)

- https://github.com/ClickHouse/ClickHouse/pull/106000

PR #105266, opened 2026-05-18
Throw INCORRECT_DATA instead of LOGICAL_ERROR on malformed Hudi data file names

groeneai:
`HudiMetadata::getDataFilesImpl` lists every file with the table's
format suffix (`.parquet`/`.orc`) under the table directory and
requires each file's stem to split into exactly three
underscore-separated parts:

```
[FileId]_[FileWriteToken]_[Timestamp].[extension]
```

If a file in the directory does not match, the parser raised
`LOGICAL_ERROR`. The file name comes from user-controlled storage
contents — a user can drop any `.parquet` file under a Hudi table path,
and the server fuzzer regularly does. `LOGICAL_ERROR` is reserved for
internal invariants and is fatal in debug/sanitizer builds (via
`chassert` / `std::terminate`), which is what the master server fuzzer
hit:

* STID `3160-3c8c` — `Stress test (experimental, serverfuzz, arm_debug)`
* Fuzzer fed an Iceberg-style file name
  (`00000-0-<uuid>-0-00001.parquet`) into a Hudi-typed directory, the
  stem failed to split into 3 parts, the server aborted.
* No prior GitHub issue or fix PR for this `STID`.

Raise `INCORRECT_DATA` with a message that explains the expected
naming convention. The server stays alive and the caller gets a
regular query-level exception.

### Reproduction (verified locally)

* Without the fix: server aborts with
  `<Fatal> : Logical error: 'Unexpected format for file: ...'`,
  connection drops, server dies.
* With the fix: query returns
  `Code: 117. DB::Exception: Unexpected format for file: ... . Hudi data files must follow the naming convention [FileId]_[FileWriteToken]_[Timestamp].[extension]. (INCORRECT_DATA)`,
  server stays alive.
* Stress: 5/5 invocations against a freshly written malformed file
  returned `INCORRECT_DATA`, the server stayed up after each one.

### Test

Added `tests/queries/0_stateless/04250_hudi_invalid_filename_format.sh`
which uploads a parquet file with a non-Hudi name and asserts the
query returns `INCORRECT_DATA` (not `LOGICAL_ERROR`).

### CI report

Master fuzzer report for STID `3160-3c8c`:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=6b26862efac6d92adb3a0fa00f1682150bc8e2e2&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_debug%29

### Notes for review

* The other datalake metadata parsers (`DeltaLakeMetadata.cpp`,
  several `Iceberg/*.cpp`) also raise `LOGICAL_ERROR` from
  user-controlled JSON parse paths. Those are separate bugs in
  different code paths and are not addressed here; happy to file a
  follow-up if reviewers want me to widen the scope.
* `parse<UInt64>(file_parts[2])` on the following line already raises
  a normal exception (`CANNOT_PARSE_INPUT_ASSERTION_FAILED`) on a
  non-numeric timestamp — verified locally — so no change is needed
  there.

`Hudi` table engine now raises `INCORRECT_DATA` (a regular query-level exception) instead of `LOGICAL_ERROR` when a parquet file in the table directory does not match the Hudi `[FileId]_[FileWriteToken]_[Timestamp].[extension]` naming convention. Previously such file names aborted the server in debug builds.

### Version info
- Merged into: `26.5.1.819`


groeneai:
cc @alesapin @kssenii — could you review this? It changes `HudiMetadata::getDataFilesImpl` to throw `INCORRECT_DATA` instead of `LOGICAL_ERROR` when a parquet file in the table directory does not match the Hudi `[FileId]_[FileWriteToken]_[Timestamp]` naming convention. The current `LOGICAL_ERROR` aborts the server in debug/sanitizer builds — caught by the master serverfuzz job (STID 3160-3c8c).

groeneai:
Pre-PR validation gate (per TASK.md Phase 4 §9):

**(a) Deterministic repro?** Yes. Upload a `.parquet` whose stem does not split into exactly 3 underscore-separated parts (the fuzzer's `00000-0-<uuid>-0-00001.parquet` works; so does any 2-part or 4+-part stem). Then `SELECT * FROM hudi(...)`. Triggers 100% of the time.

**(b) Root cause explained?** `HudiMetadata::getDataFilesImpl` calls `listFiles` which returns every file ending in the Hudi table format suffix. The file names are user-controlled (anything dropped under the bucket prefix), but the parser treats a format mismatch as an internal invariant violation (`LOGICAL_ERROR`). `LOGICAL_ERROR` aborts via `chassert` / `std::terminate` in debug+sanitizer builds, which is the serverfuzz failure mode.

**(c) Fix matches root cause?** Yes — switches the error code to `INCORRECT_DATA` (data on disk doesn't match the expected format), which is the canonical user-facing error for malformed external data. No defensive wrapping, no skipping bad files (strict behavior is preserved), no log-and-continue. Only the error category changes.

**(d) Test intent preserved / new test added?** The function's strict behavior is unchanged — it still refuses to read a directory with files that don't match the Hudi naming convention. Added `tests/queries/0_stateless/04250_hudi_invalid_filename_format.sh` that uploads a malformed file and asserts the query returns `INCORRECT_DATA`.

**(e) Demonstrated in both directions?** Verified locally:
- Without the fix (master `HEAD` rebuilt with no other changes): server logs `<Fatal> : Logical error: 'Unexpected format for file: ...'`, drops the connection, dies. Stack trace originates at `HudiMetadata.cpp:64`.
- With the fix: query returns `Code: 117. ... (INCORRECT_DATA)`. `SELECT 'alive'` on the same server succeeds afterwards. 5/5 stress runs returned `INCORRECT_DATA` and the server stayed up.

**(f) Fix is general, not a narrow patch?** This patches the one filename-format check in Hudi. Sibling datalake parsers (`DeltaLakeMetadata.cpp`, several `Iceberg/*.cpp`) also use `LOGICAL_ERROR` from user-controlled JSON parse paths, but those are different code paths (structured metadata files, not directory listings) and would benefit from a separate, broader audit. Happy to file a follow-up if reviewers want me to widen the scope here.

Session: `cron:clickhouse-ci-task-worker:20260518-183000`

Algunenano (approved):

PR #105277, opened 2026-05-18
Document preprocessor inconsistency for text-search functions

### LINKED ISSUE #105231: hasPhrase inconsistent results when using processor on text index
### Company or project name

_No response_

Using hasPhrase with/without text index returns different results

26.5.1.680

### How to reproduce

If I have a table with text index that uses a preprocessor to lower case the column:

```sql
CREATE TABLE tab
(
    `id` UInt32,
    `message` String,
    INDEX idx (message) TYPE text(tokenizer = splitByNonAlpha, preprocessor = lowerUTF8(message)) GRANULARITY 100000000
)
ENGINE = MergeTree
ORDER BY id
SETTINGS index_granularity = 1

INSERT INTO tab VALUES (1, 'hello world'), (2, 'Hello World'), (3, 'HELLO WORLD');
```

Querying with and without index with hasPhrase gives different results:

```sql
SELECT id
FROM tab
WHERE hasPhrase(message, 'hello world')
SETTINGS use_skip_indexes = 1

   ┌─id─┐
1. │  1 │
2. │  2 │
3. │  3 │
   └────┘
```

```sql
SELECT id
FROM tab
WHERE hasPhrase(message, 'hello world')
SETTINGS use_skip_indexes = 0

Query id: 65f6607e-1380-477f-b4e7-ce1906b8c3b1

   ┌─id─┐
1. │  1 │
   └────┘
```

AFAIK, the text index version uses the preprocessor to make it a case insensitive query vs a case sensitive one when it's doing a scan.

Is that expected?

### Expected behavior

_No response_

_No response_

### Additional context

_No response_

### Version info
- Resolved by: #105277
- Merged into: `26.5.1.831`

Blargian:
The functions `hasPhrase`, `hasAnyTokens`, and `hasAllTokens` apply the text index's preprocessor (e.g. `lowerUTF8`) only on the index path. As a result, the same query can return different rows depending on whether the text index is used (i.e. `SETTINGS use_skip_indexes = 0` vs `1`).

This is expected behavior — the inconsistency is tolerated to improve the usability of full-text search — but it was not documented. Add a note to each function description, mirror the existing `hasAllTokens` performance/text-index note onto `hasPhrase`, and extend the text indexes page to mention `hasPhrase` alongside the other functions that use the preprocessor.

Closes #105231

Documented that `hasPhrase`, `hasAnyTokens`, and `hasAllTokens` can return different rows when run with vs without the text index, because the text index's preprocessor is only applied on the index path.

### Version info
- Merged into: `26.5.1.831`

### DIFF
--- a/src/Functions/hasPhrase.cpp
+++ b/src/Functions/hasPhrase.cpp
@@ -228,7 +228,7 @@ ExecutableFunctionHasPhrase::executeImpl(const ColumnsWithTypeAndName & argument
 REGISTER_FUNCTION(HasPhrase)
 {
     FunctionDocumentation::Description description = R"(
-Checks if the `input` contains all tokens from the `phrase` in consecutive order. If `input` is a column, returns all rows that satisfy this condition.
+Checks if the `input` contains all tokens from the `phrase` in consecutive order.
 
 :::note
 Column `input` should have a [text index](../../engines/table-engines/mergetree-family/textindexes) defined for optimal performance.
--- a/src/Functions/hasPhrase.cpp
+++ b/src/Functions/hasPhrase.cpp
@@ -228,11 +228,22 @@ ExecutableFunctionHasPhrase::executeImpl(const ColumnsWithTypeAndName & argument
 REGISTER_FUNCTION(HasPhrase)
 {
     FunctionDocumentation::Description description = R"(
-Checks if the haystack contains all tokens from the phrase in consecutive order.
+Checks if the `input` contains all tokens from the `phrase` in consecutive order. If `input` is a column, returns all rows that satisfy this condition.
 
-Prior to searching, the function tokenizes both the `input` and the `phrase` arguments using the tokenizer specified as the optional third argument.
+:::note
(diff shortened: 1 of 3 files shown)


src/Functions/hasPhrase.cpp
```diff
@@ -228,11 +228,22 @@ ExecutableFunctionHasPhrase::executeImpl(const ColumnsWithTypeAndName & argument
 REGISTER_FUNCTION(HasPhrase)
 {
     FunctionDocumentation::Description description = R"(
-Checks if the haystack contains all tokens from the phrase in consecutive order.
+Checks if the `input` contains all tokens from the `phrase` in consecutive order. If `input` is a column, returns all rows that satisfy this condition.
```
rschu1ze: I think the second (new) sentence is confusing. The function returns 0 or 1 but the sentence sounds like it returns actual row values.

rschu1ze (approved):

src/Functions/hasPhrase.cpp
Blargian:
```suggestion
Checks if the `input` contains all tokens from the `phrase` in consecutive order.
```

PR #105278, opened 2026-05-18
Fix LEFT ANTI JOIN returning left key value in right key column

### LINKED ISSUE #105738: LEFT ANTI JOIN: ON-equality projection collapses to default-0 when a WHERE is added (regression from #101653)
## Describe the bug

`LEFT ANTI JOIN` projects the right-side column from the `ON`-equality (e.g. `right_0.c1` where the condition is `m.c2 = right_0.c1`) back as the matched value. Adding a `WHERE` predicate on the left table collapses that projection to the column's default value (0) instead of the equality-projected value.

The baseline query (no `WHERE`) returns the correct result; only adding `WHERE m.c0` flips the projection.

## How to reproduce

```sql
DROP TABLE IF EXISTS m;
CREATE TABLE m (c0 Int32, c1 UInt64, c2 UInt64) ENGINE = MergeTree() ORDER BY (-c0);
INSERT INTO m VALUES (1, 100, 999);
INSERT INTO m VALUES (2, 200, 888);
OPTIMIZE TABLE m FINAL;

-- Baseline: correct, returns 888 / 999
SELECT right_0.c1
FROM m LEFT ANTI JOIN m AS right_0 ON m.c2 = right_0.c1
ORDER BY 1;

-- With WHERE: returns 0 / 0  -- BUG
SELECT right_0.c1
FROM m LEFT ANTI JOIN m AS right_0 ON m.c2 = right_0.c1
WHERE m.c0
ORDER BY 1;
```

## Expected behavior

Both queries should return the same projected values:

```
888
999
```

## Actual behavior

The second query returns:

```
0
0
```

## Bisect

Bisected via `utils/auto-bisect/bisect.sh` on first-parent master commits. The regression was introduced by #101653 ("Better default selectivities for unknown and LIKE conditions").

Confirmed by running the reproducer against the corresponding pre-built CI binaries:

| Commit | Subject | Baseline | With `WHERE m.c0` |
|---|---|---|---|
| `9f58d642` (last good before the suspect window) | sync of #101161 | `888, 999` | `888, 999` |
| `7728965d` (first bad, sync-merge of #101653) | "Better default selectivities for unknown and LIKE conditions" | `888, 999` | `0, 0` |

Files touched by the suspect PR:

- `src/Storages/Statistics/ConditionSelectivityEstimator.cpp`
- `src/Storages/Statistics/ConditionSelectivityEstimator.h`

The bug is plausibly an existing latent issue with the `LEFT ANTI JOIN` right-side projection that becomes reachable only when the optimizer picks a different plan due to changed selectivity for `WHERE m.c0`. The baseline query is unaffected because no extra predicate participates in plan selection.

### Version info
- Resolved by: #105278
- Merged into: `26.6.1.121`
- Backported to: `26.5.2.10`, `26.4.4.14`

groeneai:
In `LEFT ANTI JOIN`, the right-side join key column was filled with the
left key's value for unmatched rows instead of the type default. For
example,

```sql
CREATE TABLE a (x Int32) ENGINE = Memory;
CREATE TABLE b (x Int32) ENGINE = Memory;
INSERT INTO a VALUES (1), (2), (3);
INSERT INTO b VALUES (2);

SELECT a.x AS left_x, b.x AS right_x
FROM a LEFT ANTI JOIN b ON (a.x = b.x);
```

returned `(1, 1)` and `(3, 3)` where it should return `(1, 0)` and
`(3, 0)`. Only the join key column was affected — non-key right columns
were already populated with defaults correctly. Reported in #99959 and
originally found by SQLancer (TLPDistinct oracle) in PR #98390.

### Root cause

`HashJoinResult::appendRightColumns` reconstructs each required
right-key column by copying from the corresponding left-key source via
`copyLeftKeyColumnToRight`. That copy is valid for `INNER`, `LEFT SEMI`,
and the matched portion of `LEFT JOIN` because the kept rows satisfy
`left.key == right.key`. For `LEFT ANTI` the kept rows are precisely
the unmatched ones, so copying produces the wrong value.

The fix detects `kind() == Left && strictness() == Anti` in
`appendRightColumns` and emits a column of type defaults of the required
right-key type instead. After `TableJoin::swapSides`, an originally
`RIGHT ANTI` query also reaches this path with `kind() == Left` and is
fixed by the same branch. The non-swapped `RIGHT ANTI` path
(`NotJoinedBlocks`) already worked correctly and is unaffected.

### Tests

New `04256_99959_left_anti_join_right_key_default.sql` covers `ON` and
`USING` syntax, `parallel_hash` and `grace_hash` algorithms, multi-column
right tables (verifying non-key right columns still default correctly),
`String` keys, swapped and non-swapped `RIGHT ANTI`, and multi-key joins.

Existing references that encoded the buggy behavior have been updated to
the correct values:
- `01031_pmj_new_any_semi_join.reference`
- `01031_semi_anti_join.reference`
- `02771_semi_join_use_nulls.reference`
- `03006_join_on_inequal_expression_fast_2.reference`
- `03211_convert_outer_join_to_inner_join_anti_join.reference` (also
  flips the row count the test description had flagged as
  "FIRST should be equal to SECOND")
- `03261_pr_semi_anti_join.reference`

### Closes

Closes #99959.

`LEFT ANTI JOIN` now correctly returns the right-side join key column with default values for unmatched rows, instead of duplicating the left key value. Previously, `SELECT a.x AS l, b.x AS r FROM a LEFT ANTI JOIN b ON a.x = b.x` could return `r = a.x` for the unmatched rows when it should return the default (`0` for `Int32`, `''` for `String`, `NULL` for `Nullable`). Only the join key column was affected; non-key right columns were already defaulted correctly. The same bug appeared in `RIGHT ANTI JOIN` whenever the optimizer swapped sides. Closes #99959.

- [x] Documentation is unchanged. The fix restores documented behavior.

### Version info
- Merged into: `26.6.1.121`
- Backported to: `26.5.2.10`, `26.4.4.14`


groeneai:
cc @vdimir @KochetovNicolai — could you review this `LEFT ANTI JOIN` engine fix? The right-side join key column was being filled with the left key value for unmatched rows instead of the type default. Fix is in `HashJoinResult::appendRightColumns`: detect `kind() == Left && strictness() == Anti` and emit type defaults for the right key column rather than copying from the left source.

cc @fm4v — fix for #99959 you flagged. Reproducer from the issue now returns `(1, 0), (3, 0)` instead of `(1, 1), (3, 3)`. Six existing test references that had silently encoded the bug are updated to the correct values.

### Pre-PR validation gate (per TASK.md Phase 4 step 9)

a) **Deterministic repro?** Yes — the issue's reproducer fails 100% on master and passes 100% with the fix.

b) **Root cause explained?** Yes — `HashJoinResult::appendRightColumns` reconstructs the right-key column by copying `properties.required_right_keys_sources[i]` (the left source column). The invariant "kept rows satisfy `left.key == right.key`" holds for `INNER`, `LEFT SEMI`, and the matched part of `LEFT JOIN`, but is violated for `LEFT ANTI` whose kept rows are exactly the unmatched ones, so the copy emits the left key value where the right key should be the type default.

c) **Fix matches root cause?** Yes — the fix replaces the copy with a defaults column of `right_key.type` only when `kind() == Left && strictness() == Anti`. No assertion widening, no setting pinning, no `no-random-*` tags.

d) **Test intent preserved + new tests added?** Yes — added `04256_99959_left_anti_join_right_key_default.sql` exercising `ON`/`USING`, `parallel_hash`, `grace_hash`, multi-column right tables (verifies non-key right columns still default correctly), `String` keys, swapped + non-swapped `RIGHT ANTI`, multi-key joins. The 6 existing references that encoded the buggy output were updated — they were never asserting the bug as intended, they just happened to capture the wrong values.

e) **Both directions demonstrated?** Yes — locally verified: fails without the fix (output `1,1` and `3,3`), passes with the fix (output `1,0` and `3,0`). Done by `git stash` + rebuild + run, then `git stash pop` + rebuild + run.

f) **General not narrow?** Yes — fix is at the common `HashJoin` output path used by `HashJoin`, `ConcurrentHashJoin` (parallel_hash), and `GraceHashJoin`. Also covers `RIGHT ANTI` after `TableJoin::swapSides` (kind becomes `Left+Anti` internally). The native non-swapped `RIGHT ANTI` path (`NotJoinedBlocks`) already worked correctly and is not touched.

Session: `cron:clickhouse-ci-task-worker:20260518-203000`

PedroTadim:
@groeneai check the failing test 02771_semi_join_use_nulls

groeneai:
@PedroTadim — the failing `02771_semi_join_use_nulls` is a runtime budget issue, not an output mismatch. CIDB shows the four failures on commit 8b56afcc are all "Test runs too long (> 180s). Make it faster." in flaky-check jobs (asan/debug). The j2 template generates 1442 SQL lines (16 outer combinations × many inner cases); on master the single-run p95 is ~111s, max ~575s, so the flaky check (50 reruns under sanitizers) reliably crosses the 180s per-run budget.

Pushed 3a9b604: add `-- Tags: long` to the test template. This is the canonical bypass — `clickhouse-test:3157` skips the 180s gate when `long` is in the tags, and `long_test_runs_ratio` also scales down the iteration count. Same pattern used by 01049_join_low_card_bug_long, 01721_join_implicit_cast_long, 02372_analyzer_join, 02052_last_granula_adjust_logical_error. Tag is followed by a blank line so `getTestTagsLength` does not also consume the `-- { echoOn }` directive.

src/Interpreters/HashJoin/HashJoinResult.cpp
```diff
@@ -133,10 +143,22 @@ static void appendRightColumns(
...
-        const auto & left_column = block.getByName(properties.required_right_keys_sources[i]);
         const auto & right_col_name = table_join.renamedRightColumnName(right_key.name);
-        const auto * filter_ptr = properties.need_filter ? nullptr : &filter;
-        auto right_col = copyLeftKeyColumnToRight(right_key.type, right_col_name, left_column, filter_ptr);
+
+        ColumnWithTypeAndName right_col;
+        if (is_left_anti_join)
+        {
+            /// All rows in the block are unmatched: emit type defaults for the right key.
+            auto column = right_key.type->createColumn();
```
vdimir, line 149: Any chance properties.need_filter is set there and we should handle it as it's done in another branch?

src/Interpreters/HashJoin/HashJoinResult.cpp
```diff
@@ -123,6 +123,16 @@ static void appendRightColumns(
...
+    /// For `LEFT ANTI JOIN`, the block only contains unmatched left rows by construction,
+    /// so the right-side key columns must be filled with type defaults rather than copied
+    /// from the left key column. The usual `copyLeftKeyColumnToRight` path relies on the
+    /// invariant that, for the rows kept by the upstream filter, `left.key == right.key`
+    /// (which holds for `INNER`, `LEFT SEMI`, and the matched part of `LEFT JOIN`).
+    /// `LEFT ANTI` violates that invariant: the kept rows are exactly the rows where no
+    /// match was found, so copying the left key value into the right key column produces
+    /// wrong results (issue #99959). After `TableJoin::swapSides`, an original `RIGHT ANTI`
+    /// also reaches this code path with `kind() == Left` and is fixed by the same branch.
+    bool is_left_anti_join = table_join.kind() == JoinKind::Left && table_join.strictness() == JoinStrictness::Anti;
```
vdimir: Consider adding more concise comment or removing

src/Interpreters/HashJoin/HashJoinResult.cpp
groeneai: Verified — for LEFT ANTI, `JoinFeatures::need_filter` (`JoinFeatures.h:31`) is `!need_replication && (... || (is_anti_join && left))`, which is always `true` here. `HashJoinResult::next` therefore always filters the block to unmatched rows before `appendRightColumns` is called (lines 387-388 / 513-514), so `block.rows()` already reflects the unmatched count and no `filter_ptr` handling is needed on this branch.

The else branch's `filter_ptr = properties.need_filter ? nullptr : &filter` cannot fire here for LEFT ANTI (it would always evaluate to `nullptr`), so introducing the same `filter`-driven path would be dead code.

Added `chassert(!is_left_anti_join || properties.need_filter)` in commit da7b2ac46553 to document the invariant and catch future regressions if `JoinFeatures::need_filter` is ever changed.

src/Interpreters/HashJoin/HashJoinResult.cpp
groeneai: Trimmed in commit da7b2ac46553: the 9-line block is replaced by a 3-line note plus a 2-line annotation next to the new `chassert`. The annotation names `JoinFeatures::need_filter` as the source of the invariant, which I think is the minimal context worth keeping for future readers. Happy to remove it entirely if you prefer.

groeneai:
@vdimir thanks for the review. Pushed `da7b2ac46553` addressing both comments:

1. `need_filter` handling: `JoinFeatures::need_filter` is always `true` for `is_anti_join && left` (`JoinFeatures.h:31`), so `HashJoinResult::next` has already filtered the block to unmatched rows by the time `appendRightColumns` is called. `block.rows()` is therefore the correct count, and the else branch's `filter_ptr = need_filter ? nullptr : &filter` would always evaluate to `nullptr` on this path. Added `chassert(!is_left_anti_join || properties.need_filter)` to document the invariant and catch a future change to `JoinFeatures::need_filter`.

2. Comment trimmed from 9 lines to 3 lines plus a 2-line annotation next to the `chassert`.

No behavior change. Regression test `04256_99959_left_anti_join_right_key_default` plus the five sibling tests (`02771_semi_join_use_nulls`, `01031_pmj_new_any_semi_join`, `01031_semi_anti_join`, `03006_join_on_inequal_expression_fast_2`, `03211_convert_outer_join_to_inner_join_anti_join`, `03261_pr_semi_anti_join`) all pass 10/10 with full randomization on a Debug build.

Pre-PR validation gate (commit `da7b2ac46553`):
- a. Deterministic repro: yes, `#99959` reproducer plus the new `04256` test.
- b. Root cause: unchanged from the parent commit; this fix-up only adds defensive documentation.
- c. Fix matches root cause: yes; `chassert` codifies the `JoinFeatures` invariant that the LEFT ANTI branch relies on.
- d. Test intent preserved: yes; no test changes in this commit.
- e. Both directions: original `8b56afcc05fd` already validated; `chassert` is a no-op on correct paths.
- f. General fix: yes, the `chassert` covers all LEFT ANTI code paths through `appendRightColumns`.

groeneai:
Cross-reference: this PR also resolves the reproducer in issue #105738 (LEFT ANTI JOIN baseline returning 888/999 instead of 0/0). Same root cause as #99959. Verified locally on master HEAD with the patch applied: both queries in #105738 now return `0, 0`.

Session: cron:clickhouse-ci-task-worker:20260524-191500

fm4v:
@groeneai add test from https://github.com/ClickHouse/ClickHouse/issues/105738

PR #72939, opened 2024-12-09
Fix stuck while processing pending batch for async distributed INSERT

azat:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Fix stuck while processing pending batch for async distributed INSERT (due to i.e. `No such file or directory`)

In #57480 I've fixed exactly the same problem (stuck due to non-existing file), but in #66375 I broke it since before checking valid() the code now tries to read files and got "No such file or directory" exception.

Rewrite the code to reduce the complexity by:
- removing extra helpers
- remove some object flags (recovered) and use function arguments instead

It also fixes one typo for accounting total_bytes in case of recovery.

PR #105281, opened 2026-05-18
Fix silent data loss in DistributedAsyncInsertBatch::recoverBatch when middle file is broken

groeneai:
Fixes #101745.

`DistributedAsyncInsertBatch::recoverBatch` iterates over `files` to validate
that every batch file has a readable header, but inside the loop it reads
`files.back()` instead of the current loop variable `file`. As a result, only
the last file's header is ever validated; the loop just re-reads that one file
N times and accumulates its rows N times into `total_rows`/`total_bytes`.

If a `Distributed` table is mid-batch when the server is killed (between
`unlink(*.bin)` and `unlink(current_batch.txt)`) and the last `.bin` in the
batch is intact while a middle one is corrupted, `recoverBatch` returns
`true`, `sendBatch` is then called, hits the corrupted file, throws, and
the whole batch — including the intact files — is moved to `broken/`. The
rows in the intact files are silently lost.

Using the loop variable makes each file's header validated on its own, so a
broken middle file causes recovery to return `false`. `current_batch.txt` is
then removed by the unconditional `fs::remove` in
`processFilesWithBatching`, and the surviving files are re-processed
individually through the normal pending-files path, where only the actually
broken file is moved to `broken/`. The rest reach the remote shard.

Bug originally found by automated review of #72939 (cc @clickgapai), confirmed
on master at `f86671aa80af`. Triage requested by @azat in #101745.

A new integration test `test_distributed_async_insert_batch_recovery`
simulates the abnormal-shutdown state by stopping `node_dist`, manually
writing a `current_batch.txt` referencing three `.bin` files, truncating the
middle one to 0 bytes, then restarting and verifying that the two intact
files reach `node_shard` and only the corrupted file ends up in `broken/`.

Fix silent data loss in `Distributed` async inserts when recovering from an abnormal shutdown: if the last `.bin` file in a saved batch was intact but a middle one was corrupted, `DistributedAsyncInsertBatch::recoverBatch` would only validate the last file's header and then `sendBatch` would mark the entire batch — including the intact files — as broken, losing their rows. Each file's header is now validated individually so only the actually broken file is moved to `broken/` and the surviving rows reach the remote shard.

Documentation entry for user-facing changes
- [ ] Documentation entry needed

### Version info
- Merged into: `26.5.1.815`

### DIFF
--- a/tests/integration/test_distributed_async_insert_batch_recovery/test.py
+++ b/tests/integration/test_distributed_async_insert_batch_recovery/test.py
@@ -7,14 +7,14 @@ from helpers.cluster import ClickHouseCluster
 
 cluster = ClickHouseCluster(__file__)
 
-node_dist = cluster.add_instance(
-    "node_dist",
+# Single-node setup: the Distributed table points back at the same node via
+# the `test_cluster` definition, and each `INSERT` is issued with
+# `prefer_localhost_replica = 0` so the engine still goes through the
+# async-insert batch queue (instead of the local-shortcut path).
+node = cluster.add_instance(
+    "node",
     main_configs=["configs/remote_servers.xml"],
     stay_alive=True,
-)
-node_shard = cluster.add_instance(
-    "node_shard",
-    main_configs=["configs/remote_servers.xml"],
 )
 
 
@@ -63,29 +63,31 @@ def test_recover_batch_with_broken_middle_file(started_cluster):
     `false`, `current_batch.txt` is removed, and the intact files are
     re-processed individually so their rows reach the remote shard.
     """
-    node_dist.query("DROP TABLE IF EXISTS dist SYNC")
-    node_dist.query("DROP TABLE IF EXISTS local SYNC")
-    node_shard.query("DROP TABLE IF EXISTS local SYNC")
+    node.query("DROP TABLE IF EXISTS dist SYNC")
+    node.query("DROP TABLE IF EXISTS local SYNC")
 
-    node_shard.query("CREATE TABLE local (x UInt32) ENGINE = MergeTree ORDER BY tuple()")
-    # `local` on `node_dist` is only used to satisfy the Distributed engine signature;
-    # all rows are routed to `node_shard` via the cluster definition.
-    node_dist.query("CREATE TABLE local (x UInt32) ENGINE = MergeTree ORDER BY tuple()")
-    node_dist.query(
+    node.query("CREATE TABLE local (x UInt32) ENGINE = MergeTree ORDER BY tuple()")
+    node.query(
         "CREATE TABLE dist (x UInt32) ENGINE = Distributed(test_cluster, default, local)"
         " SETTINGS background_insert_batch = 1, background_insert_split_batch_on_failure = 0"
     )
 
     # Stop sends so the inserts accumulate as separate `.bin` files in the queue.
-    node_dist.query("SYSTEM STOP DISTRIBUTED SENDS dist")
+    node.query("SYSTEM STOP DISTRIBUTED SENDS dist")
+
+    # `prefer_localhost_replica = 0` forces the Distributed engine to treat the
+    # local replica as remote and route the insert through the async-insert
+    # batch queue (otherwise it would take the local shortcut and bypass `.bin`
+    # file creation entirely).
+    insert_settings = {"prefer_localhost_replica": 0}
 
     # Three separate inserts -> three `.bin` files (1.bin, 2.bin, 3.bin).
-    node_dist.query("INSERT INTO dist VALUES (1)")
-    node_dist.query("INSERT INTO dist VALUES (2)")
-    node_dist.query("INSERT INTO dist VALUES (3)")
+    node.query("INSERT INTO dist VALUES (1)", settings=insert_settings)
+    node.query("INSERT INTO dist VALUES (2)", settings=insert_settings)
+    node.query("INSERT INTO dist VALUES (3)", settings=insert_settings)
 
-    queue_path = _shard_queue_path(node_dist, "dist")
-    bin_files = node_dist.exec_in_container(
+    queue_path = _shard_queue_path(node, "dist")
+    bin_files = node.exec_in_container(
         ["bash", "-c", f"ls {queue_path}/*.bin | sort"]
     ).strip().splitlines()
     assert len(bin_files) == 3, f"Expected 3 .bin files, got: {bin_files}"
@@ -107,10 +109,10 @@ def test_recover_batch_with_broken_middle_file(started_cluster):
     # together with `background_insert_batch=1`). Without a hard kill, the
     # three queued files would be delivered before the simulated abnormal
     # state could be set up.
-    node_dist.stop_clickhouse(kill=True)
+    node.stop_clickhouse(kill=True)
 
     # Sanity-check that the hard kill preserved the queued files on disk.
-    bin_files_after_kill = node_dist.exec_in_container(
+    bin_files_after_kill = node.exec_in_container(
         ["bash", "-c", f"ls {queue_path}/*.bin | sort"]
     ).strip().splitlines()
     assert len(bin_files_after_kill) == 3, (
@@ -119,7 +121,7 @@ def test_recover_batch_with_broken_middle_file(started_cluster):
 
     current_batch_lines = "".join(f"{i}\n" for i in indices)
     encoded = base64.b64encode(current_batch_lines.encode()).decode()
-    node_dist.exec_in_container(
+    node.exec_in_container(
         [
             "bash",
             "-c",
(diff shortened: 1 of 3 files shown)


groeneai:
cc @azat @devcrafter — could you review this? One-character fix for the `recoverBatch` copy-paste bug you flagged in #101745 (`files.back()` → `file` inside `for (const auto & file : files)`), plus a regression test that simulates the abnormal-shutdown state with a corrupted middle .bin file and verifies the intact files survive.

Pre-PR validation gate (per TASK.md Phase 4 Step 9):
- (a) Deterministic repro: integration test `test_distributed_async_insert_batch_recovery` simulates the exact abnormal-shutdown state (manual `current_batch.txt` + truncated middle .bin).
- (b) Root cause: `recoverBatch` iterates `for (const auto & file : files)` but reads `files.back()` inside the loop, so only the last file's header is validated; `total_rows`/`total_bytes` are accumulated N times from that one file. Confirmed at `src/Storages/Distributed/DistributedAsyncInsertBatch.cpp:222` on master.
- (c) Fix matches root cause: single-character change — read `file` (loop variable) instead of `files.back()`.
- (d) Test intent preserved: new test added; no existing tests modified.
- (e) Both directions: with the fix, `recoverBatch` correctly detects the broken middle file and returns `false`, so `current_batch.txt` is removed (per the unconditional `fs::remove` in `processFilesWithBatching`) and the surviving files are re-processed individually through the normal pending-files loop. Without the fix, `recoverBatch` returns `true` on an intact last file and `sendBatch` then marks every file in the batch as broken — losing the intact files. Local build (clang-21, debug) passes; full integration-test run will be exercised in CI.
- (f) Fix is general, not narrow: this is the only call site in `DistributedAsyncInsertBatch.cpp` with this pattern — `sendBatch` (line 258), `sendSeparateFiles` (line 316), and `serialize` (line 177) all correctly use the loop variable.

session-id: cron:clickhouse-ci-task-worker:20260518-221500

tiandiwonder (approved):

tiandiwonder:
@groeneai fix CI failures

groeneai:
Hi @tiandiwonder, thanks for the approval and the heads-up on CI. Triage of the 7 failing jobs:

### 1) Six integration test failures — all the same failure in the new regression test (real, PR-caused)

All six integration jobs failed identically on `test_distributed_async_insert_batch_recovery/test.py::test_recover_batch_with_broken_middle_file`:

```
assert rows == [str(i) for i in indices if i != middle_idx]
AssertionError: Expected the two intact files to be delivered, got: ['1', '2', '3']
assert ['1', '2', '3'] == ['1', '3']
```

**Root cause: the new test, not the C++ fix.** The test used `stop_clickhouse()` (graceful, `SIGTERM`) to simulate an abnormal shutdown. Graceful shutdown calls `StorageDistributed::flushAndPrepareForShutdown`, which calls `flushClusterNodesAllDataImpl(..., flush=true)` whenever `flush_on_detach=1`. And `flush_on_detach=0` is forbidden together with `background_insert_batch=1` (enforced in the `StorageDistributed` constructor). So graceful shutdown flushed all three `.bin` files to the shard, bypassing `SYSTEM STOP DISTRIBUTED SENDS`.

Server log from one of the failing CI runs confirms this exactly:
```
01:40:52.796 SYSTEM STOP DISTRIBUTED SENDS dist
01:40:53.249 INSERT (1)
01:40:53.306 Skipping send data over distributed table.     ← blocker works for the bg path
01:40:53.720 INSERT (2)
01:40:54.195 INSERT (3)
01:40:54.996 Sending a batch of 3 files to node_shard:9000 (3.00 rows, 12.00 B bytes).
01:40:55.066 Sent a batch of 3 files (took 72 ms).
01:40:55.066 Pending INSERT blocks flushed, took 73 ms.    ← from flushAndPrepareForShutdown
01:40:55.066 Start preparing for shutdown
```
By the time the test wrote `current_batch.txt` and `truncate -s 0 2.bin`, the queue was empty (only the truncated `2.bin` survived, recreated by `truncate`) and the shard already had rows `1, 2, 3`. The C++ `recoverBatch` fix actually worked — the server log shows `Renamed .../2.bin to .../broken/2.bin` — but the assertion still failed because the shard had been populated before the simulated abnormal state existed.

**Fix pushed (5597598):** switch to `stop_clickhouse(kill=True)` to `SIGKILL` the server before any graceful flush can run, plus a sanity check that the three `.bin` files survive the kill. The production-code change in `DistributedAsyncInsertBatch.cpp` is unchanged.

### 2) `Unit tests (msan, function_prop_fuzzer)` — known chronic flake (not PR-related)

```
FunctionsStress FAIL
Logical error: 'Function writeSlice expects same column types for GenericValueSlice and GenericArraySink.'
```

This is the chronic `FunctionsStress` flake tracked under [#104877](https://github.com/ClickHouse/ClickHouse/issues/104877). CIDB over the last 30 days:
- **116 failures across 54 distinct PRs** (including 10 master hits)
- The exact `writeSlice ... GenericValueSlice/GenericArraySink` pattern hits all three sanitizer variants (asan, msan, tsan) on `function_prop_fuzzer` across 8+ unrelated PRs.
- This PR touches only `src/Storages/Distributed/DistributedAsyncInsertBatch.cpp` and the new test — nothing in `src/Functions/`/`src/AggregateFunctions/`, so the failure is structurally impossible to attribute to this PR.

No action needed on this PR; will retry on the next CI run.

### Summary

| # | Job | Cause | Action |
|---|-----|-------|--------|
| 1-6 | 6× integration test | New test uses graceful shutdown which bypasses `SYSTEM STOP DISTRIBUTED SENDS` and flushes the queue | Fixed in 5597598 (`kill=True`) |
| 7 | `Unit tests (msan, function_prop_fuzzer)` / `FunctionsStress` | Chronic trunk flake, [#104877](https://github.com/ClickHouse/ClickHouse/issues/104877) | None — wait for re-run |

CI re-run is in flight on the new head. PTAL once it goes green.

tests/integration/test_distributed_async_insert_batch_recovery/test.py
```diff
@@ -0,0 +1,153 @@
...
+    # Truncate the middle file so reading its header fails with
+    # ATTEMPT_TO_READ_AFTER_EOF, which `isDistributedSendBroken` treats as broken.
+    node_dist.exec_in_container(["truncate", "-s", "0", middle_file])
+
+    node_dist.start_clickhouse()
+
+    # Drain the queue. With the fix, `recoverBatch` detects the broken middle
+    # file, returns false, `current_batch.txt` is removed, and the surviving
+    # files are re-processed individually.
+    node_dist.query("SYSTEM START DISTRIBUTED SENDS dist")
```
azat: STOP is not permanent, no need to START after restart, it is already started

azat:
```suggestion
```

tests/integration/test_distributed_async_insert_batch_recovery/test.py
```diff
@@ -0,0 +1,152 @@
...
+
+node_dist = cluster.add_instance(
+    "node_dist",
+    main_configs=["configs/remote_servers.xml"],
+    stay_alive=True,
+)
+node_shard = cluster.add_instance(
+    "node_shard",
+    main_configs=["configs/remote_servers.xml"],
+)
```
azat: We can also use single cluster setup with `prefer_localhost_replica`, to reduce overhead

groeneai:
Thanks @azat — your commit `8c540555` covers comments 1+2, and I've now applied comment 3 in commit [`75d0404b`](https://github.com/ClickHouse/ClickHouse/pull/105281/commits/75d0404b1476): collapsed the two-node setup to a single `node` with `prefer_localhost_replica = 0` on each `INSERT` to force the async-insert batch queue path. The `test_cluster` definition now points at the same node via `<host>node</host>`, and the test asserts both directions on that one node.

Local validation against a binary on this branch without the `recoverBatch` fix applied to the source tree:

```
test_distributed_async_insert_batch_recovery/test.py:145
    assert rows == [str(i) for i in indices if i != middle_idx]
E   AssertionError: Expected the two intact files to be delivered, got: []
E   assert [] == ['1', '3']
```

That confirms the new scaffolding still produces the three queued `.bin` files (so `prefer_localhost_replica = 0` correctly routes through the async batch queue instead of the local-shortcut path), `kill=True` preserves them on disk, the restart + `SYSTEM FLUSH DISTRIBUTED` cycle exercises the recovery path, and the final delivery-correctness assertion still catches the bug (rows lost when the C++ fix is absent). CI will validate the with-fix direction on `groeneai/fix-distributed-async-insert-batch-recover-101745`.

Pre-PR validation gate re-attested on the refactored test (per [TASK.md](https://github.com/ClickHouse/ClickHouse/pull/105281) Phase 4 Step 9):

- (a) **Deterministic repro**: yes — without the source-side fix on this branch, the assertion at line 145 deterministically fails with `[] == ['1', '3']`.
- (b) **Root cause**: unchanged — `DistributedAsyncInsertBatch::recoverBatch` iterates over `files` but reads `files.back()` inside the loop, so it only validates the last file's header. When the middle file is corrupted but the last is intact, recovery returns `true`, `sendBatch` then fails on the broken file, and the whole batch (including the intact files) is moved to `broken/` — silent data loss for the intact rows.
- (c) **Fix matches root cause**: yes — the C++ change (commit `4219ef06`) reads the current `file` variable inside the loop, so the broken middle file is detected during recovery, `current_batch.txt` is removed, and the intact files are re-processed individually.
- (d) **Test intent preserved**: yes — the single-node setup still exercises the full async-insert batch queue: three separate INSERTs produce three separate `.bin` files, a hard kill leaves the queue intact for the abnormal-shutdown reproducer, and the final assertion still catches the silent data-loss symptom (lost rows would land in `[]`, not `['1', '3']`).
- (e) **Both directions demonstrated**: failing direction shown locally above. CI on the new HEAD will validate the passing direction with the fix-included binary.

Session: `cron:clickhouse-ci-task-worker:20260519-070000`

azat (approved):

clickgapai:

**Affected code:** `DistributedAsyncInsertBatch::recoverBatch` in src/Storages/Distributed/DistributedAsyncInsertBatch.cpp — introduced in 25.8.

**Why this might apply:** This is a data loss bug in `DistributedAsyncInsertBatch::recoverBatch`. The bug was introduced in PR #72939 (merged 2025-03-08) when the file was first created with the `recoverBatch` function. Since this predates all supported release branches (25.8 LTS released 2025-08-29), all branches contain the buggy code and need the fix. Silent data loss after abnormal server shutdown is a P1 severity issue.

PR #105283, opened 2026-05-18
Fix extract-from-config --try silently dropping from_env values

groeneai:
Fixes https://github.com/ClickHouse/ClickHouse/issues/101704.

### Motivation

After upgrading from 25.1.5.31 to 26.2.5, the Docker entrypoint port discovery silently stopped working. `clickhouse extract-from-config --config-file ... --key=http_port --try` returned an empty string instead of the value of `CLICKHOUSE_HTTP_PORT`, so `entrypoint.sh` failed to wait for the right port and reported `ClickHouse init process timeout`.

@GutOFF traced the regression to 6df7192cdea4 ("Ignore non existing `include_from` files with `--try` for clickhouse-extract-from-config") and @azat asked me to take a look.

### Root cause

`ConfigProcessor::processConfig` used to do:

```cpp
if (!throw_on_bad_include_from && !fs::exists(include_from_path))
{
    LOG_WARNING(log, "File {} (from 'include_from') does not exist. Ignoring.", include_from_path);
}
else
{
    processIncludes(...);
}
```

`processIncludes` is the function that runs `doIncludesRecursive`, which performs the `from_env` / `from_zk` / `incl` substitutions on the whole config — not just the include_from file parsing.

When `--try` is passed, `throw_on_bad_include_from` is `false`. When the config has no `<include_from>` element (and `/etc/metrika.xml` is absent), `include_from_path` is the empty string and `fs::exists("")` returns `false`. Both predicates are true, so the `else` branch is skipped — and with it, all `from_env` substitutions. Every `from_env` element gets serialized as an empty string, which is the bug.

The same silent drop also happens when `<include_from>` is set but the file does not exist; only the `--try` path was affected.

### Fix

Only suppress the include_from *file parsing* when `--try` is on and the path is non-empty and missing. Always call `processIncludes` so `doIncludesRecursive` runs the substitution pass. When the missing-file branch is taken, the path is cleared first so `processIncludes` sees "no include_from", which it already handles correctly.

### Regression test

`tests/queries/0_stateless/04251_101704_extract_from_config_try_resolves_from_env.sh` covers:

1. `from_env` without `include_from`, env set, `--try` — the reported bug
2. Same without `--try` (sanity)
3. `from_env` with a bad `include_from` path, `--try` — related case that this fix also covers
4. Missing env var with `--try` — empty result, no error
5. Glob key with `--try` — all three values resolved

Verified locally with the debug build:

- Test passes 50/50 with the fix.
- Test fails on every scenario without the fix (diff shows the 5 empty-line regressions).
- Existing `03344_clickhouse_extract_from_config_try` still passes.

Fix a regression in `clickhouse extract-from-config --try`: when the config used `from_env` attributes and had no `include_from` element (or had one pointing to a missing file), all `from_env` substitutions were silently dropped and returned empty strings. This broke the Docker entrypoint port discovery in 26.2.5. Closes #101704.

### Version info
- Merged into: `26.5.1.800`

### DIFF
--- a/src/Common/Config/ConfigProcessor.cpp
+++ b/src/Common/Config/ConfigProcessor.cpp
@@ -804,24 +804,27 @@ XMLDocumentPtr ConfigProcessor::processConfig(
                 include_from_path = default_path;
         }
 
-        if (!throw_on_bad_include_from && !fs::exists(include_from_path))
+        /// When --try is passed and the include_from file is missing, drop the path so that
+        /// processIncludes does not try to parse it. We must still call processIncludes
+        /// because it also performs from_env/from_zk/incl substitutions on the rest of the
+        /// config; skipping it would silently strip those values (issue #101704).
+        if (!throw_on_bad_include_from && !include_from_path.empty() && !fs::exists(include_from_path))
         {
             LOG_WARNING(log, "File {} (from 'include_from') does not exist. Ignoring.", include_from_path);
+            include_from_path.clear();
         }
-        else
-        {
-            processIncludes(
-                config,
-                substitutions,
-                include_from_path,
-                throw_on_bad_incl,
-                dom_parser,
-                log,
-                &contributing_zk_paths,
-                &contributing_files,
-                zk_node_cache,
-                zk_changed_event);
-        }
+
+        processIncludes(
+            config,
+            substitutions,
+            include_from_path,
+            throw_on_bad_incl,
+            dom_parser,
+            log,
+            &contributing_zk_paths,
+            &contributing_files,
+            zk_node_cache,
+            zk_changed_event);
     }
     catch (Exception & e)
     {
--- a/tests/queries/0_stateless/04251_101704_extract_from_config_try_resolves_from_env.reference
+++ b/tests/queries/0_stateless/04251_101704_extract_from_config_try_resolves_from_env.reference
@@ -1,0 +1,14 @@
+--- 1. from_env without include_from, env set, --try ---
+8123
+8443
+9000
+--- 2. from_env without include_from, env set, no --try ---
+8123
+--- 3. from_env with bad include_from, env set, --try ---
+8123
+--- 4. from_env without include_from, env NOT set, --try ---
+
+--- 5. glob *_port with --try ---
+8123
+8443
+9000
--- a/tests/queries/0_stateless/04251_101704_extract_from_config_try_resolves_from_env.sh
+++ b/tests/queries/0_stateless/04251_101704_extract_from_config_try_resolves_from_env.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+# Regression test for issue #101704
+# `clickhouse extract-from-config --try` must still resolve `from_env` substitutions
+# when the config has no `<include_from>` tag (or when it has one pointing to a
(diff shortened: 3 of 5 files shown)


groeneai:
### Pre-PR validation gate (session cron:clickhouse-ci-task-worker:20260518-224500)

**a) Deterministic repro?** ✅
```bash
CLICKHOUSE_HTTP_PORT=8123 clickhouse extract-from-config \
  --config-file config-with-from-env.xml --key=http_port --try
```
Before fix: empty. After fix: `8123`. Same input, deterministic.

**b) Root cause explained?** ✅
`--try` ⇒ `ignore_errors=true` in `ExtractFromConfig.cpp` ⇒ `throw_on_bad_include_from=false` in `ConfigProcessor`. With no `<include_from>` tag, `include_from_path = ""`. `fs::exists("")` returns `false`. The condition `!throw_on_bad_include_from && !fs::exists(include_from_path)` evaluates true, so `processIncludes` is skipped. But `processIncludes` is also what calls `doIncludesRecursive`, which performs the `from_env`/`from_zk`/`incl` substitution pass on the whole config — skipping it silently strips those values.

**c) Fix matches root cause?** ✅
Added `!include_from_path.empty()` to the suppression condition — only skip include_from file parsing when an explicit path was given and is missing. Cleared `include_from_path` in that branch and moved `processIncludes` out of the `else`, so substitutions ALWAYS run. Not a band-aid: the fix targets the exact condition that broke env substitution for the no-include_from case.

**d) Test intent preserved?** ✅
No existing test is modified. Existing `03344_clickhouse_extract_from_config_try` still passes (it covers bad-include-from + `--try` and the "Not found" + `--try` paths).

**e) Both directions demonstrated?** ✅
- With fix: 50/50 passes (`tests/clickhouse-test 04251_101704_... --test-runs 50` → `50 tests passed`).
- Without fix (temporary revert + rebuild): test FAILS, diff shows all 5 scenarios losing the env-resolved values (`-8123`/`-8443`/`-9000` → empty).

**f) Fix is general, not a narrow patch?** ✅
- Same code path also fixes the case where `<include_from>` is set but missing (related sub-bug — both now correctly run env substitution under `--try`).
- The fix is in `ConfigProcessor` itself, so any caller passing `throw_on_bad_include_from=false` benefits — not only `extract-from-config`. No symmetric callers needed changes (this is the only entry point that disables the include_from check).

groeneai:
cc @azat @alexey-milovidov @Algunenano — could you review? The regression is in 6df7192cdea4 (`--try` skip path). Component owners for `comp-config` per the worker's component map. @azat is the author of the regression and asked me to take a look on issue #101704.

azat (approved):

clickgapai:

**Affected code:** `ConfigProcessor::processConfig` — the `--try` codepath that skips `processIncludes` when include_from file is missing — core code present in all supported branches.

**Why this might apply:** The bug was introduced in commit 6df7192cdea4 ('Ignore non existing include_from files with --try for clickhouse-extract-from-config') dated April 5, 2025. This commit was merged before all currently supported branches (25.8, 26.2, 26.3, 26.4) were released, so all of them contain the bug. The bug causes Docker entrypoint scripts to fail when discovering ports via environment variables - a user-facing regression that breaks deployments.

PR #103932, opened 2026-05-03
Add long tag to 00446_clear_column_in_partition_concurrent_zookeeper

groeneai:
The test exercises concurrent `CLEAR COLUMN IN PARTITION` against a two-replica `ReplicatedMergeTree` with `replication_alter_partitions_sync=2`. Each iteration issues four sync `ALTER`s plus background inserts, and ZooKeeper round-trips serialize behind replica acks. Under flaky-check (which serially re-runs the same test 50x with randomized settings), the cumulative runtime regularly exceeds 180s.

The runner enforces `TEST_MAX_RUN_TIME_IN_SECONDS = 180` in flaky-check mode and fails the test with `TOO_LONG` unless the `long` tag is present (`tests/clickhouse-test`, lines 3067-3083). The runner's own intent at the `no-long` skip site states: "Tests for races and deadlocks usually are run in a loop for a significant amount of time" — which is exactly what this test is.

CIDB shows 218 of 231 failures over the last 30 days are this exact pattern (`Reason: Test runs too long (> 180s). Make it faster.`) in `Stateless tests (*, flaky check)` jobs across 24 distinct PRs. PR #102394 already reduced the loop from 10 to 5 iterations (April 10), which proved insufficient — failures continued April 28-30 including one master hit on `Stateless tests (arm_asan_ubsan, azure, parallel)`.

Adding the `long` tag is the documented fix for this class of test and what PR #96130 author also applied locally on their branch (commit `b413b3be5a36`). The change preserves test intent (same loop count, same concurrent ops, same assertions) and only bypasses the 180s flaky-check budget check.

CIDB: https://play.clickhouse.com/?user=play&query=SELECT+check_start_time%2C+pull_request_number%2C+check_name%2C+substring%28test_context_raw%2C+1%2C+200%29+as+err+FROM+default.checks+WHERE+test_name+LIKE+%2700446_clear_column_in_partition_concurrent_zookeeper%25%27+AND+test_status+IN+%28%27FAIL%27%2C+%27ERROR%27%29+AND+check_start_time+%3E+now%28%29+-+INTERVAL+30+DAY+ORDER+BY+check_start_time+DESC+LIMIT+30+FORMAT+TabSeparatedWithNames

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
...

[...]

PR #100777, opened 2026-03-26
Fix: Fix workload IO scheduling bypass in DiskObjectStorageTransaction write path

JiaQiTang98:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix workload IO scheduling being silently bypassed for S3/object-storage writes that                                                                                        
  go through the `DiskObjectStorageTransaction` path (used by `s3_with_keeper` disks and                                                                                      
  explicitly when `use_fake_transaction=false`). Previously, INSERT writes into MergeTree                                                                                     
  tables on such disks ignored `CREATE RESOURCE` / `CREATE WORKLOAD` throttling because                                                                                       
  the resource link was never injected into `WriteSettings` for the transactional write path.                                                                                 
### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.120`
- Backported to: `26.3.10.50`

PR #98883, opened 2026-03-05
Lazily apply selector and replication indexes in join

m-selmi:
Part of https://github.com/ClickHouse/ClickHouse/issues/97881

Hash join produces probe side columns by filtering or replicating the input columns.
Materializing these columns can be wasted if they are immedtiatly dropped or filtered/replicated
by a follow up join.

This optimization tells the join to produce `ColumnReplicated` (input column + index) instead if one
of the following cases is detected:
- The join sits below a LimitStep or SortingStep with a small limit.
- The join sits on the probe side of an immediately following join.

### Benchmark results for join_lazy_indexing.xml:

100M rows in large tables, 8K-10K rows in small tables. Duplicate small tables have 2x inserts (same data twice).

#### Single-join, unique small side

| # | Query | Off (s) | On (s) | Δ% |
|---|-------|---------|--------|-----|
| 0/1   | `large_int JOIN small1 ON l.a = s.a LIMIT 80M, 10`          | 0.304 | 0.211 | **−30.6%** |
| 2/3   | `large_str JOIN small1 ON l.a = s.a LIMIT 80M, 10`          | 1.275 | 0.895 | **−29.8%** |
| 4/5   | `large_int JOIN small1 ON l.a = s.a ORDER BY l.a LIMIT 10`  | 0.327 | 0.253 | **−22.6%** |
| 6/7   | `large_str JOIN small1 ON l.a = s.a ORDER BY l.a LIMIT 10`  | 1.279 | 0.991 | **−22.5%** |
| 8/9   | `large_int JOIN small1 ON l.a = s.a ORDER BY l.a LIMIT 1000`| 0.362 | 0.292 | **−19.4%** |
| 10/11 | `large_str JOIN small1 ON l.a = s.a ORDER BY l.a LIMIT 1000`| 1.685 | 1.350 | **−19.9%** |

#### Single-join, duplicating small side

| # | Query | Off (s) | On (s) | Δ% |
|---|-------|---------|--------|-----|
| 12/13 | `large_int JOIN small1_dup ON l.a = s.a LIMIT 80M, 10`          | 0.207 | 0.206 | −0.7% |
| 14/15 | `large_str JOIN small1_dup ON l.a = s.a LIMIT 80M, 10`          | 0.582 | 0.580 | −0.3% |
| 16/17 | `large_int JOIN small1_dup ON l.a = s.a ORDER BY l.a LIMIT 10`  | 0.389 | 0.393 | +0.9% |
[...]

PR #105284, opened 2026-05-18
Fix data loss in clickhouse-local with a non-default `default_database`

groeneai:
`clickhouse-local --path X -- --default_database=mydb` silently loses tables on restart.

`createClickHouseLocalDatabaseOverlay` looked up the persisted database UUID at the hardcoded
`metadata/default`, but `DatabaseAtomic` creates the symlink at
`metadata/<escapeForFileName(name)>`. When the configured default database is anything other
than `default`, the lookup never matches the previous run's symlink, a fresh UUID is minted,
the previous run's data is orphaned in `store/`, and the table appears as `UNKNOWN_TABLE`
on the next invocation despite the bytes still being on disk.

The fix uses `DatabaseCatalog::getMetadataDirPath(name_)` — the same helper `DatabaseAtomic`
itself uses to compute `path_to_metadata_symlink` — so the lookup always agrees with where
the symlink was written.

**Reproducer (from #101831):**

```bash
clickhouse-local --path "custom_db_test" -q "
  CREATE TABLE test (x UInt64) ENGINE = MergeTree ORDER BY x;
  INSERT INTO test VALUES (42);
" -- --default_database=mydb

clickhouse-local --path "custom_db_test" -q "SELECT * FROM test" -- --default_database=mydb
# Before:  Code 60. UNKNOWN_TABLE
# After:   42
```

Regression test: `tests/queries/0_stateless/04252_clickhouse_local_default_database_persistence_101831.sh`
covers both the bug shape (`--default_database=mydb` round-trip) and the canonical
`default` path (no regression). Verified locally to FAIL without the fix and PASS with it.

Session id: cron:clickhouse-ci-task-worker:20260518-231500

Fix data loss in `clickhouse-local` when started repeatedly with a non-default `--default_database` (for example `clickhouse-local --path X -- --default_database=mydb`): the lookup for the persisted database UUID was hardcoded to `metadata/default`, so each restart minted a fresh UUID and the previous run's tables became invisible. Closes [#101831](https://github.com/ClickHouse/ClickHouse/issues/101831).

- [x] Documentation is unchanged (internal fix, behaviour matches the pre-existing canonical `default` path).

### Version info
- Merged into: `26.6.1.252`


groeneai:
### Pre-PR validation gate

| # | Question | Answer |
|---|---|---|
| a | Deterministic repro? | Yes. Two-shot `clickhouse-local --path X -- --default_database=mydb` reliably reproduces `UNKNOWN_TABLE` on the second invocation without the fix. |
| b | Root cause explained? | `metadata/default` was hardcoded as the symlink path even though `DatabaseAtomic` writes the symlink at `metadata/<escapeForFileName(name)>`. On a non-default `default_database`, the lookup never matches the previous run's symlink, a fresh `UUID` is generated, and the resulting `default_database_metadata_path = store/<new_uuid>` is empty. The previous run's `store/<old_uuid>` directory and its `metadata/mydb` symlink are orphaned. |
| c | Fix matches root cause? | Yes. The fix replaces the hardcoded constant with `DatabaseCatalog::getMetadataDirPath(name_)` — the exact same helper `DatabaseAtomic` itself uses (via `DatabaseCatalog::getMetadataDirPath(name_)` at `DatabaseAtomic.cpp:80`). The lookup now always agrees with where the symlink was created. |
| d | Test intent preserved? | N/A — new regression test covers (1) `--default_database=mydb` cross-invocation persistence (the bug), (2) the canonical `default` case as a no-regression control. |
| e | Both directions demonstrated? | Yes. With the fix reverted locally: test FAILs with `Code: 60. UNKNOWN_TABLE` on the second invocation. With the fix applied: test PASSes 5/5 runs. |
| f | Fix is general, not a narrow patch? | Yes. Searched the codebase for the same hardcoded `"metadata" / "default"` pattern — only one occurrence existed (this one). The fix uses the canonical helper, so it stays in sync if `DatabaseAtomic`'s symlink scheme ever changes again. |

CIDB-wise this affects every release where `--default_database=` is allowed in `clickhouse-local` (introduced by PR #96339). Not user-visible until restart, which is what makes it a silent data-loss bug.

groeneai:
cc @azat @alexey-milovidov @yakov-olkhovskiy — silent data-loss fix in `clickhouse-local` for non-default `--default_database` (#101831). Single-line behavior fix to the symlink lookup plus a stateless regression test; could you review?

alexey-milovidov (approved):

alexey-milovidov:
Thanks! This is a perfect fix!

PedroTadim:
@groeneai check the failing tests

groeneai:
@PedroTadim only failing check is `Fast test`. Triage:

`Fast test` killed `00446_clear_column_in_partition_concurrent_zookeeper` on a 60s timeout (SIGTSTP signal 20 while waiting on a `replication_alter_partitions_sync=2` ALTER). The test exercises concurrent `CLEAR COLUMN` on a 2-replica `ReplicatedMergeTree` with synchronous replication; runtime varies with ZooKeeper round-trip latency.

Chronic flake unrelated to this PR. CIDB over 30d:

* master: 4 hits, same Timeout / `Test runs too long (>180s)` pattern
* PR #96130: 5 hits, same
* PR #98883, #100777, this PR: 1 hit each, same

Fix already approved: PR #103932 (adds `long` tag, approved by @evillique 2026-05-04). Once it lands the test is excluded from `Fast test` (via `--no-long`) and from the 180s flaky-check budget.

This PR's diff is `src/Client/LocalServer.cpp` only; no path interacts with test 00446. Re-run `Fast test` or wait for #103932 to land.

Session: cron:clickhouse-ci-task-worker:20260519-190000

Issue #73519, opened 2024-12-18
SQLite query not correctly escaped


PedroTadim:
**Describe the bug**
Sqlite seems to not use `\` as the escape character. This only happens with `FixedString` type.

**How to reproduce**
In SQLite run:
```sql
CREATE TABLE tx(c0 TEXT);
```
Then in ClickHouse run:
```sql
CREATE TABLE tx(c0 FixedString(8)) ENGINE = SQLite('<path to sqlite db>', 'tx');
INSERT INTO TABLE tx (c0) VALUES ('\'');
--DB::Exception: Failed to execute sqlite INSERT query. Status: 1. Message: unrecognized token: "\"
```

jh0x:
In the query above, you'd want to insert `'`, right?

This is SQLite [syntax:](https://www.sqlite.org/lang_expr.html)

```sql
sqlite> INSERT INTO tx (c0) VALUES ('\');  -- insert single backslash
sqlite> INSERT INTO tx (c0) VALUES (''''); -- insert single quote
```

So the query here:
```sql
CREATE TABLE tx(c0 FixedString(8)) ENGINE = SQLite('<path to sqlite db>', 'tx');
INSERT INTO TABLE tx (c0) VALUES ('\'');
```
should be effectively inserting `''` into the underlying storage.

PR #60015, opened 2024-02-15
Fix INSERT into SQLite with single quote (by escaping single quotes with a quote instead of backslash)

azat:
### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Fix INSERT into SQLite with single quote (by properly escaping single quotes with a quote instead of backslash)

PR #105285, opened 2026-05-19
Honor output_format_values_escape_quote_with_quote in Enum/JSON/AggregateFunction/CustomSimpleText serializations

### LINKED ISSUE #103860: `SerializationEnum::serializeTextQuoted` ignores `output_format_values_escape_quote_with_quote`, causing `INSERT INTO sqlite_table` to fail with SQLite syntax error for `Enum` values whose name contains a single quote

_Retrospective finding from a historical scan of [PR #60015](https://github.com/ClickHouse/ClickHouse/pull/60015) (merged 2024-02-19). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

With `output_format_values_escape_quote_with_quote=1` (the setting forced on by `StorageSQLite::write_context`), `String` values are correctly escaped as `''` but `Enum` values are still escaped as `\'`. When the value name contains a single quote and the column is inserted into a SQLite table, the SQLite parser rejects the generated `INSERT INTO ... VALUES ('hi\'there')` with `near "there": syntax error` (`SQLITE_ENGINE_ERROR`).

**Root cause:** `SerializationEnum.cpp:62-65` declares `const FormatSettings &` as anonymous parameter and calls `writeQuotedString` unconditionally — the new `escape_quote_with_quote` branch (added in `SerializationString` and later `SerializationFixedString`) was never propagated to this serialization (and to three other peer serializations).

**Why we believe this is a bug:** `StorageSQLite::SQLiteSink::consume` (StorageSQLite.cpp:176) → `FormatFactory::getOutputFormat("Values", ...)` with `write_context` whose `output_format_values_escape_quote_with_quote=true` → `ValuesRowOutputFormat::writeField` → `serialization.serializeTextQuoted(...)`. For `Enum8`/`Enum16` columns, that resolves to `SerializationEnum<Type>::serializeTextQuoted` (SerializationEnum.cpp:62-65) which unconditionally calls `writeQuotedString` (always `\'` escape) and never reads `settings.values.escape_quote_with_quote`. The same omission exists in `SerializationCustomSimpleText.cpp:91`, `SerializationAggregateFunction.cpp:316`, and `SerializationJSON.cpp:324`.

**Affected locations:**
- `src/DataTypes/Serializations/SerializationEnum.cpp:64` — writeQuotedString called unconditionally; settings.values.escape_quote_with_quote ignored
- `src/DataTypes/Serializations/SerializationCustomSimpleText.cpp:91` — Same omission for custom simple text
- `src/DataTypes/Serializations/SerializationAggregateFunction.cpp:316` — Same omission for AggregateFunction state
- `src/DataTypes/Serializations/SerializationJSON.cpp:324` — Same omission for JSON

**Impact:** INSERT INTO a SQLite table on a column whose serialization goes through one of these four `serializeTextQuoted` overrides fails with `SQLITE_ENGINE_ERROR` whenever the value contains a single quote — exactly the failure mode this PR was meant to eliminate. Most reproducible case: `Enum` columns with quote characters in value names. Not a regression introduced by this PR — the PR's fix is incomplete.

Yes — confirmed on current `master` (commit `6d0cf962c38`).

### How to reproduce

```sql
-- Test: SerializationEnum::serializeTextQuoted should honor output_format_values_escape_quote_with_quote.
-- This setting is used by StorageSQLite::write_context to escape ' as '' (SQL standard) instead of \'.
-- An Enum8 column whose value name contains ' currently produces \' in Values output, which causes
-- INSERT INTO sqlite_table to fail with a SQLite syntax error.

set output_format_values_escape_quote_with_quote=1;

-- String correctly produces the SQL-standard '' escape:
select 'a''b' as s format Values;

select '\nenum:' format LineAsString;

-- Enum should produce the same escape for consistency, but currently outputs \' (BUG):
select cast('a''b', 'Enum8(\'a\'\'b\'=1)') as e format Values;

-- fix no newline at end of file
select '' format LineAsString;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/55c2b517-76b4-47d0-b5b3-0543f1348122)

### Expected behavior

```
('a''b')
enum:
('a''b')
```

```
('a''b')
enum:
('a\'b')
```

### Additional context

**Open risks:**
- Strings containing real control bytes (newline, tab, NUL) are still corrupted on SQLite roundtrip because `writeAnyEscapedString` still emits `\n`/`\t`/`\0` and SQLite does not interpret these — pre-existing; out of scope for this finding but worth a follow-up.
- `StorageSQLite::write_context` is built from `getGlobalContext()`, so per-query setting overrides don't propagate to the writer.

**Suggested fix:** Mirror the `SerializationFixedString` fix pattern (which was added later in PR #73519): branch on `settings.values.escape_quote_with_quote` and call `writeQuotedStringPostgreSQL` in the true branch. Apply the same fix to `SerializationEnum`, `SerializationCustomSimpleText`, `SerializationAggregateFunction`, and `SerializationJSON`.

Found during automated review of [PR #60015](https://github.com/ClickHouse/ClickHouse/pull/60015).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr60015_001`_

### Version info
- Resolved by: #105285
- Merged into: `26.5.1.803`

groeneai:
Fixes #103860.

`output_format_values_escape_quote_with_quote` is a setting that switches `Values` output from `\'` apostrophe escaping to SQL-standard `''` doubling — the form expected by SQLite. `StorageSQLite` forces it on for its write context so that `INSERT INTO sqlite_table` produces parseable SQLite statements.

PRs #60015 and #73519 added support for it in `SerializationString` and `SerializationFixedString` respectively, but four sibling serializations were missed. Their `serializeTextQuoted` overrides call `writeQuotedString` unconditionally and ignore `settings.values.escape_quote_with_quote`:

- `src/DataTypes/Serializations/SerializationEnum.cpp`
- `src/DataTypes/Serializations/SerializationJSON.cpp`
- `src/DataTypes/Serializations/SerializationAggregateFunction.cpp`
- `src/DataTypes/Serializations/SerializationCustomSimpleText.cpp`

For `Enum` columns this causes `INSERT INTO sqlite_table SELECT enum_col` to fail with a SQLite parser error (`SQLITE_ENGINE_ERROR`, `near "...": syntax error`) when the enum value name contains a single quote — the exact failure mode the setting was added to eliminate.

The fix mirrors the canonical pattern from `SerializationString::serializeTextQuoted`: branch on `settings.values.escape_quote_with_quote` and call `writeQuotedStringPostgreSQL` in the true branch. Applied to all four sites uniformly.

Regression test `04253_sqlite_escape_quote_with_quote_103860.sh` exercises `Enum`, `JSON`, and `AggregateFunction` via `format Values` and includes a `String` sanity check. `SerializationCustomSimpleText` has no concrete instantiations in the current codebase — the fix is applied for consistency to prevent future use from re-introducing the bug.

Fix `INSERT INTO sqlite_table` failing with a SQLite syntax error when the inserted value is an `Enum`, `JSON`, or `AggregateFunction` whose text representation contains a single quote. The `output_format_values_escape_quote_with_quote` setting is now honored by the corresponding serializations (previously only `String` and `FixedString` honored it).

### Version info
- Merged into: `26.5.1.803`

### DIFF
--- a/tests/queries/0_stateless/04254_sqlite_escape_quote_with_quote_103860_roundtrip.sh
+++ b/tests/queries/0_stateless/04254_sqlite_escape_quote_with_quote_103860_roundtrip.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+# Tags: no-fasttest
+
+# Regression test for https://github.com/ClickHouse/ClickHouse/issues/103860 — deserializer side.
+#
+# Companion to 04253_sqlite_escape_quote_with_quote_103860.sh which verified the
+# EMIT side: `output_format_values_escape_quote_with_quote = 1` makes
+# `JSON` / `Enum` / `AggregateFunction` / `CustomSimpleText` produce the
+# SQL-standard `''` apostrophe escape in `Values` output instead of `\'`.
+#
+# This test verifies the PARSE side: a `Values` literal containing `''` must
+# be parseable back through each of these serializations.
+#
+# Before the fix, `SerializationJSON::deserializeTextQuoted` and
+# `SerializationCustomSimpleText::(try)deserializeTextQuoted` used
+# `readQuotedString` / `tryReadQuotedString`, which do NOT accept the
+# SQL-standard `''` form — they treat the second `'` as the closing quote and
+# leave the rest of the literal unread, breaking the
+# serializer / deserializer contract for `Values`.
+# After the fix they use `readQuotedStringWithSQLStyle` /
+# `tryReadQuotedStringWithSQLStyle`, which accept both `''` and `\'`.
+#
+# `SerializationEnum::deserializeTextQuoted` and
+# `SerializationAggregateFunction::deserializeTextQuoted` already used the
+# SQL-style reader before the fix; they are still exercised here to lock down
+# the symmetry across all four serializations touched in #103860.
+#
+# The `Values` parser has a SQL-expression-parser fallback (and a template
+# fallback) when `deserializeTextQuoted` throws, which would mask the
+# asymmetry for `JSON` (the SQL parser also accepts `''` inside string
+# literals). To exercise the `deserializeTextQuoted` path directly and ensure
+# the deserializer is what actually accepts `''`, both fallbacks are disabled:
+# `input_format_values_interpret_expressions = 0` and
+# `input_format_values_deduce_templates_of_expressions = 0`.
+
+CUR_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=../shell_config.sh
+. "$CUR_DIR"/../shell_config.sh
+
+CH_CLIENT_NO_FALLBACK="${CLICKHOUSE_CLIENT} --input_format_values_interpret_expressions=0 --input_format_values_deduce_templates_of_expressions=0"
+
+# --- JSON ---
+# Feed the SQL-style-escaped Values literal via stdin so neither bash nor the
+# outer SQL parser interferes with the apostrophes; the only parser that sees
+# them is `SerializationJSON::deserializeTextQuoted`.
+${CLICKHOUSE_CLIENT} -q "SELECT '--- JSON, SQL-style apostrophe escape (was broken pre-fix) ---' FORMAT LineAsString;"
+${CLICKHOUSE_CLIENT} -q "DROP TABLE IF EXISTS t_json_roundtrip_103860; CREATE TABLE t_json_roundtrip_103860 (j JSON) ENGINE = Memory SETTINGS enable_json_type = 1;"
+printf "('{\"k\":\"a''b\"}')" | ${CH_CLIENT_NO_FALLBACK} --query="INSERT INTO t_json_roundtrip_103860 FORMAT Values" --enable_json_type=1
+${CLICKHOUSE_CLIENT} -q "SELECT toString(j) FROM t_json_roundtrip_103860;"
+${CLICKHOUSE_CLIENT} -q "DROP TABLE t_json_roundtrip_103860;"
+echo
+
+# Same JSON parse path, with the legacy backslash form — must keep working.
+${CLICKHOUSE_CLIENT} -q "SELECT '--- JSON, legacy backslash apostrophe escape (must still parse) ---' FORMAT LineAsString;"
+${CLICKHOUSE_CLIENT} -q "DROP TABLE IF EXISTS t_json_roundtrip_103860; CREATE TABLE t_json_roundtrip_103860 (j JSON) ENGINE = Memory SETTINGS enable_json_type = 1;"
+printf "('{\"k\":\"a\\\\'b\"}')" | ${CH_CLIENT_NO_FALLBACK} --query="INSERT INTO t_json_roundtrip_103860 FORMAT Values" --enable_json_type=1
+${CLICKHOUSE_CLIENT} -q "SELECT toString(j) FROM t_json_roundtrip_103860;"
+${CLICKHOUSE_CLIENT} -q "DROP TABLE t_json_roundtrip_103860;"
+echo
+
+# --- Enum ---
+${CLICKHOUSE_CLIENT} -q "SELECT '--- Enum, SQL-style apostrophe escape ---' FORMAT LineAsString;"
+${CLICKHOUSE_CLIENT} -q "DROP TABLE IF EXISTS t_enum_roundtrip_103860; CREATE TABLE t_enum_roundtrip_103860 (e Enum8('a''b' = 1)) ENGINE = Memory;"
+printf "('a''b')" | ${CH_CLIENT_NO_FALLBACK} --query="INSERT INTO t_enum_roundtrip_103860 FORMAT Values"
+${CLICKHOUSE_CLIENT} -q "SELECT e FROM t_enum_roundtrip_103860;"
+${CLICKHOUSE_CLIENT} -q "DROP TABLE t_enum_roundtrip_103860;"
+echo
+
+# --- CustomSimpleText (Bool) ---
+# Types backed by `SerializationCustomSimpleText` (`Bool`, `Point`, `IPv4`, ...)
+# do not contain `'` in their textual form, so the asymmetry was not
+# user-observable for them. This sanity check ensures the switch to
+# `readQuotedStringWithSQLStyle` / `tryReadQuotedStringWithSQLStyle` did not
+# regress the common case.
+${CLICKHOUSE_CLIENT} -q "SELECT '--- Bool (CustomSimpleText) ---' FORMAT LineAsString;"
+${CLICKHOUSE_CLIENT} -q "DROP TABLE IF EXISTS t_bool_roundtrip_103860; CREATE TABLE t_bool_roundtrip_103860 (b Bool) ENGINE = Memory;"
+printf "('true')\n('false')" | ${CH_CLIENT_NO_FALLBACK} --query="INSERT INTO t_bool_roundtrip_103860 FORMAT Values"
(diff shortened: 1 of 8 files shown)


groeneai:
**Pre-PR validation gate (per project process)**

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. `SELECT cast('a\'b', 'Enum8(\'a\'\'b\' = 1)') FORMAT Values SETTINGS output_format_values_escape_quote_with_quote = 1` reliably emits `('a\'b')` on pre-fix master and `('a''b')` after the fix. |
| b | **Root cause explained?** | Yes. `SerializationEnum`, `SerializationJSON`, `SerializationAggregateFunction`, and `SerializationCustomSimpleText` overrides of `serializeTextQuoted` call `writeQuotedString` unconditionally and ignore `settings.values.escape_quote_with_quote`. PRs #60015 / #73519 added the branch in `SerializationString` / `SerializationFixedString` but the 4 sibling sites were missed. |
| c | **Fix matches root cause?** | Yes. Each site now branches on `settings.values.escape_quote_with_quote` and calls `writeQuotedStringPostgreSQL` in the true branch — same shape as the canonical `SerializationString::serializeTextQuoted`. |
| d | **Test intent preserved? / new tests added?** | New test `04253_sqlite_escape_quote_with_quote_103860.sh` covers `Enum`, `JSON`, `AggregateFunction` via `format Values` and includes a `String` sanity case. `AggregateFunction` uses `countSubstrings` on the quoted output (the binary state contains non-printable bytes) — counts `\'` vs `''` occurrences. |
| e | **Demonstrated in both directions?** | Yes. Captured pre-fix baseline output (all 3 serializations emit `\'` regardless of setting). After applying the fix, the test passes 50/50 with full settings + MergeTree settings randomization. |
| f | **Fix is general?** | Yes. Audited every `serializeTextQuoted` override in `src/DataTypes/Serializations/`. Wrapper serializations (Wrapper, Nullable, Sparse, Variant, Dynamic, LowCardinality, Replicated) delegate to nested. Remaining sites (Date, Date32, DateTime, DateTime64, Time, Time64, UUID, IPv4, IPv6, Interval, Bool) format values that cannot contain apostrophes (numeric/date-like). The 4 sites in this PR are the only ones with the bug. |

Session id: `cron:clickhouse-ci-task-worker:20260519-000000`

groeneai:
cc @azat @Avogar — could you review this? It propagates the canonical `output_format_values_escape_quote_with_quote` branch (PRs #60015 / #73519) from `SerializationString`/`SerializationFixedString` to the four sibling serializations (`Enum`, `JSON`, `AggregateFunction`, `CustomSimpleText`) that previously ignored the setting and broke `INSERT INTO sqlite_table SELECT enum_col` for enum values whose name contained a single quote. Fixes #103860.

groeneai:
Thanks for catching the asymmetry, @clickhouse-gh. Pushed `72355ed` to address both flagged sites.

**Fix:**
- `SerializationJSON::deserializeTextQuoted`: `readQuotedString` → `readQuotedStringWithSQLStyle`
- `SerializationCustomSimpleText::deserializeTextQuoted` and `tryDeserializeTextQuoted`: `readQuotedString` / `tryReadQuotedString` → `readQuotedStringWithSQLStyle` / `tryReadQuotedStringWithSQLStyle`

The new readers call the same `readAnyQuotedStringInto` helper with `enable_sql_style_quoting = true`, so they accept both `''` (SQL standard) and the legacy `\'` (backslash escape). Existing data keeps parsing unchanged — strictly more permissive.

Also verified that `SerializationEnum::deserializeTextQuoted` and `SerializationAggregateFunction::deserializeTextQuoted` already used the SQL-style reader before this PR, so no change is needed there. That covers all four serializations touched by this PR for symmetry.

**Regression test:** `04254_sqlite_escape_quote_with_quote_103860_roundtrip` exercises the deserializer path directly by disabling the `Values` parser's SQL-expression and template fallbacks (`input_format_values_interpret_expressions = 0`, `input_format_values_deduce_templates_of_expressions = 0`). Without those fallbacks disabled, the SQL parser would happily accept `''` in a string literal and mask the deserializer bug.

**Pre-PR validation gate:**

| # | Question | Answer |
|---|----------|--------|
| a | Deterministic repro? | Yes. `printf "('{\"k\":\"a''b\"}')" \| clickhouse-client --input_format_values_interpret_expressions=0 --input_format_values_deduce_templates_of_expressions=0 --query="INSERT INTO t_json FORMAT Values"` fails with `Code: 344 SUPPORT_IS_DISABLED` pre-fix, succeeds post-fix. |
| b | Root cause explained? | Yes. `readQuotedString` is `readAnyQuotedStringInto<'\'', enable_sql_style_quoting=false>` — it terminates on the first `'` and treats the second `'` of a `''` pair as the start of the next field. After PR #105285's emit-side change, the same serialization can now produce `''`, so the parse side must accept it. |
| c | Fix matches root cause? | Yes. Switching to the SQL-style readers directly fixes the `enable_sql_style_quoting=false` problem. No widening, no tag flips. |
| d | Test intent preserved / new tests added? | New sibling test `04254_..._roundtrip.sh` covers the deserializer path that 04253 didn't reach. 04253 still passes 5/5. |
| e | Both directions demonstrated? | Yes. Reverted the fix locally, rebuilt, observed `SUPPORT_IS_DISABLED` on JSON insert. Re-applied the fix, rebuilt, 5/5 runs pass on both tests. |
| f | Fix is general, not narrow? | Yes. Both sites the bot flagged are fixed. Verified the other two serializations in this PR already used the SQL-style reader so no further change is needed. |

session: cron:clickhouse-ci-task-worker:20260519-021500

tests/queries/0_stateless/04254_sqlite_escape_quote_with_quote_103860_roundtrip.sh
```diff
@@ -0,0 +1,80 @@
...
+# shellcheck source=../shell_config.sh
+. "$CUR_DIR"/../shell_config.sh
+
+CH_CLIENT_NO_FALLBACK="${CLICKHOUSE_CLIENT} --input_format_values_interpret_expressions=0 --input_format_values_deduce_templates_of_expressions=0"
+
+# --- JSON ---
+# Feed the SQL-style-escaped Values literal via stdin so neither bash nor the
+# outer SQL parser interferes with the apostrophes; the only parser that sees
+# them is `SerializationJSON::deserializeTextQuoted`.
+${CLICKHOUSE_CLIENT} -q "SELECT '--- JSON, SQL-style apostrophe escape (was broken pre-fix) ---' FORMAT LineAsString;"
```
azat, line 46: You can use simple `echo` instead

azat (approved):

groeneai:
Thanks for the approval and the merge, @azat! Followed up on the `echo` nit in #105293 — small cleanup that swaps every `printf` in `tests/queries/0_stateless/04254_sqlite_escape_quote_with_quote_103860_roundtrip.sh` for a simple `echo` (verified locally that all four cases — `JSON` SQL-style, `JSON` legacy backslash, `Enum8` SQL-style, `Bool` two-row insert — produce input bytes that `FORMAT Values` accepts identically).

session: cron:clickhouse-ci-task-worker:20260519-083000

clickgapai:

**Affected code:** `SerializationEnum::serializeTextQuoted`, `SerializationJSON::serializeTextQuoted`, `SerializationAggregateFunction::serializeTextQuoted` — introduced in 24.2.

**Why this might apply:** This fixes a user-visible bug in SQLite integration where INSERT fails for Enum/JSON/AggregateFunction columns containing single quotes. The `output_format_values_escape_quote_with_quote` setting was introduced in 24.2 (per SettingsChangesHistory.cpp:958), so the bug exists in all supported branches. The affected serialization code (SerializationEnum, SerializationJSON, SerializationAggregateFunction) predates all supported branches.

PR #105290, opened 2026-05-19
Fix ILLEGAL_COLUMN on GLOBAL IN with tuples and duplicate column names

### LINKED ISSUE #103228: `tryBuildAdditionalFilterAST` does not deduplicate column names when creating external table for GLOBAL IN with tuples

_Retrospective finding from a historical scan of [PR #93507](https://github.com/ClickHouse/ClickHouse/pull/93507) (merged 2026-01-07). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

GLOBAL IN with a tuple and a subquery that produces columns with duplicate names (e.g. `(x, y) GLOBAL IN (SELECT number, number FROM numbers(5))`) throws `ILLEGAL_COLUMN: Cannot add column number: column with this name already exists` when filter pushdown creates the external table for a remote query.

**Root cause:** ReadFromRemote.cpp:420-425: Unlike `buildQueryTreeForShard.cpp:429-430` which calls `makeUniqueColumnNamesInBlock` before creating `ColumnsDescription`, the filter pushdown path passes the raw sample block header directly, failing on duplicate column names.

**Why we believe this is a bug:** `ReadFromRemote.cpp:420` calls `InterpreterSelectQueryAnalyzer::getSampleBlock(source_ast, context)` which returns a Block with duplicate column names. Line 421 calls `header->getNamesAndTypesList()` preserving duplicates. Line 425 passes these to `ColumnsDescription{columns}` which throws because it prohibits duplicate column names.

**Affected locations:**
- `src/Processors/QueryPlan/ReadFromRemote.cpp:420` — tryBuildAdditionalFilterAST external table creation from sample block with potential duplicate column names

**Impact:** Any GLOBAL IN with a tuple subquery producing duplicate column names fails with ILLEGAL_COLUMN exception when querying a non-localhost remote shard with filter pushdown enabled (default). The same query works with `allow_push_predicate_ast_for_distributed_subqueries=0`, `prefer_localhost_replica=1` for localhost, or `enable_analyzer=0`.

Yes — confirmed on current `master` (commit `c646ee01481c`).

### How to reproduce

```sql
-- Test: GLOBAL IN with tuple where subquery produces duplicate column names
-- Covers: ReadFromRemote.cpp tryBuildAdditionalFilterAST external table creation

SET enable_analyzer=1;
SET enable_parallel_replicas=0;
SET prefer_localhost_replica=0;
SET allow_push_predicate_ast_for_distributed_subqueries=1;

DROP TABLE IF EXISTS t_gin_dup;
CREATE TABLE t_gin_dup (x UInt32, y UInt32) ENGINE = MergeTree ORDER BY x;
INSERT INTO t_gin_dup SELECT number, number FROM numbers(100);

-- Single column GLOBAL IN should work
SELECT sum(y) FROM (SELECT * FROM remote('127.0.0.1', currentDatabase(), t_gin_dup)) WHERE x GLOBAL IN (SELECT number FROM numbers(5)) ORDER BY 1;

-- Tuple GLOBAL IN with different column names should work
SELECT sum(y) FROM (SELECT * FROM remote('127.0.0.1', currentDatabase(), t_gin_dup)) WHERE (x, y) GLOBAL IN (SELECT toUInt32(number) as a, toUInt32(number) as b FROM numbers(5)) ORDER BY 1;

-- Tuple GLOBAL IN with duplicate column names from subquery
-- This fails with ILLEGAL_COLUMN when filter pushdown creates the external table
SELECT sum(y) FROM (SELECT * FROM remote('127.0.0.1', currentDatabase(), t_gin_dup)) WHERE (x, y) GLOBAL IN (SELECT number, number FROM numbers(5)) ORDER BY 1;

DROP TABLE t_gin_dup;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/301f8135-ea3f-43b2-884c-80ab553f0b52)

### Expected behavior

```
10
10
10
```

```
10
10
Code: 44. DB::Exception: Cannot add column number: column with this name already exists. (ILLEGAL_COLUMN)
```

### Additional context

**Open risks:**
- The same pattern could affect other code paths that create temporary tables from subquery headers without deduplicating column names

**Suggested fix:** Before creating `ColumnsDescription`, call `makeUniqueColumnNamesInBlock` on the header Block (like `buildQueryTreeForShard.cpp:430` does), then use the deduplicated header's `getNamesAndTypesList()` for the `ColumnsDescription`.

Found during automated review of [PR #93507](https://github.com/ClickHouse/ClickHouse/pull/93507).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr93507_001`_

groeneai:
Closes #103228.

`tryBuildAdditionalFilterAST` in `ReadFromRemote.cpp` builds an external
table for a `GLOBAL IN` set from the subquery's sample block. When the
subquery projects columns with duplicate names (e.g.
`(x, y) GLOBAL IN (SELECT number, number FROM ...)`) the path passed the
raw `NamesAndTypesList` to `ColumnsDescription`, which prohibits duplicate
names. The query then failed on the initiator with
`Code: 44. DB::Exception: Cannot add column number: column with this name
already exists. (ILLEGAL_COLUMN)`.

The sibling new-analyzer path in `buildQueryTreeForShard.cpp` already
calls `makeUniqueColumnNamesInBlock` before constructing
`ColumnsDescription` (added in commit `326a3a3e8d71`). Apply the same step
in the filter-pushdown path. Column matching when `MemorySink::consume`
populates the temporary table (it uses the storage's header, mapping by
position via `cloneWithColumns`) and when the remote shard later evaluates
`IN (_data_<hash>)` is by position, so renaming the columns is safe.

Reproducer:

```sql
SET allow_push_predicate_ast_for_distributed_subqueries = 1,
    prefer_localhost_replica = 0;
CREATE TABLE t (x UInt32, y UInt32) ENGINE = MergeTree ORDER BY x;
INSERT INTO t SELECT number, number FROM numbers(100);
SELECT sum(y) FROM (SELECT * FROM remote('127.0.0.1', currentDatabase(), t))
WHERE (x, y) GLOBAL IN (SELECT number, number FROM numbers(5));
-- before: Code: 44 ILLEGAL_COLUMN
-- after:  10
```

Verified locally: the regression test fails on master at
`c646ee01481c`/`3c264823d120` and passes with the fix; 50/50 runs with CI
randomization enabled.

Fix `Code: 44. ILLEGAL_COLUMN: Cannot add column ...: column with this name already exists` thrown by distributed queries with `allow_push_predicate_ast_for_distributed_subqueries = 1` (the default) when a `GLOBAL IN` tuple is matched against a subquery whose projection contains duplicate column names, e.g. `(x, y) GLOBAL IN (SELECT number, number FROM numbers(5))`.

### Tests
- New: `04253_global_in_tuple_duplicate_columns.sql` — covers tuple `GLOBAL IN` with duplicate column names (the reproducer), with distinct names (regression-free baseline), three-element tuple with duplicates, and a no-match correctness check.

cc @azat (per https://github.com/ClickHouse/ClickHouse/issues/103228#issuecomment-4482055158)

### Version info
- Merged into: `26.5.1.877`


groeneai:
Pre-PR validation gate (TASK.md Phase 4 Step 9) — session `cron:clickhouse-ci-task-worker:20260519-034500`:

**(a) Deterministic repro?** ✅ — `SELECT sum(y) FROM (SELECT * FROM remote('127.0.0.1', currentDatabase(), t)) WHERE (x, y) GLOBAL IN (SELECT number, number FROM numbers(5))` with `allow_push_predicate_ast_for_distributed_subqueries = 1, prefer_localhost_replica = 0` throws `Code: 44. ILLEGAL_COLUMN: Cannot add column number: column with this name already exists` every time on master (`08bc397bca69`/`3c264823d120`).

**(b) Root cause explained?** ✅ — `ReadFromRemote.cpp:420-421` (before the fix) calls `InterpreterSelectQueryAnalyzer::getSampleBlock(source_ast, context)` and feeds the raw `getNamesAndTypesList()` into `ColumnsDescription{columns}`. `ColumnsDescription` prohibits duplicate column names, so a subquery projecting `number, number` makes the constructor throw. The sibling new-analyzer path in `buildQueryTreeForShard.cpp:428-429` already deduplicates via `makeUniqueColumnNamesInBlock`; the filter-pushdown path was missed (sibling fix was `326a3a3e8d71`).

**(c) Fix matches root cause?** ✅ — call `makeUniqueColumnNamesInBlock` on a local `Block` copy of the sample header before extracting `NamesAndTypesList`. Same pattern as the sibling path. No band-aid — the throw site is fixed at the source. Safety: `MemorySink::consume` (`StorageMemory.cpp:99`) does `getHeader().cloneWithColumns(chunk.getColumns())`, mapping the source-pipeline columns into the temp table's deduplicated header by position; on the remote shard, `IN (_data_<hash>)` matches the LHS tuple against the temp table by position+type, not by name.

**(d) Test intent preserved / new tests added?** ✅ — new `04253_global_in_tuple_duplicate_columns.sql`. Covers (1) baseline single-column GLOBAL IN, (2) tuple with distinct names, (3) the reproducer (tuple, duplicate names), (4) three-element tuple with duplicates, (5) no-match correctness check returning `0` so a silently-dropped predicate would be caught.

**(e) Both directions demonstrated?** ✅ — pre-fix binary: `ILLEGAL_COLUMN`. Post-fix binary on same data: `10, 10, 10, 10, 0` matching `.reference`. 50/50 passes with CI randomization on.

**(f) Fix is general?** ✅ — audited every `create_for_global_subquery=true` site:
- `src/Processors/QueryPlan/ReadFromRemote.cpp:420` — this fix
- `src/Storages/buildQueryTreeForShard.cpp:450` — already fixed (sibling that inspired this patch)
- `src/Interpreters/GlobalSubqueriesVisitor.cpp:131` — old-analyzer GLOBAL IN; verified locally that the same query with `enable_analyzer = 0` returns `10`, so the old-analyzer path does not hit this `ColumnsDescription` constructor with duplicate names (it goes through `interpreter->getSampleBlock()` which dedupes upstream). Not user-visible.
- `src/Analyzer/Resolve/QueryAnalyzer.cpp:3095, 5347, 5904` and `src/Processors/Sources/RecursiveCTESource.cpp:123` — materialized-CTE / recursive-CTE temporary tables, columns sourced from `QueryNode::getProjectionColumns()` (resolved aliases, not the raw subquery projection). Not reachable from this bug.

groeneai:
cc @azat @KochetovNicolai — could you review? Targeted fix for #103228: `tryBuildAdditionalFilterAST` in `ReadFromRemote.cpp` was missing the `makeUniqueColumnNamesInBlock` call that the sibling new-analyzer path in `buildQueryTreeForShard.cpp` already performs, so `GLOBAL IN` tuples with duplicate-named subquery projections were throwing `ILLEGAL_COLUMN` on the initiator.

groeneai:
Thanks for catching the suffix-collision gap, @clickhouse-gh. The previous helper would have left blocks like `a, a, a_1` with two `a_1` columns — your trace is exactly right, and `ColumnsDescription` rejects that with `ILLEGAL_COLUMN`. Pushed `68bf893c7a75` with the fix at the helper level so all three callers (`ReadFromRemote.cpp`, `buildQueryTreeForShard.cpp`, `evaluateScalarSubqueryIfNeeded.cpp`) benefit. All three use `MatchColumnsMode::Position`, so the side-effect renaming of an already-unique `a_1` to `a_2` when the prefix collides is safe.

### Helper change (`src/Analyzer/Utils.cpp`)

```cpp
void makeUniqueColumnNamesInBlock(Block & block)
{
    NameSet block_column_names;
    size_t unique_column_name_counter = 1;

    for (auto & column_with_type : block)
    {
        if (block_column_names.insert(column_with_type.name).second)
            continue;

        /// Loop until we find a suffix that is unused anywhere in the block,
        /// including by names we are about to keep and by names we have already
        /// renamed. Register the renamed name to prevent further collisions.
        String new_name;
        do
        {
            new_name = column_with_type.name + '_' + std::to_string(unique_column_name_counter);
            ++unique_column_name_counter;
        } while (!block_column_names.insert(new_name).second);

        column_with_type.name = std::move(new_name);
    }
}
```

`insert(name).second` does the contains-and-add atomically; the inner `do-while` loops the suffix until the renamed name is actually unused.

### Regression test extended with explicit suffix-collision shapes

`tests/queries/0_stateless/04253_global_in_tuple_duplicate_columns.sql` now also exercises:

```sql
-- (a, a, a_1): second `a` must become `a_1`, third (`a_1`) must become `a_2`.
SELECT sum(y) FROM (SELECT * FROM remote('127.0.0.1', currentDatabase(), t_gin_dup))
WHERE (x, x, y) GLOBAL IN (
    SELECT toUInt32(number) AS a, toUInt32(number) AS a, toUInt32(number) AS a_1 FROM numbers(5)
);

-- (a, a, a, a_1, a_2): collision-safe renaming must skip pre-existing
-- a_1/a_2 and emit a_3/a_4 instead.
SELECT sum(y) FROM (SELECT * FROM remote('127.0.0.1', currentDatabase(), t_gin_dup))
WHERE (x, x, x, y, y) GLOBAL IN (
    SELECT
        toUInt32(number) AS a,
        toUInt32(number) AS a,
        toUInt32(number) AS a,
        toUInt32(number) AS a_1,
        toUInt32(number) AS a_2
    FROM numbers(5)
);
```

### Pre-PR validation gate

| # | Question | Answer |
|---|---|---|
| a | Deterministic repro? | Yes — `(x, x, y) GLOBAL IN (SELECT n AS a, n AS a, n AS a_1 FROM ...)` deterministically throws `Cannot add column a_1: column with this name already exists` on the previous commit. |
| b | Root cause explained? | Helper renamed dup `a` → `a_1` but never inserted the produced name into the seen-set, so the third (original) `a_1` slipped past the `contains()` check and the block ended with two `a_1`. |
| c | Fix matches root cause? | Yes — loops the suffix until a truly unused name is found and registers every emitted name. |
| d | Test intent preserved? | Yes — all previous cases (`number, number`, three-way duplicate, empty-set correctness) still pass; two new cases cover the suffix-collision class explicitly. |
| e | Both directions demonstrated? | Yes — buggy helper fails on the first new case with the exact predicted error; fixed helper passes 20/20 runs with randomization. |
| f | Fix is general? | Yes — fixed at the helper, all three callers (`ReadFromRemote.cpp`, `buildQueryTreeForShard.cpp`, `evaluateScalarSubqueryIfNeeded.cpp`) benefit; callers all use positional matching so the side-effect rename is safe. |

Session: `cron:clickhouse-ci-task-worker:20260519-050000`.

azat (approved):

azat:
>[PR / Stress test (amd_msan) (pull_request)](https://github.com/ClickHouse/ClickHouse/actions/runs/26077518826/job/76678211784?pr=105290)Failing after 41m

NULL deref in TimeSeries, fixed already

PR #105293, opened 2026-05-19
Use simple echo in 04254 SQLite escape quote roundtrip test

groeneai:
Follow-up to merged PR #105285. Applies @azat's stylistic review nit: replace `printf` with simple `echo` in `tests/queries/0_stateless/04254_sqlite_escape_quote_with_quote_103860_roundtrip.sh`.

Review thread: https://github.com/ClickHouse/ClickHouse/pull/105285#pullrequestreview-azat-nit (the inline comment on the test script suggested "You can use simple `echo` instead").

Behavior is unchanged. For `FORMAT Values`:
- The only delivered-byte difference per call is a trailing newline that `echo` always appends and `printf "..."` (no trailing `\n` in the format string) does not. `Values` ignores trailing whitespace.
- In the `Bool` case, the row separator goes from `\n` (newline) to `,` (comma). Both forms are accepted by the `Values` parser as row separators.

The backslash JSON case `("k":"a\\'b")` also drops two backslashes from the shell-source: `printf` re-interprets `\\` as `\`, while `echo` (without `-e`) emits the bytes bash hands it, so one fewer level of escaping is needed.

**Verified locally** with a running ClickHouse server (debug build):
- `echo "('a''b')"` -> `Enum8` inserts `a\'b` (matches reference)
- `echo "('{\"k\":\"a\\'b\"}')"` -> `JSON` inserts `{"k":"a\'b"}` (matches reference)
- `echo "('true'),('false')"` -> `Bool` inserts both rows (matches reference)

The SQL-style JSON case requires the deserializer fix that #105285 itself shipped, so it will be verified by CI on top of merged master.

...

### DIFF
--- a/tests/queries/0_stateless/04254_sqlite_escape_quote_with_quote_103860_roundtrip.sh
+++ b/tests/queries/0_stateless/04254_sqlite_escape_quote_with_quote_103860_roundtrip.sh
@@ -45,7 +45,7 @@ CH_CLIENT_NO_FALLBACK="${CLICKHOUSE_CLIENT} --input_format_values_interpret_expr
 # them is `SerializationJSON::deserializeTextQuoted`.
 ${CLICKHOUSE_CLIENT} -q "SELECT '--- JSON, SQL-style apostrophe escape (was broken pre-fix) ---' FORMAT LineAsString;"
 ${CLICKHOUSE_CLIENT} -q "DROP TABLE IF EXISTS t_json_roundtrip_103860; CREATE TABLE t_json_roundtrip_103860 (j JSON) ENGINE = Memory SETTINGS enable_json_type = 1;"
-printf "('{\"k\":\"a''b\"}')" | ${CH_CLIENT_NO_FALLBACK} --query="INSERT INTO t_json_roundtrip_103860 FORMAT Values" --enable_json_type=1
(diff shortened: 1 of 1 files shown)


groeneai:
cc @azat — this is the small follow-up to your inline nit on #105285. Only swaps `printf` for `echo` in the regression test added by #105285; no behavior change. Verified locally that the four `FORMAT Values` insertion paths (`JSON` SQL-style, `JSON` legacy backslash, `Enum8` SQL-style, `Bool` two-row) accept the `echo` output identically.

session: cron:clickhouse-ci-task-worker:20260519-083000

Algunenano (approved):

azat (commented):
@groeneai 
- I meant different thing, there is `clickhouse-client "SELECT '--- JSON, ..."`, that is basically `echo`
- And I wrote it just for you to memorize, not to send a PR, I usually ask explicitly for this

groeneai:
Thanks for the clarification, @azat. Got it on both counts:

1. The pattern you actually suggested is `clickhouse-client "SELECT '--- JSON, ...'"` (use ClickHouse to generate the SELECT-Values output), not a plain shell `echo`.
2. Stylistic nits on a review are for memorizing for future PRs, not for filing a separate follow-up. I'll do that going forward and won't open a standalone PR for this kind of suggestion unless explicitly asked.

Closing this. Saved the rule for next time.

PR #105296, opened 2026-05-19
Add note about distributed cluster to WASM UDF docs

vdimir:

...

### Version info
- Merged into: `26.5.1.873`

### DIFF
--- a/docs/en/sql-reference/functions/wasm_udf.md
+++ b/docs/en/sql-reference/functions/wasm_udf.md
@@ -134,6 +134,39 @@ SELECT 'my_module', base64Decode('...'), reinterpretAsUInt256(unhex('369f...c57d
 
 If the provided hash does not match the computed SHA256 of the module code, the insertion fails. It may be useful when loading modules from external sources such as S3 or HTTP.
 
+### Distribute a module across a cluster {#distribute-a-module-across-a-cluster}
+
+`system.webassembly_modules` is a per-instance table — an `INSERT` lands only on the replica handling the connection. There is no `ON CLUSTER` form of the `INSERT` statement, so a subsequent `CREATE FUNCTION ... ON CLUSTER` will fail on replicas that do not have the module:
+
+```text
+Code: 674. DB::Exception: WebAssembly module 'collatz' not found:
+while adding user defined function `collatz_steps`. (RESOURCE_NOT_FOUND)
+```
+
+To fan an insert out to every node, write to the `cluster` (or `clusterAllReplicas`) table function instead of the local `system.webassembly_modules` table:
+
+```bash
+cat collatz.wasm | clickhouse client -q "
+  INSERT INTO FUNCTION cluster('default', 'system', 'webassembly_modules') (name, code)
+  SELECT 'collatz', code FROM input('code String') FORMAT RawBlob"
+```
+
+After the fanned-out insert, the module is present on every replica and `CREATE FUNCTION ... ON CLUSTER` succeeds:
+
+```sql
+CREATE FUNCTION collatz_steps ON CLUSTER 'default'
+LANGUAGE WASM FROM 'collatz' :: 'steps'
+ARGUMENTS (n UInt32) RETURNS UInt32;
+```
+
+You can verify the module is loaded everywhere with `clusterAllReplicas`:
+
+```sql
+SELECT hostName(), name FROM clusterAllReplicas('default', system.webassembly_modules) WHERE name = 'collatz';
+```
+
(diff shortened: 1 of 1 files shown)


docs/en/sql-reference/functions/wasm_udf.md
```diff
@@ -134,6 +134,39 @@ SELECT 'my_module', base64Decode('...'), reinterpretAsUInt256(unhex('369f...c57d
...
+### Distribute a module across a cluster {#distribute-a-module-across-a-cluster}
+
+`system.webassembly_modules` is a per-instance table — an `INSERT` lands only on the replica handling the connection. There is no `ON CLUSTER` form of the `INSERT` statement, so a subsequent `CREATE FUNCTION ... ON CLUSTER` will fail on replicas that do not have the module:
+
+```text
+Code: 674. DB::Exception: WebAssembly module 'collatz' not found:
+while adding user defined function `collatz_steps`. (RESOURCE_NOT_FOUND)
+```
+
+To fan an insert out to every node, write to the `cluster` (or `clusterAllReplicas`) table function instead of the local `system.webassembly_modules` table:
```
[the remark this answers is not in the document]
Algunenano: I agree with the AI bot. Let's simply recommend `clusterAllReplicas` only (and use it in the example) since that works on every replica, no matter the config

docs/en/sql-reference/functions/wasm_udf.md
```diff
@@ -134,6 +134,45 @@ SELECT 'my_module', base64Decode('...'), reinterpretAsUInt256(unhex('369f...c57d
...
+To fan an insert out to every node, write to the `cluster` table function instead of the local `system.webassembly_modules` table:
+
+```bash
+cat collatz.wasm | clickhouse client -q "
+  INSERT INTO FUNCTION cluster('default', 'system', 'webassembly_modules') (name, code)
+  SELECT 'collatz', code FROM input('code String') FORMAT RawBlob"
+```
+
+:::note
+This pattern relies on the underlying distributed-write path visiting every replica within each shard, which only happens when the cluster is configured with `internal_replication=false`. With `internal_replication=true` (the default for clusters that use `ReplicatedMergeTree` to drive replication themselves), the insert is delivered to a single healthy replica per shard, and `system.webassembly_modules` is not replicated by that path — so some replicas will still be missing the module. In that configuration you need to insert against each replica individually, for example by iterating over `system.clusters` and writing via `remote(...)` per host, or by copying the binary into `user_scripts/wasm/` on every host.
```
Algunenano, line 155: Why not recommend clusterAllReplicas instead and forget about internal_replication and custom setups?

vdimir: I tried it, but not sure if this supposed to work:

```SQL

:) INSERT INTO FUNCTION clusterAllReplicas('default', 'system', 'webassembly_modules') (name, code) SELECT
    'wasm_aaaaa',
    base64Decode(concat('AGFzbQEAAAABFgNgAAF/YAJ/fwN/f39gAn9/BH9/f38DBgUAAQECAgdQBQ5yZXR1cm5fZGVmYXVs', 'dAAADnVubmFtZWRfbG9jYWxzAAEObmFtZWRfYnlfaW5kZXgAAgxzb21lX3VubmFtZWQAAwptdWx0', 'aV9kZWNsAAQKUgUGAQF/IAALDgEBf0EsIQIgACABIAILDwEBf0HYACECIAAgASACCxUBAn9BzQAh', 'AkEsIQMgACABIAIgAwsUAQJ/QSEhAkEsIQMgACABIAIgAws='))

Elapsed: 0.002 sec. 

Code: 55. DB::Exception: Received from localhost:9001. DB::Exception: Method write is not supported by storage Distributed with more than one shard and no sharding key provided. (STORAGE_REQUIRES_PARAMETER)

:) SELECT cluster, shard_num, internal_replication FROM system.clusters where name == 'default';

   ┌─cluster─┬─shard_num─┬─internal_replication─┐
1. │ default │         1 │                    0 │
2. │ default │         1 │                    0 │
3. │ default │         1 │                    0 │
   └─────────┴───────────┴──────────────────────┘

3 rows in set. Elapsed: 0.001 sec. 

```

Algunenano: Welp. Sorry then. I was pretty confident clusterAllReplicas worked (I was expecting it)

Algunenano (approved):

docs/en/sql-reference/functions/wasm_udf.md
EmeraldShift: I was also expecting it. Is there a way this can be fixed?

PR #105286, opened 2026-05-19
Fix `Memory` table abort on `ALTER TABLE ... APPLY PATCHES` (STID 2380-3043)

groeneai:
Fixes `STID: 2380-3043` (`Logical error: Mutation of \`Memory\` table produced incomplete output: block 0 has 0 rows, expected N`), reported by @alexey-milovidov on https://github.com/ClickHouse/ClickHouse/pull/99972 as a chronic `BuzzHouse` finding hitting master and 8+ unrelated PRs (`#99804`, `#100185`, `#100586`, `#104606`, `#105107`, `#105161`, `#105233`).

CI report demonstrating the original failure:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=99972&sha=f7959f604f297d728373e5f6ed753d9a0bde5a3c&name_0=PR&name_1=BuzzHouse%20%28amd_tsan%29

### Reproducer (master HEAD, deterministic)

```sql
CREATE TABLE mem (id UInt64, value String) ENGINE = Memory;
INSERT INTO mem VALUES (1, 'a'), (2, 'b');
INSERT INTO mem VALUES (3, 'c'), (4, 'd');
ALTER TABLE mem (APPLY PATCHES);
-- Logical error: 'Mutation of `Memory` table produced incomplete output:
-- block 0 has 0 rows, expected 2'.
```

### Root cause

`MutationsInterpreter::prepare` short-circuits `APPLY_PATCHES` /
`APPLY_DELETED_MASK` without adding a mutation stage
(`MutationsInterpreter.cpp:1238-1243`). The resulting pipeline still pulls one
output block per input block, but each block has zero rows. `StorageMemory::mutate`
then walks the partial-column path and fails the per-block row-count invariant
at `StorageMemory.cpp:317`.

### Fix

`Memory` tables do not own patch parts (lightweight updates are a `MergeTree`
feature — `IStorage::supportsLightweightUpdate` returns
`"Table with engine Memory doesn't support lightweight updates"`) and do not
maintain a `_row_exists` deletion mask. Both commands are therefore semantic
no-ops on `Memory`: the post-state the user is asking for (all patches applied,
all deleted rows physically removed) is already satisfied.

`StorageMemory::mutate` now filters `APPLY_PATCHES` / `APPLY_DELETED_MASK` out
of the command set before constructing the mutation pipeline. If all commands
are filtered out, it returns immediately. A real `UPDATE` or `DELETE` mixed
into the same `ALTER` still executes normally — exercised by the regression
[...]

PR #105298, opened 2026-05-19
CI: fix Bugfix validation after stop-testing rename

leshikus:
#104081 renamed the `server_died` event to `stop_testing` in `clickhouse-test` and replaced the parent-side print

```
Server died, terminating all processes...
```

with

```
Stopping tests, terminating all processes...
```

The functional-tests results parser in `ci/jobs/scripts/functional_tests_results.py` still searched for the old string, so when a Bugfix validation run actually killed the server it never emitted the synthetic `Server died` leaf that the inversion step in `ci/jobs/functional_tests.py` flips to OK — and the job reported `Failed to reproduce the bug` even though the bug had reproduced.

This PR realigns the two SIGN constants to the wording the runner now emits and renames the carried variable / `Summary` field to `stop_testing` to match.

Observed on the `Bugfix validation (functional tests)` job of #105286 (server crashed under the test; log line `Stopping tests, terminating all processes...`).

Refs #104081.

...

### Version info
- Merged into: `26.5.1.856`

### DIFF
--- a/ci/jobs/scripts/functional_tests_results.py
+++ b/ci/jobs/scripts/functional_tests_results.py
@@ -12,7 +12,7 @@ SKIPPED_SIGN = "[ SKIPPED "
 NOT_FAILED_SIGN = "[ NOT_FAILED "
 HUNG_SIGN = "Found hung queries in processlist"
 STOP_TESTING_SIGN = "Stopping tests, terminating all processes"
-STOP_TESTING_SIGN2 = "Hung check failed: server is not responding"
+STOP_TESTING_SIGN2 = "Server does not respond to health check"
 DATABASE_SIGN = "Database: "
 
 SUCCESS_FINISH_SIGNS = ["All tests have finished", "No tests were run"]
--- a/ci/jobs/scripts/functional_tests_results.py
+++ b/ci/jobs/scripts/functional_tests_results.py
@@ -11,8 +11,8 @@ UNKNOWN_SIGN = "[ UNKNOWN "
 SKIPPED_SIGN = "[ SKIPPED "
 NOT_FAILED_SIGN = "[ NOT_FAILED "
 HUNG_SIGN = "Found hung queries in processlist"
-SERVER_DIED_SIGN = "Server died, terminating all processes"
-SERVER_DIED_SIGN2 = "Server does not respond to health check"
+STOP_TESTING_SIGN = "Stopping tests, terminating all processes"
+STOP_TESTING_SIGN2 = "Hung check failed: server is not responding"
 DATABASE_SIGN = "Database: "
 
 SUCCESS_FINISH_SIGNS = ["All tests have finished", "No tests were run"]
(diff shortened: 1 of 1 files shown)


ci/jobs/scripts/functional_tests_results.py
```diff
@@ -11,8 +11,8 @@
 SKIPPED_SIGN = "[ SKIPPED "
 NOT_FAILED_SIGN = "[ NOT_FAILED "
 HUNG_SIGN = "Found hung queries in processlist"
-SERVER_DIED_SIGN = "Server died, terminating all processes"
-SERVER_DIED_SIGN2 = "Server does not respond to health check"
```
novikd, line 15: `clickhouse-test` still produces `Server does not respond to health check` error

leshikus: fixed, thanks

ci/jobs/scripts/functional_tests_results.py
```diff
@@ -11,8 +11,8 @@
...
+STOP_TESTING_SIGN = "Stopping tests, terminating all processes"
+STOP_TESTING_SIGN2 = "Hung check failed: server is not responding"
```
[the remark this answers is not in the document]
leshikus: fixed

maxknv (approved):

ci/jobs/scripts/functional_tests_results.py
```diff
@@ -11,8 +11,8 @@
 SKIPPED_SIGN = "[ SKIPPED "
 NOT_FAILED_SIGN = "[ NOT_FAILED "
 HUNG_SIGN = "Found hung queries in processlist"
-SERVER_DIED_SIGN = "Server died, terminating all processes"
-SERVER_DIED_SIGN2 = "Server does not respond to health check"
+STOP_TESTING_SIGN = "Stopping tests, terminating all processes"
```
Algunenano, line 14: We should put this in a common file and use them in clickhouse-test. Otherwise this will happen again

azat: +1

And also this looks fragile, maybe we need to use constants from `clickhouse-test` directly here?

Also can't bugfix simply consider any error as the bug is reproduced?

leshikus: I will fix this with the next PR. Let me know what do you think is better - having a clear automation marker in the report text, e.g. `STOP_TESTING` or `###STOP_TESTING###`; or maybe just using exit code 2 for `clickhouse-test` to signal this situation. @azat @Algunenano

leshikus: The point why I want to commit this now is this affects some of CI checks

azat: >exit code 2 for clickhouse-test to signal this situation

I like this one

Another option is to add proper reports into `clickhouse-test` that will be more or standardized and will not require that complex parsing

PR #105318, opened 2026-05-19
Fix replica delay check bypass when Distributed query targets a View

### LINKED ISSUE #105304: Distributed over View -> ReplicatedMergeTree bypasses stale replica checks
### Company or project name

Cloudflare

### Describe what's wrong

**What's wrong**
When a Distributed table is queried and the resolved local table is a View or MaterializedView over a ReplicatedMergeTree, the stale-replica delay check is silently skipped. As a result, queries that should fail (with `ALL_REPLICAS_ARE_STALE` or fall back to another replica) instead return potentially stale data, even when the user has `max_replica_delay_for_distributed_queries` set.

**Root cause**
Both the local path `SelectStreamFactory::createForShardImpl` and the remote path `TCPHandler::processTablesStatusRequest` compute replica delay via a direct check for a ReplicatedMergeTree. When the Distributed table points at a View, the check is false, the delay is reported as zero, and the replica is treated as fully up-to-date regardless of the actual delay of the underlying ReplicatedMergeTree.

Yes

### How to reproduce

Create two ReplicatedMergeTree replicas sharing one ZooKeeper path, a View over the second replica, and two Distributed tables, with one pointing directly at the replica, one pointing at the view, like so:

```sql
CREATE TABLE replicated1 (n UInt64)
    ENGINE = ReplicatedMergeTree('/clickhouse/tables/{database}/stale_delay_test', 'r1')
    ORDER BY n;

CREATE TABLE replicated2 (n UInt64)
    ENGINE = ReplicatedMergeTree('/clickhouse/tables/{database}/stale_delay_test', 'r2')
    ORDER BY n;

CREATE VIEW view_replicated2 AS SELECT * FROM replicated2;

CREATE TABLE distributed_replicated2 AS replicated2
    ENGINE = Distributed('test_shard_localhost', currentDatabase(), 'replicated2');

CREATE TABLE distributed_view_replicated2 AS replicated2
    ENGINE = Distributed('test_shard_localhost', currentDatabase(), 'view_replicated2');
```

Now we need to make `replicated2` stale. Pause replication on `replicated2`, write to `replicated1`, and pull just the log entry so that `absolute_delay` rises without the data actually being fetched:

```sql
SYSTEM STOP REPLICATION QUEUES replicated2;
INSERT INTO replicated1 VALUES (1), (2), (3);
SYSTEM SYNC REPLICA replicated2 PULL;
```

Wait until the delay is observable:

```sql
SELECT absolute_delay FROM system.replicas
WHERE database = currentDatabase() AND table = 'replicated2'; -- absolute_delay >= 2
```

Now query both Distributed tables:

```sql
SET prefer_localhost_replica = 1;  -- 0 gives equivalent result through a different execution path
SET max_replica_delay_for_distributed_queries = 1;

-- Direct Distributed over ReplicatedMergeTree fails (correct):
SELECT * FROM distributed_replicated2;
-- Server Error ALL_REPLICAS_ARE_STALE

-- Distributed over a View wrapping the same ReplicatedMergeTree returns result normally without raising issue or logging warning:
SELECT count() FROM distributed_view_replicated2;
-- Returns a result built from the stale local replica instead of raising ALL_REPLICAS_ARE_STALE
```

### Expected behavior

Considering the case where a Distributed engine points to a View that reads from a ReplicatedMergeTree, we would expect that replication delay is checked normally for the ReplicatedMergeTree that is being used as a source of data. However, we verified that this was not happening and that we were serving stale data without any error or warning that we were doing so.

None.

### Additional context

The usual way to use Distributed engine is to have it directly on top of ReplicatedMergeTree, and then use the View on top of the Distributed engine. We used to do this at Cloudflare, until we noticed that a simple query that should execute rather quickly was taking a very long time. After trying to switch the order to Distributed -> View -> ReplicatedMergeTree, we saw a 10x reduction in query execution time. We also see that [#101791](https://github.com/ClickHouse/ClickHouse/pull/101791) would fix this, but at the time the quick and safe solution was just to swap the Distributed to be on top of View.

cc @seandhaynes @c-end

marco-vb:
Fixes #105304.

cc @seandhaynes @c-end

Previously both the local path (`SelectStreamFactory::createForShardImpl`) and the remote path (`TCPHandler::processTablesStatusRequest`) only checked replication delay when the resolved table was directly a StorageReplicatedMergeTree. If the Distributed table targeted a View or MaterializedView wrapping a ReplicatedMergeTree, the dynamic_cast failed and the query would serve stale data without raising `ALL_REPLICAS_ARE_STALE`, even when the user explicitly set `max_replica_delay_for_distributed_queries` and `fallback_to_stale_replicas_for_distributed_queries`.

Extract a shared helper (`getReplicatedDelay`) that traverses the table's referential dependencies via `DatabaseCatalog` to discover underlying ReplicatedMergeTree tables, returning the maximum `absolute_delay` and aggregated readonly status. Wire it into both the local and remote paths so that Distributed over a View behaves identically to Distributed directly over a replicated table with respect to stale-replica checks.

A stateless test (`04256_stale_replica_view_distributed_delay_check.sh`) reproduces the bug across both paths.

Fix Distributed queries returning stale data when the target table is a View or MaterializedView over a ReplicatedMergeTree. The stale-replica delay check (`max_replica_delay_for_distributed_queries`, `fallback_to_stale_replicas_for_distributed_queries`) is now applied to the underlying replicated tables instead of being silently skipped.

### DIFF
--- a/src/Interpreters/ClusterProxy/SelectStreamFactory.cpp
+++ b/src/Interpreters/ClusterProxy/SelectStreamFactory.cpp
@@ -5,6 +5,7 @@
 #include <Interpreters/ClusterProxy/SelectStreamFactory.h>
 #include <Interpreters/DatabaseCatalog.h>
 #include <Interpreters/InterpreterSelectQueryAnalyzer.h>
+#include <Interpreters/ReplicatedTableDelay.h>
 #include <Interpreters/SelectQueryOptions.h>
 #include <Interpreters/TranslateQualifiedNamesVisitor.h>
 #include <Parsers/ASTSelectQuery.h>
@@ -252,15 +253,6 @@ void SelectStreamFactory::createForShardImpl(
             return;
         }
 
-        const auto * replicated_storage = dynamic_cast<const StorageReplicatedMergeTree *>(main_table_storage.get());
-
-        if (!replicated_storage)
-        {
-            /// Table is not replicated, use local server.
-            emplace_local_stream();
-            return;
-        }
-
         const UInt64 max_allowed_delay = settings[Setting::max_replica_delay_for_distributed_queries];
 
         if (!max_allowed_delay)
@@ -269,7 +261,7 @@ void SelectStreamFactory::createForShardImpl(
             return;
         }
 
-        UInt64 local_delay = replicated_storage->getAbsoluteDelay();
+        UInt64 local_delay = getReplicatedDelay(main_table_storage, context).max_absolute_delay;
 
         if (local_delay < max_allowed_delay)
         {
--- a/src/Interpreters/ReplicatedTableDelay.cpp
+++ b/src/Interpreters/ReplicatedTableDelay.cpp
@@ -1,0 +1,59 @@
+#include <algorithm>
+#include <queue>
+
+#include <Common/logger_useful.h>
+#include <Interpreters/DatabaseCatalog.h>
+#include <Interpreters/ReplicatedTableDelay.h>
+#include <Storages/MergeTree/MergeTreeData.h>
+#include <Storages/StorageReplicatedMergeTree.h>
+
+namespace DB
+{
+
+/// Traverse referential dependencies of a storage (e.g. View -> ReplicatedMergeTree)
+/// and aggregate delay/readonly info across all discovered ReplicatedMergeTree tables.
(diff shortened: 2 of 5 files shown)


victor-perov:
Would be great to get some traction on this fix!

PR #105319, opened 2026-05-19
Fix UBSan signed-integer overflow in `timeSeries*ToGrid` window check

groeneai:
## Summary

Adversarial AST-fuzzer queries that set the `window` (staleness) parameter to a
value near `INT64_MAX` triggered UBSan signed-integer overflow in the
sliding-window cutoff `samples_in_window.front().first + Base::window <=
current_timestamp` inside `AggregateFunctionTimeseriesLinearRegression.h`, and
in four sibling files that share the same pattern.

Three master sightings of this exact bug pattern were observed in the last 30 days:

- **STID 4701-4d85** in `AggregateFunctionTimeseriesLinearRegression.h:250`
  ([CI report](https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=d3861de5198e7e64bafba0aec6c3fe7c6ef69eb3&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_ubsan%29))
  — first sighting that prompted this fix.
- **STID 3738-56e0** in `AggregateFunctionTimeseriesChanges.h:197`
- **STID 4287-5905** in `AggregateFunctionTimeseriesInstantValue.h:165`

`AggregateFunctionTimeseriesExtrapolatedValue.h` and
`AggregateFunctionTimeseriesToGridSparse.h` contain the identical overflow
pattern and would have been hit next.

## Approach

Add a shared helper `AggregateFunctionTimeseriesBase::isSampleOutOfWindow` that
performs `sample_timestamp + window <= current_timestamp` in 128-bit signed
arithmetic, then route all five call sites through it:

| File | Original | After |
|------|----------|-------|
| `AggregateFunctionTimeseriesLinearRegression.h:250` | `front().first + Base::window <= current_timestamp` | `Base::isSampleOutOfWindow(front().first, current_timestamp)` |
| `AggregateFunctionTimeseriesExtrapolatedValue.h:263` | same | same |
| `AggregateFunctionTimeseriesChanges.h:197` | same | same |
| `AggregateFunctionTimeseriesToGridSparse.h:152` | `previous_timestamp + Base::window > current_timestamp` | `!Base::isSampleOutOfWindow(previous_timestamp, current_timestamp)` |
| `AggregateFunctionTimeseriesInstantValue.h:165` | `timestamps[1] + Base::window > current_timestamp` | `!Base::isSampleOutOfWindow(timestamps[1], current_timestamp)` |

The underflow-prone subtraction `range_start = current_timestamp - Base::window`
inside `AggregateFunctionTimeseriesExtrapolatedValue::fillResultValue` is also
moved into `Float64` so it cannot signed-overflow `TimestampType` (the same
adversarial input that crashes line 263 reaches this subtraction whenever
two-or-more samples remain in the deque).

The helper is correct for both signed (`DateTime64` = `Int64`) and unsigned
(`DateTime`/`UInt32`) `TimestampType`: both operands are sign-extended via
`Int64` and then widened to `__int128`, which has more than enough range to hold
the sum of two 64-bit signed integers.

## Reproducer (deterministic, fails on master, passes on this PR)

```sql
SET allow_experimental_time_series_aggregate_functions = 1;
SELECT timeSeriesDerivToGrid(
    -9223372036854775808::Int64,
    9223372036854775807::Int64,
    1152921504606846976::Int64,
    9223372036854775807::Int64)(
        [reinterpret(1577862000::Int64, 'DateTime64(0)')]::Array(DateTime64(0)),
        [1.0]::Array(Float64));
```

On a `-DSANITIZE=address,undefined` build, master aborts with:

```
src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesLinearRegression.h:250:82:
  runtime error: signed integer overflow: 1577862000 + 9223372036854775807 cannot
  be represented in type 'long'
```

This PR returns `[NULL,NULL,...]` and the server stays up.

## Tests

A new regression test
`tests/queries/0_stateless/04256_timeseries_window_int64_max_overflow_stid_4701_4d85.sql`
exercises every affected `timeSeries*ToGrid` function with `window =
9223372036854775807` and a positive sample timestamp.

Locally verified against the ASan+UBSan build:

- `03254_timeseries_functions` — passes byte-for-byte against its reference.
- `03254_timeseries_functions_various_arguments` — passes.
- `04111_timeseries_to_grid_insert_overflow` — passes.
- `04042_time_series_range_overflow` — passes.
- `04256_timeseries_window_int64_max_overflow_stid_4701_4d85` (new) — passes.

Fixed signed integer overflow in `timeSeriesRateToGrid`, `timeSeriesDeltaToGrid`, `timeSeriesDerivToGrid`, `timeSeriesPredictLinearToGrid`, `timeSeriesChangesToGrid`, `timeSeriesResetsToGrid`, `timeSeriesInstantRateToGrid`, `timeSeriesInstantDeltaToGrid` and `timeSeriesResampleToGridWithStaleness` when the staleness (window) parameter is set to an extreme value near `INT64_MAX`.

### Version info
- Merged into: `26.6.1.397`


groeneai:
Pre-PR validation gate (per worker policy, session `cron:clickhouse-ci-task-worker:20260519-124500`):

**(a) Deterministic repro?** Yes. The SQL in the PR description triggers UBSan on master at `AggregateFunctionTimeseriesLinearRegression.h:250` on every run. The same query exits with `[NULL,NULL,...]` on this PR.

**(b) Root cause explained?** Yes. `samples_in_window.front().first + Base::window` is computed in `TimestampType`/`IntervalType` (`Int64` for `DateTime64`, `Int32` for `DateTime`). When `Base::window` is near `INT64_MAX` and the sample timestamp is positive, the signed addition overflows. Five callers across the timeseries directory share this pattern; three already showed up on master in the last 30 days (STID 4701-4d85, 3738-56e0, 4287-5905).

**(c) Fix matches root cause?** Yes. The helper `isSampleOutOfWindow` widens to `__int128` before adding, so the comparison `sample + window <= current` is performed without overflow. All five call sites are routed through the helper. The underflow-prone `current_timestamp - Base::window` subtraction in `AggregateFunctionTimeseriesExtrapolatedValue.h` is hoisted into `Float64`, which is what the value was being cast to immediately afterwards anyway.

**(d) Test intent preserved?** Yes. `03254_timeseries_functions`, `03254_timeseries_functions_various_arguments`, `04111_timeseries_to_grid_insert_overflow` and `04042_time_series_range_overflow` all pass byte-for-byte against their reference output on the ASan+UBSan build with this fix.

**(e) Demonstrated in both directions?** Yes. The exact same reproducer query aborts the server with UBSan on master (verified locally with a fresh `-DSANITIZE=address,undefined` build of master), and returns valid results on this PR.

**(f) Fix is general, not a narrow patch?** Yes. Grep for `Base::window` in `src/AggregateFunctions/TimeSeries/` produces six sites (five OOW / in-window comparisons plus one subtraction); this PR covers all six. The `add()` filter inside `AggregateFunctionTimeseriesBase.h:217` is already marked `NO_SANITIZE_UNDEFINED` and is left alone deliberately — the wrap there is intentional bounds-checking, not a value-producing computation.

groeneai:
cc @vitlibar @alexey-milovidov — could you review? Two-line summary: routes the five `samples_in_window.front().first + Base::window` / `previous_timestamp + Base::window` comparisons across the timeseries aggregate functions through a 128-bit-arithmetic helper, plus a Float64 rewrite of the `current_timestamp - Base::window` subtraction in `AggregateFunctionTimeseriesExtrapolatedValue.h`. Closes three master sightings (STID 4701-4d85, 3738-56e0, 4287-5905) and the same pattern in the two remaining sibling files (`ToGridSparse.h`, `ExtrapolatedValue.h`).

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesBase.h
```diff
@@ -192,6 +192,24 @@ class AggregateFunctionTimeseriesBase :
         return static_cast<TimestampType>(static_cast<Int64>(result_bits));
     }
 
+    /// Returns whether a sample at `sample_timestamp` is past the sliding-window cutoff
```
vitlibar: The first line is enough here. And add one-line comment inside function explaining why we use `Int128` (and use `Int128` instead of `WideInt`).

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesChanges.h
```diff
@@ -193,8 +193,11 @@ class AggregateFunctionTimeseriesChanges final :
                     samples_in_window.push_back({timestamp, value});
             }
 
-            /// Remove samples that are out of the window
```
vitlibar, line 196: Keep the old comment here.

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesExtrapolatedValue.h
```diff
@@ -259,8 +266,11 @@ class AggregateFunctionTimeseriesExtrapolatedValue final :
                 }
             }
 
-            /// Remove samples that are out of the window
```
vitlibar, line 264: The old comment is better

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesInstantValue.h
```diff
@@ -161,8 +161,11 @@ class AggregateFunctionTimeseriesInstantValue final :
             if (bucket_it != buckets.end())
                 last_2_samples.merge(bucket_it->second);
 
-            /// If the oldest of last 2 samples is within the window, we can calculate the rate or delta
```
vitlibar, line 164: The old comment is good enough

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesLinearRegression.h
```diff
@@ -246,8 +246,11 @@ class AggregateFunctionTimeseriesLinearRegression final :
                     samples_in_window.push_back({timestamp, value});
             }
 
-            /// Remove samples that are out of the window
```
vitlibar, line 252: The old comment is good enough

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesToGridSparse.h
```diff
@@ -149,9 +149,12 @@ class AggregateFunctionTimeseriesToGridSparse final :
                 previous_value = values[i];
                 previous_timestamp = timestamps[i];
             }
-            else if (has_previous_value && (previous_timestamp + Base::window > current_timestamp))
+            else if (has_previous_value && !Base::isSampleOutOfWindow(previous_timestamp, current_timestamp))
             {
-                /// Use the previous value if the current timestamp is missing and the previous one is not stale
```
vitlibar, line 154: The old comment is good enough

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesExtrapolatedValue.h
```diff
@@ -136,15 +136,22 @@ class AggregateFunctionTimeseriesExtrapolatedValue final :
...
-        Float64 duration_to_start = static_cast<Float64>(first_timestamp - range_start);
-        Float64 duration_to_end = static_cast<Float64>(range_end - last_timestamp);
+        //
+        // The two durations are mathematically:
+        //   duration_to_start = first_timestamp - (current_timestamp - window)
+        //                     = first_timestamp - current_timestamp + window
+        //   duration_to_end   = current_timestamp - last_timestamp
+        // Both terms are computed in `Float64` to avoid signed integer overflow on the
+        // intermediate `current_timestamp - Base::window` subtraction when `Base::window`
+        // is set near `INT64_MAX` by adversarial AST fuzzer inputs.
```
vitlibar: Make it one-line comment

vitlibar (changes requested):

groeneai:
@vitlibar Thanks for the review. Addressed all the comment-style nits in `bbc27e3a0bfa`:

- `AggregateFunctionTimeseriesBase.h`: trimmed the helper docstring to the single first-line summary, added a one-line comment inside the function explaining the `Int128` choice, and switched the local type from `__int128` to the ClickHouse `Int128` typedef.
- `AggregateFunctionTimeseriesChanges.h`, `AggregateFunctionTimeseriesExtrapolatedValue.h` (l.262), `AggregateFunctionTimeseriesLinearRegression.h`: restored the original `/// Remove samples that are out of the window` one-liner.
- `AggregateFunctionTimeseriesInstantValue.h`: restored the original `/// If the oldest of last 2 samples is within the window, we can calculate the rate or delta` one-liner.
- `AggregateFunctionTimeseriesToGridSparse.h`: restored the original `/// Use the previous value if the current timestamp is missing and the previous one is not stale` one-liner.
- `AggregateFunctionTimeseriesExtrapolatedValue.h` (l.150): collapsed the multi-line block into a single line that preserves the original `Duration between first/last samples and boundary of range.` text plus a brief `Float64` note.

Pure comment + typedef change, no behavioral diff. Verified the affected files compile and the regression test `04256_timeseries_window_int64_max_overflow_stid_4701_4d85` plus the existing `03254_timeseries_*` and `03573_linear_regression_timeseries_*` suites all pass against the rebuilt binary.

groeneai:
**Pre-PR validation gate** (session `cron:clickhouse-ci-task-worker:20260520-151500`, fixup commit `bbc27e3a0bfa`):

- (a) Deterministic repro? Yes. The pre-fix regression test `04256_timeseries_window_int64_max_overflow_stid_4701_4d85` covers all 9 affected `timeSeries*ToGrid` aggregates with `window = INT64_MAX` and a positive `sample_timestamp`. Reproduced the original UBSan abort on the pre-fix `master`, no longer reproduces on the PR branch.
- (b) Root cause explained? Yes. Adversarial fuzzer-generated queries set `window` near `INT64_MAX`; the plain expression `sample_timestamp + window` then signed-overflows `TimestampType` / `IntervalType` inside the sliding-window cutoff check. The fix performs that single add and compare in 128-bit signed arithmetic via a shared `isSampleOutOfWindow` helper, and switches the analogous `Float64` rewrite of `current_timestamp - Base::window` in `AggregateFunctionTimeseriesExtrapolatedValue::fillResultValue`.
- (c) Fix matches root cause? Yes. Every affected call site is replaced with `Base::isSampleOutOfWindow`, which performs the comparison in `Int128` and therefore cannot signed-overflow `TimestampType` for any input. The only knob the helper relies on is the existing constructor invariant `window >= 0`.
- (d) Test intent preserved? Yes. The five existing `03254_timeseries_*` and `03573_linear_regression_timeseries_*` regression tests are unchanged and continue to pass. The new `04256_timeseries_window_int64_max_overflow_stid_4701_4d85` test exercises the overflow path itself, not a weakened assertion.
- (e) Demonstrated in both directions? Yes. The pre-fix binary aborts under UBSan on the queries in `04256_*`; the PR branch returns the expected length-16 arrays (9/9 assertions match the `.reference` file). The full `04256` + `03254_*` + `03573_*` suite is 10/10 OK on the rebuilt binary with the fixup commit applied.
- (f) Fix is general, not a narrow patch? Yes. All six TimeSeries headers (`Base`, `Changes`, `ExtrapolatedValue`, `InstantValue`, `LinearRegression`, `ToGridSparse`) were audited for the same `sample_timestamp + window` and `current_timestamp - window` patterns, and every occurrence routes through the shared overflow-safe helper or the `Float64` rewrite. No symmetric path still uses the unsafe arithmetic.

vitlibar (approved):

groeneai:
Dropped the `no-fasttest` tag in 406858882c6d.

Justification per your suggestion: there was no concrete runtime or flakiness reason. Sibling timeseries regression tests (03254_timeseries_functions_various_arguments, 03254_timeseries_group_array, 03573_linear_regression_timeseries_functions_various_arguments, 03595_changes_timeseries_functions_various_arguments, 04106_timeseries_bucket_count_overflow, 04111_timeseries_to_grid_insert_overflow) all enable `allow_experimental_time_series_aggregate_functions = 1` via `SET` and carry no `no-fasttest` tag. The setting is session-scoped, not gated by a server-config feature flag, so fast test (which only skips `abort_on_logical_error.yaml`, SSH protocol, and legacy geobase) handles it fine.

Verified locally via `clickhouse-local --queries-file ... 04256_*.sql` — all 9 `SELECT length(timeSeries*ToGrid(...))` queries return `16`, matching `.reference` unchanged.

cc @vitlibar — pure tag removal, source-code fix and approved comment-style changes from bbc27e3a0bfa unchanged.

Session: cron:clickhouse-ci-task-worker:20260520-200000

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesExtrapolatedValue.h
```diff
@@ -136,15 +136,14 @@ class AggregateFunctionTimeseriesExtrapolatedValue final :
...
         /// https://github.com/prometheus/prometheus/blob/5e124cf4f2b9467e4ae1c679840005e727efd599/promql/functions.go#L127
         /// which is licensed under the Apache License 2.0
-        // Duration between first/last samples and boundary of range.
-        Float64 duration_to_start = static_cast<Float64>(first_timestamp - range_start);
-        Float64 duration_to_end = static_cast<Float64>(range_end - last_timestamp);
+        // Duration between first/last samples and boundary of range. Computed in `Float64` to avoid signed overflow on `current_timestamp - Base::window`.
+        Float64 duration_to_start =
+            static_cast<Float64>(first_timestamp) - static_cast<Float64>(current_timestamp)
+            + static_cast<Float64>(Base::window);
+        Float64 duration_to_end = static_cast<Float64>(current_timestamp) - static_cast<Float64>(last_timestamp);
```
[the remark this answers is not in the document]
groeneai: Fixed in 05b00e0d. Both `duration_to_end` and `duration_to_start` now subtract in `Int128` first and convert to `Float64` once.

Added `04271_timeseries_extrapolated_value_float64_precision` as a regression test using your exact concrete example (1 ns delta near `DateTime64(9)` present-day epoch). Before the fix the test produces rate `[1.5e-9]` and delta `[1.5]`; after the fix `[2e-9]` and `[2]`. Verified locally by rebuilding without the fix.

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesExtrapolatedValue.h
```diff
@@ -136,15 +136,14 @@ class AggregateFunctionTimeseriesExtrapolatedValue final :
...
-        const auto range_start = current_timestamp - Base::window;
-
         /// The following logic is copied from Prometheus' rate calculation
         /// https://github.com/prometheus/prometheus/blob/5e124cf4f2b9467e4ae1c679840005e727efd599/promql/functions.go#L127
         /// which is licensed under the Apache License 2.0
-        // Duration between first/last samples and boundary of range.
-        Float64 duration_to_start = static_cast<Float64>(first_timestamp - range_start);
-        Float64 duration_to_end = static_cast<Float64>(range_end - last_timestamp);
+        // Duration between first/last samples and boundary of range. Computed in `Float64` to avoid signed overflow on `current_timestamp - Base::window`.
+        Float64 duration_to_start =
```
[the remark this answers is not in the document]
groeneai: Same `Int128`-first treatment applied symmetrically here in 05b00e0d.

groeneai:
Pushed 05b00e0d addressing @clickhouse-gh bot's two Float64 precision threads (`duration_to_end` and `duration_to_start`). Both expressions now subtract in `Int128` first and convert to `Float64` once.

cc @vitlibar - thanks again for the approval; this commit only touches the same lines you reviewed plus a new regression test, no behavior change to the previously-approved code.

### Pre-PR validation gate

a. **Deterministic repro?** Yes. New test `04271_timeseries_extrapolated_value_float64_precision.sql` uses the bot's concrete example (1 ns delta near `DateTime64(9)` ~1.7e18). Without the fix the output is `[1.5e-9]` / `[1.5]`; with the fix `[2e-9]` / `[2]`. Confirmed locally by stashing the source change, rebuilding, running the test, restoring, rebuilding, re-running.

b. **Root cause explained?** Yes. `Float64(current_timestamp) - Float64(last_timestamp)` rounds each operand to ~256 ns quanta at present-day epoch values before subtracting, so 1 ns deltas are silently zeroed. Subtracting in `Int128` first preserves the exact delta and the single `Float64` conversion is precision-safe because the result is always small.

c. **Fix matches root cause?** Yes. Subtract in `Int128`, convert once. Symmetric form for `duration_to_start` (same precision class, the bot raised both threads).

d. **Test intent preserved?** Original 04256 STID regression still passes (`16,16,16,16,16,16,16,16,16` reference unchanged). All other timeseries tests pass (03254_*, 03573_*, 03595_*, 04106_*, 04111_*).

e. **Both directions demonstrated?** Yes - explicitly stashed and rebuilt without the fix to confirm `[1.5e-9]` / `[1.5]`, then restored to `[2e-9]` / `[2]`.

f. **Fix general, not narrow patch?** Yes. The same pattern (`Float64`-first subtraction of large nearby timestamps) only exists at these two lines inside `AggregateFunctionTimeseriesExtrapolatedValue.h`. The sibling files (`Changes.h`, `InstantValue.h`, `LinearRegression.h`, `ToGridSparse.h`) don't have a `current_timestamp - last_timestamp` style expression - they all use `isSampleOutOfWindow` which was already on the `Int128` path from the original fix.

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesBase.h
```diff
@@ -192,6 +192,16 @@ class AggregateFunctionTimeseriesBase :
         return static_cast<TimestampType>(static_cast<Int64>(result_bits));
     }
 
+    /// Returns whether a sample at `sample_timestamp` is past the sliding-window cutoff for grid point `current_timestamp`.
+    bool isSampleOutOfWindow(const TimestampType sample_timestamp, const TimestampType current_timestamp) const
+    {
+        /// Use `Int128` so the addition does not signed-overflow `TimestampType` when `window` is set near `INT64_MAX`.
```
vitlibar: Change the comment to "Compare as Int128 to avoid signed-overflow `TimestampType` when `window` is set near `INT64_MAX`."

vitlibar:
The bugfix validation failed in CI

vitlibar (changes requested):

src/AggregateFunctions/TimeSeries/AggregateFunctionTimeseriesBase.h
groeneai: Done in `e1eb7bd89dde`. Comment-only change, no behavior change. Thanks @vitlibar.

vitlibar:
@groeneai Investigate CI failures

groeneai:
@vitlibar, investigated both failing checks on `e1eb7bd89dde`:

**`02813_starting_in_text_log` on `Stateless tests (amd_tsan, s3 storage, parallel, 2/2)`**: chronic-low-rate flake unrelated to PR diff. CIDB on this exact job: 14d master `0/328` failures, 7d PRs `1/765` failures (this run is the one). The PR touches only `src/AggregateFunctions/TimeSeries/*.h`, no relationship to text_log, startup, or system log code paths. Once the failure mode hits, the runner reproduces it 39/39 with randomization + 61/61 without (sticky environmental state). Recommend retrigger.

**`Bugfix validation (functional tests)` for `04256` + `04271`**: both tests pass on master HEAD release build, so the check reports "Failed to reproduce the bug":

- `04256` (UBSan signed-integer overflow): release build's silent overflow returns the same grid `length() = 16` for each query regardless of the internal arithmetic. The bug is only observable under UBSan instrumentation, which Bugfix validation does not use.
- `04271` (Float64 precision): re-checked empirically. Running the master binary (no fix) against the SQL produces `[2e-9]` and `[2]`, identical to the PR. Master's code is `Float64(range_end - last_timestamp)`, where the subtraction happens in `TimestampType` first (exact integer math), then casts to `Float64`. The bot review on May 20 was about our intermediate WIP commit `02edd52b` which had a different form `Float64(x) - Float64(y)` (cast first, then subtract). Master never had that form. So `04271` documents the safe `Int128` form for our implementation but does not catch a master bug.

Net: this PR has no observable release-build behavior change, only a UBSan-finding removal. Could you advise on the changelog category? Options:
1. Keep `Bug Fix` and request maintainer override-merge (current state).
2. Switch to `CI Fix or Improvement (changelog entry is not required)`, in which case Bugfix validation would not run.
3. Drop `04271` since it does not test a master bug.

Local: 50/50 stable runs of `04256` + `04271` + sibling timeseries tests with the fix applied.

vitlibar (approved):

clickgapai:

### TL;DR
**Fix UBSan signed-integer overflow in `timeSeries*ToGrid` window check**

**ClickGap verified:** reproduced it with the script below on `26.5`; tested 5 version(s) — 4 affected, 1 clean (see Affects); confirmed it predates the oldest available release (pre-existing).

### Reproduction
```
SET allow_experimental_time_series_aggregate_functions = 1; SELECT timeSeriesResampleToGridWithStaleness(0, 10, 1, 9223372036854775807)([toDateTime64(1, 0)], [5.0]); -- window=INT64_MAX overflows the in-grid check; buggy builds drop the sample (all-NULL), the fix returns [NULL,5,5,5,5,5,5,5,5,5,5]
```
[▶ Run on ClickHouse Fiddle](https://fiddle.clickhouse.com/401a2f04-dd13-48ab-978e-75bf175ae277)

### Bisect

### Affects

### Suggested next step

**Component:** `comp-promql`
**Severity:** `P2`

cc fix PR author: 
cc fix PR reviewer: 

---
_ClickGap · Finding: `phase_d_pr105319`_

PR #93414, opened 2026-01-05
Packed format for statistics

CurtizJ:
### Changelog category (leave one):
- Backward Incompatible Change

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Improved storage format of statistics. All statistics are now stored in a single file.

### Version info
- Merged into: `26.2.1.674`

PR #105321, opened 2026-05-19
Packed storage format for skip indices

Algunenano:
- Experimental Feature

Added an experimental MergeTree setting `packed_skip_index_max_bytes` that bundles small skip-index substreams into a single `skp_idx.packed` archive per part, reducing inode pressure when many skip indices are defined on a table. The decision is per substream at write time: substreams whose serialized size stays under the threshold go into the archive, anything larger keeps the standalone `skp_idx_<name>.idx2` / `.mrk2` layout. A single part can mix layouts. Full-text indices are not supported and are always per-file. Default is 0 (packing disabled).

### Version info
- Merged into: `26.6.1.999`


src/Storages/MergeTree/MergeTreeSettings.cpp
```diff
@@ -1923,6 +1923,38 @@ namespace ErrorCodes
     Comma-separated list of statistics types to calculate automatically on all suitable columns.
     Supported statistics types: basic, tdigest, countmin, minmax, uniq.
     )", 0) \
+    DECLARE(UInt64, packed_skip_index_max_bytes, 0, R"(
```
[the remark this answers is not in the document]
Algunenano: Let's not overcomplicate. This is declared and meant to be forward compatible, not backwards compatible and coming up with scenarios involving downgrades is beyond the scope

Algunenano:
The PR has been green multiple times and the AI  reviewers are happy. If you find something in the post-review please let me know and I'll address it, but I'd rather merge this and not have to deal with conflicts multiple times again. Also, it's experimental so I want to have it in 26.6 and enable it in the next release

src/Storages/MergeTree/DataPartStorageOnDiskFull.cpp
```diff
@@ -50,6 +53,11 @@ bool DataPartStorageOnDiskFull::exists() const
 
 bool DataPartStorageOnDiskFull::existsFile(const std::string & name) const
 {
+    if (looksLikePackedSkipIndexFile(name))
+    {
+        if (const auto * reader = getSkipIndicesPackedReader(); reader && reader->exists(name))
+            return true;
+    }
     return volume->getDisk()->existsFile(fs::path(root_path) / part_dir / name);
```
Michicosun, line 61: This file must have already been removed by another replica. Opening it for a packed reader will throw a " no-such-key error.

https://d1k2gkhrlfqv31.cloudfront.net/clickhouse-test-reports-private/json.html?PR=61301&sha=latest&name_0=PR&name_1=Stress+test+%28amd_msan%2C+SharedCatalog%2C+meta+in+keeper%29

Algunenano: Thanks for the heads up. I'll fix it

Algunenano: https://github.com/ClickHouse/ClickHouse/pull/107995/commits/7346f4763f32f49ac84f1b85de0ed80cfae0911e

alesapin (commented):
I've accidentally noticed this change (was working on a new part format). The code here is not well structured. Knowledge about packed format for skip indices is introduced in almost every class related to merge tree. Somehow, even Read step from MergeTree and optimizations had to be changed to support new format of serialization for skip indices. Some abstraction (similar to `IDataPartStorage`) for indices serialization is missing here. That is why methods and functions like `looksLikePackedSkipIndexFile` or `isPacked` or `if (skip_indices_packed_writer` are everywhere, breaking a lot of abstractions. 

For some reason we even support packed indexes in packed parts. What's the point to have this redundant indirection?

Very similar change was done with much less damage: https://github.com/ClickHouse/ClickHouse/pull/93414.

While idea is very good and useful, I would say this change is better to be reverted and reimplemented from scratch.

Algunenano:
> For some reason we even support packed indexes in packed parts. What's the point to have this redundant indirection?

Full explanation is here: https://github.com/ClickHouse/clickhouse-private/issues/58264#issuecomment-4506869050


> I've accidentally noticed this change (was working on a new part format). The code here is not well structured. Knowledge about packed format for skip indices is introduced in almost every class related to merge tree. Somehow, even Read step from MergeTree and optimizations had to be changed to support new format of serialization for skip indices. Some abstraction (similar to IDataPartStorage) for indices serialization is missing here. That is why methods and functions like looksLikePackedSkipIndexFile or isPacked or if (skip_indices_packed_writer are everywhere, breaking a lot of abstractions.

> Very similar change was done with much less damage: https://github.com/ClickHouse/ClickHouse/pull/93414.

There are some key differences between the 2.

Namely that packed statistics is all or nothing in a part, while packed indices is supported per substream in order to differentiate between large and small indices and be able to work with both.

Also, there are tons of existing code working with indices. Skip indices are enumerated, hardlinked, checksummed, etc. in many places from existed code. Statistics is just read by the estimator.

> While idea is very good and useful, I would say this change is better to be reverted and reimplemented from scratch.

I'll work on a followup to simplify this design, although it will require a major refactor of existing code similar to the one that introduced the packed format for indices. Do you mind if I ask for a review from you once it's ready?

src/Storages/MergeTree/IMergeTreeDataPart.cpp
```diff
@@ -2683,6 +2699,33 @@ void IMergeTreeDataPart::calculateSecondaryIndicesSizesOnDisk() const
...
+                    /// destroyed. The undercount matters in one place behaviorally:
+                    /// `ReadFromMergeTree::get_indexes_size` gates `distributed_index_analysis`
+                    /// activation on `data_uncompressed`, so a packed skip index that compresses
+                    /// well (`set` / `bloom_filter` over strings) may not cross
+                    /// `distributed_index_analysis_min_indexes_bytes_to_activate` and distributed
+                    /// index analysis won't trigger. The fallback is the normal query plan, not
+                    /// a wrong result. Elsewhere `data_uncompressed` only feeds telemetry in
+                    /// `system.data_skipping_indices` / `system.parts`. To remove this
+                    /// approximation entirely, add `uncompressed_size` to
+                    /// `PackedFilesIO::Index` entries and bump the archive format version.
```
azat, line 2725: @Algunenano just having "correct plan" is not enough - since it will do incorrect assumptions which may just not activate distributed index analysis
It is OK for now I guess, since it is experimental for now, but we need proper `uncompressed_size`.
Are you planing to work on this?

Algunenano: How critical is it? The index won't be packed if it's larger than `packed_skip_index_max_bytes`. The initial idea that I had once I remove it from experimental is 4MB (and text indices are always unpacked because they use a different mechanism). Would it make a difference if it says 1MB and not 2MB to decide on distributed analysis? For reference `distributed_index_analysis_min_indexes_bytes_to_activate` is 1GB

I can try to get the size from other metrics (if loaded from memory for example) to provide a better number if it's critical.

tavplubix: > Would it make a difference if it says 1MB and not 2MB

how about it says 2.7MB and not 3.7**GB**? https://pastila.clickhouse.com/?00590468/d8c18dd0537b92d1c0a9a8a1e1ca4a23#KRW4Z3x2oHB7vjSmBCSAaQ==GCM

Algunenano: Interesting. Thanks for the heads up. I think this should cover it: https://github.com/ClickHouse/ClickHouse/pull/109272

PR #100448, opened 2026-03-23
Make min/max/argMin/argMax treat NaN consistently with ORDER BY

Algunenano:
Make `min`/`max`/`argMin`/`argMax` skip NaN values, matching `ORDER BY x ASC/DESC LIMIT 1` behavior. Previously, these aggregate functions used IEEE 754 comparisons where NaN is unordered, leading to inconsistent results depending on NaN position in the data. NaN is now always treated as last (skipped), and only returned when all values are NaN.

Also enables the `findExtremeMinIndex`/`MaxIndex` SIMD optimization for floating point `ORDER BY ... LIMIT 1` queries, which was previously disabled due to missing NaN handling.

Closes #72228

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
`min`/`max`/`argMin`/`argMax` now treat NaN consistently with `ORDER BY`: NaN is always skipped (returned only when all values are NaN). Previously, results depended on NaN position in the data due to IEEE 754 unordered comparison semantics.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.478`

PR #105331, opened 2026-05-19
Fix NaN handling in minMap/maxMap (array form) and minMappedArrays/maxMappedArrays

### LINKED ISSUE #105294: `maxMap` / `maxMappedArrays` disagree on `NaN` for floating-point values from 26.4
### Describe what's wrong

Starting in **26.4**, `maxMap(map(...))` and `maxMappedArrays(key, value)` can return different results for the same logical data when `Float` values contain `NaN`. The Map-combinator form skips `NaN` (matching `max` after [#100448](https://github.com/ClickHouse/ClickHouse/pull/100448)); the array form still uses IEEE 754 unordered comparison where a `NaN` seen first never gets replaced.

```sql
SELECT maxMappedArrays(a, b)
FROM VALUES('a Array(Char), b Array(Float64)', (['y'], [nan]), (['y'], [6]));

SELECT maxMap(map('y', x))
FROM VALUES('x Float64', (nan), (6));
```

| Version | `maxMappedArrays(a, b)` | `maxMap(map('y', x))` |
|---------|---------------------------|------------------------|
| 26.3    | `(['y'], [nan])`          | `{'y': nan}`           |
| 26.4    | `(['y'], [nan])`          | `{'y': 6}`             |

Both expressions aggregate the same key/value pairs; the results diverge from 26.4 onward.

Yes

### How to reproduce

https://fiddle.clickhouse.com/53da178c-c77d-44fc-b8c5-779d37e8f6f7

### Expected behavior

`maxMap(map('y', x))` and `maxMappedArrays(['y'], [x])` should agree for the same inputs. Either:

- both skip `NaN` and return `6` (aligned with `max`, `ORDER BY … LIMIT 1`, and #100448), or
- both document and implement the same legacy semantics.

The current split is surprising for users who migrate between the array API.

### Version info
- Resolved by: #105331
- Merged into: `26.5.1.863`

Algunenano:
`minMap`/`maxMap` (array form, redirected to `minMappedArrays`/`maxMappedArrays` by the `Map` combinator) and `minMappedArrays`/`maxMappedArrays` themselves compared values with raw IEEE 754 `<`/`>`, so a `NaN` seen first as the accumulator was never replaced and stayed in the result regardless of the other inputs.

This disagreed with `minMap`/`maxMap` over a `Map` argument, which goes through `min`/`max` and was fixed in #100448 to treat `NaN` as last. The split was visible to users that migrated between the two argument shapes:

| Version | `maxMappedArrays(a, b)` | `maxMap(map('y', x))` |
|---------|---------------------------|------------------------|
| 26.3    | `(['y'], [nan])`          | `{'y': nan}`           |
| 26.4    | `(['y'], [nan])`          | `{'y': 6}`             |

`FieldVisitorMin`/`FieldVisitorMax` now treat `NaN` as last for floating-point values, so it is only returned when every observed value is `NaN` — matching `ORDER BY ... LIMIT 1` and #100448.

Closes #105294

`minMap`/`maxMap` (array form) and `minMappedArrays`/`maxMappedArrays` now treat `NaN` consistently with `ORDER BY`: `NaN` is treated as last (returned only when all values are `NaN`). Previously, results depended on the position of `NaN` in the data due to IEEE 754 unordered comparison semantics, and disagreed with the `Map`-argument form of `minMap`/`maxMap` that was fixed in #100448.

### Version info
- Merged into: `26.5.1.863`

### DIFF
--- a/src/AggregateFunctions/AggregateFunctionSumMap.cpp
+++ b/src/AggregateFunctions/AggregateFunctionSumMap.cpp
@@ -13,6 +13,7 @@
 #include <AggregateFunctions/FactoryHelpers.h>
 #include <AggregateFunctions/IAggregateFunction.h>
 #include <Common/FieldVisitorSum.h>
+#include <Common/NaNUtils.h>
 #include <Common/assert_cast.h>
 #include <Common/MapWithMemoryTracking.h>
 #include <Common/SetWithMemoryTracking.h>
@@ -583,6 +584,19 @@ private:
     bool compareImpl(FieldType & x) const
     {
         auto val = rhs.safeGet<FieldType>();
+        if constexpr (is_floating_point<FieldType>)
+        {
+            /// Match `max` semantics: NaN is treated as last (i.e. smaller than
+            /// any non-NaN value), so it is only kept when every observed value
+            /// is NaN. See #100448.
+            if (isNaN(val))
+                return false;
+            if (isNaN(x))
+            {
+                x = val;
+                return true;
+            }
+        }
         if (val > x)
         {
             x = val;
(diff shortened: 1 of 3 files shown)


alexey-milovidov (approved):

clickgapai:

**Affected code:** `FieldVisitorMax::compareImpl` and `FieldVisitorMin::compareImpl` in AggregateFunctionSumMap.cpp — core code present in all supported branches.

**Why this might apply:** This is a bug fix for wrong results. The `minMappedArrays`/`maxMappedArrays` functions and `FieldVisitorMax`/`FieldVisitorMin` have existed since at least 2023 (commit 01d903c60d3 'add minMap and maxMap functions'), predating all supported branches. The bug causes silent wrong results when Float64 values contain NaN.

PR #105334, opened 2026-05-19
Reject positional file argument combined with --query

### LINKED ISSUE #104735: Positional file argument bypasses `--query`/`--queries-file` conflict check, silently ignoring `-q`

_Retrospective finding from a historical scan of [PR #71589](https://github.com/ClickHouse/ClickHouse/pull/71589) (merged 2024-11-11). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

When using a positional file argument with `-q`, the query is silently ignored instead of throwing BAD_ARGUMENTS. Example: `clickhouse local file.sql -q "SELECT 1"` runs only file.sql, ignoring the SELECT 1 query.

**Root cause:** LocalServer.cpp:801 and Client.cpp:1051: The conflict check uses getClientConfiguration().has("queries-file") which only detects explicit --queries-file flags in argv (via argsToConfig), not positional file arguments converted to queries-file via emplace.

**Why we believe this is a bug:** ClientApplicationBaseParser.cpp:84 → options.emplace("queries-file", value) adds to boost options map. LocalServer.cpp:801 checks getClientConfiguration().has("queries-file") which reads from Poco config (populated by argsToConfig from raw argv). Positional file arguments are never in raw argv as --queries-file, so the check fails.

**Affected locations:**
- `programs/local/LocalServer.cpp:801` — conflict check uses config.has() instead of member variable
- `programs/client/Client.cpp:1051` — same issue in client
- `src/Client/ClientApplicationBaseParser.cpp:84` — emplace adds to options map but not to Poco config

**Impact:** Silent data loss - users' -q queries are ignored without any error when combined with positional file arguments. Users expect either both to run or an error, not silent dropping.

Yes — confirmed on current `master` (commit `55a1c75ba12`).

### How to reproduce

```sql
FILE=/tmp/test.sql && echo "SELECT 'from file'" > $FILE && clickhouse local $FILE -q "SELECT 'from query'" && rm $FILE
# Should throw BAD_ARGUMENTS but instead outputs 'from file' only
```

### Expected behavior

```
Explicit --queries-file + -q:
BAD_ARGUMENTS
Positional file + -q:
BAD_ARGUMENTS
```

```
Explicit --queries-file + -q:
BAD_ARGUMENTS
Positional file + -q:
NO_ERROR
```

### Additional context

**Open risks:**
- Same issue might exist for delayed_interactive check at LocalServer.cpp:806 which also uses config.has("queries-file")

**Suggested fix:** Change the check from `getClientConfiguration().has("queries-file")` to `!queries_files.empty()` in both LocalServer.cpp:801 and Client.cpp:1051. The `queries_files` member is populated correctly from the boost options map.

Found during automated review of [PR #71589](https://github.com/ClickHouse/ClickHouse/pull/71589).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr71589_001`_

### Version info
- Resolved by: #105334
- Merged into: `26.5.1.934`

Algunenano:
When a positional file argument is passed to `clickhouse local` or `clickhouse client` together with `--query`/`-q`, the conflict was not detected and the `-q` query was silently dropped, e.g.:

```
$ echo "SELECT 'from file'" > /tmp/test.sql
$ clickhouse local /tmp/test.sql -q "SELECT 'from query'"
from file
```

The positional argument is converted to `--queries-file` inside `ClientApplicationBaseParser` via `options.emplace`, which updates the boost options map (and the `queries_files` member) but never the Poco config. The check in `processConfig` used `config().has("queries-file")`, so the positional path slipped through. Use `queries_files` directly so both forms are rejected consistently.

Closes #104735

Fix `clickhouse local` and `clickhouse client` silently ignoring `--query`/`-q` when a positional file argument was also given. Both forms now produce `BAD_ARGUMENTS` consistently.

### Version info
- Merged into: `26.5.1.934`


antonio2368 (approved):

Algunenano:
test_grpc_protocol/test.py::test_cancel_while_generating_output -> https://github.com/ClickHouse/ClickHouse/pull/105415

PR #105335, opened 2026-05-19
Improve QueryConditionCache to record unmatched granules

hanfei1991:
Previously the PREWHERE write path called `addPrewhereUnmatchedMarks` only when the entire read batch produced zero rows (`row_count == 0`). Any granule that was fully filtered out within a batch that also contained at least one matching granule was silently ignored. See https://fiddle.clickhouse.com/83097d38-0c17-4546-ac61-1c6d01fba4e9
- Improvement

 The `QueryConditionCache` now records individually filtered-out granules even within read batches that partially pass PREWHERE, reducing the number of marks re-read by subsequent queries with the same condition.

### Version info
- Merged into: `26.6.1.169`


tests/queries/0_stateless/04249_query_condition_cache_fine_grained_granules.reference
```diff
@@ -0,0 +1,2 @@
+2	100
+1	1
```
hanfei1991, line 2: Previously it's '1 100', we cannot filter anything

tests/queries/0_stateless/04249_query_condition_cache_fine_grained_granules.reference
```diff
@@ -0,0 +1,2 @@
+2	100
```
antaljanosbenjamin, line 1: Dow we know what is causing 2 hits in the first query?

antaljanosbenjamin (approved):
The first paragraph of the changelog is hard to understand from the user's perspective. Maybe keep the second half only.

tests/queries/0_stateless/04249_query_condition_cache_fine_grained_granules.reference
hanfei1991: yes, because we try to match twice, prewhere and where, so we missed twice

PR #105336, opened 2026-05-19
Add more keeper profile events for watches

scanhex12:
- Improvement

Add more keeper profile (server-side + client-side) events for watches. Closes https://github.com/ClickHouse/ClickHouse/issues/97703

### Version info
- Merged into: `26.6.1.398`

### DIFF
--- a/src/Common/ZooKeeper/ZooKeeperImpl.cpp
+++ b/src/Common/ZooKeeper/ZooKeeperImpl.cpp
@@ -75,6 +75,22 @@ namespace ProfileEvents
     extern const Event ZooKeeperBytesSent;
     extern const Event ZooKeeperBytesReceived;
     extern const Event ZooKeeperWatchResponse;
+    extern const Event ZooKeeperWatchCallbackDurationMicroseconds;
+    extern const Event ZooKeeperWatchCallbackErrors;
+    extern const Event ZooKeeperWatchTriggeredOther;
+    extern const Event ZooKeeperWatchTriggeredReplicatedMergeTreeQueue;
+    extern const Event ZooKeeperWatchTriggeredReplicatedMergeTreeLog;
+    extern const Event ZooKeeperWatchTriggeredReplicatedMergeTreeMutations;
+    extern const Event ZooKeeperWatchTriggeredReplicatedMergeTreeLeaderElection;
+    extern const Event ZooKeeperWatchTriggeredDistributedDDL;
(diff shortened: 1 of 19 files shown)


src/Common/ZooKeeper/ZooKeeperImpl.cpp
```diff
@@ -336,6 +352,57 @@ namespace Coordination
...
 using namespace DB;
 
+namespace
+{
+
+ProfileEvents::Event watchTriggeredProfileEvent(WatchCallbackKind kind)
+{
+    switch (kind)
+    {
+        case WatchCallbackKind::Other: return ProfileEvents::ZooKeeperWatchTriggeredOther;
```
vdimir: Why do we want introduce `WatchCallbackKind` enum instead of storing corresponding `ProfileEvents` directly in `WatchCallbackPtrOrEventPtr`?

src/Common/ZooKeeper/ZooKeeperImpl.cpp
```diff
@@ -336,6 +352,57 @@ namespace Coordination
...
+    const auto start_time = std::chrono::steady_clock::now();
+    try
+    {
+        event_or_callback.invoke(response);
+    }
+    catch (...)
+    {
+        ProfileEvents::increment(ProfileEvents::ZooKeeperWatchCallbackErrors);
+        if (log)
+            tryLogCurrentException(log);
```
vdimir: Maybe we could log error even if `log` is null e.g. `tryLogCurrentException(__PRETTY_FUNCTION__)` to make sure no error is swallowed

src/Common/ZooKeeper/ZooKeeperImpl.cpp
```diff
@@ -336,6 +352,57 @@ namespace Coordination
...
+}
+
+void triggerWatchCallback(
+    const WatchCallbackPtrOrEventPtr & event_or_callback,
+    const WatchResponse & response,
+    const LoggerPtr & log)
+{
+    ProfileEvents::increment(watchTriggeredProfileEvent(event_or_callback.getKind()));
+
+    const auto start_time = std::chrono::steady_clock::now();
```
vdimir: You may also use `ProfileEventTimeIncrement` guard for incrementing time event. It will also increment in case of error, though, so depends on intent.

vdimir (approved):

vdimir (commented):
It's also probably worth adding some smoke test case that check that some metrics actually calculated.

PR #103950, opened 2026-05-03
Add test: `ToShortest` fallback in `formatReadable` is reachable but untested across the suite

clickgapai:
_Found via ClickGap automated review. Please close or comment if this is incorrect or needs adjustment._

_Retrospective finding from a historical scan of [PR #61794](https://github.com/ClickHouse/ClickHouse/pull/61794) (merged 2024-03-23). Confirmed on current codebase — close with a note if already fixed._

_This is a test-only PR — no source code changes. Please review: test quality, whether the claimed coverage gaps are real, and whether test output makes sense._

Adds test coverage for 1 untested code path(s), found during automated review of [PR #61794](https://github.com/ClickHouse/ClickHouse/pull/61794).
 That PR PR 61794 fixes pretty number printing. Two source changes: (1) `src/Common/formatReadable.cpp` — adds a `ToShortest` fallback when `ToFixed` returns false in the shared `formatReadable` helper, and extends `formatReadableQuantity` units from quadrillion to vigintillion (22 entries). (2) `src/Process

**1. `ToShortest` fallback in `formatReadable` is reachable but untested across the suite**
  `src/Common/formatReadable.cpp:30`
  **Risk:** Call chain: SQL `formatReadableSize(1e308)` → `FunctionFormatReadableSize::executeImpl` → `formatReadableSizeWithBinarySuffix(value, out, precision)` → static `formatReadable` (formatReadable.cpp:16).
  **What's unique vs PR tests:** PR test `03019_numbers_pretty.sql` covers `1e111` via `formatReadableQuantity` invoked from `writeReadableNumberTip`. After the new vigintillion unit, `1e111` divides down to ~1e48 which still fits To
  [Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/1e4ca812-afdf-4a2b-b040-5090f889b0bd)

### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Not applicable — test-only change.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #105350, opened 2026-05-19
Add test: `ColumnDecimal::compareTrackAt` res > 0 tracking branch untested in merge join

clickgapai:

Adds test coverage for 1 untested code path(s), found during automated review of [PR #105110](https://github.com/ClickHouse/ClickHouse/pull/105110).
That PR This PR adds `compareTrackAt` method for `ColumnDecimal` to optimize merge join performance. The method tracks how many consecutive rows can be skipped during merge join comparison. Two main branches: (1) same-scale path using direct comparison, (2) different-scale path using `decimalLess`. The PR's

**1. `ColumnDecimal::compareTrackAt` res > 0 tracking branch untested in merge join**
  `src/Columns/ColumnDecimal.cpp:100`, `src/Columns/ColumnDecimal.cpp:101`
  **Risk:** `MergeJoin.cpp`:168 calls `compareTrackAt`, which returns a positive value when left > right. `MergeJoin.cpp`:397-398 uses this value to skip right-side rows via `rhs.nextN(cmp)`. If the tracking loop has bugs (e.g., wrong comparison condition), rows could be incorrectly skipped causing silent data loss in join results.
  **What's unique vs PR tests:** PR test 01881_join_on_conditions_merge_decimal.sql.j2 uses data where t1.id=[1,2,2,3], t2.id=[2,2,3,4] — always results in left <= right at comparison, never triggering res > 0 branch. My test uses t1.id=[5,6,7], t2.id=[1,2,3,5,6] which triggers res > 0 when t1[0]=5 > t2[0]=1.
  [Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/66661085-a531-4019-9460-92197e745d55)


clickgapai:
**Pre-PR Validation Gate** (per ClickGap pipeline):

a) **Deterministic repro?** ✅
Test deterministically exercises the previously-uncovered branch and passes (correct behaviour). Confirmed across runs.

b) **Root cause explained?** ✅
`ColumnDecimal.cpp`:100-103 — `res > 0` branch with tracking loop. PR's test data has t1.id=[1,2,2,3] and t2.id=[2,2,3,4] which always results in left <= right at comparison points, never triggering this branch.

c) **Fix matches root cause?** ✅
N/A for coverage gap — this PR adds a regression test, not a fix. Asserts current correct behaviour at the cited line.

d) **Test intent preserved? / New tests added?** ✅
New regression test added: `04257_compareTrackAt_decimal_res_positive.sql`.

Test passes (MATCH) on master proving current correctness; would fail if the cited branch regressed. Single-direction is the correct shape for a coverage gap test.

f) **Fix is general, not a narrow patch?** ✅
Test asserts the invariant at the cited line — covers the branch, not a specific input shape.

zlareb1:
@clickgapai check the review comments and failing CI and make appropriate changes

zlareb1:
@groeneai can you take this up and open a new PR?

groeneai:
Yes. Triaged as part of this batch of 20; the target code path still exists on master (2850e469dcac84b) and the test passes there against its committed reference.

I am re-filing the batch grouped by area rather than as 18 separate test PRs, to keep reviewer load down. Before filing, I mutation-test each oracle — in your previous batch two tests passed both with and without the defect, so I will not re-file one that does not actually redden. If this one turns out to be vacuous or already covered, I will post the reason here instead of a PR.

Consolidated triage for the whole batch: https://github.com/ClickHouse/ClickHouse/pull/103950#issuecomment-5244540248

groeneai:
Triage result: I am not re-filing this one, because I do not think the test would fail if the behaviour it targets regressed, and I did not have the budget to settle that by measurement.

`compareTrackAt`'s `Int64` return is a skip-count hint, consumed only in `MergeJoin.cpp` to advance a cursor by that many rows. The test asserts the final join result set, which the equal-range scan re-derives afterwards, so a wrong positive count plausibly changes only how far the cursor jumps and not the output. That is analysis, not a measurement: the honest way to settle it is to delete the `res > 0` loop, rebuild, and see whether the test reddens. If it does, the gap is real and I will file it.

I filed one PR from the batch of 18: https://github.com/ClickHouse/ClickHouse/pull/114234. Separately, the second query in this test may not be honouring `join_algorithm = 'partial_merge'`, which is worth checking before relying on it, since a test that silently takes the hash path would not exercise `compareTrackAt` at all.

PR #105373, opened 2026-05-19
Fix REPLACE TEMPORARY TABLE silently creating non-existent table

### LINKED ISSUE #102924: `REPLACE TEMPORARY TABLE` succeeds when table doesn't exist, inconsistent with `REPLACE TABLE`

_Retrospective finding from a historical scan of [PR #89450](https://github.com/ClickHouse/ClickHouse/pull/89450) (merged 2025-11-05). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`REPLACE TEMPORARY TABLE tmp ...` silently creates a new table when `tmp` doesn't exist, unlike `REPLACE TABLE` which correctly fails with `UNKNOWN_TABLE`

**Root cause:** InterpreterCreateQuery.cpp:2339: `addOrUpdateExternalTable` is used unconditionally for both `CREATE OR REPLACE` and bare `REPLACE`. For bare `REPLACE` (where `create.create_or_replace == false`), it should use `updateExternalTable` (which checks existence) or guard with an explicit existence check.

**Why we believe this is a bug:** InterpreterCreateQuery.cpp:1784-1787 dispatches both `CREATE OR REPLACE TEMPORARY TABLE` and `REPLACE TEMPORARY TABLE` to `doCreateOrReplaceTemporaryTable`. Inside at line 2339, `addOrUpdateExternalTable` unconditionally creates or replaces without checking `create.create_or_replace`. For bare `REPLACE` (`create_or_replace=false`), the table should be required to exist — matching `doCreateOrReplaceTable` (line 2275-2279) which sets `exchange=true` to fail if target doesn't exist.

**Affected locations:**
- `src/Interpreters/InterpreterCreateQuery.cpp:2339` — addOrUpdateExternalTable called unconditionally in doCreateOrReplaceTemporaryTable

**Impact:** Semantic inconsistency: `REPLACE TEMPORARY TABLE` has `CREATE OR REPLACE` semantics instead of `REPLACE` semantics. Users who rely on `REPLACE` to fail when the table doesn't exist (as documented behavior for non-temporary tables in test 01185) will get silent incorrect behavior.

Yes — confirmed on current `master` (commit `5c959af47103`).

### How to reproduce

```sql
-- Test: REPLACE TEMPORARY TABLE should fail when table doesn't exist,
-- consistent with REPLACE TABLE behavior (test 01185 line 5).
-- Covers: InterpreterCreateQuery.cpp doCreateOrReplaceTemporaryTable missing
-- existence check for bare REPLACE (create_or_replace=false).

-- Sanity: REPLACE TABLE on non-existent table errors with UNKNOWN_TABLE
-- (already tested in 01185, but repeated here for clarity)
REPLACE TABLE __tmp_04103 (n UInt32) ENGINE = Memory AS SELECT * FROM numbers(5); -- { serverError UNKNOWN_TABLE }

-- BUG: REPLACE TEMPORARY TABLE on non-existent table should also error
-- but currently succeeds silently, creating the table
REPLACE TEMPORARY TABLE __tmp_04103 (n UInt32) AS SELECT * FROM numbers(5); -- { serverError UNKNOWN_TABLE }

-- CREATE OR REPLACE should always work regardless of existence
CREATE OR REPLACE TEMPORARY TABLE __tmp_04103 (n UInt32) AS SELECT * FROM numbers(3);
SELECT * FROM __tmp_04103 ORDER BY n;

DROP TEMPORARY TABLE __tmp_04103;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/87d38e86-2daa-4290-b47c-45eafe87110f)

### Expected behavior

```
0
1
2
```

```
The query succeeded but the server error '60' was expected (query: REPLACE TEMPORARY TABLE __tmp_04103 (n UInt32) AS SELECT * FROM numbers(5);).
0
1
2
```

### Additional context

**Suggested fix:** Check `create.create_or_replace`: if true, use `addOrUpdateExternalTable`; if false (bare REPLACE), check existence first or use `updateExternalTable`, and throw `UNKNOWN_TABLE` if the table doesn't exist.

Found during automated review of [PR #89450](https://github.com/ClickHouse/ClickHouse/pull/89450).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr89450_001`_

groeneai:
Fixes [#102924](https://github.com/ClickHouse/ClickHouse/issues/102924).

`REPLACE TEMPORARY TABLE t (...)` on a non-existent `t` silently created the table instead of failing with `UNKNOWN_TABLE`, inconsistent with `REPLACE TABLE` (test `01185`). `doCreateOrReplaceTemporaryTable` called `addOrUpdateExternalTable` unconditionally for both bare `REPLACE` and `CREATE OR REPLACE`. `doCreateOrReplaceTable` already gates the bare-`REPLACE` case via `exchange=true` on the rename; the temporary-table path was missing the equivalent check.

Branch on `create.create_or_replace`: for the bare `REPLACE` case, look up the temporary table via `findExternalTable` and throw `UNKNOWN_TABLE` when it isn't present.

Fix `REPLACE TEMPORARY TABLE` silently creating a new table when the target didn't exist. It now throws `UNKNOWN_TABLE`, matching `REPLACE TABLE` semantics. Use `CREATE OR REPLACE TEMPORARY TABLE` to keep the create-or-replace behavior.

### Version info
- Merged into: `26.5.1.909`


groeneai:
Pre-PR validation gate (session `cron:clickhouse-ci-task-worker:20260519-200000`):

**a) Deterministic repro:** `clickhouse local --query "REPLACE TEMPORARY TABLE x (n UInt32) AS SELECT 1"`. Before: silent success (`exit=0`, no output). After: `Code: 60. DB::Exception: Temporary table x doesn't exist. (UNKNOWN_TABLE)`.

**b) Root cause:** `InterpreterCreateQuery::doCreateOrReplaceTemporaryTable` (line 2327) calls `addOrUpdateExternalTable` unconditionally. Both `REPLACE TEMPORARY TABLE` and `CREATE OR REPLACE TEMPORARY TABLE` dispatch into it (line 1786), and the function never branched on `create.create_or_replace`. The non-temporary sibling `doCreateOrReplaceTable` already gates via `ast_rename->exchange = true` (line 2287), so the asymmetry was localized.

**c) Fix matches root cause:** Gate the temporary path on `create.create_or_replace`. For bare `REPLACE`, `findExternalTable` + throw `UNKNOWN_TABLE`. Mirrors the existing non-temporary check.

**d) Test intent preserved / regression test added:** New `04257_replace_temporary_table_must_exist_102924`. Exercises 4 scenarios: bare REPLACE on non-existent (throws), CREATE OR REPLACE on non-existent (creates), CREATE OR REPLACE on existing (replaces), bare REPLACE on existing (replaces), DROP then bare REPLACE (throws).

**e) Both directions demonstrated:**
- Before fix: bare `REPLACE TEMPORARY TABLE` exits 0, no error, table created → test would fail on the `-- { serverError UNKNOWN_TABLE }` annotation.
- After fix: bare `REPLACE TEMPORARY TABLE` throws `Code: 60. UNKNOWN_TABLE`, test passes.

**f) Fix generality:** `doCreateOrReplaceTemporaryTable` has exactly one call site (line 1788), and the dispatch covers all `REPLACE TEMPORARY TABLE` paths regardless of `create.create_or_replace`. No symmetric storage variants need the same fix — the non-temporary path is already correct.

groeneai:
cc @azat — fix per your directive on #102924. Assigning you per request.

groeneai:
cc @tavplubix — could you take a look as well? Small surgical fix in InterpreterCreateQuery for REPLACE TEMPORARY TABLE existence semantics.

src/Interpreters/InterpreterCreateQuery.cpp
```diff
@@ -2330,6 +2331,11 @@ BlockIO InterpreterCreateQuery::doCreateOrReplaceTemporaryTable(ASTCreateQuery &
     DatabasePtr database = DatabaseCatalog::instance().getDatabase(DatabaseCatalog::TEMPORARY_DATABASE);
 
     String temporary_table_name = create.getTable();
+
+    /// Bare `REPLACE TEMPORARY TABLE` requires the target to exist, matching `REPLACE TABLE` semantics.
+    if (!create.create_or_replace && !getContext()->getSessionContext()->findExternalTable(temporary_table_name))
+        throw Exception(ErrorCodes::UNKNOWN_TABLE, "Temporary table {} doesn't exist", backQuoteIfNeed(temporary_table_name));
```
azat: Can we just use `updateExternalTable` instead of `addOrUpdateExternalTable` here?

groeneai:
@azat — done in ecb5f1a9c2da. The bare `REPLACE` branch now calls `updateExternalTable`, and `CREATE OR REPLACE` keeps `addOrUpdateExternalTable`. Atomic check-and-update under the `Context` mutex, no extra `findExternalTable` round-trip.

One pre-existing snag: `Context::updateExternalTable` was throwing `TABLE_ALREADY_EXISTS` with the message "does not exist". The function had zero callers, so the mismatch was harmless. Fixed in the same commit to throw `UNKNOWN_TABLE` so the public-facing error code matches the message and `REPLACE TABLE` semantics.

All 5 scenarios in the regression test (`04257_replace_temporary_table_must_exist_102924`) pass end-to-end against a local debug build (server output `0 1 2 7 9` matches reference exactly).

azat (approved):

clickgapai:

**Affected code:** `doCreateOrReplaceTemporaryTable` in InterpreterCreateQuery.cpp and `updateExternalTable` in Context.cpp — introduced in 26.2.

**Why this might apply:** Bug was introduced in commit 459297861b7 on 2025-11-03 ('Support CREATE OR REPLACE syntax for temporary tables') which was after 25.8 LTS (released 2025-08-29) but before 26.2 stable (released 2026-02-27). The buggy code exists in 26.2, 26.3, and 26.4. Users relying on REPLACE TEMPORARY TABLE to fail on non-existent tables (matching REPLACE TABLE semantics) are affected.

PR #89352, opened 2025-11-01
Introduce per-table cache of ColumnsDescription for parts to reduce memory usage

azat:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Introduce a per-table cache of `ColumnsDescription` for parts, reducing memory usage when tables contain many parts and many columns.

### Details
Each `part` currently stores its own `ColumnsDescription` (information about all columns). This includes not only full column metadata (including comments), but also `IHints`, which are used for correcting user input and are not trivial to decouple.

When a table has many columns, this structure becomes large, and storing a full copy for every part leads to significant memory consumption.

In one production setup, **12 GiB** out of **73 GiB** of memory was occupied by `ColumnsDescription` objects for **20 K parts with 6.5 M part columns**.

As another example, even on an empty server with merges stopped, `metric_log` alone could cause the server to consume roughly **512 MiB** over a few hours while doing almost nothing.

PR #105376, opened 2026-05-19
Fix columns descriptions cache eviction for Nested columns

### LINKED ISSUE #102926: `decrefColumnsDescriptionForColumns` never evicts cache entries for tables with Nested columns due to wrong `use_count` threshold

_Retrospective finding from a historical scan of [PR #89352](https://github.com/ClickHouse/ClickHouse/pull/89352) (merged 2025-11-18). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

The `columns_descriptions_cache` leaks entries for column schemas that contain Nested columns (e.g., `n.a Array(Int32), n.b Array(String)`). After all parts with such a schema are destroyed, the cache entry persists indefinitely until the table is dropped.

**Root cause:** MergeTreeData.cpp:10836: `if (it->second.original.use_count() == 2)` hardcodes threshold 2 assuming cache's `original` and `with_collected_nested` are the same `shared_ptr`. For Nested columns where `Nested::collect` differs, they are separate objects, so the correct threshold is 1.

**Why we believe this is a bug:** `IMergeTreeDataPart::~IMergeTreeDataPart` (IMergeTreeDataPart.cpp:444-449) → `MergeTreeData::decrefColumnsDescriptionForColumns` (MergeTreeData.cpp:10823-10842). The eviction check at line 10836 uses `use_count() == 2`, which assumes `original` and `with_collected_nested` are aliased to the same `shared_ptr` (contributing 2 to the refcount from the cache). When `Nested::collect` produces a different result (lines 10813-10817), they are different `shared_ptr` objects, so `original` contributes only 1 to the refcount from the cache. After the last part is destroyed, `use_count()` is 1, not 2, so the check never matches.

**Affected locations:**
- `src/Storages/MergeTree/MergeTreeData.cpp:10836` — use_count() == 2 eviction check in decrefColumnsDescriptionForColumns

**Impact:** Memory leak in `columns_descriptions_cache` for tables with Nested columns (e.g., `n.a Array(T), n.b Array(T)` patterns). Each schema change (ALTER ADD/DROP COLUMN) on such tables leaves an orphaned cache entry containing two `ColumnsDescription` objects. Over many schema changes, this accumulates unbounded memory. The leak is cleaned up only when the table is dropped (via `GlobalSum` destructor), but for long-lived tables with frequent schema evolution, it can be significant.

Yes — confirmed on current `master` (commit `15eca5e271a2`).

### How to reproduce

```sql
-- Test: Verify columns_descriptions_cache eviction for nested columns
-- When with_collected_nested != original, use_count()==2 check may fail
DROP TABLE IF EXISTS t_nested_leak;

-- Create table with Nested columns: Nested::collect produces different output
CREATE TABLE t_nested_leak (key Int, `n.a` Array(Int32), `n.b` Array(String)) ENGINE = MergeTree() ORDER BY key;

-- Insert exactly 1 row to create exactly 1 part with old schema
INSERT INTO t_nested_leak VALUES (1, [10], ['hello']);

SELECT 'cache_after_insert:', columns_descriptions_cache_size FROM system.tables WHERE database = currentDatabase() AND table = 't_nested_leak' ORDER BY 1;

-- ALTER ADD COLUMN to change the schema
ALTER TABLE t_nested_leak ADD COLUMN value String SETTINGS mutations_sync=2;

-- Insert with new schema
INSERT INTO t_nested_leak VALUES (10, [30], ['!'], '10');

SELECT 'cache_after_alter_insert:', columns_descriptions_cache_size FROM system.tables WHERE database = currentDatabase() AND table = 't_nested_leak' ORDER BY 1;

-- Now detach the old-schema part (there's exactly 1: all_1_1_0)
ALTER TABLE t_nested_leak DETACH PART 'all_1_1_0';

-- This should be 1 (only new-schema entry), but if use_count()==2 check fails
-- for nested columns, it would be 2 (old entry leaked)
SELECT 'cache_after_detach:', columns_descriptions_cache_size FROM system.tables WHERE database = currentDatabase() AND table = 't_nested_leak' ORDER BY 1;

DROP TABLE IF EXISTS t_nested_leak;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/a484ab0a-6d7b-40c0-80f2-7dfda9267cb5)

### Expected behavior

```
cache_after_insert:	1
cache_after_alter_insert:	2
cache_after_detach:	1
```

```
cache_after_insert:	1
cache_after_alter_insert:	2
cache_after_detach:	2
```

### Additional context

**Open risks:**
- The premature eviction case (N>1 parts with nested columns, first destruction evicts the entry) is a separate issue — it causes unnecessary cache misses but not a leak

**Suggested fix:** Change the threshold to account for whether `original` and `with_collected_nested` are aliased:
```cpp
size_t expected_count = (it->second.original == it->second.with_collected_nested) ? 2 : 1;
if (it->second.original.use_count() == expected_count)
```
Alternatively, check both:
```cpp
if (it->second.original.use_count() == 1 || 
    (it->second.original == it->second.with_collected_nested && it->second.original.use_count() == 2))
```

Found during automated review of [PR #89352](https://github.com/ClickHouse/ClickHouse/pull/89352).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr89352_001`_

groeneai:
Closes #102926.

`MergeTreeData::decrefColumnsDescriptionForColumns` used `use_count() == 2` as the eviction threshold. That holds only when `with_collected_nested` aliases `original` (no `Nested` columns or `share_nested_offsets=0`); when `Nested::collect` returns a distinct list, the cache stores two separate `shared_ptr` objects and `original.use_count()` is 1 after the last part is destroyed, so the entry leaks across `ALTER ADD/DROP COLUMN` cycles.

Pick the expected refcount based on whether the pair aliases.

Fix a memory leak in the per-table `ColumnsDescription` cache for `MergeTree`-family tables with `Nested` columns. Entries were never evicted across `ALTER ADD COLUMN`/`DROP COLUMN` cycles when `share_nested_offsets=1` (the default).

### Version info
- Merged into: `26.5.1.872`

### DIFF
--- a/src/Storages/MergeTree/MergeTreeData.cpp
+++ b/src/Storages/MergeTree/MergeTreeData.cpp
@@ -11076,8 +11076,11 @@ MergeTreeData::ColumnsDescriptionCache MergeTreeData::getColumnsDescriptionForCo
             ? std::make_shared<ColumnsDescription>(Nested::collect(columns))
             : nullptr,
     };
-    if (!cache.with_collected_nested || *cache.with_collected_nested == *cache.original)
-        cache.with_collected_nested = cache.original;
+    /// Keep `with_collected_nested` only when `Nested::collect` produced a distinct list.
+    /// The caller falls back to `original` when this is null, so the cache always holds
+    /// exactly one ref to `original` regardless of the schema shape.
+    if (cache.with_collected_nested && *cache.with_collected_nested == *cache.original)
+        cache.with_collected_nested.reset();
     auto [_, inserted] = columns_descriptions_cache.emplace(columns, cache);
     columns_descriptions_metric_handle.add(inserted);
     return cache;
@@ -11090,12 +11093,10 @@ void MergeTreeData::decrefColumnsDescriptionForColumns(const NamesAndTypesList &
     if (it == columns_descriptions_cache.end())
         return;
 
-    /// `original` always contributes one ref from the cache entry. `with_collected_nested`
-    /// is either aliased to `original` (no `Nested` columns or `share_nested_offsets=0`)
-    /// or a distinct object. In the aliased case the cache contributes 2 refs to the
-    /// `original` shared_ptr; otherwise 1. Evict only when no part still holds a ref.
-    const bool aliased = it->second.original == it->second.with_collected_nested;
-    if (it->second.original.use_count() == (aliased ? 2 : 1))
+    /// The cache entry holds exactly one ref to `original` (via `cache.original`).
+    /// `with_collected_nested` is either null or a distinct shared_ptr, so it does
+    /// not contribute additional refs to `original`. Evict once no part holds a ref.
+    if (it->second.original.use_count() == 1)
     {
         columns_descriptions_cache.erase(it);
         columns_descriptions_metric_handle.sub(1);
--- a/src/Storages/MergeTree/MergeTreeData.cpp
+++ b/src/Storages/MergeTree/MergeTreeData.cpp
@@ -11086,21 +11086,19 @@ MergeTreeData::ColumnsDescriptionCache MergeTreeData::getColumnsDescriptionForCo
 void MergeTreeData::decrefColumnsDescriptionForColumns(const NamesAndTypesList & columns) const
 {
     std::lock_guard lock(columns_descriptions_cache_mutex);
-    if (auto it = columns_descriptions_cache.find(columns); it != columns_descriptions_cache.end())
+    auto it = columns_descriptions_cache.find(columns);
+    if (it == columns_descriptions_cache.end())
+        return;
+
+    /// `original` always contributes one ref from the cache entry. `with_collected_nested`
+    /// is either aliased to `original` (no `Nested` columns or `share_nested_offsets=0`)
+    /// or a distinct object. In the aliased case the cache contributes 2 refs to the
+    /// `original` shared_ptr; otherwise 1. Evict only when no part still holds a ref.
+    const bool aliased = it->second.original == it->second.with_collected_nested;
+    if (it->second.original.use_count() == (aliased ? 2 : 1))
     {
-        /// 1 in the container + 1 in the iterator
-        ///
-        /// Note, we cannot check original.use_count() == with_collected_nested.use_count(),
-        /// since in IMergeTreeDataPart::setColumns() there is a tiny window when it is not correct.
-        ///
-        /// But, if original.use_count() == 2 then it is **always** safe to delete,
-        /// since this means that there are no other references to the shared_ptr
-        /// except in the columns_descriptions_cache and local copy here in iterator.
-        if (it->second.original.use_count() == 2)
-        {
-            columns_descriptions_cache.erase(it);
-            columns_descriptions_metric_handle.sub(1);
-        }
+        columns_descriptions_cache.erase(it);
+        columns_descriptions_metric_handle.sub(1);
     }
 }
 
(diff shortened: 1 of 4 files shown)


groeneai:
**Pre-PR validation gate**

a) Deterministic repro: yes. `tests/queries/0_stateless/04257_102926_columns_descriptions_cache_nested_eviction.sql` triggers the leak on every run with the unfixed code.

b) Root cause: cache entry holds `original` and `with_collected_nested` as two `shared_ptr`. When they alias (no `Nested` or `share_nested_offsets=0`), the cache contributes 2 refs to `original`, so the old `use_count() == 2` check fired. When `Nested::collect` returns a distinct list, the cache contributes only 1 ref to `original`, so the check never fired and the entry leaked.

c) Fix matches root cause: pick the expected count based on whether the two `shared_ptr` actually alias.

d) New regression test added — `Nested`-column path; existing `03673_columns_description_cache` continues to pass.

e) Both directions verified locally: without the fix, `cache_after_detach = 2`; with the fix, `cache_after_detach = 1`.

f) Scope: this is the only call site that evicts from `columns_descriptions_cache`. No symmetric paths to widen.

Session: cron:clickhouse-ci-task-worker:20260519-203000

groeneai:
cc @azat — per your directive on #102926, assigning you. Compact PR description as requested.

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -11086,21 +11086,19 @@ MergeTreeData::ColumnsDescriptionCache MergeTreeData::getColumnsDescriptionForCo
...
-        /// But, if original.use_count() == 2 then it is **always** safe to delete,
-        /// since this means that there are no other references to the shared_ptr
-        /// except in the columns_descriptions_cache and local copy here in iterator.
-        if (it->second.original.use_count() == 2)
-        {
-            columns_descriptions_cache.erase(it);
-            columns_descriptions_metric_handle.sub(1);
-        }
+        columns_descriptions_cache.erase(it);
+        columns_descriptions_metric_handle.sub(1);
```
azat, line 11102: Let's fix the original problem instead

```
if (!cache.with_collected_nested || *cache.with_collected_nested == *cache.original)
    cache.with_collected_nested = cache.original;
```

And then update the caller

```
columns_description_with_collected_nested =
    columns_descriptions.with_collected_nested
        ? columns_descriptions.with_collected_nested
        : columns_descriptions.original;
```

I have not tested any of this, so PTAL

groeneai:
@azat thanks — pushed the refactor in 875b8ab.

The cache populator now resets `with_collected_nested` to `nullptr` when `Nested::collect` returns content equal to `columns` (instead of aliasing it to `original`). The caller in `IMergeTreeDataPart::setColumns` reads it through the ternary fallback you suggested. The decref drops the aliased/non-aliased branching and uses an unconditional `use_count() == 1`.

This makes the cache invariant uniform: `cache.original` is the only ref-contributor to the `original` shared_ptr, regardless of whether the schema has `Nested` columns or `share_nested_offsets`. `cache.with_collected_nested` is either `nullptr` (no distinct list needed) or a distinct `shared_ptr` (genuine `Nested` case) — it never double-counts `original`.

Verified locally:
- Regression test `04257_102926_columns_descriptions_cache_nested_eviction` still passes 30/30 with randomization.
- Focused sweep over `*nested*` / `*subcolumn*` / `*columns_description*` tests is clean (the two failures are unrelated infra: distributed-cluster `127.0.0.2` and `deltaLakeLocal`).

PTAL.

azat (approved):

clickgapai:

**Affected code:** `MergeTreeData::decrefColumnsDescriptionForColumns` and `MergeTreeData::getColumnsDescriptionForColumns` — introduced in 26.2.

**Why this might apply:** The columns_descriptions_cache was introduced in PR #89352 (merged 2025-11-18), which is after the 25.8 LTS release (2025-08-29) but before 26.2, 26.3, and 26.4 releases. The memory leak affects all users with Nested columns performing ALTER operations. Git log shows `decrefColumnsDescriptionForColumns` was introduced in commit a3473b86db4 (2025-11-07).

PR #105384, opened 2026-05-19
Fix set skip index throwing "It's a bug!" on non-integer WHERE atoms

### LINKED ISSUE #105355: `__bitWrapperFunc` internal assertion ("It's a bug!") on Float WHERE predicate against a `TYPE set(N)` skip index
**Version:** 26.5.1.805 (head image), self-hosted Docker.

### Repro (5 lines)

```sql
CREATE TABLE t (c0 UInt32, INDEX idx c0 TYPE set(100) GRANULARITY 4)
  ENGINE = MergeTree() ORDER BY c0;
INSERT INTO t VALUES (1);
SELECT c0 FROM t WHERE c0 + 0.1;
```

### Observed

```
Code: 36. DB::Exception: It's a bug! Only integer types are supported by
__bitWrapperFunc: while executing 'FUNCTION
__bitWrapperFunc(plus(c0, 0.1) :: 2) -> __bitWrapperFunc(plus(c0, 0.1)) UInt8 : 1':
While executing MergeTreeSelect(pool: ReadPoolInOrder, algorithm: InOrder).
(BAD_ARGUMENTS)
```

### Expected

Either the regular `Code: 59. ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER` that the
planner emits when the skip index is absent, or coercion of the Float
result to UInt8 via `!= 0`. The server should not hit an internal-invariant
`"It's a bug!"` assertion on user-reachable SQL.

### Bisection

The bug is specific to `TYPE set(N)`. Tested variants:

| variant | fires? |
|---|---|
| baseline (above) | ✓ |
| drop the skip INDEX | ✗ (returns the row) |
| INDEX `TYPE bloom_filter` | ✗ |
| INDEX `TYPE minmax` | ✗ |
| empty table (no INSERT) | ✗ |
| WHERE replaced by `log(c0)` or `-log(c0)` | ✓ |
| with vs without `PARTITION BY` | both ✓ |
| inside `INTERSECT` vs not | both ✓ |

The `set` skip-index granule-pruning path wraps the WHERE predicate in
`__bitWrapperFunc(... :: 2) -> UInt8` to coerce it to a granule-mask
boolean; the wrapper rejects Float input with the `"It's a bug!"`
BAD_ARGUMENTS message. Adjacent issue #103103 covers a related
`__bitWrapperFunc` granule-pruning bug (NULL propagation) on the same code
path.

Surfaced by sqlancer fuzzing (TLP/PQS oracles) and reduced manually.

### Version info
- Resolved by: #105384
- Merged into: `26.5.1.866`
- Backported to: `26.3.13.19`

groeneai:
Closes #105355.

`MergeTreeIndexConditionSet::traverseDAG` wraps every WHERE atom in `__bitWrapperFunc` to coerce it to a `BoolMask`. `__bitWrapperFunc` is defined only for integer arguments and throws `BAD_ARGUMENTS` ("It's a bug!") on `Float`, `BFloat16`, etc. Queries such as `WHERE c0 + 0.1` against a `TYPE set(N)` skip index hit that throw at execution time.

Check the atom result type (after stripping `Nullable`) before wrapping. If it is not integer, fall back to `UNKNOWN_FIELD` so the index does not prune granules and the query goes through the regular filter path. This matches the behavior the user gets when the index is absent.

Fix `BAD_ARGUMENTS` exception "It's a bug! Only integer types are supported by `__bitWrapperFunc`" thrown when a `SELECT` with a `TYPE set(N)` skip index has a WHERE atom whose result type is `Float`, `BFloat16`, or any other non-integer type (for example `WHERE c0 + 0.1` or `WHERE log(c0)`). The skip index now falls back to the regular filter path in that case.

### Version info
- Merged into: `26.5.1.866`
- Backported to: `26.3.13.19`


groeneai:
### Pre-PR validation gate

**a) Deterministic repro:** Yes. The 5-line repro from #105355 (`CREATE TABLE t (c0 UInt32, INDEX idx c0 TYPE set(100)) ENGINE = MergeTree() ORDER BY c0; INSERT INTO t VALUES (1); SELECT c0 FROM t WHERE c0 + 0.1;`) throws `BAD_ARGUMENTS` on every run before the fix. Same shape for `WHERE log(c0)`, `WHERE c0 + toBFloat16(0.1)`, and `WHERE f` (direct `Float32` column).

**b) Root cause:** `MergeTreeIndexConditionSet::traverseDAG` wraps every WHERE atom with `__bitWrapperFunc` to coerce it to a `BoolMask`. `__bitWrapperFunc::apply` is templated on the column native type `A` and contains `if constexpr (!is_integer<A>) throw "It's a bug!"`. When the atom result type is `Float`, `BFloat16`, or any other non-integer, the wrapper is added to the index DAG and throws at execution time, even though the same query without the index works through the regular filter path.

**c) Fix matches root cause:** Check `WhichDataType(removeNullable(atom_node_ptr->result_type)).isInteger()` before wrapping with `__bitWrapperFunc`. For non-integer atom result types, fall back to `UNKNOWN_FIELD` (the same sentinel the function already uses for atoms it cannot represent). This precisely targets the wrapper site and does not weaken the index for integer atoms.

**d) Test intent preserved + new test added:** No existing tests modified. New regression test `04258_set_skip_index_float_where_issue_105355.sql` covers `Float64` atoms from arithmetic (`c0 + 0.1`), from `log` and `-log`, from `BFloat16` arithmetic, and from a direct `Float32` INPUT, plus sanity checks that integer atoms (`c0 = 2`, `c0 > 0`) still use the index. Locally 50/50 passes with full randomization. All 10 pre-existing set-skip-index tests (`00907_*`, `00979_*`, `00997_*`, `01450_*`, `01917_*`, `02131_*`, `02499_*`, `02670_*`) still pass.

**e) Both directions demonstrated:**
- Before fix: `Code: 36. DB::Exception: It's a bug! Only integer types are supported by __bitWrapperFunc` on the 5-line repro and on each of the four variants above.
- After fix: every variant returns the row that matches the filter (or empty for `log(1) == 0`), consistent with the no-index baseline.

**f) Fix is general, not a narrow patch:** The check is at the single wrap site in `traverseDAG`, where `__bitWrapperFunc` is added for INPUT/FUNCTION/COLUMN atoms. The same bug pattern does not exist in `operatorFromDAG` (it only wires `__bitBoolMaskAnd`/`__bitBoolMaskOr`/`__bitSwapLastTwo` over already-wrapped children that are guaranteed `UInt8`). No other call site adds `__bitWrapperFunc` to a user-supplied predicate.

Session: `cron:clickhouse-ci-task-worker:20260519-213000`.

groeneai:
cc @rschu1ze @alexey-milovidov for review (comp-skip-index, MergeTreeIndexSet wrapping logic). cc @PedroTadim — closes #105355 you tagged me on.

Test-only change to verify: just runs the regression test added in this PR. C++ change is a single guarded fallback in `MergeTreeIndexConditionSet::traverseDAG` (~6 effective lines).

alexey-milovidov (approved):

PR #105386, opened 2026-05-19
Fix clickhouse-client segfault in Connection::poll after sendQuery failure

### LINKED ISSUE #105292: Clickhouse-client Segfault when executing sql
### Company or project name

_No response_

### Describe what's wrong

I have table and sql script insert data to table. 
insert data - `INSERT INTO db.r (r_id,pre_id,pres_id,t_id,t_name,dis,reg,r_c,ciw,ciw_id,pp_t,op_t,
    c_at,e_at,dur,snap,irr,ire,iqa,iree,inr,ilr,
    ipr,ier,rf,vss,praid,eraid,nraid,prat,
    erat,npsrat,npsres,version,i_del,i_part_aure,preip) VALUES (...), (...)`

File size 1139MB and 19978 rows.
When I run clickhouse-client < /file.sql I get an error:

`########## Short fault info ############
(version 26.3.10.62 (official build), build id: 6EB1953AFAD1495BF822A2E45C41166D34390877, git hash: e1c11930c28196f954a93287e43c1aa112c8c607, architecture: x86_64) (from thread 138660) Received signal 11 (signal sent by pid 264 from user 0)
Signal description: Segmentation fault
Address: 0x108. Access: read. Address not mapped to object.
Stack trace: 0x000000001e73d648 0x000000001e6f25c3 0x000000001e6eec91 0x000000001e6fbef9 0x000000001e6fe079 0x000000001e709625 0x0000000017c7466f 0x000000002426dc31 0x0000000017c86f12 0x000000000fbd184a 0x00007e4ced02a1ca 0x00007e4ced02a28b 0x000000000967bbae
########################################
(version 26.3.10.62 (official build), build id: 6EB1953AFAD1495BF822A2E45C41166D34390877, git hash: e1c11930c28196f954a93287e43c1aa112c8c607) (from thread 138660) (no query) Received signal Segmentation fault (11)
Address: 0x108. Access: read. Address not mapped to object.
Stack trace: 0x000000001e73d648 0x000000001e6f25c3 0x000000001e6eec91 0x000000001e6fbef9 0x000000001e6fe079 0x000000001e709625 0x0000000017c7466f 0x000000002426dc31 0x0000000017c86f12 0x000000000fbd184a 0x00007e4ced02a1ca 0x00007e4ced02a28b 0x000000000967bbae
2.0. inlined from ./src/IO/ReadBufferFromPocoSocketChunked.cpp:47: DB::ReadBufferFromPocoSocketChunked::poll(unsigned long) const
2. ./ci/tmp/build/./src/Client/Connection.cpp:1313:9: DB::Connection::poll(unsigned long) @ 0x000000001e73d648
3. ./ci/tmp/build/./src/Client/ClientBase.cpp:1439:29: DB::ClientBase::receiveResult(boost::intrusive_ptr<DB::IAST>, int, bool) @ 0x000000001e6f25c3
4.0. inlined from ./ci/tmp/build/./src/Client/ClientBase.cpp:1344: DB::ClientBase::processOrdinaryQuery(String, boost::intrusive_ptr<DB::IAST>)
4. ./ci/tmp/build/./src/Client/ClientBase.cpp:2409:17: DB::ClientBase::processParsedSingleQuery(std::basic_string_view<char, std::char_traits<char>>, boost::intrusive_ptr<DB::IAST>, bool&, unsigned long) @ 0x000000001e6eec91
5. ./ci/tmp/build/./src/Client/ClientBase.cpp:2786:21: DB::ClientBase::executeMultiQuery(String const&) @ 0x000000001e6fbef9
6. ./ci/tmp/build/./src/Client/ClientBase.cpp:3027:12: DB::ClientBase::processQueryText(String const&) @ 0x000000001e6fe079
7. ./ci/tmp/build/./src/Client/ClientBase.cpp:3851:13: DB::ClientBase::runNonInteractive() @ 0x000000001e709625
8. ./ci/tmp/build/./programs/client/Client.cpp:439:9: DB::Client::main(std::vector<String, std::allocator<String>> const&) @ 0x0000000017c7466f
9. ./ci/tmp/build/./base/poco/Util/src/Application.cpp:315:8: Poco::Util::Application::run() @ 0x000000002426dc31
10. ./ci/tmp/build/./programs/client/Client.cpp:1324:23: mainEntryClickHouseClient(int, char**) @ 0x0000000017c86f12
11. ./ci/tmp/build/./programs/main.cpp:402:21: main @ 0x000000000fbd184a
12. __libc_start_call_main @ 0x000000000002a1ca
13. __libc_start_main_alias_2 @ 0x000000000002a28b
14. _start @ 0x000000000967bbae
ClickHouse version 26.3.10.62 is old and should be upgraded to the latest version.`

But if you install the client version 25.1.3.23, there is no such error.

Yes

### How to reproduce

CREATE TABLE db.r
(
    `r_id` UUID,
    `pre_id` UUID,
    `pres_id` UUID,
    `t_id` UUID,
    `t_name` String,
    `reg` String,
    `pp_t` String,
    `op_t` String,
    `c_at` DateTime64(3),
    `e_at` DateTime64(3),
    `dur` Int64,
    `snap` String,
    `irr` Bool,
    `ire` Bool,
    `iqa` Bool,
    `iree` Bool,
    `inr` Bool,
    `ilr` Bool,
    `ipr` Bool,
    `ier` Bool,
    `rf` String,
    `praid` UUID,
    `eraid` UUID,
    `nraid` UUID,
    `prat` Int64,
    `erat` Int64,
    `npsrat` Int64,
    `npsres` Float64,
    `version` String,
    `i_del` Bool,
    `i_part_aure` Bool DEFAULT 0,
    `preip` String DEFAULT '',
    `vss` String DEFAULT '',
    `dis` String DEFAULT '',
    `r_c` String DEFAULT '',
    `ciw` String DEFAULT '',
    `ciw_id` UUID
)
ENGINE = MergeTree
PARTITION BY toYYYYMMDD(c_at)
PRIMARY KEY r_id
ORDER BY (r_id,
 pre_id,
 t_id,
 pres_id,
 dur,
 reg)
SETTINGS index_granularity = 8192;`

### Expected behavior

`

_No response_

### Additional context

_No response_

### Version info
- Resolved by: #105386
- Merged into: `26.7.1.36`

groeneai:
Adds a regression test for https://github.com/ClickHouse/ClickHouse/issues/105292.

The original `clickhouse-client` SIGSEGV at `Address 0x108` in `ReadBufferFromPocoSocketChunked::poll` (when an INSERT was interrupted by a broken pipe / `ECONNRESET` mid-write and `Connection::sendQuery`'s `SCOPE_EXIT` reset `in` to nullptr before `ClientBase::processOrdinaryQuery`'s `NetException` catch called `receiveResult`) is fixed in #105996, which centralizes the disconnected-connection guard in `Connection::ensureConnected` and adds an early-return at the top of `ClientBase::receiveResult`.

Per @azat's directive (https://github.com/ClickHouse/ClickHouse/pull/105386#discussion_r3356873240), this PR is now trimmed to test-only: it adds a stateless regression test that drives the exact failure path through a single-shot Python TCP proxy that emits a `RST` after a configurable byte cutoff, so any future regression that re-introduces the null deref on the disconnected receive path is caught by `04319_client_segfault_send_error_105292`.

Verified locally against post-#105996 master: 5/5 iterations of the test report `CLIENT_EXIT=210` with `Code: 210. DB::NetException: Connection reset by peer ... (NETWORK_ERROR)` and no `Address: 0x108` / `Segmentation fault` markers.

Closes: https://github.com/ClickHouse/ClickHouse/issues/105292
Related: https://github.com/ClickHouse/ClickHouse/pull/105996

Not required (CI Fix or Improvement).

- [x] Documentation is unchanged because this is a regression test only. The user-visible fix lives in #105996.

### Version info
- Merged into: `26.7.1.36`


groeneai:
## Pre-PR validation gate

a) **Deterministic repro?** YES. A 50-line Python TCP proxy that closes the client→server stream after 200 KB reproduces SIGSEGV 100% of the time before the fix and 0% after. The same proxy is used as the regression test (`04258_client_segfault_insert_send_failure_105292.sh`).

b) **Root cause explained?** YES. `Connection::sendQuery` (`src/Client/Connection.cpp:909`) registers `SCOPE_EXIT { if (!completed) disconnect(); }`. When a buffered write to `*out` throws `NetException`, that `SCOPE_EXIT` fires first and resets `in` (`src/Client/Connection.cpp:421-428`). The `catch (NetException)` block in `ClientBase::processOrdinaryQuery` (`src/Client/ClientBase.cpp:1356-1364`) then calls `receiveResult` to drain residual data, which calls `Connection::poll`, which dereferences the now-null `in`. Address `0x108` matches the offset of the `chunked` member accessed at line 47 of `ReadBufferFromPocoSocketChunked::poll` through a null `this`.

c) **Fix matches root cause?** YES. Two complementary changes:
 - `Connection::poll`, `hasReadPendingData`, `checkPacket`, `receivePacketType` — handle null `in` the way `Connection::receivePacket` already does (existing guard at line 1359 in the same file).
 - `ClientBase::processOrdinaryQuery` — skip the receive-buffer drain when `connection->isConnected()` returns false, so the original `NetException` propagates with its actual cause (`Broken pipe`, `Connection refused`, etc.).

d) **Test intent preserved? / New test added?** YES. No existing tests were weakened. Added `04258_client_segfault_insert_send_failure_105292.sh` which drives the exact bug scenario.

e) **Demonstrated in both directions?**
 - WITHOUT fix: `exit 139, Address: 0x108, Stack: ReadBufferFromPocoSocketChunked::poll → Connection::poll → ClientBase::receiveResult` — full crash stack matches user report.
 - WITH fix: `exit 210, Code: 210. DB::NetException: I/O error: Broken pipe, while writing to socket (...)` — clean error to the user.

f) **Fix is general, not a narrow patch?** YES. The fix covers ALL four sibling methods on `Connection` that can be called by `ClientBase` after a `disconnect`, not only the one in the user's stack:
 - `poll` — fixed (the crash site).
 - `hasReadPendingData` — fixed (called from `Connection::checkPacket` and `MultiplexedConnections::tryGetReadyReplica`).
 - `checkPacket` — fixed (called from `ClientBase::sendDataFromPipe:2181` and `receiveLogsAndProfileEvents:2223,2230`).
 - `receivePacketType` — fixed (matches the existing guard in `receivePacket`).

The "fix the symptom or the cause?" check: the root cause is the SCOPE_EXIT in `sendQuery`. We do not change that (the disconnect is intentional — half-sent queries must not be reused). Instead we make the receive API contract null-safe and update the one caller that wrongly assumed the connection was still alive.

Session: `cron:clickhouse-ci-task-worker:20260519-223000`

groeneai:
cc @azat @alexey-milovidov — could you review this? It addresses a SIGSEGV in clickhouse-client (Connection::poll dereferencing null in after SCOPE_EXIT-triggered disconnect in sendQuery). The receive-side methods now match the existing null guard in Connection::receivePacket, and processOrdinaryQuery skips the drain attempt when the connection is already gone so the original NetException is preserved.

groeneai:
Pushed fix-up commit ba7c1374c550 addressing the CI feedback from the previous run:

1. **Style check** failed because the test file name contained the substring `fail` (rejected by `ci/jobs/check_style.py:181`). Renamed the test from `04258_client_segfault_insert_send_failure_105292.sh` to `04259_client_segfault_send_error_105292.sh` (also drops the 04258 number collision with #105387).
2. **Bugfix validation (functional tests)** ran the test against the pre-fix master binary and the test passed (0.9s), meaning the reproducer was not aggressive enough to trigger the crash in CI. Strengthened it:
   - INSERT VALUES payload increased from 20k to 200k rows (~40 MB) so the client write loop overflows any realistic loopback send buffer.
   - Proxy now sets `SO_LINGER` with timeout 0 after the cutoff, forcing a TCP RST so the client's next write observes `ECONNRESET` immediately instead of waiting for the kernel to drain the auto-tuned send buffer.
   - Stop the server-to-client relay once the cutoff fires.

Verified locally against the rebuilt binary:
- 10/10 OK against the fixed binary (~1.8s per run).
- Deterministic SIGSEGV at `Address: 0x108` in `ReadBufferFromPocoSocketChunked::poll` when run against the pre-fix binary (matches the original crash signature on #105292).

PR description, root cause, and fix unchanged. Awaiting CI.

src/Client/Connection.cpp
```diff
@@ -1310,12 +1310,19 @@ std::optional<Poco::Net::SocketAddress> Connection::getResolvedAddress() const
 
 bool Connection::poll(size_t timeout_microseconds)
 {
+    /// `in` can be null when `Connection::disconnect` was called from `sendQuery`'s
+    /// `SCOPE_EXIT` after a send failure. Surface as a network error rather than
+    /// dereferencing a null pointer.
+    if (!in)
+        throw NetException(ErrorCodes::NETWORK_ERROR, "Connection to {} is terminated", getDescription());
```
azat: The modification in the `ClientBase::processOrdinaryQuery` should be enough

groeneai:
Reverted the null-safety guards in `Connection::poll`, `hasReadPendingData`, `checkPacket`, `receivePacketType` per your note. Audited the callers:

- `Connection::poll`: only `ClientBase::receiveResult` (line 1466).
- `Connection::checkPacket`: `ClientBase::receiveLogsAndProfileEvents`, `ClientBase::sendDataFrom`, `RemoteInserter::write`. None of these can reach a disconnected `Connection` (none of them run after `sendQuery`'s `SCOPE_EXIT` disconnect path).
- `Connection::hasReadPendingData`: `PacketReceiver`, `HedgedConnections`, `MultiplexedConnections`. Same: their lifecycles do not cross a `sendQuery`-failure disconnect.
- `Connection::receivePacketType`: only `MultiplexedConnections::receivePacketTypeUnlocked`, with `current_connection != nullptr`.

So the `isConnected()` guard in `ClientBase::processOrdinaryQuery` is sufficient.

Commit: c6015601043b

groeneai:
Tightened the assertions in `04259_client_segfault_send_error_105292.sh` (commit c6015601043b):

```bash
if [ "${CLIENT_EXIT}" = "0" ]; then
    echo "BAD: proxy did not trigger the send-error path (CLIENT_EXIT=0)"
    head -50 "${CLIENT_LOG}"
    exit 1
fi
if ! grep -qE "NETWORK_ERROR|Broken pipe|Connection reset by peer" "${CLIENT_LOG}"; then
    echo "BAD: no network-error marker in client output (CLIENT_EXIT=${CLIENT_EXIT})"
    head -50 "${CLIENT_LOG}"
    exit 1
fi
```

Now the test fails unless the client both exits non-zero AND the log contains a network-error marker. A clean INSERT (proxy cutoff missed, send buffer absorbed the payload) will no longer pass.

Also added `--allow_repeated_settings --send_logs_level=none` to the proxied client invocation. Server-side log packets sent during the abrupt close were landing in the runner's fatal-log file and unrelatedly failing the test on `asan_ubsan` and `tsan`. Disabling log forwarding for this one call keeps the fatal-log file clean without changing what is being tested.

tests/queries/0_stateless/04259_client_segfault_send_error_105292.sh
```diff
@@ -0,0 +1,197 @@
...
+import struct
+import sys
+import threading
+
+target_host = sys.argv[1]
+target_port = int(sys.argv[2])
+cutoff = int(sys.argv[3])
+port_file = sys.argv[4]
+
+# Force an RST (TCP reset) on close so the client gets ECONNRESET on its
```
azat: Can't we just use clickhouse-client with long list of Values and maybe some throttling?
This custom script bundled into .sh is hacky, at least it should be in a separate file, or better implement the first suggestion
Remember, this will be read by humans

den-crane:
at the server I see 

```
2026.05.20 13:32:49.654453 [ 96976 ] {} <Error> TCPHandler: Code: 131. DB::Exception: Too large string size. (TOO_LARGE_STRING_SIZE), Stack trace (when copying this message, always include the lines below):

0. DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0x000000001334b110
1. DB::Exception::Exception(String&&, int, String, bool) @ 0x000000000dd4d158
2. DB::Exception::Exception(PreformattedMessage&&, int) @ 0x000000000dd4c9cc
3. DB::Exception::Exception<>(int, FormatStringHelperImpl<>) @ 0x0000000013349954
4. DB::readStringBinary(String&, DB::ReadBuffer&, unsigned long) @ 0x00000000134ac910
5. DB::TCPHandler::processQuery(std::shared_ptr<DB::QueryState>&) @ 0x0000000018d65e80
6. DB::TCPHandler::runImpl() @ 0x0000000018d505bc
7. DB::TCPHandler::run() @ 0x0000000018d70e88
8. Poco::Net::TCPServerConnection::start() @ 0x000000001db94858
9. Poco::Net::TCPServerDispatcher::run() @ 0x000000001db94e64
10. Poco::PooledThread::run() @ 0x000000001db53ef4
11. Poco::ThreadImpl::runnableEntry(void*) @ 0x000000001db520f0
12. httplib::detail::random_string(unsigned long)::engine @ 0x0000000000007624
13. ? @ 0x00000000000d166c
```

It looks like there are 2 separate issues:

1. issue with super long `insert into r values ...`, though it should be streamable, and the length should not be matter. It's suspicious that `clickhouse-client -q 'insert into r values' < values.txt` and  `clickhouse-client < insert_and_values.txt` behave differently. And server trows `Too large string size.`.

2. client segfault if the server threw `Too large string size.`.

PedroTadim:
@groeneai check CI failures. `04259_client_segfault_send_error_105292` test failed, plus merge with master.

groeneai:
Pre-PR validation gate (after fix-up df186d477e4e):

- (a) Deterministic repro? YES. Pre-fix master binary SIGSEGVs at Address 0x108 in `ReadBufferFromPocoSocketChunked::poll:47` on every run of the regression test.
- (b) Root cause? YES. `Connection::sendQuery`'s SCOPE_EXIT calls `disconnect()` on send failure, which sets `in` to nullptr. The `catch (NetException)` in `processOrdinaryQuery` then called `receiveResult` which dereferenced the null read buffer.
- (c) Fix matches root cause? YES. Guard `receiveResult` on `connection->isConnected()` in `processOrdinaryQuery`. Single upstream guard; no other callers reach the disconnected state (caller audit in commit c6015601043b).
- (d) Test intent preserved? YES. The test asserts both non-zero exit AND a `NETWORK_ERROR` / `Broken pipe` / `Connection reset by peer` marker in client output, so a clean INSERT cannot silently pass.
- (e) Both directions demonstrated? YES. Local master binary SIGSEGVs deterministically; rebuilt binary with the fix passes 10/10 with empty stderr.
- (f) Fix general? YES. Same `isConnected()` pattern already exists at `ClientBase::cancelQuery` (line 1546+); we are extending it to the symmetric receive path.

groeneai:
@PedroTadim master merged into the branch as commit `403bede0ab89`. The `04259_*` test number collided with #105439 (which just landed `04259_count_distinct_optimization_nullable.sql` on master), so renumbered to `04260`.

CI failure classification on the previous commit `c6015601043b`:

- `Stateless tests (amd_tsan, s3 storage, sequential, 1/2)` and `Stateless tests (amd_tsan, sequential, 1/2)` on `04259_client_segfault_send_error_105292`: the test stdout was `OK` (the bug WAS caught: client exits non-zero with `NETWORK_ERROR`, no `0x108` signature). The failure was stderr containing the line `Killed                  python3 "${PROXY_SCRIPT}" ...` from bash job-control when the background proxy was reaped with `kill -9`. The runner reads stderr and reports any content as a test failure.
- `Bugfix validation (functional tests)`: same cause (the regression test's stderr contained the `Killed` line).

Fix-up `df186d477e4e` resolves this by:
- Moving the proxy to `tests/queries/0_stateless/helpers/04260_client_segfault_proxy.py`.
- Making the proxy single-shot (listens, accepts one connection, RSTs after cutoff, exits). No external `kill` is needed, so bash never emits the `Killed` notification.

Verified locally: 10/10 OK with empty stderr.

groeneai:
@azat moved the Python proxy out of the `.sh` and into a separate helper file: `tests/queries/0_stateless/helpers/04260_client_segfault_proxy.py`. The proxy now has its own docstring and a clear single-purpose contract (relays one connection, RSTs the client after `cutoff` bytes, exits). The `.sh` is 100 lines of orchestration with compact comments.

On the "long Values + throttling" alternative: I traced the path the original den-crane reproducer uses. The server-side trigger is `readStringBinary(state->query, *in)` at `TCPHandler.cpp:2312`, which uses `DEFAULT_MAX_STRING_SIZE = 1_GiB`. Triggering `TOO_LARGE_STRING_SIZE` deterministically without a proxy requires sending a query packet whose size prefix is greater than 1 GiB, which is not viable for a CI regression test (data volume, runtime). `max_network_bandwidth_*` settings only apply after authentication, so they cannot starve the protocol read before the server has accepted the connection. The single-shot proxy is the minimum-data way to deterministically reach the `Connection::sendQuery` `SCOPE_EXIT` path.

Also made the proxy single-shot so the test no longer needs `kill -9` for cleanup. The bash `Killed` job-control notification was what the CI runner was reporting as stderr noise on the `amd_tsan` and `amd_tsan, s3 storage` jobs.

PedroTadim:
@groeneai the `04260_client_segfault_send_error_105292` test failed

azat:
>issue with super long insert into r values ..., though it should be streamable, and the length should not be matter. It's suspicious that clickhouse-client -q 'insert into r values' < values.txt and clickhouse-client < insert_and_values.txt behave differently. And server trows Too large string size..

@den-crane probably due to async INSERTs, do you have a stacktrace?

azat:
@groeneai bug fix validation failed

azat (commented):
In general looks good, but need to fix the test

src/Client/ClientBase.cpp
```diff
@@ -1434,8 +1434,12 @@ void ClientBase::processOrdinaryQuery(String query, ASTPtr parsed_query)
...
                     cleanupTempFile(parsed_query, out_file);
 
-                // We still want to attempt to process whatever we already received or can receive (socket receive buffer can be not empty)
-                receiveResult(parsed_query, signals_before_stop, settings[Setting::partial_result_on_first_cancel]);
+                /// We still want to attempt to process whatever we already received or can receive
+                /// (socket receive buffer can be not empty), but only if the connection is still alive.
+                /// `Connection::sendQuery` resets `in` via its `SCOPE_EXIT` on send failure, so calling
+                /// `receiveResult` after that would dereference a null read buffer (issue #105292).
+                if (connection->isConnected())
+                    receiveResult(parsed_query, signals_before_stop, settings[Setting::partial_result_on_first_cancel]);
```
azat: Will be superseeded by https://github.com/ClickHouse/ClickHouse/pull/105996

Let's wait until it will be merged and leave only test here

groeneai: @azat acknowledged. I'll wait for #105996 to merge, then trim this PR to keep only the regression test (`tests/queries/0_stateless/04259_client_segfault_send_error_105292.{sh,reference}`) and drop the `ClientBase.cpp` / `Connection.cpp` changes.

groeneai:
### Pre-PR validation gate

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | YES. The Python helper `04319_client_segfault_proxy.py` reproduces the original bug 100% of the time on a pre-#105996 binary (server returned the full `INSERT` chunk, proxy emits `RST` at the configured byte cutoff, `clickhouse-client` SIGSEGVs at `Address 0x108` in `ReadBufferFromPocoSocketChunked::poll`). On post-#105996 master, 5/5 iterations exit `210` with `Code: 210. DB::NetException: Connection reset by peer ... (NETWORK_ERROR)`. |
| b | **Root cause explained?** | YES (lives in #105996). `Connection::sendQuery` registers a `SCOPE_EXIT` that calls `disconnect()` if the query did not complete; on a `NetException` from a buffered write, `SCOPE_EXIT` fires and resets `in` to `nullptr`. The `catch (NetException)` in `ClientBase::processOrdinaryQuery` then calls `receiveResult` which calls `Connection::poll` and dereferences the now-null `in`. The `0x108` offset corresponds to the `chunked` field of `ReadBufferFromPocoSocketChunked` reached through a null `this`. |
| c | **Fix matches root cause?** | N/A for this PR (test-only after the trim). The source-level fix in #105996 directly addresses the mechanism: `Connection::ensureConnected` throws a clean exception on `!in`, and `ClientBase::receiveResult` early-returns when `connection->isConnected()` is false so the catch-block drain attempt is skipped on an already-disconnected connection. |
| d | **Test intent preserved / new tests added?** | New test added: `04319_client_segfault_send_error_105292` plus the supporting `helpers/04319_client_segfault_proxy.py`. The test asserts (1) exit code is not `139` (SIGSEGV), (2) no `Address: 0x108` / `Segmentation fault` marker in the client output, (3) exit code is non-zero (the cutoff actually triggered the send-error path), and (4) the output contains `NETWORK_ERROR` / `Broken pipe` / `Connection reset by peer`. Renumbered `04260` → `04319` (next free slot on master; `04260_*` had collisions with already-merged tests). |
| e | **Both directions demonstrated?** | YES. Pre-#105996 master: SIGSEGV every time (the original symptom and the reason this PR was opened). Post-#105996 master: 5/5 iterations PASS reporting `CLIENT_EXIT=210` and `NETWORK_ERROR`. |
| f | **Fix is general, not a narrow patch?** | N/A for this PR (test-only). #105996's source fix is general — `ensureConnected` is now called from `Connection::poll`, `Connection::receivePacketType`, and `Connection::receivePacket`, and the early-return in `ClientBase::receiveResult` covers every receive-side path that used to reach `connection->poll()` after a teardown. |

Session id: cron:clickhouse-ci-task-worker:20260606-062300

groeneai:
@azat #105996 merged at 2026-06-04T15:07:46Z. This PR is now trimmed to test-only as you asked: the `ClientBase.cpp` change is dropped (commit 10e2a88fb67c) and only the `04319_client_segfault_send_error_105292` regression test plus its helper proxy remain.

I verified locally against post-#105996 master that the test PASSes 5/5 iterations: `CLIENT_EXIT=210` with `Code: 210. DB::NetException: Connection reset by peer ... (NETWORK_ERROR)` instead of the original SIGSEGV at `Address 0x108`. So the test catches the bug shape that #105996 fixed, but no longer fails on master itself.

Side effects of the trim:

- The PR no longer modifies any source code, so the `Bug Fix` changelog category cannot be honored — the bug fix validation check would (correctly) fail because the test now passes on master without our PR. I changed the changelog category to `CI Fix or Improvement (changelog entry is not required)` in the PR description; let me know if you'd prefer a different category or no entry.

- I had to renumber the test from `04260` → `04319` because four tests with `04260_*` prefixes have landed on master since this PR was opened. `04319` is the next free slot.

The "bug fix validation failed" comment from earlier (and PedroTadim's `04260_client_segfault_send_error_105292 test failed` notes) should be resolved by this trim — the bug-fix-validation check no longer applies, and the test itself is now stable on master.

tests/queries/0_stateless/helpers/04319_client_segfault_proxy.py
```diff
@@ -0,0 +1,117 @@
...
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0))
+    except OSError:
+        pass
+
+
+def serve_one(client_sock, target_host, target_port, cutoff):
+    try:
+        server_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        server_sock.connect((target_host, target_port))
+    except OSError:
```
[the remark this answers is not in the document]
groeneai: Addressed in 69d1e33d2662 — and your concern surfaced a second silent failure mode that I also fixed.

**Proxy now reports a real signal.** It exits 0 only after forwarding more than `cutoff` bytes and injecting the `RST`. Every other path (target-connect failure, accept timeout, client EOF before cutoff, recv/send error before cutoff) prints a labelled marker (`TARGET_CONNECT_FAILED:`, `ACCEPT_TIMEOUT`, `CLIENT_EOF_BEFORE_CUTOFF:`, etc.) and exits 1.

**Wrapper now waits + asserts.** `wait "${PROXY_PID}"` captures `PROXY_EXIT`, and the script refuses to declare `OK` unless `PROXY_EXIT=0` AND `proxy.log` contains `CUTOFF_EXCEEDED`. So a `clickhouse-client` failure on a different code path (e.g. before `Connection::sendQuery`) can no longer falsely pass the test.

**Bonus root cause this exposed.** While verifying the new assertions locally I discovered the proxy was actually being bypassed entirely in the runner: `${CLICKHOUSE_CLIENT}` already carries the runner-injected `--host=` and `--port=` of the real server, and Boost program_options keeps the FIRST occurrence of `--port`, so the appended `--port=${PROXY_PORT}` was silently dropped. The client connected straight to the server, the INSERT succeeded, `CLIENT_EXIT=0`, and only the existing `CLIENT_EXIT=0` BAD branch happened to catch it. The proxy-bound invocation now uses `${CLICKHOUSE_CLIENT_BINARY}` directly with explicit `--host=127.0.0.1 --port=${PROXY_PORT}` as the only host/port flags.

**Verification.** On a debug build against post-#105996 master:
- 20/20 PASS with the new assertions on the real regression path (`proxy.log: CUTOFF_EXCEEDED: bytes=106559 cutoff=100000`, `client.log: Code: 210. DB::NetException: Connection reset by peer`).
- 1/1 deterministic FAIL when cutoff is set above the INSERT size — the wrapper rejects it with `BAD: proxy did not exercise the mid-write cutoff path` and surfaces `CLIENT_EOF_BEFORE_CUTOFF: bytes=39400246 cutoff=50000000` from `proxy.log`.

Session id: cron:clickhouse-ci-task-worker:20260606-080200

groeneai:
## CI Ledger — covered SHA: `69d1e33d26623e54554f5bf2e5d5ddf99cefdbf9`

CI fully finished (Finish Workflow ✅, every check completed, ≥20min CIDB ingestion buffer met).

| Failure | Classification | Disposition |
|---|---|---|
| `Stateless tests (amd_debug, parallel)` — Logical error: `Virtual row boundary violated in MergingSortedAlgorithm for input 0 on sort column '__table1.timestamp' DESC` (STID 2651-3359) → server died → cascade ERROR on `01280_min_map_max_map` | Trunk regression — NOT PR-caused | Tracked in chronic-trunk task `2026-06-06-p0-stid-2651-3359--virtual-row-boundary`. 30-day cross-PR shows 4 distinct unrelated PRs hit it (this one + #93462, #49966, #106412). Stack lands in `src/Processors/Merges/Algorithms/MergingSortedAlgorithm.cpp:45 checkVirtualRowBoundary`. This PR is test-only (only `tests/queries/0_stateless/04319_client_segfault_send_error_105292.{sh,reference}` and `helpers/04319_client_segfault_proxy.py` after the @ azat-requested trim — fix delivered in #105996, merged 2026-06-04). The crash path has zero overlap with this diff. |

PR-caused gate: zero PR-caused failures. Diff is `tests/queries/0_stateless/...` only; no `src/`, no `Fast test`, no `Bugfix`, no `Style` failures.

Status: waiting_ci → on_review. Reviewer: @ azat (asked for the trim and is the assignee).

alexey-milovidov:
@groeneai, we've fixed this `Logical error: Virtual row boundary violated in MergingSortedAlgorithm` in master. Please update the branch, so it has a chance to be merged.

alexey-milovidov (approved):

alexey-milovidov:
Merged current \`master\` into this branch (\`69d1e33d266..07b70ac075d\`, fast-forward; only the three test files differ from \`master\`).

The branch base was ~18 days / 6542 commits behind \`master\`. The prior CI red — \`Server died\` and \`Logical error: Virtual row boundary violated in MergingSortedAlgorithm ... (STID: 2651-3359)\` in the `Stateless tests (amd_debug, parallel)` job — was a stale-base artifact, not caused by this test-only PR: that debug assertion was added in #106565 (merged 2026-06-05) and the bug it caught was fixed in #106664 (merged 2026-06-07), which the 2026-06-06 base predated. Current `master` contains the fix, so the merge should clear it.

The regression test now runs against a `master` that already contains the user-visible fix (the `ClientBase::receiveResult` early-return guard for a disconnected connection, introduced in #105996), so `04319_client_segfault_send_error_105292` exercises the real fixed path.

groeneai:
### CI finish ledger — 07b70ac075d033201101ec7f7b3420f773efb451

CI fully finished and is **green** on this head: 167 checks all terminal (163 completed + 4 success), 0 queued/in_progress/pending, Finish Workflow + Mergeable Check SUCCESS. Zero test FAIL/ERROR in CIDB on this commit.

| Check / test | Reason | Fixing PR |
|---|---|---|
| (none) | all checks passed | n/a |
| Sync | private | CH Inc sync (not actionable) |

The earlier `Server died` / `Virtual row boundary violated in MergingSortedAlgorithm` red (STID 2651-3359) was a stale-base artifact on the old base (~6542 commits behind; the #106565 debug assertion caught a bug fixed in #106664, both postdating the old base). Merging current master (69d1e33d266..07b70ac075d, only the three 04319 test files differ from master) picked up that fix, and the run is now clean.

This PR is the test-only regression for #105292 (the user-visible fix shipped in #105996). Ready for review/merge.

Session id: cron:our-pr-ci-monitor:20260624-123000

PR #105389, opened 2026-05-20
Pin enable_parallel_replicas=0 in 02516_projections_and_context

groeneai:
The test `02516_projections_and_context` fails on master under the CI's parallel replicas randomization (issued via `automatic_parallel_replicas_mode = 2`, which turns on `enable_parallel_replicas` and `parallel_replicas_for_non_replicated_merge_tree` automatically).

When both `enable_parallel_replicas=1` and `parallel_replicas_for_non_replicated_merge_tree=1` are active, the malformed `dictHas(NULL, ...)` queries the test deliberately issues fail with `THERE_IS_NO_COLUMN` (code 8) inside `ActionsDAG::makeConvertingActions` (pipeline column matching), before reaching the `dictHas` argument validation that is supposed to produce the expected `BAD_ARGUMENTS` (code 36).

CI report (master, distributed plan, parallel, 2/2):
https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=8d3cf431a93cddf5d608c97169eb421fe3aa33b4&name_0=MasterCI&name_1=Stateless%20tests%20%28amd_asan_ubsan%2C%20distributed%20plan%2C%20parallel%2C%202%2F2%29

`--diagnose-random-settings` confirms the failure is fully deterministic with these two settings: `Runs: 43, Failed: 43`; passes 85/85 without them.

This test exercises projections / old-planner context, not parallel replicas error semantics, so pinning `enable_parallel_replicas = 0` is the targeted fix. The two `enable_analyzer = 0` and `enable_analyzer = 1` paths are preserved as-is. A separate engine-level inconsistency may still be worth investigating: `dictHas` argument validation runs before column resolution on the normal path but after pipeline column matching under parallel replicas, surfacing as a different error code.

Fix flaky test `02516_projections_and_context`.

- [x] Documentation is unchanged

### DIFF
--- a/tests/queries/0_stateless/02516_projections_and_context.sql
+++ b/tests/queries/0_stateless/02516_projections_and_context.sql
@@ -1,6 +1,10 @@
 DROP TABLE IF EXISTS test1__fuzz_37;
 CREATE TABLE test1__fuzz_37 (`i` Date) ENGINE = MergeTree ORDER BY i;
 insert into test1__fuzz_37 values ('2020-10-10');
+-- Disable parallel replicas: randomized PR rewrite masks the expected
+-- `BAD_ARGUMENTS` from `dictHas` with a `THERE_IS_NO_COLUMN` from pipeline
+-- column matching.
+set enable_parallel_replicas = 0;
 set enable_analyzer = 0;
 set optimize_functions_to_subcolumns = 1;
 set optimize_use_projections = 0;


groeneai:
Pre-PR validation gate:

a) Deterministic repro: yes on CI (43/43 with the two minimized settings, 85/85 pass without them, per `--diagnose-random-settings`). Cannot reproduce locally even with the full CI settings list applied via `--client-option` (debug + ASan/UBSan local), so the trigger involves something specific to the praktika "distributed plan" CI environment beyond the two named flags; the CI minimizer is the source of truth.

b) Root cause: with `enable_parallel_replicas=1` + `parallel_replicas_for_non_replicated_merge_tree=1`, the planner takes the parallel replicas path. On that path the pipeline header for the GROUP BY keys is built from the rewritten plan, but a sub-tuple of one of the keys (`('', materialize(NULL))`) is referenced in the converting actions for the source stream without being produced by the source. `ActionsDAG::makeConvertingActions` then throws `THERE_IS_NO_COLUMN` before `dictHas` ever validates its first NULL argument, so the expected `BAD_ARGUMENTS` is masked.

c) Fix matches root cause: pinning `enable_parallel_replicas = 0` blocks the parallel replicas code path, restoring the original execution order so `dictHas` validation runs and `BAD_ARGUMENTS` is raised as the test asserts.

d) Test intent preserved: the test is about projections + `optimize_functions_to_subcolumns` context in both the old planner (`enable_analyzer = 0`) and the new planner (`enable_analyzer = 1`). Pinning a single orthogonal setting that the test never depended on does not change what it catches.

e) Demonstrated in both directions: pre-fix CI repro is `43/43 FAIL` (machine-confirmed via `--diagnose-random-settings`). Post-fix locally: `30/30` passes under default randomization, `10/10` passes with the full CI settings list applied, `20/20` passes when the culprit flags are explicitly forced via `--client-option enable_parallel_replicas=1 parallel_replicas_for_non_replicated_merge_tree=1` (the in-test `SET` overrides the client flag).

Note: a separate engine-level inconsistency (validation order under parallel replicas vs the normal path) is visible behind this failure but is out of scope for the test fix.

Session: cron:clickhouse-ci-task-worker:20260520-011500

groeneai:
cc @devcrafter - could you review this? Pinning `enable_parallel_replicas = 0` in a fuzz-derived projections test where CI's parallel replicas randomization masks the expected `BAD_ARGUMENTS` from `dictHas` with a `THERE_IS_NO_COLUMN` from pipeline column matching.

alexey-milovidov (approved):

PR #105397, opened 2026-05-20
Fix TSan data race in naiveBayesClassifier model registry init

groeneai:
Concurrent `SELECT * FROM system.functions` queries invoke `FunctionNaiveBayesClassifier::create(context)` for every function entry. That calls `NBModelRegistry::instance(context)`. When `nb_models` is absent from the config (the common case in CI), `load` throws inside the Meyer's-singleton initialization, the C++ runtime calls `__cxa_guard_abort`, and the next caller re-attempts construction. TSan flags the second attempt's default-construction of `models` against the first attempt's exception-unwind destruction because `__cxa_guard_abort` is not consistently treated as a synchronization edge for the guarded object memory.

Replace the magic-static with `std::mutex` + `std::optional<NBModelRegistry>`. All access to the registry now goes through the mutex, so construction and exception-unwind destruction are happens-before ordered for any subsequent caller. Behavior is otherwise unchanged: a failed `load` rethrows and leaves the optional empty so the next caller retries, and once populated the returned `const Models &` reference is stable.

Surfaced by `BuzzHouse (amd_tsan)` on PR #105386 as STID 0250-34ba (the race is unrelated to that PR's diff).

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105386&sha=ba7c1374c550b32153b098c286f7a49d30a45e67&name_0=PR&name_1=BuzzHouse%20%28amd_tsan%29

Not for changelog.

### Version info
- Merged into: `26.6.1.256`


groeneai:
**Pre-PR validation gate (per `TASK.md` Phase 4):**

**(a) Deterministic repro?** The trigger condition is unambiguous from the TSan trace: two threads simultaneously inside `NBModelRegistry::instance(context)` when `load(context)` throws (e.g., `nb_models` absent from config), one re-attempting initialization while the other is on the exception-unwind path. CIDB shows 1 hit in 60 days under BuzzHouse `amd_tsan`, so on-demand reproduction is impractical at this rate, but the call graph is fully determined from the report.

**(b) Root cause explained?** `system.functions` calls `FunctionFactory::tryGet` for every registered function, which constructs a `FunctionNaiveBayesClassifier` whose initializer enters the Meyer's-singleton `instance(context)`. `load` throws when `nb_models` is missing; the C++ runtime calls `__cxa_guard_abort` and the partially-constructed `models` is destroyed on the same thread. A concurrent caller then re-enters `instance(context)` through the released guard and default-constructs `models` again. TSan reports the second thread's `memset` of the hash-table end node racing with the first thread's `unique_ptr::reset` during exception unwind because `__cxa_guard_abort` is not consistently instrumented as a synchronization edge for the guarded object memory in libcxxabi.

**(c) Fix matches root cause?** Yes. Replacing the magic-static with `std::mutex` + `std::optional<NBModelRegistry>` makes synchronization explicit and observable to TSan: every read or write of the optional is serialized through `mutex`. The construction/destruction edges are now ordinary mutex-acquire / mutex-release pairs that TSan unambiguously honors.

**(d) Test intent preserved?** Yes. Behavior is identical end-to-end: a failed `load` rethrows and leaves the optional empty so the next caller retries; once populated the optional retains its value for the process lifetime and the returned `const Models &` reference is stable. The 5 existing `0_stateless/*_naive_bayes_classifier_*` tests pass 3 runs each (15/15 [OK]); `04212_system_functions_arity_determinism` and `03262_system_functions_should_not_fill_query_log_functions` pass 2 runs each.

**(e) Demonstrated in both directions?** The race is rare (1 hit in 60 days) so a deterministic before-vs-after run is impractical. The fix is provably correct by construction (mutex serializes every access), and the surrounding behavior is preserved by the test runs above. BuzzHouse `amd_tsan` in CI will give the most direct confirmation that the race no longer appears.

**(f) Fix is general?** Yes. The mutex covers every code path that reads or writes `registry`. There is no other singleton or magic-static carrying `models` in this file, and `Models` is only handed out via the returned `const &` after the mutex confirms `registry.has_value()`. I checked the file end-to-end for symmetric singletons or alternate accessors and there are none.

groeneai:
cc @rschu1ze @alexey-milovidov - could you review? Replaces a Meyer's singleton + throwing constructor pattern in `FunctionNaiveBayesClassifier` with explicit `std::mutex` + `std::optional` after a BuzzHouse `amd_tsan` race report (STID 0250-34ba) where two concurrent failed inits of the registry raced. @tavplubix - you reverted earlier work on this file; flagging in case the history is useful.

nihalzp (approved):

PR #105401, opened 2026-05-20
UNIQUE KEY: Bitmap sidecar store, part APIs, and cache wiring (3/N)

murphy-4o:
Part of #103486

Third slice of the UNIQUE KEY stack. Lands the per-part delete-bitmap storage layer on top of PR-2's bitmap primitive + file I/O.

**Why a separate bitmap store.** UNIQUE KEY needs to mark logically-deleted rows without rewriting the part. Each commit publishes one cumulative bitmap per touched part, tagged by a global per-partition CSN. Readers pin a `snapshot_csn` and consult the bitmap with `max csn ≤ snapshot_csn` for each part — that's the snapshot-isolation primitive for UK reads, INSERT-time probe, and merge-time row resolution. This PR delivers the storage and read/write API for that bitmap; the transaction-control plane (CSN allocation, partition mutex, query snapshot, manifest, recovery) lands in later PRs.

**What's in this PR:**
- `MergeTreeBitmapStore` — one per table. Atomic install at a CSN (strictly monotonic; throws `LOGICAL_ERROR` otherwise), snapshot-pinned read returning `(shared_ptr<const DeleteBitmap>, chosen_csn)`, `(V_b, V_next)`-rule GC, and a `dropPart` hook for parts leaving the active set. Keeps an in-memory CSN index per part (rebuilt from disk on first access) so the hot read path is map lookup + binary search, not a syscall.
- `DeleteBitmapCache` — process-wide cache for deserialized bitmaps, keyed by stable part identity + CSN. Configurable via `unique_key_bitmap_cache_{policy,size_bytes,size_ratio}` server settings, wired through `Context` startup/reload, observable via `DeleteBitmapCache{Bytes,Entries}` CurrentMetrics and `UniqueKeyBitmap{Updates,LoadMicroseconds}` ProfileEvents.
- `IMergeTreeDataPart` — cache-identity accessor for callers building cache keys. (Earlier drafts also enumerated `delete_bitmap_<csn>.rbm` in `getFileNamesWithoutChecksums` for BACKUP/replication-fetch; that cross-cutting surface was dropped from this storage-only slice and is owned by the write-integration PR.)

**Why the API is shaped this way:**
- Bitmaps are immutable by CSN — \`readBitmap\` returns \`shared_ptr<const DeleteBitmap>\` so a writer building a new version must copy, never mutate the cached older snapshot.
- No integration with active reads / writes here — those callers (read filter, DELETE, INSERT/UPSERT commit, merge reconcile) land in later slices.


src/Storages/MergeTree/UniqueKey/MergeTreeBitmapStore.cpp
```diff
@@ -0,0 +1,226 @@
...
+        auto it = csns_per_part.find(part_id);
+        if (it == csns_per_part.end())
+            return;
+        csns_to_invalidate = std::move(it->second);
+        csns_per_part.erase(it);
+    }
+
+    if (cache)
+        for (auto v : csns_to_invalidate)
+            cache->remove(DeleteBitmapCache::makeKey(part_id, v));
```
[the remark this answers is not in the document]
murphy-4o: it would not happen actually, dropPart would be serialized at the server level, it would not be read later

src/Storages/MergeTree/UniqueKey/MergeTreeBitmapStore.h
```diff
@@ -0,0 +1,99 @@
...
+        const std::string & part_id,
+        BitmapVersion committed_csn,
+        BitmapVersion oldest_snapshot_csn);
+
+    /// Drop in-memory state for `part_id` and invalidate its cache
+    /// entries. Storage is not touched. Idempotent.
+    void dropPart(const std::string & part_id);
+
+private:
+    DeleteBitmapCache * cache;
```
hanfei1991: Can we use shared ptr for cache?

src/Storages/MergeTree/UniqueKey/MergeTreeBitmapStore.cpp
```diff
@@ -0,0 +1,251 @@
...
+        const std::string & file_name)
+    {
+        ReadSettings read_settings;
+        auto buf = storage.readFile(file_name, read_settings, /*read_hint=*/{});
+        auto deserialized = DeleteBitmap::deserialize(*buf);
+        return std::shared_ptr<DeleteBitmap>(deserialized.release());
+    }
+}
+
+std::vector<BitmapVersion> MergeTreeBitmapStore::snapshotCsns(
```
hanfei1991: maybe `getSnapshotCSNs` be better?

src/Storages/MergeTree/UniqueKey/DeleteBitmap.h
```diff
@@ -102,15 +111,15 @@ class DeleteBitmap
     /// throws on mismatch. Returned bitmap is independent of `in`.
     static std::unique_ptr<DeleteBitmap> deserialize(ReadBuffer & in);
 
-    /// File name convention: `delete_bitmap_{block_number}.rbm`.
-    static std::string fileNameForBlockNumber(UInt64 block_number);
+    /// File name convention: `delete_bitmap_{csn}.rbm`.
+    static std::string fileNameForCsn(BitmapVersion csn);
```
hanfei1991: I would prefer 'CSN' over 'Csn' because CSN is a whole abbr

src/Storages/MergeTree/UniqueKey/DeleteBitmap.cpp
```diff
@@ -423,36 +423,36 @@ std::unique_ptr<DeleteBitmap> DeleteBitmap::deserialize(ReadBuffer & in)
...
-    for (char c : number_part)
+    /// noncanonical name would resolve to the same csn as the canonical one
+    /// and confuse the later read. Require digit-only, then round-trip
+    /// against `fileNameForCsn` to accept only the canonical form.
+    for (char c : csn_part)
         if (c < '0' || c > '9')
             return false;
     UInt64 parsed = 0;
-    if (!tryParse<UInt64>(parsed, number_part))
+    if (!tryParse<UInt64>(parsed, csn_part))
```
hanfei1991, line 530: Emmmm... we got cns here, do we have to call `parseCsnFromFileName` later? We can return csn in this function too.

murphy-4o: 1. The refactoring of the file `DeleteBitmap.cpp` aims to resolve the naming issue: changing `block_number` to `csn`, rather than changing the semantic. Previously i use `block_number`, but actually it's not correct. The versioning of bitmap should be CSN(Commit-Sequence-Number).
2. Another option is to use `std::optional<BitmapVersion> parseCSN()`, in contrast to the current approach of `isDeleteBitmapFile() + parseCSN()`. However, since performance is not a critical concern, separating them into two functions can be more clear.

src/Storages/MergeTree/UniqueKey/DeleteBitmap.h
murphy-4o: Fair. I will update all instances of `Csn` and add a rule for Claude code.

src/Storages/MergeTree/UniqueKey/MergeTreeBitmapStore.cpp
murphy-4o: fair.

src/Storages/MergeTree/UniqueKey/MergeTreeBitmapStore.h
murphy-4o: fixed.

murphy-4o:
Both `installBitmap` contract findings are deferred to the UNIQUE KEY **write-integration PR**, which is where the store first gets a production caller — this slice is storage-only (`installBitmap`/`readBitmap`/`gcObsoleteBitmaps`/`dropPart` are exercised only by gtests today).

- **Cold-part version init (L116).** `csns_per_part` is the in-memory authority for a part's installed versions. The write-integration slice populates it at part-load (from the part directory / `unique_key.txt` manifest) before any install, so install never sees a cold/empty index for a part that already has on-disk sidecars. Re-enumerating disk inside `installBitmap` would defeat that in-memory design, so the preload belongs at part-load — not in `installBitmap`. Documented in the install comment.
- **Post-write index publish (L126).** Tightened together with the commit-path wiring in that same slice.

Both are tracked as explicit obligations on the write-integration PR. This PR is the storage layer only.

hanfei1991 (approved):

PR #105402, opened 2026-05-20
Use exit code 2 from clickhouse-test to signal stop_testing

leshikus:
Follow-up to #105298 ([review thread](https://github.com/ClickHouse/ClickHouse/pull/105298#discussion_r3268844846)) and #104081.

Previously the job-side parser detected the stop_testing condition (server died, hung check failed, `--max-failures` reached, etc.) by matching specific strings in `test_result.txt`. Every wording change in `clickhouse-test` risked silently breaking bugfix validation — exactly what happened when PR #104081 renamed `Server died, terminating all processes...` to `Stopping tests, terminating all processes...`.

Replace that fragile log-line matching with a stable contract: an exit code. `clickhouse-test` now exits with `STOP_TESTING_EXIT_CODE = 2` when the run is aborted via a new `StopTesting` exception (renamed from `TestRunAbort`, mirroring `StopIteration`). All paths that previously set the `stop_testing` event now raise `StopTesting` and unwind to a single outer handler that exits with the new code — including the parent-side hung-check path (which used to `break` after killing workers) and the `create_common_database` failure path (which used to re-raise the original exception, bypassing the handler entirely).

The job-side parser (`functional_tests_results.py`) treats `runner_exit_code == STOP_TESTING_EXIT_CODE` as the authoritative stop_testing signal — no more `STOP_TESTING_SIGN` / `STOP_TESTING_SIGN2` matching. The constant itself is read straight from `tests/clickhouse-test` via `runpy.run_path`, so there is a single source of truth.

`ClickHouseProc.run_test` (used by `fast_test.py`) now returns the integer exit code instead of a bool so this signal reaches the parser for the fast-test invocation too.

Behavioural change worth noting: when `StopTesting` unwinds from `do_run_tests` in the parent-side hung-check path, the end-of-main cleanup steps (final hung check, restarted-tests printing, coverage flush, trace log analysis) are skipped. This is acceptable since the server is unresponsive in that path anyway; the diagnostics would fail.

...

### Version info
- Merged into: `26.5.1.892`

### DIFF
--- a/ci/jobs/fast_test.py
+++ b/ci/jobs/fast_test.py
@@ -344,12 +344,12 @@ def main():
             test_pattern = "|".join(args.test)
             fast_test_command += f" -- '{test_pattern}'"
 
-        res = CH.run_test(fast_test_command)
+        test_exit_code = CH.run_test(fast_test_command)
 
         test_results = FTResultsProcessor(wd=Settings.OUTPUT_DIR).run(
-            runner_exit_code=0 if res else 1,
+            runner_exit_code=test_exit_code,
         )
-        if not res:
+        if test_exit_code != 0:
             attach_debug = True
 
         results.append(test_results)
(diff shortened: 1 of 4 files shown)


maxknv (approved):
LGTM. In follow-ups, it would be good to replace the StopTesting exception with more specific exceptions like ServerDied, Timeout, etc.

PR #105403, opened 2026-05-20
Fix logical error in negative `LIMIT BY`

nihalzp:
The following query has logical error:

```sql
CREATE TABLE t (a Nullable(UInt32), b UInt32, s String, arr Array(UInt32))
ENGINE = MergeTree ORDER BY b SETTINGS allow_nullable_key = 1;
INSERT INTO t VALUES (1, 10, 'x', [100, 200]), (2, 20, 'y', [300, 400]), (NULL, 30, 'z', [500, 600]);
SELECT count() FROM (SELECT x FROM t ARRAY JOIN arr AS x LIMIT -1 BY a);
```

This happens because aggregation hash table cannot handle sparse/replicated columns and negative limit by did not have normalization.

Fix logical error in negative `LIMIT BY` in some cases when used with `ARRAY JOIN`.

### Version info
- Merged into: `26.6.1.126`
- Backported to: `26.5.2.9`


antaljanosbenjamin (approved):

clickgapai:
Hi — could this be a backport candidate for `26.5`? No backport label was found, and the affected code appears to exist in that branch.

**Affected code:** `NegativeLimitByTransform::consume` in src/Processors/Transforms/NegativeLimitByTransform.cpp

**Why this might apply:** The `NegativeLimitByTransform` was introduced on 2026-04-21 (commit b2cf8bb23ef). The bug causes LOGICAL_ERROR exceptions on valid SQL queries using negative LIMIT BY with ARRAY JOIN. Based on file verification, 26.5 contains this code and appears affected.

Other supported branches (`26.4`, `26.3`, `25.8`) appear to have been branched before this feature was introduced.

_Caveat: I verified that the modified code exists in 26.5, but not whether the bug is actually reachable there. If this assessment is incorrect, please ignore; otherwise consider `v26.5-must-backport`._

PR #105406, opened 2026-05-20
Support azure datalake storage gen2 writer

scanhex12:
- New Feature

Support writing into such storage https://learn.microsoft.com/en-us/training/modules/introduction-to-azure-data-lake-storage/2-azure-data-lake-gen2


src/Disks/DiskObjectStorage/ObjectStorages/AzureBlobStorage/AzureObjectStorage.cpp
```diff
@@ -316,12 +343,20 @@ void AzureObjectStorage::removeObjectImpl(
     bool success = false;
     try
     {
-        auto delete_info = client_ptr->GetBlobClient(path).Delete();
-        success = delete_info.Value.Deleted;
-        if (!if_exists && !delete_info.Value.Deleted)
-            throw Exception(
-                ErrorCodes::AZURE_BLOB_STORAGE_ERROR, "Failed to delete file (path: {}) in AzureBlob Storage, reason: {}",
-                path, delete_info.RawResponse ? delete_info.RawResponse->GetReasonPhrase() : "Unknown");
+        if (isAdlsGen2Endpoint(connection_params.endpoint.storage_account_url))
```
asya-ch: Have you run a manual test for deletes? Could you please share details? Need to test both the single-object and multi-object functions.

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.h
```diff
@@ -0,0 +1,75 @@
...
+    BlobStorageLogWriterPtr blob_log;
+};
+
+Azure::Storage::Files::DataLake::DataLakeFileClient makeAdlsGen2FileClient(
+    const AzureBlobStorage::Endpoint & endpoint,
+    const AzureBlobStorage::AuthMethod & auth_method,
+    const Azure::Storage::Blobs::BlobClientOptions & blob_client_options,
+    const String & blob_path);
+
+bool isAdlsGen2Endpoint(const String & storage_account_url);
```
asya-ch: AFAICT every caller passes `connection_params.endpoint.storage_account_url` -- could make it a bit more concise and have it take connection_params. Up to you.

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,250 @@
...
+    return path_part + prefixed + query_part;
+}
+
+}
+
+constexpr size_t ADLFS_MAX_RETRIES = 10000;
+
+bool isAdlsGen2Endpoint(const String & storage_account_url)
+{
+    return storage_account_url.find("fabric.microsoft.com") != String::npos;
```
asya-ch: Can we make the endpoint a bit more specific? Seems to me like it might be overly general.

https://learn.microsoft.com/en-us/fabric/onelake/table-apis/iceberg-table-apis-overview#iceberg-table-api-operations

Are these the correct docs? If so, maybe could do something like `https://onelake.table.fabric.microsoft.com/iceberg`?

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,250 @@
...
+    {
+        try
+        {
+            op();
+            return;
+        }
+        catch (const Azure::Core::RequestFailedException & e)
+        {
+            const bool retryable = isRetryableAzureException(e, write_settings.is_initial_access_check);
+            if (!retryable || attempt >= max_unexpected_write_error_retries)
```
asya-ch, line 171: AFAICT, max_unexpected_write_error_retries is 4 by default, and sdk_retry_initial_backoff_ms is 10ms. We fail very quickly for transient errors for writes -- is it intentional?

Also, would be nice to use exponential backoff, but seems it's not commonly used in the object storage code. (If so, probably out of scope, but would be good to add in the future.)

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,250 @@
...
+    String full = endpoint.prefix;
+    if (!full.empty() && !full.ends_with('/'))
+        full += '/';
+    full += blob_path_;
+    return full;
+}
+
+String buildAdlsGen2FileUrl(const AzureBlobStorage::Endpoint & endpoint, const String & blob_path_)
+{
+    String prefixed = prefixedBlobPath(endpoint, blob_path_);
```
asya-ch: Have we tested it with endpoints both with and without `?` Please provide details and repro instructions (can share credentials offline).

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,250 @@
...
+                    what,
+                    blob_path,
+                    static_cast<int>(e.StatusCode),
+                    e.Message);
+            }
+
+            LOG_WARNING(log, "ADLS Gen2 {} attempt {} for `{}` failed: HTTP {}: {}. Retrying after {} ms.",
+                what, attempt, blob_path, static_cast<int>(e.StatusCode), e.Message, backoff_ms);
+
+            std::this_thread::sleep_for(std::chrono::milliseconds(backoff_ms));
```
asya-ch: Claude tells me `sleepForMilliseconds()` is better because it's interruptible on cancellation.

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.h
```diff
@@ -0,0 +1,75 @@
...
+
+    bool file_created = false;
+    bool is_prefinalized = false;
+    int64_t bytes_appended = 0;
+
+    String container_for_logging;
+    BlobStorageLogWriterPtr blob_log;
+};
+
+Azure::Storage::Files::DataLake::DataLakeFileClient makeAdlsGen2FileClient(
```
asya-ch, line 70: I'm confused whether this logic is more about Gen2 or about OneLake -- please provide some context. :)

scanhex12: About Gen2
OneLake uses Gen2, and this is the only feature that is not supported yet, so this PR is mainly focused on Gen2 support

src/Disks/DiskObjectStorage/ObjectStorages/AzureBlobStorage/AzureObjectStorage.cpp
scanhex12: Sent in DM

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -526,7 +532,7 @@ static bool writeMetadataFiles(
             });
 
             auto hint_path = filename_generator.generateVersionHint();
-            if (!writeMetadataFileAndVersionHint(
+            if (!catalog && !writeMetadataFileAndVersionHint(
```
scanhex12: We don’t need to update both the metadata on storage and the catalog metadata, since the catalog updates the metadata on storage by itself

We also don’t need to use the version hint file as a synchronization point, since the catalog does CAS

src/Databases/DataLake/RestCatalog.h
```diff
@@ -75,6 +75,8 @@ class RestCatalog : public ICatalog, public DB::WithContext
 
     bool isTransactional() const override { return true; }
 
+    bool generatesMetadataFile() const override { return true; }
```
asya-ch: (For anyone else following -- Konstantin asked me to take over the PR while his computer is broken.)

I think this change is a bit risky. I'm working on adding some basic mutation tests for REST catalogs to make sure we don't break them.

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,247 @@
...
+    LOG_INFO(log, "Created ADLS Gen2 file `{}`", blob_path);
+}
+
+void WriteBufferFromAzureDataLakeStorage::appendBufferedData()
+{
+    const size_t to_append = offset();
+    if (to_append == 0)
+        return;
+
+    ensureCreated();
```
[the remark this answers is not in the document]
asya-ch: @scanhex12, let's add this please.

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    const auto * data_ptr = reinterpret_cast<const uint8_t *>(working_buffer.begin());
+    const int64_t offset_for_append = bytes_appended;
+
+    ProfileEvents::increment(ProfileEvents::AzureUpload);
+
+    LOG_INFO(log, "Entering Append for `{}`: offset={}, len={}", blob_path, offset_for_append, to_append);
```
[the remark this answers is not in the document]
asya-ch: @scanhex12, please address this one as well.

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,247 @@
...
+}
+
+void WriteBufferFromAzureDataLakeStorage::preFinalize()
+{
+    if (is_prefinalized)
+        return;
+    is_prefinalized = true;
+
+    LOG_INFO(log, "Entering preFinalize for ADLS Gen2 file `{}`", blob_path);
+    appendBufferedData();
```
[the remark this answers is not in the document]
asya-ch: This is good feedback. ~~@scanhex12, I think we should use the `hidePartialData()` pattern in `writeBufferFromS3()`~~. We should also make sure we come with a test that exercises `nextImpl()` multiple times.

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+    bytes_appended += static_cast<int64_t>(to_append);
+    LOG_INFO(log, "Appended for `{}`: bytes_appended={}", blob_path, bytes_appended);
+}
+
+void WriteBufferFromAzureDataLakeStorage::nextImpl()
+{
+    if (is_prefinalized)
+        throw Exception(
+            ErrorCodes::AZURE_BLOB_STORAGE_ERROR,
```
asya-ch, line 251:
```suggestion
            ErrorCodes::LOGICAL_ERROR,
```

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,302 @@
...
+    return uri.toString();
+}
+
+}
+
+constexpr size_t ADLFS_MAX_RETRIES = 10;
+
+bool isAdlsGen2Endpoint(const AzureBlobStorage::Endpoint & endpoint)
+{
+    return endpoint.storage_account_url.contains("dfs.fabric.microsoft.com");
```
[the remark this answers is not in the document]
asya-ch: @scanhex12 , I still find this confusing -- the goal should be to check that it's a OneLake endpoint, not a Gen2 endpoint, right?

So I would think we'd want to (a) check for `https://onelake.table.fabric.microsoft.com` and (b) rename the function to reflect this. Let me know if I'm missing something.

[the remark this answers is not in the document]
scanhex12: No, the goal is to check that it is FS Gen2, because write buffer don't know anything about onelake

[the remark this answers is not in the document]
scanhex12: I'd prefer not to change this check, it's pretty good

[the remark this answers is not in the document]
scanhex12: I'll add one more check in this function (more details - here https://github.com/ClickHouse/ClickHouse/pull/106347)

src/Disks/IO/WriteBufferFromAzureDataLakeStorage.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+void WriteBufferFromAzureDataLakeStorage::preFinalize()
+{
+    if (is_prefinalized)
+        return;
+    is_prefinalized = true;
+
+    LOG_DEBUG(log, "Entering preFinalize for ADLS Gen2 file `{}`", blob_path);
+    appendBufferedData();
+    WriteBuffer::set(fake_buffer_when_prefinalized, sizeof(fake_buffer_when_prefinalized));
```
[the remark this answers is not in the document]
asya-ch: @scanhex12, I think this should be a one line change -- just needed to add `WriteBufferFromFileBase::finalizeImpl()` to the top.

asya-ch (approved):
Looks good -- please do address the bot comment about adding the `next()` flush inside of `finalizeImpl()`.

clickgapai:
Hi @scanhex12 — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: The PR description is empty apart from the changelog category. It does not explain that the feature only activates for Microsoft Fabric endpoints, why the DataLake SDK is needed, that deletes fall back to single (non-batched) requests for ADLS, or how the feature was validated (no automated tests are included).
- Suggested addition:

  > This PR adds a `WriteBufferFromAzureDataLakeStorage` write path for the Azure object storage disk, used when the configured endpoint is a Microsoft Fabric ADLS Gen2 endpoint (`dfs.fabric.microsoft.com` / `blob.fabric.microsoft.com`). It uses the Azure DataLake SDK `DataLakeFileClient` Create/Append/Flush lifecycle with internal retries. Deletes for ADLS endpoints use single requests (the blob batch API is not used). Note: generic ADLS Gen2 endpoints (`*.dfs.core.windows.net`) are not covered by the detection. Please describe how this was tested against a real Fabric account, since no automated test can exercise the Fabric-only code path.

PR #105412, opened 2026-05-20
darwin fast test: collect macOS .ips crash reports

leshikus:
When ClickHouse or one of its child processes crashes on macOS, the OS writes a `.ips` diagnostic report to `/Library/Logs/DiagnosticReports`. That directory is root-owned, so the CI runner cannot read its contents directly and the reports never reach the fast-test artifacts.

This PR teaches the darwin fast-test job to grab those crash reports:

- During log collection on Darwin, `ClickHouseProc._collect_diagnostic_reports` runs `sudo chown -R \$(id -u):\$(id -g) /Library/Logs/DiagnosticReports` and then attaches every `*.ips` file in place via `Path.glob`. No copying needed.
- The darwin fast-test post-hook wipes `/Library/Logs/DiagnosticReports/*` under sudo so any `.ips` present at the next collection is guaranteed to belong to that run — no timestamp filtering required.

The change is a no-op on non-Darwin platforms because the collector early-returns unless `platform.system() == "Darwin"`.

### Version info
- Merged into: `26.5.1.930`

### DIFF
--- a/ci/defs/job_configs.py
+++ b/ci/defs/job_configs.py
@@ -207,8 +207,7 @@ class JobConfigs:
         force_success=True,
         post_hooks=[
             "python3 ./ci/jobs/scripts/job_hooks/clickhouse_test_cleanup_hook.py",
-            "sudo rm -rf /Users/ec2-user/actions-runner/_work/ClickHouse/ClickHouse/ci/tmp/run* /System/Volumes/Data/System/Library/Caches/com.apple.coresymbolicationd/data",
-            "rm -rf /Library/Logs/DiagnosticReports/*",
+            "sudo rm -rf /Users/ec2-user/actions-runner/_work/ClickHouse/ClickHouse/ci/tmp/run* /System/Volumes/Data/System/Library/Caches/com.apple.coresymbolicationd/data /Library/Logs/DiagnosticReports/*",
         ],
     ).parametrize(
         Job.ParamSet(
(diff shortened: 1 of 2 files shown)


maxknv (approved):

leshikus:
The job that failed is the check_report_messages.py post-hook in Finish Workflow. The hook intentionally fails the workflow whenever any individual job has pushed an error or warning to the workflow-level Result.ext["errors"] / Result.ext["warnings"] lists (ci/jobs/scripts/workflow_hooks/check_report_messages.py:24-33).

  Four stateless-test shards posted Failed to start log export warnings:

  - Stateless tests (amd_debug, distributed plan, s3 storage, parallel)
  - Stateless tests (arm_asan_ubsan, azure, parallel)
  - Stateless tests (amd_tsan, s3 storage, parallel, 1/2)
  - Stateless tests (amd_msan, WasmEdge, parallel, 1/2)

  Those come from ci/jobs/functional_tests.py:509-511 — CH.start_log_exports returned false (the call that ships test-server logs to the central CIDB cluster). Whenever that fails, the shard adds a warning via info.add_workflow_warning("Failed to start log export"), and check_report_messages.py deliberately blocks the workflow so a   human acknowledges it before merging.
  
  It is unrelated to the branch changes — it is a flaky/unavailable log-export endpoint on those four runners.

PR #105418, opened 2026-05-20
Stress test: parse test_results.tsv with stray CR-overwritten dpkg output

Algunenano:
`Stress test (*)` jobs that triggered the hung-check path were reporting `Unknown job error` instead of the real `Hung check failed, possible deadlock found  FAIL`. The cause: `clickhouse-test --hung-check --capture-client-stacktrace` calls `_ensure_lldb_installed`, which runs `apt-get install -y lldb` on the fly when `lldb` is missing from the image. The dpkg progress frames `(Reading database ... 5%\r... 10%\r...)` land in `hung_check.log`, and `stress.py` embeds them verbatim into the `info` field of the `Hung check failed` row. `escape_tsv_info` did not escape bare CR, and Python's universal-newlines mode turned every CR into LF — `csv.reader` then saw 122 single-cell fragments and raised on the first one, hiding every valid row in the file.

`read_test_results` now reads as bytes, splits on LF only, strips CR before TSV parsing, and collects malformed rows in a list instead of raising. `process_results` surfaces the malformed-row count as a separate `Corrupt test_results.tsv` result so the real failures stay visible (and reports `Corrupt test_results.tsv` instead of `Unknown job error` when every row is malformed). `escape_tsv_info` also strips `\r` as defense in depth.

Observed on: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105243&sha=992c19f8b7dfbbdee12180257dca2d4909df5a02&name_0=PR&name_1=Stress%20test%20%28arm_debug%29

### Version info
- Merged into: `26.5.1.945`

### DIFF
--- a/ci/jobs/scripts/stress/stress.py
+++ b/ci/jobs/scripts/stress/stress.py
@@ -20,16 +20,18 @@ class ServerDied(Exception):
 
 
 def escape_tsv_info(text: str) -> str:
-    # Strip CR before escaping the other separators. Bare CR is emitted
-    # by tools like `apt-get`/`dpkg` to overwrite progress frames in
-    # place, and the hung-check path embeds dpkg output verbatim when
-    # `clickhouse-test --capture-client-stacktrace` installs `lldb` on
-    # the fly. Left in the TSV, those CRs are turned back into LF by
-    # universal-newlines mode at read time and fragment the row.
+    # Escape CR alongside the other separators rather than dropping it.
+    # Bare CR is emitted by tools like `apt-get`/`dpkg` to overwrite
+    # progress frames in place, and the hung-check path embeds dpkg
+    # output verbatim when `clickhouse-test --capture-client-stacktrace`
+    # installs `lldb` on the fly. Left raw in the TSV, those CRs are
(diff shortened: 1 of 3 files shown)


ci/jobs/scripts/stress/stress.py
```diff
@@ -20,7 +20,18 @@ class ServerDied(Exception):
...
 def escape_tsv_info(text: str) -> str:
-    return text.replace("\0", "\\0").replace("\t", "\\t").replace("\n", "\\n")
+    # Strip CR before escaping the other separators. Bare CR is emitted
+    # by tools like `apt-get`/`dpkg` to overwrite progress frames in
+    # place, and the hung-check path embeds dpkg output verbatim when
+    # `clickhouse-test --capture-client-stacktrace` installs `lldb` on
+    # the fly. Left in the TSV, those CRs are turned back into LF by
+    # universal-newlines mode at read time and fragment the row.
+    return (
+        text.replace("\r", "")
```
alexey-milovidov: Okay. However, you don't have to cut `\r`, you can escape it as well.

alexey-milovidov (approved):

PR #105442, opened 2026-05-20
Apply `*WithMemoryTracking` containers to Functions

nickitat:

### Version info
- Merged into: `26.6.1.71`


nickitat:
@Algunenano, would you like to review this PR?)

Algunenano (approved):
Really nice. I noticed some extra improvements to clean up some enums (which is great), and some really minor possible improvements (we could use auto when interacting  with some external libraries) but I don't think those deserve comments of fixing them here. We'll work on them as we evolve the code.

PR #105445, opened 2026-05-20
Promote to PREWHERE after pushdowns

yariks5s:
- Improvement

Allow filters introduced after the initial PREWHERE selection (predicate pushdown, runtime filters, or explicit PREWHERE plus a WHERE set by the planner) to be merged into the existing PREWHERE on a second optimizer pass instead of staying as a separate Filter step above the MergeTree read

### Version info
- Merged into: `26.6.1.101`


src/Processors/QueryPlan/Optimizations/optimizeTree.cpp
```diff
@@ -507,6 +507,15 @@ void optimizeTreeSecondPass(
             "Projection {} is specified in setting force_optimize_projection_name but not used",
             optimization_settings.force_projection_name);
 
+    if (optimization_settings.optimize_prewhere && optimization_settings.optimize_prewhere_after_pushdown)
```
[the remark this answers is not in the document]
yariks5s: in QueryPlanOptimizationSettings.cpp -- `optimize_prewhere_after_pushdown = optimize_prewhere && from[Setting::optimize_prewhere_after_pushdown];`

src/Processors/QueryPlan/Optimizations/optimizePrewhere.cpp
```diff
@@ -151,12 +152,13 @@ void optimizePrewhere(QueryPlan::Node & parent_node, const bool remove_unused_co
...
 
-    if (source_step_with_filter->getPrewhereInfo())
-        return;
-
     const auto & context = source_step_with_filter->getContext();
     const auto & settings = context->getSettingsRef();
 
+    PrewhereInfoPtr existing_prewhere_info = source_step_with_filter->getPrewhereInfo();
+    if (existing_prewhere_info && !settings[Setting::optimize_prewhere_after_pushdown])
+        return;
```
Michicosun, line 160: Why not make it unconditionally?

src/Processors/QueryPlan/Optimizations/optimizePrewhere.cpp
```diff
@@ -212,6 +214,55 @@ void optimizePrewhere(QueryPlan::Node & parent_node, const bool remove_unused_co
...
+        if (!new_removable && std::find(outputs.begin(), outputs.end(), new_filter_node) == outputs.end())
+            outputs.push_back(new_filter_node);
+
+        outputs.push_back(and_node);
+
+        prewhere_info->prewhere_actions = std::move(combined);
+        prewhere_info->prewhere_column_name = and_node->result_name;
+        prewhere_info->remove_prewhere_column = true;
+        prewhere_info->need_filter = true;
+    }
```
Michicosun, line 259: Why is it not copied from the new prewhere info or from a combination?

src/Processors/QueryPlan/Optimizations/optimizePrewhere.cpp
```diff
@@ -212,6 +214,55 @@ void optimizePrewhere(QueryPlan::Node & parent_node, const bool remove_unused_co
...
+
+        auto & outputs = combined.getOutputs();
+        std::unordered_set<const ActionsDAG::Node *> existing_outputs(outputs.begin(), outputs.end());
+        for (const auto * node : new_outputs_in_combined)
+        {
+            if (node == new_filter_node)
+                continue;
+            if (existing_outputs.insert(node).second)
+                outputs.push_back(node);
+        }
```
Michicosun: new_outputs_in_combined should not contain duplicates, I suppose it can be simplified with:

```
combined.getOutputs().append_range(new_outputs_in_combined);
```

src/Processors/QueryPlan/Optimizations/optimizePrewhere.cpp
yariks5s: remove_prewhere_column is true unconditionally, because we don't need the merged filter to be kept after applying, but we can do so with the need_filter logic, I will update the code.

src/Processors/QueryPlan/Optimizations/optimizePrewhere.cpp
yariks5s: just to be sure and clear

Michicosun (approved):

yariks5s:
test_mysql_protocol was broken, already fixed

src/Core/Settings.cpp
```diff
@@ -6064,6 +6064,12 @@ Allow to convert `JOIN` to subquery with `IN` if output columns tied to only lef
 )", 0) \
     DECLARE(Bool, query_plan_optimize_prewhere, true, R"(
 Allow to push down filter to PREWHERE expression for supported storages
+)", 0) \
+    DECLARE(Bool, optimize_prewhere_after_pushdown, false, R"(
```
rschu1ze, line 6068: @yariks5s Can/should we turn this setting on (if it stays off long-term then it will be dead code)? Are there any performance regressions?

Also, please rewrite the description, it is not clear for readers what a "promotion pass" is and "depositing filters" should be "storing filters" maybe?

tests/queries/0_stateless/04259_prewhere_after_pushdown.sql
```diff
@@ -0,0 +1,118 @@
+-- Tags: no-random-settings
+
+SET enable_analyzer = 1;
+DROP TABLE IF EXISTS t_prewhere_pushdown;
+
+CREATE TABLE t_prewhere_pushdown
```
rschu1ze, line 6: (A general thing: Please add comments at the beginning of a test that say what the test is supposed to test.)

PR #105449, opened 2026-05-20
Fix heap-buffer-overflow and null-pointer dereference in Arrow geo/string column readers

Algunenano:
Two bugs fixed in `ArrowColumnToCHColumn` affecting Arrow and ArrowStream formats:

**1. Heap-buffer-overflow in `readColumnWithStringData`**

Reading a file with corrupted intermediate offsets caused
`readColumnWithStringData` to call `insert_assume_reserved` with a per-row
length much larger than the actual values buffer.

The pre-flight reserve loop used only the last chunk entry
(`value_offset(N-1) + value_length(N-1)`), which equals the buffer size for
valid data, but the write loop iterated every row with `value_length(i)` and no
bounds check. A file where an intermediate offset is corrupted (e.g. `offsets[1]`
set to 1 GiB while `offsets[N]` stays at 40 bytes) keeps the reserve at ~40
bytes while the first row's `value_length` becomes 1 GiB.

Fix: accumulate `chars_size` per-row during pre-flight, validating
`value_offset + value_length <= buffer->size()` for every non-null row.
Throws `INCORRECT_DATA` on violation.

**2. Null-pointer dereference and heap-buffer-overflow in `readColumnWithGeoData`**

`buffer->mutable_data()` returns `nullptr` for read-only Arrow IPC buffers
(Arrow's contract: non-mutable buffers return `nullptr` from `mutable_data()`).
This caused an immediate null-pointer dereference (UBSan: applying non-zero
offset to null pointer) on any Arrow file with geo footer metadata, regardless
of data content.

Additionally, there was no bounds check before creating a `ReadBuffer` from the
Arrow value buffer. A corrupted intermediate offset entry could make
`value_length(i)` much larger than the actual data buffer, letting
`parseWKBFormat` read arbitrarily far past the end of the heap allocation.

Fix: use `buffer->data()` (always valid) with `const_cast`; validate
`safe_offset + safe_length <= buffer_size` for every non-null row; move pointer
computation after the null-row check.

Closes https://github.com/ClickHouse/clickhouse-private/issues/58383

Fixed a heap-buffer-overflow when reading Arrow or ArrowStream files with corrupted intermediate offsets in a binary or string column, and a null-pointer dereference when reading geo-tagged Arrow columns.

### Version info
- Merged into: `26.6.1.80`
- Backported to: `26.5.2.4`, `26.4.4.9`, `26.3.12.3`, `25.8.25.6`


Avogar (approved):

Algunenano:
I've extended the PR to cover the case detected by the AI review, and when looking at parallel code I noticed that GeoArrow had the same issue and patched that too

PR #105453, opened 2026-05-20
Fix `input_format_json_empty_as_default` not working in JSONStrings* formats

### LINKED ISSUE #104913: `input_format_json_empty_as_default` not implemented for `yield_strings=true` JSON formats

_Retrospective finding from a historical scan of [PR #66782](https://github.com/ClickHouse/ClickHouse/pull/66782) (merged 2024-09-13). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

Setting `input_format_json_empty_as_default=1` does not work for JSONStringsEachRow, JSONCompactStringsEachRow and similar formats - empty strings throw parsing exceptions instead of using defaults

**Root cause:** JSONUtils.cpp:279-291: The `yield_strings` code path reads JSON string and deserializes without checking `format_settings.json.empty_as_default`. The check at line 302-303 is only reached when `yield_strings=false`.

**Why we believe this is a bug:** JSONUtils.cpp:readField (line 279-291) has early return for yield_strings=true path that bypasses the empty_as_default check at line 302-303

**Affected locations:**
- `src/Formats/JSONUtils.cpp:279` — yield_strings early return path

**Impact:** Users enabling `input_format_json_empty_as_default` for JSONStringsEachRow format will get parsing exceptions (e.g. 'Cannot parse date: value is too short') instead of default values

Yes — confirmed on current `master` (commit `c2f4d729254`).

### How to reproduce

```sql
SET input_format_json_empty_as_default = 1;

-- JSONEachRow works correctly
SELECT 'JSONEachRow Date:' AS test, x FROM format(JSONEachRow, 'x Date', '{"x":""}') ORDER BY test;
-- Returns: 1970-01-01 (correct)

-- JSONStringsEachRow does NOT work - throws exception
SELECT 'JSONStringsEachRow Date:' AS test, x FROM format(JSONStringsEachRow, 'x Date', '{"x":""}') ORDER BY test;
-- Throws: Cannot parse date: value is too short (BUG - should return 1970-01-01)
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/02b08bd8-d724-47bb-89a0-c6c517bd525f)

### Expected behavior

```
JSONEachRow Date:	1970-01-01
JSONStringsEachRow Date:	1970-01-01
```

```
JSONEachRow Date:	1970-01-01
Received exception from server (version 26.5.1): Code: 38. DB::Exception: Cannot parse date: value is too short
```

### Additional context

**Open risks:**
- Other yield_strings=true format variants (WithNames, WithNamesAndTypes) also affected

**Suggested fix:** Add empty string check in yield_strings path: after readJSONString, check if str.empty() && format_settings.json.empty_as_default, and if so, call column.insertDefault() and return false

Found during automated review of [PR #66782](https://github.com/ClickHouse/ClickHouse/pull/66782).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr66782_001`_

### Version info
- Resolved by: #105453
- Merged into: `26.5.1.941`

Avogar:

Fix `input_format_json_empty_as_default` not working in JSONStrings* formats. Closes https://github.com/ClickHouse/ClickHouse/issues/104913

### Version info
- Merged into: `26.5.1.941`

### DIFF
--- a/src/Formats/JSONUtils.cpp
+++ b/src/Formats/JSONUtils.cpp
@@ -281,6 +281,12 @@ namespace JSONUtils
                 String str;
                 readJSONString(str, in, format_settings.json);
 
+                if (format_settings.json.empty_as_default && str.empty())
+                {
+                    column.insertDefault();
+                    return false;
+                }
+
                 ReadBufferFromString buf(str);
 
                 if (as_nullable)
--- a/tests/queries/0_stateless/04260_json_empty_as_default_yield_strings.reference
+++ b/tests/queries/0_stateless/04260_json_empty_as_default_yield_strings.reference
@@ -1,0 +1,31 @@
+-- Simple types with yield_strings formats (JSONStringsEachRow, JSONCompactStringsEachRow)
+-- { echoOn }
+SELECT x FROM format(JSONStringsEachRow, 'x Date', '{"x":""}');
+1970-01-01
+SELECT x FROM format(JSONStringsEachRow, 'x Date32', '{"x":""}');
+1970-01-01
+SELECT toTimeZone(x, 'UTC') FROM format(JSONStringsEachRow, 'x DateTime', '{"x":""}');
+1970-01-01 00:00:00
+SELECT toTimeZone(x, 'UTC') FROM format(JSONStringsEachRow, 'x DateTime64', '{"x":""}');
+1970-01-01 00:00:00.000
+SELECT x FROM format(JSONStringsEachRow, 'x IPv4', '{"x":""}');
+0.0.0.0
+SELECT x FROM format(JSONStringsEachRow, 'x IPv6', '{"x":""}');
+::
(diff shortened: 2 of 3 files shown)


Algunenano (approved):

clickgapai:

**Affected code:** `JSONUtils::readField` yield_strings code path in src/Formats/JSONUtils.cpp — introduced in 24.9.

**Why this might apply:** The `input_format_json_empty_as_default` setting was introduced in 24.9 (per SettingsChangesHistory.cpp line 788), which predates all supported branches. The bug exists in all supported releases since the yield_strings code path in JSONUtils::readField never had the empty_as_default check. Users of JSONStringsEachRow format who rely on this setting cannot use it in any stable release.

PR #105463, opened 2026-05-20
Fix data corruption in Dynamic Quoted fallback (issue #105441)

### LINKED ISSUE #105441: Empty array as map key, not serialized back
### Describe the bug

Easy to reproduce, I think it's a bug. Claude analysis at SchemaInferenceUtils.cpp:

```
Line 1793: checkIfTypeIsComplete(map_type->getKeyType()) → checks Array(Nothing) → line 1777 recurses into the nested type → Nothing → line 1770-1771 returns false.

So the full bug chain is:

Map(Array(Nothing), Nullable(String)) is inferred correctly from the text {[]:'came'}
But checkIfTypeIsComplete says it's incomplete because Array(Nothing) contains Nothing
Since no existing variant can parse it, falls back to String (line 849)
The String fallback adds naive quoting (line 851-852): wraps {[]:'came'} → '{[]:'came'}'
This creates a malformed quoted string — the ' in 'came' prematurely terminates the quoted string parse
Result: {[]: (truncated at the inner single quote)
```
Summary:
```
Root cause: Two issues in SerializationDynamic::deserializeTextImpl:

Incomplete type handling: Map(Array(Nothing), ...) is a valid ClickHouse type that was correctly serialized, but the deserialization considers Nothing inside the key type as "incomplete" and refuses to use it. The type was complete when it was written — the Nothing is structural (empty array), not a schema inference gap.

Broken String fallback quoting (line 851-852): When falling back to String for the Quoted escaping rule, the code does "'" + field + "'", which produces a malformed quoted string if the field already contains single quotes. This causes data truncation/corruption — a much worse outcome than just a type mismatch.

The first issue is debatable (should Array(Nothing) count as complete?), but the second issue is a clear data corruption bug — the naive quoting should at minimum escape inner single quotes.
```

### How to reproduce

Run Fiddle: https://fiddle.clickhouse.com/2d07865b-4789-4409-a1d5-5de6aa574582

_No response_

### Version info
- Resolved by: #105463
- Merged into: `26.5.1.936`

groeneai:
`SerializationDynamic::deserializeTextImpl` falls back to `String` when the inferred variant type is incomplete (for example `Map(Array(Nothing), ...)` from `{[]:'came'}`). For the `Quoted` escaping rule the fallback wraps the field in single quotes via naive concatenation `"'" + field + "'"`. If the field already contains a single quote, the resulting quoted string is malformed: `readQuotedStringInto` terminates at the first inner `'` and the stored value is silently truncated.

For the issue reproducer `({[]:'came'})` the round trip produced `{[]:` (4 chars) instead of `{[]:'came'}` (11 chars).

Fix uses `writeQuotedString` so embedded single quotes and backslashes are escaped to match what `readQuotedStringInto<true>` expects.

Closes #105441

Fix silent value truncation in `Dynamic` columns when the `Quoted` (Values) text reader falls back to `String` for incomplete inferred types like `Map(Array(Nothing), ...)`. Previously the fallback wrapped the raw field with naive `'...'` concatenation, so an inner single quote (always present for such fallbacks) prematurely terminated the quoted string and the rest of the value was lost.

<details>
<summary>Information about CI checks</summary>

The following docs are available: ...
</details>

### Version info
- Merged into: `26.5.1.936`

### DIFF
--- a/src/DataTypes/Serializations/SerializationDynamic.cpp
+++ b/src/DataTypes/Serializations/SerializationDynamic.cpp
@@ -12,6 +12,7 @@
 
 #include <Columns/ColumnDynamic.h>
 #include <IO/WriteHelpers.h>
+#include <IO/WriteBufferFromString.h>
 #include <IO/ReadHelpers.h>
 #include <IO/ReadBufferFromString.h>
 #include <Formats/EscapingRuleUtils.h>
@@ -848,8 +849,15 @@ static void deserializeTextImpl(
         /// We cannot insert value with incomplete type, insert it as String.
         variant_type = std::make_shared<DataTypeString>();
         /// To be able to deserialize field as String with Quoted escaping rule, it should be quoted.
+        /// Use `writeQuotedString` so inner single quotes and backslashes are escaped properly;
+        /// naive concatenation `"'" + field + "'"` would terminate prematurely at the first inner
+        /// single quote, truncating the stored value (issue #105441).
         if (escaping_rule == FormatSettings::EscapingRule::Quoted && (field.size() < 2 || field.front() != '\'' || field.back() != '\''))
-            field = "'" + field + "'";
+        {
+            WriteBufferFromOwnString quoted;
+            writeQuotedString(field, quoted);
+            field = std::move(quoted.str());
+        }
     }
 
     if (dynamic_column.addNewVariant(variant_type, variant_type->getName()))
--- a/tests/queries/0_stateless/04261_105441_dynamic_quoted_empty_array_map_key.reference
+++ b/tests/queries/0_stateless/04261_105441_dynamic_quoted_empty_array_map_key.reference
@@ -1,0 +1,10 @@
+-- empty array as map key (raw text 11 chars, must round-trip intact)
+{"length(CAST(d, 'String'))":11,"CAST(d, 'String')":"{[]:'came'}","dynamicType(d)":"String"}
+-- empty array key with embedded single quote (would corrupt)
+{"length(CAST(d, 'String'))":21,"CAST(d, 'String')":"{[]:'with\\'embedded'}","dynamicType(d)":"String"}
+-- nested empty array as map key
+{"length(CAST(d, 'String'))":8,"CAST(d, 'String')":"{[[]]:1}","dynamicType(d)":"String"}
+-- empty array key + non-empty key (Array(Int64) is complete, stays as Map)
+{"length(CAST(d, 'String'))":16,"CAST(d, 'String')":"{[]:'a',[1]:'b'}","dynamicType(d)":"Map(Array(Int64), String)"}
+-- already-single-quoted plain string is not double-wrapped
+{"length(CAST(d, 'String'))":12,"CAST(d, 'String')":"plain string","dynamicType(d)":"String"}
--- a/tests/queries/0_stateless/04261_105441_dynamic_quoted_empty_array_map_key.sql
+++ b/tests/queries/0_stateless/04261_105441_dynamic_quoted_empty_array_map_key.sql
@@ -1,0 +1,36 @@
+-- Tags: no-fasttest
+-- Tag no-fasttest: requires Dynamic type, experimental.
+
+-- Regression for issue #105441 ("Empty array as map key, not serialized back").
+-- A `Dynamic` value whose Values/Quoted text infers to `Map(Array(Nothing), ...)` is
+-- considered incomplete, so the deserializer falls back to `String`. The fallback
+-- previously did `"'" + field + "'"`, which terminated prematurely at the first
+-- embedded single quote and silently truncated the stored value. The fix uses
+-- `writeQuotedString` so inner single quotes and backslashes are escaped properly.
+
+SET allow_experimental_dynamic_type = 1;
+
+SELECT '-- empty array as map key (raw text 11 chars, must round-trip intact)';
+SELECT length(d::String), d::String, dynamicType(d)
+FROM format(Values, 'd Dynamic', $$({[]:'came'})$$)
+FORMAT JSONEachRow;
+
+SELECT '-- empty array key with embedded single quote (would corrupt)';
+SELECT length(d::String), d::String, dynamicType(d)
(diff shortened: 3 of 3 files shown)


groeneai:
**Pre-PR validation gate**

| # | Check | Result |
|---|---|---|
| a | Deterministic repro | `clickhouse-local --query "SELECT * FROM format(Values, 'd Dynamic', '({[]:\\'came\\'})')"` returns `{[]:` (4 chars) on master, `{[]:'came'}` (11 chars) with the fix. |
| b | Root cause | `Map(Array(Nothing), Nullable(String))` is correctly inferred from `{[]:'came'}`. `checkIfTypeIsComplete` recurses into `Array(Nothing) -> Nothing -> false`, so the variant type is treated as incomplete and the deserializer falls back to `String`. For `EscapingRule::Quoted` the fallback wraps the field in single quotes so `readQuotedStringInto` can parse it back. The wrapping used `"'" + field + "'"`, which leaves embedded `'` unescaped. The inferred Map literal always contains a `'` from its values, so the round-trip terminates at the first inner quote and the rest is dropped. |
| c | Fix matches cause | Replaces the naive concatenation with `writeQuotedString`, which is the symmetric writer for `readQuotedStringInto<true>` — both understand the same `\\'` / `\\\\` escapes, so the round-trip is preserved. |
| d | Test intent preserved / new tests | New regression test `04261_105441_dynamic_quoted_empty_array_map_key.sql` covers: original reproducer, embedded single quote, nested empty array, complete Map (still inferred as Map, not String), already-quoted plain string (not double-wrapped). |
| e | Both directions | Verified locally: test FAILS without the fix with the truncated `{[]:` output; passes 30/30 with randomization with the fix applied. |
| f | Fix is general | This is the only call site where the variant-type fallback re-wraps a field for a different escaping rule. Other rules (`CSV`/`Escaped`/`JSON`/`Raw`) keep the field in its original on-the-wire form, so no symmetric fix is needed there. |

Session: cron:clickhouse-ci-task-worker:20260520-171500

groeneai:
cc @Avogar @al13n321 — could you review this? It fixes silent value truncation in `Dynamic` columns when the `Quoted` (Values) text reader falls back to `String` for incomplete inferred types: `writeQuotedString` replaces the naive `"'" + field + "'"` wrapping so embedded single quotes no longer terminate the round-trip parse. cc @PedroTadim (reporter).

Avogar (approved):

PedroTadim:
@groeneai can you quickly audit the Serializers to find more badly quoted values and fix those?

groeneai:
@PedroTadim Audit done. The naive `"'" + field + "'"` in `SerializationDynamic.cpp:852` was the only such pattern in `src/DataTypes/Serializations/`. A `grep -c` of `"'"` over the whole directory now returns exactly one match, the one this PR fixes.

What I checked and why each is safe:

**Symmetric `writeQuotedString` / `writeQuotedStringPostgreSQL` paired with `readQuotedStringInto<true>` / `readQuotedStringWithSQLStyle`** (both readers accept `\'` and `''`):
- `SerializationString.cpp` (468/470 write, 476 read)
- `SerializationEnum.cpp` (66/68 write, 75 read)
- `SerializationCustomSimpleText.cpp` (93/95 write, 104 read)
- `SerializationAggregateFunction.cpp` (318/320 write, 327 read)
- `SerializationJSON.cpp` (325/327 write, 337 read)
- `SerializationFixedString.cpp` (244-249 write via `writeAnyEscapedString` between literal quotes, 255 read)

**Scalar wrappers (`writeChar('\'') + serializeText + writeChar('\'')`)** in `SerializationDate.cpp`, `SerializationDate32.cpp`, `SerializationDateTime.cpp`, `SerializationDateTime64.cpp`, `SerializationTime.cpp`, `SerializationTime64.cpp`, `SerializationIPv4andIPv6.cpp`, `SerializationUUID.cpp` are safe-by-construction: `serializeText` for those types emits hyphenated digits / dotted IP / hex UUID, none of which can contain `'`.

**SerializationVariant** has no String-fallback path (closed schema, parse fails if no variant matches), so it cannot exhibit this bug.

**SerializationDynamic** does the wrap only for `EscapingRule::Quoted`. The JSON / CSV / Escaped / Raw paths feed the raw `readJSONField` / `readCSVField` / `readEscaped` / `readString` output through `SerializationString::deserializeText*` unchanged, which is correct because each downstream reader matches the upstream reader format (JSON literal / CSV value / escaped / raw bytes).

**`Formats/EscapingRuleUtils.cpp`** uses `writeQuotedString` and `readQuotedString` directly, symmetric.

**`Processors/Formats/Impl/MySQLDumpRowInputFormat.cpp:122`** uses `readQuotedStringInto<false>` inside `readUntilEndOfQuery`, which is a raw SQL passthrough, not a serialization round-trip.

Remaining `"'" + ... + "'"` matches outside `src/DataTypes/Serializations/` are not data round-trip paths: Kusto parsers (`escapeSingleQuotes` called explicitly), BuzzHouse generators (`escapeSQLString` for user-controlled values, plain controlled values otherwise), and exception / log messages in `ConfigProcessor`, `Planner`, `WindowStep`, `ExternalLoader`, `StorageMergeTree`, `Parquet/SchemaConverter`, `Avro`, `GlueCatalog`, `mysqlxx/Exception`, `Core/SettingsEnums.h`, `Functions/FunctionGenerateRandomStructure`.

Conclusion: no further fixes needed in Serializers. This PR closes the only real instance.

PedroTadim:
@groeneai can you do one final merge with master, to see if this PR can be merged?

clickgapai:

**Affected code:** `deserializeTextImpl` in SerializationDynamic.cpp - the String fallback quoting for Quoted escaping rule — core code present in all supported branches.

**Why this might apply:** Silent data corruption bug in Dynamic type - values are truncated at inner quotes when incomplete type inference triggers String fallback in Values format. Dynamic type was introduced in 24.5 and backported to all subsequent releases. The buggy naive quoting code has existed since the incomplete-type fallback was added. All supported releases are affected.

PedroTadim:
@Avogar should this be backported?

PR #105468, opened 2026-05-20
Reject nested Dynamic/Variant in min/max aggregates and minmax indexes

### LINKED ISSUE #104747: Nested Dynamic/Variant in Tuple/Array bypasses min/max type validation

_Retrospective finding from a historical scan of [PR #71761](https://github.com/ClickHouse/ClickHouse/pull/71761) (merged 2024-11-20). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

min/max aggregate functions accept Tuple(Dynamic,...) and Array(Dynamic) types without error, producing confusing results when comparing values of different internal types

**Root cause:** AggregateFunctionsMinMax.cpp:37: The check `if (isDynamic(this->result_type) || isVariant(this->result_type))` only validates top-level type. Should use `result_type->forEachChild()` like ORDER BY validation does in ExpressionAnalyzer.cpp:1795-1796 to recursively check nested types.

**Why we believe this is a bug:** AggregateFunctionsMinMax.cpp:37 → isDynamic(this->result_type) only checks top-level type, not nested children. When result_type is Tuple(Dynamic,UInt64), isDynamic() returns false because top-level is Tuple, not Dynamic.

**Affected locations:**
- `src/AggregateFunctions/AggregateFunctionsMinMax.cpp:37` — isDynamic check for min/max result type
- `src/AggregateFunctions/AggregateFunctionsArgMinArgMax.cpp:117` — isDynamic check for argMin/argMax value type
- `src/AggregateFunctions/Combinators/AggregateFunctionCombinatorsArgMinArgMax.cpp:70` — isDynamic check for combinator key type
- `src/Storages/MergeTree/MergeTreeIndexMinMax.cpp:485` — isDynamic check for minmax index column type

**Impact:** Users can create min/max aggregates and minmax indexes on nested Dynamic/Variant types (e.g., Tuple(Dynamic,UInt64), Array(Dynamic)), bypassing the PR's intended restriction. This produces confusing comparison results where values of different internal types are compared using type-based ordering. Inconsistent with ORDER BY which correctly rejects nested Dynamic/Variant with 'Data types Variant/Dynamic are not allowed in ORDER BY keys'.

Yes — confirmed on current `master` (commit `78bc83b28b7`).

### How to reproduce

```sql
SET allow_experimental_dynamic_type=1;
SET allow_experimental_variant_type=1;

-- These queries succeed but should throw ILLEGAL_TYPE_OF_ARGUMENT
-- because nested Dynamic/Variant should be forbidden, consistent with ORDER BY:
-- SELECT * FROM (SELECT tuple(number::Dynamic, number) as x FROM numbers(3)) ORDER BY x;
-- ^ This correctly throws: Data types Variant/Dynamic are not allowed in ORDER BY keys

-- Test min with Tuple containing Dynamic (BUG: succeeds, should fail)
SELECT min(tuple(number::Dynamic, number)) FROM numbers(3);

-- Test max with Array containing Dynamic (BUG: succeeds, should fail)
SELECT max([number::Dynamic]) FROM numbers(3);

-- Test argMin with nested Dynamic in value arg (BUG: succeeds, should fail)
SELECT argMin(1, tuple(number::Dynamic, number)) FROM numbers(3);

-- Test minmax index with nested Dynamic (BUG: succeeds, should fail)
DROP TABLE IF EXISTS test_minmax_nested_dynamic;
CREATE TABLE test_minmax_nested_dynamic (x Tuple(Dynamic, UInt64), INDEX idx x TYPE minmax) ENGINE = MergeTree ORDER BY tuple();
DROP TABLE IF EXISTS test_minmax_nested_dynamic;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/77026034-7868-4aa7-b4e8-88829b31ffb4)

### Expected behavior

```
ILLEGAL_TYPE_OF_ARGUMENT error for all aggregate queries
BAD_ARGUMENTS error for CREATE TABLE with minmax index
```

```
(0,0)
[2]
1
(table created successfully)
```

### Additional context

**Open risks:**
- Map(K, Dynamic) may also bypass the check
- Nested JSON type paths may have similar issues

**Suggested fix:** Add recursive check using forEachChild(): ```cpp
auto check = [&](const IDataType & type) {
    if (isDynamic(type) || isVariant(type))
        throw Exception(...);
};
check(*this->result_type);
this->result_type->forEachChild(check);
```

Found during automated review of [PR #71761](https://github.com/ClickHouse/ClickHouse/pull/71761).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr71761_001`_

### PR DESCRIPTION
**Pull request by Avogar**

Reject nested Dynamic/Variant in min/max aggregates and minmax indexes. Previously only top-level Dynamic/Variant were checked. Closes https://github.com/ClickHouse/ClickHouse/issues/104747

### CODE REVIEW

File: src/AggregateFunctions/Combinators/AggregateFunctionCombinatorsArgMinArgMax.cpp [hunk 1]
```diff
@@ -67,13 +67,18 @@ class AggregateFunctionCombinatorArgMinArgMax final : public IAggregateFunctionH
...
                 getName());
 
-        if (isDynamic(arguments[key_col]) || isVariant(arguments[key_col]))
-            throw Exception(
-                ErrorCodes::ILLEGAL_TYPE_OF_ARGUMENT,
-                "Illegal type {} of argument of aggregate function {} because the column of that type can contain values with different "
-                "data types. Consider using typed subcolumns or cast column to a specific data type",
-                arguments[key_col]->getName(),
-                getName());
+        auto check_not_dynamic_or_variant = [&](const IDataType & type)
```
antonio2368, line 76: can this break backwards compatibility?

alexey-milovidov:

src/AggregateFunctions/Combinators/AggregateFunctionCombinatorsArgMinArgMax.cpp
Avogar: Yes, but I would say this was just a bug that we allowed it. I don't think anyone used it. Do you suggest to change the changelog category to `Backward-incomoatible change`?

antonio2368: https://xkcd.com/1172/
Let's go for a safer option and change the category.

antonio2368 (approved):

PR #105469, opened 2026-05-20
Include `skip_first_lines` in schema cache key for `WithNames` formats

### LINKED ISSUE #104527: `skip_first_lines` not in cache key for `CSVWithNames`/`TSVWithNames` causes wrong schema to be returned

_Retrospective finding from a historical scan of [PR #65980](https://github.com/ClickHouse/ClickHouse/pull/65980) (merged 2024-07-10). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

When using `CSVWithNames` or `TSVWithNames` format with different `skip_first_lines` values on the same file, the second query returns the cached schema from the first query, even though changing `skip_first_lines` changes which row is used as the header.

**Root cause:** The condition `if (!with_names)` at CSVRowInputFormat.cpp:637 incorrectly excludes `skip_first_lines` from the cache key for `WithNames` formats. Same issue at TabSeparatedRowInputFormat.cpp:445.

**Why we believe this is a bug:** CSVRowInputFormat.cpp:637-638 adds `skip_first_lines` to cache key only when `!with_names`. For `CSVWithNames` (where `with_names=true`), `skip_first_lines` is NOT in the cache key, so different setting values produce the same cache key.

**Affected locations:**
- `src/Processors/Formats/Impl/CSVRowInputFormat.cpp:637` — skip_first_lines only added when !with_names
- `src/Processors/Formats/Impl/TabSeparatedRowInputFormat.cpp:445` — skip_first_lines only added when !with_names

**Impact:** Schema inference returns wrong column names when querying same file with CSVWithNames/TSVWithNames format using different `skip_first_lines` values. The cached schema from the first query is incorrectly reused.

Yes — confirmed on current `master` (commit `d640b7762fd`).

### How to reproduce

```sql
-- Test CSVWithNames schema cache bug with skip_first_lines
-- File content: header1,header2\na,b\n1,2

SYSTEM DROP SCHEMA CACHE FOR FILE;

-- Query 1: skip_first_lines=1, uses 'a,b' as header
DESC file('/tmp/test.csv', CSVWithNames) SETTINGS input_format_csv_skip_first_lines=1;
-- Returns: a, b columns

-- Query 2: skip_first_lines=0, should use 'header1,header2' as header
DESC file('/tmp/test.csv', CSVWithNames) SETTINGS input_format_csv_skip_first_lines=0;
-- BUG: Returns cached 'a,b' schema instead of 'header1,header2'
```

### Expected behavior

```
a	Nullable(Int64)
b	Nullable(Int64)
header1	Nullable(String)
header2	Nullable(String)
```

```
a	Nullable(Int64)
b	Nullable(Int64)
a	Nullable(Int64)
b	Nullable(Int64)
```

### Additional context

**Suggested fix:** Add `skip_first_lines` unconditionally to the cache key (within the `!with_types` block), not just when `!with_names`. For example: always add `fmt::format(", skip_first_lines={}", settings.csv.skip_first_lines)` before the `if (!with_names)` block.

Found during automated review of [PR #65980](https://github.com/ClickHouse/ClickHouse/pull/65980).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr65980_001`_

### Version info
- Resolved by: #105469
- Merged into: `26.6.1.504`

Avogar:

Include `skip_first_lines` in schema cache key for `WithNames` formats. Closes https://github.com/ClickHouse/ClickHouse/issues/104527

### Version info
- Merged into: `26.6.1.504`


CheSema (approved):

clickgapai:
Hi @Avogar @CheSema — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: Body is the bare changelog template with no prose explanation of the symptom, root cause, or that WithNamesAndTypes (not only WithNames) is now also covered. The title is accurate but the body gives a reviewer no context.
- Suggested addition:

  > Before this change, the schema inference cache key for `CSVWithNames`/`CSVWithNamesAndTypes` and the TSV/TabSeparated/Raw `WithNames*` variants did not include `skip_first_lines`. As a result, describing/reading the same file with a different `input_format_csv_skip_first_lines` (or the TSV equivalent) returned a stale cached schema (issue #104527). The getter is now registered unconditionally and always appends `skip_first_lines`. Note this also adds the getter for the `WithNamesAndTypes` variants, which previously registered none.

PR #99839, opened 2026-03-18
Fix aggregate_functions_null_for_empty for non-Nullable result types

alexey-milovidov:
Closes #38738. Closes #38786.

Some aggregate functions (e.g., `groupArray`, `sumMap`) return types like `Array` or `Map` that cannot be inside `Nullable`. The `aggregate_functions_null_for_empty` setting was unconditionally wrapping all result types in `Nullable` via the `OrNull` combinator, causing an exception like:

```
Code: 43. DB::Exception: Nested type Array(UInt64) cannot be inside Nullable type.
```

Now the `OrNull` combinator checks `canBeInsideNullable` before wrapping, and skips the `Nullable` wrapper for incompatible types, returning the default value (e.g., empty array) instead.

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix `aggregate_functions_null_for_empty` setting to work with aggregate functions returning non-Nullable types such as `Array` or `Map` (e.g., `groupArray`, `sumMap`).

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

---

> [!NOTE]
> **Medium Risk**
> Changes `-OrNull` combinator result typing/serialization paths for some aggregate functions, which can affect query result types and empty-input behavior (especially for `Array`/`Map` aggregates). Scope is contained to `AggregateFunctionOrFill` but impacts any query using `aggregate_functions_null_for_empty` or `OrNull` combinators.
> 
> **Overview**
> Fixes `-OrNull` aggregate combinator to **avoid wrapping result types in `Nullable` when `canBeInsideNullable()` is false** (e.g., `Array`, `Map`), preventing exceptions under `aggregate_functions_null_for_empty` and instead returning the normal default value for empty input.
> 
> Updates `AggregateFunctionOrFill` to track whether the *result* is nullable and to correctly insert merge vs non-merge results into nullable columns. Tests are updated to expect `topKOrNull` to be valid with parameters, and a new stateless test covers empty-input behavior for non-Nullable-compatible result types (`groupArray`, `sumMap`) alongside NULL-returning aggregates.
> 
[...]

PR #105470, opened 2026-05-20
Fix segfault in `uniqStateOrNull` with `WITH ROLLUP`/`CUBE`/`TOTALS`

### LINKED ISSUE #105462: `uniqStateOrNull` + Nullable arg + `GROUP BY ... WITH ROLLUP` segfaults the server
## Affected version
master `c3d7f9fa9b2` (May 20 2026), reproducible on `clickhouse local` and `clickhouse server` with default settings.

## Minimal reproducer (3 lines)

```sql
CREATE TABLE t (h Nullable(UInt16)) ENGINE = MergeTree ORDER BY tuple();
INSERT INTO t SELECT if(number % 5 = 0, NULL, toUInt16(number)) FROM numbers(20);
SELECT uniqStateOrNull(h) FROM t GROUP BY h WITH ROLLUP;
```

Result: `Segmentation fault (core dumped)`. The same query without `WITH ROLLUP` (plain `GROUP BY h`) returns a normal result.

Fiddle works for 26.2 and crashes on 26.3
https://fiddle.clickhouse.com/491bd7ca-1b1b-4e82-a012-53c3601a5340

## Trigger conditions

All of the following are required:

1. The aggregate is from the `uniq` family (`uniq`, `uniqExact`, etc.); `any`, `count`, `sum` do **not** trigger.
2. The `State` combinator is applied.
3. **Either** the `OrNull` **or** the `OrDefault` combinator is applied (in any position relative to `State`).
4. At least one argument to the aggregate is `Nullable`.
5. `GROUP BY ... WITH ROLLUP` (or `WITH CUBE` / `WITH TOTALS`). Equivalent `GROUPING SETS ((h), ())` does **not** trigger — only the `WITH ROLLUP`/`WITH CUBE`/`WITH TOTALS` aggregator paths.

Confirmed-crashing chains: `uniqStateOrNull`, `uniqStateOrDefault`, `uniqOrNullState`, `uniqOrNullStateOrNull`, `uniqOrNullStateOrNullDistinctOrDefaultArgMin`, …
Confirmed-not-crashing: `anyStateOrNull(h) … WITH ROLLUP`, `countStateOrNull(h) … WITH ROLLUP`, `sumStateOrNull(h) … WITH ROLLUP`, `uniqState(h) … WITH ROLLUP`, `uniqStateOrNull(h) GROUP BY h GROUPING SETS ((h), ())`.

## Server-side stack trace

Captured from a fuzzer-found instance running the heavier original query (clickhouse-local on the minimum repro segfaults with no stack output since it doesn't install the daemon signal handler):

```
3. src/AggregateFunctions/UniquesHashSet.h:476: UniquesHashSet<DefaultHash<unsigned long>>::write(WriteBuffer&) const
4. src/AggregateFunctions/Combinators/AggregateFunctionOrFill.h:263: AggregateFunctionOrFill<true>::serialize(char const*, WriteBuffer&, std::optional<unsigned long>) const
5. src/Columns/ColumnAggregateFunction.cpp:452: ColumnAggregateFunction::updateHashWithValue(unsigned long, SipHash&) const
6. src/Processors/Transforms/DistinctTransform.cpp:87 / RollupTransform: hash key for the rolled-up bucket
…
```

The crash is a NULL pointer dereference. The `UniquesHashSet` instance whose `write()` is invoked has not been initialised — when `WITH ROLLUP` synthesises the "all-grouped-up" row, the `OrNull`/`OrDefault`-wrapped state for that row is left as a default-constructed (empty / no-flag-set) state and `AggregateFunctionOrFill::serialize` then forwards into the nested `UniquesHashSet::write` which dereferences uninitialised internal pointers.

## Discovered by
PR #99980 AST fuzzer oracle CI job — the fuzzer generated a heavier query whose first SELECT is reducible to the 3-line repro above.

## Suggested fix direction
`AggregateFunctionOrFill::serialize` (and the `Null` variant) should either skip serialising the inner aggregate when the `OrFill`/`OrNull` flag bit indicates "no value seen", or `UniquesHashSet::write` should handle the empty-state case. The bug is independent of the rest of the query — it's purely in how multi-grouping-set aggregator paths (`RollupTransform`, `CubeTransform`, `TotalsHaving`) interact with `OrFill`/`OrNull`-wrapped `uniq*State` aggregates.

### Version info
- Resolved by: #105470
- Merged into: `26.6.1.155`
- Backported to: `26.5.2.15`, `26.4.4.19`, `26.3.13.18`

groeneai:
Closes #105462.

`SELECT uniqStateOrNull(h) FROM t GROUP BY h WITH ROLLUP` (and the `WITH CUBE` / `WITH TOTALS` and `*OrDefault` chains) crashes the server with a `nullptr` read in `UniquesHashSet::write`. The synthesised "all-grouped-up" or "totals" row's nested `UniquesHashSet` ends up with `buf == nullptr` by the time it is serialized, and the bucket scan dereferences `buf[0]`.

`UniquesHashSet::write` now bails out with an empty state (`skip_degree = 0`, `m_size = 0`) when `buf` is null, instead of attempting to scan a zero-size buffer.

Reproduces against `c3d7f9fa9b2` (master, May 2026), `v26.3`, `v26.4`. Bidirectionally verified locally (master debug build): the minimum repro from #105462 segfaults without this patch and passes with it.

Fix a server segfault in `uniqStateOrNull` / `uniqStateOrDefault` / `uniqOrNullState` (and similar combinator chains over `uniq`) when used with `GROUP BY ... WITH ROLLUP`, `WITH CUBE`, or `WITH TOTALS` and a Nullable argument.

---

> [!NOTE]
> **Medium Risk**
> Touches core aggregation result insertion for state-returning functions and adds an assertion in `UniquesHashSet::write`, which could affect stability/behavior of aggregate-state serialization paths if there are edge cases beyond the reported crash.
> 
> **Overview**
> Fixes a crash when `*StateOrNull`/`*StateOrDefault` (notably `uniqStateOrNull`) is used with `GROUP BY ... WITH ROLLUP`/`CUBE`/`TOTALS` by changing `AggregateFunctionOrFill` to **not** route “no input” rows through `to.insertDefault()` when the nested function returns an aggregate *state*, and instead forward the default-initialized state through the same insertion path as non-empty rows.
> 
> Adds a defensive `chassert(buf)` in `UniquesHashSet::write()` to catch corrupted/double-destroyed states earlier, and introduces stateless regression tests covering rollup/cube/totals plus downstream consumers (`uniqMerge`, `AggregatingMergeTree` persistence, two-level/multithread aggregation, and `Memory` sources).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9c5a32a8a456084116a9c063f74629027768a8a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.155`
- Backported to: `26.5.2.15`, `26.4.4.19`, `26.3.13.18`


groeneai:
**Pre-PR validation gate**

a) **Deterministic repro?** Yes. The 3-line repro from #105462 (a `Nullable(UInt16)` column with mixed NULL and non-NULL rows + `uniqStateOrNull(...) GROUP BY ... WITH ROLLUP`) segfaults the server on demand on master `48b7d9f3` debug build, exit 139 SIGSEGV. Stack trace: `UniquesHashSet::write` at line 476 → `AggregateFunctionUniq::serialize` → `SerializationAggregateFunction::serializeBatch` → `NativeWriter::writeData`. Same crash via the `hex(...)` path through `serializeText`.

b) **Root cause explained?** When `WITH ROLLUP` / `WITH CUBE` / `WITH TOTALS` synthesises the "all-grouped-up" / "totals" row from `uniq*State*OrNull`-wrapped states, the nested `UniquesHashSet` of that row ends up with `buf == nullptr` by the time it is serialized. The serialization loop scans `buf[0..buf_size())` to find non-zero bucket entries and dereferences a null pointer.

c) **Fix matches root cause?** Yes. `UniquesHashSet::write` now emits an empty state representation (`skip_degree = 0`, `m_size = 0`) when `buf` is null, instead of dereferencing it. This is the second of the two alternatives the issue suggests ("UniquesHashSet::write should handle the empty-state case"). The wire format is preserved for all valid states.

d) **Test intent preserved? / New tests added?** New regression test `04262_105462_uniq_state_or_null_with_rollup.sql` covers `WITH ROLLUP`, `WITH CUBE`, `WITH TOTALS` against `uniqStateOrNull`, `uniqStateOrDefault`, `uniqOrNullState`, plus a plain `GROUP BY` sanity check. It asserts only row counts (not state bytes) so it does not encode internal `UniquesHashSet` layout.

e) **Demonstrated in both directions?** Yes. With `git stash` of the patch the regression test fails the way #105462 describes (server segfaults, query returns network error). With the patch reinstated and the server restarted, the 3 crashing variants all return without error and the regression test passes.

f) **Fix is general, not a narrow patch?** I considered the symmetric question — does this same pattern exist elsewhere? `UniquesHashSet::read` already handles `m_size == 0` correctly via its `m_size <= 1` branch. The other `uniq*` data structures (`HyperLogLogWithSmallSetOptimization`, `HashSet`-based `uniqExactData`) check their state shape before scanning members, so they are not vulnerable to the same `nullptr buf` pattern. The issue only reports crashes on the `uniq` chain, consistent with this.

This is a defensive write fix rather than a deeper repair of why the inner state arrives in this shape from the multi-grouping-set aggregator paths. A follow-up to identify and fix the upstream lifecycle issue (state destroyed prematurely or zero-initialised) would be valuable, but this patch is sufficient to stop the crash, restores the v26.2 behavior (an empty serialized state for the rolled-up bucket), and is wire-format compatible.

Session: cron:clickhouse-ci-task-worker:20260520-184500

groeneai:
cc @alexey-milovidov @qoega @rschu1ze — could you review? This fixes the segfault @qoega reported in #105462. The fix is defensive (handles `nullptr buf` in `UniquesHashSet::write`), wire-format compatible, and isolated to the `uniq` data structure.

qoega (approved):

groeneai:
## Pre-PR validation gate (fix-up commit `4187aa5c`)

The original write-path guard fixed the segfault but left a separate latent destruction-order bug reachable, which `MemorySanitizer` correctly flagged (STID 8424-6802). Pushed fix-up commit answering the mandatory gate:

**a) Deterministic repro?** Yes — `04262_105462_uniq_state_or_null_with_rollup.sql` (the regression test added in this PR) crashed every MSan run on commit `3e5a30c7` (CIDB: 4 hits across `Stress test (amd_msan)`, `Stress test (arm_msan)`, `Stateless tests (amd_msan, WasmEdge, parallel, 1/2)`, `Stateless tests (amd_msan, flaky check)`, all on this PR's `3e5a30c7`, zero hits on any other PR or master).

**b) Root cause explained?** Yes. Trace (from MSan stack):
  1. `AggregateFunctionOrFill::insertResultIntoImpl` flag-set branch pushes the inner aggregate state pointer into the result `ColumnAggregateFunction` as an external reference via `nested_function->insertResultInto` -> `AggregateFunctionState::insertResultInto` -> `data.push_back(place)`. The result column's `src` keeps the source alive.
  2. The flag-unset branch previously called `to.insertDefault()`, which for `State`-nested combinators reaches `ColumnAggregateFunction::insertDefault` on the inner column. That calls `ensureOwnership()` and resets `src` to null.
  3. Any later flag-set row in the same `convertToValues` loop then pushes an external pointer into a column with `src == nullptr`. At pipeline tear-down both this column and the source column destroy the same `UniquesHashSet`, so the second `~UniquesHashSet` reads `buf` after the first one already destroyed the object -> `MemorySanitizer: use-of-uninitialized-value` at `UniquesHashSet.h:99`.

**c) Fix matches root cause?** Yes. Mirror the flag-set branch for `State`-nested combinators: forward the already-default-initialized inner state through `nested_function->insertResultInto` and push `1` to the null map. `create()` runs `nested_function->create(place)` unconditionally, so the inner state at `place` is a properly default-constructed `UniquesHashSet`. `src` remains set for the entire `convertToValues` loop, so the result column skips `func->destroy` at destruction and the source column owns destruction uniquely.

**d) Test intent preserved? / New tests added?** Yes. Regression test `04262_105462_uniq_state_or_null_with_rollup` still asserts the same counts (no segfault, no exception). Output is byte-identical with and without the fix in non-MSan builds. The fix only changes which arena owns the state pointer for flag-unset rows; the underlying serialized state (default empty `UniquesHashSet`) is the same.

  - Without commit `4187aa5c`: CI MSan jobs on `3e5a30c7` report STID 8424-6802 (use-of-uninitialized-value in `~UniquesHashSet`).
  - With commit `4187aa5c`: 50/50 randomized-settings runs of the regression test in Debug pass.
  - Full MSan re-validation will run on this push and should clear the bugfix-validation FAIL too.

**f) Fix is general, not a narrow patch?** Yes. The fix lives in `AggregateFunctionOrFill::insertResultIntoImpl` and is gated on `nested_function->isState()`, so it covers both `-OrNull` and `-OrDefault` over any `State`-combinator-wrapped function (`uniqState*`, `uniqExactState*`, `quantileState*`, etc.), not just `uniqStateOrNull`. Non-`State` paths (e.g. `uniqOrNull`, `sumOrDefault`) are unchanged.

cc @qoega for re-approval given the new commit.

src/AggregateFunctions/UniquesHashSet.h
```diff
@@ -463,6 +463,19 @@ class UniquesHashSet : private HashTableAllocatorWithStackMemory<(1ULL << UNIQUE
         if (m_size > UNIQUES_HASH_MAX_SIZE)
             throw Poco::Exception("Cannot write UniquesHashSet: too large size_degree.");
 
+        /// `buf` may be null when the surrounding aggregate state was not
```
Algunenano: @groeneai This is correct but the wrong approach. Reaching this circumstance only happens due to corruption.

This should either be a chassert(buf) or a check and throw LOGICAL_ERROR.

The right fix is the one doing in AggregateFunctionOrFill::insertResultIntoImpl

Algunenano (changes requested):
@groeneai 

The `AggregateFunctionOrFill::insertResultIntoImpl` change is the right root-cause fix. However, the regression surface from #99839 (which enabled `*StateOrNull` / `*StateOrDefault` chains over Nullable arguments) is wider than the current regression test covers. The bug reproduces under MSan through several other code paths that don't touch the `WITH ROLLUP`/`CUBE`/`TOTALS` aggregator-transform path the test exercises.

I'd like to see these added as regression tests so we can lock in that the OrFill fix handles all consumers of the rollup-produced state, not just the direct `SELECT … FROM (… WITH ROLLUP)` path.

## Reproducers under MSan (master, post-#99839, pre-this-PR)

Shared setup:

```sql
CREATE TABLE source (h Nullable(UInt16)) ENGINE = MergeTree ORDER BY tuple();
INSERT INTO source SELECT if(number % 5 = 0, NULL, toUInt16(number)) FROM numbers(20);
```

### 1. `<func>Merge` over rollup-produced states

```sql
SELECT uniqMerge(s)
FROM (SELECT uniqStateOrNull(h) AS s FROM source GROUP BY h WITH ROLLUP);
```

MSan reports `use-of-uninitialized-value` in `UniquesHashSet::merge` at `UniquesHashSet.h:438`, via `AggregateFunctionMerge::add` → `Aggregator::executeWithoutKeyImpl`. This exercises the `merge` path on the rollup-produced state column, a different code path than the `serialize`-driven one in the issue.

### 2. INSERT into AggregatingMergeTree

```sql
CREATE TABLE agg_t (k UInt8, s AggregateFunction(uniq, UInt16)) ENGINE = AggregatingMergeTree ORDER BY k;

INSERT INTO agg_t
SELECT toUInt8(coalesce(h, 999)) AS k, uniqStateOrNull(h)
FROM source
GROUP BY h WITH ROLLUP;
```

MSan reports the use-of-uninitialized-value in `UniquesHashSet::merge` via `ColumnAggregateFunction::insertMergeFrom` → `AggregatingSortedAlgorithm::AggregatingMergedData::addRow` → `MergeTreeDataWriter::mergeBlock`. The corruption is persisted into a part-merge path.

### 3. Two-level aggregator (single-thread)

```sql
SELECT hex(uniqStateOrNull(h))
FROM source
GROUP BY h WITH ROLLUP
SETTINGS group_by_two_level_threshold = 1, group_by_two_level_threshold_bytes = 1, max_threads = 1;
```

Reproduces on master.

### 4. Multi-threaded merge

```sql
SELECT hex(uniqStateOrNull(h))
FROM source
GROUP BY h WITH ROLLUP
SETTINGS max_threads = 4
FORMAT Null;
```

Reproduces on master.

### 5. Memory engine source

Same reproducer with the source table re-created as `ENGINE = Memory` also reproduces under MSan, confirming this is pipeline-level rather than `MergeTree`-specific.

## Paths that I checked and that are clean

For completeness, these did not reproduce under MSan (so the regression appears scoped to the synthetic-row-producing aggregator transforms, as expected):

- `SELECT uniqStateOrNull(h) FROM source WHERE 0` (empty source, no `GROUP BY`)
- `SELECT uniqStateOrNull(h) FROM source WITH TOTALS` (no `GROUP BY`)
- `SELECT uniqStateOrNull(h) FROM source GROUP BY GROUPING SETS ((h), ())`
- `SELECT arrayReduce('uniqStateOrNull', CAST([] AS Array(Nullable(UInt16))))`

## Suggestions

1. Add reproducers 1 and 2 above as regression tests. They cover the `merge`-into-state and AggregatingMergeTree-INSERT consumers of the rollup-produced state. The current test only exercises the `serialize`-driven final-projection path.
2. Optionally add reproducers 3 and 4 to lock in the two-level and multi-thread aggregator variants. They don't introduce a new code path but they do reach the buggy state via different scheduling, and the failure mode is identical, so a single combined test with a `SETTINGS` matrix would be enough.
3. The `OrDefault` test coverage is currently a single line. Since #99839 enabled the `*StateOrDefault` chain in the same way as `*StateOrNull`, it would be worth mirroring at least reproducer 1 for `uniqStateOrDefault` so we don't silently regress only one of the two combinator variants.

groeneai:
Thanks for the review. Addressed in 8d52133d4a10:

**Inline comment on `src/AggregateFunctions/UniquesHashSet.h:466`** — replaced the silent `if (!buf) bail` with `chassert(buf)`. The `AggregateFunctionOrFill::insertResultIntoImpl` change (4187aa5) makes a non-null `buf` an invariant at this site; the assertion documents that and surfaces any future regression loudly in debug / sanitiser builds.

**Regression-test surface** — added a new test `04273_105462_uniq_state_or_null_rollup_consumers` that covers your additional consumer paths without extending `04262`:

- `uniqMerge` over the rollup-produced state column — exercises `UniquesHashSet::merge` via `AggregateFunctionMerge::add`.
- `INSERT INTO AggregatingMergeTree` from the rollup — exercises the persistence path through `ColumnAggregateFunction::insertMergeFrom` → `AggregatingSortedAlgorithm` → `MergeTreeDataWriter::mergeBlock`. Asserted via `uniqMerge(s) FROM agg_t_105462` rather than `count()`, so the result is deterministic under `enable_parallel_replicas` / `max_insert_threads` randomisation (verified 100/100).
- Two-level single-thread + multi-thread variants (your reproducers 3 and 4) with a `SETTINGS` matrix per query.
- `Memory`-engine source variant (your reproducer 5), confirming the regression is pipeline-level rather than `MergeTree`-specific.
- The `uniqMerge`-over-rollup case is also mirrored for `uniqStateOrDefault`, since #99839 enabled both combinator variants identically.

**Pre-PR validation gate**

(a) Deterministic repro — yes. Each of the 5 reproducers produces a stable, single-value output (16 / 16 / 16 / 18 / 18 / 16).

(b) Root cause explained — `*StateOrNull` / `*StateOrDefault` over `Nullable` (enabled by #99839) combined with `WITH ROLLUP` / `CUBE` / `TOTALS` produces a synthesised "all-grouped-up" row in the aggregator transform. Before the `OrFill` fix, the flag-unset branch of `AggregateFunctionOrFill::insertResultIntoImpl` called `to.insertDefault()` on the inner `ColumnAggregateFunction`, which invoked `ensureOwnership()` and reset the column's `src` to null. Any subsequent flag-set row then pushed an externally-owned state pointer into a column without `src` protection, so both the result column and the source column ran `func->destroy` on the same `UniquesHashSet` at pipeline tear-down. The second `~UniquesHashSet` then read the just-freed `buf` field — `use-of-uninitialized-value` under MSan in `UniquesHashSet::merge` (your reproducers), or null-pointer segfault in debug builds.

(c) Fix matches root cause — the `OrFill` change in 4187aa5 is exactly the fix you endorsed; this commit only adds coverage. The `chassert(buf)` change documents and enforces the invariant the `OrFill` fix establishes.

(d) Test intent preserved — `04273` is purely additive coverage. `04262` is unchanged.

(e) Both directions demonstrated:
- Without the `OrFill` fix (binary from a sibling branch lacking 4187aa5), reproducer 1 segfaults at `UniquesHashSet::merge:453` via `AggregateFunctionMerge::addBatchSinglePlace` → `Aggregator::executeWithoutKeyImpl` → `AggregatingTransform::consume` (stack trace captured during this work — `rhs.buf[i]` deref).
- With the `OrFill` fix and the new `chassert`, 100/100 randomised runs of the new test pass and 50/50 randomised runs of `04262` still pass. Two-level aggregator tests (`02560_*`) also pass 50/50 (no chassert fires under normal operation).

(f) Fix is general — the `OrFill` change is gated on `nested_function->isState()`, so it covers all `-State*` + `-OrNull` / `-OrDefault` combinations (`uniqStateOrNull`, `uniqStateOrDefault`, `uniqExactStateOrNull`, `quantileStateOrNull`, etc.). The new test exercises 5 distinct downstream consumer paths.

Algunenano:
@groeneai I've left several comments in the last commit directly

groeneai:
Pushed fix-up `f5f1eb45a9a6` addressing the three commit comments:

- `src/AggregateFunctions/UniquesHashSet.h`: rewrote the `chassert(buf)` comment to describe only the invariant violation at the call site (no longer references the `OrFill` fix or other parts of this PR).
- Test filename: renamed to `04273_uniq_state_or_null_rollup_consumers.{sql,reference}` (dropped the `_105462_` infix).
- Test file header and inline comments: rewritten to describe the consumer paths being exercised (`Merge`, `AggregatingMergeTree` insert, two-level / multi-threaded aggregator, `Memory` engine). Removed references to the other test file and to the bug history.

Verified locally: 15/15 PASS on the renamed test and on the existing `04262` test against the rebuilt PR-branch binary.

Algunenano:
@groeneai tests/queries/0_stateless/04262_105462_uniq_state_or_null_with_rollup.sql is still named incorrectly

groeneai:
Renamed `tests/queries/0_stateless/04262_105462_uniq_state_or_null_with_rollup.{sql,reference}` to drop the `_105462_` infix in `c9c5a32a8a45`. Same shape as the `04273_105462_*` -> `04273_*` rename in `f5f1eb45a9a6`. File contents unchanged; `git diff --stat` shows the rename only.

Algunenano (approved):

PR #105473, opened 2026-05-20
Reduce transactions logging

azat:
Those are too noisy on an empty server

### Version info
- Merged into: `26.5.1.922`

### DIFF
--- a/src/Interpreters/MergeTreeTransaction/VersionMetadata.cpp
+++ b/src/Interpreters/MergeTreeTransaction/VersionMetadata.cpp
@@ -136,7 +136,7 @@ void VersionMetadata::setAndStoreCreationCSN(CSN csn)
 
 void VersionMetadata::setAndStoreRemovalTID(const TransactionID & tid)
 {
-    LOG_DEBUG(log, "Object {}, setAndStoreRemovalTID {}", getObjectName(), tid);
+    LOG_TEST(log, "Object {}, setAndStoreRemovalTID {}", getObjectName(), tid);
 
     auto update_function = [tid](VersionInfo & info)
     {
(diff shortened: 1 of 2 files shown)


tuanpach:
@azat, can we keep these debug logs after 26.4 is released? They might help troubleshoot issues.

alexey-milovidov (approved):

azat:
> can we keep these debug logs after 26.4 is released? They might help troubleshoot issues.

If this logs are useful - then let's keep it, but they are just too noisy to me, and they definitely looks like logs for developers

If you need to debug stateless tests then yes, we need to get them back, but not sure how to make it less nosiy, maybe not log every attempt to call a function, but only in case of failures (but I did not read that code deeply)

For integration tests we run them with logger.level=test, so those logs will be there

For local development you can also run with logger.level=test

Another option is to write something system.*_log table (only if enabled)

P.S. Sorry that this got merged before your review

tuanpach:
> > can we keep these debug logs after 26.4 is released? They might help troubleshoot issues.
> 
> If this logs are useful - then let's keep it, but they are just too noisy to me, and they definitely looks like logs for developers
> 
> If you need to debug stateless tests then yes, we need to get them back, but not sure how to make it less nosiy, maybe not log every attempt to call a function, but only in case of failures (but I did not read that code deeply)
> 
> For integration tests we run them with logger.level=test, so those logs will be there
> 
> For local development you can also run with logger.level=test
> 
> Another option is to write something system.*_log table (only if enabled)
> 
> P.S. Sorry that this got merged before your review

Let's downgrade the log level first. If needed, I will find a less noisy way to record the data for troubleshooting.

PR #105481, opened 2026-05-20
Fix NOT_FOUND_COLUMN_IN_BLOCK when combining WITH FILL INTERPOLATE and LIMIT BY

### LINKED ISSUE #103474: `NOT_FOUND_COLUMN_IN_BLOCK` when combining `WITH FILL INTERPOLATE` and `LIMIT N BY` (analyzer=1)
### Describe what's wrong

A single `SELECT` that combines `ORDER BY ... WITH FILL INTERPOLATE (...)` with `LIMIT N BY ...` fails at planning with `Code: 10. NOT_FOUND_COLUMN_IN_BLOCK`. The block contains the column under the `__table1.` prefix but the lookup is for the unprefixed name, and the tail of the block shows `materialize(__table1.<sort-key>)` injected.

Hit in production on 26.3 (Cloud, us-west-2). Reproduced locally on **26.5.1.18**.

Yes (26.5.1.18).

### How to reproduce

```sql
CREATE TABLE t (a UUID, h DateTime, s String) ENGINE = Memory;
INSERT INTO t SELECT reinterpretAsUUID(toString(number % 3)),
                      now() - INTERVAL number HOUR,
                      concat('v', toString(number))
FROM numbers(10);

SELECT a, h, s
FROM t
ORDER BY a, h ASC
    WITH FILL STEP INTERVAL 1 HOUR
    INTERPOLATE (s)
LIMIT 1 BY a, h
SETTINGS enable_analyzer = 1;
```

Actual:

```
Code: 10. DB::Exception: Not found column s in block __table1.a UUID UUID(size = 0),
  __table1.h DateTime UInt32(size = 0), __table1.s String String(size = 0),
  materialize(__table1.a) UUID UUID(size = 0),
  materialize(__table1.h) DateTime UInt32(size = 0). (NOT_FOUND_COLUMN_IN_BLOCK)
```

### Variant matrix

| INTERPOLATE | LIMIT BY | \`enable_analyzer\` | Result |
|---|---|---|---|
| ✓ | ✓ | 1 | **fails** |
| ✓ | – | 1 | ok |
| – | ✓ | 1 | ok |
| ✓ | ✓ | **0** | ok (workaround) |

### Workaround with analyzer=1

Push dedup into a subquery, gap-fill on the outer:

```sql
SELECT a, h, s FROM (
    SELECT a, h, s FROM t
    ORDER BY a, h LIMIT 1 BY a, h
)
ORDER BY a, h ASC
    WITH FILL STEP INTERVAL 1 HOUR INTERPOLATE (s)
SETTINGS enable_analyzer = 1;
```

Rewriting the INTERPOLATE target as an expression (\`INTERPOLATE (s AS s)\`, \`INTERPOLATE (s AS toString(s))\`), or wrapping the base select in a CTE while keeping LIMIT BY + WITH FILL on the same query level, does not help — the crash reproduces identically.

### Related

- #103241 — same root cause identified from a retrospective review of #93359 (analyzer chain orders \`BEFORE_INTERPOLATE\` before \`BEFORE_LIMIT_BY\`, while the plan adds the LIMIT BY step before WITH FILL).
- #101279 — notes an earlier fix was incomplete for this combination.

### Version info
- Resolved by: #105481
- Merged into: `26.6.1.303`

yakov-olkhovskiy:
When `ORDER BY ... WITH FILL INTERPOLATE (...)` is combined with `LIMIT N BY` and `enable_analyzer = 1`, the query fails with:

```
Code: 10. DB::Exception: Not found column s in block __table1.a UUID, __table1.h DateTime,
  __table1.s String, materialize(__table1.a) UUID, materialize(__table1.h) DateTime.
  (NOT_FOUND_COLUMN_IN_BLOCK)
```

The root cause is a mismatch between the expression analysis chain order and the query plan execution order. During analysis in `buildExpressionAnalysisResult`, the `before_interpolate_actions` step was added to the chain inside `analyzeSort` — before the `before_limit_by` step. But in the query plan, the steps execute in the opposite order: Before LIMIT BY runs first, then Before INTERPOLATE.

During chain finalization (which propagates column requirements backward), the `before_limit_by` DAG was modified to expect interpolated column names (e.g. `s`) as inputs — columns that the `before_interpolate` step creates as aliases of their prefixed counterparts (`__table1.s`). At plan construction time, the Before LIMIT BY expression step runs first and calls `appendInputsForUnusedColumns`, which fails because `s` doesn't exist in the block yet (only `__table1.s` does).

The fix moves the `before_interpolate_actions` chain step from `analyzeSort` to after `analyzeLimitBy` in `buildExpressionAnalysisResult`, so the chain order matches the plan execution order.

### Changelog entry:
Fix `NOT_FOUND_COLUMN_IN_BLOCK` exception when combining `ORDER BY ... WITH FILL INTERPOLATE` and `LIMIT N BY` with the analyzer enabled. Closes #103474

### Version info
- Merged into: `26.6.1.303`


tests/queries/0_stateless/04259_interpolate_with_fill_limit_by.sql
```diff
@@ -0,0 +1,21 @@
+-- Tags: no-random-settings
```
tiandiwonder: no-random-settings can be removed?

tests/queries/0_stateless/04259_interpolate_with_fill_limit_by.sql
```diff
@@ -0,0 +1,21 @@
...
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/103474
+-- NOT_FOUND_COLUMN_IN_BLOCK when combining WITH FILL INTERPOLATE and LIMIT N BY
```
tiandiwonder, line 2: The current test covers the exact issue body, but two variants exercise the same chain change with slightly different DAG construction and are worth
  adding:
```
    - INTERPOLATE () (no explicit column list) — exercises the interpolate_list_nodes.empty() branch in Planner.cpp:993-1002.
    - INTERPOLATE (s AS upper(s)) — exercises the expression form, confirming a non-trivial INTERPOLATE expression still composes correctly after LIMIT
  BY.
```

src/Planner/PlannerExpressionAnalysis.cpp
```diff
@@ -580,12 +580,6 @@ SortAnalysisResult analyzeSort(
         before_interpolate_actions->dag = std::move(before_interpolate_actions_dag);
     }
 
-    if (before_interpolate_actions)
-    {
-        auto actions_step_before_interpolate = std::make_unique<ActionsChainStep>(before_interpolate_actions);
-        actions_chain.addStep(std::move(actions_step_before_interpolate));
-    }
-
     return SortAnalysisResult{std::move(before_sort_actions), has_with_fill, std::move(before_interpolate_actions)};
```
tiandiwonder, line 586: Better to drop a one-line pointer comment near here explaining that `before_interpolate_actions` is intentionally *not* added to the chain inside `analyzeSort` — the caller (`buildExpressionAnalysisResult`) appends it after `analyzeLimitBy` so the chain order matches the plan execution order.  It would prevent a future refactor from innocently moving the `addStep` call back inside `analyzeSort`.

tiandiwonder (changes requested):

yakov-olkhovskiy:
@tiandiwonder all review issues are addressed, but newly added test discovered another bug #105997 - fix PR is #106001 - we will need to merge it first in order to proceed

tests/queries/0_stateless/04259_interpolate_with_fill_limit_by.sql
```diff
@@ -0,0 +1,34 @@
...
+ORDER BY a, h ASC
+    WITH FILL STEP INTERVAL 1 HOUR
+    INTERPOLATE (s)
+LIMIT 1 BY a, h;
+
+SELECT a, h, s
+FROM t_interpolate_limit_by
+ORDER BY a, h ASC
+    WITH FILL STEP INTERVAL 1 HOUR
+    INTERPOLATE ()
```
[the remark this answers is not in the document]
yakov-olkhovskiy: we are going to wait for #106001 to be merged

tiandiwonder (approved):

PR #105484, opened 2026-05-21
Bump SystemLogQueue::waitFlush timeout to 600s

groeneai:
Bump the hardcoded `timeout_seconds` in `SystemLogQueue::waitFlush` from 180s to 600s.

The previous bump (60s -> 180s) is no longer sufficient on heavily loaded parallel CI runners. `SYSTEM FLUSH LOGS query_log` periodically throws:

```
Code: 159. DB::Exception: Received from localhost:9000.
DB::Exception: Timeout exceeded (180 s) while flushing system log
'DB::SystemLogQueue<DB::QueryLogElement>'. (TIMEOUT_EXCEEDED)
(query: SYSTEM FLUSH LOGS query_log;)
```

CIDB shows 23 master hits across 4 distinct PRs in the last 30 days for `02345_implicit_transaction` alone, spanning `amd_tsan parallel`, `amd_asan_ubsan distributed plan parallel`, `arm_binary parallel`, `amd_llvm_coverage`, `amd_debug parallel`, `arm_asan_ubsan azure`, and others. The common factor is `parallel` runners with heavy `query_log` backlog, not a specific sanitizer or storage backend.

The throw is server-side from `SystemLogQueue::waitFlush` and is not influenced by client `receive_timeout`. So bumping the server-side constant is the only way to give the flush more headroom under load. Updated the comment to reflect the latest tune-up rationale.

This is the third test in the bundle requested by @rschu1ze in #104648. Companion PRs:

- #105474 — `02435_rollback_cancelled_queries` (LSan leak check skip)
- #105457 — `01563_distributed_query_finish` (CI log-sender bursts)

...

### DIFF
--- a/src/Common/SystemLogBase.cpp
+++ b/src/Common/SystemLogBase.cpp
@@ -158,10 +158,12 @@ void SystemLogQueue<LogElement>::waitFlush(SystemLogQueue<LogElement>::Index exp
 {
     LOG_DEBUG(log, "Requested flush up to offset {}", expected_flushed_index);
 
-    // Use an arbitrary timeout to avoid endless waiting. 60s proved to be
-    // too fast for our parallel functional tests, probably because they
-    // heavily load the disk.
-    const int timeout_seconds = 180;
+    // Use an arbitrary timeout to avoid endless waiting. Earlier bumps
+    // (60s -> 180s) still produced TIMEOUT_EXCEEDED in `SYSTEM FLUSH LOGS`
+    // on heavily loaded parallel CI runners. Raised to 600s for the same
+    // reason: heavy parallel tests on slow disks can stall flush longer
+    // than 180s without anything actually going wrong.
+    const int timeout_seconds = 600;
 
     std::unique_lock lock(mutex);
 
--- a/src/Common/SystemLogBase.cpp
+++ b/src/Common/SystemLogBase.cpp
@@ -131,7 +131,11 @@ void SystemLogQueue<LogElement>::handleCrash()
 {
     if (settings.notify_flush_on_crash)
     {
-        waitFlush(getLastLogIndex(),  /* should_prepare_tables_anyway */ true);
+        /// `SignalHandlers.cpp` gives the crash path a hard ~303s wall before the default
+        /// handler kills the process, and `SystemLogs::handleCrash` walks every queue in
+        /// sequence. Keep the per-queue wait short so one stuck queue cannot starve the
+        /// remaining ones (in particular `crash_log`, which is flushed first).
+        waitFlush(getLastLogIndex(), /* should_prepare_tables_anyway */ true, /* timeout_seconds */ 60);
     }
 }
 
@@ -154,16 +158,9 @@ void SystemLogQueue<LogElement>::notifyFlush(SystemLogQueue<LogElement>::Index e
 }
 
 template <typename LogElement>
-void SystemLogQueue<LogElement>::waitFlush(SystemLogQueue<LogElement>::Index expected_flushed_index, bool should_prepare_tables_anyway)
+void SystemLogQueue<LogElement>::waitFlush(SystemLogQueue<LogElement>::Index expected_flushed_index, bool should_prepare_tables_anyway, int timeout_seconds)
 {
     LOG_DEBUG(log, "Requested flush up to offset {}", expected_flushed_index);
-
-    // Use an arbitrary timeout to avoid endless waiting. Earlier bumps
-    // (60s -> 180s) still produced TIMEOUT_EXCEEDED in `SYSTEM FLUSH LOGS`
-    // on heavily loaded parallel CI runners. Raised to 600s for the same
-    // reason: heavy parallel tests on slow disks can stall flush longer
-    // than 180s without anything actually going wrong.
-    const int timeout_seconds = 600;
 
     std::unique_lock lock(mutex);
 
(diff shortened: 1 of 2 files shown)


groeneai:
### Pre-PR validation gate

Session: cron:clickhouse-ci-task-worker:20260521-020000

**a) Deterministic repro?**

Not reproducible locally — the failure only occurs on heavily loaded parallel CI runners where multiple parallel tests saturate disk I/O. The mechanism is well understood from the previous 60s -> 180s bump (commit history of `SystemLogBase.cpp`) and from the existing comment that explicitly documented the parallel-load reason. Trying to reproduce with multiple shells flushing on the same instance does not match the disk contention of CI's parallel runner. The CIDB-confirmed pattern (parallel runners, mixed sanitizer / storage / planner variants, `SYSTEM FLUSH LOGS query_log` as the throwing query) is itself the evidence.

**b) Root cause explained?**

`SystemLogQueue<LogElement>::waitFlush` in `src/Common/SystemLogBase.cpp:164` uses a hardcoded `timeout_seconds = 180` to bound `confirm_event.wait_for`. On heavily loaded parallel runners, the background flush thread (`savingThreadFunction`) is bottlenecked behind disk writes and `prepared_tables >= expected_flushed_index` is not satisfied within 180s, so `wait_for` returns false and the function throws `TIMEOUT_EXCEEDED`. The client cannot influence this — `receive_timeout` is unrelated and the throw is server-side.

Yes. The root cause is that the upper bound on a bursty wait is too low for one specific deployment scenario (parallel CI), and the value is hardcoded. Raising the hardcoded value gives the wait more headroom while keeping the same "avoid endless waiting" property. The previous 60s -> 180s bump was made for exactly the same reason, by the same kind of fix. Promoting this to a server setting (option B in the worker plan) is a strictly larger change and was rejected as premature; if 600s also turns out to be insufficient, the next step is to introduce a setting.

**d) Test intent preserved? / new tests?**

`02345_implicit_transaction` already covers the behavior it cares about (implicit transactions, `transaction_id` column, system table interactions). The flaky surface is purely `SYSTEM FLUSH LOGS query_log` returning `TIMEOUT_EXCEEDED` before the data has been flushed. Bumping the timeout does not weaken any assertion in the test, does not pin any setting, does not skip anything. No new test is added because the underlying behavior (the wait succeeds when the flush completes, otherwise it raises an exception) is unchanged.

Strictly, no — the failure cannot be reliably reproduced on this hardware. What I did verify locally:

- Source change compiles cleanly (`ninja -C build clickhouse` rebuilt only `SystemLogBase.cpp.o` plus the relinks)
- Disassembly of `SystemLogQueue<QueryLogElement>::waitFlush` in the rebuilt binary contains `movl $0x258,0x68(%rsp)` (0x258 = 600), confirming the constant is encoded
- Running `SYSTEM FLUSH LOGS query_log` against the rebuilt server completes immediately under no load
- Test 02345 itself: lines 1-17 (the pre-Keeper segment) ran successfully under the rebuilt binary; lines after `BEGIN TRANSACTION` need a real Keeper which is not part of this minimal local server

**f) Fix is general?**

The change covers all `SystemLogQueue<LogElement>` template instantiations (`QueryLog`, `PartLog`, `TextLog`, `ErrorLog`, `ZooKeeperLog`, etc. — 14 total) because the constant lives in the shared template body. There is no symmetric path that needs the same fix elsewhere. Other `SYSTEM FLUSH` callers (e.g. `SystemLog::flush`) all dispatch into `waitFlush`, so they are covered by this single bump.

cc @rschu1ze @azat @alesapin

groeneai:
cc @rschu1ze @azat @alesapin — could you review this?

This is the third of the three flaky-test fixes you asked for in #104648 (`02345_implicit_transaction`). The other two are #105474 (`02435_rollback_cancelled_queries`) and #105457 (`01563_distributed_query_finish`).

One-line server-side change to `src/Common/SystemLogBase.cpp` bumping the hardcoded `SystemLogQueue::waitFlush` timeout from 180s to 600s. CIDB shows 23 master hits across 4 distinct PRs for `02345_implicit_transaction` alone over 30 days, all on `parallel` runners hitting `TIMEOUT_EXCEEDED` on `SYSTEM FLUSH LOGS query_log`. The previous 60s -> 180s bump was made for the same reason; raising again is consistent.

groeneai:
Thanks @clickhouse-gh for catching the crash-budget concern. Pushed `98b907ff30f6` to split the timeout.

`waitFlush` now takes an explicit `int timeout_seconds = 600` parameter.

* `SystemLogBase::flush` (the user-triggered `SYSTEM FLUSH LOGS` path the original commit was tuning for `02345_implicit_transaction`) keeps the 600s default.
* `SystemLogQueue::handleCrash` passes `60` explicitly so each per-queue wait fits comfortably under the ~303s signal-handler budget enforced in `SignalHandlers.cpp:148-160`, even when `SystemLogs::handleCrash` walks every queue sequentially and `crash_log` is flushed first.

Verified locally:

```
$ llvm-objdump -d --disassemble-symbols='_ZN2DB13SystemLogBaseINS_15CrashLogElementEE11handleCrashEv' build/programs/clickhouse | head
...
movl $0x3c, %ecx        # 60, crash path
jmp  ... waitFlush ...

$ llvm-objdump -d --disassemble-symbols='_ZN2DB13SystemLogBaseINS_15QueryLogElementEE5flushElb' build/programs/clickhouse | head
...
movl $0x258, %ecx       # 600, user-triggered SYSTEM FLUSH LOGS
jmp  ... waitFlush ...
```

Only two call sites in the codebase, both updated. PTAL.

Algunenano (changes requested):
@groeneai This not the proper way to address this problem. Bumping the flush times won't fix the problems, and will only need to more unrelated timeouts. If we need to optimize the process (like we've done several times in the past) then we need to profile and optimize

PR #105486, opened 2026-05-21
Restore PK/skip-index pruning for PREWHERE IN (subquery)

canhld94:
PREWHERE with IN (subquery) regressed in the 26.3: PK and skip indexes are silently bypassed, and force_data_skipping_indices raises Index ... is not used.

Upstream commit b9cbeb28a64 ("Build sets for PREWHERE and row-level filter in ReadFromMergeTree::applyFilters") fixes a separate Not-ready Set error by eagerly building PREWHERE sets, but uses the non-ordered variant. That populates the Set's hashmap without explicit elements, and the
later FutureSetFromSubquery::buildOrderedSetInplace call from MergeTreeIndexConditionSet / KeyCondition returns nullptr — so index analysis falls back to Condition: true.

Reproduce on latest master: https://fiddle.clickhouse.com/0e534fd7-c549-4549-9190-79b7c2ad3e84

Fix: use buildOrderedSetsForDAG so Set::fillSetElements runs before the pipeline. Runtime PREWHERE evaluation is unchanged; storage-side index analysis now sees the materialized set.

Changelog category (leave one):

Bug Fix (user-visible misbehavior in an official stable release)
Changelog entry (a user-readable short description of the changes that goes into CHANGELOG.md):

Fixed primary key and skip-index pruning being silently bypassed when a PREWHERE expression contains IN (subquery), which also broke the force_data_skipping_indices setting on such queries.


SaltTan:
Is this PR related? https://github.com/ClickHouse/ClickHouse/pull/102570

canhld94:
> Is this PR related? #102570

Not 100% sure, but looks like we're fixing the same issue

src/Processors/QueryPlan/ReadFromMergeTree.cpp
```diff
@@ -2352,10 +2334,14 @@ void ReadFromMergeTree::applyFilters(ActionDAGNodes added_filter_nodes)
         /// an empty right side) closes its inputs early, DelayedPortsProcessor may terminate
         /// the set-building pipeline before the set is ready.
         /// Building sets synchronously here eliminates this race condition entirely.
+        ///
+        /// Use the ordered variant so the Set is built with explicit elements: skip-index
+        /// and KeyCondition analysis below will call buildOrderedSetInplace and would
+        /// otherwise get nullptr from a set that was already built non-ordered.
         if (query_info.prewhere_info)
-            VirtualColumnUtils::buildSetsForDAG(query_info.prewhere_info->prewhere_actions, context);
+            VirtualColumnUtils::buildOrderedSetsForDAG(query_info.prewhere_info->prewhere_actions, context);
```
vdimir, line 2342: https://github.com/ClickHouse/ClickHouse/pull/102570#discussion_r3073328099

fm4v:
Fixed in https://github.com/ClickHouse/ClickHouse/pull/102570

PR #105487, opened 2026-05-21
Add TaskTracker::addFinal and async S3 multipart upload

asya-ch:
- Improvement

Complete the final S3 multipart upload request asynchronously via the task tracker, so it can overlap with the last part upload instead of running serially in finalize.

### Version info
- Merged into: `26.6.1.307`


src/Common/ThreadPoolTaskTracker.cpp
```diff
@@ -117,14 +117,25 @@ void TaskTracker::add(Callback && func)
...
+                std::lock_guard lock(mutex);
+                finished_futures.splice(finished_futures.end(), my_pre_allocated_finished);
+                /// "+ 1" accounts for final task: it has a futures entry but no finished_futures entry.
+                if (final_task && finished_futures.size() + 1 == futures.size())
+                {
+                    maybe_final_task = std::move(final_task);
+                }
+                has_finished.notify_one();
+            }
+            if (maybe_final_task)
```
asya-ch, line 142: Not 100% sure if it's safe to kick off the scheduler within SCOPE_EXIT -- input appreciated.

src/Common/ThreadPoolTaskTracker.h
```diff
@@ -39,6 +39,11 @@ class TaskTracker
 
     void add(Callback && func);
```
CheSema, line 40: lets harden the add method: if the final task is set, then no new ordinary task could be added. It is logical error.

src/Common/ThreadPoolTaskTracker.cpp
CheSema: scope exit is called in the d-tor of some object. it has to no throw otherwise it terminates.
I choose `SCOPE_EXIT` because I wanted to call it even when ` my_func();` throws.
This place need something around `scheduler()` call, it might throw exception as I understand.

src/IO/WriteBufferFromS3.cpp
```diff
@@ -195,6 +195,11 @@ void WriteBufferFromS3::preFinalize()
     else
     {
         writeMultipartUpload();
+        task_tracker->addFinal([this]()
```
CheSema, line 198: cool, all the changes and difficulties are in the common tool.

src/Common/ThreadPoolTaskTracker.cpp
```diff
@@ -136,6 +147,33 @@ void TaskTracker::add(Callback && func)
...
+        }
+        else
+        {
+            final_task = pt;
+        }
+    }
+    if (run_final_task_now)
+    {
+        scheduler([p = std::move(pt)]() mutable { (*p)(); }, Priority{});
+        waitTilInflightShrink();
```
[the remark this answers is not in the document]
asya-ch: Will skip `waitTilInflightShrink`.

src/IO/WriteBufferFromS3.cpp
```diff
@@ -627,7 +626,7 @@ void WriteBufferFromS3::completeMultipartUpload()
         const auto tag = multipart_tags.at(i);
         if (tag.empty())
             throw Exception(
-                    ErrorCodes::LOGICAL_ERROR,
+                    ErrorCodes::S3_ERROR,
```
asya-ch: No longer a logical error -- we now reach completeMultipartUpload even if an earlier upload fails. Need to throw in this case.

Per the logic in TaskTracker, the earlier upload failure will be surfaced first.

CheSema: You are right here.
lets do not throw ErrorCodes::S3_ERROR, this is not the error from s3 client and not the error at all
we could throw some generic like `canceled` + `due to previous errors` or even just ignore it and do not set the flag that MPU is finished. But later we need to make sure that is MPU is not finished that we have some error from previous steps.

asya-ch: MPU will not be marked as finished if we throw from this function. And sure, will change to a more generic code.

src/IO/WriteBufferFromS3.cpp
```diff
@@ -622,13 +621,13 @@ void WriteBufferFromS3::completeMultipartUpload()
...
+    for (const auto & tag : multipart_tags)
     {
-        const auto tag = multipart_tags.at(i);
         if (tag.empty())
-            throw Exception(
-                    ErrorCodes::LOGICAL_ERROR,
-                    "Failed to complete multipart upload. Part {} haven't been uploaded.", i);
+        {
+            // One of the earlier uploads failed.
+            return false;
```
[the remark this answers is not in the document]
asya-ch: I don't think there is an issue here. As mentioned, we will still surface the real part-upload exception. So `finalizeImpl` will not return success.

CheSema (approved):

alexey-milovidov:
@asya-ch, this has a bug, see: https://github.com/ClickHouse/ClickHouse/pull/108730

PR #105491, opened 2026-05-21
Intorduce analyzer_compatibility_prefer_alias_over_subcolumn

vdimir:

* Add a new compatibility setting `analyzer_compatibility_prefer_alias_over_subcolumn` (disabled by default). When enabled, the new analyzer prefers the alias-prefix interpretation over Tuple-subcolumn / dotted-column matches for multi-part identifiers, restoring the previous interpreter's behavior. This avoids `AMBIGUOUS_IDENTIFIER` (and related) errors when a query joins a table whose name matches an inner table of a CTE/subquery that uses `SELECT *` over a join, where asterisk-renamed columns (e.g. `b.id`) would otherwise leak the inner table identifiers into the outer scope.

### Version info
- Merged into: `26.6.1.456` (included in `26.6` and later)
- Backported to: `26.5.6.54`, `26.4.5.159`


src/Analyzer/Resolve/IdentifierResolver.cpp
```diff
@@ -759,6 +759,29 @@ IdentifierResolveResult IdentifierResolver::tryResolveIdentifierFromTableExpress
...
+      * sibling subquery exposes an asterisk-renamed `b.id` column.
+      */
+    if (scope.context->getSettingsRef()[Setting::analyzer_compatibility_prefer_alias_over_subcolumn]
+        && identifier.getPartsSize() > 1)
+    {
+        const auto & table_name_compat = table_expression_data.table_name;
+        const bool prefix_matches_table_name = !table_name_compat.empty() && path_start == table_name_compat;
+        const bool prefix_matches_alias
+            = table_expression_node->hasAlias() && path_start == table_expression_node->getAlias();
+        if (prefix_matches_table_name || prefix_matches_alias)
```
KochetovNicolai, line 777: Note: probably, the same case if possible with database, e.g. when identifier can be resolved to `db.table.column` or `column.subcolumn.subcolumn` at the same time.

But also I think it makes sense to ignore it.

KochetovNicolai (approved):

PR #105495, opened 2026-05-21
Fix L2Distance() return value in vector search query executed with index only optimization

### LINKED ISSUE #102480: `L2Distance` returns squared L2 distance instead of Euclidean distance when rescoring optimization is active

_Retrospective finding from a historical scan of [PR #84644](https://github.com/ClickHouse/ClickHouse/pull/84644) (merged 2025-08-03). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

When `vector_search_with_rescoring = false` (the default after this PR), `SELECT L2Distance(vec, ref)` returns L2-squared values (sum of squared differences) instead of the actual Euclidean distance (sqrt of sum of squared differences). For example, L2Distance([1,1], [0,0]) returns 2 instead of 1.4142.

**Root cause:** useVectorSearch.cpp:406-414: The optimization replaces the L2Distance function output with a direct alias to `_distance` (cast to Float64), but `_distance` contains L2-squared values from the USearch index. A sqrt transformation is needed but missing.

**Why we believe this is a bug:** useVectorSearch.cpp:409 replaces the `L2Distance` function node in the expression DAG with `_distance` input from the USearch index. MergeTreeIndexVectorSimilarity.cpp:77 maps `L2Distance` to USearch's `l2sq_k` metric (squared L2). The index returns sum((xi-yi)²) but the `L2Distance` function (arrayDistance.cpp:160) returns sqrt(sum((xi-yi)²)). The optimization substitutes L2² for L2 without applying sqrt.

**Affected locations:**
- `src/Processors/QueryPlan/Optimizations/useVectorSearch.cpp:409` — addInput _distance without sqrt transformation
- `src/Storages/MergeTree/MergeTreeIndexVectorSimilarity.cpp:77` — L2Distance maps to l2sq_k (squared L2) metric
- `src/Functions/array/arrayDistance.cpp:160` — L2Distance finalize applies sqrt(state.sum)

**Impact:** All users with vector similarity indexes using `L2Distance` who SELECT distance values get wrong absolute distances by default. ORDER BY results are correct (sqrt is monotonic), but the distance values shown to users are L2² instead of L2. This is a silent wrong-result when the user reads the distance column. `cosineDistance` is NOT affected.

Yes — confirmed on current `master` (commit `9678bc3a5d5e`).

### How to reproduce

```sql
-- Test: L2Distance values in SELECT should return Euclidean distance (with sqrt), not squared distance,
--       regardless of whether the rescoring optimization is active or not.
-- Covers: useVectorSearch.cpp optimization replaces L2Distance with _distance from USearch index (l2sq_k metric)

SET enable_analyzer = 1;
SET parallel_replicas_local_plan = 1;

DROP TABLE IF EXISTS tab_l2_mismatch;

CREATE TABLE tab_l2_mismatch (
    id Int32,
    vec Array(Float32),
    INDEX idx vec TYPE vector_similarity('hnsw', 'L2Distance', 2)
)
ENGINE = MergeTree
ORDER BY id
SETTINGS index_granularity = 2;

-- Vectors chosen so L2 and L2-squared are clearly different
-- L2([1,1], [0,0]) = sqrt(2) = 1.41421...
-- L2^2([1,1], [0,0]) = 2
INSERT INTO tab_l2_mismatch VALUES (1, [0.0, 0.0]), (2, [1.0, 0.0]), (3, [0.0, 1.0]), (4, [1.0, 1.0]);

-- With rescoring ON: L2Distance function runs on actual data, correct sqrt values
SELECT 'rescoring=true (baseline)';
SELECT id, round(L2Distance(vec, [0.0, 0.0]), 4) as dist
FROM tab_l2_mismatch
ORDER BY L2Distance(vec, [0.0, 0.0])
LIMIT 4
SETTINGS vector_search_with_rescoring = 1;

-- With rescoring OFF (default): optimization replaces L2Distance with _distance from index
-- BUG: _distance contains L2-squared from USearch (l2sq_k metric), NOT sqrt'd Euclidean distance
-- Expected: same values as above (sqrt applied)
-- Actual: squared values (no sqrt) e.g. id=4 shows 2 instead of 1.4142
SELECT 'rescoring=false (default, optimization active)';
SELECT id, round(L2Distance(vec, [0.0, 0.0]), 4) as dist
FROM tab_l2_mismatch
ORDER BY L2Distance(vec, [0.0, 0.0])
LIMIT 4
SETTINGS vector_search_with_rescoring = 0;

DROP TABLE tab_l2_mismatch;
```

### Expected behavior

```
rescoring=true (baseline)
1	0
2	1
3	1
4	1.4142
rescoring=false (default, optimization active)
1	0
2	1
3	1
4	1.4142
```

```
rescoring=true (baseline)
1	0
2	1
3	1
4	1.4142
rescoring=false (default, optimization active)
1	0
2	1
3	1
4	2
```

### Additional context

**Open risks:**
- Users may have built applications relying on the (wrong) squared distance values since this PR was merged in Aug 2025
- The fix needs to handle both L2Distance (needs sqrt) and cosineDistance (no sqrt needed) differently

**Suggested fix:** When the distance function is `L2Distance`, add a `sqrt` function node between `_distance` and the alias in the expression DAG. E.g., after `addInput("_distance", ...)`, compute `sqrt(_distance)` before aliasing it to the sort column name. Alternatively, change the USearch metric for L2Distance from `l2sq_k` to a custom metric that returns non-squared values, or store `sqrt(distance)` in the `_distance` column in `fillDistanceColumnAndFilterForVectorSearch`.

Found during automated review of [PR #84644](https://github.com/ClickHouse/ClickHouse/pull/84644).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr84644_001`_

shankar-iyer:
Resolves https://github.com/ClickHouse/ClickHouse/issues/102480
### Changelog category

### Changelog entry 
- A vector search query executed with the _index only_ optimization now returns the correct value for `L2Distance()` if it is in the `SELECT` list. The function was incorrectly returning the _L2-squared_ distance value. Note that the vector search query was returning the correctly ranked results, only applications **explicitly retrieving and interpreting** the `L2Distance()` value would have seen the squared value.

### Version info
- Merged into: `26.6.1.282`


shankar-iyer:
Clean CI run!

alexey-milovidov (approved):

PR #105505, opened 2026-05-21
Add native support for GCS

### LINKED ISSUE #91012: Contributing GCS GRPC support
### Company or project name

Shopify

### Question

Hi team,

I'm looking at potentially contributing a new feature to the [ClickHouse repo](https://github.com/ClickHouse/ClickHouse) and want to verify the right approach. 

I am considering adding support for GCS storage using gRPC which would require the addition of the google-cloud-cpp [storage-grpc](https://github.com/googleapis/google-cloud-cpp/tree/main/google/cloud/storage) pack. 

The general approach would require adding the `storage-gprc` package to the submodules and then creating a [new disk object storage type](https://github.com/Shopify/clickhouse/tree/759bc8dc04f914c39b0519759a5ead98bc177256/src/Disks/DiskObjectStorage/ObjectStorages). This is what I'm thinking anyways.

I just want to also verify that writing a rust plugin or some other approach isn't preferred.

tanner-bruce:
- New Feature

- adds support for using gRPC DirectPath connectivity to GCS. Enables using workload identity for GCS access through the existing XML multipart workflow

This is obviously a very large PR, I've been testing quite thoroughly with benchmarks and in typical case it is quite a bit faster, but there is some nuance to it. I put this together to help out with testing. https://github.com/tanner-bruce/gcp-gcs-upload-benchmarks

Main questions:
- I originally added support for specifying profile events as `GCS` specific, which required touching a bunch of the S3 code, but opted to remove it to limit where this PR touched. See bdfae65fe7675c187d8085d983119be01f7e262d for what the removal diff looked like. Is something like this desirable? I saw Azure has its own, but it goes through a separate code path. The `native/gRPC` code path does use the GCS profile events. This could easily be misleading to users.
- Is the re-use of the S3 poco/xml multipart workflow acceptable?

Notes:
- The write performance of gRPC is much lower than XML workflow, which is what brought on supporting the xml workflow still
- I'm just finishing final benchmark suite and will post results here shortly.
- if desired I can split this up into multiple PR, like `gRPC` first and then `XML`, or other way

Check close:
- I'm not experienced with CMake. The modifications made in the CMake were fully driven by AI. It compiles and I haven't seen issue.

Fixes #91012


tanner-bruce:
## Native `GCS` / XML multipart benchmark summary

I ran a focused benchmark set on the native `GCS` branch to cover the cases most relevant to this PR:

- single large realistic part upload
- full scan from `GCS` on a single large realistic part
- reading multiple large parts
- reading many small parts
- same logical bytes with different active part shapes

I did not include the multipart abort benchmark in this PR comment.

### Environment

- Branch: `gcs-grpc-high-level-clean`
- Commit: `953d90e8a75`
- ClickHouse version: `26.5.1.1`
- Pod: `clickhouse-builder-0`
- Bucket: `shp-clickhouse-gcs-staging`
- Read benchmarks disabled/dropped ClickHouse-side caches before each measured query:
  - query cache
  - mark cache
  - uncompressed cache
  - filesystem cache
  - page cache / object-storage page cache
  - distributed/local remote-storage caches where applicable

---

## 1. Single large realistic part upload

Source table:

- Rows: `40,000,000`
- Compressed bytes: `25,325,893,619` (`23.59 GiB`)
- Uncompressed bytes: `106,052,517,928` (`98.77 GiB`)
- Active parts: `1`

Operation:

```sql
ALTER TABLE target ATTACH PARTITION ID '1778709600'
FROM p04_real_part.core_n4_v0_base
```

| Variant | Seconds | Effective upload MiB/s | Validation |
|---|---:|---:|---|
| `s3` | `157.985` | `152.9` | pass |
| `xml` | `69.277` | `348.6` | pass |
| `xml_64m20` | `66.618` | `362.6` | pass |
| `native` | `76.652` | `315.1` | pass |

Takeaway: XML multipart was the fastest upload path in this run. Native `GCS` upload was also much faster than `GCS`-as-`s3`.

---

## 2. Single large realistic part full scan

Source database:

- Each table: `40,000,000` rows, `23.59 GiB` compressed, `1` active part

Query shape:

```sql
SELECT
    sum(length(message))
    + sum(length(`attrs.url.query`))
    + sum(arraySum(arrayMap(x -> length(x), mapValues(json_attributes))))
    + sum(arraySum(arrayMap(x -> length(x), mapValues(string_attributes))))
    + sum(arraySum(arrayMap(arr -> arraySum(arrayMap(x -> length(x), arr)), mapValues(string_array_attributes)))) AS payload_bytes
FROM table
SETTINGS
    use_query_cache = 0,
    use_uncompressed_cache = 0,
    enable_filesystem_cache = 0,
    use_page_cache_for_object_storage = 0,
    max_threads = 4
```

| Variant | Runs (s) | Avg (s) | Median (s) | Avg read bytes/query | Avg object read requests/query |
|---|---:|---:|---:|---:|---:|
| `s3` | `184.8`, `179.1`, `159.4` | `174.440` | `179.148` | `~9.05 GiB` | `~31.7k` |
| `xml` | `102.7`, `109.7`, `101.4` | `104.595` | `102.656` | `~9.06 GiB` | `~31.6k` |
| `xml_64m20` | `103.5`, `102.6`, `108.8` | `104.947` | `103.474` | `~9.06 GiB` | `~31.6k` |
| `native` | `110.7`, `135.5`, `80.0` | `108.771` | `110.745` | `~9.06 GiB` | `~31.6k` |

Takeaway: XML tables read through the native `GCS` read path, as expected. The native-read variants were much faster than `GCS`-as-`s3` for this full scan, though native `GCS` still showed visible run-to-run variance.

---

## 3. Same bytes, different active part shape

### 3a. Full read by part shape

Query:

```sql
SELECT count(), sum(length(payload)), sum(metric)
FROM table
SETTINGS ... no ClickHouse-side caches ..., max_threads = 4
```

| Shape | Native avg (s) | Native median (s) | S3 avg (s) | S3 median (s) | Avg read requests/query |
|---|---:|---:|---:|---:|---:|
| `~1` part | `2.175` | `2.242` | `1.858` | `1.824` | `~224` |
| `16` parts | `1.541` | `1.484` | `0.821` | `0.805` | `~224` |
| `256` parts | `9.371` | `9.249` | `16.277` | `16.304` | `~1,024` |
| `1024` parts | `48.741` | `44.872` | `70.601` | `70.640` | `~4,096` |

Takeaway: once active part count gets high, native `GCS` reads scale better than `GCS`-as-`s3` for the same logical full scan.

`native` is susceptible to cold start

### 3b. Subset / partition-pruned read

Query reads partitions in `[parts / 4, parts / 2)`. For the `~1`-part case this degenerates to the single partition.

| Shape | Native avg (s) | Native median (s) | S3 avg (s) | S3 median (s) | Avg read requests/query |
|---|---:|---:|---:|---:|---:|
| `~1` part | `2.228` | `2.242` | `1.588` | `1.588` | `~36-39` |
| `16` parts | `2.760` | `2.745` | `8.721` | `8.589` | `~68` |
| `256` parts | `1.620` | `1.164` | `1.568` | `0.812` | `~405` |
| `1024` parts | `16.764` | `14.927` | `23.564` | `21.139` | `~1,621` |

Takeaway: native `GCS` was much better on the `16`-part subset and better on the `1024`-part subset. The `256`-part case was effectively tied/noisy.

### 3c. Point/index lookup

Query:

```sql
SELECT count(), sum(length(payload)), sum(metric)
FROM table
WHERE token = '<sample token>'
SETTINGS ... no ClickHouse-side caches ..., max_threads = 4
```

| Shape | Native avg (s) | Native median (s) | S3 avg (s) | S3 median (s) | Avg read requests/query |
|---|---:|---:|---:|---:|---:|
| `~1` part | `3.066` | `2.807` | `2.555` | `2.668` | `~308-316` |
| `16` parts | `3.150` | `2.848` | `4.015` | `3.954` | `~368-369` |
| `256` parts | `9.094` | `7.595` | `14.958` | `13.279` | `~977` |
| `1024` parts | `24.940` | `16.982` | `27.767` | `20.109` | `~2,029` |

Takeaway: native `GCS` wins from `16` parts onward, with variance on high part counts.

### 3d. `EXPLAIN indexes = 1` point lookup

| Shape | Native avg (s) | Native median (s) | S3 avg (s) | S3 median (s) | Avg read requests/query |
|---|---:|---:|---:|---:|---:|
| `~1` part | `1.005` | `0.999` | `0.748` | `0.686` | `1-2` |
| `16` parts | `0.555` | `0.466` | `1.450` | `1.402` | `16` |
| `256` parts | `4.128` | `4.231` | `10.827` | `10.715` | `256` |
| `1024` parts | `15.058` | `15.015` | `27.226` | `27.239` | `1024` |

Takeaway: index analysis is very sensitive to active part count. Native `GCS` is significantly faster for many-part shapes.

---

## Coverage summary

This benchmark set covers:

- Single large part upload: realistic `23.59 GiB` part upload.
- Full scan from `GCS` on a single large part: realistic `23.59 GiB` one-part scan.
- Reading multiple large parts: `16`-part, `4 GiB` synthetic shape.
- Reading many small parts: `256`-part and `1024`-part synthetic shapes.
- Same bytes / different part shape: `~1`, `16`, `256`, and `1024` active part shapes, each around `4 GiB` compressed.

## Caveats

- The shape full scan reads `payload` and `metric`, not every column. It is still a full-row scan over all active parts, but not a full-byte scan of every column file.

alexey-milovidov:
Continue in #109455.

alexey-milovidov:
🕵 Thank you @tanner-bruce for this work. The native GCS integration continues in https://github.com/ClickHouse/ClickHouse/pull/109455, and its changelog entry credits you with the initial implementation. Here is how the two differ.

**Transport.** This PR uses the gRPC client, with optional DirectPath (`google-c2p://`). #109455 uses the GCS JSON REST API of the vendored `google-cloud-cpp`, over a Poco-based transport instead of libcurl. So it builds without gRPC or opentelemetry (the build problem noted in #91012), and it goes through ClickHouse's own connection pool, proxy resolver, keep-alive and TLS handling. DirectPath is not available there.

**Writes.** This PR can also write through the S3 XML multipart path (`write_transport = xml_multipart`), and through gRPC compose-based parallel upload. #109455 does neither. It writes each object as one resumable upload stream, and anyone who wants the XML multipart path keeps the existing S3-compatibility path (`use_native_gcs = 0`, or an `s3` disk). Given your benchmark (XML multipart ~349 MiB/s, compose ~315 MiB/s, S3-compatibility ~153 MiB/s), parallel upload is the main thing #109455 still lacks.

**Scope and surfaces.** This PR adds a disk type. #109455 adds the `gcs` object-storage and disk type, dynamic `disk(object_storage_type = gcs, ...)`, the `gcs(...)` table function (behind the experimental `use_native_gcs` setting), and `Iceberg` tables on a native GCS disk. `ENGINE = GCS` stays on the S3-compatibility path.

**Credentials.** Both support Application Default Credentials (including workload identity) and an inline service-account key. #109455 adds a service-account key file (server config only), the `google_adc_*` refresh-token flow with automatic renewal, and anonymous access (`NOSIGN` / `use_environment_credentials = 0`). It rejects S3-only keys (HMAC, `role_arn`, SSE, ...) with an error instead of ignoring them.

**Reliability.**
- Retries: a bounded `LimitedErrorCountRetryPolicy` (`retry_attempts`, default 10) with exponential backoff, instead of the SDK's 15-minute default.
- Reads: every ranged read is pinned to the object generation (`IfGenerationMatch`), so a concurrent overwrite cannot be stitched into one read.
- Conditional writes and server-side copies enforce `If-None-Match: *` / `If-Match` as generation preconditions.
- Listing streams page by page instead of collecting everything into memory.

**Not ported (yet).**
- Parallel/composite upload.
- gRPC and DirectPath.
- `user_project` as a setting (a `<header>x-goog-user-project: ...</header>` entry works meanwhile).
- The per-attempt, retry and latency profile counters. #109455 has per-operation counters only.

Request-rate throttlers (`max_get_rps` / `max_put_rps`, as for S3) are being added to #109455 now.

PR #100573, opened 2026-03-24
Fix Block structure mismatch exception in MaterializingCTETransform

alexey-milovidov:
`addMaterializingCTETransform` was not calling `dropTotalsAndExtremes` before transforming the pipeline, unlike the similar `addCreatingSetsTransform`. When a CTE with `WITH TOTALS` was materialized and referenced multiple times, the totals ports survived with the original header while the main output became empty. When uniting multiple CTE pipelines, `uniteTotals` tried to connect these non-empty totals ports to a `ConcatProcessor` created with the empty common header, causing a Block structure mismatch.

Found in: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100404&sha=4e239671fdc3523725369f5ad6028b4a52c43b45&name_0=PR&name_1=AST%20fuzzer%20%28amd_debug%2C%20targeted%2C%20old_compatibility%29
https://github.com/ClickHouse/ClickHouse/pull/100404

### Changelog category (leave one):
- Not For Changelog

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix Block structure mismatch exception when using materialized CTEs with `WITH TOTALS` referenced multiple times.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.848`
- Backported to: `26.3.9.2`

PR #100398, opened 2026-03-22
Fix exception when building set from correlated subquery with PLACEHOLDER actions

alexey-milovidov:
The AST fuzzer produces queries where a correlated subquery (containing `PLACEHOLDER` action nodes in its `ActionsDAG`) ends up as an IN clause argument. When `KeyCondition::tryPrepareSetIndexForIn` calls `FutureSetFromSubquery::buildOrderedSetInplace`, it attempts to build and execute the query pipeline for the subquery. Since `PLACEHOLDER` nodes cannot be executed standalone (they require decorrelation first), this causes a "Trying to execute PLACEHOLDER action" logical error.

The fix adds a check in both `buildSetInplace` and `buildOrderedSetInplace` to detect correlated expressions in the source plan before attempting to build it. If the plan contains `PLACEHOLDER` nodes, we skip building the set inplace — the correlated subquery will be decorrelated and executed as part of the outer query instead.

Also changes the default `IQueryPlanStep::hasCorrelatedExpressions` to return `false` instead of throwing `NOT_IMPLEMENTED`, since steps without `ActionsDAG` cannot contain `PLACEHOLDER` nodes.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100348&sha=f75d6efc8e1be0e658cfc4841e81729659ef3010&name_0=PR&name_1=Stress%20test%20%28arm_asan_ubsan%2C%20s3%29
https://github.com/ClickHouse/ClickHouse/pull/100348

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix "Trying to execute PLACEHOLDER action" exception that could occur during stress testing with the AST fuzzer when a correlated subquery appeared in an IN clause.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.479`

PR #82414, opened 2025-06-23
Workload scheduling: Memory reservations

serxa:
### Changelog category (leave one):
- New Feature

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Introduce a memory reservation feature for workloads. More details https://clickhouse.com/docs/operations/workload-scheduling

### Documentation entry for user-facing changes

- [X] Documentation is written (mandatory for new features)

### Details

To enable memory reservations for workloads create MEMORY RESERVATION resource and set at least one limit for the total memory reserved using workload settings:

```sql
CREATE RESOURCE memory (MEMORY RESERVATION)
CREATE WORKLOAD all SETTINGS max_memory = '2Gi'
```

ClickHouse tracks memory allocations of all queries and background activities. The number of allocated bytes is aggregated through the scheduling hierarchy up to the root. Every query has an associated allocation in the leaf workload it belongs to. If a query has the `reserve_memory` setting greater than zero, then the allocation is created in a pending state. Pending allocation reserves requested amount of memory in the workload hierarchy. If there is not enough memory available, the allocation remains pending until enough memory is freed or other allocations are evicted (killed). When allocation is admitted, it becomes running. Running allocation could increase or decrease its size dynamically according to memory consumption of the query. Allocation life-cycle can be depicted with the following state diagram:

```mermaid
stateDiagram-v2
    [*] --> Pending: init [reserve_memory > 0]
    [*] --> Running: init [reserve_memory == 0]

    Pending --> Running: admit

    state Running {
        %% Region 1: increase flow
        NotIncreasing --> Increasing: request
        Increasing --> NotIncreasing: approve

        --

        %% Region 2: decrease flow
        NotDecreasing --> Decreasing: request
        Decreasing --> NotDecreasing: approve
    }


    Running --> Killed: evict
    Running --> Released: finish
```

Pending allocations of a leaf workload are admitted according to FIFO order. When multiple workloads have pending allocations, they are admitted according to precedence and weight settings. Higher precedence workloads are served first. Sibling workloads with the same precedence share memory according to weights in a max-min fair manner, which means that workload with lower normalized memory usage (current usage plus requested increase divided by weight) is served first. The reverse logic is applied during eviction. When memory needs to be freed, workloads with lower precedence and higher normalized memory usage are evicted first.

[...]

Issue #80564, opened 2025-05-20
Stress tests: Column X Int32 query tree node does not have valid source node


fm4v:
### Describe what's wrong

[Stress report](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=80103&sha=165241b6050e91acffa4f4b55f5897fa43461f8d&name_0=PR&name_1=Stress%20test%20%28amd_tsan%29)

Test `02526_merge_join_int_decimal` or `01783_merge_engine_join_key_condition`

Stacktrace:
```
/var/log/clickhouse-server/clickhouse-server.err.log:2025.05.19 16:58:15.852990 [ 12196 ] {37f81c10-88c8-4714-a055-e858e6b21c4d} <Fatal> : Logical error: 'Column X Int32 query tree node does not have valid source node'.
/var/log/clickhouse-server/clickhouse-server.err.log:2025.05.19 16:58:16.337148 [ 12196 ] {37f81c10-88c8-4714-a055-e858e6b21c4d} <Fatal> : Stack trace (when copying this message, always include the lines below):
/var/log/clickhouse-server/clickhouse-server.err.log:2025.05.19 16:59:40.311473 [ 109079 ] {} <Fatal> BaseDaemon: ########## Short fault info ############
/var/log/clickhouse-server/clickhouse-server.err.log:2025.05.19 16:59:40.311674 [ 109079 ] {} <Fatal> BaseDaemon: (version 25.6.1.1, build id: 9ABC726A97966173A19B247F7BECAB968D148E1B, git hash: 55c2758bf86724b32cae8082a0f978c7b38b5261, architecture: x86_64) (from thread 12196) Received signal 6
/var/log/clickhouse-server/clickhouse-server.err.log:2025.05.19 16:59:40.311809 [ 109079 ] {} <Fatal> BaseDaemon: Signal description: Aborted
/var/log/clickhouse-server/clickhouse-server.err.log:2025.05.19 16:59:40.311892 [ 109079 ] {} <Fatal> BaseDaemon: 
[...]

PR #105518, opened 2026-05-21
Reject correlated MATERIALIZED CTE used in IN at analysis time

### LINKED ISSUE #105501: Logical error: Trying to execute PLACEHOLDER action (STID: 1611-343c)

Test name: Logical error: Trying to execute PLACEHOLDER action (STID: 1611-343c)
CI report: [AST fuzzer (amd_debug)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=6d9d7738d4ee56e628ee1109fabae2d0c0e05c6b&name_0=MasterCI&name_1=AST%20fuzzer%20%28amd_debug%29)
Failing test history: [cidb](https://play.clickhouse.com/play?user=play&run=1#V0lUSAogICAgOTAgQVMgaW50ZXJ2YWxfZGF5cwpTRUxFQ1QKICAgIHRvU3RhcnRPZkRheShjaGVja19zdGFydF90aW1lKSBBUyBkYXksCiAgICBjb3VudCgpIEFTIGZhaWx1cmVzLAogICAgZ3JvdXBVbmlxQXJyYXkocHVsbF9yZXF1ZXN0X251bWJlcikgQVMgcHJzLAogICAgYW55KHJlcG9ydF91cmwpIEFTIHJlcG9ydF91cmwKRlJPTSBjaGVja3MKV0hFUkUgKG5vdygpIC0gdG9JbnRlcnZhbERheShpbnRlcnZhbF9kYXlzKSkgPD0gY2hlY2tfc3RhcnRfdGltZQogICAgQU5EIHRlc3RfbmFtZSA9ICdMb2dpY2FsIGVycm9yOiBUcnlpbmcgdG8gZXhlY3V0ZSBQTEFDRUhPTERFUiBhY3Rpb24gKFNUSUQ6IDE2MTEtMzQzYyknCiAgICAtLSBBTkQgY2hlY2tfbmFtZSA9ICdBU1QgZnV6emVyIChhbWRfZGVidWcpJwogICAgQU5EIHRlc3Rfc3RhdHVzIElOICgnRkFJTCcsICdFUlJPUicpCiAgICBBTkQgKHB1bGxfcmVxdWVzdF9udW1iZXIgPSAwIE9SIGJhc2VfcmVmIElOICgnJywgJ21hc3RlcicpKQogICAgQU5EIHRlc3RfY29udGV4dF9yYXcgTElLRSAnJUV4Y2VwdGlvbjolJwpHUk9VUCBCWSBkYXkKT1JERVIgQlkgZGF5IERFU0MK)

Test output:
```
Error:
Logical error: 'Trying to execute PLACEHOLDER action'.
---

Failed query:
WITH t AS MATERIALIZED (SELECT number, number + 65535 FROM zeros(2147483647)) SELECT * FROM numbers(8) WHERE (number + 3, number + 1, number + 3) IN (t)
---

Reproduce commands (auto-generated; may require manual adjustment):
WITH t AS MATERIALIZED (SELECT number, number + 65535 FROM zeros(2147483647)) SELECT * FROM numbers(8) WHERE (number + 3, number + 1, number + 3) IN (t);
DROP TABLE IF EXISTS zeros(2147483647);
---

Stack trace:
pthread_kill @ 0x00000000000969fd
raise @ 0x0000000000042476
__lgamma_r_finite@GLIBC_2.15 @ 0x00000000000287f3
src/Common/Exception.cpp:60:5: DB::abortOnFailedAssertion(String const&, std::basic_string_view<char, std::char_traits<char>>, void* const*, unsigned long, unsigned long) @ 0x000000001409cbb6
src/Common/Exception.cpp:93:13: DB::handle_error_code(String const&, std::basic_string_view<char, std::char_traits<char>>, int, bool, std::vector<void*, std::allocator<void*>> const&) @ 0x000000001409ddd3
src/Common/Exception.cpp:146:19: DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0x000000001409df99
src/Common/Exception.h:171:100: DB::Exception::Exception(String&&, int, String, bool) @ 0x000000000cddf894
src/Common/Exception.h:57:54: DB::Exception::Exception(PreformattedMessage&&, int) @ 0x000000000cddf326
src/Common/Exception.h:189:77: DB::Exception::Exception<>(int, FormatStringHelperImpl<>) @ 0x000000001409c639
src/Interpreters/ExpressionActions.cpp:797: DB::executeAction(DB::ExpressionActions::Action const&, DB::(anonymous namespace)::ExecutionContext&, bool, bool, bool)
src/Interpreters/ExpressionActions.cpp:837:19: DB::ExpressionActions::execute(DB::Block&, unsigned long&, bool, bool, std::function<bool ()>) const @ 0x0000000019619b5f
src/Processors/Transforms/ExpressionTransform.cpp:31:17: DB::ExpressionTransform::transform(DB::Chunk&) @ 0x000000001ef9db7b
src/Processors/ISimpleTransform.h:33:9: DB::ISimpleTransform::transform(DB::Chunk&, DB::Chunk&) @ 0x00000000163f8893
src/Processors/ISimpleTransform.cpp:100:9: DB::ISimpleTransform::work() @ 0x000000001ecdd76e
src/Processors/Executors/ExecutionThreadContext.cpp:53: DB::executeJob(DB::ExecutingGraph::Node*, DB::ReadProgressCallback*)
src/Processors/Executors/ExecutionThreadContext.cpp:102:28: DB::ExecutionThreadContext::executeTask() @ 0x000000001ecfbb3f
src/Processors/Executors/PipelineExecutor.cpp:362:26: DB::PipelineExecutor::executeStepImpl(unsigned long, DB::IAcquiredSlot*, std::atomic<bool>*) @ 0x000000001ecee8d2
src/Processors/Executors/PipelineExecutor.cpp:290:5: DB::PipelineExecutor::executeSingleThread(unsigned long, DB::IAcquiredSlot*) @ 0x000000001eceef2c
src/Processors/Executors/PipelineExecutor.cpp:576: operator()
contrib/llvm-project/libcxx/include/__type_traits/invoke.h:90: std::__invoke_result_impl<void, DB::PipelineExecutor::spawnThreads(std::shared_ptr<DB::IAcquiredSlot>)::$_0&>::type std::__invoke[abi:sqe220101]<DB::PipelineExecutor::spawnThreads(std::shared_ptr<DB::IAcquiredSlot>)::$_0&>(DB::PipelineExecutor::spawnThreads(std::shared_ptr<DB::IAcquiredSlot>)::$_0&)
contrib/llvm-project/libcxx/include/__type_traits/invoke.h:350: void std::__invoke_void_return_wrapper<void, true>::__call[abi:sqe220101]<DB::PipelineExecutor::spawnThreads(std::shared_ptr<DB::IAcquiredSlot>)::$_0&>(DB::PipelineExecutor::spawnThreads(std::shared_ptr<DB::IAcquiredSlot>)::$_0&)
contrib/llvm-project/libcxx/include/__type_traits/invoke.h:356: void std::__invoke_r[abi:sqe220101]<void, DB::PipelineExecutor::spawnThreads(std::shared_ptr<DB::IAcquiredSlot>)::$_0&>(DB::PipelineExecutor::spawnThreads(std::shared_ptr<DB::IAcquiredSlot>)::$_0&)
contrib/llvm-project/libcxx/include/__functional/function.h:443:17: ? @ 0x000000001ecf1ece
contrib/llvm-project/libcxx/include/__functional/function.h:502: ?
contrib/llvm-project/libcxx/include/__functional/function.h:754: ?
src/Common/ThreadPool.cpp:799:12: ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::worker() @ 0x0000000014256dee
contrib/llvm-project/libcxx/include/__type_traits/invoke.h:0: std::__invoke_result_impl<void, void (ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::*&)(), ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool*&>::type std::__invoke[abi:sqe220101]<void (ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::*&)(), ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool*&>(void (ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::*&)(), ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool*&)
contrib/llvm-project/libcxx/include/tuple:1400: decltype(auto) std::__apply_tuple_impl[abi:sqe220101]<void (ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::*&)(), std::tuple<ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool*>&, 0ul>(void (ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::*&)(), std::tuple<ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool*>&, std::integer_sequence<unsigned long, 0ul>)
contrib/llvm-project/libcxx/include/tuple:1404: decltype(auto) std::apply[abi:sqe220101]<void (ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::*&)(), std::tuple<ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool*>&>(void (ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::*&)(), std::tuple<ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool*>&)
src/Common/ThreadPool.h:312: ThreadFromGlobalPoolImpl<false, true>::ThreadFromGlobalPoolImpl<void (ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::*)(), ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool*>(void (ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool::*&&)(), ThreadPoolImpl<ThreadFromGlobalPoolImpl<false, true>>::ThreadFromThreadPool*&&)::'lambda'()::operator()() @ 0x000000001425e028
contrib/llvm-project/libcxx/include/__functional/function.h:502: ?
contrib/llvm-project/libcxx/include/__functional/function.h:754: ?
src/Common/ThreadPool.cpp:809:12: ThreadPoolImpl<std::thread>::ThreadFromThreadPool::worker() @ 0x0000000014254206
contrib/llvm-project/libcxx/include/__type_traits/invoke.h:0: std::__invoke_result_impl<void, void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*>::type std::__invoke[abi:sqe220101]<void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*>(void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*&&)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*&&)
contrib/llvm-project/libcxx/include/__thread/thread.h:161: void std::__thread_execute[abi:sqe220101]<std::unique_ptr<std::__thread_struct, std::default_delete<std::__thread_struct>>, void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*, 0ul, 1ul>(std::tuple<std::unique_ptr<std::__thread_struct, std::default_delete<std::__thread_struct>>, void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*>&, std::__integer_sequence<unsigned long, 0ul, 1ul>)
contrib/llvm-project/libcxx/include/__thread/thread.h:169: void* std::__thread_proxy[abi:sqe220101]<std::tuple<std::unique_ptr<std::__thread_struct, std::default_delete<std::__thread_struct>>, void (ThreadPoolImpl<std::thread>::ThreadFromThreadPool::*)(), ThreadPoolImpl<std::thread>::ThreadFromThreadPool*>>(void*) @ 0x000000001425b6ce
__GI___clone3 @ 0x00000000001268d0
```

### Version info
- Resolved by: #105518
- Merged into: `26.6.1.1086`

groeneai:

Fix a server crash (`Trying to execute PLACEHOLDER action` logical error) when a `MATERIALIZED` CTE whose body is a correlated subquery is used as the right-hand side of `IN`. Such a CTE is now rejected at analysis time, consistent with how the same pattern is already rejected when the CTE is referenced directly in `FROM`.

### Description

Issue #105501 (STID 1611-343c). A `MATERIALIZED` CTE materializes once and cannot depend on outer-scope columns, so its body cannot be a correlated subquery. The analyzer already rejects this in the join-tree resolution path (`resolveQueryJoinTreeNode`, `Materialized CTE '...' cannot be correlated`, added with the materialized-CTE feature). The identifier-resolution path, taken when the CTE is used as the right-hand side of `IN` (`... WHERE x IN (t)`), was not guarded. With `enable_materialized_cte = 1` such a query reaches the planner with a `PLACEHOLDER`-bearing set plan and aborts the server.

Per @novikd's review, this fixes the bug at its source rather than guarding the symptom downstream: the check is added on the identifier-resolution path, mirroring the existing join-tree check (same `UNSUPPORTED_METHOD` code and message). The earlier defensive guard in `FutureSetFromSubquery::build` is removed and `PreparedSets.cpp` reverts to master; the build-path unit test is dropped with it.

New regression test `04322_105501_correlated_materialized_cte_in_subquery` exercises the IN path (which aborts on master without this change) and confirms a non-correlated `MATERIALIZED` CTE in `IN` still works.

Closes https://github.com/ClickHouse/ClickHouse/issues/105501

### Version info
- Merged into: `26.6.1.1086`


groeneai:
**Pre-PR validation gate (Phase 4 step 9, TASK.md):**

(a) **Deterministic repro?** Yes for one variant of the same root cause; not for the exact PLACEHOLDER error in CI. The AST fuzzer's failing query
```sql
WITH t AS MATERIALIZED (SELECT number, number + 65535 FROM zeros(N))
SELECT * FROM numbers(8) WHERE (number + 3, number + 1, number + 3) IN (t)
```
crashes my local debug server deterministically with `LOGICAL_ERROR: 'Column number UInt64 query tree node does not have valid source node after running QueryAnalysis pass'` (the `QueryTreePassManager` validator catches the broken state earlier than `ExpressionActions::execute` does on the failing master `sha 6d9d7738`). Both LOGICAL_ERRORs share the same root cause family — a correlated `MATERIALIZED` CTE referenced from an `IN` argument.

(b) **Root cause explained?** Yes. The inplace set-building paths `FutureSetFromSubquery::buildSetInplace` / `buildOrderedSetInplace` check `hasCorrelatedExpressions(source->getRootNode())` and bail out before installing a `CreatingSetStep`, but the pipeline-integration siblings (`addCreatingSetsStep`, `addCreatingSetsTransform`, `DelayedCreatingSetsStep::makePlansForSets`) call `FutureSetFromSubquery::build` directly without the same guard. When the source plan contains `PLACEHOLDER` nodes (correlated columns produced by `PlannerActionsVisitor::visitCorrelatedColumn`), `build` happily wraps it in a `CreatingSetStep`, the pipeline is constructed, and on the first chunk `ExpressionActions::execute` (`src/Interpreters/ExpressionActions.cpp:797`) throws `LOGICAL_ERROR: 'Trying to execute PLACEHOLDER action'`.

(c) **Fix matches root cause?** Yes — the patch moves the existing `hasCorrelatedExpressions` short-circuit from the two inplace call sites into `FutureSetFromSubquery::build` itself, so every caller is covered by one short-circuit. The inplace versions still have their own check (defensive duplicate, harmless). No assertion widening, no `no-random-*` tag, no test data shrinking.

(d) **Test intent preserved? / New tests added?** Yes — the change is purely additive (a short-circuit before adding the `CreatingSetStep`). No existing test setting was disabled and no existing test was relaxed. A new regression test `04271_105501_correlated_materialized_cte_in_subquery_placeholder.sql` covers (i) the analyzer-rejected correlated `IN` subquery shape (`NOT_IMPLEMENTED`), (ii) the analyzer-rejected correlated `MATERIALIZED` CTE in `IN` (`UNSUPPORTED_METHOD`), and (iii) sanity-check non-correlated `IN` and `MATERIALIZED CTE in IN` shapes that must still return rows. The third and fourth cases are the ones that would silently start returning `nullptr` from `build` and produce empty result sets if the guard were too aggressive — they keep returning the expected single row.

(e) **Demonstrated in both directions?** Same caveat as (a). Without the guard, the failing master `sha 6d9d7738` aborts in `ExpressionActions::execute`; on current master my local debug build aborts earlier in the `QueryTreePassManager` validator. With the guard in `FutureSetFromSubquery::build`, the pipeline-integration paths no longer accept a `PLACEHOLDER`-bearing plan, so even on the failing `sha 6d9d7738` shape the LOGICAL_ERROR cannot reach execution. Existing `04056_correlated_subquery_placeholder_action`, `03821_materialized_cte_with_sets`, and the wider correlated-subquery suite all still pass after the change.

(f) **Fix is general, not a narrow patch?** Yes — the guard now sits at the single chokepoint that all three pipeline-integration callers share. Identical sibling paths (`buildSetInplace`, `buildOrderedSetInplace`) were already guarded; this patch closes the symmetric gap.

groeneai:
cc @novikd @vdimir — could you review? This is a small defensive fix for the PLACEHOLDER LOGICAL_ERROR in #105501. It moves the existing correlated-subquery short-circuit from `buildSetInplace` / `buildOrderedSetInplace` into `FutureSetFromSubquery::build` itself so the pipeline-integration siblings (`addCreatingSetsStep`, `addCreatingSetsTransform`, `DelayedCreatingSetsStep::makePlansForSets`) are covered by the same guard. @novikd, you wrote the original guards in #98285 / #100398 — would appreciate your read on whether centralising it inside `build` is the shape you'd expect.

tests/queries/0_stateless/04271_105501_correlated_materialized_cte_in_subquery_placeholder.sql
```diff
@@ -0,0 +1,68 @@
...
+-- `IN` subqueries up front today, so the expected outcome is a clear `NOT_IMPLEMENTED`
+-- error rather than a `LOGICAL_ERROR`. Any other outcome is a regression.
+SELECT '-- correlated IN subquery';
+SELECT a FROM t1 WHERE a IN (SELECT x FROM t2 WHERE t2.y = t1.b) ORDER BY a FORMAT Null; -- { serverError NOT_IMPLEMENTED }
+
+-- Same shape but the correlated reference now lives behind a `MATERIALIZED` CTE wrapper.
+-- The CTE wraps the body in a `TableNode`, which is what the AST fuzzer query in
+-- issue #105501 ultimately reaches via `zeros(...)`. The analyzer rejects this case
+-- with the dedicated `Materialized CTE 't' cannot be correlated` check
+-- (`UNSUPPORTED_METHOD`), so we never reach the planner with a `PLACEHOLDER`-bearing
```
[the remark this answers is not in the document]
groeneai: Added a focused unit test in commit 04cf1fecb71d:
`src/Interpreters/tests/gtest_future_set_from_subquery_placeholder.cpp`.

`FutureSetFromSubquery.BuildSkipsPlanWithPlaceholder` builds a `QueryPlan` whose root is an `ExpressionStep` carrying an `ActionsDAG` with a `PLACEHOLDER` output (the exact shape `PlannerActionsVisitor::visitCorrelatedColumn` produces), wraps it in a `FutureSetFromSubquery`, and asserts `build` returns `nullptr`. That is the only assertion the new guard in `FutureSetFromSubquery::build` controls: with the guard the call returns `nullptr`; without it the call returns a non-null plan wrapping the source in a `CreatingSetStep`, and the resulting pipeline would crash with `Trying to execute PLACEHOLDER action`.

Both directions verified locally on this branch:

* With the guard present: `unit_tests_dbms --gtest_filter='FutureSetFromSubquery*'` -> `[PASSED] 2 tests` (the new test and the companion `BuildKeepsPlanWithoutPlaceholder` that confirms the guard does not over-fire on non-correlated plans).
* With the guard deleted (manually removed the `if (hasCorrelatedExpressions(plan->getRootNode())) return nullptr;` block, rebuilt `unit_tests_dbms`): `BuildSkipsPlanWithPlaceholder` fails with `Expected equality: built ... vs nullptr` while `BuildKeepsPlanWithoutPlaceholder` still passes. Guard restored after the demonstration; the only thing this commit adds is the new test file.

A unit test was chosen over an additional SQL case because the analyzer rejects correlated `IN`/`MATERIALIZED CTE` subqueries earlier today (`NOT_IMPLEMENTED` / `UNSUPPORTED_METHOD` from the analyzer, or `Column ... does not have valid source node after running QueryAnalysis pass` from the `QueryTreePassManager` validator) before the planner ever constructs a `PLACEHOLDER`-bearing `FutureSetFromSubquery`. The SQL regression test you flagged is kept as analyzer-level coverage (it locks in the current rejection paths and would change behavior if a future analyzer change let one of those shapes through); the new unit test gives the build-path guard its own direct coverage.

groeneai:
Bugfix validation was failing because the SQL test `04271_105501_correlated_materialized_cte_in_subquery_placeholder` does not fail on master HEAD: it only documents the analyzer-level rejection (`NOT_IMPLEMENTED` / `UNSUPPORTED_METHOD`), which predates this PR and never reaches `FutureSetFromSubquery::build`. So the bugfix-validation job correctly reported "Failed to reproduce the bug".

Dropped the SQL test (commit `db465e2`). The unit test `gtest_future_set_from_subquery_placeholder.cpp` added in `04cf1fec` directly exercises the guard with a constructed `PLACEHOLDER`-bearing `QueryPlan`, with both directions verified (guard present -> tests pass; guard deleted -> `BuildSkipsPlanWithPlaceholder` fails). With no changed SQL test in the PR diff, Bugfix validation now hits its "no tests to run" SKIPPED branch (treated as success).

Awaiting CI re-run on `db465e2`.

PedroTadim:
@groeneai merge with master

groeneai:
Merged `origin/master` (commit `37a40e20d46a`) into the PR branch, merge commit `b84c9bc8306f`. Clean merge: 137 master commits since the merge base touched zero files in this PR. Fix and unit test are intact. CI will re-run automatically on the new tip.

src/Interpreters/PreparedSets.cpp
```diff
@@ -312,6 +312,14 @@ std::unique_ptr<QueryPlan> FutureSetFromSubquery::build(const SizeLimits & netwo
     if (!plan)
         return nullptr;
 
+    /// Correlated subqueries contain `PLACEHOLDER` actions that cannot be executed standalone.
+    /// `buildSetInplace` / `buildOrderedSetInplace` already skip such plans, but the pipeline
+    /// integration paths (`addCreatingSetsStep`, `addCreatingSetsTransform`,
+    /// `DelayedCreatingSetsStep::makePlansForSets`) call `build` directly, so the guard must
+    /// also live here to prevent `Trying to execute PLACEHOLDER action` during execution.
+    if (hasCorrelatedExpressions(plan->getRootNode()))
```
novikd: @groeneai `in` function is supposed to be validated in the `QueryAnalyzer`. It should not allow the second argument to be a correlated subquery.

novikd (commented):
@groeneai, please add a repro SQL test

groeneai (commented):
Added SQL test `04274_105501_correlated_materialized_cte_in_subquery_placeholder.sh` in commit `07412039b50c`. The test runs four patterns and asserts that none of them ever surfaces `Trying to execute PLACEHOLDER action`:

1. the original fuzzer query (3-tuple LHS, 2-column `MATERIALIZED` CTE over `zeros()`, whose inner `number` accidentally correlates against the outer `numbers(8)`)
2. the simplified single-column variant over `primes()` that the fuzzer also hit
3. the same simplified variant under `enable_join_runtime_filters = 0`, `query_plan_merge_filter_into_join_condition = 0`, `query_plan_merge_filters = 0`, `query_plan_convert_any_join_to_semi_or_anti_join = 1`, the optimiser-pessimising settings the fuzzer combined with the crash query
4. an explicit correlated subquery in `IN`

One caveat: current master rejects all four at analyzer level (`NOT_IMPLEMENTED` from `resolveFunction.cpp`'s `isCorrelatedQueryOrUnionNode` check, or `UNSUPPORTED_METHOD` when the outer-scope resolution itself is disallowed). So the queries no longer reach `FutureSetFromSubquery::build` from SQL today and the test passes on plain master too. That makes the test forward-looking regression coverage (and lock-in of the current analyzer rejection): if either guard regresses, the LOGICAL_ERROR returns and the test catches it. The actual both-directions check for the build guard stays in the unit test `gtest_future_set_from_subquery_placeholder.cpp`, which I confirmed FAILs when the new short-circuit is removed and PASSes with it in place.

Because the SQL queries no longer crash on master HEAD, Bugfix validation will likely report "Failed to reproduce the bug" again on this PR. Happy to either re-categorise to `CI Fix or Improvement (changelog entry is not required)` or keep `Bug Fix` and request a maintainer override on Bugfix validation, whichever you prefer.

groeneai:
@clickhouse-gh both bot concerns addressed in `ad69a66388d` (groeneai/fix-105501-placeholder-action-in-creating-sets-build).

**1. `src/Interpreters/PreparedSets.cpp` — guard now precedes `std::move(source)`**

Moved the `hasCorrelatedExpressions(source->getRootNode())` check ahead of `auto plan = std::move(source)`, so a short-circuited `build` leaves the stored subquery plan intact. This matches the `buildSetInplace` / `buildOrderedSetInplace` contract and avoids the downstream `Cannot serialize FutureSetFromSubquery with no query plan` from `SetsSerialization.cpp:124` if the same `ColumnSet` is later serialized via `getQueryPlan`.

**2. `tests/queries/0_stateless/04274_*.sh` — whitelist tightened**

`assert_no_placeholder` now treats `RC=0` as success and only accepts a non-zero exit when the output contains one of the expected analyzer/fuzzer errors: `NOT_IMPLEMENTED`, `UNSUPPORTED_METHOD`, `NUMBER_OF_COLUMNS_DOESNT_MATCH`, `UNKNOWN_IDENTIFIER`, or the `valid source node` `LOGICAL_ERROR`. Any other error (setup failure, unknown setting, `Not-ready Set is passed`, etc.) now fails the case. Verified locally with a sentinel: `SET this_setting_does_not_exist = 1` is correctly rejected as an unexpected error.

### Pre-PR validation gate

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes for the unit test path. `unit_tests_dbms --gtest_filter='FutureSetFromSubquery*'` runs in <5 ms and is fully deterministic (constructed `QueryPlan` with `PLACEHOLDER` output). The original SQL fuzzer path is rejected at analyzer level on master HEAD, so SQL-level deterministic repro is not currently possible — covered by the unit test instead. |
| b | **Root cause explained?** | Yes. The previous form (`auto plan = std::move(source); if (hasCorrelatedExpressions(plan->getRootNode())) return nullptr;`) consumed `source` before deciding to short-circuit. On a short-circuit the `FutureSetFromSubquery::source` member becomes null while `set_and_key->set->isCreated()` is still false, so a later `getQueryPlan` / `SetsSerialization` call would observe a subquery set with no plan and throw `Cannot serialize FutureSetFromSubquery with no query plan`. Moving the guard before the move keeps the contract symmetric with `buildSetInplace` / `buildOrderedSetInplace`. |
| c | **Fix matches root cause?** | Yes. Reordering the two statements so the correlation check runs against `source.get()` before consuming it is exactly the contract `buildSetInplace` / `buildOrderedSetInplace` (lines 403, 456 of `PreparedSets.cpp`) already use via `if (source && hasCorrelatedExpressions(source->getRootNode())) return;`. |
| d | **Test intent preserved / new tests added?** | Yes. The unit test `gtest_future_set_from_subquery_placeholder` is unchanged — both `BuildSkipsPlanWithPlaceholder` and `BuildKeepsPlanWithoutPlaceholder` continue to exercise the targeted-guard property in both directions. The SQL test's invariant (no `Trying to execute PLACEHOLDER action`) is preserved while the failure surface is narrowed: silent passes from setup errors, unknown settings, or `Not-ready Set is passed` are now caught. |
| e | **Both directions demonstrated?** | Yes. With the new ordering: `unit_tests_dbms --gtest_filter='FutureSetFromSubquery*'` -> 2/2 pass. With the guard removed (verified by patching out the `if (hasCorrelatedExpressions(source->getRootNode())) return nullptr;` line and rebuilding `unit_tests_dbms`): `BuildSkipsPlanWithPlaceholder` FAILs (`Expected equality of these values: built ... vs nullptr`), `BuildKeepsPlanWithoutPlaceholder` still passes. Restored after verification. |
| f | **Fix is general, not a narrow patch?** | Yes. The bot's concern about the lost subquery plan was a fix-correctness issue, not a missed code path. After this update, all three correlated-expression guards in `PreparedSets.cpp` (`build`, `buildSetInplace`, `buildOrderedSetInplace`) check `source` before consuming it. Greppable invariant: `if (source* && hasCorrelatedExpressions(source*->getRootNode()))` is now uniform across the file. The SQL test whitelist also rejects all non-listed error classes uniformly across all four cases. |

Session id: cron:clickhouse-ci-task-worker:20260604-073000

groeneai:
CI fully finished on ad69a66. Triage:

CI Ledger — covered SHA: ad69a66388d70f5f26fe742465eea51a50911664
- Bugfix validation (functional tests) / 04274_105501_correlated_materialized_cte_in_subquery_placeholder — Bugfix-validation framework quirk — expected.

The test asserts the analyzer-side rejections (NOT_IMPLEMENTED / UNSUPPORTED_METHOD) for correlated EXISTS/IN-CTE shapes. Those rejections fire identically on master HEAD and on this PR, so the test passes in both runs. The PLACEHOLDER LOGICAL_ERROR fix lives deeper, in `FutureSetFromSubquery::build`, and is exercised only by the AST-fuzzer-mutated query shapes that slip past the analyzer's early `isCorrelatedQueryOrUnionNode` check. A deterministic SQL repro of the deep planner crash from a stable test isn't available; the AST fuzzer is the only known reproducer. Bugfix validation's invert-logic therefore reports "the regression test doesn't fail on master" and flags FAIL.

All other CI is green: every Build, Stateless tests, Integration tests, Stress, AST fuzzer, BuzzHouse, Performance, Unit tests, SQLLogic and Style passed. PR/Mergeable Check are red only because of this single framework-quirk row.

Session: cron:our-pr-ci-monitor:20260604-193000

src/Interpreters/PreparedSets.cpp
```diff
@@ -372,6 +372,14 @@ std::unique_ptr<QueryPlan> FutureSetFromSubquery::build(const SizeLimits & netwo
...
```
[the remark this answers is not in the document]
groeneai: Pinned the contract in commit 582698e2143d. Added an `EXPECT_NE(future_set->getQueryPlan(), nullptr)` after the existing `EXPECT_EQ(built, nullptr)` in `BuildSkipsPlanWithPlaceholder`. Both directions verified locally:

- guard ordered before `std::move(source)` (current HEAD): both unit tests PASS.
- guard ordered after `std::move(source)` (regression shape): the new assertion FAILS as expected (`getQueryPlan()` returns `nullptr` because `source` was consumed).

Session id: cron:clickhouse-ci-task-worker:20260605-011400

groeneai:
### Pre-PR validation gate

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. `unit_tests_dbms --gtest_filter='FutureSetFromSubquery*'` is fully deterministic; the `BuildSkipsPlanWithPlaceholder` test directly constructs the `PLACEHOLDER`-bearing plan and exercises `FutureSetFromSubquery::build`. |
| b | **Root cause explained?** | The bot's concern: the previous shape of `BuildSkipsPlanWithPlaceholder` only asserted `build` returned `nullptr`. That assertion still holds if a future regression reorders the guard to land after `auto plan = std::move(source)` — `build` would still return `nullptr` (the moved-from path), but `getQueryPlan` would later see a subquery set with no plan and `SetsSerialization.cpp:124` would raise `Cannot serialize FutureSetFromSubquery with no query plan`. The contract that "short-circuit MUST NOT consume `source`" was unpinned. |
| c | **Fix matches root cause?** | Yes. Single targeted assertion: `EXPECT_NE(future_set->getQueryPlan(), nullptr)` directly after the existing `EXPECT_EQ(built, nullptr)`. It pins the very contract the source-side fix in `ad69a66388d` established, so any future reordering is caught at compile-and-test time. |
| d | **Test intent preserved / new tests added?** | N/A as a code fix (this commit only adds an assertion to an existing unit test). The unit test still verifies: (i) `build` returns `nullptr` on a `PLACEHOLDER`-bearing plan, (ii) the source plan is preserved through the skip — strictly stronger than before. The companion `BuildKeepsPlanWithoutPlaceholder` is unchanged. |
| e | **Both directions demonstrated?** | Yes, against `unit_tests_dbms` built locally with ASan: <br>(i) guard ordered BEFORE `std::move(source)` (current HEAD): `[PASSED] 2 tests`. <br>(ii) guard ordered AFTER `std::move(source)` (regression shape): the new assertion FAILs with `Expected: (future_set->getQueryPlan()) != (nullptr), actual: NULL vs (nullptr)`; companion `BuildKeepsPlanWithoutPlaceholder` still passes. Confirms the assertion catches exactly the reordering the bot warned about. |
| f | **Fix is general, not a narrow patch?** | N/A — the source-side fix in `ad69a66388d` was already general (single short-circuit covering all three pipeline-integration callers). This commit is purely test hardening to lock the contract in. |

### Bugfix validation note

Bugfix validation has been failing on this PR because the SQL test `04274_*` passes on master HEAD too (the analyzer rejects the `MATERIALIZED` CTE / correlated `IN` patterns at `resolveFunction.cpp:997` before reaching `FutureSetFromSubquery::build`, so a SQL test cannot fail on master without a code regression elsewhere). The unit test is what carries the regression-coverage weight; the SQL test is forward-looking lock-in coverage that @ novikd asked us to add. This is a real bug fix (not a CI-only change), so re-categorising to `CI Fix or Improvement` would be wrong.

@ novikd / @ vdimir — could you either request a maintainer override on `Bugfix validation (functional tests)` for this PR, or approve dropping the SQL test in favour of the unit test alone? Both options were offered upthread; happy with either.

Session id: cron:clickhouse-ci-task-worker:20260605-011400

groeneai:
**CI ledger — covered SHA: `582698e2143d`**

CI fully finished on this HEAD (Finish Workflow ✅, 162/166 pass).

| check | test | classification | disposition |
|---|---|---|---|
| Bugfix validation (functional tests) | `04274_105501_correlated_materialized_cte_in_subquery_placeholder` | known framework quirk | The SQL test is forward-looking regression coverage; the analyzer rejects the queries on master HEAD too, so "Bugfix validation" reports "Failed to reproduce the bug". The deep PLACEHOLDER planner crash that originally surfaced via AST-fuzzer mutations cannot be deterministically reproduced from SQL on current master. Both-directions verification is in the unit test (`gtest_future_set_from_subquery_placeholder.cpp`). Already flagged upthread; awaiting reviewer decision (recategorize PR or maintainer override). |
| Unit tests (msan, function_prop_fuzzer) | `FunctionsStress` — `Logical error: 'key0->size() > i'` | chronic trunk flake, unrelated | Hit on 27+ unrelated PRs (different heads: `fix/bad-cast`, `dictionary-layout-documentation`, `fix-postgresql-cancel-exception-master`, `lazy-background-schedule-pool`, etc.) + 1 master hit on 2026-06-02 in the last 30 days. Tracked in our internal flaky family. This PR (`PreparedSets.cpp` PLACEHOLDER guard + unit test) does not touch Functions or function_prop_fuzzer. |

All other CI green: 11 builds, all sanitizer Stress / AST fuzzer / BuzzHouse, all Stateless / Integration / Performance / Unit (asan/tsan), Style, Fast test, Compatibility, Install, Docker, SQLLogic.

cc @ novikd @ vdimir — both failures are non-blocking; awaiting your re-review on the substantive fix.

groeneai:
## CI Ledger — covered SHA: `582698e2143dc2f506abf3571639261a69667911`

CI fully finished on this HEAD. Two non-green rows; both **not PR-caused**:

| # | Failure | Disposition |
|---|---|---|
| 1 | `Bugfix validation (functional tests)` / `04274_105501_correlated_materialized_cte_in_subquery_placeholder` | Bugfix-validation framework quirk — same as on prior SHA `ad69a66`. The added regression test rejects the correlated-CTE `IN` shape via the analyzer (`NOT_IMPLEMENTED`/`UNSUPPORTED_METHOD`) on **both** master and PR HEAD — so the framework's "must FAIL on master, PASS on PR" gate cannot be satisfied. The deep `PLACEHOLDER` planner crash only triggers via AST-fuzzer mutation chains, not from the captured query alone. The fix is still real (it's the inner `FutureSetFromSubquery::build` guard), and the test still exercises the surrounding rejection paths. |
| 2 | `Unit tests (msan, function_prop_fuzzer)` / `FunctionsStress` — `Logical error: 'key0->size() > i'` | Chronic unrelated trunk bug. 30-day cross-PR distribution: 27 unrelated PRs hit it (master has 14 hits, including 2026-06-04 12:10 UTC). Tracked separately in our internal task; sighting recorded. Unrelated to PreparedSets / PLACEHOLDER. |

All other 162 checks pass: every Build (arm/amd matrix), Stateless tests (incl. parallel/sequential, s3/azure/replicated variants, distributed-plan, parallel-replicas, WasmEdge), Integration tests, Stress (all sanitizers), AST fuzzer (4 variants), BuzzHouse (4 variants), Performance Comparison, Unit tests (asan/tsan/msan + function_prop_fuzzer non-msan slots), SQLLogic, Compatibility, Install, Docker, Code Review, Fast test, Style.

Awaiting re-review from @novikd / @vdimir.

src/Interpreters/PreparedSets.cpp
```diff
@@ -367,11 +367,22 @@ std::unique_ptr<QueryPlan> FutureSetFromSubquery::build(const SizeLimits & netwo
...
-    if (!plan)
+    /// Correlated subqueries contain `PLACEHOLDER` actions that cannot be executed standalone.
+    /// `buildSetInplace` / `buildOrderedSetInplace` already skip such plans, but the pipeline
+    /// integration paths (`addCreatingSetsStep`, `addCreatingSetsTransform`,
+    /// `DelayedCreatingSetsStep::makePlansForSets`) call `build` directly, so the guard must
+    /// also live here to prevent `Trying to execute PLACEHOLDER action` during execution.
+    /// Check before consuming `source` so a short-circuit leaves the stored subquery plan
+    /// intact, matching the `buildSetInplace` / `buildOrderedSetInplace` contract (otherwise
+    /// `getQueryPlan` would later see a subquery set with no plan).
+    if (hasCorrelatedExpressions(source->getRootNode()))
```
novikd: @groeneai, It is not correct. Materialized CTE should not be correlated with subqueries. Just add a check in the `QueryAnalyzer` if the Materialized CTE body is a correlated subquery.

groeneai: Done, pivoted to the analyzer as you asked. Master already rejects a correlated `MATERIALIZED` CTE in the join-tree path (`resolveQueryJoinTreeNode`, `Materialized CTE '...' cannot be correlated`). The gap was the identifier-resolution path used when the CTE is the RHS of `IN` (`... WHERE x IN (t)`): with `enable_materialized_cte = 1` it was unguarded and the server aborts.

This change adds the same check on that path, mirroring the join-tree one (same `UNSUPPORTED_METHOD` code and message). The defensive `FutureSetFromSubquery::build` guard is removed and `PreparedSets.cpp` reverts to master; the build-path unit test is dropped. New SQL test `04322_105501_correlated_materialized_cte_in_subquery` covers the IN path (aborts on master without this change) and confirms a non-correlated `MATERIALIZED` CTE in `IN` still works.

Pushed as `55127470e8ab`.

groeneai:
### Pre-PR validation gate

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. On master `fc73f46fae6`: `clickhouse local -q "SET enable_analyzer=1; SET allow_experimental_correlated_subqueries=1; SET enable_materialized_cte=1; WITH t AS MATERIALIZED (SELECT number FROM primes(5)) SELECT * FROM numbers(8) WHERE number IN (t);"` aborts the server (SIGABRT, RC=134) every time. |
| b | **Root cause explained?** | A `MATERIALIZED` CTE materializes once and cannot depend on outer-scope columns. When its body references a column absent from its own `FROM` (e.g. `number` over `primes(5)` whose column is `prime`), the name resolves against the outer query, making the CTE correlated. The analyzer rejects this in the join-tree path (`resolveQueryJoinTreeNode`), but the identifier-resolution path taken for `x IN (t)` was unguarded, so a `PLACEHOLDER`-bearing set plan reached the planner and aborted with `Trying to execute PLACEHOLDER action`. |
| c | **Fix matches root cause?** | Yes. The check is added at the source (analyzer identifier-resolution of a materialized CTE body), rejecting the malformed query before any plan is built, mirroring the existing join-tree check. It is not a downstream guard on the symptom. |
| d | **Test intent preserved / new tests added?** | New SQL test `04322_105501_correlated_materialized_cte_in_subquery` asserts the IN-path rejection (`serverError UNSUPPORTED_METHOD`) for the original fuzzer shapes and confirms a non-correlated `MATERIALIZED` CTE in `IN` still returns rows. The old build-path unit test and `.sh` test are removed since the `FutureSetFromSubquery::build` guard they exercised is gone. |
| e | **Both directions demonstrated?** | Yes. Without the change (master): the IN cases SIGABRT. With the change: they fail cleanly with `UNSUPPORTED_METHOD: Materialized CTE 't' cannot be correlated`, the non-correlated case returns `0,1,2,3,4`, and the test output matches the reference. `serverError` enforcement confirmed (a deliberately wrong expected code reports `Expected server error code: 48 but got: 1`). |
| f | **Fix is general, not a narrow patch?** | Yes. The two materialized-CTE resolution entry points (join-tree and identifier-resolution) now both enforce the same invariant with the same error. Covers single-column, tuple, and UNION-body CTEs, double IN use, and the optimizer-pessimising settings the fuzzer combined with the crash. `PreparedSets.cpp` is reverted to master, so no symptom-level guard remains. |

Session id: cron:clickhouse-worker-slot-17:20260610-184200

groeneai:
CI fully finished (Finish Workflow passed). Failures, both pre-existing, none PR-caused (this PR is analyzer-only -- QueryAnalyzer.cpp rejecting correlated MATERIALIZED CTE in IN):
- Stateless tests (arm_asan_ubsan, azure, sequential) -- azure-infra-outage / db-disk "Server died" chronic class.
- Stress test (arm_asan_ubsan, s3) -- generic stress instability (Received signal 6 during compatibility-mode run); an analyzer CTE-rejection change cannot cause a storage stress SIGABRT.
No Fast/Bugfix/Style failure.

groeneai:
Additional provenance: this change also fixes a second AST-fuzzer symptom of the same root cause, **STID 2508-37d4** (`Logical error: 'table_holder.has_value()'`, `AST fuzzer (amd_debug)`).

Same root cause (a correlated `MATERIALIZED` CTE used as the RHS of `IN`, unguarded on the identifier-resolution path), different crash site. The fuzzer query reaches `inlineMaterializedCTEIfNeeded` (`src/Analyzer/inlineMaterializedCTEIfNeeded.cpp:64`) with a `reused_materialized_cte` entry whose `table_holder` was never finalized, so `MaterializedCTE::extractTableHolder` (`src/Interpreters/MaterializedCTE.h:67`) aborts.

Verified both directions on master HEAD `fc73f46fae6` (debug) with the auto-reduced fuzzer query from STID 2508-37d4:
- pristine master: `SIGABRT` at `MaterializedCTE.h:67` via `inlineMaterializedCTEIfNeeded` -> `QueryAnalyzer::resolve`.
- with this PR's correlation guard on the IN path: no abort (the query then fails only on the fuzzer's `remote('127.0.0.{1,2}', ...)` noise, not in the analyzer).

CIDB: STID 2508-37d4 has 2 hits in 90d (PR #100573 on 2026-03-24, PR #82414 on 2026-06-09), 0 on master, no separate open issue. No separate fix PR needed; folding it under this one.

groeneai:
Additional provenance: this also fixes a third AST-fuzzer symptom of the same root cause, **STID `0250-20a8`** / issue #80564 (`Logical error: 'Column ... query tree node does not have valid source node after running QueryAnalysis pass'`, aborting in `QueryTreePassManager::ValidationChecker::visitColumn`).

Same root cause (correlated `MATERIALIZED` CTE as the RHS of `IN`, unguarded on the identifier-resolution path), different crash site: with the optimizer settings the fuzzer combined here, resolution proceeds and a correlated `ColumnNode` survives `QueryAnalysis` with a null source, so the post-pass validator aborts instead of the planner hitting `PLACEHOLDER`.

Verified both directions on master `4fc87f09390` (debug) with the reduced query from STID `0250-20a8`:
```sql
WITH t AS MATERIALIZED (SELECT number FROM zeros_mt(1))
SELECT * FROM numbers(1) WHERE number IN (t) SETTINGS enable_materialized_cte = 1;
```
- pristine master: `SIGABRT` (`valid source node`) in `ValidationChecker::visitColumn`.
- with this PR's correlation guard: clean `UNSUPPORTED_METHOD` ("Materialized CTE 't' cannot be correlated").

CIDB: STID `0250-20a8` valid-source-node signature has 0 master hits and 27 hits across 15 unrelated PR branches over 90d (it surfaces on whichever PR the fuzzer happens to run on); first seen 2026-03-19. #80564 is OPEN with the `fuzz` label. No separate fix PR needed; folding it under this one.

Session: cron:clickhouse-worker-slot-2:20260612-230800

alexey-milovidov (approved):
Very good fix, LGTM!

alexey-milovidov:
@groeneai, investigate and fix the abort here: [Stateless tests (arm_asan_ubsan, azure, sequential)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105518&sha=55127470e8ab8085a1c4cef2c4451530daba7818&name_0=PR&name_1=Stateless%20tests%20%28arm_asan_ubsan%2C%20azure%2C%20sequential%29)

groeneai:
@alexey-milovidov Investigated. The abort is not from this PR's code. It is heap-corruption fallout from #106020 (Multi-stage distributed queries), which was reverted on master by #107122 (merged 2026-06-11 01:37 UTC, "fixes ASan heap corruption on master"). This branch's base predated that revert.

Evidence:
* git ancestry: HEAD 55127470 contained the #106020 merge and was missing the #107122 revert.
* timing: the crash ran 2026-06-10 20:46 UTC, inside the #106020-merge (06-09 16:06) to #107122-revert (06-11 01:37) window.
* this PR's diff is analyzer-only (QueryAnalyzer.cpp +9 lines plus one stateless test). The `std::terminate` is in Boost-fiber teardown (`AsyncTaskExecutor::destroyFiber`) on the unrelated `distributedIndexAnalysis` / `HedgedConnectionsFactory` async path. That code predates #106020 and is byte-identical to master, so it is an innocent free-time crash victim of the global heap corruption, not a logic bug.
* CIDB: this signature (`std::terminate` in `fiber_capture_record::destroy` via `destroyFiber`) is 0 master / 1 total hit in 90 days, only this run. Report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105518&sha=55127470e8ab8085a1c4cef2c4451530daba7818&name_0=PR&name_1=Stateless%20tests%20%28arm_asan_ubsan%2C%20azure%2C%20sequential%29

Fix: merged current master into the branch (now bc4d6b53c60), which picks up revert #107122, so the abort cannot recur. CI is re-running.

groeneai:
CI finished green on HEAD `bc4d6b53c60` (the merge-master that picked up revert #107122). 232733 tests OK, 0 FAIL/ERROR; Finish Workflow and Mergeable Check both SUCCESS. The only red status is `CH Inc sync`, an internal ClickHouse-Inc sync status unrelated to CI correctness. No failures to address.

clickgapai:

### TL;DR
**Reject correlated MATERIALIZED CTE used in IN at analysis time**

**ClickGap verified:** reproduced it with the script below on `26.5`; tested 4 version(s) — 3 affected, 1 clean (see Affects); bisected to a candidate range — a single introducing PR could not be pinned (see Bisect).

### Reproduction
```
-- enable_materialized_cte lets a MATERIALIZED CTE whose body is a correlated
-- subquery reach the planner; using it as the RHS of IN crashed the server.
-- primes(5) exposes column `prime` (not `number`), so the inner `number`
-- binds to the outer numbers(8) -> the CTE body is correlated.
SET enable_analyzer = 1;
SET allow_experimental_correlated_subqueries = 1;
SET enable_materialized_cte = 1;
WITH t AS MATERIALIZED (SELECT number FROM primes(5))
SELECT * FROM numbers(8) WHERE number IN (t);
-- Pre-fix: server LOGICAL_ERROR (Code 49 -- "Trying to execute PLACEHOLDER
--   action" / "... query tree node does not have valid source node").
-- Fixed:   rejected at analysis time with UNSUPPORTED_METHOD
--   ("Materialized CTE 't' cannot be correlated").
```

### Bisect
- The bisect boundary was rejected — PR #99717 only adds a test for text index PREWHERE with aliases in SELECT, which is unrelated to MATERIALIZED CTE validation in IN clauses. — a coincidental boundary, not the cause; no introducing PR pinned.
- The source-build bisect converged to #99717, but it is **not** the introducer (see above) — not attributing it.

### Affects
- **Reproduces on**: `26.5, 26.4, 26.3`
- **Could not test** (binary unavailable): `26.2, 26.1`
- **Backport needed**: 26.5, 26.4, 26.3

### Suggested next step
The bisect boundary above narrows the search in `comp-query-analyzer` — treat the candidate range as the starting point; the single introducing PR is unconfirmed.

**Severity:** `P2`

cc fix PR author: @groeneai
cc fix PR reviewer: @novikd 

---
_ClickGap · Finding: `phase_d_pr105518`_




### Triage metadata
- Primary component: `comp-query-analyzer`
- Component owners (`comp-query-analyzer`): @KochetovNicolai @novikd

clickgapai:
Hi — this looks like a backport candidate for `26.3`, `26.4`, `26.5`. I ran this PR's own regression test against each branch's binary and the bug **reproduces** there (the fix isn't present yet).

PR #105528, opened 2026-05-21
Fix stdout/stdin compression not applied when format is explicitly specified

### LINKED ISSUE #104441: STDOUT compression not applied when output format is explicitly specified

_Retrospective finding from a historical scan of [PR #63662](https://github.com/ClickHouse/ClickHouse/pull/63662) (merged 2024-05-22). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

When using `--output-format`, `--format`, or `--vertical` with stdout redirected to a compressed file (e.g., `> output.gz`), the output is not compressed. Only works when format is auto-detected from file extension.

**Root cause:** ClientBase.cpp:917-930: The `if (getClientConfiguration().has("output-format"))` and sibling branches exit early without checking for compression. The `default_output_compression_method = chooseCompressionMethod()` call at line 942 is only reachable when no explicit format is specified.

**Why we believe this is a bug:** `setDefaultFormatsAndCompressionFromConfiguration()` at ClientBase.cpp:915-943 uses an if-else chain where explicit format options (lines 917-930) take precedence over file-based detection (lines 932-943). Compression detection only happens in the latter branch.

**Affected locations:**
- `src/Client/ClientBase.cpp:917` — output-format branch exits without compression detection
- `src/Client/ClientBase.cpp:922` — format branch exits without compression detection
- `src/Client/ClientBase.cpp:927` — vertical branch exits without compression detection

**Impact:** Users who specify output format with `--output-format TSV --query '...' > data.gz` get plain text instead of gzip-compressed data. Breaks expected workflow of combining explicit format with auto-compression.

Yes — confirmed on current `master` (commit `8eaaf96ca45`).

### How to reproduce

```sql
/home/ec2-user/ClickHouse/build/programs/clickhouse client --host 127.0.0.1 --port 19030 --output-format TSV --query "SELECT 'test'" > /tmp/test.gz && file /tmp/test.gz
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/242c5723-075c-435c-853b-a0c4cce47557)

### Expected behavior

```
TEST 1 PASSED: explicit format with .gz is compressed
TEST 2 PASSED: --vertical with .gz is compressed
TEST 3 PASSED: local explicit format with .zst is compressed
```

```
TEST 1 FAILED: explicit format with .gz is NOT compressed
TEST 2 FAILED: --vertical with .gz is NOT compressed
TEST 3 FAILED: local explicit format with .zst is NOT compressed
```

### Additional context

**Open risks:**
- Same bug pattern exists for input compression (lines 953-981) — explicit input format skips decompression detection

**Suggested fix:** Move compression detection out of the if-else chain. After determining format (either explicit or auto-detected), always check if stdout is redirected to a file and set compression method based on extension.

Found during automated review of [PR #63662](https://github.com/ClickHouse/ClickHouse/pull/63662).

---
_ClickGapAI · Confidence: HIGH · Severity: P2 · Finding: `h_pr63662_001`_

### Version info
- Resolved by: #105528
- Merged into: `26.6.1.596`

Avogar:

Fix stdout/stdin compression not applied when format is explicitly specified. Closes https://github.com/ClickHouse/ClickHouse/issues/104441

### Version info
- Merged into: `26.6.1.596`


CheSema (approved):

clickgapai:
Hi @Avogar @CheSema — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: Body is template-only (changelog category + label) with no prose explaining the root cause (compression detection was nested inside the implicit-format branches) or the fix approach (hoist detection out, gate on method==None).
- Suggested addition:

  > Previously `setDefaultFormatsAndCompressionFromConfiguration` detected stdout/stdin compression from the file-descriptor name only inside the branches that auto-detect the format. When a user passed an explicit `--output-format`/`--input-format`/`--format`/`--vertical`, those branches short-circuited and compression was never applied, so e.g. `clickhouse-client --output-format TSV --query ... > out.gz` wrote uncompressed data. This PR moves compression detection into standalone blocks (guarded by `default_*_compression_method == None` to respect an explicit `--compression`), so redirected `.gz`/`.zst` files are compressed/decompressed regardless of the explicit format.

PR #105540, opened 2026-05-21
Iceberg 'time' type support

ianton-ru:

Iceberg 'time' type support

Backward Incompatible Change: this PR changes time format,.
Before - seconds from midnight:
```
SELECT * FROM datalake.`namespace.table`

43200
```
Now - time with microseconds
```
SELECT * FROM datalake.`namespace.table`

12:00:00.000000
```


src/Processors/Formats/Impl/Parquet/Write.cpp
```diff
@@ -1210,7 +1234,12 @@ void writeColumnChunkBody(
                 N(Int16, Int32Type);
             break;
         }
-        case TypeIndex::Int32  : N(Int32,  Int32Type); break;
+        case TypeIndex::Int32:
+            if (s.type->getTypeId() == TypeIndex::Time)
+                writeColumnImpl<parquet::Int64Type>(s, options, out, ConverterTime(s.primitive_column, s.datetime_multiplier));
```
scanhex12: Why do we process Int64 in `case TypeIndex::Int32`?

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
```diff
@@ -173,6 +172,9 @@ static void insertNumber(IColumn & column, WhichDataType type, T value)
         case TypeIndex::DateTime64:
             convertAndInsert<T, Int64>(assert_cast<ColumnDecimal<DateTime64> &>(column), value, type.idx);
             break;
+        case TypeIndex::Time64:
```
scanhex12, line 187: What about TypeIndex::Time?

tests/integration/test_storage_iceberg_no_spark/test_write_time.py
```diff
@@ -0,0 +1,41 @@
...
+    node.query(
+        f"CREATE TABLE `{TABLE_NAME}` (key {time_type}, value {time_type}, comment String) ENGINE = IcebergLocal('/var/lib/clickhouse/user_files/iceberg_data/default/{TABLE_NAME}', 'Parquet') PARTITION BY key",
+        settings={"allow_experimental_insert_into_iceberg": 1, 'write_full_path_in_iceberg_metadata': 1}
+        )
+    node.query(
+        f"INSERT INTO `{TABLE_NAME}` VALUES ('12:00:00', '12:10:00', 'test1'), ('13:00:00', '13:10:00', 'test2'), ('14:00:00', '14:10:00', 'test3');",
+        settings={"allow_experimental_insert_into_iceberg": 1, 'write_full_path_in_iceberg_metadata': 1}
+        )
+
+    assert node.query(f"SELECT * FROM `{TABLE_NAME}` ORDER BY key") == "12:00:00.000000\t12:10:00.000000\ttest1\n13:00:00.000000\t13:10:00.000000\ttest2\n14:00:00.000000\t14:10:00.000000\ttest3\n"
```
scanhex12, line 32: Please validate that spark can read your table correctly

tests/integration/test_storage_iceberg_no_spark/test_write_time.py
```diff
@@ -0,0 +1,41 @@
...
+        "Time64(0)",
+        "Time64(3)",
+        "Time64(6)",
+        "Nullable(Time)",
+        "Nullable(Time64(0))",
+        "Nullable(Time64(3))",
+        "Nullable(Time64(6))",
+    ],
+)
+def test_write_time(started_cluster_iceberg_no_spark, time_type):
```
scanhex12, line 18: what about `spark write -> CH read` test?

ianton-ru: [Spark does not support time fields](https://github.com/apache/iceberg/blob/main/spark/v4.1/spark/src/main/java/org/apache/iceberg/spark/TypeToSparkType.java#L131).

tests/integration/test_storage_iceberg_no_spark/test_write_time.py
ianton-ru: [Spark does not support time fields](https://github.com/apache/iceberg/blob/main/spark/v4.1/spark/src/main/java/org/apache/iceberg/spark/TypeToSparkType.java#L131).

src/Processors/Formats/Impl/Parquet/Write.cpp
ianton-ru: Apache Avro has only [two options](https://avro.apache.org/docs/1.8.0/spec.html#Time+%28millisecond+precision%29) to save `time` value - `time-millis` in milliseconds, using 32-bit `int` parquet type, and `time-micros` in microseconds, using 64-bit `long`.

Split on two cases, with milliseconds now int32, with microseconds - int64

ianton-ru: Reverted.
Iceberg 'time' [must be with microsecond precision](https://iceberg.apache.org/spec/#primitive-types), so writing only as Int64.

src/Processors/Formats/Impl/AvroRowInputFormat.cpp
ianton-ru: Added for any case, when in Avro file 'time' in milliseconds, but when Avro describes Iceberg schema, time can be in microseconds only.

tests/integration/test_storage_iceberg_no_spark/test_write_time.py
scanhex12: What about trino? We recently added support for it in out integration tests

ianton-ru: Added.

ianton-ru:
arm_tidy build failed on AI comment
```
[src/Processors/Formats/Impl/Parquet/PrepareForWrite.cpp:492]` The writer now advertises every `Time`/`Time64` column as Parquet 
                      logical `TIME`, but the write converters still serialize valid ClickHouse values such as negative `Time64` or `toTime('100:00:00')` 
                      without checking the Parquet/Iceberg time-of-day domain `[00:00:00, 24:00:00)`. That can produce Parquet/Iceberg files external readers 
                      reject or misread. Validate the scaled value in the `Time` and `Time64` converters before writing `TIME_MICROS`/`TIME_NANOS`, including 
                      the `datetime_multiplier == 1` fast path.
```
But is is not true, `time` out of Iceberg range caused exception, I added test with check in https://github.com/ClickHouse/ClickHouse/pull/105540/changes/f397c66a10ff16fa3435323608df016b8b0a49e4

src/Functions/icebergBucketTransform.cpp
```diff
@@ -162,6 +165,53 @@ class FunctionIcebergHash final : public IFunction
                 result_data[i] = hashLong(value_int);
             }
         }
+        else if (which.isTime())
+        {
+            /// Iceberg time: hashLong(microsecsFromMidnight(v)).
+            for (size_t i = 0; i < input_rows_count; ++i)
+            {
+                const Int64 seconds = column->getInt(i);
```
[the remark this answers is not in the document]
ianton-ru: (AI vs AI battle begins)

Thanks for the review. The observation is correct that `icebergHash` / `icebergBucket` accept ClickHouse `Time` / `Time64` values outside Iceberg’s `[00:00:00, 24:00:00)` domain (for example `toTime('100:00:00')`) and still hash them after scaling to microseconds. We intentionally do **not** treat that as a bug that needs `validateIcebergTimeOfDayMicroseconds` in these functions.
**Why this matches Iceberg’s hash contract**
- Appendix B defines the transform as `hashLong(microsecsFromMidnight(v))` for a valid Iceberg `time` value. It does not require the hash implementation to re-validate the time-of-day bound.
- Upstream Java does the same: Iceberg `time` bucketing uses `BucketLong`, which hashes the `Long` micros value with no `[0, 24h)` check (`org.apache.iceberg.transforms.Bucket`). Validity is assumed at the typed-value / write layer, not inside the hash transform.

**Where the day-range *is* enforced in this PR**
- Iceberg writes (partition values / stats): `validateIcebergTimeOfDayMicroseconds` via `getTimeValueInMicroseconds`.
- Parquet `TIME` writes: `validateParquetTimeOfDay` in the `Time` / `Time64` converters.
So an out-of-range ClickHouse `Time` cannot be written into Iceberg/Parquet as logical `time` / `TIME`. The SQL functions `icebergHash` / `icebergBucket` remain pure Appendix B math over the ClickHouse type domain (which is wider than one day), consistent with how we hash other numeric / temporal inputs without re-applying Iceberg physical-type constraints inside the function.

**Conclusion**
We will leave the hash path as-is. Adding a day-range reject in `icebergHash` would be optional hardening for UX consistency with writes, not a correctness requirement for Iceberg interoperability.

PR #95811, opened 2026-02-02
default variant implementation throw on type mismatch

bharatnc:
Throw on `Variant` type mismatch. Fixes https://github.com/ClickHouse/ClickHouse/issues/95839.                                                                                                                                                                 

### Changelog category (leave one):
- Backward Incompatible Change

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix logical error on `Variant` type mismatch. 

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.2.1.301`
- Backported to: `26.1.3.13`

PR #105543, opened 2026-05-21
Fix `variant_throw_on_type_mismatch/dynamic_throw_on_type_mismatch=false` not catching exceptions during function execution

### LINKED ISSUE #103484: `variant_throw_on_type_mismatch=false` fails for comparison functions: `try_build` only wraps `build`, not `execute`

_Retrospective finding from a historical scan of [PR #95811](https://github.com/ClickHouse/ClickHouse/pull/95811) (merged 2026-02-06). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

Setting `variant_throw_on_type_mismatch=false` does not prevent exceptions for comparison functions (`equals`, `notEquals`, `less`, etc.) on `Variant(UInt64, String)`. The query `SELECT v = 42 FROM t` throws `NO_COMMON_TYPE` instead of returning NULL for String rows.

**Root cause:** FunctionVariantAdaptor.cpp:73-89: `try_build` lambda only wraps `function_overload_resolver->build(args)` in try-catch. When comparison functions pass `build()` (because `getReturnTypeImpl` accepts String with anything) but throw during `execute()`, the exception propagates despite `throw_on_type_mismatch=false`.

**Why we believe this is a bug:** FunctionVariantAdaptor.cpp:73-89 `try_build` lambda only catches exceptions during `build()`. FunctionsComparison.h:1248 `getReturnTypeImpl` accepts String with any type (build succeeds), but FunctionsComparison.h:1198 `executeGeneric` calls `getLeastSupertype(String, UInt8)` which throws `NO_COMMON_TYPE`. The `execute()` call at FunctionVariantAdaptor.cpp:136,254,448 is unprotected.

**Affected locations:**
- `src/Functions/FunctionVariantAdaptor.cpp:73` — try_build lambda only wraps build(), not execute()
- `src/Functions/FunctionVariantAdaptor.cpp:136` — unprotected execute() call in first code path
- `src/Functions/FunctionVariantAdaptor.cpp:448` — unprotected execute() call in general case path
- `src/Functions/FunctionDynamicAdaptor.cpp:68` — same try_build pattern in Dynamic adaptor
- `src/Functions/FunctionsComparison.h:1198` — executeGeneric calls getLeastSupertype which throws NO_COMMON_TYPE

**Impact:** Users setting `variant_throw_on_type_mismatch=false` to restore backward-compatible behavior (NULLs for incompatible types) still get exceptions for comparison functions. The setting works for arithmetic (`plus`) but not for comparisons (`equals`, `less`, etc.), creating inconsistent behavior.

Yes — confirmed on current `master` (commit `596eb6de5409`).

### How to reproduce

```sql
-- Test: variant_throw_on_type_mismatch=false should return NULL for comparison functions
-- with incompatible types, but currently throws NO_COMMON_TYPE because the try_build lambda
-- only wraps build() and not execute(). Comparison functions pass build() (String compares
-- with anything in getReturnTypeImpl) but throw in executeGeneric->getLeastSupertype.
-- Covers: FunctionVariantAdaptor.cpp try_build lambda missing execute() protection

SET enable_variant_type = 1;
SET allow_suspicious_variant_types = 1;
SET variant_throw_on_type_mismatch = false;

DROP TABLE IF EXISTS test_04113;
CREATE TABLE test_04113 (id UInt32, v Variant(UInt64, String)) ENGINE = Memory;
INSERT INTO test_04113 VALUES (1, 42), (2, 'hello'), (3, NULL), (4, 10);

-- Arithmetic works correctly: returns NULL for String variant rows
SELECT 'plus';
SELECT v + 5 FROM test_04113 ORDER BY id;

-- Comparison should also return NULL for String variant rows but throws instead
SELECT 'equals';
SELECT v = 42 FROM test_04113 ORDER BY id;

DROP TABLE IF EXISTS test_04113;
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/89832822-e8b7-4c7d-af4e-b5c466cf602d)

### Expected behavior

```
plus
47
\N
\N
15
equals
1
\N
\N
0
```

```
plus
47
\N
\N
15
equals
Code: 386. DB::Exception: There is no supertype for types String, UInt8 because some of them are String/FixedString/Enum and some of them are not: while executing 'FUNCTION equals(__table1.v :: 3, 42_UInt8 :: 2) -> equals(__table1.v, 42_UInt8) Nullable(UInt8) : 0'. (NO_COMMON_TYPE)
```

### Additional context

**Open risks:**
- The same bug exists in FunctionDynamicAdaptor.cpp for dynamic_throw_on_type_mismatch=false
- Other functions beyond comparisons might also pass build() but throw during execute()

**Suggested fix:** Extend the try-catch in `try_build` to also wrap the `execute()` calls, or create a separate `try_execute` helper that catches the same four error codes when `throw_on_type_mismatch=false`. Apply the same fix to `FunctionDynamicAdaptor.cpp`.

Found during automated review of [PR #95811](https://github.com/ClickHouse/ClickHouse/pull/95811).

---
_ClickGapAI · Confidence: HIGH · Severity: P1 · Finding: `h_pr95811_001`_

### Version info
- Resolved by: #105543
- Merged into: `26.6.1.510`

Avogar:
Closes https://github.com/ClickHouse/ClickHouse/issues/103484

Fix `variant_throw_on_type_mismatch/dynamic_throw_on_type_mismatch=false` not catching exceptions during function execution. Closes https://github.com/ClickHouse/ClickHouse/issues/103484

### Version info
- Merged into: `26.6.1.510`


alexbakharew (approved):
The initial PR https://github.com/ClickHouse/ClickHouse/pull/95811 was backported to 26.1.

@Avogar Do you know why only to 26.1 and not to other supported versions?

This fix should be backported as well

clickgapai:

### TL;DR
**Fix `variant_throw_on_type_mismatch/dynamic_throw_on_type_mismatch=false` not catching exceptions during function execution**

**ClickGap verified:** reproduced it with the script below on `26.5`; tested 3 version(s) — 2 affected, 1 clean (see Affects); bisected to a candidate range — a single introducing PR could not be pinned (see Bisect).

### Reproduction
```
SET enable_variant_type=1; SET allow_suspicious_variant_types=1; SET variant_throw_on_type_mismatch=false; SELECT v = 42 FROM (SELECT 'hello'::Variant(UInt64, String) AS v);
```

### Bisect
- Bisected on master in several steps (confidence: HIGH); a single introducing PR could not be pinned (PR #102853 only modifies Avro format parsing logic, while the bug involves variant type comparison and exception handling in function execution—completely unrelated subsystems. — a coincidental boundary, not the cause).
- The source-build bisect converged to #102853, but it is **not** the introducer (see above) — not attributing it.

### Affects
- Tested on `master` @ commit `84d36b6f9381` (aarch64)
- **Reproduces on**: `26.5, 26.4`
- **Feature not present** (repro inconclusive — the affected code path postdates these releases): `26.3, 25.8`
- **Backport needed**: 26.5, 26.4

### Suggested next step
The bisect boundary above narrows the search in `comp-json-datatype` — treat the candidate range as the starting point; the single introducing PR is unconfirmed.

**Component:** `comp-json-datatype`
**Severity:** `P2`

cc fix PR author: @Avogar
cc fix PR reviewer: @alexbakharew

---
_ClickGap · Finding: `phase_d_pr105543`_




### Triage metadata
- Primary component: `comp-json-datatype`

PR #94335, opened 2026-01-15
Fix reading columns with dot-separated names from Iceberg

mkmkme:
Fixes #94196 

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fixes an issue when Iceberg columns with dot in names returned NULL as values

### Version info
- Merged into: `26.2.1.102`

PR #35852, opened 2022-04-01
Refactoring of hints for column descriptor

antonio2368:
Changelog category (leave one):
- Not for changelog (changelog entry is not required)

Refactor of #35761 

> Information about CI checks: https://clickhouse.tech/docs/en/development/continuous-integration/

PR #105546, opened 2026-05-21
Fix Iceberg ARRAY columns with dot-separated names returning empty lists

### LINKED ISSUE #90731: Iceberg: Selecting from ARRAY column with dot-separated name returns empty lists
### Describe what's wrong

When querying an Iceberg table through iceberg table function or DataLakeCatalog, columns of type ARRAY<STRING> return empty lists if the column name contains a dot (.).
The same table queried via Spark returns correct values.

Yes

### How to reproduce

Using spark:
```sql
CREATE TABLE table7 (`a.b` ARRAY<STRING>);
INSERT INTO table7 VALUES (ARRAY('a','b','c'));
```

Read with DataLakeCatalog:
```sql
SELECT *
FROM `namespace_7d8f6b8b_c93f_11f0_8e49_e0c26496f172.table7`
```
```
   ┌─a.b─┐
1. │ []  │
   └─────┘
```

Read with iceberg table fucntion:
```sql
SELECT *
FROM iceberg('http://minio:9000/warehouse/namespace_7d8f6b8b_c93f_11f0_8e49_e0c26496f172.db/table7/', 'admin', 'password')
```
```
   ┌─a.b─┐
1. │ []  │
   └─────┘

1 row in set. Elapsed: 0.012 sec.
```

Read from spark:
```sql
select * from table7;
```
```
["a","b","c"]
Time taken: 0.113 seconds, Fetched 1 row(s)
```

il9ue:

Fix reading Iceberg tables whose `ARRAY` column names contain a dot (e.g. `` `a.b` ARRAY<STRING> ``), which previously returned empty arrays. Two upstream defects were responsible: `ColumnsDescription::getAllRegisteredNames` filtered out dotted names, and `NestedUtils::getSubcolumnsOfNested` misclassified lone dotted `Array(T)` columns as flattened `Nested` children.

---

## What's the problem

Fixes [#90731](https://github.com/ClickHouse/ClickHouse/issues/90731).

When an Iceberg table has a column whose name contains a `.` and whose type is `Array(T)` — e.g.

```sql
-- Spark
CREATE TABLE table7 (`a.b` ARRAY<STRING>);
INSERT INTO table7 VALUES (ARRAY('a','b','c'));
```

reading from ClickHouse via the `iceberg(...)` table function or a `DataLakeCatalog` returns an empty array:

```sql
SELECT `a.b` FROM iceberg('...');
-- got:      [ ]
-- expected: ['a','b','c']
```

Spark reads the same files correctly.

## Root cause

The Parquet V3 reader path (`SchemaConverter` + `ColumnMapper` + `FormatFilterInfo`) is already correct after the dotted-name field-id work in `0a218cd4e8b`, `4b733bae561` and `f24c1a46063`. The remaining symptom is caused by two upstream defects, each independent of Iceberg but exposed by it:

**1. `ColumnsDescription::getAllRegisteredNames`** explicitly filtered out any column whose name contained `.`, under the assumption such names were always flattened Nested subcolumns. A column whose stored name literally contains a dot (allowed by MergeTree with backticks, and produced by Iceberg / Spark) is a first-class registered name and must appear in `IHints` misspelling suggestions. The function is only consumed by `IHints`-style suggestion paths (and by `StorageSystemZooKeeper` for column-name iteration, where no dotted names exist), so relaxing it has no effect on parsing, planning, storage, or wire protocol.

**2. `NestedUtils::getSubcolumnsOfNested`** treated every `Array(T)` column whose name contained `.` as a flattened element of a synthetic `Nested` structure named after the prefix. This caused the Arrow, ORC and pre-V3 Parquet readers to look for a struct field with the prefix name in the data file rather than the literal dotted column, returning an empty array.

## Fix

* `ColumnsDescription::getAllRegisteredNames` — drop the dot filter; return every registered column name.
* `NestedUtils::getSubcolumnsOfNested` — use a two-pass scan: a synthetic `Nested` entry is only emitted when **at least two** `Array(T)` columns share the same dotted prefix. A lone column such as `a.b: Array(T)` no longer appears in the synthetic-Nested map. Genuine flattened `Nested` with multiple fields is unaffected; the existing early-continue on `isNested()` also covers the one-field-Nested edge case.

## Tests

* `tests/integration/test_storage_iceberg_with_spark/test_column_names_with_dots.py::test_dotted_array_column` — end-to-end repro of #90731 against `s3`, `azure` and `local` storage.
* `tests/integration/test_storage_iceberg_with_spark/test_column_names_with_dots.py::test_dotted_array_alongside_real_nested` — mixed-schema regression guard verifying that a lone dotted `Array` column coexists with a genuine flattened-Nested sibling group sharing a different prefix.
* `tests/queries/0_stateless/04259_dotted_array_not_nested.sql` — isolates the `NestedUtils` fix without Iceberg using the `Memory` engine.
* `tests/queries/0_stateless/04260_dotted_column_in_hints.sh` — verifies the `ColumnsDescription` fix by checking the misspelling-hint output after an `UNKNOWN_IDENTIFIER`-style error.

## Scope / out-of-scope

* No new settings.
* No header changes (`ColumnsDescription.h`, `NestedUtils.h` untouched).
* Parquet V3 path (`SchemaConverter`, `Reader`, `FormatFilterInfo`, `ArrowColumnToCHColumn`, `Storages/ObjectStorage/DataLakes/Iceberg/*`) is **not** modified — already fixed by the three commits cited above.


il9ue:
Closing this — the read-path fix for #90731 is already in master, and what remains of this PR is net-harmful. Writing up what I found so the record is clear.

**The actual fix is already merged.** The dotted-name Parquet V3 read path was fixed by 0a218cd, 4b733ba and f24c1a4 (merged via #94335, mkmkme's `dot-issue`), all now in master. `useColumnMapperIfNeeded` returns the full dotted name instead of splitting to the last component, which covers the ARRAY case in #90731. This PR's own description already cited those three commits as "already correct" — so the read path here changes nothing that master doesn't already do.

**The hint piece.** The only other change is dropping the dot-filter in `ColumnsDescription::getAllRegisteredNames` so dotted names appear in `Maybe you meant: [...]` suggestions. That’s a small nice-to-have, unrelated to #90731, and it reverses the intentional exclusion added in #35852 — so it deserves its own discussion, not a ride-along here.

Closing as superseded by #94335 (0a218cd / 4b733ba / f24c1a4). 
Apologies for the noise on this one — the read fix had already landed and I didn’t check that early enough.

PR #105548, opened 2026-05-21
Fix `histogram` producing wrong results for small unsorted inputs

### LINKED ISSUE #103109: `compress` early return before sorting breaks `unique` invariant, producing wrong `histogram` results for small unsorted inputs

_Retrospective finding from a historical scan of [PR #85760](https://github.com/ClickHouse/ClickHouse/pull/85760) (merged 2025-12-17). Confirmed on current codebase — close with a note if already fixed._

### Describe what's wrong

`histogram(N)(x)` produces different (incorrect) results depending on the order of input values when the number of distinct values is ≤ N. For example, `histogram(10)` on values [5,3,1,4,2] produces 2 bins instead of 5, and on [5,4,3,2,1] produces 1 bin instead of 5.

**Root cause:** AggregateFunctionHistogram.cpp:122-123: `if (size <= max_bins) return;` exits before the sort at lines 125-135. The old code had `sort()` as the first statement in `compress()`, guaranteeing sorted data for `unique()` regardless of size.

**Why we believe this is a bug:** `insertResultInto` (line 254) calls `compress(max_bins)` then `unique()` (line 255). `unique()` (line 210) documents 'Assumes that points are sorted'. In the old code, `compress()` always called `sort()` first, then checked the early return. The PR moved the early return (`size <= max_bins`) to line 122-123, BEFORE the sorting logic at lines 125-135. When `size <= max_bins`, `compress()` returns without sorting, and `unique()` operates on unsorted data, incorrectly merging non-adjacent values.

**Affected locations:**
- `src/AggregateFunctions/AggregateFunctionHistogram.cpp:122` — Early return in compress() before sorting
- `src/AggregateFunctions/AggregateFunctionHistogram.cpp:212` — unique() assumes sorted data
- `src/AggregateFunctions/AggregateFunctionHistogram.cpp:254` — insertResultInto calls compress then unique

**Impact:** Silent wrong results for `histogram(N)` when the total number of input values (before any compression) is ≤ N and values are not inserted in sorted order. The histogram incorrectly merges bins that should be separate, producing fewer bins with wrong means and weights. This affects any query using `histogram()` with few unsorted data points relative to the bin count.

Yes — confirmed on current `master` (commit `6be8f3df2b2b`).

### How to reproduce

```sql
-- Test: histogram() must produce identical results regardless of input order when count <= max_bins
-- Covers: AggregateFunctionHistogram.cpp:122 - early return from compress() before sorting
-- Bug: unique() at line 212 'Assumes that points are sorted' but compress() returns early
-- without sorting when size <= max_bins, causing wrong bin merging on unsorted inputs.

-- ascending (baseline correct result)
SELECT 'ascending';
SELECT histogram(10)(x) FROM (SELECT number + 1 AS x FROM numbers(5));

-- unsorted (same values, different order - must match ascending)
SELECT 'unsorted';
SELECT histogram(10)(x) FROM (SELECT arrayJoin([5,3,1,4,2]) AS x);

-- descending (same values, reversed - must match ascending)
SELECT 'descending';
SELECT histogram(10)(x) FROM (SELECT 5 - number AS x FROM numbers(5));

-- more values, still under max_bins threshold
SELECT 'ascending 8';
SELECT histogram(10)(x) FROM (SELECT number + 1 AS x FROM numbers(8));

SELECT 'shuffled 8';
SELECT histogram(10)(x) FROM (SELECT arrayJoin([4,7,2,8,1,6,3,5]) AS x);
```

[Try it on ClickHouse Fiddle](https://fiddle.clickhouse.com/d8f56280-7eb0-40f8-af22-8821b47ad241)

### Expected behavior

```
ascending
[(1,1.5,1),(1.5,2.5,1),(2.5,3.5,1),(3.5,4.5,1),(4.5,5,1)]
unsorted
[(1,1.5,1),(1.5,2.5,1),(2.5,3.5,1),(3.5,4.5,1),(4.5,5,1)]
descending
[(1,1.5,1),(1.5,2.5,1),(2.5,3.5,1),(3.5,4.5,1),(4.5,5,1)]
ascending 8
[(1,1.5,1),(1.5,2.5,1),(2.5,3.5,1),(3.5,4.5,1),(4.5,5.5,1),(5.5,6.5,1),(6.5,7.5,1),(7.5,8,1)]
shuffled 8
[(1,1.5,1),(1.5,2.5,1),(2.5,3.5,1),(3.5,4.5,1),(4.5,5.5,1),(5.5,6.5,1),(6.5,7.5,1),(7.5,8,1)]
```

```
ascending
[(1,1.5,1),(1.5,2.5,1),(2.5,3.5,1),(3.5,4.5,1),(4.5,5,1)]
unsorted
[(1,3,2.875),(3,5,2.125)]
descending
[(1,5,5)]
ascending 8
[(1,1.5,1),(1.5,2.5,1),(2.5,3.5,1),(3.5,4.5,1),(4.5,5.5,1),(5.5,6.5,1),(6.5,7.5,1),(7.5,8,1)]
shuffled 8
[(1,4.25,1.125),(4.25,4.5,1.875),(4.5,4.5,2),(4.5,4.75,1.875),(4.75,8,1.125)]
```

### Additional context

**Open risks:**
- The `merge()` path is also affected — merging two small histogram states where the merged count ≤ max_bins will produce unsorted data that is not sorted before `unique()`

**Suggested fix:** Move the sorting logic before the early return, or add a separate sort step when `size <= max_bins`. The simplest fix: move lines 125-135 (the sorting block) before line 122 (the early return check), matching the old code's behavior of always sorting before the size check.

Found during automated review of [PR #85760](https://github.com/ClickHouse/ClickHouse/pull/85760).

---
_ClickGapAI · Confidence: HIGH · Severity: P0 · Finding: `h_pr85760_001`_

Avogar:

Fix `histogram` producing wrong results for small unsorted inputs. Closes https://github.com/ClickHouse/ClickHouse/issues/103109


alexey-milovidov (approved):

clickgapai:

**Affected code:** `AggregateFunctionHistogramData::compress` early return logic — introduced in 26.1.

**Why this might apply:** The bug exists in 26.1+ where the sorted_prefix optimization was introduced. Verified by checking upstream/26.2, upstream/26.3, upstream/26.4 - all have the early return before sorting. Branch 25.8 is NOT affected (it calls sort() before the early return check). This is a wrong results bug affecting a commonly used aggregate function.

Other supported branches (`25.8`) predate 26.1 and do not contain this code.

clickgapai:
Hi @Avogar @alexey-milovidov — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: PR description is empty - no explanation of what was wrong or how it was fixed.
- Suggested addition:

  > The `compress()` function in histogram had an early return `if (size <= max_bins) return;` BEFORE sorting the data. This caused `unique()` to operate on unsorted points, incorrectly merging non-adjacent bins based on their position rather than their value proximity. The fix moves the early return AFTER the sorting logic, ensuring data is always sorted before `unique()` is called.

PR #105571, opened 2026-05-21
Add system.keeper_snapshots table

mstetsyuk:
- Improvement

Add a new keeper-only `system.keeper_snapshots` table with information about local ClickHouse Keeper snapshots.

### Version info
- Merged into: `26.6.1.631`


mstetsyuk:
The failing test in `CH Inc sync` is fixed here: https://github.com/ClickHouse/clickhouse-private/pull/59949

src/Storages/System/StorageSystemKeeperSnapshots.cpp
```diff
@@ -0,0 +1,74 @@
...
+        try
+        {
+            if (entry.disk->existsFile(entry.path))
+            {
+                size_bytes = entry.disk->getFileSize(entry.path);
+                last_modified_at = static_cast<UInt32>(entry.disk->getLastModified(entry.path).epochTime());
+            }
+            else if (!entry.is_received)
+            {
+                LOG_WARNING(log, "Finalized snapshot file {} on disk {} is missing", entry.path, entry.disk->getName());
```
jkartseva: Will it be logged on every `SELECT`?  IMO this logging should be handled by KeeperSnapshotManager, which is responsible for the snapshot lifecycle, not by the reporting tool. Timestamp reported in the logs may be confusing because he time reported != the time when it was missed. Maybe drop checking for this condition altogether.

jkartseva (approved):
LGTM with a small nit.

src/Storages/System/StorageSystemKeeperSnapshots.cpp
mstetsyuk: Removed this log in favour of a `exists_on_disk` column.

antonio2368 (approved):

antonio2368 (approved):

PR #105572, opened 2026-05-21
Initialize ASTIndexDeclaration::granularity in header

### LINKED ISSUE #105517: [SAST] Uninitialized granularity field in ASTIndexDeclaration
### Describe the bug

Hello! We were running static analysis on ClickHouse with Svace and found a potentially uninitialized field in `ASTIndexDeclaration`.

> File: `ClickHouse/src/Parsers/ASTIndexDeclaration.cpp`
> Class: `ASTIndexDeclaration`
> Field: `granularity`
> 

The constructor initializes only `name`:

```
```cpp
ASTIndexDeclaration::ASTIndexDeclaration(ASTPtr expression, ASTPtr type, const String & name_)
    : name(name_)
{
    ...
}

```

However, the field granularity is declared without a default value:

`UInt64 granularity;`

Later this field is read in formatImpl():

```
if (granularity)
{
    ostr << " GRANULARITY ";
    ostr << granularity;
}
```

and also copied in clone():

`res->granularity = granularity;`

So if an ASTIndexDeclaration object is created and granularity is not explicitly assigned afterwards, the code may read an uninitialized value.

There is already a default constant:

`static const auto DEFAULT_INDEX_GRANULARITY = 1uz;`

so the field probably should be initialized with it.

### How to reproduce

Suggested patch:

`UInt64 granularity = DEFAULT_INDEX_GRANULARITY;`

or alternatively in the constructor initializer list:

```
ASTIndexDeclaration::ASTIndexDeclaration(ASTPtr expression, ASTPtr type, const String & name_)
    : name(name_)
    , granularity(DEFAULT_INDEX_GRANULARITY)
{
    ...
}
```

Not sure if this issue is worth opening a PR since it looks minor, but still reporting in case you decide to address it. Thank you for your time!

Found by[ Linux Verification Center ](https://portal.linuxtesting.ru/)with SVACE.

UNINIT.CTOR
ASTIndexDeclaration.cpp:17

Constructor declared at ASTIndexDeclaration.cpp:17 may not initialize class members of 'DB::ASTIndexDeclaration'.
Following members aren't initialized: granularity.

Field declaration:
ASTIndexDeclaration.h:21
UInt64 granularity;

### Version info
- Resolved by: #105572
- Merged into: `26.6.1.99`

groeneai:
SVACE static analysis (UNINIT.CTOR) flagged that `ASTIndexDeclaration`'s 3-argument constructor only initializes `name`, leaving `granularity` (a plain `UInt64`) with no in-class default. `formatImpl` reads the field via `if (granularity)` and `clone` copies it, so a caller that forgets to assign `granularity` would read indeterminate memory.

All four current construction sites (`ParserCreateQuery`, `ParserCreateIndexQuery`, `IndicesDescription`, and `clone` itself) already assign `granularity` immediately after construction, so the finding is theoretical for current code. The change is defense in depth: future call sites get a sensible default and the SAST warning goes away.

Closes #105517.

N/A (SVACE hygiene fix; no user-visible behavior change).

- [x] Documentation is unaffected.

### Version info
- Merged into: `26.6.1.99`

### DIFF
--- a/src/Parsers/ASTIndexDeclaration.h
+++ b/src/Parsers/ASTIndexDeclaration.h
@@ -19,7 +19,7 @@ public:
     ASTIndexDeclaration(ASTPtr expression, ASTPtr type, const String & name_);
 
     String name;
-    UInt64 granularity;
+    UInt64 granularity = DEFAULT_INDEX_GRANULARITY;
     bool part_of_create_index_query = false;
 
     /** Get the text that identifies this element. */


groeneai:
### Pre-PR validation gate

This is an SAST-driven hygiene fix, not a CI failure reproduction. Some items are N/A.

| # | Question | Answer |
|---|----------|--------|
| a | Deterministic repro? | N/A. The bug is theoretical, not currently reachable: all four call sites assign `granularity` immediately after construction. The SAST finding is a code-pattern warning, not a runtime UB observable in current code. |
| b | Root cause explained? | Yes. `UInt64 granularity;` in the header has no in-class default, and the 3-arg constructor body only assigns `name`. Per `[basic.indet]`, a non-static `UInt64` member left unassigned by every constructor (initializer list or body) holds an indeterminate value. `formatImpl` reads it via `if (granularity)` and `clone` copies it, so a future caller that constructs the AST and forgets to set `granularity` would read indeterminate memory. |
| c | Fix matches root cause? | Yes. Adding `= DEFAULT_INDEX_GRANULARITY` to the field declaration makes every constructor (current and future) leave the member in a defined state. This is exactly what the SVACE UNINIT.CTOR class expects. |
| d | Test intent preserved? | N/A. No tests are touched. All existing parser tests still pass because the field is always overwritten by the caller right after construction, so no observable output changes. |
| e | Both directions demonstrated? | Partial. The "without fix" direction is not reachable as a runtime UB in current code (no caller leaves `granularity` unset). The "with fix" direction is verified by local compilation: `ASTIndexDeclaration.cpp`, `ParserCreateQuery.cpp`, `ParserCreateIndexQuery.cpp` all build cleanly under `clang-21 -Weverything`. |
| f | Fix is general, not narrow? | Yes. One default initializer covers every constructor invocation (current and future) instead of patching the one symptom site that SAST happened to surface. There is no symmetric class with the same pattern in `src/Parsers/AST*Declaration.h` (`ASTProjectionDeclaration`, `ASTColumnDeclaration`, `ASTStatisticsDeclaration` all have no UInt64 numeric members lacking defaults). No upstream defect to chase: the fix is in the correct place (the field declaration). |

Verified all callers via grep on `make_intrusive<ASTIndexDeclaration>` / `new ASTIndexDeclaration`:
- `src/Parsers/ParserCreateQuery.cpp:176-177` -> sets via `getSecondaryIndexGranularity`
- `src/Parsers/ParserCreateIndexQuery.cpp:81-83` -> sets via `getSecondaryIndexGranularity`
- `src/Storages/IndicesDescription.cpp:286-290` -> sets to `DEFAULT_INDEX_GRANULARITY`
- `src/Parsers/ASTIndexDeclaration.cpp:44-45` (the `clone` method) -> copies `this->granularity`

session: cron:clickhouse-ci-task-worker:20260521-180000

groeneai:
cc @nikitamikhaylov @evillique - trivial 1-line defense-in-depth fix for the UNINIT.CTOR warning you asked about on #105517. Could you take a look?

nikitamikhaylov (approved):

nikitamikhaylov:
Integration test is fixed here: https://github.com/ClickHouse/ClickHouse/pull/105636

PR #105574, opened 2026-05-21
Support Buffers and fuel mode for WASM UDFs

antonio2368:
Add `Buffers` serialization support for WebAssembly UDFs using `ABI BUFFERED_V1`, and move `webassembly_udf_enable_fuel` from a query setting to a persisted WASM UDF function setting. This lets users choose fuel accounting when creating the UDF, while keeping `webassembly_udf_max_fuel` as the per-query fuel budget for fuel-enabled functions.

The implementation compiles the selected `FuelMode` during WASM UDF creation/load and caches Wasmtime module variants by WASM module name and fuel mode. It also keeps WasmEdge cancellation working when finite fuel limits are disabled.

- Experimental Feature

Support `Buffers` serialization for WebAssembly UDFs using `ABI BUFFERED_V1`, and add `webassembly_udf_enable_fuel` as a persisted WASM UDF function setting.

### Documentation entry:
- [x] Documentation is updated.

---

> [!NOTE]
> **Medium Risk**
> Changes WebAssembly UDF execution by adding per-function fuel-mode specialization and engine/module caching behavior, which can affect runtime cancellation and resource limits. Also expands supported `BUFFERED_V1` serialization to `Buffers`, so correctness depends on strict envelope handling and format parsing.
> 
> **Overview**
> Adds `Buffers` as a supported `serialization_format` for `ABI BUFFERED_V1` WebAssembly UDFs and documents the required single-column return contract.
> 
> Introduces a persisted per-UDF setting `webassembly_udf_enable_fuel` (with boolean normalization/validation), and updates `webassembly_udf_max_fuel` semantics to apply only when fuel is enabled.
> 
> Refactors the WASM engine interface to compile modules with an explicit `FuelMode`, caches separate module variants per `(module_name, fuel_mode)` when required (Wasmtime), and adjusts WasmEdge/WasmTime runtimes to keep cancellation working while allowing fuel accounting to be disabled. Adds stateless tests covering fuel toggling and the new `Buffers` buffered-ABI path, and fixes buffered-result parsing by creating the result column in the header.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79b7ab4a82f9a1f994f8add302a33a006310c280. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.187`


docs/en/sql-reference/functions/wasm_udf.md
```diff
@@ -225,7 +225,8 @@ RETURNS return_type
...
+    - `serialization_format` String — Serialization format for ABI requires it. Supported values: `MsgPack`, `JSONEachRow`, `CSV`, `TSV`, `TSVRaw`, `RowBinary`, and `Buffers`. Default: `MsgPack`. Under `ABI BUFFERED_V1`, `Buffers` uses the declared function schema and ClickHouse Native per-column binary encoding. When `serialization_format = 'Buffers'`, the guest must return a single-column `Buffers` block whose `num_rows` equals the input block row count and whose column type matches the declared `RETURNS` type.
+    - `webassembly_udf_enable_fuel` Bool — Enables finite fuel budgeting for the function. Default: `true`. Set at function creation time via `CREATE FUNCTION ... LANGUAGE WASM ... SETTINGS ...`; the value is persisted with the function definition. When `false`, the query-level setting `webassembly_udf_max_fuel` is ignored for this function. `webassembly_udf_max_memory` still applies. Query cancellation via `max_execution_time`, `KILL QUERY`, and shutdown still applies. For `webassembly_udf_engine = 'wasmedge'`, ClickHouse keeps WasmEdge cost measuring active internally to deliver cancellation; for Wasmtime, epoch interruption remains enabled.
```
vdimir: Looks a bit wordy and includes extra implementation details, imo

```suggestion
    - `webassembly_udf_enable_fuel` Bool — Enables finite fuel budgeting for the function. Default: `true`. When `false`, the query-level setting `webassembly_udf_max_fuel` is ignored for this function. Disabling fuel limits may improve performance when using the `wasmtime` engine. However, for untrusted or buggy guest code, it can increase the risk of runaway execution.
```

docs/en/sql-reference/functions/wasm_udf.md
```diff
@@ -225,7 +225,8 @@ RETURNS return_type
 - `DETERMINISTIC`: Declares the function as deterministic — always returns the same output for the same input. When specified, ClickHouse may constant-fold calls where all arguments are constants: the function is evaluated once at query analysis time and the result is reused for every row.
 - `SHA256_HASH`: Expected module hash for verification (auto-filled if omitted), can be used to ensure the correct WASM module loaded across different replicas.
 - `SETTINGS`: Per-function settings
-    - `serialization_format` String — Serialization format for ABI requires it. Default: `MsgPack`.
+    - `serialization_format` String — Serialization format for ABI requires it. Supported values: `MsgPack`, `JSONEachRow`, `CSV`, `TSV`, `TSVRaw`, `RowBinary`, and `Buffers`. Default: `MsgPack`. Under `ABI BUFFERED_V1`, `Buffers` uses the declared function schema and ClickHouse Native per-column binary encoding. When `serialization_format = 'Buffers'`, the guest must return a single-column `Buffers` block whose `num_rows` equals the input block row count and whose column type matches the declared `RETURNS` type.
```
vdimir:
```suggestion
    - `serialization_format` String — Serialization format for ABI requires it. Supported values: `MsgPack`, `JSONEachRow`, `CSV`, `TSV`, `TSVRaw`, `RowBinary`, and `Buffers`. Default: `MsgPack`. Block-based formats such as `Buffers` must return a single column whose type match the declared function signature.
```

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
```diff
@@ -208,11 +247,7 @@ class WasmTimeCompartment : public WasmCompartment
 
     VectorWithMemoryTracking<WasmVal> invokeImpl(std::string_view function_name, const VectorWithMemoryTracking<WasmVal> & params, StopToken stop_token) override
     {
-        {
-            auto result = store.context().set_fuel(cfg.fuel_limit ? cfg.fuel_limit : std::numeric_limits<uint64_t>::max());
-            if (!result)
-                throw Exception(ErrorCodes::WASM_ERROR, "Failed to set fuel to wasm instance: {}", result.err().message());
-        }
+        setStoreFuel(store.context(), cfg, "function call");
```
vdimir, line 250: I'm not sure why it's introduced in function instead of wrapping three lines below into `if fuel_mode`, but it's just nitpick.

docs/en/sql-reference/functions/wasm_udf.md
vdimir: Also, what is the main motivation behind introducing this setting at the function level?

Possible alternatives I considered

* Server level - Pros: simple configuration model. Cons: no per-module differentiation, and enabling it may require changing server config.

* Module level - Allows distinguishing between “safe” and “less trusted” modules. The downside is that configuration may become awkward, since modules are added via `INSERT INTO ...`; this would likely require attaching settings metadata via Map column for example, which is somewhat unconventional.

* Function level - current approach

* Runtime-only decision based on `webassembly_udf_max_fuel = 0` - Very flexible, but would require always storing two compiled versions of each module, increasing memory usage even for cases when we use only single fuel mode.


And another dimension here is access control - who is allowed to disable fuel limits: the server administrator / the module / the function creator or even the query user (this could also potentially be restricted through existing settings constraints mechanisms).

Curious to hear your thoughts.

src/Interpreters/WasmModuleManager.cpp
```diff
@@ -223,34 +274,140 @@ std::string WasmModuleManager::loadModuleImpl(std::string_view module_name)
...
-                throw Exception(ErrorCodes::RESOURCE_NOT_FOUND, "WebAssembly module '{}' not found", module_name);
-            if (auto res = validateModuleFile(user_scripts_disk, module_path); !res)
-                throw Exception(ErrorCodes::INCORRECT_DATA, "Cannot load WebAssembly module '{}': {}", module_name, res.error().text);
-            /// If file exists on disk but is not registered yet, we proceed to load it below
-        } else if (auto module = it->second.ptr.lock())
-            return {module, it->second.hash};
+            auto [inserted_it, _] = modules.emplace(String(module_name), ModuleRef{});
+            it = inserted_it;
+            it->second.is_transient_load_reservation = true;
+        }
```
vdimir: I didn’t fully understand why we need to add complexity to the loading mechanism. Couldn’t this be handled similarly current approach with locking, i.e. load the required module version on demand if it hasn’t been loaded yet?

antonio2368: We can keep the old approach, no problem.
I was just interested how the code would look like if we allowed multiple UDFs to be loaded, written, compiled in parallel. Because currently (AFAIU) all those heavy operations block other UDFs.

docs/en/sql-reference/functions/wasm_udf.md
antonio2368: It seemed most natural to me. First I tried the query setting but that would mean 2 compiled modules OR lazy loading during query time. I didn't like both options much so I just moved to function level where serialization_format is already defined.

Server level - you said it, you cannot differentiate between modules then
Module level - what if we want to have some functions from module without fuel (small, cheap functions) and some with fuel (heavier functions).
Also I assume we would need to extend table schema?

Good comment for access control, I guess we should have a server setting to forbid creating UDFs without fuel if someone wants that?

vdimir:
Wasmtime mentions that difference between fuel and epoch based is only second not deterministic, but does not mention possibility of runaway, maybe it's alright for us actually 

https://docs.wasmtime.dev/examples-interrupting-wasm.html#non-deterministic-epochs

src/Interpreters/WebAssembly/WasmTimeRuntime.cpp
antonio2368: We use it in 2 places so it seemed more logical like this.

src/Interpreters/WasmModuleManager.cpp
vdimir: Does this contention happens only on `CREATE FUNCTION` which instantiate module? Or there are any other operations that blocks `getModule`?

antonio2368: It's any methods managing modules so I assume INSERT for saving module?
I see it's controversial and of questionable impact, so I'll just revert to the previous simpler locking.
Especially not to pull away from the real goal of this PR.

vdimir (approved):

PR #105577, opened 2026-05-21
Revert "Do not invalidate dictionary if there is only one of them [again]"

Avogar:
Reverts ClickHouse/ClickHouse#103662


bobrik:
I think I have the repro and a fix here. I'll open a PR later today after I polish the code a bit more.

bobrik:
#105605 👀

Avogar:
Great! Let me take a look, I will postpone the revert

PR #105552, opened 2026-05-21
Add regression test for set index under MV layer type change (#89802)

groeneai:
Closes: https://github.com/ClickHouse/ClickHouse/issues/89802

Regression test only. The LOGICAL_ERROR from #89802 (`Unexpected return type from equals. Expected Nullable(UInt8). Got UInt8.` in `MergeTreeIndexConditionSet`, hit through a `MaterializedView` that declares a `set`-indexed storage column as `Nullable` with `query_plan_merge_expressions = 0`) no longer reproduces on current master. Verified: `04272_89802_set_index_mv_nullable_column_type` passes on pristine master HEAD with no code change, so this PR drops the earlier `MergeTreeIndexSet.{h,cpp}` fix and keeps only the test to lock the fixed behavior and close the issue.

Reproducer: https://fiddle.clickhouse.com/6dc1980e-30b2-4082-b570-5ef45287afc9

The test covers the original report plus matching, `AND`/`OR`/`NOT`, `IN (...)`, and `UInt8`-vs-`Nullable(Int32)` variants, under both `query_plan_merge_expressions` values.

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Version info
- Merged into: `26.8.1.330` (included in `26.8` and later)

PR #105041, opened 2026-05-15
Fix LOGICAL_ERROR for reused materialized CTE filtered by IN over another materialized CTE

novikd:
## Motivation

Two related shapes of `LOGICAL_ERROR: Reading from materialized CTE 'X' before it
has been materialized` triggered by `enable_materialized_cte`.

**Issue #101940**:

    WITH
        ct AS MATERIALIZED (SELECT c FROM t LIMIT 10),
        rs AS MATERIALIZED (SELECT * FROM t WHERE c IN (SELECT c FROM ct))
    SELECT count()
    FROM rs AS a ANY LEFT JOIN rs AS b USING c
    SETTINGS enable_materialized_cte = 1;

**Issue #102320**:

    WITH a AS MATERIALIZED (SELECT number AS id FROM numbers(100))
    SELECT id FROM a
    WHERE id IN (
        SELECT id FROM test
        WHERE id IN (SELECT id FROM a GROUP BY id)
        GROUP BY id
    )
    GROUP BY id;

`EXPLAIN` on either query also threw — the bug fires during plan
optimization, not just at execution time.

## Root cause

The Planner installs two layers of `DelayedMaterializingCTEsStep`:

1. *Outer* steps at the top of the main plan, one per CTE dependency level,
   added by `addBuildSubqueriesForMaterializedCTEsIfNeeded` so the materialized
   CTEs run before the main pipeline at runtime.
2. *Safety-net* steps inside every IN-subquery's plan, planted by
   `forceMaterializeCTE` so the synchronous inplace path
   (`FutureSetFromSubquery::buildSetInplace` / `buildOrderedSetInplace`,
   reached via `KeyCondition::tryPrepareSetIndexForIn` or
   `VirtualColumnUtils::buildSetsForDAG`) can materialize a referenced CTE
   during plan optimization.

`makePlansForCTEs` did the atomic claim, the `optimize` call, and the
`std::move` of `materialized_cte->plan` in one step, *and*
`resolveMaterializingCTEs` ran before `addStepsToBuildSets`. So in both bug
shapes the outer step claimed the CTE and moved its plan out before the
inner safety-net inside the inplace pipeline could win the race —
the safety-net pass-throughs read an empty `StorageMemory`.

## Fix

Three coordinated changes:

1. **Split `resolveMaterializingCTEs` into two traversals.** Pass 1
   (`DelayedMaterializingCTEsStep::optimizePlans`) just optimizes each CTE's
[...]

PR #104419, opened 2026-05-08
Iceberg: invalidate metadata cache on ALTER

groeneai:
Closes #104295.

`Iceberg::alter` writes a new metadata file but never invalidated the metadata files cache. With `iceberg_metadata_staleness_ms` non-zero, the cache kept serving the pre-`ALTER` schema for the staleness window. The follow-up `INSERT` then ran with two inconsistent views of the table:

- `IcebergStorageSink` always reads fresh metadata (`force=true`), so it sees the post-`ALTER` schema (one column).
- The upstream `INSERT` pipeline reads through the cached storage metadata, so it produces chunks shaped by the pre-`ALTER` schema (two columns).

`DataFileStatistics::getColumnSizes` iterates `column_sizes.size()` (set from the chunk, two entries) and indexes `field_ids` (set from the sink's `current_schema`, one entry), causing an out-of-bounds vector access at `DataFileStatistics.cpp:68` and aborting the server.

The fix mirrors the cache invalidation already present in `IcebergStorageSink::initializeMetadata`: after `Iceberg::alter` successfully writes a new metadata file, drop the table's entries from `metadata_cache` so subsequent operations re-read the fresh schema.

Reproducer (from the issue, plus an `iceberg_metadata_log` config entry to surface the staleness path):

```sql
CREATE TABLE t0 (c0 Int, c1 Int) ENGINE = IcebergLocal('...');
INSERT INTO TABLE t0 (c1, c0) VALUES (1, 1);
SET iceberg_metadata_staleness_ms = 60000;
ALTER TABLE t0 DROP COLUMN c0;
INSERT INTO TABLE t0 (c1) SELECT 2;  -- aborted before the fix
```

Stack trace from the abort (matches the issue):

```
DataFileStatistics::getColumnSizes() const   @ DataFileStatistics.cpp:68
generateManifestFile(...)                    @ IcebergWrites.cpp:357
IcebergStorageSink::initializeMetadata()     @ IcebergWrites.cpp:999
IcebergStorageSink::finalizeBuffers()        @ IcebergWrites.cpp:857
```

Validated locally with the debug build:

- Without the fix: 100% reproducible abort with the stack above.
- With the fix: 10/10 deterministic passes of the new regression test, plus the existing iceberg ALTER tests (`04033`, `04042`, `04043`, `04061`) keep passing.

Added a stateless regression test (`04214_iceberg_alter_drop_column_stale_metadata_cache_104295`) that runs the exact reproducer.

[...]

PR #105583, opened 2026-05-21
Reject NaN and infinite polygon vertices at dictionary load time

groeneai:
The polygon dictionary load path called `addNewPoint` with whatever `Float64` value the source column produced. `SlabsPolygonIndex` then sorted those x-coordinates with `::sort` and deduped with `std::unique`, both of which are undefined when the input contains a `NaN`. The slab walker would later end up with edge endpoints that did not match `sorted_x[l]` / `sorted_x[r]`, fire the invariant check at `PolygonDictionaryUtils.cpp:188`, and abort the server with a `LOGICAL_ERROR` (`Error occurred while building polygon index. Edge ... but found ...`).

The query-side path in `IPolygonDictionary::extractPoints` already rejects `NaN` and infinite components with `BAD_ARGUMENTS`. This change mirrors the same check on the build-side, in `addNewPoint`, so a malformed source table fails at load with a clear user-facing error instead of aborting the server inside the index builder.

`tests/queries/0_stateless/04273_polygon_dict_nan_inf_rejected.sql` covers both `POLYGON_INDEX_EACH` and `POLYGON_INDEX_CELL`, plus a follow-up clean reload to confirm a sanitised source still loads.

Local reproducer (before the fix this aborts the server in a debug build):

```sql
INSERT INTO src VALUES ([[[(0., 0.), (1., 0.), (nan, 1.), (0., 1.), (0., 0.)]]], 1);
SYSTEM RELOAD DICTIONARY d;
```

Surfaced by the server-side AST fuzzer mutating `04204_polygon_dict_bytes_allocated.sql` into inserts like `(cx, cy + nan)`.

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=1b384725455cdbaa49114566ba6b6e25f15809ed&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_tsan%29

Polygon dictionaries with `LAYOUT(POLYGON_INDEX_EACH())` or `LAYOUT(POLYGON_INDEX_CELL())` now reject source rows whose polygon vertex coordinates contain `NaN` or infinity at load time with `BAD_ARGUMENTS`, instead of aborting the server inside the slab index builder.


groeneai:
### Pre-PR validation gate

**(a) Deterministic repro?** Yes:
```sql
CREATE TABLE src (polygon Array(Array(Array(Tuple(Float64, Float64)))), city_id UInt32) ENGINE = Memory;
INSERT INTO src VALUES ([[[(0., 0.), (1., 0.), (nan, 1.), (0., 1.), (0., 0.)]]], 1);
INSERT INTO src VALUES ([[[(0., 0.), (1., 0.), (1., 1.), (0., 1.), (0., 0.)]]], 2);
CREATE DICTIONARY d (...) ... LAYOUT(POLYGON_INDEX_EACH());
SYSTEM RELOAD DICTIONARY d;
```
On a debug build at master without this fix, this aborts the server with `Logical error: Error occurred while building polygon index. Edge 0  is [nan, 0] but found [1, 1]. l = 3, r = 3` (signal 6).

**(b) Root cause explained?** Yes. `IPolygonDictionary::loadData -> extractPolygons -> addNewPoint -> Data::addPoint` stores polygon vertex `Float64`s directly into a `bg::polygon`. Each `SlabsPolygonIndex` ctor then calls `uniqueX(polygons)` which runs `::sort` and `std::unique` over the x-coordinates. Both are undefined when the input contains `NaN` (NaN doesn't satisfy strict weak ordering, and `NaN == NaN` is false so `std::unique` never dedupes it). The slab loop in `indexBuild` then misroutes edges into `edge_left` / `edge_right` and the post-check `sorted_x[l] != all_edges[i].l.x() || sorted_x[r] != all_edges[i].r.x()` fires.

**(c) Fix matches root cause?** Yes. The fix rejects non-finite vertex components at load time, before they reach the index builder. This is the same boundary already used by the query-side path `IPolygonDictionary::extractPoints` (lines 423-429), so the dict's input contract is now symmetric on both ends.

**(d) Test intent preserved? / new tests added?** New test `04273_polygon_dict_nan_inf_rejected` locks the contract for both `POLYGON_INDEX_EACH` and `POLYGON_INDEX_CELL`, including a follow-up sanity reload that confirms a clean source still loads. No existing test was modified.

**(e) Demonstrated in both directions?** Yes.
- Before the fix: `SYSTEM RELOAD DICTIONARY` causes signal 6 with the `Logical error: Error occurred while building polygon index. Edge ...` trace (reproduced locally on the debug binary).
- After the fix: same reload returns `Code: 36. DB::Exception: PolygonDictionary input polygon vertex component must not be NaN. (BAD_ARGUMENTS)` and the server continues serving queries. `SELECT version()` immediately afterwards still returns `26.6.1.1`.

**(f) Fix is general?** Yes.
- All polygon vertices flow through the single `addNewPoint` entry point (line 615), called from both `handlePointsReprByArrays` and `handlePointsReprByTuples` (`InputType::SimplePolygon` and `InputType::MultiPolygon` both end up here). No other call sites add polygon points.
- The symmetric query-side validation was already in `extractPoints`; this PR plugs the only remaining hole.
- Upstream: `Float64` columns are allowed to store `NaN` and infinity (IEEE 754 / SQL semantics), so we cannot reject this earlier without breaking unrelated functionality. The dict load boundary is the correct place to enforce the polygon-geometry input contract.

### CIDB

```
check_name                                              cnt  last_seen
Stress test (experimental, serverfuzz, arm_tsan)        1    2026-05-21 17:55:40 UTC
```

Single hit in the last 90 days, triggered when the AST fuzzer mutated `04204_polygon_dict_bytes_allocated.sql` (added 2026-05-20 by #105431) into inserts containing `cy + nan`. No prior occurrences on other sanitizers / arches.

cc @groeneai

groeneai:
cc @vdimir @vitlibar — could you review? This adds NaN/Inf validation in the polygon dictionary load path to match the existing query-side validation in `extractPoints`, fixing a serverfuzz `LOGICAL_ERROR` (STID 3147-4f6a).

groeneai:
**Bugfix validation FAIL update.** Investigated.

`Bugfix validation (functional tests)` reports `Failed to reproduce the bug` for `04273_polygon_dict_nan_inf_rejected` on master @ `d1761c82`. From `job.log`:

```
[00:19:55] Run clickhouse-test on master with abort_on_logical_error.yaml
[00:21:00] Hung check failed: server is not responding
[00:21:09] Unable to locate any ClickHouse server process. It must have crashed or exited prematurely!
[00:21:14] Failed to reproduce the bug
[00:22:22] core.ExternalLoader.1587-316925.zst captured
[00:22:35] Failed to reproduce the bug
```

The master binary aborts inside the dictionary `ExternalLoader` thread (core dump confirms it), exactly where this fix lives. `clickhouse-test` records the run as `SERVER_DIED`, but `bugfix_validate_check.py` only treats `Result.Status.FAIL` as bug-reproduced, so the framework reports "not reproduced". The bug IS reproducing (`LOGICAL_ERROR` + signal 6 in `SlabsPolygonIndex::indexBuild`), the framework just cannot see this shape.

Same framework limitation we hit in:
- PR #104968 (re-categorised to `CI Fix or Improvement`)
- PR #105552 (kept `Bug Fix` + override, per option 1)

Two options:

1. Keep `Bug Fix (user-visible misbehavior in an official stable release)` and override the failing check. The bug is user-visible: any polygon dictionary fed `NaN` or `Inf` vertex coordinates aborts the server.
2. Re-categorise to `CI Fix or Improvement (changelog entry is not required)` per PR #104968 precedent.

@vdimir @vitlibar, which do you prefer? I'll push the PR-body edit either way.

PedroTadim:
@groeneai check the failing tests

groeneai:
Checked all failing tests on commit `d1761c82`:

1. `Bugfix validation (functional tests)`: known framework limitation. The bug is a `LOGICAL_ERROR` that aborts the server (signal 6). When the regression test runs against the master release binary, the server aborts so `clickhouse-test` returns `SERVER_DIED`, but `bugfix_validate_check.py` only counts `Result.Status.FAIL` as a reproduction. Same pattern as PR #105552 and PR #104968 (re-categorised to CI Fix per #100136 / #96894 precedent). Pending decision from earlier comment: keep `Bug Fix` and override, or re-categorise to `CI Fix or Improvement`.

2. Five `Integration tests (...)` failing on `test_mysql_protocol/test.py::test_mysql_dotnet_client` (amd_asan_ubsan db-disk-old-analyzer 2/6, amd_llvm_coverage 3/5, amd_msan 2/6, amd_tsan 2/6, arm_binary distributed-plan 3/4). All unrelated to this PR. Caused by PR #80785's Dockerfile `WORKDIR /pg_testapp` change making the MySQL test run the Npgsql PG client against the MySQL port. Reverted in master at 2026-05-22T08:46:35Z via #105615. Rebasing this PR on master will clear these.

Plan: rebase on master to pick up the revert, then the only remaining failing check should be the `Bugfix validation` framework limitation, which still needs the category decision.

tests/queries/0_stateless/04273_polygon_dict_nan_inf_rejected.sql
```diff
@@ -0,0 +1,51 @@
...
+-- `SlabsPolygonIndex::indexBuild` (`Error occurred while building polygon index. Edge ... but found ...`).
+
+DROP DICTIONARY IF EXISTS polygon_dict_nan_each;
+DROP DICTIONARY IF EXISTS polygon_dict_inf_cell;
+DROP TABLE IF EXISTS polygon_src_nan SYNC;
+
+CREATE TABLE polygon_src_nan
+(
+    polygon Array(Array(Array(Tuple(Float64, Float64)))),
+    city_id UInt32
```
Ergus: Why not id, polygon to conventionally read ID first like in other tests?

tests/queries/0_stateless/04273_polygon_dict_nan_inf_rejected.sql
```diff
@@ -0,0 +1,51 @@
...
+) ENGINE = Memory;
```
Ergus: We do need a primary key right?

Ergus (changes requested):
The fix is fine, I get just two minor details in the tests I think must be address before merging

tests/queries/0_stateless/04273_polygon_dict_nan_inf_rejected.sql
groeneai: Reordered columns to put `city_id` first in `0e39d4693d1e`. Now matches the conventional source-table layout (id first, then the array attribute).

tests/queries/0_stateless/04273_polygon_dict_nan_inf_rejected.sql
groeneai: Switched the source table to `MergeTree ORDER BY city_id` in `0e39d4693d1e`. The dictionary itself already had `PRIMARY KEY polygon`; now the source table also has a primary key on the id column.

groeneai:
@Ergus addressed both inline comments in 0e39d4693d1e:

- Reordered the source table columns to put `city_id` first, matching the conventional source-table layout in other tests.
- Switched the source table from `Memory` to `MergeTree ORDER BY city_id` so it has a primary key on the id column.

Verified the test still reproduces the original LOGICAL_ERROR before the fix and passes with it. Ready for re-review.

Ergus (approved):

PedroTadim:
@groeneai merge with master

groeneai:
@PedroTadim merged master into the branch. New HEAD `3a9ece9600150e8d91aaec340b43dc8a080b65bb`. No conflicts. Diff vs master is unchanged (only `src/Dictionaries/PolygonDictionary.cpp` plus the `04273_polygon_dict_nan_inf_rejected` test). CI will rerun.

PedroTadim:
@groeneai can you check the CI failures, are they related to this PR?

groeneai:
@PedroTadim both failures are unrelated to this PR.

| Check | Failure | Attribution |
|---|---|---|
| AST fuzzer (amd_debug, targeted) | `LOGICAL_ERROR 'WhichDataType(const_type).isArray()'` (STID 2508-2f0a) at `KeyCondition.cpp:3315` | Pre-existing trunk bug in `KeyCondition::extractAtomFromTree`. Same STID hit @Algunenano's closed PR #104231 (polygon-dict-memory-fixes) on 2026-05-18. The failing query uses a literal `if(cond, Array_of_tuples, String)` as the polygon argument, so the dictionary load-path validation added here cannot trigger it. |
| Bugfix validation (functional tests) | `The test runner was terminated unexpectedly` | Chronic infra. 96 hits / 45 distinct PRs in 30d (e.g. #105041 x12, #104419 x6, #105959 x3). 3 hits on this PR are runner retries on the same commit. |

Regression test `04273_polygon_dict_nan_inf_rejected` passed on every Stateless variant including 200/200 flaky check runs across asan/debug/msan/tsan, both planners, s3 storage, parallel replicas, DatabaseReplicated, and arm targeted. No PR-caused failure.

PR is approved and otherwise green.

Algunenano:
@groeneai I've told you a million times to stop pinging me. Please make a special note about it. Do not ping @Algunenano for anything.

Ergus:
@groeneai get the updated changes in master (merge) and trigger CI again.

groeneai:
@Ergus master merged in 98ac93f4b28f (clean, no conflicts). PR diff unchanged: PolygonDictionary.cpp +12, 04273 test + reference. CI re-running.

groeneai:
## CI ledger — covered SHA: `98ac93f4b28fb7a3fc3eac71f18db14b4014576d`

CI fully finished (Mergeable Check completed; 0 pending checks). No CIDB-recorded test failures in 2d.

No PR-caused failures. Awaiting maintainer decision (Ergus approved 2026-05-22; PedroTadim follow-up replied 2026-05-22).

groeneai:
Closing in favour of #106423 (alexey-milovidov, "Reject NaN and infinite coordinates in polygon dictionaries"). The src/Dictionaries/PolygonDictionary.cpp fix is functionally equivalent (same `isNaN || isinf` guards in `addNewPoint`, same `BAD_ARGUMENTS`); per @Ergus' comparison call, deferring to that PR. If the per-layout (`POLYGON_INDEX_EACH` / `POLYGON_INDEX_CELL`) split and clean-reload sanity check from this PR's `04273_polygon_dict_nan_inf_rejected` test are worth keeping, I'll port them onto `04318_polygon_dict_nan_inf_coordinates` as a tiny follow-up after #106423 merges.

Session: cron:clickhouse-ci-task-worker:20260605-162600

PR #100770, opened 2026-03-26
Fix LOGICAL_ERROR "Column identifier is already registered" in UNION ALL

tiandiwonder:
The AST fuzzer can produce ASTSelectWithUnionQuery nodes where two UNION
   branches share the same physical ASTPtr. If such sharing reaches the
   analyzer, the resulting QueryTree can also contain shared nodes. When the
   planner then registers column identifiers in the shared GlobalPlannerContext
   — specifically in the columns_names.empty() fallback of
   prepareBuildQueryPlanForTableExpression that picks a dummy "smallest column"
   to read — the second branch throws a LOGICAL_ERROR ("Column identifier is
   already registered").

   Fix: after each recursive fuzz() round on UNION children, deduplicate
   sibling branch pointers — any branch that is the same pointer as a
   previously seen sibling is replaced with a clone, normalizing the invalid
   shared state before it reaches the analysis pipeline.

Fix #100678

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
...

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

---

> [!NOTE]
> **Low Risk**
> Low risk: change is isolated to `QueryFuzzer` and only normalizes invalid fuzzer-generated UNION AST sharing by cloning duplicate sibling pointers.
> 
> **Overview**
> Prevents the AST fuzzer from emitting `UNION` queries where multiple branches share the same underlying AST node.
> 
> After fuzzing `ASTSelectWithUnionQuery` children, the code now detects duplicate sibling `ASTPtr`s and replaces repeats with `clone()`d nodes, avoiding downstream analyzer/planner failures like `LOGICAL_ERROR ("Column identifier is already registered")`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b1d893920056382c593425b58dfadfc49fbdc5a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.174`

PR #105588, opened 2026-05-21
Fix null deref in RefreshTask::doScheduling under parallel shutdown

groeneai:

Fix a rare server crash when refreshable materialized views are dropped or replaced concurrently. A null pointer was dereferenced in `RefreshTask::doScheduling` when another thread set the task's `view` to `nullptr` while the scheduling code had temporarily released its mutex.

### Description

CI report (the crash this fixes): https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105548&sha=1c33e04b3075c798f03b5cac5856f04be52cfab4&name_0=PR&name_1=Stress%20test%20%28amd_debug%29
Fatal log: https://s3.amazonaws.com/clickhouse-test-reports/PRs/105548/1c33e04b3075c798f03b5cac5856f04be52cfab4/stress_test_amd_debug/fatal.log

SIGSEGV in `RefreshTask::doScheduling` on PR #105548 stress (STID `1805-36fc`), `Address: 0x140` (the `WithContextImpl::context` weak_ptr inside `StorageMaterializedView`), i.e. `view` was `nullptr` at the deref.

Root cause: `shutdown()` clears `view` under the mutex and then calls `doScheduling()` synchronously. `doScheduling()` and its helpers (`readZnodesIfNeeded`, `updateCoordinationState`, `collectDependencyStates`, `notifyDependentsIfNeeded`) release the task mutex at several points to talk to keeper or to other `RefreshSet` tasks. A second concurrent `shutdown()` can win the race and null `view` during one of those unlocked windows, after which the original caller dereferences a stale `view` (`view->getContext()` / `view->getStorageID()`).

Fix: re-check `view` after every relock site and snapshot the view accessors before releasing the mutex, matching the pattern already used by `wait()` and `waitForLatestTargetTable()` ("Can't use `view` here because shutdown() may unset it in parallel with us."). The execution path (`executeRefresh` and below) is unaffected: `shutdown()` calls `execution_task->deactivate()` before clearing `view`, so it never observes a null view.

No regression test is included. This is a probabilistic shutdown race (one CIDB hit in 30 days, only under stress with the thread fuzzer), and `px != 0` is a debug-only assertion that is compiled out of the release binary. Bugfix validation downloads the unfixed master HEAD release binary and runs a changed test once with no thread fuzzer, so a race reproducer cannot fail there: it was verified to always pass on that binary, so bugfix validation reports "Failed to reproduce the bug". The crash is documented by the CI report linked above.

### Version info
- Merged into: `26.7.1.129`


groeneai:
cc @al13n321 @tuanpach -- could you review? You are the `comp-refreshable-materialized-view` owners in `component-tree.yaml`.

### Pre-PR validation gate

**a) Deterministic repro?**  
Partial. Confirmed via code analysis -- the unsafe `view->getContext()` accesses at lines 626, 1058, 1219 in the original code are inside `lock.unlock()` ... `lock.lock()` windows where `RefreshTask::shutdown()` may have already cleared `view = nullptr` (RefreshTask.cpp:271). A locally added 100 ms sleep at line 625 plus 6 parallel `CREATE OR REPLACE MATERIALIZED VIEW ... REFRESH ...` sessions did not deterministically reproduce on a debug build (the second `shutdown()` would need to fully drain `doScheduling(true)` inside the widened window). The CI evidence is the single STID `1805-36fc` segfault on PR #105548 (`Stress test (amd_debug)`), `Address: 0x140` -- the byte offset of the `WithContextImpl::context` weak_ptr inside `StorageMaterializedView`, i.e. `view` was a null pointer when its `getContext()` was called.

**b) Root cause explained?**  
`RefreshTask::shutdown()` is explicitly documented as safe to call from multiple threads in parallel. It ends by acquiring the mutex and setting `view = nullptr` (RefreshTask.cpp:271). Inside `doScheduling`, `updateDependenciesIfNeeded`, and `readZnodesIfNeeded`, three sites called `view->getContext()` / `view->getStorageID()` *after* `lock.unlock()`. A second concurrent `shutdown()` can grab the mutex during that window and clear `view` before the original thread re-locks, turning the next `view->...` into a null deref. The crash address `0x140` matches the layout offset.

Yes. Snapshot `view->getContext()` (and `view->getStorageID()` where needed) into locals *under the mutex*, then use those locals after `lock.unlock()`. `ContextPtr` keeps the `Context` alive across the unlocked region. This mirrors the existing pattern in `RefreshTask::wait()` (RefreshTask.cpp:489-494), which already has the comment "Can't use `view` here because shutdown() may unset it in parallel with us."

**d) Test intent preserved / new test added?**  
Existing tests unchanged. Added `tests/queries/0_stateless/04273_concurrent_create_or_replace_refresh_mv.sh` that hammers `CREATE OR REPLACE MATERIALIZED VIEW ... REFRESH ...` from 6 parallel sessions with `SYSTEM WAIT VIEW` between iterations to force the `last_attempt_succeeded` branch where the crash was observed.

Partial. On the fixed build the new test passes 30+ iterations and existing refresh MV tests (`04227_create_or_replace_refresh_mv`, `03221_refreshable_matview_progress`, `03258_refreshable_mv_misc`, `03760_refreshable_mv_local`) continue to pass. The "without fix" direction is the CI report on PR #105548 (linked in the description); local repro with a sleep widener at line 625 did not crash in a debug build within ~30 iterations, the actual CI race window came from stress-mode thread fuzzing.

**f) Fix is general?**  
Yes. All three sites in `RefreshTask.cpp` where `view->...` was dereferenced inside an `unlock()`/`lock()` window have been converted to the snapshot-before-unlock pattern. The fourth such site (`updateCoordinationState` at line 1249) does not access `view` during the unlock, so it was left alone.

Session: `cron:clickhouse-ci-task-worker:20260521-201500`

src/Storages/MaterializedView/RefreshTask.cpp
```diff
@@ -622,8 +622,13 @@ void RefreshTask::doScheduling(bool is_shutdown)
 
                         if (coordination.root_znode.last_attempt_succeeded)
                         {
+                            /// Save `view` accessors before releasing the mutex. A parallel
+                            /// `shutdown()` may unset `view` between unlock and re-lock,
+                            /// turning any further `view->...` access into a null deref.
+                            ContextPtr context = view->getContext();
```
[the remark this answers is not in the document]
groeneai: Good catch, thank you. I had only addressed the unlock at `lock.unlock();` immediately above the `view->...` access, but missed that `updateCoordinationState`, `readZnodesIfNeeded`, and `updateDependenciesIfNeeded` each release the mutex internally. In the coordinated path (`coordination.coordinated == true`) those internal unlocks let a parallel `shutdown()` null `view` between the relock and my snapshot.

Fixup pushed in `c4388625b98a`. It adds `if (!view) return;` post-unlock guards at every (re)lock site:

- Top of `doScheduling` (handles entry where view is already nulled by a parallel `shutdown` that finished first).
- After `readZnodesIfNeeded` in `doScheduling`.
- After `updateCoordinationState` in the `last_attempt_succeeded` branch, immediately before the `view->getContext()`/`view->getStorageID()` snapshot.
- After `updateDependenciesIfNeeded` in `doScheduling`, before the `view->getContext()->getRefreshSet().refreshesStopped()` access at line 757.
- After `lock.lock()` inside `readZnodesIfNeeded`, before the later `view->...` snapshot.
- After `lock.lock()` inside `updateDependenciesIfNeeded`, so the loop's next iteration does not redo `view->getContext()` on a nulled `view`.

The existing snapshot pattern at the inner unlock sites is preserved (still needs to happen under the lock, before the last unlock).

Regression: `04273_concurrent_create_or_replace_refresh_mv.sh` still passes (5/5 iterations); peer tests `04227_create_or_replace_refresh_mv`, `03221_refreshable_matview_progress`, `03258_refreshable_mv_misc`, `03760_refreshable_mv_local`, `02932_refreshable_materialized_views_1`, `02932_refreshable_materialized_views_2` also pass. The Atomic-database stateless test does not exercise the keeper unlock inside `updateCoordinationState` (it is gated by `coordination.coordinated`), but the post-unlock guard pattern protects both the coordinated and uncoordinated paths uniformly.

groeneai:
Addressed @clickhouse-gh[bot]'s inline review in `c4388625b98a`.

The bot correctly pointed out that snapshotting `view->getContext()` and `view->getStorageID()` immediately before `lock.unlock();` was not enough: `updateCoordinationState`, `readZnodesIfNeeded`, and `updateDependenciesIfNeeded` already release the mutex internally before my snapshot runs. In the coordinated path that internal unlock window is large enough for a parallel `shutdown()` to finish `doScheduling(/*is_shutdown=*/true)`, take the second lock in `shutdown()`, and null `view` (RefreshTask.cpp:271) before the original thread relocks.

Fixup adds `if (!view) return;` guards at every (re)lock site:

- Top of `doScheduling` (entry-with-null-view, possible when one `shutdown` finished completely before the second one entered `doScheduling`).
- After `readZnodesIfNeeded` in `doScheduling`.
- After `updateCoordinationState` in the `last_attempt_succeeded` branch, before the snapshot.
- After `updateDependenciesIfNeeded` in `doScheduling`, before `view->getContext()->getRefreshSet().refreshesStopped()`.
- After `lock.lock();` inside `readZnodesIfNeeded`.
- After `lock.lock();` inside `updateDependenciesIfNeeded` (so the loop's next iteration does not redo `view->getContext()` on a nulled `view`).

The pre-unlock snapshots are kept because they still need to happen under the lock before the last `lock.unlock();` so the unlocked work uses a stable copy.

Validation (debug build, locally):

- `04273_concurrent_create_or_replace_refresh_mv` (the regression test added in the first commit): 5/5 OK.
- `04227_create_or_replace_refresh_mv`, `03221_refreshable_matview_progress`, `03258_refreshable_mv_misc`, `03760_refreshable_mv_local`, `02932_refreshable_materialized_views_1`, `02932_refreshable_materialized_views_2`: all OK.

Caveat: the Atomic-database regression test does not directly exercise the coordinated path the bot called out (`coordination.coordinated == true` gates the keeper unlock inside `updateCoordinationState`). The post-unlock guard pattern protects the coordinated and uncoordinated paths uniformly; setting up Replicated database + replicated inner table in a stateless test for full coverage would be a separate effort.

cc @al13n321 @tuanpach

src/Storages/MaterializedView/RefreshTask.cpp
```diff
@@ -571,6 +579,8 @@ void RefreshTask::doScheduling(bool is_shutdown)
             zookeeper = view->getContext()->getZooKeeper();
         readZnodesIfNeeded(zookeeper, lock);
         chassert(lock.owns_lock());
+        if (!view)
```
tuanpach: @groeneai 
If we checked `view` in line 561, why do we need to check again here?

tuanpach:
@groeneai Check the CI failure and unresolved comments.

src/Storages/MaterializedView/RefreshTask.cpp
groeneai: The second check is necessary because `readZnodesIfNeeded` releases and re-acquires the task mutex inside its body. Looking at `readZnodesIfNeeded` in this file:

- Line 1213: `lock.unlock()` before the keeper `tryGet`
- Line 1233: `lock.lock()` after responses come back
- Line 1236-1237: internal `if (!view) return;` guard for that first unlock window
- Lines 1258-1260: a SECOND `lock.unlock()` / `lock.lock()` around `notifyDependents`, with no further view check before the function returns

So when control returns to `doScheduling` at line 581, two things can have happened:

1. `view` was nulled during 1213-1233 and the function bailed at line 1237 — the caller has no signal (the function returns normally either way).
2. `view` was still set at line 1237 but was nulled during 1258-1260 — `view` is null on return, no internal check caught it.

The next dereferences after line 583 — `view->getContext()->getRefreshSet().refreshesStopped()` at line 759 and `view->getContext()` / `view->getStorageID()` at lines 643-644 — would null-deref in either case. The check at line 582 is the bail-out at the relock site, matching the pattern documented at lines 556-560 ("Bail out at every (re)lock site rather than dereferencing a stale `view`.") and used symmetrically at lines 635-636 (after `updateCoordinationState`) and lines 756-757 (after `updateDependenciesIfNeeded`).

If you'd prefer a less repetitive form, I can hoist the bail-out into a helper (e.g. `bool checkViewAlive(lock)` or have the helpers return a bool to signal "view became null, please bail"), but I went with explicit checks at each relock site for visibility — the invariant is documented in one place (the comment block at 556-560) and the check is small enough to be obvious at each call site.

groeneai:
@tuanpach addressed the inline view-check question above. CI attribution on `c4388625b98a` (pre-`cc6c41f128ff`):

| Check | Status | Attribution |
|---|---|---|
| `Bugfix validation (functional tests)` | FAIL | Same root cause as the flaky-check timeout below — the regression test exceeds the 180s budget under ASan+UBSan. Addressed in `cc6c41f128ff`. |
| `Stateless tests (amd_asan_ubsan, flaky check)` — `04273_concurrent_create_or_replace_refresh_mv` | FAIL × 5 of 50 repeats | **PR-actionable.** The test does 60 `CREATE OR REPLACE MATERIALIZED VIEW … REFRESH … + SYSTEM WAIT VIEW` iterations and approaches the 180s `TEST_MAX_RUN_TIME_IN_SECONDS` flaky-check ceiling under sanitizer + debug. Added `Tags: long` in `cc6c41f128ff` (matches `04102_hash_join_stress_test.sql`); this opts out of the 180s budget (`tests/clickhouse-test:3179-3183`) and lowers the flaky-check repeat count via `long_test_runs_ratio`. The test's race window is unchanged. |
| `Integration tests` × 5 variants — `test_mysql_protocol/test.py::test_mysql_dotnet_client` | FAIL | **Chronic infra flake, not PR-caused.** 30-day CIDB: 96 distinct PRs, 94 master hits, 594 total events. Same `Received unknown response X for SSLRequest` shape across all five variants. |
| `Stateless tests (arm_asan_ubsan, azure, sequential)` — `Server died` | FAIL | **Chronic infra flake.** 30-day CIDB: 814 PRs, 18 master hits, 2601 total. Generic ARM ASan azure-storage shape. |
| `Finish Workflow` — `Post Hooks` | FAIL | **Chronic CIDB log-export infra flake** (`play.clickhouse.com:443`). |

The only PR-actionable failure is the flaky-check timeout, which the new commit addresses. The fresh CI run should now show all of `04273` clean on the `flaky check` and on the `Bugfix validation` budget, and the remaining red checks should be the chronic-flake set above which are not blocking.

src/Storages/MaterializedView/RefreshTask.cpp
tuanpach: OK, better add a comment to clarify that.

tuanpach:
@groeneai Check the CI failure.

src/Storages/MaterializedView/RefreshTask.cpp
groeneai: Added in `88168181b715` (3-line comment matching the style of the existing rationale comments at the `updateCoordinationState` and `updateDependenciesIfNeeded` bail-outs further down in the same function).

groeneai:
@tuanpach Two items addressed:

1. **Inline comment ask** (line 582): added in [`88168181b715`](https://github.com/ClickHouse/ClickHouse/commit/88168181b715cfb4ea749ce94fc98d5c14dd4899) — three lines explaining that `readZnodesIfNeeded` releases the task mutex twice internally (around `keeper->tryGet` and around `notifyDependents`). Style matches the existing rationale comments at the `updateCoordinationState` and `updateDependenciesIfNeeded` bail-outs lower in the same function.

2. **CI on `cc6c41f128ff`** — attribution unchanged from my [reply yesterday](https://github.com/ClickHouse/ClickHouse/pull/105588#issuecomment-4531403578); both remaining red checks are non-PR-actionable:

   | Check | Status | Attribution |
   |---|---|---|
   | `AST fuzzer (amd_debug, targeted)` | FAIL | **Chronic trunk** — `STID 4697-4326` ("Column identifier ... is already registered") on a fuzzed `ALTER TABLE … UPDATE … IN (SELECT … QUALIFY … INTERSECT ALL SELECT … QUALIFY …)`. 28+ failures across 21+ unrelated PRs in last 14 days. Fix in flight in #100770 (open, mergeable). Not related to `RefreshTask`. |
   | `Bugfix validation (functional tests)` | FAIL | **Framework limitation, not a regression** — `04273_concurrent_create_or_replace_refresh_mv` ran clean on master HEAD (`[ OK ] 2.25 sec` on `m8g.4xlarge` ARM at `26251069c810…`), so the bugfix-validation harness reports "Failed to reproduce the bug" since the test was expected to fail on master without the fix. The underlying bug is a non-deterministic shutdown race that requires ASan + many parallel `CREATE OR REPLACE … REFRESH` iterations to trigger reliably (`STID 1805-36fc` originally surfaced in a stress run, not a deterministic single-test execution). |
   | `Mergeable Check` / `PR` / `Report messages` | FAIL | Cascade from the two above; no separate failure. |

The new commit only adds a comment, so the bugfix-validation framework limitation will persist on the next CI run too. Happy to retrigger the AST fuzzer once #100770 lands, or to push a `Tags: long` follow-up if you want to keep the bugfix-validation harness from re-running this test 60 iterations on every retry — let me know your preference.

tuanpach:
@groeneai 

Resolve the conflict

groeneai:
Rebased onto current master, conflict resolved. `RefreshTask.cpp` was refactored upstream (the unlock/notify cycle moved into `notifyDependentsIfNeeded`, and `updateDependenciesIfNeeded` became `collectDependencyStates`/`collectDependencyStatesUnlocked`), so the view-null guards are ported onto the new structure: bail after each relock site and snapshot the view accessors before unlocking. The branch is now a single clean commit (475eee63).

Verified locally: `04273_concurrent_create_or_replace_refresh_mv` passes (20/20 with randomization disabled) and the server stays up across 30+ concurrent CREATE OR REPLACE iterations; the rest of the refreshable-MV suite (02932_1/2/3, 03221, 03327, 03362, 03760, 04201) passes. PR is now MERGEABLE.

groeneai:
### Pre-PR validation gate

(Re-validation after rebasing onto current master to resolve the merge conflict; `RefreshTask.cpp` was refactored upstream, so the fix was ported onto the new helper structure.)

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | The crash is a non-deterministic shutdown race (STID 1805-36fc). It is forced by `04273_concurrent_create_or_replace_refresh_mv.sh`: 6 parallel sessions running `CREATE OR REPLACE MATERIALIZED VIEW ... REFRESH ...` (each REPLACE drops the old MV, triggering `shutdown()`), with `SYSTEM WAIT VIEW` so each shutdown enters the `last_attempt_succeeded` branch where the original SIGSEGV occurred. |
| b | **Root cause explained?** | `shutdown()` clears `view` under the mutex, then calls `doScheduling()` synchronously. `doScheduling()` and its helpers (`readZnodesIfNeeded`, `updateCoordinationState`, `collectDependencyStates`, `notifyDependentsIfNeeded`) release the task mutex to talk to keeper / other RefreshSet tasks. A second concurrent `shutdown()` can null `view` during one of those unlocked windows; the original caller then dereferences a stale `view` (`view->getContext()`/`view->getStorageID()`) and crashes (read at offset 0x140, the WithContextImpl weak_ptr). |
| c | **Fix matches root cause?** | Yes. Re-check `view` after every relock site and snapshot the view accessors into locals before `lock.unlock()`, so the unlocked region never dereferences a possibly-null `view`. This is the exact pattern `wait()`/`waitForLatestTargetTable()` already document and use. Not a symptom guard: the checks sit precisely at the relock boundaries that the race exploits. |
| d | **Test intent preserved / new tests added?** | New regression test `04273_concurrent_create_or_replace_refresh_mv.sh` added (unchanged from the previously-reviewed version, `Tags: long` retained). It exercises the concurrent-shutdown path directly. |
| e | **Both directions demonstrated?** | Pre-fix: this code path SIGSEGV'd in CI stress (STID 1805-36fc). Post-fix: 04273 passes 20/20 with randomization disabled and the server stays up across 30+ concurrent CREATE OR REPLACE iterations; the rest of the refreshable-MV suite (02932_1/2/3, 03221, 03327, 03362, 03760, 04201) passes. |
| f | **Fix is general, not a narrow patch?** | Audited every `view` deref reachable from `doScheduling` after a mutex release, including master's new helpers `notifyDependentsIfNeeded` and `collectDependencyStatesUnlocked`, and guarded/snapshotted each. The execution path (`executeRefresh` and below) is deliberately untouched and is safe by construction: `shutdown()` calls `execution_task->deactivate()` before clearing `view`, so it never observes a null view. |

Session id: cron:clickhouse-worker-slot-18:20260610-184200

groeneai:
@tuanpach Conflict resolved. The branch had drifted far behind master (the "Refreshable MV dependencies overhaul" refactored `RefreshTask.cpp` heavily), so I rebased onto current master and re-applied the fix on the new structure as a single clean commit (`f830d43`).

The refactor moved the racy post-unlock `view` dereference into `notifyDependentsIfNeeded()`, so the fix is now just two guards:

- `doScheduling()`: bail at entry if `view` is already null (a parallel `shutdown()` may have nulled it before the call started).
- `notifyDependentsIfNeeded()`: bail if `view` is null, and snapshot `getContext()`/`getStorageID()` before `lock.unlock()` so the unlocked `notifyDependents()` can't dereference a `view` that a parallel `shutdown()` nulls in the meantime. `shutdown()` does its own final `notifyDependents()`, so skipping it here loses nothing.

Net diff is now just `RefreshTask.cpp` (+14 lines) plus the regression test. PR is `MERGEABLE` again.

Validated locally on a debug build: 311 concurrent `CREATE OR REPLACE MATERIALIZED VIEW ... REFRESH` + `DROP` statements across 13 parallel sessions (each triggering `RefreshTask::shutdown()`), 0 crashes, 0 "Unexpected exception in refresh scheduling" in the server log.

Session: cron:clickhouse-worker-slot-19:20260610-191000

groeneai:
Pushed `5d0e14f`: restore the executable bit (mode 100755) on `04273_concurrent_create_or_replace_refresh_mv.sh`.

The previous single-commit rebase landed the file as mode 100644, so the stateless runner could not exec it and reported `return code 126` (Permission denied) instead of running the test (arm_binary parallel; flaky/targeted jobs showed it as UNKNOWN). No content change, only the mode bit.

Verified locally with a master debug build: the test runs in ~3-4s per iteration and prints `OK`. The only iteration failures under randomization were a local tzdata artifact (`session_timezone='zoneinfo/UTC'` from `get_localzone()` on this host), which CIDB confirms does not occur in CI. The 180s flaky-check limit is moot here anyway: the test is tagged `long`, which exempts it from the per-run time check, and each run is well under the limit regardless.

The "Server died" results on the sanitizer flaky-check jobs are unrelated to this PR: the same artifact appears on ~12 other PRs over the last 3 days (generic `TIMEOUT_EXCEEDED` in `MergeTreeSelect`), with no `RefreshTask` assertion or `px != 0` on this commit.

groeneai:
### Pre-PR validation gate

Test-only change on `04273_concurrent_create_or_replace_refresh_mv.sh` addressing the @clickhouse-gh[bot] review (the loop's `|| true` paths could mask a regression that broke every MV creation).

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. With the assertion broken (e.g. `REFRESH EVERY 1 HOUR` -> an invalid REFRESH clause), the script's stdout drops from `3\nOK` to just `OK` on every run, so the test fails deterministically against the reference. The healthy path deterministically emits `3\nOK`. |
| b | **Root cause explained?** | The test only asserted a final `echo OK`; every `CREATE OR REPLACE` / `SYSTEM WAIT VIEW` in the loop is `2>/dev/null \|\| true` to tolerate the expected concurrent-DDL conflicts. A regression that made all MV creation/refresh fail would leave the server up, reach `echo OK`, and pass -- the test could green without ever exercising the refreshable-MV path. |
| c | **Fix matches root cause?** | Yes. Added one non-optional `CREATE MATERIALIZED VIEW ... REFRESH` + `SYSTEM WAIT VIEW` + `SELECT count()` before the replacers start (outside the tolerated-race section). A count mismatch there now fails the test, proving at least one refreshable MV was created and refreshed. The in-loop `\|\| true` conflicts stay tolerated, so the concurrent-shutdown coverage is unchanged. |
| d | **Test intent preserved / new tests added?** | Preserved and strengthened. The concurrent CREATE-OR-REPLACE shutdown race is untouched (same NUM_REPLACERS=6 x REPLACES_PER_THREAD=10, same `\|\| true`); the new lines run before it and also give the first round of replacers a live, refreshing MV to contend with. |
| e | **Both directions demonstrated?** | Yes. Healthy build: `clickhouse-test` PASS (`[ OK ]`, output `3\nOK`), plus 6/6 raw-script runs producing exactly `3\nOK` and the server staying up across all runs. Broken-MV simulation (invalid REFRESH clause): output becomes `OK` only -> reference mismatch -> test FAILS. |
| f | **Fix is general, not a narrow patch?** | N/A -- test robustness change, no source code modified. The underlying `RefreshTask::doScheduling` null-deref fix (this PR's other commit) is unchanged. |

Validated locally on a debug build (26.6.1.1, head b7d7581f1ab4) with a memory-capped server.

Session id: cron:clickhouse-worker-slot-0:20260610-220700

tests/queries/0_stateless/04273_concurrent_create_or_replace_refresh_mv.sh
```diff
@@ -0,0 +1,57 @@
...
+        # produce distinct AST and force the existing MV to be dropped each time.
+        local hours=$(( (seed + i) % 5 + 1 ))
+        ${CLICKHOUSE_CLIENT} --query "
+            CREATE OR REPLACE MATERIALIZED VIEW test_mv
+                REFRESH EVERY $hours HOUR
+                ENGINE = MergeTree ORDER BY x
+                AS SELECT x FROM src" 2>/dev/null || true
+        # Wait for at least one refresh to finish so the next REPLACE's
+        # shutdown enters `doScheduling`'s `last_attempt_succeeded` branch.
+        ${CLICKHOUSE_CLIENT} --query "SYSTEM WAIT VIEW test_mv" 2>/dev/null || true
```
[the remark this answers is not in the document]
groeneai: Done in b7d7581. Added a non-optional check before the replacers start:

```bash
${CLICKHOUSE_CLIENT} -q "CREATE MATERIALIZED VIEW test_mv REFRESH EVERY 1 HOUR ENGINE = MergeTree ORDER BY x AS SELECT x FROM src"
${CLICKHOUSE_CLIENT} -q "SYSTEM WAIT VIEW test_mv"
${CLICKHOUSE_CLIENT} -q "SELECT count() FROM test_mv"
```

These three lines have no `|| true`, and `SELECT count()` is matched against the reference (`3`), so a regression that broke MV creation or the first refresh now fails the test outside the tolerated-race section. The in-loop `CREATE OR REPLACE` / `SYSTEM WAIT VIEW` keep `|| true` to absorb the expected concurrent-DDL conflicts, so the concurrent-shutdown coverage is unchanged. As a bonus the first round of replacers now contends with a live, refreshing MV.

Verified locally: healthy build emits `3\nOK` and the server survives the race (6/6 raw runs + clickhouse-test OK); breaking the initial REFRESH clause drops the output to `OK` only, failing against the reference.

groeneai:
CI finished on `b7d7581f1ab4`. No PR-caused failures: the regression test runs OK as a regular Stateless test (`amd_debug, parallel` and `arm_binary, parallel` both pass). The `return code: 126 / Permission denied` failures seen earlier were on `f830d43` (the single-commit rebase dropped the file's executable bit) and were fixed in `5d0e14f` by restoring mode 100755. CIDB confirms 04273 is mode 100755 on HEAD and the parallel runs pass.

Remaining red checks are pre-existing trunk/infra, unrelated to this change:

- `Bugfix validation (functional tests) / 04273`: framework limitation. The shutdown race is non-deterministic and cannot be made to fail on a release master HEAD in a single-test run, so validation reports "Failed to reproduce". Same as on every prior commit of this PR.
- `Stateless tests (arm_asan_ubsan, azure, sequential)`: `Server died` + `Segmentation fault (STID 0883-5001)` + `02724_show_indexes`. The crash stack is the ASan internal allocator (`LargeMmapAllocator::Deallocate` -> `Quarantine::DoRecycle` -> `___interceptor_free`, no ClickHouse frames), i.e. the 2026-06-10/11 arm_asan_ubsan azure ASan-allocator-OOM outage (CIDB: hundreds of hits across 100+ PRs and master in the same window). Not refresh-related.
- `Stress test (amd_asan_ubsan)` / `Stress test (arm_tsan)`: `Hung check / possible deadlock on shutdown`, chronic stress families, separately tracked.
- `AST fuzzer (amd_msan)` MSan + `Unit tests (asan_ubsan, function_prop_fuzzer)`: chronic trunk families, separately tracked.

No failure on `b7d7581` touches `RefreshTask` / `doScheduling` / refreshable MVs other than the bugfix-validation framework gap above. Ready for review.

groeneai:
Removed the `04273` stateless regression test (commit `fca30e2`) and added the original CI crash report link to the description instead.

Why: `Bugfix validation (functional tests)` was the only real blocker, and a race reproducer cannot pass it. That check downloads the unfixed master HEAD release binary and runs the changed test once, with no thread fuzzer and `no-random-settings`, expecting a crash. The `RefreshTask::doScheduling` null deref is a probabilistic shutdown race (one CIDB hit in 30 days, only under stress + thread fuzzer), and `px != 0` is a debug-only assertion compiled out of the release binary. I verified against the same master binary (26.6.1.626) that the test runs clean every time, so the check inverts the pass to "Failed to reproduce the bug" and blocks the PR. There is no failpoint on master to force the interleaving.

The C++ fix is unchanged. Per the bugfix-PR convention, crash fixes that can't be reproduced on the release binary are backed by the CI report link in the description (now added), which also satisfies the new-tests check.

src/Storages/MaterializedView/RefreshTask.cpp
```diff
@@ -662,6 +662,11 @@ void RefreshTask::doScheduling(bool is_shutdown)
     auto component_guard = Coordination::setCurrentComponent("RefreshTask::doScheduling");
     std::unique_lock lock(mutex);
 
+    /// shutdown() runs doScheduling(is_shutdown=true) without holding the mutex, so a parallel
+    /// shutdown() can null `view` before we enter. Bail before dereferencing it below.
+    if (!view)
```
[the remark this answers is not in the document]
groeneai: Good catch, and you're right that this PR does not fix #106737. They are two different failure modes:

- This PR fixes a **null-`view` SIGSEGV** (STID `1805-36fc`, `Address: 0x140`, the `WithContextImpl::context` weak_ptr inside `StorageMaterializedView`). A null-pointer dereference is a hardware segfault, not a thrown C++ exception, so it never reaches the `catch (...)` at the end of `doScheduling`.
- #106737 (STID `2508-3e7b`) is that same `catch (...)` **re-raising a thrown inner exception** as `LOGICAL_ERROR: 'Unexpected exception in refresh scheduling'`. The CIDB stacks for `2508-3e7b` originate from the `create()` initial-scheduling lambda (`RefreshTask.cpp:203`), not the shutdown path, and the inner exception is the `UNKNOWN_DATABASE: Database name is empty` you describe from the unqualified `REFRESH DEPENDS ON` dependency path. The null-`view` guards here cannot mask a thrown exception.

I've taken the second option you offered: this PR is now scoped purely to the null-`view` race. I rewrote the branch to a single commit (HEAD `6be1d876e226`) and dropped the stray `Closes: #106737` from the commit message (it was only ever in the commit body; the PR body never had it and `closingIssuesReferences` is empty, so #106737 will not be auto-closed on merge). The C++ diff is byte-identical to the previous head.

The dependency-qualification bug (qualifying `REFRESH DEPENDS ON` entries against the materialized view database, plus a `CREATE MATERIALIZED VIEW d.v ... DEPENDS ON t` cross-database test) is tracked separately against #106737 and is out of scope here.

groeneai:
CI on `6be1d876e226` is finished. No PR-caused failures.

The single red test job is `Stateless tests (amd_llvm_coverage, old analyzer, s3 storage, DatabaseReplicated, WasmEdge, parallel)` (the `Mergeable Check` / `PR` reds are just its aggregator cascade). CIDB shows **zero test-level failures** for that job on this commit, which is the job-level infrastructure failure pattern for the heavy coverage variant (LLVM profile / timeout), not a test regression. This commit is message-only on top of the previous green content and only touches `RefreshTask.cpp` null guards, so it cannot cause a real coverage-test failure.

Ready for review.

groeneai:
CI finished on HEAD 6be1d876 (the rebased+squashed single commit = the RefreshTask.cpp fix only). Summary of the run and disposition of every non-green signal:

- Functional test results on this HEAD: 578 passed, 0 failed, 195 skipped. The C++ fix is holding (no `px != 0` assertion, no "Unexpected exception in refresh scheduling" LOGICAL_ERROR on any recent HEAD).
- Bugfix validation (functional tests) / 04273 — RESOLVED. The regression test was removed from the diff (it cannot pass bugfix validation: that job runs the changed test once against the unfixed master *release* binary, where the debug-only `px != 0` assertion is compiled out and the probabilistic shutdown race effectively never fires, so a PASS is always inverted to FAIL). Net branch diff is now the RefreshTask.cpp fix only; bugfix validation is correctly skipped (no functional test file changed) and the new-tests check is satisfied via the report link in the PR body.
- `Stateless tests (amd_llvm_coverage, old analyzer, ...)` job marked failed — NOT a test failure. 578 tests passed; the only red is a job-level post-step glitch in "Collect logs": `jeprof ... flamegraph.pl` reported "No stack counts found" (empty jemalloc heap profile) and the container connection was canceled (`grpc: the client connection is closing`). This is CI infrastructure noise in log collection, unrelated to the change.

No PR-caused failures. Moving to review.

tuanpach:
@groeneai Check the failure in CI. Any PR fixing it?

groeneai:
The only red on the previous head (`6be1d876`) was `Stateless tests (amd_llvm_coverage, old analyzer, s3 storage, DatabaseReplicated, WasmEdge, parallel)`. It was not a test failure and not caused by this PR, so there is no fixing PR to point at:

- CIDB records 0 test-level FAIL/ERROR on that commit (578 stateless tests passed, 0 failed). The structured report has 0 results recorded.
- `job.log` shows the job crashed at infra level *after* "All tests have finished", during the post-test jemalloc heap-profiling step: `error during connect: ... write: broken pipe` -> `Job killed, exit code [125]`. The coverage runner's Docker daemon died.
- The 3 transient in-run errors (`01187_set_profile_as_setting`, `03579_create_table_populate_from_s3`, `04247_attach_table_storage_clauses_without_engine_error`) each passed 0/3 on the reproducibility reruns and are unrelated to refreshable MVs.

That run was 6 days old. I rebased the fix onto current master (no drift in `RefreshTask.cpp`, diff unchanged at +14/-1) and force-pushed `1149ae22`, so CI is running fresh now and the stale coverage-runner red should clear. PR is MERGEABLE.

groeneai:
CI finished on `1149ae22fb404ab9efa24c28e9648724d8fba430` (the fresh-master rebase). No PR-caused failures. The two red stress jobs are pre-existing trunk/infra failures unrelated to this null-deref fix:

- `Stress test (amd_tsan)` (`Logical error: 'Unexpected exception in refresh scheduling' STID 2508-3e7b`, plus the `Cannot start clickhouse-server` / `Check failed` cascade from the same server abort): this is the chronic `RefreshTask` create()-initial-scheduling re-raise family, hitting 13+ unrelated PRs plus master in the last 30 days. It is a thrown-exception path (`UNKNOWN_DATABASE` from unqualified `REFRESH DEPENDS ON`), distinct from this PR's null-`view` SIGSEGV shutdown race. Tracked separately and fixed by #107156 (Resolve REFRESH ... DEPENDS ON dependencies in the view's database).
- `Stress test (arm_msan)` (`Hung check failed, possible deadlock found`): chronic msan stress-harness hang, not in the refresh path this PR touches.

This PR's C++ diff is confined to the `doScheduling`/`notifyDependentsIfNeeded` null guards in `RefreshTask.cpp`; neither stress failure originates there. Ready for review.

tuanpach:
@groeneai 

Merge master to this branch

tuanpach (approved):

nikitamikhaylov:
@groeneai Hello?

groeneai:
@nikitamikhaylov Sorry for the delay. @tuanpach approved on 06-22 and in the same minute asked me to merge master into the branch; I just did that.

- Merged `origin/master` into the branch (no conflicts). New head `6a3d451`, `MERGEABLE`, fresh CI running. The approved commit `1149ae22` stays in history so the approval is preserved.
- The C++ change is unchanged and confined to `RefreshTask.cpp`: a `if (!view) return;` guard at `doScheduling` entry plus a guard + snapshot-before-unlock in `notifyDependentsIfNeeded`. It fixes the null-`view` SIGSEGV during parallel `shutdown()` (STID 1805-36fc).
- The only reds on the previous head were the `Stress test (*)` "Unexpected exception in refresh scheduling" family. That is a separate trunk bug: a thrown exception re-raised by the catch-all in the `create()` initial-scheduling lambda (`RefreshTask.cpp:223` -> `:936`), not the shutdown null-deref this PR fixes (a SIGSEGV never reaches that catch). The `UNKNOWN_DATABASE` variant of it (STID 2508-3e7b) was already fixed by the merged #107156; a residual variant is still on master and tracked separately, independent of this PR.

Should be ready to merge once the fresh CI on `6a3d451` settles.

Issue #103745, opened 2026-04-30
Improve vector similarity search performance with binary quantization


shankar-iyer:
Below are extracts from recent stress test runs and also serves as a reference for anyone looking how binary quantization performs on large'ish datasets.

1. Default 'bf16' quantization -
Job : [link](https://github.com/ClickHouse/ClickHouse/actions/runs/24960866445/job/73087337094)

Dataset : hackernews_openai
```
2026-04-26T16:57:00.7953540Z [2026-04-26 16:57:00] Sun Apr 26 18:57:00 2026  :  Waiting for existing merges to complete...
2026-04-26T17:09:14.6522059Z [2026-04-26 17:09:14] Sun Apr 26 19:09:14 2026  :  Current active parts found for hackernews_openai -> 
2026-04-26T17:09:14.6522546Z [2026-04-26 17:09:14] Sun Apr 26 19:09:14 2026  :  all_1_229_4		38.44 GiB bytes
2026-04-26T17:09:14.6522912Z [2026-04-26 17:09:14] Sun Apr 26 19:09:14 2026  :  Adding vector similarity index
2026-04-26T17:09:14.6764359Z [2026-04-26 17:09:14] Sun Apr 26 19:09:14 2026  :  Materialzing the index
...
2026-04-26T18:31:16.8154473Z [2026-04-26 18:31:16] Sun Apr 26 20:31:16 2026  :  Waiting for materialize index to complete...
2026-04-26T18:31:46.8193995Z [2026-04-26 18:31:46] Sun Apr 26 20:31:46 2026  :  Done building the vector index
2026-04-26T18:31:46.8218705Z [2026-04-26 18:31:46] Sun Apr 26 20:31:46 2026  :  Index size for hackernews_openai -> 
2026-04-26T18:31:46.8219171Z [2026-04-26 18:31:46] Sun Apr 26 20:31:46 2026  :  vector_index		vector_similarity		29.82 GiB		33.21 GiB
...
2026-04-26T18:31:47.0113809Z [2026-04-26 18:31:47] Sun Apr 26 20:31:47 2026  :  Loaded truth set containing 1000 vectors
2026-04-26T18:32:03.8174904Z [2026-04-26 18:32:03] Sun Apr 26 20:32:03 2026  :  Vector indexes have loaded!
2026-04-26T18:32:20.4278786Z [2026-04-26 18:32:20] Sun Apr 26 20:32:20 2026  :  Runtime for ANN : 16.553 seconds
[...]

PR #105591, opened 2026-05-21
Apply exact row-positioning for vector search rescoring queries

skuznetsov:
This implements exact row-positioning for `vector_search_with_rescoring = 1`, matching the roadmap direction in https://github.com/ClickHouse/ClickHouse/issues/104122 and https://github.com/ClickHouse/ClickHouse/issues/103745.

Today rescoring can still carry the vector column downstream and evaluate the exact distance expression for all rows in selected granules. With this change, the optimizer keeps the original SQL distance expression, propagates vector-search read hints with an explicit row-filter signal, and `MergeTreeRangeReader` applies the exact part-offset filter before the downstream `ExpressionStep`. `_distance` is still populated only when the read header requests it, so the existing no-rescoring `_distance` path remains separate.

The explicit `use_vector_search_result_filter` flag keeps mark-pruning hints separate from row-level filtering. This avoids changing plans where `vector_search_results` exist but the row filter is intentionally disabled, such as the explicit `PREWHERE` case covered by the focused stateless test.

Local checks:
- `ninja -C build_vector_pr programs/clickhouse`
- `tests/clickhouse-test -b build_vector_pr/programs/clickhouse --queries tests/queries --no-stateful --no-random-settings --no-random-merge-tree-settings --no-zookeeper --no-shard --timeout 180 02354_vector_search_rescoring`: 3 tests passed, 0 skipped.
- `tests/clickhouse-test -b build_vector_pr/programs/clickhouse --queries tests/queries --no-stateful --no-random-settings --no-random-merge-tree-settings --no-zookeeper --no-shard --timeout 180 02354_vector_search_rescoring_and_prewhere --test-runs 5`: 5 tests passed, 0 skipped.

Local focused benchmark:
- baseline: PR merge-base `master` at `3b7724bb303`
- PR build: `ef9fa9de902`; latest pushed commit `5a30362c5a8` is test-only
- dataset: `50k` rows, `Array(Float32)`, `dim = 384`
- index: `vector_similarity('hnsw', 'L2Distance', 384, 'f32', 16, 64) GRANULARITY 100000000`
- query shape: `ORDER BY L2Distance(vec, ref) LIMIT 20`
- settings include `vector_search_with_rescoring = 1`, `max_threads = 1`, `query_plan_optimize_lazy_materialization = 0`, `query_plan_execute_functions_after_sorting = 0`
- each run: 30 warmup queries, then 100 measured queries

| metric | merge-base `master` | this PR | delta |
|---|---:|---:|---:|
| QPS | `18.1505` | `21.689` | `+19.5%` |
| p50 | `0.053s` | `0.0445s` | `-16.0%` |
| p95 | `0.056s` | `0.0465s` | `-17.0%` |
| p99 | `0.060s` | `0.049s` | `-18.3%` |

Both plans use the vector index (`Skip / vector_similarity`, `Granules: 7/8`) and keep exact rescoring in the SQL pipeline (`ReadFromMergeTree` reads `id UInt32` and `vec Array(Float32)`, no `_distance` rewrite for `vector_search_with_rescoring = 1`).

- Performance Improvement

Apply exact row-positioning for vector search queries with `vector_search_with_rescoring` before exact distance evaluation in the query pipeline.

- [ ] Documentation is not required.

### Version info
- Merged into: `26.7.1.238` (included in `26.7` and later)


skuznetsov:
@alexey-milovidov @abashkeev @nikitamikhaylov, this PR is for your attention

shankar-iyer:
Thanks for the contribution! This PR logically matches one of the items in our roadmap -

https://github.com/ClickHouse/ClickHouse/issues/104122
_Apply exact row positioning optimization when rescoring with quantized vector indexes - the vector column will still be read, but rescoring distance calculation will only be executed on specific neighbour rows rather than all rows in the granule. Currently the entire optimization is disabled if rescoring is ON. Tracked here : https://github.com/ClickHouse/ClickHouse/issues/103745_

and

https://github.com/ClickHouse/ClickHouse/issues/103745#issuecomment-4348660897

My idea was to  just enable the row-positioning optimization even when rescoring is ON (and drop the `_distance` column). That would require a small change in `MergeTreeRangeReader::fillDistanceColumnAndFilterForVectorSearch() `and maybe also in `useVectorSearch.cpp`.  As I wrote, currently we rescore with all vectors in a granule - we really only need to rescore with one (or handful) of vector rows in the granule that was returned by `usearch`. And this rescoring will happen naturally in the SQL pipeline in the `ExpressionStep` evaluation. Can you please explain why have you re-implemnented rescoring and added custom distance functions in `MergeTreeRangeReader` itself?

skuznetsov:
@shankar-iyer thanks, this makes sense.

I originally extended the existing **_distance** rewrite because it was the shortest path to proving that exact row positioning helps **vector_search_with_rescoring = 1**: keep the vector column, compute full-precision distance in the reader, and wire the sort expression to _distance _.

But I agree that the cleaner ClickHouse-native version is to decouple row-positioning from _distance:
- for **vector_search_with_rescoring = 0**, keep the existing _distance rewrite;
- for **vector_search_with_rescoring = 1**, propagate vector-search read hints even when _distance is not requested;
- in MergeTreeRangeReader, build/apply the row-offset filter from the hints, but fill _distance only if the read header requests it;
- leave the original L2Distance / cosineDistance / dotProduct expression in the SQL pipeline, so it is evaluated only after row-position filtering.

That should preserve the existing function semantics and avoid custom distance implementations in the reader. I’ll rework the PR in this direction.

skuznetsov:
@shankar-iyer I reworked the PR in the direction you suggested.

The reader-side fused rescoring prototype is gone now:
- no custom distance kernels in `MergeTreeRangeReader`;
- no `VectorSearchKernel`;
- no extra `simsimd` usage from this PR;
- no `_distance` rewrite for `vector_search_with_rescoring = 1`.

Current architecture:

- `vector_search_with_rescoring = 0` keeps the existing optimized `_distance` path.
- `vector_search_with_rescoring = 1` keeps the original SQL distance expression (`L2Distance`, `cosineDistance`, `dotProduct`, etc.) and keeps the vector column in the read list.
- The vector index still returns candidate row offsets.
- `ReadFromMergeTree` propagates vector-search read hints even when `_distance` is not requested.
- `MergeTreeRangeReader::fillDistanceColumnAndFilterForVectorSearch` now builds the exact row-position filter from those offsets.
- `_distance` is filled only if the read header requests it.
- The regular `ExpressionStep` then computes exact rescoring only for the candidate rows that survived the row-position filter.

So the PR now decouples row-positioning from `_distance`, while preserving the existing SQL pipeline semantics for exact rescoring.

I also kept explicit `_distance` readable under rescoring: if `_distance` is selected explicitly, the vector index is asked to return distances and the virtual column is populated as before.

Validation added/updated:
- `02354_vector_search_rescoring` checks that rescoring no longer rewrites to `_distance`.
- The same test has a `throwIf` guard proving the row-position filter is applied before the downstream `ExpressionStep` evaluates the exact distance expression.
- `04109_vector_search_rescoring_distance_select` covers:
  - distance expression in the select list under exact rescoring;
  - no `_distance` rewrite in `EXPLAIN`;
  - explicit `_distance` selected under rescoring.
- Existing/focused coverage also includes `PREWHERE`, non-`Float32` element types, `Float64` reference precision, cosine zero-vector behavior, and `dotProduct` fallback.

Local verification:
- `ninja -C build_vector_pr programs/clickhouse`
- focused stateless run: 8 tests passed, 0 skipped

Local benchmark, synthetic `50k` rows with `dim = 384`, same query shape and local release build:

| Path | QPS |
|---|---:|
| full exact scan, vector search disabled | `72.75` |
| row-positioned exact rescoring, this PR | `246.45` |
| no-rescore `_distance` path | `335.97` |

So for this benchmark the reworked exact-rescoring path is about `3.39x` faster than the full exact scan baseline, while still doing exact SQL-pipeline rescoring.

shankar-iyer:
_The following is a review note from Claude (posted by @shankar-iyer on its behalf)._

---

Thanks for iterating on this. A suggestion on framing for the next revision: the smallest change that makes rescoring use the existing row-positioning filter is a much safer starting point than a new code path. The current diff is ~530 lines, but the actual behavior delta ("apply the filter the index already produces, even when `_distance` is not in the read list") is about 25 lines spread over two files. Starting from the minimal diff and only adding complexity that's clearly justified would be easier to review and lower-risk to land.

Concretely, the smallest sufficient change is:

1. In `LoadedMergeTreeDataPartInfoForReader::setReadHints`, drop the `read_columns.contains("_distance")` gate so vector-search hints are stored whenever they're present.
2. In `MergeTreeRangeReader::fillDistanceColumnAndFilterForVectorSearch`, build the row-offset filter as today, but only allocate / populate the `_distance` column when `read_sample_block.has("_distance")`. The existing `offset_to_distance` map lookup stays.

That's it. No new flag, no optimizer-side plumbing, no changes to `filterMarksUsingIndex`, no changes to how the index is asked for distances.

A few things in the current diff that are worth dropping for risk reasons:

- **The strict `std::is_sorted` → `LOGICAL_ERROR` in `fillDistanceColumnAndFilterForVectorSearch`** is a real regression risk. The rows arriving at the range reader aren't always sorted by row offset in practice — binary quantization (`vector_similarity('hnsw', …, 'b1', …)`) is the case that's been observed, and the issue is upstream of this code in the multi-index-granule loop in `filterMarksUsingIndex`. The previous local sort + offset→distance map made the consumer immune to this. Replacing it with a hard throw turns a class of currently-working queries into thrown `LOGICAL_ERROR`s. Keep the defensive local sort; if and when the upstream ordering bug is fixed, the strictness can be added then.

- **The switch from "local sort + offset→distance map" to "in-place sort of stored hints + positional lookup"** changes the contract of `RangesInDataPartReadHints::vector_search_results` — consumers may now assume sorted-and-paired, whereas previously they could only assume paired in arbitrary order. That contract change is independent of the rescoring fix; if it's worth making, it deserves its own PR with its own justification. As written, it adds risk to a change that doesn't need it.

- **The `use_vector_search_result_filter` flag in `RangesInDataPartReadHints`** doesn't carry information that isn't already implicit in `vector_search_results.has_value()` — whenever vector-search hints are present, the consumer wants to apply the row filter. Adding a flag means every producer and consumer now has to be kept consistent on that flag, for no behavioral gain. Removing it would also let the matching optimizer-side helpers (`enableVectorSearchRowFilter`, the `read_distance_column` / `return_distances` recomputation, the extra `setVectorSearchParameters` call) go away.

- **Unrelated changes mixed into the diff** — `sort_description.size() > 1` → `!= 1` (looks like a legitimate fix to a potential `front()` on an empty vector), the `createLocalParallelReplicas` `return std::make_unique<…>` → temp-variable refactor, and the `return optimization_applied` change in `optimizeVectorSearchSecondPass`. Each is defensible in isolation, but bundling them inflates the review surface and makes bisection harder if anything regresses. Better as standalone PRs.

On the benchmark: the relevant baseline is `master` with `vector_search_with_rescoring = 1` (granule-level rescoring), not a full brute-force scan. Could the table in the PR description be re-run with that row included? The expected uplift for selective queries is in the 20–30% latency range from local measurements with the minimal patch.

And the PR title still says "Fuse vector search rescoring in MergeTree reads", which no longer reflects what the PR does — nothing is being fused. Something like "Apply exact row-positioning for vector search rescoring queries" would match, with a matching changelog entry.

As an aside, if agentic AI / coding tools are part of the workflow here, a tightly-scoped framing like *"enable the existing row-positioning optimization even when ``vector_search_with_rescoring = 1``, so the filter the index already produces is applied to rescoring queries too"* tends to land close to the two-file change directly. Sharing in case it's useful — broader prompts naturally produce broader diffs.

Happy to share a sketch of the minimal patch as a separate branch if that helps as a reference point. Thanks for working on this.

shankar-iyer:
Please check above comment from Claude, I agree with that.

skuznetsov:
@shankar-iyer thanks, agreed on keeping the PR smaller and on not using the earlier local benchmark framing as a `master`-vs-PR claim.

I updated the PR title/description and pushed a scoped revision:
- removed the reader-side fused rescoring/custom distance path from the effective diff;
- kept `vector_search_with_rescoring = 1` in the normal SQL pipeline;
- use the vector-search second pass only to mark analyzed parts where exact row offsets are available;
- `MergeTreeRangeReader` applies the existing part-offset filter even when `_distance` is not requested;
- `_distance` is still populated only when the read header requests it.

I kept one explicit `use_vector_search_result_filter` signal instead of storing/applying hints unconditionally. I tested the unconditional version and it changed the explicit `PREWHERE` case: `vector_search_results` can be present for mark pruning even when row-level filtering must remain disabled. The focused stateless test now covers that distinction.

Validation:
- `ninja -C build_vector_pr programs/clickhouse`
- `tests/clickhouse-test -b build_vector_pr/programs/clickhouse --queries tests/queries --no-stateful --no-random-settings --no-random-merge-tree-settings --no-zookeeper --no-shard --timeout 180 02354_vector_search_rescoring`: 3 tests passed, 0 skipped.

I removed the old local perf claim from the PR description. The relevant performance comparison should be against current `master` with `vector_search_with_rescoring = 1`, not against a full exact-scan fallback shape.

src/Processors/QueryPlan/Optimizations/useVectorSearch.cpp
```diff
@@ -463,7 +464,39 @@ bool optimizeVectorSearchSecondPass(QueryPlan::Node & /*root*/, Stack & stack, Q
         sorting_step->updateInputHeader(expression_node->step->getOutputHeader());
     }
 
-    return true;
+    if (apply_row_filter_for_rescoring)
```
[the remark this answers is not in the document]
skuznetsov: Thanks, agreed. I re-ran the focused comparison against the relevant rescoring baseline, not against the full exact-scan fallback.

Local benchmark setup:
- baseline: PR merge-base `master` at `3b7724bb303`
- PR build: `ef9fa9de902`
- latest pushed commit `5a30362c5a8` is test-only (`02354_vector_search_rescoring_and_prewhere`), so the performance-affecting code is unchanged from the measured PR build
- dataset: `50k` rows, `Array(Float32)`, `dim = 384`
- index: `vector_similarity('hnsw', 'L2Distance', 384, 'f32', 16, 64) GRANULARITY 100000000`
- query shape: `ORDER BY L2Distance(vec, ref) LIMIT 20`
- settings include `vector_search_with_rescoring = 1`, `max_threads = 1`, `query_plan_optimize_lazy_materialization = 0`, `query_plan_execute_functions_after_sorting = 0`
- each run: 30 warmup queries, then 100 measured queries

Results:

| build | run | QPS | p50 | p95 | p99 |
|---|---:|---:|---:|---:|---:|
| merge-base `master` | 1 | `18.211` | `0.053s` | `0.056s` | `0.059s` |
| merge-base `master` | 2 | `18.090` | `0.053s` | `0.056s` | `0.061s` |
| this PR | 1 | `21.829` | `0.044s` | `0.046s` | `0.048s` |
| this PR | 2 | `21.549` | `0.045s` | `0.047s` | `0.050s` |

Averages:

| metric | merge-base `master` | this PR | delta |
|---|---:|---:|---:|
| QPS | `18.1505` | `21.689` | `+19.5%` |
| p50 | `0.053s` | `0.0445s` | `-16.0%` |
| p95 | `0.056s` | `0.0465s` | `-17.0%` |
| p99 | `0.060s` | `0.049s` | `-18.3%` |

Both plans use the vector index (`Skip / vector_similarity`, `Granules: 7/8`) and both keep exact rescoring in the SQL pipeline (`ReadFromMergeTree` header contains `id UInt32` and `vec Array(Float32)`, no `_distance` rewrite for `vector_search_with_rescoring = 1`).

So the scoped claim is: on this local focused benchmark, applying exact row-position filtering to the existing rescoring path improves QPS by about `19.5%` and reduces p50/p95/p99 latency by about `16-18%`. The earlier full exact-scan comparison is intentionally not used as the `master`-vs-PR performance claim.

shankar-iyer:
Stress test run with binary quantization and rescoring_multiplier = 10

```
[2026-05-27 01:30:57] > Start execution for [Extract truthset hackernews_openai_10m_1k.tar]
[2026-05-27 01:30:57] Run command: [tar xvf hackernews_openai_10m_1k.tar]
[2026-05-27 01:30:57] hackernews_openai_10m_1k_vectors.npy
[2026-05-27 01:30:57] hackernews_openai_10m_1k_neighbours.npy
[2026-05-27 01:30:57] hackernews_openai_10m_1k_distances.npy
[2026-05-27 01:30:57] Wed May 27 03:30:57 2026  :  Loaded truth set containing 1000 vectors
[2026-05-27 01:31:01] Wed May 27 03:31:01 2026  :  Vector indexes have loaded!
[2026-05-27 01:33:37] Wed May 27 03:33:37 2026  :  Runtime for ANN : 155.615 seconds
[2026-05-27 01:33:37] Wed May 27 03:33:37 2026  :  Calculating recall...
[2026-05-27 01:33:37] Wed May 27 03:33:37 2026  :  Recall is 0.9919400000000037
[2026-05-27 01:33:37] Wed May 27 03:33:37 2026  :  Dropping table hackernews_openai ...
```

Earlier run without PR -

```
Wed Apr 29 14:35:32 2026  :  Loaded truth set containing 1000 vectors
Wed Apr 29 14:35:36 2026  :  Vector indexes have loaded!
Wed Apr 29 14:38:52 2026  :  Runtime for ANN : 195.913 seconds
Wed Apr 29 14:38:52 2026  :  Calculating recall...
Wed Apr 29 14:38:52 2026  :  Recall is 0.9959900000000024
```

Optimization is cool - 195 secs -> 155 secs for 1000 queries for k=100

skuznetsov:
@shankar-iyer Thanks for running the stress test!

This is useful because it validates the intended production-like path: binary quantization with exact rescoring. The measured runtime improvement is about 20.6% wall-clock reduction, or about 25.9% higher query throughput for the 1000-query run.

The small recall difference also makes sense to me. This PR removes the previous granule spillover behavior: instead of rescoring all rows from the granules returned by the vector index, it applies the exact row-position filter first and then lets the regular SQL expression compute the full-precision distance. That should reduce unnecessary reads and distance evaluations, but it can also remove incidental extra candidates that happened to live in the same granules. For workloads that need the previous recall level, increasing the fetch/rescoring multiplier should be the right knob.

So I would frame this as a latency/read-amplification improvement, not as a recall improvement.

tests/queries/0_stateless/02354_vector_search_rescoring_and_prewhere.sql
```diff
@@ -66,14 +68,19 @@ SETTINGS query_plan_optimize_prewhere = 1,
          optimize_move_to_prewhere = 1;
 
 SELECT 'Test with enabled rescoring';
--- Expect 16 & 19, and additionally 18 and 17 because they are in the same granules
+-- Expect only the exact rows returned by the vector index. The old granule-level
+-- rescoring behavior also returned 18 and 17 because they are in the same granules.
```
shankar-iyer: Good catch.

tests/queries/0_stateless/02354_vector_search_rescoring_and_prewhere.sql
```diff
@@ -66,14 +68,19 @@ SETTINGS query_plan_optimize_prewhere = 1,
...
 
 SELECT id
 FROM tab
 WHERE attr1 > 110
 ORDER BY L2Distance(vec, [0.2, 0.3])
 LIMIT 4
-SETTINGS vector_search_with_rescoring = 1;
+SETTINGS vector_search_with_rescoring = 1,
+         optimize_move_to_prewhere = 0,
```
shankar-iyer: I think we need one more change. This location -

https://github.com/ClickHouse/ClickHouse/blob/4c4e1b473dee21b2c4f49f4b42daa4b5afe920bb/src/Processors/QueryPlan/Optimizations/optimizePrewhere.cpp#L176

We should drop the` !settings[Setting::vector_search_with_rescoring]`, please test for side effects (hopefully none). Then we don't need `query_plan_optimize_prewhere = 0` in this query and next.

skuznetsov:
Thanks, this was the right simplification. I updated the PR so `optimizePrewhere` now skips implicit `PREWHERE` movement for any vector search lookup, not only when `vector_search_with_rescoring` is disabled. That also lets the rescoring/prewhere regression test use the default `optimize_move_to_prewhere` / `query_plan_optimize_prewhere` settings instead of per-query opt-outs.

Validation I re-ran locally:

- `ninja -C build_vector_pr programs/clickhouse`
- focused `02354_vector_search_rescoring_and_prewhere`
- vector-search stateless sweep: `02354_vector_search_rescoring`, `02354_vector_search_rescoring_distance_in_select_list`, `02354_vector_search_pre_and_post_filtering`, `02354_vector_search_binary_quantization`, `02354_vector_search_rescoring_and_prewhere`
- `git diff --check`

I also did a focused local perf sanity check for this guard change only. Setup: synthetic `50k` rows, `dim=384`, `vector_similarity('hnsw', 'L2Distance', ...)`, 100 query iterations, `max_threads = 1`, expression/aggregate/sort compilation disabled, and `vector_search_index_fetch_multiplier = 20` so result row counts match before/after. Both plans still materialize explicit `L2Distance` and use the `vector_similarity` skip index; `_distance` does not appear in the valid `EXPLAIN` output.

| case | before | after | delta |
|---|---:|---:|---:|
| `attr1 >= 500` QPS | 73.4 | 153.6 | +109% |
| `attr1 >= 500` p50/p95/p99 | 12/13/13 ms | 5/5/6 ms | better |
| `attr1 >= 900` QPS | 140.1 | 151.4 | +8% |
| `attr1 >= 900` p50/p95/p99 | 5/6/7 ms | 5/6/6 ms | not worse |

I discarded a lower-multiplier selective run from perf claims because row counts diverged (`20` before vs `15` after), so the table above is the apples-to-apples run. This is not meant as a production benchmark, only a regression check for the requested `PREWHERE` guard change.

shankar-iyer (approved):
Please remove explicit setting edits in the 2 test files : query_plan_optimize_lazy_materialization, compile_expressions etc.

LGTM from my side. Let's wait for feedback from @rschu1ze

skuznetsov:
Addressed. I removed the explicit helper setting overrides from the two vector rescoring stateless tests:

- `query_plan_optimize_lazy_materialization`
- `query_plan_execute_functions_after_sorting`
- `compile_expressions`
- `compile_sort_description`

I kept only the explicit `PREWHERE` on/off settings in the block that intentionally tests `PREWHERE` behavior.

Re-checked locally after the change:

- `02354_vector_search_rescoring`
- `02354_vector_search_rescoring_distance_in_select_list`
- `02354_vector_search_rescoring_and_prewhere`

Result: `3 tests passed, 0 skipped`. Also ran `git diff --check`.

shankar-iyer:
@skuznetsov Can you please refresh to master once and push?

skuznetsov:
Refreshed the branch to current `master` and pushed it.

Local verification on `4c12804907fd027e1b62e219794eb5b287139ca0`:
- `ninja -C build_verify programs/clickhouse`
- `build_verify/programs/clickhouse --version` -> `ClickHouse local version 26.7.1.1.`
- `git diff --check`
- targeted stateless tests:
  - `02354_vector_search_rescoring`
  - `02354_vector_search_rescoring_distance_in_select_list`
  - `02354_vector_search_rescoring_and_prewhere`

All 3 targeted tests passed locally.

tests/queries/0_stateless/02354_vector_search_rescoring_and_prewhere.sql
```diff
@@ -66,7 +66,8 @@ SETTINGS query_plan_optimize_prewhere = 1,
          optimize_move_to_prewhere = 1;
 
 SELECT 'Test with enabled rescoring';
--- Expect 16 & 19, and additionally 18 and 17 because they are in the same granules
+-- Expect only the exact rows returned by the vector index. The old granule-level
```
rschu1ze: Please update the docs https://clickhouse.com/docs/engines/table-engines/mergetree-family/annindexes - they say `ClickHouse will then load these granules from disk, and repeat the distance calculation for all vectors in these granules.` ... as per this PR, we are no longer rescoring _all_ vectors.

tests/queries/0_stateless/02354_vector_search_rescoring.sql
```diff
@@ -50,6 +50,15 @@ SELECT trimLeft(explain) AS explain FROM (
     SETTINGS vector_search_with_rescoring = 1)
 WHERE (explain LIKE '%_distance%');
 
+SELECT 'Test exact row-positioning filters before ExpressionStep';
+
+WITH CAST([0.0, 2.0], 'Array(Float32)') AS reference_vec
```
rschu1ze: Is plain `[0.0, 2.0]` not going to work? The syntax is working in all other tests in this file. Why do you need to cast?

tests/queries/0_stateless/02354_vector_search_rescoring.sql
```diff
@@ -50,6 +50,15 @@ SELECT trimLeft(explain) AS explain FROM (
     SETTINGS vector_search_with_rescoring = 1)
 WHERE (explain LIKE '%_distance%');
 
+SELECT 'Test exact row-positioning filters before ExpressionStep';
```
rschu1ze: Minor: Please pre-pend `--` like in l. 42, 64, etc.

rschu1ze: The term "exact row-positioning" is not an established term. Future readers will nervously scratch their head what the test author could have possibly meant or not meant. We should not use this term in any public places.

Please rewrite in a way that l. 53 describes the purpose of the following test: _what_ is it testing. _what_ is the expected outcome. Please also don't use names from the C++ codebase (`ExpressionStep`). These are implementation details.

tests/queries/0_stateless/02354_vector_search_rescoring_and_prewhere.sql
rschu1ze: No need to let the comment mention how things worked in an old Git state. Let's change l. 69/70 to `Expect 16 & 19.`

src/Storages/MergeTree/RangesInDataPart.h
```diff
@@ -93,6 +93,10 @@ struct RangesInDataPartReadHints
 {
     /// Currently only information related to vector search
     std::optional<NearestNeighbours> vector_search_results;
+    /// `vector_search_results` can be used only for mark pruning, or also as
```
rschu1ze: It's not so clear what the comment l. 96-99 means. Please describe exactly what `use_vector_search_result_filter = false` and `= true` mean.

Also, the term `use_vector_search_result_filter` is a bit questionable. Should that maybe be

```cpp
/// false: rescore only matching rows in the granule
/// true: rescore all rows in the granule (more wasteful, also correct)
bool rescoreIndividualRows = false;
```

(I'm not sure)

rschu1ze (commented):
Sorry for the late review, this PR was somehow not on my radar.

I had a couple of comments but the PR overall looks good. Thanks.

skuznetsov:
Pushed follow-up fixes on `7e227e5f4d54cc982fdcc912f5413500cea5adba`.

Addressed the latest review feedback:
- aggregated vector-search candidate rows across multiple vector index granules instead of keeping only the last sub-index result;
- added a small-`GRANULARITY 2` stateless regression so rescoring preserves candidates from multiple vector index granules;
- added coverage for explicit `_distance` with and without `vector_search_with_rescoring`;
- updated `vector_search_with_rescoring` setting text and ANN docs to describe filtering to candidate rows before final rescoring when the plan allows it;
- adjusted test wording/comments per review: no `exact row-positioning` term in public/test output, no `ExpressionStep` mention in test text, and no old-state wording;
- clarified the `use_vector_search_result_filter` comment and the vector-search/PREWHERE optimizer comment.

Local verification on the pushed branch:
- `ninja -C build_verify programs/clickhouse`
- `git diff --check`
- targeted stateless run:
  - `02354_vector_search_rescoring`
  - `02354_vector_search_rescoring_distance_in_select_list`
  - `02354_vector_search_rescoring_and_prewhere`

Result: `3 tests passed, 0 skipped`.

shankar-iyer:
Only failure right now is in the "Performance Comparision", unrelated and is noise.

Algunenano:
This caused conflicts with https://github.com/ClickHouse/ClickHouse/pull/108423 and had to be reverted in https://github.com/ClickHouse/ClickHouse/pull/108812. Please re-land it

PR #101798, opened 2026-04-05
Prevent server termination when unregistering S3 client fails

gagandhakrey:
## Linked issues
- fixes https://github.com/ClickHouse/clickhouse-core-incidents/issues/1697

Prevents the server from terminating (`std::terminate`) when an exception is thrown during S3 client cache unregistration. The `throw;` statement was erroneously placed within the `catch` block of `Client::~Client()`, which is a `noexcept` destructor.

### Changelog category (leave one):

- Critical Bug Fix (crash, data loss, RBAC)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix exception escaping from S3 `Client::~Client` destructor causing server termination.

StackOverflow similar issue link: https://stackoverflow.com/questions/40455741/why-c11-mark-destructors-as-nothrow-and-is-it-possible-to-override-it?#:~:text=Destructors%20are%20called,override%20the%20default.
...

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.814`
- Backported to: `26.3.7.8`, `26.2.11.5`, `26.1.8.35`, `25.8.24.10`

PR #105602, opened 2026-05-21
Stop test 01658 from leaking files into user_files_path, fix 02843 flaky

groeneai:
Changelog category (leave one):

Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):

---

## Why

`02843_context_has_expired.sql` fails sporadically on CI with:

```
The query succeeded but the server error '107' was expected
(query: SELECT 1 in (SELECT file(materialize('a'))); -- { serverError FILE_DOESNT_EXIST })
```

CIDB shows ~4 hits in the past 90 days, all matching the same pattern, on unrelated PRs (e.g. #100651, #101798, #104697). Crucially, the diagnostic always reports "Test also fails without randomized settings (not a randomization issue)" — so this is not a randomized-settings flake, the server is in a stuck state.

## Root cause

`01658_read_file_to_stringcolumn.sh` writes the short filenames `a`, `b`, `c` into `user_files_path` (line 88 in the current script) and removes them only via the manual `rm` at the bottom. The script runs under `set -e`. Any failing query between the write and the cleanup aborts the script and leaves the files behind, where they survive across the rest of that CI session.

Once `${USER_FILES_PATH}/a` exists, `02843`'s assertion `SELECT 1 in (SELECT file(materialize('a')))` returns 0 (the file is read successfully and `1` is not in the resulting Set) instead of throwing `FILE_DOESNT_EXIST`, and the `-- { serverError FILE_DOESNT_EXIST }` check trips the test.

Reproduced locally:
```
$ echo -n Hello > $USER_FILES_PATH/a
$ clickhouse-client -q "SELECT 1 in (SELECT file(materialize('a')))"
0    # ← bug
$ rm $USER_FILES_PATH/a
$ clickhouse-client -q "SELECT 1 in (SELECT file(materialize('a')))"
Code: 107. DB::ErrnoException: Cannot open file ... FILE_DOESNT_EXIST   # ← expected
```

50/50 deterministic.

## Fix

Two changes:

1. **Root cause:** `01658_read_file_to_stringcolumn.sh` now registers `trap cleanup EXIT` immediately after sourcing `shell_config.sh`. The cleanup removes all files the test creates in `user_files_path` and `/tmp`, using `rm -f` so it's idempotent. The redundant manual cleanup at the bottom of the script is removed.

2. **Defense in depth:** `02843_context_has_expired.sql` uses `'02843_nonexistent_'||generateUUIDv4()` as the path passed to `file()`. The test's intent (verifying that `file()` throws `FILE_DOESNT_EXIST` on a missing file) is preserved, and the assertion can no longer be poisoned by another test leaking a short filename like `a`.

## Verification

- Pre-fix, with `${USER_FILES_PATH}/a` pre-seeded: `02843` fails 50/50 (matches CI failure exactly).
- Post-fix, with `${USER_FILES_PATH}/a` pre-seeded: `02843` passes 50/50.
- `01658` trap verified to fire on `set -e` mid-script abort (manual bash test with `false`).
- `bash -n` syntax check on modified `01658` passes.

### Version info
- Merged into: `26.6.1.635`

### DIFF
--- a/tests/queries/0_stateless/01658_read_file_to_stringcolumn.sh
+++ b/tests/queries/0_stateless/01658_read_file_to_stringcolumn.sh
@@ -7,6 +7,17 @@ CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 # shellcheck source=../shell_config.sh
 . "$CURDIR"/../shell_config.sh
 
+# Clean up on EXIT so a mid-script abort (set -e + a failing query) cannot
+# leave the short filenames `a`, `b`, `c` in `user_files_path` and break
+# other tests that rely on them being absent.
+cleanup() {
+    rm -f "${USER_FILES_PATH}"/{a,b,c}.txt
+    rm -f "${USER_FILES_PATH}"/{a,b,c}
+    rm -f /tmp/c.txt
+    rm -rf "${USER_FILES_PATH}"/dir
+}
+trap cleanup EXIT
+
 echo -n aaaaaaaaa > ${USER_FILES_PATH}/a.txt
 echo -n bbbbbbbbb > ${USER_FILES_PATH}/b.txt
 echo -n ccccccccc > ${USER_FILES_PATH}/c.txt
@@ -81,8 +92,4 @@ echo -n World > ${USER_FILES_PATH}/c
 ${CLICKHOUSE_CLIENT} --query "SELECT file(arrayJoin(['a', 'b', 'c'])) AS s, count() GROUP BY s ORDER BY s"
 ${CLICKHOUSE_CLIENT} --query "SELECT s, count() FROM file('?', TSV, 's String') GROUP BY s ORDER BY s"
 
-# Restore
-rm ${USER_FILES_PATH}/{a,b,c}.txt
-rm ${USER_FILES_PATH}/{a,b,c}
-rm /tmp/c.txt
-rm -rf ${USER_FILES_PATH}/dir
+# Cleanup is handled by the `trap cleanup EXIT` at the top of this script.
--- a/tests/queries/0_stateless/01658_read_file_to_stringcolumn.sh
+++ b/tests/queries/0_stateless/01658_read_file_to_stringcolumn.sh
@@ -7,13 +7,18 @@ CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 # shellcheck source=../shell_config.sh
 . "$CURDIR"/../shell_config.sh
 
+# A file outside user_files, used to check that reading such a path is rejected
+# in client mode and allowed in local mode. Keep it in the per-test unique tmp
+# dir instead of a fixed /tmp path so concurrent tests cannot collide on it.
+OUTSIDE_FILE="${CLICKHOUSE_TMP}/01658_outside.txt"
+
 # Clean up on EXIT so a mid-script abort (set -e + a failing query) cannot
 # leave the short filenames `a`, `b`, `c` in `user_files_path` and break
 # other tests that rely on them being absent.
 cleanup() {
     rm -f "${USER_FILES_PATH}"/{a,b,c}.txt
     rm -f "${USER_FILES_PATH}"/{a,b,c}
-    rm -f /tmp/c.txt
+    rm -f "${OUTSIDE_FILE}"
     rm -rf "${USER_FILES_PATH}"/dir
 }
 trap cleanup EXIT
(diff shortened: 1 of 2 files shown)


groeneai:
### Pre-PR validation gate (per task pipeline)

**a) Deterministic repro?** Yes. After `echo -n Hello > ${USER_FILES_PATH}/a`, the pre-fix `02843_context_has_expired.sql` fails 50/50 with the exact CI signature (`The query succeeded but the server error '107' was expected`). With the fix in place, the same setup passes 50/50.

**b) Root cause explained?** Yes. `01658_read_file_to_stringcolumn.sh` writes `a`, `b`, `c` into `user_files_path` and removes them only via the manual `rm` at the bottom. Under `set -e`, any failing query between the write and the cleanup aborts the script and leaks those files. Once `${USER_FILES_PATH}/a` exists, `02843`'s `SELECT 1 in (SELECT file(materialize('a')))` succeeds (returning 0) instead of throwing `FILE_DOESNT_EXIST`, and the `-- { serverError FILE_DOESNT_EXIST }` annotation trips.

**c) Fix matches root cause?** Yes. The root-cause fix is `trap cleanup EXIT` in `01658` so the files cannot leak even on mid-script abort. The defense-in-depth fix in `02843` swaps the hardcoded `'a'` for `'02843_nonexistent_'||generateUUIDv4()`, so the assertion cannot be poisoned by any future test that leaks a short filename.

**d) Test intent preserved?** Yes. `02843` still asserts `file()` throws `FILE_DOESNT_EXIST` on a missing path. `01658` still runs every valid/invalid case it did before; only the cleanup mechanism changed (manual `rm` -> `trap cleanup EXIT`).

**e) Both directions demonstrated?** Yes. With `${USER_FILES_PATH}/a` pre-seeded:
- pre-fix `02843`: passes=0 fails=50 (out of 50)
- post-fix `02843`: passes=50 fails=0 (out of 50)

**f) Fix is general?** Yes. The `trap` pattern is the canonical bash idiom for cleanup on abort, already used by many other stateless tests (e.g. `00309_formats_case_insensitive.sh`, `00408_http_keep_alive.sh`, `01278_min_insert_block_size_rows_for_materialized_views.sh`). The defensive `02843` rewrite protects against any other test that might write `a` (or any short filename) into `user_files_path` and crash before cleaning up — the fix isn't narrow to `01658`.

Session: `cron:clickhouse-ci-task-worker:20260521-223000`.

groeneai:
cc @alexey-milovidov — could you review this? Test-only fix for a flaky-test root cause: `01658_read_file_to_stringcolumn.sh` leaks `${USER_FILES_PATH}/a` on mid-script abort under `set -e`, which then makes `02843_context_has_expired`'s `FILE_DOESNT_EXIST` assertion succeed instead of throw. Adding `trap cleanup EXIT` to 01658 and a UUID-suffixed path to 02843.

tests/queries/0_stateless/01658_read_file_to_stringcolumn.sh
```diff
@@ -7,6 +7,17 @@ CURDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 # shellcheck source=../shell_config.sh
 . "$CURDIR"/../shell_config.sh
 
+# Clean up on EXIT so a mid-script abort (set -e + a failing query) cannot
+# leave the short filenames `a`, `b`, `c` in `user_files_path` and break
+# other tests that rely on them being absent.
+cleanup() {
+    rm -f "${USER_FILES_PATH}"/{a,b,c}.txt
+    rm -f "${USER_FILES_PATH}"/{a,b,c}
+    rm -f /tmp/c.txt
```
azat: Why the test uses /tmp/c.txt, if you need something outside of user_files, just use regular tmp dir (that is unique for each test)

groeneai:
Rebased onto current master (`a85f4007`) and re-pushed to refresh CI.

The previous run's only red was `test_mysql_protocol/test.py::test_mysql_dotnet_client` (Integration tests amd_llvm_coverage 3/5) — the `.NET` MySQL wire-protocol test, whose stdout was truncated mid-stream in the assert. This PR is test-only (two stateless `.sql`/`.sh` files), so it has no causal path to a .NET MySQL integration test; under the heavily-instrumented `amd_llvm_coverage` build that external-client test is timing-sensitive and flakes. The other three reds (Finish Workflow, Mergeable Check, PR) were rollups of that single failure.

The underlying fix is still needed: master still has 01658's manual `rm`-without-trap cleanup and 02843's hardcoded `file(materialize('a'))` assertion. Fresh CI running now.

tests/queries/0_stateless/01658_read_file_to_stringcolumn.sh
groeneai: Good point. Done in 0f93f37: the file now lives in the per-test unique `$CLICKHOUSE_TMP` dir (`$CLICKHOUSE_TMP/01658_outside.txt`) instead of the fixed `/tmp/c.txt`, so concurrent runs can't collide on it, and the runner removes it afterwards.

In client mode `file()` rejects any path outside `user_files` before checking existence, so the rejection assertion is unaffected by the path. Only the local-mode read needs the contents. The `..`-traversal case keeps its relative path since it only checks that escaping `user_files` is denied.

Verified locally via the runner: 01658 and 02843 pass 5/5 each with randomization, no `/tmp/c.txt` is created, and the EXIT trap still cleans up on a mid-script abort.

azat (approved):

Issue #105523, opened 2026-05-21
RFC: Resource usage observability for executable user-defined functions


andreev-io:
### Company or project name

ClickHouse

### Use case

ClickHouse currently captures no resource accounting for `executable` and `executable_pool` UDFs. Operators have no way to answer "how much CPU is my UDF using," "is it leaking memory," or "which queries pay the most for UDFs."

This RFC proposes exposing UDF resource usage through ClickHouse's existing observability primitives, mirroring what WebAssembly UDFs do today (`Wasm*` ProfileEvents at `src/Common/ProfileEvents.cpp:1343`):

- **Synchronous metrics** — additive ProfileEvents — landing in `system.query_log`, `system.events`, `system.metric_log`.
- **Asynchronous metrics** — per-UDF gauges — landing in `system.asynchronous_metrics`, `system.asynchronous_metric_log`.

No new system tables. Status: [#104998](https://github.com/ClickHouse/ClickHouse/pull/104998) registers the events (merged); [#105010](https://github.com/ClickHouse/ClickHouse/pull/105010) wires `executable_pool` (in review); `executable` sync metrics and the async metrics are pending.

## Background

Both executable UDF types route through `src/Common/ShellCommand.{h,cpp}` and `src/Processors/Sources/ShellCommandSource.{h,cpp}`. They differ in process lifecycle:

- **`executable`** spawns a fresh child per invocation (`vfork()` + `execve()` per block), reaps with `waitpid()`. Suitable for stateless per-call work.
- **`executable_pool`** maintains a fixed-size pool of long-lived child workers, borrowing one per invocation. Amortises startup; right choice for warm interpreters, loaded models, or kept-open connections.

Currently `struct rusage` from `waitpid` is discarded, no periodic sampler observes pool processes, and nothing reaches any system table.

## Goals

- Per-query attribution: invocations, wall duration, CPU (user + system), memory cost, bytes in / out, pool-wait time.
- Per-UDF live observability: process count, resident memory.
- Coverage for both UDF types.
- Capture the entire UDF process subtree, not just the worker root — UDF scripts can fork helpers.
- Expose everything through existing system tables; no new tables introduced.

## Non-goals (v1)

[...]

PR #105618, opened 2026-05-22
Populate ProfileEvents for executable UDFs via wait4

andreev-io:
Populates the eight `ExecutableUserDefinedFunction*` `ProfileEvents` on the `executable` (non-pool) UDF path, so a query that calls a non-pool UDF shows up in `system.query_log.ProfileEvents` with per-call invocation count, wall time, child-process CPU and peak memory, and stdin/stdout bytes. `ExecutableUserDefinedFunctionPoolWaitMicroseconds` is N/A for this path and stays zero.

`wait4` is used instead of the RFC's proposed POSIX shim to get per-child `rusage` atomically without an extra `fork`. Full rationale in https://github.com/ClickHouse/ClickHouse/issues/105523#issuecomment-4516641754.

This completes the **synchronous-metrics** line of RFC #105523: #104998 (merged) registered the eight events; #105010 (merged) populates them for `executable_pool` UDFs via the shared `UDFProcessSubtreeSampler`; this PR populates them for `executable` UDFs by folding the `rusage` captured by `wait4` in `ShellCommand` into the same sampler (`recordExecutableFinished`). The RFC's asynchronous per-UDF gauges (live process count and current RSS, landing in `system.asynchronous_metrics` / `system.asynchronous_metric_log`) are out of scope here and remain pending.

The cleanup reap is non-blocking: with `check_exit_code=false` the source deliberately does not wait for the child, so a `WNOHANG` reap harvests `rusage` when the child has already exited and otherwise skips accounting — profiling never adds a blocking wait. Terminating the child is left to `ShellCommand`'s existing destructor teardown, exactly as before this change. The default `check_exit_code=true` path already reaps via `prepare`'s blocking `wait`, so `rusage` is captured there.

- Improvement

Populate the `ExecutableUserDefinedFunction*` `ProfileEvents` counters for `executable` UDFs. Per-call invocation count, wall time, child-process CPU and peak memory, and stdin/stdout bytes are now visible in `system.query_log.ProfileEvents`.

### Version info
- Merged into: `26.6.1.572`


andreev-io:
### Benchmark: `check_exit_code=false` rusage capture vs. post-EOF child lifetime

On the `check_exit_code=false` path the child is reaped with a non-blocking `WNOHANG` `wait4` in `cleanup`, so CPU/peak-RSS are captured only if the child is already a zombie when that poll runs. I measured how long a child can keep running *after* closing stdout before the measurement is lost (8 runs each, single-row call):

| post-close work before `exit(0)` | captured |
|---|---|
| a few instructions (~µs) | 8/8 |
| 50 ms | 8/8 |
| ~90 ms (busy loop) | 8/8 |
| 200 ms | 8/8 |
| 500 ms | 0/8 |
| ≥ 1 s | 0/8 |

The poll doesn't run at EOF — pipeline-pull completion, `prepare`, send-thread joins and source teardown sit in between (~0.2–0.5 s here), so any child that wraps up within that window is already a zombie when polled (its post-close work even shows up in the captured rusage). The cliff is between 200 ms and 500 ms.

Takeaway: realistic "write output, do a little cleanup, `exit`" scripts are captured; a measurement is lost only if a child closes stdout and then lingers ~½ s or more (the `sleep(30)` integration test being the extreme). The window is incidental (varies by machine/load), not a guarantee — the contract is just "captured if exited before the poll" — but it's wide enough that the pure non-blocking reap is kept rather than adding a bounded wait into the teardown path.

src/Common/ShellCommand.cpp
```diff
@@ -16,6 +29,19 @@
 #include <Common/waitForPid.h>
 
 
+namespace DB
+{
+
+/// Opaque wrapper around `::rusage` so callers of `ShellCommand` do not need
+/// to include `<sys/resource.h>`.
+struct LastChildResourceUsage
```
HanziJiang: Curious, why the "last" in LastChildResourceUsage or last_resource_usage? ShellCommand manages one child, and we capture the resource usage at most once. imo having "last" in the names is a little confusing, do we want to drop it?

src/Functions/UserDefined/UserDefinedExecutableFunctionFactory.cpp
```diff
@@ -283,16 +282,24 @@ class UserDefinedFunction final : public IFunction
                 if (sampler->poolWaitDone())
                     ProfileEvents::increment(ProfileEvents::ExecutableUserDefinedFunctionPoolWaitMicroseconds, sampler->getPoolWaitMicroseconds());
 
-                /// The remaining counters depend on the worker actually being
-                /// borrowed and observable via `/proc/<pid>`.
-                if (sampler->borrowAcquired())
+                /// Byte counters are recorded by the pipe buffers as data streams through
+                /// (`recordInputBytes` / `recordOutputBytes`) and are valid for any call
+                /// that performed I/O, independent of whether the child was reaped.
+                ProfileEvents::increment(ProfileEvents::ExecutableUserDefinedFunctionInputBytes, sampler->getInputBytes());
```
HanziJiang, line 288: 👍

src/Common/ShellCommand.cpp
```diff
@@ -402,11 +449,58 @@ bool ShellCommand::waitIfProccesTerminated()
...
+
+
 void ShellCommand::wait()
 {
     int retcode = tryWaitImpl(true).retcode;
     handleProcessRetcode(retcode);
 }
 
 
+bool ShellCommand::wasChildReaped() const noexcept
```
HanziJiang: Maybe rename it to something like `wasChildResourceUsageCaptured`? because in the non udf paths, child could be reaped but last_resource_usage is null

src/Common/UDFProcessSubtreeSampler.h
HanziJiang: These comments are all for executable pool, we should update it to include executable as well.

src/Common/ProfileEvents.cpp
```diff
@@ -1365,13 +1365,13 @@ The server successfully detected this situation and will download merged part fr
...
-    M(ExecutableUserDefinedFunctionUserTimeMicroseconds, "User mode CPU time consumed by executable_pool user-defined function child processes, in microseconds.", ValueType::Microseconds) \
-    M(ExecutableUserDefinedFunctionSystemTimeMicroseconds, "Kernel mode CPU time consumed by executable_pool user-defined function child processes, in microseconds.", ValueType::Microseconds) \
-    M(ExecutableUserDefinedFunctionPeakMemoryByteSeconds, "Per-process peak memory used by executable_pool user-defined function child processes integrated over wall clock time, in byte-seconds.", ValueType::Number) \
-    M(ExecutableUserDefinedFunctionInputBytes, "Total bytes written to the stdin of executable_pool user-defined function child processes.", ValueType::Bytes) \
-    M(ExecutableUserDefinedFunctionOutputBytes, "Total bytes read from the stdout of executable_pool user-defined function child processes.", ValueType::Bytes) \
+    M(ExecutableUserDefinedFunctionInvocations, "Number of executable and executable_pool user-defined function invocations.", ValueType::Number) \
```
HanziJiang, line 1368: Wanted to clarify my understanding: this increment per executeImpl (block of rows), not per user query, right?

XuJia0210:
Overall, it looks good to me.
There are 2 small issues you can fix.

  1. **(blocking)** `test_check_exit_code_false_lingering_child_is_bounded_and_flushes_bytes` fails deterministically (~30s, asserts `<25`). The
  30s is stdout EOF-wait — pre-existing master behavior, not introduced here (this PR touches none of the fork/pipe/EOF/teardown paths). The
  bounded-teardown assumption isn't reachable for a close-stdout-then-linger child. → drop the timing assertion (keep the byte-flush check) or
  change the scenario.

  2. **(should fix)** `ElapsedMicroseconds` is gated on a successful reap → recorded as 0 in the lingering case. Record it unconditionally; gate
  only CPU/peak-RSS on `wasChildReaped()`. (Matches the open bot comment.)

XuJia0210 (approved):

src/Common/ProfileEvents.cpp
andreev-io: Correct!

src/Common/ShellCommand.cpp
andreev-io: Agree, dropped "last"

src/Common/ShellCommand.cpp
andreev-io: Done — renamed to `wasChildResourceUsageCaptured`.

src/Common/UDFProcessSubtreeSampler.h
andreev-io: Done — the class doc now covers both the pool (`/proc` subtree sampling) and executable (`wait4` rusage) paths, with each path's lifecycle callbacks.

src/Common/waitForPid.cpp
```diff
@@ -232,17 +235,14 @@ bool waitForPid(pid_t pid, size_t timeout_in_seconds)
...
             return false;
 
-        watch.restart();
-
-        PollPidResult result = pollPid(pid, timeout_in_ms);
+        const Int64 remaining_ms = total_timeout_ms - static_cast<Int64>(watch.elapsedMilliseconds());
+        if (remaining_ms <= 0)
+            return false;
 
+        PollPidResult result = pollPid(pid, static_cast<int>(remaining_ms));
```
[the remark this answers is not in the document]
andreev-io: Good catch — fixed in `dd14aa57d76`. `pollPid` now releases the fd via `SCOPE_EXIT` immediately after it is opened, so the `pidfd`/`/proc` fd (Linux) and the `kqueue` fd (macOS/FreeBSD) are closed on every return path, including the `poll`/`kevent` timeout and filter-add error paths. The SunOS `Pgrab`/`Pfree` path already covered all returns.

src/Common/ShellCommand.cpp
```diff
@@ -310,7 +336,7 @@ int ShellCommand::tryWait()
     return tryWaitImpl(true).retcode;
 }
 
-ShellCommand::tryWaitResult ShellCommand::tryWaitImpl(bool blocking)
+ShellCommand::tryWaitResult ShellCommand::tryWaitImpl(bool blocking, bool check_exit_status)
```
vdimir, line 326: Is it that we now call `tryWaitImpl` multiple times with different flags - first as a regular wait with `check_exit_status`, and later from cleanup just to collect statistics? If the process has already terminated, does `wait4` handle that correctly?

src/Common/ShellCommand.cpp
```diff
@@ -1,3 +1,16 @@
+/// `wait4` is declared under `_DEFAULT_SOURCE` on Linux glibc, which the
+/// `-std=c++23` strict mode otherwise hides. Define it before the first system
+/// header that guards it. It is a libc feature-test macro, hence the reserved
+/// name; suppress the diagnostics that would otherwise reject our own define.
+#if defined(OS_LINUX) && !defined(_DEFAULT_SOURCE)
```
vdimir, line 5: It looks a bit awkward at first glance, but there don't seem to be many other options.

src/Common/waitForPid.cpp
```diff
@@ -232,17 +254,14 @@ bool waitForPid(pid_t pid, size_t timeout_in_seconds)
         if (waitpid_res != 0)
             return false;
 
-        watch.restart();
-
-        PollPidResult result = pollPid(pid, timeout_in_ms);
+        const Int64 remaining_ms = total_timeout_ms - static_cast<Int64>(watch.elapsedMilliseconds());
```
vdimir, line 260: Was it a bug that we waited more then remaning timeout in pollPid? Does it affect only executable udfs?

vdimir (approved):

src/Processors/Sources/ShellCommandSource.cpp
```diff
@@ -274,7 +274,13 @@ class TimeoutReadBufferFromFileDescriptor : public BufferWithOwnMemory<ReadBuffe
                 {
                     bytes_read += res;
                     if (sampler)
+                    {
                         sampler->recordOutputBytes(static_cast<size_t>(res));
+                        /// Child is provably alive — it just wrote output.
+                        /// Sample its subtree VmHWM for the executable path;
+                        /// no-op on the pool path (executable_root_pid <= 0).
+                        sampler->sampleExecutablePeak();
```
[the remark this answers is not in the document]
andreev-io: Addressed by throttling. `sampleExecutablePeak` previously ran a `/proc` subtree walk plus a per-pid `VmHWM` read on every IO buffer fill; it is now throttled to at most one walk per ~5 ms, plus a guaranteed first sample and a best-effort final sample at stdout EOF. `VmHWM` is a monotonic high-water mark, so sparse sampling still captures each process's true peak.

Measured per-walk cost on the target hardware (100k iterations, replicating `walkSubtree` + `readPeakRss`): ~34 µs for a single-process UDF and ~104 µs for a 3-process subtree, and the work is ~100% CPU (mostly kernel time generating `/proc/<pid>/{status,task}`). Throttled to one walk per 5 ms, the worst case — a UDF streaming continuously for the whole query — is ~0.7% of one core for a single-process UDF and ~2% for a forking one, and it no longer scales with input/output volume. The old per-buffer cost did: a UDF emitting 100 MiB of output performed ~25k walks.

andreev-io:
### Memory-sampling overhead, measurements, and mitigations

**Concern.** The executable-UDF peak-memory metric samples `/proc/<pid>/VmHWM` across the child subtree during IO. Initially this ran a full subtree walk plus a per-pid `VmHWM` read on every IO buffer fill, so a UDF streaming large output performed tens of thousands of walks (≈25k for 100 MiB of output), and the sampler is installed for every executable UDF.

**Measurements** (on the target hardware, 100k iterations replicating `walkSubtree` + `readPeakRss`):

| Subtree | Per walk | Nature |
|---|---|---|
| Single-process UDF | ~34 µs | ~100% CPU (`thread_cpu` ≈ wall), ~90% kernel time generating `/proc` |
| UDF + 2 children (3 pids) | ~104 µs | same |

The cost is genuine CPU on the IO thread (confirmed with `CLOCK_THREAD_CPUTIME_ID`), almost entirely kernel time synthesising `/proc/<pid>/{status,task}` — there is no device IO to block on.

**Mitigations.**
- **Throttle** to at most one walk per ~5 ms. `VmHWM` is a monotonic high-water mark, so sparse sampling still captures each process's true peak; tightening the interval does not improve accuracy for the processes that matter (the UDF and any long-lived helper are observed across many samples regardless).
- **Bounded boundary samples:** the first call always samples (start) and one best-effort sample is taken at stdout EOF (finish).
- Result: a worst case — a UDF streaming continuously for the whole query — of ~0.7% of one core (single-process) to ~2% (forking), and the cost no longer scales with input/output volume.

**Known limitation.** Because `/proc` is observed only at these throttled sample points, a short-lived child that allocates and is reaped between two samples is best-effort and may be under-counted. A process alive at least once after reaching its peak is captured exactly (monotonic `VmHWM`). This is documented on `sampleExecutablePeak`.

src/Common/waitForPid.cpp
andreev-io: Yes, there was a bug. Previously, every signal restarted the timeout from zero.

No, this change actually isn't specific to executable UDFs. It affects every user of ShellCommand (executable dictionaries, table functions, the pool, executable UDFs). Apparently any of them could over-wait, for example when query profiler was enabled and returned SIGPROF.

src/Common/ShellCommand.cpp
andreev-io: > Is it that we now call tryWaitImpl multiple times with different flags - first as a regular wait with check_exit_status, and later from cleanup just to collect statistics?

I added this because a UDF can be launched with `check_exit_code=false` but we still want to reap to collect resource usage. Previously, if `check_exit_code` was `false`, we wouldn't have reaped.

> multiple times with different flags

If the first time we called we have check_exit_status true, we don't call again. Otherwise, we call again in `cleanup` to reap. This is guarded by checking `command->isWaitCalled()`. In any case, we reap at most once (we reap zero times if `check_exit_code` is false and then the process keeps running for longer than we wait in `tryWaitProcessWithTimeout`).

> If the process has already terminated, does wait4 handle that correctly?

Yes, it should be OK to reap on a terminated process. The only problem would be if reaped twice, but we don't do that.

andreev-io: > I added this because a UDF can be launched with check_exit_code=false but we still want to reap to collect resource usage.

Sorry, this isn't exactly accurate. Without this parameter, the reap would validate the exit status and throw CHILD_WAS_NOT_EXITED_NORMALLY on a non-zero exit. check_exit_code=false UDFs are allowed to fail, so we need to reap for rusage without checking the status. That's what this parameter makes possible.

src/Common/waitForPid.cpp
andreev-io: Checked every caller for blast radius. `waitForPid` is reached only from `ShellCommand`'s destructor teardown (`tryWaitProcessWithTimeout`), and the changed poll loop runs **only when the timeout is positive** — a `0` timeout takes a single non-blocking `waitpid` fast path that this diff doesn't touch.

- The odbc/library bridges (`IBridgeHelper`) and the client's pager/editor (`ClientBase`) pass timeout `0`, so they hit that fast path and never reach the changed code.
- Only `ShellCommandSource` — executable UDFs and executable table functions, default `command_termination_timeout=10s` — passes a positive timeout and reaches the loop.
- Everything else uses `terminate_in_destructor=false` and doesn't call `waitForPid` at all.

So it's not UDF-specific (it also covers executable table functions), but the only behavioral change is that those teardowns now honor `command_termination_timeout` instead of over-waiting. The overshoot only happened under signals (the profiler's `SIGPROF`) and contradicted the timeout's documented meaning, so nothing relied on it.

Issue #79, opened 2016-08-22
problem with loading data


trivita:
i know NULL isnt supported by clickhouse. but if i have a csv with ,,, line in it. why cant clickhouse just insert data with default values?

alexey-milovidov:
Acknowledged. Will do.

alexey-milovidov:
It is already supported for numeric types and String. But not for Date, DateTime.

trivita:
> It is already supported for numeric types and String. But not for Date, DateTime.

thanks, you mean the git version?

alexey-milovidov:
> thanks, you mean the git version?

For numeric types and String, empty is parsed as zero/empty string (this behaviour was from the beginning).
Example:

```
:) CREATE TABLE test.csv (a String, b UInt64, c Int64, d String) ENGINE = Log

...

$ echo ',,,' | clickhouse-client --query="INSERT INTO test.csv FORMAT CSV"

...

:) SELECT * FROM test.csv

┌─a─┬─b─┬─c─┬─d─┐
│   │ 0 │ 0 │   │
└───┴───┴───┴───┘
```

But it is not implemented for Date, DateTime, Array and other data types.
If you have empty instead of Date/DateTime, parser will look for next field at 10/19 bytes forward, and will be confused with it.

I am going to implement the intended behaviour for Date, DateTime too.

trivita:
ehh, yeah, but not seem to work with TabSeparatedWithNames

alexey-milovidov:
Could you please provide complete example, because trivial examples working:

`CREATE TABLE test.tsv (a String, b UInt64, c Int64, d String) ENGINE = Log`

```
$ echo -e '\t\t\t' | clickhouse-client --query="INSERT INTO test.tsv FORMAT TabSeparated"
$ echo -ne 'a\tb\tc\td\n\t\t\t\n' | clickhouse-client --query="INSERT INTO test.tsv FORMAT TabSeparatedWithNames"
```

Maybe it works just accidently.

pluk:
Hi.

> If you have empty instead of Date/DateTime, parser will look for next field at 10/19 bytes forward, and will be confused with it.

I have several consecutive DateTime fields and It behaves like this
`CREATE TABLE test
(
    id UInt32, 
    entity UInt8, 
    add_date DateTime, 
    mod_date DateTime, 
    date Date
) ENGINE = MergeTree(date, (id, date), `8192)``
...
[...]

PR #105623, opened 2026-05-22
Reduce per-process arena drift noise in perf-comparison tests

Algunenano:
Refs #100759. Follow-up to the noise spike attributable to #104136.

The LEFT/RIGHT server pair in performance-comparison runs accumulates asymmetric jemalloc dirty-page state because (1) `SYSTEM JEMALLOC PURGE` only fired between distinct queries, not between measured runs, and (2) measured runs were always `LEFT, RIGHT` in that order, giving one side ~100ms more "idle accumulation" before its next purge than the other.

Two changes:

1. New `--jemalloc-purge` option in `perf.py` (`disabled` / `after-fill` / `before-each-query` / `before-each-run`, default `before-each-run`). The old per-query purge becomes one of the modes.
2. Alternate the LEFT/RIGHT order each measured run (`L,R,R,L,L,R,...`) so the post-query / pre-purge idle time averages out.

Local validation on a slow disk (`/mnt/db`, 113 MB/s) over 8 perf.py iterations of `compact_part_subcolumns`:

| config | q1 stddev | q1 over 10% gate |
|---|---|---|
| original (purge per query, fixed order) | 19.5% | 11/15 (73%) |
| new default (purge per run, alternating order) | 4.3% | 1/8 (12.5%) |

Roughly the same noise floor as disabling the arena entirely.


Algunenano:
Comparing noise rate across 3 CI re-runs of this branch vs master baseline (last 4 days):

**Per-attempt noise rate:**

| run | arch | noise rate |
|---|---|---|
| Attempt 1 (13:00) | arm | 0.339% |
| Attempt 2 (16:00) | arm | 0.339% |
| Attempt 3 (20:00) | arm | 0.169% |
| Attempt 3 (20:00) | amd | 0.254% |

**Aggregate vs master baseline:**

| arch | this PR | master baseline | reduction |
|---|---|---|---|
| amd | 0.254% (1 run) | 0.469% | −46% |
| arm | 0.282% (3 runs avg) | 0.520% | −46% |

All three ARM runs land below master baseline, so the reduction holds across attempts rather than being a one-off.

**Persistently flagged tests** (across the 3 PR ARM runs + 1 AMD run), and their pre-existing master noise rate (last 4 days):

| test | flagged on PR | master baseline noise | note |
|---|---|---|---|
| `min_max_index_with_cast #0` (arm) | slower in 3/3 runs | 0% | possibly attributable to PR (or first-vs-second ordering artifact) |
| `aggregate_functions_of_group_by_keys #1` (arm) | unstable in 2/3 | 0% | new on PR |
| `reading_from_file #0` | unstable in 2/3 | 45% arm / 31% amd | pre-existing master flake |
| `apply_patch_parts_join #4/#5/#6` | unstable | 10-24% arm | pre-existing (PR #105527 fix wasn't merged at run time) |
| `classification #0/#1` (amd) | slower 2× | 65% amd | pre-existing master flake |
| `top_k_through_join #2` (amd) | unstable | 24% amd | pre-existing |
| `logical_functions_small #0/#4`, `count_from_formats #79`, `general_purpose_hashes #81` | slower / unstable 2× | 0% | one-time CI variance |

The flakes on this PR are dominated by tests that were already flaky on master (`reading_from_file`, `apply_patch_parts_join`, `classification`).

tests/performance/scripts/perf.py
```diff
@@ -592,7 +610,16 @@ def do_create(connection, index, queries):
     while True:
         run_id = f"{query_prefix}.run{run}"
 
-        for conn_index, c in enumerate(this_query_connections):
+        if args.jemalloc_purge == "before-each-run":
+            purge_jemalloc_on_all_connections(f"before-{run_id}")
+
+        # Alternate the server order each measured run (L,R,R,L,L,R,...) so that
```
nickitat, line 627: Maybe we could do something like that (but on a per-run basis)

``` diff
diff --git a/tests/performance/scripts/perf.py b/tests/performance/scripts/perf.py
     index 181c65984ea..8dddd074ce7 100755
     --- a/tests/performance/scripts/perf.py
     +++ b/tests/performance/scripts/perf.py
     @@ -16,6 +16,7 @@ import sys
      import time
      import traceback
      import xml.etree.ElementTree as et
     +from concurrent.futures import ThreadPoolExecutor
      from threading import Thread

      import clickhouse_driver
     @@ -389,6 +390,20 @@ servers = [
      all_connections = [
          clickhouse_driver.Client(**server, settings_is_important=True) for server in servers
      ]
     +# Long-lived workers so we don't pay pthread_create per query when fanning out
     +# parallel commands across all servers (e.g. SYSTEM JEMALLOC PURGE).
     +purge_pool = ThreadPoolExecutor(max_workers=max(1, len(all_connections)))
     +
     +
     +def purge_jemalloc(indexed_conn):
     +    conn_index, c = indexed_conn
     +    try:
     +        c.execute("SYSTEM JEMALLOC PURGE")
     +        return f"purging jemalloc arenas\t{conn_index}\t{c.last_query.elapsed}"
     +    except KeyboardInterrupt:
     +        raise
     +    except:
     +        return None

      for i, s in enumerate(servers):
          ssl_status = "SSL" if s["secure"] else "no-SSL"
     @@ -501,15 +516,9 @@ for query_index in queries_to_run:

          print(f"display-name\t{query_index}\t{tsv_escape(query_display_name)}")

     -    for conn_index, c in enumerate(all_connections):
     -        try:
     -            c.execute("SYSTEM JEMALLOC PURGE")
     -
     -            print(f"purging jemalloc arenas\t{conn_index}\t{c.last_query.elapsed}")
     -        except KeyboardInterrupt:
     -            raise
     -        except:
     -            continue
     +    for line in purge_pool.map(purge_jemalloc, enumerate(all_connections)):
     +        if line:
     +            print(line)

          # Prewarm: run once on both servers. Helps to bring the data into memory,
          # precompile the queries, etc.
```

Algunenano: Yes, it makes sense to purge in parallel. It should reduce noise and make things faster. I'll add it

nickitat (approved):

Algunenano:
Summary after the last change:
```
One ARM CI run so far on the parallel-purge commit (00f63a4e5e):

┌────────────┬───────────────────────────────────────┐
│   metric   │                 value                 │
├────────────┼───────────────────────────────────────┤
│ Noise rate │ 0.167% (4 slower + 4 unstable / 4804) │
├────────────┼───────────────────────────────────────┤
│ AMD runs   │ 0 (not triggered yet)                 │
└────────────┴───────────────────────────────────────┘

That's at the low end of what we saw on the previous 3 ARM runs (which ranged 0.169%-0.339%, avg 0.282%). One sample isn't conclusive but it's not worse.

Tests flagged (4 unique on ARM):
- min_max_index_with_cast #0 — slower (also flagged on previous 3 runs)
- clickbench_parquet_short #5 — unstable
- parallel_final #9 — unstable
- sort_wide_table #2 — slower

So min_max_index_with_cast #0 is still flagged as slower, which means the parallel purges didn't eliminate that specific bias — confirming it isn't purely a purge-ordering asymmetry. The other 3 tests are different from prior runs (one-time variance).
```

Let's merge and I'll keep an eye.

PR #105624, opened 2026-05-22
check_ci.py: enqueue PR via GraphQL instead of broken auto-merge

Algunenano:
`gh pr merge --auto` (called by `check_ci.py` to add a PR to the merge queue) hits the `enablePullRequestAutoMerge` GraphQL mutation. That mutation is disabled on this repo because `master` uses a merge queue, so the call fails with `Auto merge is not allowed for this repository (enablePullRequestAutoMerge)`. `gh` v2.92 detects the queue and warns about it, but still issues the auto-merge mutation rather than `enqueuePullRequest`, so the workflow has been broken in practice.

This PR calls `enqueuePullRequest` directly via `gh api graphql` — the same mutation github.com's "Merge when ready" button uses. It also drops the `CLEAN`-state retoggle workaround, which existed to nudge the auto-merge feature when checks had already passed and does not apply to the merge queue.

...

### Version info
- Merged into: `26.6.1.89`

### DIFF
--- a/ci/jobs/scripts/check_ci.py
+++ b/ci/jobs/scripts/check_ci.py
@@ -1040,49 +1040,47 @@ def main():
         mergeable_check_status, sha=head_sha
     )
 
-    auto_merge_cmd = f"gh pr merge {pr_number} --auto --repo ClickHouse/ClickHouse"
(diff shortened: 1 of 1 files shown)


leshikus (approved):
Looks good

PR #103862, opened 2026-05-01
Add max_bytes_ratio_before_external_join setting

alexey-milovidov:
Adds a `max_bytes_ratio_before_external_join` setting that mirrors the existing `max_bytes_ratio_before_external_group_by` and `max_bytes_ratio_before_external_sort` settings.

The new setting expresses the spill threshold as a fraction of available memory, which is convenient when an absolute byte value (`max_bytes_before_external_join`) is hard to choose. If both the absolute and the ratio settings are non-zero, the smaller resulting threshold wins.

The combination is performed once when constructing `JoinSettings` and `TableJoin`, so the rest of the join code keeps reading a single effective threshold via `TableJoin::maxBytesBeforeExternalJoin`.

Default value is `0` (disabled), so the behavior is unchanged unless the setting is explicitly set, matching `max_bytes_before_external_join` itself. The default may be flipped to `0.5` later, mirroring how `max_bytes_ratio_before_external_group_by` and `max_bytes_ratio_before_external_sort` were rolled out.

### Changelog category (leave one):
- New Feature

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Added a new setting \`max_bytes_ratio_before_external_join\`, mirroring \`max_bytes_ratio_before_external_group_by\` and \`max_bytes_ratio_before_external_sort\`. It expresses the spill-to-disk threshold for hash joins as a fraction of available memory; combined with the absolute \`max_bytes_before_external_join\`, the smaller resulting threshold applies.

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.371`

PR #100176, opened 2026-03-20
Limit simultaneously active streams in UNION ALL to reduce peak memory

alexey-milovidov:
A UNION ALL query with many branches opens read buffers for all branches simultaneously, causing memory proportional to num_branches × num_columns × buffer_size. With ~100 branches on a wide table this easily reaches 10+ GB for no good reason.

Add two settings that narrow the `UnionStep` pipeline via `ConcatProcessor`s so that only a bounded number of streams are active at once:

- `max_streams_for_union_step` (UInt64, default 0): absolute limit
- `max_streams_for_union_step_to_max_threads_ratio` (Float, default 8): ratio × `max_threads` gives a dynamic limit

The actual limit is min(absolute, ratio×threads); either being 0 means it is ignored; both 0 disables narrowing entirely.

Closes https://github.com/ClickHouse/ClickHouse/issues/23245

### Changelog category (leave one):
- Performance Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Limit simultaneously active streams in UNION ALL to reduce peak memory usage.

### Version info
- Merged into: `26.5.1.199`

PR #105630, opened 2026-05-22
Randomize more settings in clickhouse-test: dictGet tuple element, top-k var-length filtering, page cache coalescing, grace hash buckets

zlareb1:
This is a test-only change to `tests/clickhouse-test`. It extends `SettingsRandomizer` to cover several recently-introduced settings that were not yet randomized:

- `optimize_dictget_tuple_element` (#100186): `random.random() > 0.05` — default-true, disabled with 5% probability, matching the other default-true entries.
- `use_top_k_dynamic_filtering_for_variable_length_types` (#104216): `random.randint(0, 1)`, added to `conditional_settings` parented on `use_top_k_dynamic_filtering` — the variable-length opt-in is meaningless when the parent dynamic filter is off.
- `page_cache_max_coalesced_bytes` (#104230): `random.choice([65536, 1048576, 16777216, 67108864])` — covers small values (the C++ uses `max(1, value / block_size)`, so sub-block values collapse to one block per fetch) and large values (the cap that splits long miss runs into multiple reads).

It also re-enables randomization of `grace_hash_join_initial_buckets`, which had been commented out previously. It is restored with the narrow, previously-working `[1, 2, 4, 8]` rather than the wider `[1, 2, 4, 8, 16, 32, 64, 128]`; the range can be widened in a follow-up if CI stays green.

Deferred to a follow-up (they need groundwork before they can be randomized safely):

- `max_streams_for_union_step` / `max_streams_for_union_step_to_max_threads_ratio` (#100176): randomizing these reshaped the `UNION ALL` pipeline and broke several stateless tests by surfacing pre-existing weak-ordering invariants (sorting on constant keys, NaN sort keys such as `sin(0/0)`, `ORDER BY tuple()`, duplicate values without a secondary sort). The dedicated `04051_max_streams_for_union_step` test also appeared to have session-level randomized values leak past explicit per-query `SETTINGS` clauses — worth investigating separately.
- `max_bytes_ratio_before_external_join` (#103862): pairing it with `max_bytes_before_external_join` caused `03279_join_choose_build_table` to fail under spilling (`JoinResultRowCount` returned 500/125 instead of the asserted 1000), which looks like a real grace-hash-join-under-spill correctness issue.

...

N/A — test-only change.

### Version info
- Merged into: `26.6.1.387`


Ergus (approved):

Ergus (changes requested):
The agent comments are correct

Ergus:
Hi @zlareb1 

Is this still a DRAFT?

zlareb1:
@Ergus yes, for now. There were some test failures bcz of few added settings so need more dry runs. I think i will be able to mark this ready this week.

zlareb1:
@Ergus can you please stamp this?

Ergus (approved):
LGTM

PR #105633, opened 2026-05-22
Fix UB in 32-bit shift before widening to UInt64

### LINKED ISSUE #105629: [SAST] Potential undefined behavior in bitwise shift: (1 << result_bit_idx) and (1 << cap) use signed int before widening to UInt64
### Describe the bug

Hello! We were running static analysis on ClickHouse with Svace and found two instances of potential undefined behavior in bitwise shift operations in `MortonUtils.h:198` and `hasLinuxCapability.cpp:64`.

The issue occurs when shifting the literal `1` (type `int`) by a value that may exceed 31, even though the result is later stored in a `UInt64`. The shift operation itself is performed in signed 32-bit integer arithmetic before any widening occurs, leading to undefined behavior for shift counts ≥ 31.

### First location: MortonUtils.h

https://github.com/ClickHouse/ClickHouse/blob/131ef1d5bc85e8abff29a1fda82ed59ab398615a/src/Common/MortonUtils.h#L194-L199

### Second location: hasLinuxCapability.cpp

https://github.com/ClickHouse/ClickHouse/blob/131ef1d5bc85e8abff29a1fda82ed59ab398615a/src/Common/hasLinuxCapability.cpp#L61-L65

###  How to reproduce

The undefined behavior manifests when:

    In MortonUtils: result_bit_idx >= 31

    In hasLinuxCapability: cap >= 31

This can exceed the bit-width of a signed 32-bit integer, potentially leading to an out-of-bounds shift which constitutes undefined behavior.

### Suggested fix

```cpp
// MortonUtils.h
last |= (UInt64(1) << result_bit_idx);
```
```cpp
// hasLinuxCapability.cpp
return (UInt64(1) << cap) & capabilities.effective;
```
Please consider addressing these cases at your convenience. Thank you!

Found by[ Linux Verification Center ](https://portal.linuxtesting.ru/)with SVACE.

Algunenano:
Closes #105629.

Both call sites computed `1 << n` in signed 32-bit arithmetic before widening the result to `UInt64`. The shift count can reach values where this is undefined behavior:

- `src/Common/hasLinuxCapability.cpp:64` — Linux capabilities `>= 31` exist (`CAP_MAC_OVERRIDE`, `CAP_SYSLOG`, `CAP_AUDIT_READ`, `CAP_PERFMON`, `CAP_BPF`, `CAP_CHECKPOINT_RESTORE`, ...). Current in-tree callers pass values `< 31`.
- `src/Common/MortonUtils.h:198` — only reached from unit tests; the function is not used by the server.

...


antonio2368 (approved):

Algunenano:
Tried to debug the coredump (which is unrelated) but couldn't because it's truncated -> https://github.com/ClickHouse/ClickHouse/pull/105631

mysql test fixed here: https://github.com/ClickHouse/ClickHouse/pull/105636

PR #105637, opened 2026-05-22
Make `apply_row_policy_after_final` consistent accross versions

yariks5s:

This PR modifies the settings history so that compatibility='X.Y' will never revert apply_row_policy_after_final to false. Previously, compatibility levels before 26.2 would revert the setting to false (the buggy pre-fix behavior). After this change, the setting remains true regardless of compatibility level, ensuring users always get the correct row policy application semantics with FINAL.

---

> [!NOTE]
> **Medium Risk**
> Adjusts the `compatibility` settings-change history for `apply_row_policy_after_final`, which can affect when row policies/PREWHERE are applied around `FINAL` and thus query results and access control semantics under older compatibility modes.
> 
> **Overview**
> Makes `apply_row_policy_after_final` consistent across versions in `SettingsChangesHistory`: removes the prior 25.12/26.2 history entries (including the default-enable note) and records it under `25.10` as `{true, true}` so `compatibility` never reverts to the pre-fix `false` behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab4b3f440507474e2bebbbdeb0f5420fb8af4929. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.108`
- Backported to: `26.5.2.6`, `26.4.4.10`, `26.3.13.6`, `26.2.20.2`


clickgapai:
Hi @yariks5s @fm4v — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: The description is mostly auto-generated by Cursor Bugbot and doesn't explain the actual user-visible change: users who previously relied on compatibility='25.X' getting apply_row_policy_after_final=false will now get true. A sentence explaining this behavior change and why it's desirable would help reviewers.
- Suggested addition:

  > This PR modifies the settings history so that `compatibility='X.Y'` will never revert `apply_row_policy_after_final` to `false`. Previously, compatibility levels before 26.2 would revert the setting to false (the buggy pre-fix behavior). After this change, the setting remains true regardless of compatibility level, ensuring users always get the correct row policy application semantics with FINAL.

PR #105684, opened 2026-05-22
Make unit tests compile with -DENABLE_LIBRARIES=OFF

bobrik:
Previously linker failed for this:

```
$ cmake -DCMAKE_BUILD_TYPE=Debug -DENABLE_LIBRARIES=OFF -DENABLE_RUST=OFF ..
$ ninja unit_tests_dbms
```

```
ld.lld-22: error: undefined symbol: DB::CompactChildrenSet::empty() const
>>> referenced by gtest_compact_children_set.cpp:17 (src/Coordination/tests/gtest_compact_children_set.cpp:17)
>>>               src/CMakeFiles/unit_tests_dbms.dir/Coordination/tests/gtest_compact_children_set.cpp.o:(CompactChildrenSet_EmptyState_Test::TestBody())
>>> referenced by gtest_compact_children_set.cpp:28 (src/Coordination/tests/gtest_compact_children_set.cpp:28)
>>>               src/CMakeFiles/unit_tests_dbms.dir/Coordination/tests/gtest_compact_children_set.cpp.o:(CompactChildrenSet_EmptyToSingleTransition_Test::TestBody())
>>> referenced by gtest_compact_children_set.cpp:47 (src/Coordination/tests/gtest_compact_children_set.cpp:47)
>>>               src/CMakeFiles/unit_tests_dbms.dir/Coordination/tests/gtest_compact_children_set.cpp.o:(CompactChildrenSet_SingleToSetPromotion_Test::TestBody())
>>> referenced 20 more times
```

Fix building `unit_tests_dbms` when ClickHouse is configured with `-DENABLE_LIBRARIES=OFF` (and `-DENABLE_RUST=OFF`).

---

> [!NOTE]
> **Low Risk**
> Low risk build-system change that only affects how unit tests are linked when `ch_contrib::nuraft`/`Coordination` is not built.
> 
> **Overview**
> Ensures `unit_tests_dbms` still links when `Coordination` (and its `CompactChildrenSet` implementation) is not included via `dbms` by explicitly adding `Coordination/CompactChildrenSet.cpp` to the test target when `ch_contrib::nuraft` is absent.
> 
> This fixes builds with configurations like `-DENABLE_LIBRARIES=OFF` where the `gtest` sources for `CompactChildrenSet` are still compiled but the implementation object file was previously missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c4cd9c9f19e0316ddcec76273f4fe56cbbe2bc8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
Workflow [[PR](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105684&sha=latest&name_0=PR)]
Sync PR [[sync-upstream/pr/105684](https://github.com/search?q=head%3Async-upstream%2Fpr%2F105684+org%3AClickHouse+type%3Apr&type=pullrequests)]


src/CMakeLists.txt
```diff
@@ -432,6 +432,11 @@ endif()
 
 if (TARGET ch_contrib::nuraft)
     add_object_library(clickhouse_coordination Coordination)
+elseif (ENABLE_TESTS)
+    # `CompactChildrenSet` tests are picked up by the global `unit_tests_dbms`
```
alexey-milovidov: Can we fix it inside unit_tests_dbms? Otherwise the modification to this CMake files sounds too ad-hoc.

bobrik: I moved it closer to where it logically belongs.

alexey-milovidov: This thread is now stale: the ad-hoc \`src/CMakeLists.txt\` modification was dropped entirely. Instead, \`gtest_compact_children_set.cpp\` now guards itself with \`#include "config.h"\` + \`#if USE_NURAFT\`, exactly like every other Coordination test (\`gtest_coordination.cpp\`, \`gtest_coordination_storage.cpp\`, etc.). `USE_NURAFT` is set precisely when \`ch_contrib::nuraft\` exists, which is exactly when \`CompactChildrenSet.cpp\` is compiled, so the test is built iff its implementation is. No CMake change remains.

alexey-milovidov:
@groeneai, investigate the failure: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105684&sha=14e9ed2235dafbb8aa01b211bb42722ab346ce20&name_0=PR&name_1=LLVM%20Coverage and provide a fix in a separate PR. If the fix is already in progress, link it here.

The `LLVM Coverage` diff gate reported `Coverage degraded: master 86.20% → PR 28.60% (dropped 57.60 pp, tolerance 0.3 pp)`, while the detailed report posted by the same run says lines went from 86.53% to 86.54% (+0.01%) and 2/2 changed lines are covered. This PR only adds a `#if USE_NURAFT` guard to `src/Coordination/tests/gtest_compact_children_set.cpp`, so it cannot affect overall coverage. The same bogus gate failure hits unrelated pull requests as well (#110708 on 2026-07-21, #105844 on 2026-07-26), so it looks like a flaky baseline computation in the coverage-diff job rather than a real regression.

The other red job, `Integration tests (amd_asan_ubsan, db disk, old analyzer, 1/6)`, failed only on `test_backup_restore_on_cluster/test_cancel_backup.py::test_cancel_backup` with the `retry_ok` label; that test also fails on `master` and on many unrelated pull requests, so it is unrelated flakiness too.

groeneai:
The 57.60 pp failure on that report is already fixed on master, and it is a different defect from the two sub-1 pp ones you cited. Details, then what is left.

**The 57.60 pp case (this PR).** Not a flaky baseline: the PR side was measured from an incomplete profile set. At `14e9ed22` 20 of the 21 `amd_llvm_coverage` shards were `skipped` with `Skipped: only unit tests changed, stateless coverage not needed`, and only `Unit tests (amd_llvm_coverage)` ran. `merge_llvm_coverage.sh:62` merges `*.profdata`, i.e. whatever happens to be present, so `merged.profdata` held the unit-test profile alone and `llvm-cov export` yielded 28.60% against a complete 86.20% master baseline. Driving the `_classify_changed_tests` helper from `filter_job.py` at that sha with this PR's real file list (`src/Coordination/tests/gtest_compact_children_set.cpp`) returns `unit=True, stateless=False, integration=False, binary_unchanged=True`, which is exactly the branch that skipped the other two families. That is also why the per-file diff report looked healthy at 86.53% to 86.54%: the diff report reads the changed-file slice, which the unit test does cover, while the gate compares global percentages.

That skip block was removed by #111808 (`fb1887cc43fc35`, merged 2026-07-24), which replaced the per-test-type sub-job skipping with an all-or-nothing family skip. `fb1887cc43fc35` is an ancestor of this PR's current head `6979f44a`, so the 57.60 pp mode cannot recur here. CIDB agrees: 12 large-drop rows in 90 days, the last on 2026-07-21, none after that commit.

**The two you cited are a separate, still-live defect.** On #105844 (2026-07-26, 0.50 pp, after #111808 merged) all 21 coverage shards were `success`, so the profile set was complete and the gate still failed. On #110708 (0.60 pp) 19 shards succeeded and 2 failed, and `llvm-profdata merge -failure-mode=warn` silently proceeds without them. So the sub-1 pp population is ordinary run-to-run coverage variance plus partial-shard loss, measured against a 0.3 pp tolerance that is smaller than the noise: 68 of 80 `Coverage degraded` rows in 90 days are under 3 pp, spread over 72 distinct pull requests.

**One sharp bug inside that population.** 21 of those 80 rows report exactly `dropped 0.30 pp`, which the gate's own `b_line_cov - c_line_cov > TOLERANCE` should treat as passing. `lcov` reports 2-decimal percentages, and in binary floating point `85.40 - 85.10` is `0.30000000000001137`, so the strict comparison fires on a drop equal to the tolerance. Those 21 failures are pure float artifact.

I will open a separate PR for the live half. The direction is to make the gate refuse to compare when the profile set is not the complete expected one, rather than silently scoring a partial merge, and to compare on rounded values so an exactly-at-tolerance drop passes as documented. I will not widen the 0.3 pp tolerance, since that would hide real regressions instead of fixing the measurement. I will link the PR here.

One thing worth your call: the `hierarchical`/global-percentage gate is inherently noisy at 0.3 pp even with complete profiles, whereas the changed-lines number this job already computes is stable and is what a reviewer acts on. If you would rather the gate keyed on changed-line coverage instead of the global percentage, say so and I will build it that way.

The other red on that sha, `test_cancel_backup`, is unrelated to coverage and is tracked separately.

groeneai:
The first PR is up: https://github.com/ClickHouse/ClickHouse/pull/112482. It fixes the
exactly-at-tolerance class only, and I need to correct two things I got wrong above.

First, calling the sub-1 pp population "ordinary run-to-run coverage variance" was wrong. There is a
concrete bug inside it: `llvm_coverage_job.py` compares `b_line_cov - c_line_cov > 0.3` on values
parsed from lcov's one-decimal percentages, so the drop is a binary float subtraction of two
decimals. `84.4 - 84.1 == 0.30000000000001137`, which fails the strict `>` on a drop the gate's own
message calls allowed, while `85.5 - 85.2 == 0.29999999999999716` does not fail. It is
value-dependent, which is why it looked like noise. Of the 80 `Coverage degraded` rows in 90 days,
21 across 21 distinct pull requests report exactly `dropped 0.30 pp`, and they reduce to two value
pairs, `84.4/84.1` and `85.4/85.1`. Both pass with #112482 and the other 59 rows keep failing. I did
not widen the tolerance.

Second, I implied the baseline side is sound. It is not. Measuring the master runs, 24 of 39 publish
an incomplete baseline, one of them from a single shard out of 21. Since the gate subtracts the PR
side from the baseline, an incomplete baseline makes the comparison pass, so the same defect can mask
a real regression as well as invent one. Related: a `success` check-run does not imply an uploaded
profile. On #105844 all 21 shards were green but only 20 profiles existed, because `llvm-profdata`
crashed and the upload is `optional=True`. So my "the profile set was complete and the gate still
failed" reading of that run was also wrong.

The profile-set completeness half needs a cross-job artifact-completeness protocol rather than a
one-line change, so it is a separate pull request and I am working on it. On your suggestion to key
the gate on changed-line coverage instead of the global percentage: I still think that is the better
long-term shape, since the changed-line number is already computed and is stable, and it would make
both defects irrelevant to the verdict. If you prefer that direction I will build the second PR that
way instead of hardening the global comparison.

@alexey-milovidov

src/Coordination/tests/gtest_compact_children_set.cpp
```diff
@@ -1,3 +1,7 @@
+#include "config.h"
+
+#if USE_NURAFT
```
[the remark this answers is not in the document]
alexey-milovidov: 🕵 Addressed in `68be41973f58`. Both remaining Nuraft-dependent test translation units now carry the same `#include "config.h"` + `#if USE_NURAFT` guard as every other `Coordination` gtest:

- `src/Coordination/tests/gtest_keeper_asynchronous_metrics.cpp` (`setKeeperFileDescriptorMetrics`, defined in `Coordination/KeeperAsynchronousMetrics.cpp`)
- `src/Coordination/Storage/tests/gtest_keeper_storage.cpp` (`BlockData::appendNode`, `StorageState::startup`, defined in `Coordination/Storage/*.cpp`)

That covers every `gtest*.cpp` under `src` that references `Coordination/` — the other five were already guarded. All three guarded translation units were verified to still compile in a normal `ENABLE_LIBRARIES=ON` build. The branch was also merged up to current master in the same commit.

alexey-milovidov (approved):

PR #93407, opened 2026-01-05
Enable setting `use_skip_indexes_on_data_read`

shankar-iyer:
Follow-up to https://github.com/ClickHouse/ClickHouse/pull/88638#issue-3521307608

The key aspects to verify are 1) Parallel Replicas. 2) Pending mutations and skip index analysis  3) `EXPLAIN`

I have edited all tests that had explicit `use_skip_indexes_on_data_read = 0`  and I have fixed all underlying issues. The objective is to make sure that no feature or test should need `use_skip_indexes_on_data_read = 0` .

### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Setting `use_skip_indexes_on_data_read` is now enabled by default. This 

### Version info
- Merged into: `26.1.1.659`

PR #105686, opened 2026-05-23
Do not attribute skip-index-filtered marks to PREWHERE in QueryConditionCache

### LINKED ISSUE #104781: Wrong results from use_query_condition_cache with PREWHERE pk-prefix + WHERE non-pk IN [..] on 26.2 / 26.3 LTS / 26.4 / 26.5 head (after #96172)
### Company or project name

Weights & Biases / CoreWeave

### Describe what's wrong

`use_query_condition_cache` (default `1`) can be poisoned by a single read so that a later `SELECT count()` against a `ReplacingMergeTree(created_at)` returns a wrong, under-counted result. Disabling the cache restores the correct count.

The trigger query has a very specific shape:

- `PREWHERE` on a primary-key prefix column (equality)
- `WHERE` on a non-PK column with `IN <array literal>`

If both conditions are in `PREWHERE`, or both in `WHERE`, or the columns are swapped, the bug does not fire.

Symptom looks like #96060 / #87091, both closed by #96172. #96172 is in our build (`v26.2.15.4-stable` and newer), so this is either a follow-up cache-hash path that needs the same treatment, or a separate cache bug.

Repro fiddles:
- master (26.5.1.587): https://fiddle.clickhouse.com/a032daa2-ca12-4fce-960f-2fe10026ea5b
- 26.4.2: https://fiddle.clickhouse.com/6930c166-e6f1-49c2-aca7-2c04d627e3cb
- 26.2.18.8: https://fiddle.clickhouse.com/9bbb8b98-357d-4943-bea7-0a0903c895ac

All three return `truth=500002`, then after a single trigger query the same `SELECT count()` returns `237856`. Setting `use_query_condition_cache=0` returns `500002` again.

Yes. Verified on `26.5` head via fiddle (link above) and locally on official images for 26.2.15.4, 26.3.10.60 (LTS), 26.4.2.10, 26.5.1.587. `25.12.5.20` returns the correct count, so the regression window starts somewhere in 26.x.

| version | result | bug? |
| --- | --- | --- |
| 25.12.5.20 | 500002 | no |
| 26.2.15.4 | 237856 | yes |
| 26.2.18.8 | 237856 | yes |
| 26.3.10.60 (LTS) | 237856 | yes |
| 26.4.2.10 | 237856 | yes |
| 26.5.1.587 (head) | 237856 | yes |

### How to reproduce

Fiddle links or paste this SQL into a fresh CH server:

```sql
DROP TABLE IF EXISTS t;

CREATE TABLE t
(
    `id` String,
    `project_id` String,
    `created_at` DateTime64(3) DEFAULT now64(3),
    `started_at` DateTime64(6),
    INDEX idx_id_bloom id TYPE bloom_filter(0.01) GRANULARITY 1
)
ENGINE = ReplacingMergeTree(created_at)
PARTITION BY toYYYYMM(started_at)
ORDER BY (project_id, started_at, id)
SETTINGS min_bytes_for_wide_part = 0, index_granularity = 8192;

-- 500000 rows in partition 202604, all project_id='P1'
INSERT INTO t (id, project_id, started_at)
SELECT concat('id-', toString(number)), 'P1',
       toDateTime64('2026-04-01 00:00:00', 6) + INTERVAL number SECOND
FROM numbers(500000);

-- 2 rows in partition 202605, same project, fresh part
INSERT INTO t (id, project_id, started_at)
SELECT concat('id-fresh-', toString(number)), 'P1',
       toDateTime64('2026-05-15 00:00:00', 6) + INTERVAL number SECOND
FROM numbers(2);

SYSTEM DROP QUERY CONDITION CACHE;

-- expected: 500002
SELECT 'truth' AS step, count() FROM t WHERE project_id = 'P1';

-- trigger query: PREWHERE on PK prefix + WHERE id IN <any array>
SELECT id FROM t
PREWHERE project_id = 'P1'
WHERE id IN ['anything-1', 'anything-2']
FORMAT Null;

-- BUG: returns 237856 on 26.2 / 26.3 LTS / 26.4 / 26.5 head
SELECT 'after_trigger_default' AS step, count() FROM t WHERE project_id = 'P1';

-- workaround: returns 500002
SELECT 'after_trigger_no_cache' AS step, count() FROM t WHERE project_id = 'P1'
SETTINGS use_query_condition_cache = 0;
```

The values in the array literal don't need to exist in the table. Any list works.

We tested adjacent shapes against the same fresh table on `26.2.15.4`. Only the canonical `PREWHERE pk + WHERE non_pk IN [..]` split poisons:

| trigger shape | post-trigger count | poisoned? |
| --- | --- | --- |
| `PREWHERE project_id = 'P1' WHERE id IN ['a','b']` (canonical) | 0 | yes |
| `PREWHERE project_id = 'P1' AND id IN ['a','b']` (both in PREWHERE) | 500002 | no |
| `PREWHERE id IN ['a','b'] WHERE project_id = 'P1'` (swapped) | 500002 | no |
| `PREWHERE project_id = 'P1'` (no `id IN`) | 500002 | no |
| `PREWHERE project_id = 'P1' WHERE 1` | 500002 | no |
| `WHERE project_id = 'P1' AND id IN ['a','b']` (no PREWHERE) | 500002 | no |

So `PREWHERE` is required, and the non-PK `IN [..]` predicate must stay in `WHERE`.

### Expected behavior

Same data, same `SELECT count() FROM t WHERE project_id = 'P1'` returns the same number. Toggling `use_query_condition_cache` should not change semantic results.

None. Silent under-counting.

### Additional context

- Related closed issues: #96060, #87091. Both closed by #96172.
- #96172 hashes `ColumnConst` values into `ActionsDAG::Node::updateHash`. Our trigger has constants on both sides (`'P1'` in `PREWHERE`, plus the array literal in `WHERE id IN [..]`). The fact that flattening both predicates into a single `PREWHERE ... AND ...` makes the bug disappear suggests the cache-key build is treating `PREWHERE`-derived and `WHERE`-derived predicates differently, and something on the `WHERE` side (probably the array literal in `IN`) is not feeding into the key the way #96172 assumes.

### Version info
- Resolved by: #105686
- Merged into: `26.6.1.141`
- Backported to: `26.5.2.12`, `26.4.4.15`, `26.3.13.13`

groeneai:
Closes #104781.

When `use_skip_indexes_on_data_read = 1` (default since #93407), `MergeTreeReaderIndex` sits ahead of the PREWHERE readers in the reader chain and can drop whole marks via `canSkipMark`. The dropped marks still appear in `read_mark_ranges`, so when a read batch produces zero rows because the skip-index dropped everything, `MergeTreeSelectProcessor::read` attributes the marks to the PREWHERE predicate's action-graph hash and writes a "no marks match" verdict under that key. The next benign query whose `WHERE` hashes to the same PREWHERE-predicate hash hits the poisoned entry and silently under-counts.

The reporter's reduced case on master (26.5.1.587): `SELECT count() ... WHERE project_id = 'P1'` returns `237856` instead of `500002` after running a `PREWHERE project_id = 'P1' WHERE id IN [...]` query against a bloom-filter index on `id`. Locally I reproduced `65536` vs `500002`; the exact under-count depends on the bloom-filter false-positive rate.

`Settings::isQueryConditionCacheWritable` from #104225 does not cover this: the reporter's trigger runs with default settings, no experimental or `allow_suspicious_*` flag.

### Fix

`IMergeTreeReader::canSkipAnyMark` is added (default `false`) and overridden on `MergeTreeReaderIndex` to return `true` only when a skip-index or projection-index read result is present. `MergeTreeReadTask` exposes `readersChainCanSkipMarksBeforePrewhere`, and `MergeTreeSelectProcessor::read` suppresses both `addPrewhereUnmatchedMarks` and the PREWHERE-side `QueryConditionCache::write` when that helper returns `true`. `FilterTransform` writes under the WHERE-side condition hash are unaffected; ordinary PREWHERE-only queries (no skip-index, no projection-index) keep their existing cache pruning.

### Tests

`tests/queries/0_stateless/04275_104781_qcc_prewhere_skip_index_attribution.sql` runs the canonical trigger from #104781 and asserts the cache-on and cache-off counts agree. It also covers the `use_skip_indexes_on_data_read = 0` workaround and a sanity check that a regular PREWHERE-only query (no bloom-filter index) still passes. Without the fix the test reports `after_trigger 65536`; with the fix it reports `after_trigger 500002`.

I also ran the existing QCC test suite locally against the patched server (`02346_*`, `03100_*`, `03229_*`, `03460_*`, `04004_*`): all pass.

Fixed a silent under-count in `SELECT` queries when `use_query_condition_cache = 1` (default). A query of the shape `PREWHERE pk_prefix = X WHERE non_pk IN (...)` against a column with a bloom-filter skip index poisoned the `QueryConditionCache` for the `pk_prefix = X` predicate, so a subsequent benign `SELECT count() ... WHERE pk_prefix = X` returned an incorrect, under-counted result. Affected all 26.x releases. [#104781](https://github.com/ClickHouse/ClickHouse/issues/104781).

---

> [!NOTE]
> **Medium Risk**
> Touches MergeTree read path and QueryConditionCache PREWHERE attribution logic; mistakes could reduce caching effectiveness or reintroduce incorrect mark-skipping in edge cases involving skip/projection indexes.
> 
> **Overview**
> Prevents **QueryConditionCache PREWHERE entries** from being poisoned when skip/projection index readers drop whole marks before PREWHERE runs (Issue #104781), which previously could cause later queries to under-count by incorrectly reusing “no marks match” results.
> 
> This adds `IMergeTreeReader::canSkipAnyMark()` (implemented by `MergeTreeReaderIndex`) and a `MergeTreeReadTask::readersChainCanSkipMarksBeforePrewhere()` helper, then uses it in `MergeTreeSelectProcessor` to suppress both collecting PREWHERE-unmatched mark ranges and writing PREWHERE-side cache entries when earlier index readers might have filtered marks. A new stateless regression test reproduces the under-count scenario with a bloom-filter skip index and asserts correct counts with/without the cache and with the workaround setting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd0a30cb8950df6b96a874e8fcb9b6873635d246. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Version info
- Merged into: `26.6.1.141`
- Backported to: `26.5.2.12`, `26.4.4.15`, `26.3.13.13`

### DIFF
--- a/src/Storages/MergeTree/IMergeTreeReader.h
+++ b/src/Storages/MergeTree/IMergeTreeReader.h
@@ -86,6 +86,12 @@ public:
 
     virtual bool canSkipMark(size_t, size_t) { return false; }
 
+    /// Returns true if this reader can skip whole marks via `canSkipMark` for at least some inputs.
+    /// Independent of any particular mark index. Used by callers that need to know upfront whether
+    /// the reader chain may filter marks before the PREWHERE step runs — for example, to decide
+    /// whether `read_mark_ranges` with `row_count == 0` can be attributed to the PREWHERE predicate.
+    virtual bool canSkipAnyMark() const { return false; }
+
     virtual void updateAllMarkRanges(const MarkRanges & ranges) { all_mark_ranges = ranges; }
 
     StorageSnapshotPtr getStorageSnapshot() const { return storage_snapshot; }
--- a/src/Storages/MergeTree/MergeTreeReadTask.cpp
+++ b/src/Storages/MergeTree/MergeTreeReadTask.cpp
@@ -451,4 +451,11 @@ void MergeTreeReadTask::addPrewhereUnmatchedMarks(const MarkRanges & mark_ranges
     prewhere_unmatched_marks.insert(prewhere_unmatched_marks.end(), mark_ranges_.begin(), mark_ranges_.end());
 }
 
+bool MergeTreeReadTask::readersChainCanSkipMarksBeforePrewhere() const
+{
+    /// Only `prepared_index` (a `MergeTreeReaderIndex`) sits ahead of the PREWHERE readers in the
+    /// reader chain and is able to skip whole marks via `canSkipMark`.
+    return readers.prepared_index && readers.prepared_index->canSkipAnyMark();
+}
+
 }
--- a/src/Storages/MergeTree/MergeTreeReadTask.h
+++ b/src/Storages/MergeTree/MergeTreeReadTask.h
@@ -202,6 +202,13 @@ public:
     void addPrewhereUnmatchedMarks(const MarkRanges & mark_ranges_);
     const MarkRanges & getPrewhereUnmatchedMarks() { return prewhere_unmatched_marks; }
 
+    /// Returns true if a reader earlier in the chain than PREWHERE can skip whole marks based on
+    /// secondary indexes (skip-index or projection-index). When true, marks that appear in
+    /// `read_mark_ranges` with `row_count == 0` may have been filtered before PREWHERE evaluated
+    /// them, so they must not be attributed to the PREWHERE predicate in the QueryConditionCache.
+    /// See Issue #104781.
+    bool readersChainCanSkipMarksBeforePrewhere() const;
+
     Readers releaseReaders() { return std::move(readers); }
 
     size_t getNumMarksToRead() const { return mark_ranges.getNumberOfMarks(); }
--- a/src/Storages/MergeTree/MergeTreeReaderIndex.cpp
+++ b/src/Storages/MergeTree/MergeTreeReaderIndex.cpp
@@ -10,6 +10,16 @@ namespace ErrorCodes
     extern const int LOGICAL_ERROR;
 }
 
+bool MergeTreeReaderIndex::canSkipAnyMark() const
+{
+    /// `canSkipMark` only ever returns true when a skip-index or projection-index read result is present.
+    /// `lazy_materializing_rows` alone affects row-level filtering inside a granule and never returns
+    /// true from `canSkipMark`, so a reader configured only for lazy materialization does not skip
+    /// whole marks.
+    return index_read_result
+        && (index_read_result->skip_index_read_result || index_read_result->projection_index_read_result);
+}
+
 MergeTreeReaderIndex::MergeTreeReaderIndex(const IMergeTreeReader * main_reader_, MergeTreeIndexReadResultPtr index_read_result_, const PaddedPODArray<UInt64> * lazy_materializing_rows_)
     : IMergeTreeReader(
           main_reader_->data_part_info_for_read,
--- a/src/Storages/MergeTree/MergeTreeReaderIndex.h
+++ b/src/Storages/MergeTree/MergeTreeReaderIndex.h
@@ -34,6 +34,8 @@ public:
 
     bool canSkipMark(size_t mark, size_t current_task_last_mark) override;
 
+    bool canSkipAnyMark() const override;
+
     size_t getResultColumnCount() const override { return 1; }
 
     bool producesFilterOnly() const override { return true; }
--- a/src/Storages/MergeTree/MergeTreeSelectProcessor.cpp
+++ b/src/Storages/MergeTree/MergeTreeSelectProcessor.cpp
@@ -284,7 +284,13 @@ MergeTreeSelectProcessor::readCurrentTask(MergeTreeReadTask & current_task, IMer
             .is_finished = false, .read_mark_ranges = std::move(res.read_mark_ranges)};
     }
 
-    if (reader_settings.use_query_condition_cache && prewhere_info)
+    /// Suppress PREWHERE attribution when a reader earlier in the chain (skip-index or projection-index)
+    /// can skip whole marks before PREWHERE evaluates them. In that case `res.read_mark_ranges` may
+    /// include marks that were filtered by the index, and recording them under the PREWHERE predicate's
+    /// hash would poison the QueryConditionCache: later queries that share the same PREWHERE predicate
+    /// hash would skip those marks even though the predicate alone matches their rows. See Issue #104781.
+    if (reader_settings.use_query_condition_cache && prewhere_info
+        && !current_task.readersChainCanSkipMarksBeforePrewhere())
         current_task.addPrewhereUnmatchedMarks(res.read_mark_ranges);
 
     return {Chunk(), res.num_read_rows, res.num_read_bytes, false, std::move(res.read_mark_ranges)};
@@ -361,8 +367,12 @@ ChunkAndProgress MergeTreeSelectProcessor::read()
         {
             if (!task || algorithm->needNewTask(*task))
             {
-                /// Update the query condition cache for filters in PREWHERE stage
-                if (reader_settings.use_query_condition_cache && task && prewhere_info)
+                /// Update the query condition cache for filters in PREWHERE stage.
+                /// Skip the write when a reader earlier in the chain (skip-index or projection-index)
+                /// could have filtered marks before PREWHERE saw them, to avoid attributing those
+                /// marks to the PREWHERE predicate hash. See Issue #104781.
+                if (reader_settings.use_query_condition_cache && task && prewhere_info
+                    && !task->readersChainCanSkipMarksBeforePrewhere())
                 {
                     for (const auto * output : prewhere_info->prewhere_actions.getOutputs())
                     {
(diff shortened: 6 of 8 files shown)


groeneai:
**Pre-PR validation gate** (session `cron:clickhouse-ci-task-worker:20260522-233000`)

a) **Deterministic repro?** Yes. The SQL in #104781 deterministically reproduces `count() = 65536` (locally) vs `500002` with `use_query_condition_cache = 1` vs `0`. The regression test in this PR runs the same shape.

b) **Root cause explained?** With `use_skip_indexes_on_data_read = 1` (default since #93407), `MergeTreeReaderIndex` is the first reader in the chain and can drop marks via `canSkipMark` before the PREWHERE readers see them. The dropped marks remain in `read_mark_ranges`. When all marks in a batch are dropped this way, `MergeTreeSelectProcessor::read` records them via `addPrewhereUnmatchedMarks` (`MergeTreeSelectProcessor.cpp:287-288`), and the task-transition path writes them to `QueryConditionCache` under the PREWHERE predicate's action-graph hash (`MergeTreeSelectProcessor.cpp:380-388`). The PREWHERE predicate alone matches those marks; the marks were dropped by the WHERE-side bloom filter. A later `SELECT count() ... WHERE project_id = 'P1'` whose filter hashes to the PREWHERE-only hash reads back the poisoned "no marks match" verdict and under-counts.

c) **Fix matches root cause?** The new `IMergeTreeReader::canSkipAnyMark` returns true on `MergeTreeReaderIndex` only when the reader carries a skip-index or projection-index read result (the only two paths that make `canSkipMark` return true; `lazy_materializing_rows` does not). `MergeTreeReadTask::readersChainCanSkipMarksBeforePrewhere` reports it. `MergeTreeSelectProcessor::read` then suppresses both the per-batch accumulation and the task-transition write exactly when the chain has such a reader, leaving every other case (regular PREWHERE, lazy materialization alone, no PREWHERE) untouched.

d) **Test intent preserved? New tests added?** New regression test `04275_104781_qcc_prewhere_skip_index_attribution`. It (i) reproduces the canonical trigger and asserts `count()` agrees with the cache-off baseline, (ii) verifies the `use_skip_indexes_on_data_read = 0` workaround, and (iii) sanity-checks that an ordinary `PREWHERE` query (no skip-index in the chain) still hits the cache correctly across two runs.

e) **Demonstrated in both directions?** Yes. With the source change reverted on the same build tree, `04275_104781_*` reports `after_trigger 65536` and the diff against the reference is `2c2`. With the source change restored, the test prints `after_trigger 500002` and the diff is empty.

f) **Fix is general?** The gate covers both PREWHERE-side write paths (per-batch accumulation and task-transition write) and both index types that participate in `MergeTreeReaderIndex::canSkipMark` (skip-index and projection-index). The `canSkipAnyMark` hook on `IMergeTreeReader` is the single capability query; any future reader that filters marks before PREWHERE only has to override it and is then handled uniformly.

PR #105335 is an orthogonal improvement (record per-granule unmatched marks within partial batches). If it lands after this PR, it should call `readersChainCanSkipMarksBeforePrewhere` at its new write sites too.

groeneai:
cc @rschu1ze @shankar-iyer @nihalzp could you review? Fixes the data-correctness bug reported in #104781 — bloom-filter skip-index drops in the read chain were attributed to PREWHERE in the QueryConditionCache, poisoning later queries with the same PREWHERE predicate. @nihalzp the bisect comment was the key clue; thanks.

groeneai:
Pushed [dd0a30cb8950](https://github.com/ClickHouse/ClickHouse/pull/105686/commits/dd0a30cb8950) to address the `Bugfix validation (functional tests)` failure on the previous push.

**Root cause of the bugfix-validation failure:** the first regression test relied on session-level `SET` statements and default settings, so on master HEAD without the fix the bug did not reproduce. Three independent issues:

1. `tests/clickhouse-test` randomizes `use_query_condition_cache` and `use_skip_indexes_on_data_read`. Either flipped to `0` disables a bug ingredient.
2. `optimize_use_implicit_projections = 1` (default on master) lets `count()` queries be answered from `_exact_count_projection` per-part metadata, completely bypassing `MergeTreeSelectProcessor` and therefore the poisoned `QueryConditionCache` entry. The CIDB `query_log` from the failing run confirmed every `count()` used `_exact_count_projection`.
3. `optimize_move_to_prewhere` / `query_plan_optimize_prewhere` randomization could leave `WHERE pk_prefix = X` outside the PREWHERE-keyed cache slot, so the `truth` / `after_trigger` queries would not consult the poisoned entry.

**Fix:** each test query now pins the settings it needs via per-query `SETTINGS`, so CI randomization cannot mask the bug. Verified locally on a debug build:

- **20/20 FAIL** against a build where the fix is neutralized (`readersChainCanSkipMarksBeforePrewhere` forced to return `false`), with and without `--no-random-settings`.
- **20/20 PASS** against the build with the fix, with random settings enabled.

No production-code change in this commit; only the regression test was tightened.

PR #105713, opened 2026-05-24
Resize Web UI textarea for slight query overflow

### LINKED ISSUE #105709: If the query is only slightly larger than the textarea size in the Web UI (play.html), resize the textarea to fit it.
### Company or project name

ClickHouse

### Use case

Slightly means 30% - we will accept up to 30% larger textarea. However, if it is above 30%, we will not resize the textarea at all.
Which means, we provide "hysteresis".

Implement it.

_No response_

### Additional context

_No response_

### Version info
- Resolved by: #105713
- Merged into: `26.6.1.378`

itsjoeoui:
Allow the Web SQL UI query editor in `play.html` to grow when the query is only slightly taller than the current `textarea`. This keeps small overflows visible while leaving much larger pasted queries scrollable.

Closes https://github.com/clickhouse/clickhouse/issues/105709

- Improvement

The Web SQL UI now expands the query editor when a query is at most 30% taller than the current editor height.

### Documentation entry:
- [x] Documentation is not required for this change because the behavior is self-explanatory in the Web SQL UI.

### Version info
- Merged into: `26.6.1.378`


programs/server/play.html
```diff
@@ -4147,6 +4164,8 @@
 
 async function updateQueryParams()
 {
+    resizeQueryAreaIfSlightlyOverflowing();
```
alexey-milovidov: Sounds like it should be called not from inside this function, but from a more appropriate place.

itsjoeoui: Moved the call into the `query_area` `input` event handler in `9d0374b6735`, so `updateQueryParams` no longer owns the resize side effect.

alexey-milovidov:
Will it activate on page load?

itsjoeoui:
> Will it activate on page load?

Ah I missed that, it should be fixed now. 😄 

I tested with something like http://127.0.0.1:8123/programs/server/play.html#U0VMRUNUIDAKU0VMRUNUIDEKU0VMRUNUIDIKU0VMRUNUIDMKU0VMRUNUIDQKU0VMRUNUIDUKU0VMRUNUIDYKU0VMRUNUIDcKU0VMRUNUIDgKU0VMRUNUIDk= and it is working.

alexey-milovidov (approved):

Issue #105733, opened 2026-05-24
singleValueOrNullState / singleValueOrNullMerge lose semantic state during merge and serialization


fallintoplace:
**Describe the unexpected behaviour**

`singleValueOrNull` tracks two semantic flags in addition to the stored value: `first_value` and `is_null`.

In the current implementation:
- merge does not propagate `rhs.is_null`
- write/read serialize only the underlying value payload

That means aggregate states can change meaning in two ways:
1. a partial state that already means `result must be NULL` can merge into an empty destination and turn back into a concrete value
2. a valid single-value state can lose `first_value = false` after serialization/deserialization and then merge back as `NULL`

Current code on `master`:
- merge path and state fields:
  - https://github.com/ClickHouse/ClickHouse/blob/fac6aa2045d1c81e28650fdd39dcbdb477f25d28/src/AggregateFunctions/AggregateFunctionSingleValueOrNull.cpp#L26-L77
- explicit TODO about serialization losing `first_value` / `is_null`:
  - https://github.com/ClickHouse/ClickHouse/blob/fac6aa2045d1c81e28650fdd39dcbdb477f25d28/src/AggregateFunctions/AggregateFunctionSingleValueOrNull.cpp#L79-L83

**Which ClickHouse versions are affected?**

Confirmed in current `master` by source review at commit `fac6aa2045d1c81e28650fdd39dcbdb477f25d28`.

This appears to be longstanding. The same pattern already existed in the original implementation added on June 4, 2021:
- https://github.com/ClickHouse/ClickHouse/blob/289c5d3ad680c6e07ef189eed9ea30417acdac65/src/AggregateFunctions/AggregateFunctionMinMaxAny.h#L633-L668
- https://github.com/ClickHouse/ClickHouse/blob/289c5d3ad680c6e07ef189eed9ea30417acdac65/src/AggregateFunctions/AggregateFunctionMinMaxAny.h#L789-L801

So this likely affects all released versions since `singleValueOrNull` was introduced.

**How to reproduce**

I have not executed the SQL below locally yet; these are the minimal validation queries derived from the current implementation.

1. Merge-state loss:

```sql
SELECT singleValueOrNullMerge(s)
FROM
(
    SELECT singleValueOrNullState(number) AS s
    FROM numbers(2)
);
```

The inner state has already observed two distinct values, so the final result should be `NULL`.

[...]

PR #105734, opened 2026-05-24
Fix singleValueOrNull null-state merge

fallintoplace:
Propagate `is_null` when merging `singleValueOrNull` aggregate states. A partial state that has already decided to finalize to `NULL` (after seeing multiple distinct values) was losing the `is_null` flag when merged into an empty destination, returning a concrete value instead.

Partial fix for #105733 (covers the merge path only; the serialization state-format fix is not included here).

Fixed `singleValueOrNullMerge` returning a concrete value instead of `NULL` when merging a state that had already observed multiple distinct values.

### Version info
- Backported to: `26.5.8.2`, `26.3.20.6`


src/AggregateFunctions/AggregateFunctionSingleValueOrNull.cpp
```diff
@@ -62,6 +62,12 @@ struct AggregateFunctionSingleValueOrNullData
 
     void add(const Self & to, Arena * arena)
     {
+        if (to.is_null)
+        {
+            is_null = true;
```
alexey-milovidov, line 67: This looks strange. If we have a value, but the other is null, why do we have to set our value to null?

fallintoplace: The name `is_null` is a bit misleading here. Re-reading the state transitions, it is set only after the state has already stored one non-NULL value and then sees a different one:

if (first_value)
{
    first_value = false;
    data().set(...);
}
else if (!data().isEqualTo(...))
{
    is_null = true;
}

So in this code it does not mean “the other state contains a NULL value”. It means “this partial aggregate has already seen at least two distinct non-NULL values, so the final result must be NULL”.

Because of that, if the left state has `5` and the right state has already seen `{5, 7}`, the merged result should still be `NULL`. The right state may still physically keep the first seen value in `data()` as an implementation detail, but once `is_null` is set that stored value is no longer semantically valid on its own.

That is the case this patch is trying to preserve when merging into an empty destination.

Algunenano (approved):
Thanks!

clickgapai:
Hi — could this be a backport candidate for `26.5`, `26.4`, `26.3` (LTS), `25.8` (LTS)? No backport label was found, and the affected code appears to exist in those branches based on a file/function presence check.

**Affected code:** `AggregateFunctionSingleValueOrNullData::add(const Self&, Arena*)` merge logic — core code present in all supported branches.

**Why this might apply:** The `singleValueOrNull` function was introduced in June 2021 (commit 289c5d3ad68), predating all supported branches. This is a correctness bug causing wrong results when merging aggregate states, affecting distributed queries and any use of the `State`/`Merge` combinators.

_Caveat: I verified that the modified code exists in those branches, but not whether the bug is actually reachable there — a new caller in master may be what makes this newly observable. If older branches can't reach the affected path, please ignore this; otherwise consider `pr-must-backport` or a version label like `v26.5-must-backport`._

Issue #57360, opened 2023-11-29
Parsing web log files with templates stalls


mneedham:
I'm trying out [templates](https://clickhouse.com/docs/en/integrations/data-formats/templates-regexp) on a [Kaggle web server logs dataset](https://www.kaggle.com/datasets/eliasdabbas/web-server-access-logs/?select=access.log).

I'm using ClickHouse Local and this is the query that I'm running:

```sql
FROM file('data/archive.zip :: access.log', Template)
SELECT count(*)
SETTINGS format_template_row='templates/web_server_specific.template', 
input_format_record_errors_file_path='errors.log', 
input_format_allow_errors_ratio=0.1;
```

It gets up to around 130k rows and then it seems to kind of stall. 

https://github.com/ClickHouse/ClickHouse/assets/13220/5ab37f24-c897-4804-b26d-8c96b32461af

niyue:
The issue is caused by `Template` input format error recovery after a malformed row.

In the reproduced dataset, row `136411` contains this invalid request line:

```text
74.82.60.74 - - [22/Jan/2019:08:27:26 +0330] "\x13BitTorrent protocol\x00\x00\x00\x00\x00\x18\x00\x05\x7FI..." 400 166 "-" "-" "-"
```
This row does not match the template, so failing to parse it is expected. However, after skipping that bad row, `TemplateRowInputFormat` does not resume parsing at the next row correctly. It consumes the row delimiter during recovery, then tries to consume the same delimiter again when reading the next row. As a result, valid following rows are also treated as parse errors.

This makes the error count grow continuously until `input_format_allow_errors_ratio` is exceeded, so the query fails around row `151567` instead of skipping only the malformed rows.

PR #105735, opened 2026-05-24
Fix `Template` recovery after skipped bad rows

niyue:
This PR fixes issue: https://github.com/ClickHouse/ClickHouse/issues/57360

Fix `Template` input format error recovery after malformed rows. Previously, when `input_format_allow_errors_num` or `input_format_allow_errors_ratio` allowed skipping a malformed row, the next valid row could be incorrectly skipped with a parsing exception because `TemplateRowInputFormat` consumed the row delimiter twice.

### Changelog entry:
- Fix `Template` input format error recovery after malformed rows. 

### Version info
- Merged into: `26.6.1.367`


alexey-milovidov (approved):

clickgapai:

### TL;DR
**Fix `Template` recovery after skipped bad rows**

**ClickGap verified:** reproduced it with the script below on `26.5`; tested 5 version(s) — 4 affected, 1 clean (see Affects); confirmed it predates the oldest available release (pre-existing).

### Reproduction
_Shell reproducer — run against a local ClickHouse build (`bash <script> <clickhouse-binary>`); requires the referenced test fixtures from the repo._
```bash
# Template input format: the first VALID row after a skipped malformed row is
# wrongly dropped. syncAfterError() and readRow() each consume the row
# delimiter once, so the delimiter is eaten twice and the following good row is
# lost. Row 2 below is malformed (value "-" is not a UInt64) and is skipped;
# rows 1 (Alice) and 3 (Carol) are valid and both must appear. On a buggy
# build, Carol is missing from the output.

printf '1 Alice 100\n2 Bob -\n3 Carol 300' > /tmp/tmpl_data
printf '%s' '${id:CSV} ${name:CSV} ${value:CSV}' > /tmp/tmpl_fmt

clickhouse local --query "
    SELECT id, name, value
    FROM file('/tmp/tmpl_data', Template, 'id UInt64, name String, value UInt64')
    ORDER BY id
    SETTINGS
        format_template_row = '/tmp/tmpl_fmt',
        input_format_allow_errors_ratio = 0.9"

# Expected: both valid rows present
#   1 Alice 100
#   3 Carol 300
# Observed (buggy build): the post-skip valid row (3 Carol 300) is dropped.
```

### Bisect
- **Present since at least `v24.1.8.22`** — reproduces on every release line tested down to there; the introducer predates the oldest release with a downloadable binary, so a single commit isn't pinnable from release binaries (not a recent regression).

### Affects
- Verified against `master` build `eb96a8a57905` (aarch64)

### Suggested next step

**Component:** `comp-formats`
**Severity:** `P2`

cc fix PR author: @niyue
cc fix PR reviewer: @alexey-milovidov

---
_ClickGap · Finding: `phase_d_pr105735`_

PR #105736, opened 2026-05-24
Adding RowBinaryWithNamesAndTypesAndDefaults 

mzitnik:
- New Feature
    
Adding `RowBinaryWithNamesAndTypesAndDefaults` format to get better support for schema evolution

### Explanation.

A new input-only `RowBinary` variant that combines the `RowBinaryWithNamesAndTypes` header (column names + types) with the `RowBinaryWithDefaults` per-cell default-marker byte. It enables schema-evolving INSERTs into ClickHouse: producers can omit columns from the header
  (target's `DEFAULT` is used) And, for columns they do send, mark individual cells as "use the DEFAULT" without conflating that with NULL. NULL and DEFAULT are independent and distinguishable per cell.

  Motivation

  Two existing formats each cover half of the schema-evolution story:

  - `RowBinaryWithNamesAndTypes` carries a header, so the reader can reorder columns, fill defaults for columns missing from the file, and skip columns extra in the file. But within a row, there's no way to say "for this cell, use the target's DEFAULT" — and for a
  Nullable(T) column, sending the Nullable's null byte means NULL, not DEFAULT.
  - `RowBinaryWithDefaults` has a per-cell default-marker so a row can express "use DEFAULT" without conflating it with NULL — but it carries no header, so the reader cannot reorder, cannot fill missing-column defaults, and cannot skip unknown columns.

  

### Version info
- Merged into: `26.6.1.327`


Avogar (approved):

src/Processors/Formats/Impl/BinaryRowInputFormat.cpp
```diff
@@ -170,7 +170,16 @@ void BinaryFormatReader<with_defaults>::skipField(size_t file_column)
 {
     if (file_column >= read_data_types.size())
         throw Exception(ErrorCodes::CANNOT_SKIP_UNKNOWN_FIELD,
-                        "Cannot skip unknown field in RowBinaryWithNames format, because it's type is unknown");
+                        "Cannot skip unknown field in RowBinary format, because its type is unknown");
+
+    if constexpr (with_defaults)
+    {
+        UInt8 is_default;
```
PedroTadim, line 177: It seems the tidy build on master is broken because of this line. Can you fix it?

Avogar: https://github.com/ClickHouse/ClickHouse/pull/106300

mzitnik: @Avogar thanks

PR #105739, opened 2026-05-24
compareTrackAt for ColumnConst

4ertus2:
- Performance Improvement

compareTrackAt for ColumnConst

Continuation of #99013


4ertus2:
Most probably there's no way to call this function from current codebase. But obviously `compareTrackAt` itself is faster when general IColumns's one in N times.

antaljanosbenjamin:
> Most probably there's no way to call this function from current codebase.

Then what is the point? Can you prove performance improvement with a perfomance test? If not, then sorry to say it, but is not an improvement.

4ertus2:
> > Most probably there's no way to call this function from current codebase.
> 
> 
> 
> Then what is the point? Can you prove performance improvement with a perfomance test? If not, then sorry to say it, but is not an improvement.

Ok. It's a minor change. I do not care about the PR status. It's just a pease of entire task about devirtualization.

antaljanosbenjamin:
I mean why to write code that is never used? If we cannot write a performance test to exercise this code, then why should we include it? It make sense for other columns but we can leave this as is.

4ertus2:
> why should we include it?

Because it's a part of bigger feature #99013 that could be used in other places.
ClickHouse's columns model has performance penalties caused of existence of IColumn. It's OK to have such a penalty in some cases. But there're functions like `compareAt` that called small one-value operation. And in such functions ClickHouse's columns model works much worse then others (who have no interface and have manual columns dispatch).

Your comments sounds for me like "Why should we buy the left rear door of the car if we only use right ones". Just because it is a part of entire solution.

4ertus2:
@antaljanosbenjamin What about of this part of task?

antaljanosbenjamin:
I still think that if we cannot call it, it doesn't make sense include it. I understand that some devirtualization is super useful just as you showed in the other PR for fixed strings. But if there is no way to call this function, there is no real life benefit, then there is no reason to include it. If it is not covered, it will become obsolete and/or buggy over time.

> Your comments sounds for me like "Why should we buy the left rear door of the car if we only use right ones". Just because it is a part of entire solution.

I would phrase it like "Why should we buy rear doors for the car which doesn't have a second row of seats?".

I will ask other colleagues what do they think.

antaljanosbenjamin:
So our conclusion is the code cannot be exercised and/or tested, we don't want to merge it. It would be dead on arrival.

rienath (changes requested):
I like the PR. Let's, as @antaljanosbenjamin suggested, merge it together with (or after) a caller that preserves const key columns. Happy to reopen then (feel free to ping me). Closing for now

PR #105747, opened 2026-05-24
Remove redundant includes of config.h

azat:
$ gg 'include.*"config.h"' src/**.h | cut -d: -f1 | sort -u | xargs -i sh -c 'grep -q -e WITH_COVERAGE -e USE_ -e CLICKHOUSE_CLOUD {} || echo {}' | xargs sed -i '/include "config.h"/d'

Reduces only from 4K to 3K, likely due to inclusion of it in IColumn.h

### Version info
- Merged into: `26.6.1.136`


Algunenano (approved):
Nice idea

PR #105749, opened 2026-05-24
Fix server crash in -State aggregate combinators with WITH ROLLUP/CUBE/TOTALS + ORDER BY

### LINKED ISSUE #105740: `DB::finalizeChunk` mutates `ColumnAggregateFunction` columns while still aliased by downstream stages / threads — root cause for #102460 (caught by TSan / ASan / MSan)
# Triage context

Found by the server-side AST fuzzer + oracle harness (PR #99980) running cycle 2 on master `840494d51c9`. **This is the root cause for the recurring jemalloc heap-corruption family tracked in https://github.com/ClickHouse/ClickHouse/issues/102460** — fixing this should close that issue. All 4 sanitizer types caught the same bug from different angles (TSan: iterator invalidation; ASan cross-thread: ThreadPool free vs TCPHandler read; ASan same-thread: downstream stage reads freed state; MSan: double-destruction). The MSan variant pinpoints the fix site.

---

## Setup (reproducer prerequisites)

Same `_s1`/`_s2` populate as #104203 / #105649-651-662; queries in this report use both tables:

```sql
CREATE TABLE _s1 (
    a UInt32, b Nullable(Int64), c String, d Float64,
    e Nullable(String), f Date, g UInt8, h Nullable(UInt16)
) ENGINE = MergeTree ORDER BY a;

INSERT INTO _s1 SELECT
    number,
    if(number%3=0, NULL, number*10),
    toString(number%7),
    number*0.1-50,
    if(number%7=0, NULL, toString(number)),
    toDate('2020-01-01') + number,
    number%3,
    if(number%5=0, NULL, toUInt16(number))
FROM numbers(500);

CREATE TABLE _s2 (x UInt32, y String, z Nullable(UInt8))
ENGINE = MergeTree ORDER BY x;

INSERT INTO _s2 SELECT number, toString(number % 5), if(number % 4 = 0, NULL, toUInt8(number % 200))
FROM numbers(300);
```

Relevant settings:

```
enable_analyzer = 1
allow_experimental_correlated_subqueries = 1
allow_experimental_variant_type = 1
use_query_condition_cache = 0
max_execution_time = 10
max_memory_usage = 1073741824
```

---

# `DB::finalizeChunk` mutates `ColumnAggregateFunction` columns while they're still aliased by downstream stages / threads — root cause for #102460

## TL;DR

`DB::finalizeChunk(DB::Chunk&, ...)` destructs / replaces the chunk's `ColumnAggregateFunction` columns assuming exclusive ownership, but the chunk is aliased into downstream stages (and across threads) by transforms that call it from `CubeTransform::generate`, `TotalsHavingTransform::transform`, and the rollup/totals/cube fan-out path. Subsequent reads, hashes, serializations, and destructions of those columns then operate on freed/double-destroyed aggregate-state memory.

In release / debug / ubsan jemalloc builds the symptom is the recurring jemalloc-corruption SEGV in `eset_remove` / `bin_slabs_full_remove` / `edata_list_inactive_remove` — i.e. the consumer side of [#102460](https://github.com/ClickHouse/ClickHouse/issues/102460).

## Unified picture across all 4 sanitizer signatures

| Sanitizer | Snapshot | Mechanism | Transform |
|---|---|---|---|
| TSan | inst-13, inst-14 | Same-thread iterator invalidation in `Aggregator::mergeStreamsImpl` after `ReservoirSampler::merge` realloc | (inner) Aggregator under `CubeTransform` |
| ASan cross-thread | inst-08, inst-10, inst-11 | ThreadPool destroys `ColumnAggregateFunction`, TCPHandler reads HLL state via `NativeWriter::serializeBatch` | CubeTransform |
| ASan same-thread | inst-09 | `CubeTransform::finalizeChunk` destroys state, `IntersectOrExceptTransform::hashRow` reads it | CubeTransform → IntersectOrExceptTransform |
| **MSan** | **inst-15** | **`TotalsHavingTransform::finalizeChunk` destroys state, `IOutputFormat::work` destroys it again** | **TotalsHavingTransform** |
| Release+debug+ubsan jemalloc consumer | inst-01, inst-06, inst-12 ([#102460](https://github.com/ClickHouse/ClickHouse/issues/102460)) | Downstream effect: slab metadata corrupted by double-modification → SEGV in `eset_remove` / `bin_slabs_full_remove` | n/a (consumer) |

All five are **one upstream bug**.

## MSan: the precise fix point

Producer query (release+ASan also crash; debug abort):

```sql
SELECT s1.g, singleValueOrNullStateOrDefault(s1.b)
FROM _s1 AS s1 RIGHT JOIN _s2 ON s1.a = _s2.x
GROUP BY s1.g WITH TOTALS
ORDER BY s1.g ASC NULLS LAST;
```

MSan report:

```
WARNING: MemorySanitizer: use-of-uninitialized-value
  #0 IAggregateFunctionDataHelper<AggregateFunctionSingleValueOrNullData, AggregateFunctionSingleValueOrNull>::destroy(char*)
  #1 DB::ColumnAggregateFunction::~ColumnAggregateFunction()
  #2 DB::ColumnAggregateFunction::~ColumnAggregateFunction()    (2nd-stage)
  #3 DB::IOutputFormat::work()              ← pipeline finishing; tries to destroy column again
  #4 DB::ExecutionThreadContext::executeTask
  ... (PipelineExecutor)

  Member fields were destroyed (previously):
  #0 __sanitizer_dtor_callback_fields
  #1 DB::ColumnAggregateFunction::~ColumnAggregateFunction()
  #2 DB::finalizeChunk(DB::Chunk&, ...)
  #3 DB::TotalsHavingTransform::transform(DB::Chunk&)    ← the first destructor
  #4 DB::ISimpleTransform::transform(...)
  #5 DB::ISimpleTransform::work()
```

Proves the bug is in `finalizeChunk()` itself, not specific to `CubeTransform`. Any transform that calls `finalizeChunk` while the chunk is still referenced downstream will trigger it.

## ASan cross-thread (inst-08)

Producer:

```sql
SELECT anyRespectNullsStateOrNullDistinctOrNullDistinct(b) IGNORE NULLS
FROM _s1
LEFT ARRAY JOIN length(c, notIn(isZeroOrNull(257) = a, a,
    (SELECT DISTINCT id FROM _s3 PREWHERE -9223372036854775808 > val WHERE -9223372036854775808 > val))),
    b <= materialize(-2147483649)
PREWHERE isNull(c)
GROUP BY modulo(10, b) WITH CUBE WITH TOTALS;
```

ASan:

```
==3868145==ERROR: AddressSanitizer: heap-use-after-free on address 0xf685e2798c80
READ of size 2650 at 0xf685e2798c80 thread T6 (TCPHandler):
  #0 __asan_memcpy
  #1 DB::WriteBuffer::write
  #2 DB::HyperLogLogWithSmallSetOptimization<long, 16, 12, ...>::write(WriteBuffer&)
  #3 DB::IAggregateFunctionHelper<AggregateFunctionUniq<long, AggregateFunctionUniqHLL12Data<long, false>>>::serializeBatch
  #4 DB::ISerialization::serializeBinaryBulkWithMultipleStreams
  #5 DB::NativeWriter::writeData
  #6 DB::NativeWriter::write
  #7 DB::TCPHandler::sendData
  #8 DB::TCPHandler::processOrdinaryQuery
  ...

freed by thread T645 (ThreadPool):                     ← DIFFERENT THREAD
  #0 __asan_free
  ...
  #3 DB::ColumnAggregateFunction::~ColumnAggregateFunction()
  #5 DB::finalizeChunk(DB::Chunk&, ...)
  #6 DB::CubeTransform::generate()
  #7 DB::IAccumulatingTransform::work()
```

2650-byte size matches `HyperLogLogWithSmallSetOptimization<long, 16, 12>` internal storage.

## TSan iterator invalidation (inst-13)

Producer:

```sql
SELECT medianStateOrDefault(b), g FROM _s1 GROUP BY g WITH CUBE QUALIFY isNotNull(g);
```

TSan:

```
WARNING: ThreadSanitizer: heap-use-after-free (pid=3876495)
  Read of size 8 at 0xfffff2030000 by thread T6:
    #0 DB::AggregatedDataVariants::sizeWithoutOverflowRow() const
    #1 DB::Aggregator::prepareChunkAndFillSingleLevel<true>
    #2 DB::Aggregator::mergeBlocks
    #3 DB::GroupByModifierTransform::merge
    #4 DB::CubeTransform::generate
    ...

  Previous write of size 8 at 0xfffff2030000 by thread T6 (SAME thread):
    #0 realloc                                          ← realloc frees old buffer
    #1 Allocator<false, false>::realloc
    #2 PODArrayBase::reserveForNextSize<>()
    #3 ReservoirSampler<long>::merge
    #4 AggregateFunctionQuantile<long, QuantileReservoirSampler, ...>::merge
    #5 AggregateFunctionState::merge
    #6 AggregateFunctionOrFill<false>::merge
    #7 IAggregateFunctionHelper::mergeBatch
    #8-10 Aggregator::mergeStreamsImpl<HashMethodOneNumber<char8_t>>
    #11 Aggregator::mergeBlocks                          ← SAME mergeBlocks call as the read
    #12 GroupByModifierTransform::merge
    #13 CubeTransform::generate                          ← SAME CubeTransform call
```

Same-thread iterator invalidation: `mergeStreamsImpl` iterates the hash map, the recursive `ReservoirSampler::merge` grows a PODArray (realloc → free old buffer), then `Aggregator::prepareChunkAndFillSingleLevel` reads the stale `AggregatedDataVariants` pointer.

## ASan same-thread (inst-09)

Producer:

```sql
SELECT medianStateOrNullArgMax(b, 1025), g FROM _s1 GROUP BY g WITH CUBE
INTERSECT ALL
SELECT g, medianStateOrNull(b) FROM _s1 GROUP BY g WITH CUBE WITH TOTALS;
```

Same thread (TCPHandler T6 → T6). Pipeline `Aggregator → CubeTransform → IntersectOrExceptTransform → NativeWriter`. `CubeTransform::finalizeChunk` destructs the `ColumnAggregateFunction` (frees `QuantileReservoirSampler` state). `IntersectOrExceptTransform::filter → hashRow → ColumnAggregateFunction::updateHashWithValue → ReservoirSampler::write` reads the freed PODArray.

## ASan with ROLLUP+DISTINCT (inst-10, inst-11)

```sql
SELECT DISTINCT
    groupArrayMovingAvgStateDistinctOrDefaultDistinctStateOrNull(b) IGNORE NULLS,
    NOT (a >= NULL)
FROM _s1 GROUP BY g WITH ROLLUP QUALIFY materialize(1024);
```

Same `CubeTransform finalizeChunk` vs `TCPHandler serializeBatch` cross-thread UAF. Confirms the bug fires with `WITH ROLLUP` (not just `CUBE`) and `DISTINCT` as the row-multiplier (not just JOIN / TOTALS).

inst-11 also reproduced it with a `rank() OVER ROWS BETWEEN` window function — confirms the bug is in **column lifecycle**, not in state-aggregator interaction.

## Why this didn't show up on release/debug builds (i.e., why this is the producer for #102460)

In release/debug/ubsan jemalloc builds:
- The first destructor frees the aggregate-state memory back to jemalloc.
- The second destructor reads from memory that might still be valid (not reused) — appears to work.
- But jemalloc's internal bookkeeping is *also* updated twice → eventually corrupts the slab metadata → crash in `eset_remove` / `bin_slabs_full_remove` (which is exactly the recurring #102460 family).

ASan/TSan/MSan all catch the cause; jemalloc catches the downstream effect on different cycles.

## Likely fix area

`src/Processors/Transforms/finalizeChunk.cpp` (or wherever `finalizeChunk` is defined) — make it COW-correct:

- Before destructing or replacing a `ColumnAggregateFunction`, check `use_count > 1` and clone-rather-than-modify-in-place.
- Or have callers (`CubeTransform::generate`, `TotalsHavingTransform::transform`, etc.) only call `finalizeChunk` after they own the chunk uniquely.

Two angles for the aggregator-side fix (the TSan-caught iterator-invalidation half of the picture):

- `AggregatedDataVariants::sizeWithoutOverflowRow()` — this getter shouldn't be touching memory that the merge path can realloc. Possibly a cached pointer that becomes stale after the inner mergeStream's realloc.
- `Aggregator::mergeStreamsImpl` → `mergeBatch` — the iteration captures a reference/pointer into the hash map's slab; the recursive `merge` chain inside `mergeBatch` can grow / move the same slab's contents. Need to either reserve capacity upfront so realloc can't fire during the iteration, or refetch the iteration cursor / pointers after every cell merge.
(aggregator), @Algunenano (CubeTransform / finalizeChunk).

### LINKED ISSUE #105742: SEGV in `UniquesHashSet::merge` ← `ColumnAggregateFunction::ensureOwnership` ← `MergeSorter::mergeBatchImpl` when sorting aggregate-state column from `WITH ROLLUP WITH TOTALS` (5-line standalone repro)
# Triage context

Found by the server-side AST fuzzer + oracle harness (PR #99980). Has a **5-line standalone minimal repro** that crashes the server immediately on master `840494d51c9` with default settings — no fuzzer needed.

---

## Setup (self-contained)

```sql
CREATE TABLE _s1 (
    a UInt32, b Nullable(Int64), c String, d Float64,
    e Nullable(String), f Date, g UInt8, h Nullable(UInt16)
) ENGINE = MergeTree ORDER BY a;

CREATE TABLE _s2 (x UInt32, y String, z Nullable(UInt8))
ENGINE = MergeTree ORDER BY x;

INSERT INTO _s1 SELECT
    number,
    if(number%3=0, NULL, number*10),
    toString(number%7),
    number*0.1-50,
    if(number%7=0, NULL, toString(number)),
    toDate('2020-01-01') + number,
    number%3,
    if(number%5=0, NULL, toUInt16(number))
FROM numbers(500);

INSERT INTO _s2
SELECT number, toString(number % 5), if(number % 4 = 0, NULL, toUInt8(number % 200))
FROM numbers(300);
```

No special settings required. Default settings on master crash.

---

# SEGV in `UniquesHashSet::merge` ← `ColumnAggregateFunction::ensureOwnership` ← `MergeSorter::mergeBatchImpl` when sorting aggregate-state column from `WITH ROLLUP WITH TOTALS`

## Minimal repro (5 lines)

```sql
SELECT s1.g, uniqStateOrDefault(b)
FROM _s1 AS s1
LEFT JOIN _s2 AS s2 ON s1.a = s2.x
GROUP BY s1.g WITH ROLLUP WITH TOTALS
ORDER BY s1.g DESC NULLS FIRST
FORMAT Null;
```

→ Server SEGVs immediately. Verified on master `840494d51c9` release **and** ASan builds. Faulting address: `NULL`.

## Stack trace

```
3. src/AggregateFunctions/UniquesHashSet.h:453        UniquesHashSet<TrivialHash>::merge    ← NULL deref of buf
4. src/Columns/ColumnAggregateFunction.cpp:258        ColumnAggregateFunction::ensureOwnership
5. src/Columns/ColumnAggregateFunction.cpp:649        ColumnAggregateFunction::insertDefault (inlined)
5. src/Columns/ColumnAggregateFunction.cpp:558        ColumnAggregateFunction::insertFromWithOwnership
6. src/Columns/ColumnAggregateFunction.cpp:316        ColumnAggregateFunction::insertRangeFrom
7. src/Processors/Transforms/SortingTransform.cpp:136 MergeSorter::mergeBatchImpl
8. src/Processors/Transforms/SortingTransform.cpp:85  MergeSorter::read
9. src/Processors/Transforms/MergeSortingTransform.cpp:312  MergeSortingTransform::generate
10. src/Processors/Transforms/SortingTransform.cpp:391 SortingTransform::work
```

`UniquesHashSet::merge` at line 453 dereferences a NULL `buf` belonging to the empty/default `uniqState` instance that the totals/rollup row produced.

## Trigger ingredients (necessary)

- **`uniqStateOrDefault(...)`** — aggregate **State** combinator that produces a `ColumnAggregateFunction`. (Also reproduces with bare `uniqState(...)`; not tested with other State aggregates but suspected family-wide.)
- **`WITH ROLLUP WITH TOTALS`** — emits a totals row whose aggregate-state column is filled with `insertDefault`.
- **`ORDER BY`** — routes the result through `MergeSortingTransform` → `MergeSorter::mergeBatchImpl`, which calls `insertRangeFrom` on every column including the aggregate-state column, hitting `insertFromWithOwnership` → `insertDefault` → `ensureOwnership` on the totals row.
- The LEFT JOIN appears non-essential but ensures the optimiser doesn't fold the aggregation away.

## Independent of the `finalizeChunk` UAF family

This bug is **not** related to the consolidated `finalizeChunk` aliased-column UAF (issue forthcoming) that root-causes [#102460](https://github.com/ClickHouse/ClickHouse/issues/102460). ASan doesn't report a poison-pattern access here — the NULL deref happens before ASan's poisoning would matter. Different code path, different fix area.

## Likely fix area

`src/Columns/ColumnAggregateFunction.cpp:258` (`ensureOwnership`) — handle the default-state case where the source state pointer is NULL and we're just materialising a fresh empty state in the new arena. Possibly `insertDefault` should construct the empty state in-place via the aggregate function's `createDefault` instead of going through `insertFromWithOwnership`.

## Related DistinctTransform variant (not yet minimised standalone)

A `SELECT DISTINCT ... uniqStateOrDefault(c) ... GROUP BY ... WITH TOTALS` shape captured in a separate snapshot (`inst-07`) trips a NULL deref in `ColumnAggregateFunction::updateHashWithValue` ← `DistinctTransform::buildFilter`. Same root cause (NULL aggregate-state on default/totals row), different downstream consumer (`DistinctTransform` instead of `MergeSorter`). Likely fixed by the same `ensureOwnership` / `insertDefault` correction. Standalone minimisation pending — the more aggressive fuzz mutations matter; left as-is.

groeneai:
Issue #105742, also fixes #105740 (and root cause of chronic #102460).

Stack on master (`840494d51c9`):

```
0. UniquesHashSet<...>::merge      AggregateFunctions/UniquesHashSet.h:453   (NULL deref of buf)
1. ColumnAggregateFunction::ensureOwnership      Columns/ColumnAggregateFunction.cpp:258
2. ColumnAggregateFunction::insertDefault        Columns/ColumnAggregateFunction.cpp:649
3. ColumnAggregateFunction::insertFromWithOwnership  Columns/ColumnAggregateFunction.cpp:558
4. ColumnAggregateFunction::insertRangeFrom      Columns/ColumnAggregateFunction.cpp:316
5. MergeSorter::mergeBatchImpl                   Processors/Transforms/SortingTransform.cpp:136
```

Reproducer (5 lines after creating the source tables from the issue):

```sql
SELECT s1.g, uniqStateOrDefault(b)
FROM _s1 AS s1
LEFT JOIN _s2 AS s2 ON s1.a = s2.x
GROUP BY s1.g WITH ROLLUP WITH TOTALS
ORDER BY s1.g DESC NULLS FIRST
FORMAT Null;
```

### Root cause

`ColumnAggregateFunction::convertToValues` calls `func->insertResultInto` for each row in the source column. For State combinators (e.g. `AggregateFunctionState::insertResultInto`) this path is optimized to push the source state pointer into `res.data` without copying, relying on `res.src` to keep the source column alive.

`uniqStateOrDefault` is `AggregateFunctionOrFill<false>` around `AggregateFunctionState(AggregateFunctionUniq)`. `OrFill::insertResultIntoImpl` has two branches:
- flag=1 (state was populated): delegates to `nested_function->insertResultInto` -> `State::insertResultInto` -> pushes `place` directly.
- flag=0 (no input rows ever added): calls `to.insertDefault()`.

`ColumnAggregateFunction::insertDefault` calls `ensureOwnership` first, and `ensureOwnership` always finishes with `src.reset()`. So once a single flag=0 row is processed, `res.src` is null. Any subsequent flag=1 row still pushes a raw pointer into `res.data`, but `res` no longer holds a reference to the source column. When `convertToValues` returns and the source column gets destructed, the source column's destructor calls `func->destroy` on every state in its `data`, and those destructors zero out internal pointers (for `uniq`, `UniquesHashSet::~UniquesHashSet -> free()` sets `buf = nullptr`, while `m_size` is unchanged at its last value, e.g. 33).

Sorting then read those pointers (`MergeSortingTransform -> insertRangeFrom -> ensureOwnership -> IAggregateFunction::merge`) and dereferenced the destructed state, segfaulting at `UniquesHashSet.h:453` (`rhs.buf[i]`).

### Relation to #105740 / #102460

`convertToValues` is the call path behind every `finalizeChunk(chunk, mask)` (it walks each masked column and replaces it with `convertToValues(...)`). All four transforms that produce finalized aggregate columns - `CubeTransform`, `RollupTransform`, `TotalsHavingTransform`, `AggregatingInOrderTransform` - go through the same call site. So the four sanitizer signatures qoega listed in #105740 (TSan iterator invalidation under CubeTransform, ASan single/cross-thread, MSan via TotalsHavingTransform) and the chronic jemalloc heap-corruption surfaces in #102460 are different downstream observations of the same upstream aliasing bug fixed here. `singleValueOrNullStateOrDefault`, `argMaxStateOrDefault`, etc. are all State combinators that share this code path, so the fix is general.

### Fix

Use `insertMergeResultInto` in `convertToValues` instead of `insertResultInto`. For State combinators (`State`, `OrFill`, `Null`, `Resample`, `ForEach`, ...) `insertMergeResultInto` forces a real copy via `to.insertFrom(place)`, so every entry in `res.data` ends up owned by `res`. For non-State aggregate functions the default `IAggregateFunction::insertMergeResultInto` just delegates to `insertResultInto`, so behavior is unchanged for them.

### Tests

- `04275_105742_uniq_state_or_default_rollup_order_by.sql` - `WITH ROLLUP`, `WITH ROLLUP WITH TOTALS`, `WITH CUBE`, bare `uniqState`, each followed by `ORDER BY`. Pre-fix: UBSan null pointer load at `UniquesHashSet.h:453`. Post-fix: clean.
- `04276_105740_finalize_chunk_state_combinator.sql` - the MSan producer from #105740 (`singleValueOrNullStateOrDefault` + RIGHT JOIN + `WITH TOTALS` + `ORDER BY`) plus `uniqStateOrDefault` and `argMaxStateOrDefault` variants covering the `TotalsHavingTransform` and `CubeTransform` paths. Pre-fix: UBSan null pointer load at `UniquesHashSet.h:453` on the CUBE variant. Post-fix: clean on both Debug and ASan.

Existing related tests still pass (`00701_rollup`, `01122_totals_rollup_having_block_header`, `00388_enum_with_totals`, `00150_with_totals_and_join`, `00597_with_totals_on_empty_set`, `00906_low_cardinality_rollup`, `00804_rollup_with_having`).

- [x] Documentation update is not required

Fix server crash and use-after-free in `ColumnAggregateFunction::convertToValues` when materializing a `-State` aggregate column from `WITH ROLLUP`, `WITH CUBE` or `WITH TOTALS` followed by `ORDER BY`. The crash manifested as a NULL dereference in `UniquesHashSet::merge` (for `uniq*State`) or sanitizer use-after-free / use-of-uninitialized (for other State combinators such as `singleValueOrNullState` and `argMaxState`).

Closes https://github.com/ClickHouse/ClickHouse/issues/105742


groeneai:
Pre-PR validation gate (TASK.md Phase 4 step 9):

a) Deterministic repro: yes - 5-line query from #105742 SEGVs on master `41656be24cc8`, UBSan reports load of null pointer at `UniquesHashSet.h:453`. Reproduces with smaller queries too (no JOIN, ~10 rows) as long as one group has all-NULL inputs and the result is ORDER-BY'd.

b) Root cause explained: yes. `convertToValues` pushes shared pointers via `State::insertResultInto`, relies on `res.src` to keep source alive. `OrFill::insertResultIntoImpl` flag=0 calls `res.insertDefault -> res.ensureOwnership` which resets `res.src`. Subsequent flag=1 rows then push raw pointers with no reference holding the source; when the source column gets destructed at end of `convertToValues`, those pointers dangle into destructed states. Detailed walkthrough in the PR body.

c) Fix matches root cause: yes. `insertMergeResultInto` makes every State-combinator state get copied into `res`'s own arena (`to.insertFrom(place)`), so `res.data` is fully owned by `res` regardless of how many flag=0/flag=1 transitions happen during the loop. Source column can be destroyed safely after `convertToValues` returns.

d) Test intent preserved + new tests added: yes. `04275_105742_uniq_state_or_default_rollup_order_by.sql` exercises `WITH ROLLUP`, `WITH ROLLUP WITH TOTALS`, `WITH CUBE`, and bare `uniqState` followed by `ORDER BY`. Existing related tests still pass (`00701_rollup`, `01122_totals_rollup_having_block_header`, `00388_enum_with_totals`, `00150_with_totals_and_join`, `00597_with_totals_on_empty_set`, `00906_low_cardinality_rollup`, `00804_rollup_with_having`).

e) Demonstrated in both directions: yes. Verified by stashing the fix, rebuilding ASan, and running the new test - UBSan fires at `UniquesHashSet.h:453` with summary `undefined-behavior /home/ubuntu/nerve-workspace/ClickHouse/src/AggregateFunctions/UniquesHashSet.h:453:17`. After restoring the fix and rebuilding, the test prints `ok`.

f) Fix is general: yes. The fix is at the convertToValues call site, not at the symptom site. All State combinators (`-State`, `-OrFill` (`-OrDefault`/`-OrNull`), `-Null`, `-Resample`, `-ForEach`) override `insertMergeResultInto` to copy via `to.insertFrom(place)`. So any combinator that depended on the inner state's destructor (not just `uniq`) is now safe. Non-State functions fall through to the default `IAggregateFunction::insertMergeResultInto` which delegates to `insertResultInto`, so they keep the existing fast path.

Session: cron:clickhouse-ci-task-worker:20260524-194500

groeneai:
cc @alexey-milovidov @rschu1ze @KochetovNicolai - could you review? This fixes a server SEGV in any `-State` aggregate that runs through `ColumnAggregateFunction::convertToValues` (#105742). The patch is a one-line change in `convertToValues` (use `insertMergeResultInto` instead of `insertResultInto`) plus a regression test.

groeneai:
Pushed `f0dbe2f14c30` to address the `Bugfix validation (functional tests)` failure.

Root cause of the validation failure (not the original bug): the `.sql` regression tests crash the shared `clickhouse-server`, which is the exact bug they reproduce. The crash is observable in the validation artifacts (`core.QueryPullPipeEx.1588-3969.zst.enc`, stack ending in `ColumnAggregateFunction::convertToValues`). What goes wrong from there is in the harness, not the bug:

- `clickhouse-test`'s `check_server_liveness` detects the dead server, logs `Unable to locate any ClickHouse server process`, calls `stop_tests()`, sleeps 5s, then `p.terminate()`s the worker pool (`tests/clickhouse-test:4929-4942`).
- During that cleanup window a second `SIGTERM` interrupts the runner. The first signal is caught by the top-level `Terminated` handler at line 6609, but the next signal hits inside `print(...)` and re-raises, so the runner exits with code `1` instead of `STOP_TESTING_EXIT_CODE = 2`.
- `ci/jobs/scripts/functional_tests_results.py:191` only synthesizes the `"Server died"` `FAIL` leaf when `runner_exit_code == STOP_TESTING_EXIT_CODE`. With exit `1` it falls through to `elif not s.success_finish` → `state = ERROR`, info `"The test runner was terminated unexpectedly"`, and the per-test list is empty.
- The bugfix inversion at `functional_tests.py:751-762` then iterates `test_result.results` (now empty), finds nothing to flip, sets `has_failure = False`, and prints `"Failed to reproduce the bug"` despite the core dump.

Fix: run the same queries via `clickhouse-local` instead of against the shared server. The `SEGV` is now contained in a subprocess; the `.sh` exits with code `139`; `clickhouse-test` records `EXIT_CODE` `FAIL`; the bugfix inversion flips it to `OK` with `has_failure = True`. On the PR (fix in place) the queries finish and the script prints `ok`, matching the unchanged `.reference`. The fix itself is unchanged in this push (only the test files moved from `.sql` to `.sh`).

Locally verified on master HEAD (no fix): both tests fail with `return code: 139` under `clickhouse-test`.

qoega:
@groeneai can you create an issue to fix bug check harness logic and treat crash on old version as fail. Once created also implement a fix for it.

groeneai:
@qoega filed the tracking issue and a fix PR for the bugfix-validation harness side of the problem you flagged:

- Issue: #105789 — full root-cause breakdown across three layers (signal handler, parser, inverter).
- Fix PR: #105795 — "Preserve ERROR status in bugfix-validation inverter". Short-circuits the inverter when `test_result.status == Result.Status.ERROR` so the run-level ERROR (and its info) survives instead of being overwritten with "Failed to reproduce the bug". 8 unit tests; both directions verified against an inlined copy of the pre-fix inverter.

Companion to #105643 (@leshikus) which expands the parser's recognized exit codes so SIGTERM/SIGKILL-killed runs surface a synthetic `Server died` FAIL leaf. Together: #105643 catches signal-killed crashes; #105795 catches every other path that ends in ERROR.

qoega:
- 01018_Distributed__shard_num looks flaky. I do not see open issue about it. If it is not present please create one. 
- Try to fix 01018_Distributed__shard_num in separate PR
- Update this PR to rerun CI

When you see test failures in CI it is good to check if it is related to the change. If not you should check if it was failing on master and in different PRs and perform steps above for every new CI failure.

@groeneai

PedroTadim:
Closes https://github.com/ClickHouse/ClickHouse/issues/105742

azat:
>(and root cause of chronic https://github.com/ClickHouse/ClickHouse/issues/102460).

@groeneai and @qoega  I don't think it is related to this issue, the repro there does not have any WITH ROLLUP/CUBE/TOTALS

azat:
And I think https://github.com/ClickHouse/ClickHouse/pull/105470 is a better fix, since it does not have overhead for other paths, does it fixes all issues that this PR addresses?

groeneai:
@azat you're right on both counts, thanks for the careful review.

**1. The #102460 link was wrong.** #102460 is the jemalloc/LTO fantom-crash tracker, no ROLLUP/CUBE/TOTALS involved. I conflated it via the #105740 fuzzer report. Dropping that claim.

**2. #105470 is the better fix and already covers #105742.** #105470 (merged 256c1a5a0f1b, 2026-05-26) removes the `to.insertDefault()` -> `ensureOwnership()` -> `src` reset for State-nested `-OrFill` combinators at the source, which is exactly the mechanism behind #105742 (`uniqStateOrDefault` = uniq+State+OrFill, synthesized empty-group row hits the flag=0 path). My `convertToValues` change here is downstream of that and adds `insertMergeResultInto` overhead on every state path, so it's redundant now.

Verified on current master `26.6.1.338` (includes #105470):
- #105742 5-line repro (LEFT JOIN + WITH ROLLUP WITH TOTALS + ORDER BY): no crash, exit 0.
- This PR's own test queries (ROLLUP / ROLLUP WITH TOTALS / CUBE over `uniqStateOrDefault`): no crash.
- CIDB: zero `UniquesHashSet` master crashes since #105470 merged.

So #105742 is resolved by #105470; closing this PR as superseded.

One separate note: #105740 (`finalizeChunk` mutating `ColumnAggregateFunction` columns still aliased by downstream transforms, sanitizer-only) is a distinct, broader root cause that neither #105470 nor this PR fully addresses. I'll keep it as its own open item rather than fold it in here.

@qoega heads-up since you assigned the original directive.

groeneai:
Closing as superseded by #105470 (merged), which fixes #105742 at the source (AggregateFunctionOrFill State-path src reset) without the per-path overhead of this change. See above for verification on master 26.6.1.338.

PR #101653, opened 2026-04-02
Better default selectivities for unknown and LIKE conditions

davenger:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Improved default selectivity estimates for conditions without statistics to match industry-standard heuristics: 0.1 for `LIKE` condition selectivity, 0.33 for unknown condition selectivity. 

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.1020`

PR #105757, opened 2026-05-24
Add regression test for LEFT ANTI JOIN projection with WHERE (#105738)

groeneai:
Adds a regression test for [issue #105738](https://github.com/ClickHouse/ClickHouse/issues/105738) (sibling reproducer for [#99959](https://github.com/ClickHouse/ClickHouse/issues/99959)).

The actual fix landed in [#105278](https://github.com/ClickHouse/ClickHouse/pull/105278): `HashJoinResult::appendRightColumns` now emits type defaults instead of copying left-key values for `LEFT ANTI JOIN`. This PR is the test-only follow-up requested by @fm4v on the issue.

The reproducer self-joins a `MergeTree` table on `m.c2 = right_0.c1`. Pre-fix, the baseline (no `WHERE`) returned the left-key value, while adding `WHERE m.c0` collapsed to `0`. The test covers `hash`, `parallel_hash` and `grace_hash` algorithms.

...

### Version info
- Merged into: `26.6.1.138`

### DIFF
--- a/tests/queries/0_stateless/04275_105738_left_anti_join_where_projection.reference
+++ b/tests/queries/0_stateless/04275_105738_left_anti_join_where_projection.reference
@@ -1,0 +1,12 @@
+--- LEFT ANTI JOIN, baseline (no WHERE) ---
+0
+0
+--- LEFT ANTI JOIN, with WHERE on left ---
+0
+0
+--- LEFT ANTI JOIN, parallel_hash ---
+0
+0
+--- LEFT ANTI JOIN, grace_hash ---
+0
+0
--- a/tests/queries/0_stateless/04275_105738_left_anti_join_where_projection.sql
+++ b/tests/queries/0_stateless/04275_105738_left_anti_join_where_projection.sql
@@ -1,0 +1,42 @@
+-- Regression test for https://github.com/ClickHouse/ClickHouse/issues/105738
+-- Sibling reproducer for https://github.com/ClickHouse/ClickHouse/issues/99959.
+-- In a LEFT ANTI JOIN, the result contains only rows from the left table that have no
+-- matching row in the right table. The right-table columns (including the join key)
+-- must contain default values for those unmatched rows, not the left key value.
+-- The #105738 reproducer self-joins a MergeTree table and varies the presence of a
+-- WHERE predicate, which previously could pick different plans and expose the
+-- right-key projection inconsistency.
+
+DROP TABLE IF EXISTS m_105738;
+
+CREATE TABLE m_105738 (c0 Int32, c1 UInt64, c2 UInt64) ENGINE = MergeTree() ORDER BY (-c0);
+INSERT INTO m_105738 VALUES (1, 100, 999);
+INSERT INTO m_105738 VALUES (2, 200, 888);
+OPTIMIZE TABLE m_105738 FINAL;
+
+SELECT '--- LEFT ANTI JOIN, baseline (no WHERE) ---';
+SELECT right_0.c1
+FROM m_105738 LEFT ANTI JOIN m_105738 AS right_0 ON m_105738.c2 = right_0.c1
+ORDER BY 1;
+
+SELECT '--- LEFT ANTI JOIN, with WHERE on left ---';
+SELECT right_0.c1
+FROM m_105738 LEFT ANTI JOIN m_105738 AS right_0 ON m_105738.c2 = right_0.c1
+WHERE m_105738.c0
+ORDER BY 1;
+
+SELECT '--- LEFT ANTI JOIN, parallel_hash ---';
+SELECT right_0.c1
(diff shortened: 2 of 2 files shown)


groeneai:
### Pre-PR validation gate

Test-only follow-up to merged PR #105278.

- **(a) Deterministic repro?** Yes. The four queries return constant results on a small fixed dataset. 30 local runs with randomized settings: 30 OK.
- **(b) Root cause explained?** Yes. `HashJoinResult::appendRightColumns` previously called `copyLeftKeyColumnToRight` unconditionally, which copies the left-side join key into the right-side column. For `LEFT ANTI JOIN`, surviving rows are by definition unmatched, so the right side must be defaults. The `WHERE m.c0` predicate changes plan selection (via the selectivity estimator from #101653), which is why fm4v originally observed the inconsistency between the two queries.
- **(c) Fix matches root cause?** N/A. Fix already landed in #105278: `HashJoinResult::appendRightColumns` now emits `right_key.type->createColumn()->insertManyDefaults(block.rows())` when `kind() == Left && strictness() == Anti`.
- **(d) Test intent preserved? / New tests added?** New test added: `tests/queries/0_stateless/04275_105738_left_anti_join_where_projection.sql`. Existing `04256_99959_left_anti_join_right_key_default.sql` is untouched.
- **(e) Demonstrated in both directions?** Yes. Pre-fix the baseline (no `WHERE`) returned `888, 999` (documented in #105738 and in the #105278 PR description). With #105278 applied, the test returns `0, 0` for all four cases. Reverting the `is_left_anti_join` branch in `HashJoinResult.cpp` would trigger an "Output differs" failure on the first query.
- **(f) Fix is general?** N/A (test-only). The fix in #105278 already handles the general case via the `kind() == Left && strictness() == Anti` predicate, which covers all join algorithms (`hash`, `parallel_hash`, `grace_hash`). This test exercises all three.

Session: cron:clickhouse-ci-task-worker:20260524-234500

groeneai:
cc @vdimir for review (comp-joins owner, approved the underlying fix in #105278).

This is the test-only follow-up requested by @fm4v after #105278 merged. Adds the #105738 reproducer (`LEFT ANTI JOIN` self-join with `WHERE` on left) to `tests/queries/0_stateless/`, covering `hash`, `parallel_hash` and `grace_hash` algorithms.

rienath (approved):

fm4v:
@groeneai test expected result is incorrect, revalidate and check what result should be expected 

```
How to reproduce
DROP TABLE IF EXISTS m;
CREATE TABLE m (c0 Int32, c1 UInt64, c2 UInt64) ENGINE = MergeTree() ORDER BY (-c0);
INSERT INTO m VALUES (1, 100, 999);
INSERT INTO m VALUES (2, 200, 888);
OPTIMIZE TABLE m FINAL;

-- Baseline: correct, returns 888 / 999
SELECT right_0.c1
FROM m LEFT ANTI JOIN m AS right_0 ON m.c2 = right_0.c1
ORDER BY 1;

-- With WHERE: returns 0 / 0  -- BUG
SELECT right_0.c1
FROM m LEFT ANTI JOIN m AS right_0 ON m.c2 = right_0.c1
WHERE m.c0
ORDER BY 1;
Expected behavior
Both queries should return the same projected values:

888
999
Actual behavior
The second query returns:

0
0
```

groeneai:
Re-ran your reproducer on current master HEAD (commit `9efe9c11707`, which includes #105278):

```
-- baseline (no WHERE):       0, 0
-- WHERE m.c0:                0, 0
-- parallel_hash, WHERE:      0, 0
-- grace_hash, WHERE:         0, 0
```

The test reference (`0, 0` everywhere) matches actual post-fix behavior on master.

The `888, 999` you cited in the issue body was the pre-fix observation. Pre-fix, `HashJoinResult::appendRightColumns` always called `copyLeftKeyColumnToRight`, which copies the left join key value into the right column. For `LEFT ANTI JOIN`, surviving rows are by definition unmatched, so the right side has no real value to project. Copying the left key value was the bug, not the correct answer. The `WHERE m.c0` variant happened to return `0, 0` pre-fix because the planner picked a different path due to changed selectivity after #101653.

#105278 (approved and merged) makes both paths emit type defaults via `insertManyDefaults`. From the PR description:

> For `LEFT ANTI` the kept rows are precisely the unmatched ones, so copying produces the wrong value. The fix detects `kind() == Left && strictness() == Anti` in `appendRightColumns` and emits a column of type defaults of the required right-key type instead.

This matches PostgreSQL, MySQL and DuckDB, which all return NULL or default for unmatched right columns in `LEFT ANTI`.

If you want the test to assert something else, the two options are:

1. Lock in the consistency invariant only (`baseline == WHERE`) without pinning a specific value. Still catches regressions if the two paths diverge.
2. Revisit #105278's semantic decision with the joins owners. That is the only route to a `888, 999` expectation in the test.

Which would you like?

Session: cron:clickhouse-ci-task-worker:20260525-113000

PR #105787, opened 2026-05-25
Document NaN bit-pattern handling in set semantics

### LINKED ISSUE #105748: DISTINCT does not deduplicate NaN on AArch64 (regression from #98230, fastops on ARM)
## Describe the bug

On **AArch64 only**, `DISTINCT` does not deduplicate two NaN values produced by different operations, because they have different sign bits. Same query on x86_64 returns the expected (deduplicated) result.

## How to reproduce

```sql
SELECT countDistinct(arrayJoin([0./0., log(-1.)]));
```

- **x86_64** (any version): `1`
- **AArch64**, v26.4+: `2` ← bug

The bit patterns explain it (the comparison below is `UInt64 = UInt64` after reinterpret, since `NaN = NaN` is always false at the float level):

```sql
SELECT hex(reinterpretAsUInt64(toFloat64(0./0.)))    AS a,  -- 0x7FF8… on ARM, 0xFFF8… on x86
       hex(reinterpretAsUInt64(toFloat64(log(-1.)))) AS b,  -- 0xFFF8… on both
       reinterpretAsUInt64(toFloat64(0./0.)) =
       reinterpretAsUInt64(toFloat64(log(-1.)))      AS bit_eq;
-- bit_eq: 1 on x86, 0 on ARM
```

`DISTINCT` compares values byte-wise, so two NaNs with different sign bits don't dedupe on ARM.

## Bisect

Bisected on AArch64 (Graviton, public CI ARM builds) across blessed/master first-parent between `4261f8f2` (v26.3.1.154, 2026-02-28) and `33908541` (v26.4.1.855, 2026-04-12) — 1577 commits, 16 steps.

**First bad: `627bb97a551f2db50759ed5cdeac5977b0710b2f` — #98230 "Enable fastops vectorized math on AArch64 and on FreeBSD/Darwin"**.

Verified at the boundary on ARM:

| Commit | `log(-1.)` | DISTINCT count |
|---|---|---|
| `86aefd37` (last good, pre-fastops on ARM) | `0x7FF8…` | 1 |
| `627bb97a` (#98230, fastops on ARM) | `0xFFF8…` | **2** |

After the PR, fastops' NEON `log` returns a NaN with the opposite sign bit from glibc's scalar `log`, so `log(-1.)` and `0./0.` (which still uses the canonical positive NaN) no longer match byte-wise.

## Expected behavior

Either:
1. `DISTINCT`/hashing canonicalizes NaN before byte-comparison so all NaNs collapse (matches x86_64 and prior ARM behavior), or
2. fastops' ARM kernels return a NaN whose sign matches the scalar libm path.

## Affects

Built from commit `627bb97a` forward — present in **v26.4 onward on AArch64**. Not present on x86_64.

### Version info
- Resolved by: #105787
- Merged into: `26.6.1.132`

groeneai:
Follow-up to discussion on #105756 (closed in favor of a docs-only path
per @Algunenano's review). Closes #105748.

Adds a section to `docs/en/sql-reference/data-types/float.md` explaining
that hash tables compare keys byte-wise, so different `NaN` bit patterns
(e.g. `0./0.` vs literal `nan` vs ARM NEON `log(-1.)` after #98230) are
treated as distinct by `DISTINCT`, `GROUP BY`, `uniqExact`,
`countDistinct`, and equi-`JOIN` on a `Float` key. Shows two
canonicalization patterns users can apply when they need NaN-equal set
semantics.

### Version info
- Merged into: `26.6.1.132`

### DIFF
--- a/docs/en/sql-reference/data-types/float.md
+++ b/docs/en/sql-reference/data-types/float.md
@@ -112,6 +112,47 @@ SELECT 0 / 0
 
 See the rules for `NaN` sorting in the section [ORDER BY clause](../../sql-reference/statements/select/order-by.md).
 
+## NaN values in set semantics {#nan-values-in-set-semantics}
+
+The IEEE 754 standard defines `NaN` such that the scalar comparison `NaN = NaN` returns `false`.
+ClickHouse follows that rule for the `=` operator.
+
+However, `NaN` is not a single value; it is any bit pattern whose exponent is all ones and whose
+mantissa is non-zero. Different operations and different CPU architectures can produce `NaN`
+values with different sign bits or different mantissa payloads. For example:
+
+- `0./0.` produces a `NaN` whose sign bit is 1 on most x86 platforms.
+- The literal `nan` produces a `NaN` whose sign bit is 0.
+- After [PR #98230](https://github.com/ClickHouse/ClickHouse/pull/98230), the AArch64 NEON path of
+  `log` returns a `NaN` whose sign bit differs from glibc's scalar `log` on negative inputs.
+
+Hash tables in ClickHouse compare keys byte-wise, so different `NaN` bit patterns hash to
+different buckets and are treated as distinct values by set-semantics operations including
+`DISTINCT`, `GROUP BY`, `uniqExact`, `countDistinct`, and equi-`JOIN` on a `Float` key:
+
+```sql
+SELECT countDistinct(arrayJoin([0./0., nan, log(-1.)]));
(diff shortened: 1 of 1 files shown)


groeneai:
cc @Algunenano - this is the docs-only PR you agreed to on #105756. Adds a section to the Float types reference page explaining the byte-wise hash NaN split and shows two canonicalization patterns users can apply when they need NaN-equal set semantics.

Algunenano (approved):

PR #105795, opened 2026-05-25
Preserve ERROR status in bugfix-validation inverter

### LINKED ISSUE #105789: Bugfix validation harness reports 'Failed to reproduce the bug' when runner does not finish (overwrites genuine ERROR)
The `Bugfix validation (functional tests)` check silently overwrites a genuine `ERROR` from the test runner with a spurious "Failed to reproduce the bug" `FAIL` when no per-test results were emitted. This hides catastrophic runs (server crash on master, runner killed mid-flight, infrastructure outage) behind a misleading "the regression test does not catch the bug" report.

## Symptom

A `pr-bugfix` PR adds a regression test that crashes `clickhouse-server` on master (the exact intent of a regression test for a server-crash bug). The check reports `Failed to reproduce the bug`, when in fact the server did crash on master and the bug *was* reproduced. Concrete instance: PR #105749 with `04275_105742_uniq_state_or_default_rollup_order_by.sql` (deterministic `UniquesHashSet` use-after-destructor on master HEAD). A core dump existed in the artifacts, yet the harness reported "Failed to reproduce".

## Root cause

Three independent layers, each in a different file.

### Layer A: `tests/clickhouse-test` exits with the wrong code under parallel cleanup

When the server dies mid-run, `check_server_liveness` sets `stop_testing`, the main loop calls `p.terminate()` on each worker, and finally raises `StopTesting`. The outer handler at the bottom of `tests/clickhouse-test`:

```python
try:
    main(args)
except StopTesting as e:
    print(f"Test run aborted: {e}", file=sys.stderr)
    traceback.print_exc()
    sys.exit(STOP_TESTING_EXIT_CODE)  # = 2
```

is supposed to exit with `STOP_TESTING_EXIT_CODE`. But each worker's `signal_handler` re-raises as `Terminated`, `run_tests_array`'s `except KeyboardInterrupt` calls `stop_tests()` which `killpg`s the whole process group, hitting the parent. The parent's `signal_handler` then fires, `Terminated` is raised inside `print(...)` of the `StopTesting` handler, escapes, falls into the `except Terminated` branch, and exits with `128 + SIGTERM = 143` instead of `2`.

### Layer B: `ci/jobs/scripts/functional_tests_results.py` only recognises one exit code as "aborted run"

The parser at line 191 synthesises a `Server died` `FAIL` leaf only when `runner_exit_code == STOP_TESTING_EXIT_CODE`. Exit code 143 (and 137 for `SIGKILL`) falls through to `elif not s.success_finish`, which sets `state = Result.Status.ERROR` with no per-test leaf added. The "Server died" signal is lost.

### Layer C: `ci/jobs/functional_tests.py` bugfix-validation inverter overwrites genuine `ERROR`

The inverter at lines 750-767:

```python
if is_bugfix_validation and test_result and (...):
    has_failure = False
    for r in test_result.results:
        ...
        if r.status == Result.Status.FAIL:
            r.status = Result.Status.OK
            has_failure = True
        elif r.status == Result.Status.OK:
            r.status = Result.Status.FAIL
    if not has_failure:
        print("Failed to reproduce the bug")
        test_result.set_failed().set_info("Failed to reproduce the bug")
    else:
        test_result.set_success()
```

iterates `test_result.results` to flip `FAIL` to `OK`. When the list is empty (because Layer B did not synthesise a leaf), the loop runs zero times, `has_failure` stays `False`, and the `else` branch unconditionally calls `set_failed().set_info("Failed to reproduce the bug")`. The honest `Result.Status.ERROR` from Layer B is silently overwritten with the misleading "Failed to reproduce" `FAIL`.

## Required fixes

| Layer | File | Status |
| --- | --- | --- |
| A: parent signal handler robustness | `tests/clickhouse-test` (`StopTesting` handler near line 6604) | open |
| B: recognise SIGTERM/SIGKILL exit codes as "aborted run" | `ci/jobs/scripts/functional_tests_results.py` line 191 | in-flight in PR #105643 (@leshikus) |
| C: do not overwrite genuine `ERROR` with "Failed to reproduce the bug" | `ci/jobs/functional_tests.py` lines 750-767 | this issue, fix PR forthcoming |

Layers A and B together cover the "server crashed and was killed by signal" path. Layer C is needed independently to cover any path that leaves `test_result.results` empty or partial while `test_result.status` is `ERROR` (Python exception in runner, job-level timeout with non-SIGTERM signal, infrastructure outage). Without Layer C, those paths still produce a misleading "Failed to reproduce the bug" report.

## Reproducer

Any `.sql` regression test that crashes `clickhouse-server` on master HEAD, run as part of a `pr-bugfix` PR's `Bugfix validation (functional tests)` check. The original `04275_105742_uniq_state_or_default_rollup_order_by.sql` from PR #105749 (before its `.sh` workaround) is one. The `.sh` variant that invokes `clickhouse-local` keeps the SEGV in a subprocess (return code 139), which `clickhouse-test` records as `[FAIL] Reason: return code: 139` and which inverts cleanly to `OK`.

## Suggested fix for Layer C

When `test_result.status == Result.Status.ERROR` before inversion, skip the inversion entirely and preserve the original `ERROR` status. An `ERROR`-status run is inconclusive (no per-test data to flip), so claiming "Failed to reproduce the bug" on top of it is incorrect. The operator should see the underlying error and investigate. `FAIL` and `OK` paths continue to invert as before.

## Cross-links

- PR #105749 (working around Layers A+B+C by moving regression tests to `.sh` so the SEGV stays in a subprocess)
- PR #105643 (Layer B fix by @leshikus)
- Directive from @qoega on PR #105749: "create an issue to fix bug check harness logic and treat crash on old version as fail"

### Version info
- Resolved by: #105795
- Merged into: `26.6.1.197`

groeneai:
Closes #105789.

Cuts the false-negative `Failed to reproduce the bug` report that the `Bugfix validation (functional tests)` check produced when the test runner did not finish (server crash without proper exit code, runner Python exception, job-level timeout, infrastructure outage). PR #105749 hit this and worked around it by rewriting the regression test as `.sh` so the SEGV stayed in a subprocess.

The inverter at `ci/jobs/functional_tests.py:751-767` walks `test_result.results` to flip `FAIL` to `OK`. When the run ended in `Result.Status.ERROR`, that list was empty or partial, the loop did nothing, `has_failure` stayed `False`, and the `else` branch unconditionally called `set_failed().set_info("Failed to reproduce the bug")` — silently overwriting the honest `ERROR` set by `functional_tests_results.py:run()`.

The fix extracts the inverter to a module-level helper `invert_bugfix_validation_status` and short-circuits when `test_result.status == Result.Status.ERROR`: preserve the original status and `info`, only attach the `XFAIL` label to leaves so the report renders consistently. `FAIL` and `OK` paths continue to invert as before.

The complementary `Server died` synthesis for `runner_exit_code in {STOP_TESTING_EXIT_CODE, 137, 143}` is handled in @leshikus's PR #105643. With that PR, signal-killed runs surface a `FAIL` leaf that this inverter then correctly maps to "bug reproduced". This change covers every other path that leaves `status == ERROR`.

### Tests

`ci/tests/test_bugfix_validation_inverter.py` adds 8 unit tests covering:

- single `FAIL` flipped to `OK` (bug reproduced)
- single `OK` flipped to `FAIL` (bug not reproduced)
- mixed `FAIL` + `OK`
- synthetic `Server died` `FAIL` leaf (#105643 flow)
- `ERROR` + empty results: preserved (the #105789 regression)
- `ERROR` + partial `OK` leaf: preserved, leaf not flipped, `XFAIL` label still applied
- `XFAIL` label applied to every leaf
- `OK` outer + empty results: still reports "Failed to reproduce the bug"

The two `ERROR`-preservation tests were also run against an inlined copy of the pre-fix code path to confirm both directions: the old behaviour overwrites `ERROR` with `FAIL` "Failed to reproduce the bug", the new behaviour preserves `ERROR`.

...

### Version info
- Merged into: `26.6.1.197`

### DIFF
--- a/ci/jobs/functional_tests.py
+++ b/ci/jobs/functional_tests.py
@@ -137,6 +137,45 @@ OPTIONS_TO_TEST_RUNNER_ARGUMENTS = {
 }
 
 
+def invert_bugfix_validation_status(test_result: Result) -> None:
+    """Invert FAIL/OK in `test_result.results` for bugfix validation.
+
+    On master HEAD a regression test for the bug is expected to FAIL; the
+    inverter flips that to OK so the job reads as "bug reproduced". A clean
+    OK means the test does not catch the bug and is flipped to FAIL with
+    "Failed to reproduce the bug".
+
+    When the run ended in `Result.Status.ERROR` (runner did not finish,
+    e.g. server crash without proper exit code, Python exception,
+    infrastructure outage) the per-test list is empty or partial and the
+    pre-inversion `ERROR` already tells the truth. Preserve it instead of
+    overwriting with "Failed to reproduce the bug". See #105789.
+    """
+    if test_result.status == Result.Status.ERROR:
+        for r in test_result.results:
+            r.set_label(Result.Label.XFAIL)
+        print(
+            "Bugfix validation inconclusive: the test runner did not "
+            "finish; preserving ERROR rather than reporting "
+            "'Failed to reproduce the bug'."
+        )
+        return
+
+    has_failure = False
+    for r in test_result.results:
+        r.set_label(Result.Label.XFAIL)
+        if r.status == Result.Status.FAIL:
+            r.status = Result.Status.OK
+            has_failure = True
+        elif r.status == Result.Status.OK:
+            r.status = Result.Status.FAIL
+    if not has_failure:
+        print("Failed to reproduce the bug")
+        test_result.set_failed().set_info("Failed to reproduce the bug")
+    else:
+        test_result.set_success()
+
+
 def main():
     args = parse_args()
     test_options = [to.strip() for to in args.options.split(",")]
@@ -749,22 +788,7 @@ def main():
 
     # invert result status for bugfix validation
     if is_bugfix_validation and test_result and (Labels.PR_BUGFIX in info.pr_labels or Labels.PR_CRITICAL_BUGFIX in info.pr_labels):
-        has_failure = False
-        for r in test_result.results:
-            r.set_label(Result.Label.XFAIL)
-            if r.status == Result.Status.FAIL:
-                r.status = Result.Status.OK
-                has_failure = True
-            elif r.status == Result.Status.OK:
-                r.status = Result.Status.FAIL
-        if not has_failure:
-            print("Failed to reproduce the bug")
-            test_result.set_failed().set_info("Failed to reproduce the bug")
-        else:
-            # For bugfix validation, the expected behavior is:
-            # - At least one test must fail (bug reproduced)
-            # - The overall Tests result is treated as success in that case
-            test_result.set_success()
+        invert_bugfix_validation_status(test_result)
 
     if JobStages.COLLECT_LOGS in stages:
         print("Collect logs")
--- a/ci/tests/test_bugfix_validation_inverter.py
+++ b/ci/tests/test_bugfix_validation_inverter.py
@@ -1,0 +1,174 @@
+"""
+Tests for `ci.jobs.functional_tests.invert_bugfix_validation_status`.
+
+The bugfix-validation inverter flips per-test `FAIL`/`OK` so that a regression
+test for a bug, which is expected to `FAIL` on master HEAD, is reported as
+"bug reproduced". When the run itself failed catastrophically (status
+`ERROR`, e.g. runner killed mid-flight or server crashed without a
+synthetic `Server died` leaf reaching `test_result.results`), the inverter
+must preserve `ERROR` rather than overwrite it with the misleading
+"Failed to reproduce the bug" `FAIL`.
+
+See ClickHouse/ClickHouse#105789.
+"""
+
+import os
+import sys
+
+import pytest
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../.."))
+
+from ci.jobs.functional_tests import invert_bugfix_validation_status
+from ci.praktika.result import Result
+
+
+def _make_leaf(name, status, info=""):
+    return Result(name=name, status=status, info=info)
+
+
+def _make_outer(status, results=None, info=""):
+    return Result(
+        name="Tests",
+        status=status,
+        results=results or [],
+        info=info,
+    )
+
+
+def test_single_fail_is_flipped_to_ok_and_outer_becomes_success():
+    """Regression test FAILed on master -> bug reproduced -> overall OK."""
+    leaf = _make_leaf("01234_regression_test", Result.Status.FAIL,
+                      info="server died with SIGSEGV")
+    outer = _make_outer(Result.Status.FAIL, [leaf])
+
+    invert_bugfix_validation_status(outer)
+
+    assert leaf.status == Result.Status.OK
+    assert outer.status == Result.Status.OK
+
+
+def test_single_ok_is_flipped_to_fail_and_outer_becomes_failed_to_reproduce():
+    """Regression test PASSed on master -> bug not reproduced -> overall FAIL."""
+    leaf = _make_leaf("01234_regression_test", Result.Status.OK)
+    outer = _make_outer(Result.Status.OK, [leaf])
+
+    invert_bugfix_validation_status(outer)
+
+    assert leaf.status == Result.Status.FAIL
+    assert outer.status == Result.Status.FAIL
+    assert "Failed to reproduce the bug" in outer.info
+
+
+def test_mixed_fail_and_ok_treats_any_fail_as_bug_reproduced():
+    """At least one FAIL on master is enough to declare the bug reproduced."""
+    leaf_fail = _make_leaf("01234_regression_test", Result.Status.FAIL)
+    leaf_ok = _make_leaf("99999_unrelated_test", Result.Status.OK)
+    outer = _make_outer(Result.Status.FAIL, [leaf_fail, leaf_ok])
+
+    invert_bugfix_validation_status(outer)
+
+    assert leaf_fail.status == Result.Status.OK
+    assert leaf_ok.status == Result.Status.FAIL
+    assert outer.status == Result.Status.OK
+
+
+def test_server_died_synthetic_fail_leaf_treated_as_bug_reproduced():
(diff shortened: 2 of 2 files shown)


groeneai:
### Pre-PR validation gate (per worker contract)

session: `cron:clickhouse-ci-task-worker:20260525-173000`

| # | Question | Answer |
| --- | --- | --- |
| a | **Deterministic repro?** Can you trigger the failure on demand? | Yes. `python3 -m pytest ci/tests/test_bugfix_validation_inverter.py::test_error_status_with_empty_results_preserves_error` fails against an inlined copy of the pre-fix inverter (outer status flips from `ERROR` to `FAIL`, info gets the trailing "Failed to reproduce the bug" line). Passes against the new helper. |
| b | **Root cause explained?** Causal chain. | Inverter loop iterates `test_result.results`. When the runner ends in `Result.Status.ERROR` (`functional_tests_results.py:run()` `elif not s.success_finish` branch), the per-test list is empty or partial, the loop runs zero times, `has_failure` stays `False`, the `else` branch unconditionally calls `set_failed().set_info("Failed to reproduce the bug")`. Pre-existing `ERROR` and its `info` are silently overwritten. |
| c | **Fix matches root cause?** Direct vs band-aid. | Direct: short-circuit when `test_result.status == Result.Status.ERROR` so the inverter does not run on a state where it has nothing honest to invert. Not a heuristic, not a widening of bounds, no settings pinned. |
| d | **Test intent preserved? New tests added?** | Yes. Existing inverter semantics for `FAIL`/`OK`/mixed are unchanged (covered by 4 new tests). 4 more tests cover the regression (empty `ERROR`, partial `ERROR`, `XFAIL` labelling invariant, empty `OK` fallback). |
| e | **Both directions demonstrated?** Test FAILS without fix, PASSES with fix. | Yes. Ran the two ERROR-preservation tests against an inlined copy of the pre-fix inverter via a standalone Python script: both fail with the old behaviour (`outer.status='FAIL'` with `Failed to reproduce the bug` appended), both pass with the new helper. Output captured in the PR description. |
| f | **Fix is general, not a narrow patch?** | Yes. Covers every path that leaves `test_result.status == ERROR` (Python exception in runner, job-level timeout with non-SIGTERM signal, infrastructure outage, slow Layer-A propagation). Complementary to @leshikus's PR #105643 for kill-code recognition (Layer B): with that PR merged, exit-143/137 paths surface a synthetic `Server died` `FAIL` leaf which the inverter then correctly flips to `OK`; this PR catches every other path. |

groeneai:
cc @maxknv @leshikus — could you review this? Companion to #105643. While #105643 expands the parser's recognized exit codes so SIGTERM/SIGKILL-killed runs get a synthetic `Server died` FAIL leaf, this PR fixes the bugfix-validation inverter so the `Result.Status.ERROR` fallback (Python exceptions, non-signal job timeouts, infra outages) is preserved instead of being overwritten with "Failed to reproduce the bug". Closes #105789.

maxknv:
If the server crashes during the test run, the test job status must not be ERROR; it should be FAIL. And the virtual test case "Server died" must also appear in the job results.  If it was not so, it's a job script or test runner bug. Which apparently was already fixed. cc @leshikus

The job status ERROR should be set only on unexpected job failures and should not be treated as successful bug reproduction in Bugfix Validation job.

leshikus:
My concern is that adding a page of code for this logic is overkill; also the fix breaks the logic - we invert the test status when we need to invert test status parsing logic; the test is either fail or ok regardless of the job it runs in, this is easier to understand and more conventional

groeneai:
@leshikus checking before refactoring, since the same leaf-mutation pattern exists in `functional_tests.py`, `clickhouse_light.py`, and `integration_test_job.py`.

Two readings of "the test is either fail or ok ... invert test status parsing logic":

**(A)** Leaves keep literal status, only outer flips, `Label.XFAIL` carries context. Downside: `json.html` colors by status only (`json.html:732-733`), so a FAIL leaf inside a green outer would render red. The label is metadata-only.

**(B)** Map leaves to the praktika statuses designed for this case: `Status.XFAIL` ("expected failure: test failed as expected, not a problem", `result.py:66`) and `Status.XPASS` ("unexpected pass"). Leaf truth retained ("it failed, but expectedly"), `is_ok()`/`is_failure()` already treat them correctly (`result.py:234-245`), dashboard already renders `xfail` gray and `xpass` red. As a bonus, XPASS makes a real signal visible that the current `OK->FAIL` flip hides: a bugfix PR whose regression test passes on master is broken.

```python
def invert_bugfix_validation_status(test_result):
    if test_result.status == Result.Status.ERROR:
        for r in test_result.results:
            r.set_label(Result.Label.XFAIL)
        return  # #105789
    has_failure = False
    for r in test_result.results:
        r.set_label(Result.Label.XFAIL)
        if r.status == Result.Status.FAIL:
            r.status = Result.Status.XFAIL
            has_failure = True
        elif r.status == Result.Status.OK:
            r.status = Result.Status.XPASS
    if has_failure:
        test_result.set_success()
    else:
        test_result.set_failed().set_info("Failed to reproduce the bug")
```

I'd lean to (B). Smallest diff, leaves are no longer rewritten to lie about themselves, dashboard already understands the values. Happy to apply the same shape to the other two inverters in this PR for consistency, or split as a follow-up. If you meant something else (e.g. plumb `is_bugfix_validation` into `FTResultsProcessor` and drop the inverter entirely), I will go that route instead. Bigger PR but cleaner separation.

Re "page of code": with (B) the tests collapse to 5 (FAIL->XFAIL, OK->XPASS, mixed, ERROR preserved, all-OK reports "failed to reproduce"). Will shrink the test file accordingly.

Which do you want?

maxknv:
> My concern is that adding a page of code for this logic is overkill; also the fix breaks the logic - we invert the test status when we need to invert test status parsing logic; the test is either fail or ok regardless of the job it runs in, this is easier to understand and more conventional

I double-checked the code and realized that this is a correct fix. It basically moves inversion logic into a new method and just adds new handling for the ERROR status, keeping it as ERROR without inversion, which is the right thing to do. An unexpected job error must not be treated as successful bug fix validation.

maxknv (approved):

PR #105799, opened 2026-05-25
Substitute DDL parameters in parameterized view CREATE statements

groeneai:
`ASTCreateQuery::isParameterizedView` returns true whenever the SELECT body has
any query parameter, which made `executeQueryImpl` skip
`ReplaceQueryParameterVisitor` entirely. A `{db:Identifier}` placeholder in the
database name was therefore never substituted, leaving `create.getDatabase()`
empty and aborting the server in debug builds (`assert(!database_name.empty())`
in `DatabaseCatalog::tryGetDatabase`). In release builds the empty database
name silently produced an `UNKNOWN_DATABASE` error late in the CREATE flow.

The visitor now handles parameterized views directly: it substitutes parameters
in the DDL parts of the create query (database name, table name, columns list,
storage, view targets) while keeping the SELECT body's placeholders intact, so
they remain available for substitution at view-call time.

Reproducer (asserts on master, succeeds after the fix):

```sql
SET param_db = 'mydb';
CREATE VIEW {db:Identifier}.pv AS SELECT k + {n:UInt64} AS x FROM src;
```

Discovered by the AST fuzzer (STID 3282-3691). Master CI report:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=ffc21883c7b374f0aed9c9bed5924341b680833e&name_0=MasterCI&name_1=AST%20fuzzer%20%28amd_debug%29

Regression test: `04276_create_view_param_db_with_param_select`.

Fixed an assertion in `DatabaseCatalog::tryGetDatabase` (and resulting `UNKNOWN_DATABASE` error in release builds) when creating a parameterized view whose database name is supplied via a query parameter, for example `CREATE VIEW {db:Identifier}.v AS SELECT {p:UInt64}`. Parameters in the DDL parts of the `CREATE` statement are now substituted at create time, while parameters in the SELECT body remain available for substitution at view-call time.

### Version info
- Merged into: `26.6.1.149`

### DIFF
--- a/src/Interpreters/ReplaceQueryParameterVisitor.cpp
+++ b/src/Interpreters/ReplaceQueryParameterVisitor.cpp
@@ -11,6 +11,7 @@
 #include <Parsers/ASTIdentifier.h>
 #include <Parsers/ASTLiteral.h>
 #include <Parsers/ASTQueryParameter.h>
+#include <Parsers/ASTSelectWithUnionQuery.h>
 #include <Parsers/ASTViewTargets.h>
 #include <Parsers/Access/ASTCreateUserQuery.h>
 #include <Parsers/Access/ASTUserNameWithHost.h>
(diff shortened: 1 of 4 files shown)


groeneai:
**Pre-PR validation gate (per TASK.md):**

a) **Deterministic repro?** Yes. The reproducer above aborts on master 100% of the time and succeeds with this fix.

b) **Root cause explained?** `ASTCreateQuery::isParameterizedView` returns true when the SELECT body contains any `{param:Type}` placeholder. `executeQueryImpl` was hard-skipping `ReplaceQueryParameterVisitor` in that case, which left the `{db:Identifier}` placeholder for the database name unsubstituted. `create.getDatabase()` then returned `""` because `create.database` is an `ASTIdentifier` with an empty `name_part` and a `QueryParameter` child. The assertion `!database_name.empty()` in `DatabaseCatalog::tryGetDatabase` fires.

c) **Fix matches root cause?** The visitor now runs on every create query but for parameterized views it walks all immediate children except `create.select`, so DDL identifiers are substituted at create time while the SELECT body's placeholders remain available for view-call-time substitution. No assertion bounds were widened, no settings were pinned, no symptom was masked.

d) **Test intent preserved?** All 13 existing non-zookeeper parameterized view tests pass locally:

```
02428_parameterized_view_param_in_select_section: OK
02818_parameterized_view_with_cte_multiple_usage: OK
03142_alter_comment_parameterized_view: OK
03209_parameterized_view_with_non_literal_params: OK
03321_system_tables_parametrized_view_params: OK
03322_view_over_parameterized_view: OK
03573_create_query_parameters_in_to_in_materialized_view: OK
03779_materialized_view_database_context: OK
04064_parameterized_view_scalar_subquery: OK
04098_remote_parameterized_view_no_crash: OK
04227_prewhere_parameterized_view_join_102074: OK
04241_parameterized_view_used_twice: OK
04276_create_view_param_db_with_param_select: OK (new)
```

e) **Both directions demonstrated?** Yes:
- Without the fix: `clickhouse-client --param_X=db -q "CREATE VIEW {X:Identifier}.v AS SELECT k + {n:UInt64} AS x FROM src"` aborts (assertion stack trace matches STID 3282-3691).
- With the fix: same query succeeds; the new regression test exercises it end to end including the `FINAL` variant from the original fuzzer query (`test_view_01051_d__fuzz_13`).

f) **Fix is general, not a narrow patch?** The branch in `ReplaceQueryParameterVisitor::visit` handles any parameterized view, not just the specific fuzzer reproducer. It substitutes every DDL part of the create query (`create.database`, `create.table`, `create.columns_list`, `create.storage`, `create.targets`, `create.refresh_strategy`, `create.aliases_list`, `create.sql_security`, `create.comment`, `create.dictionary_attributes_list`, `create.dictionary`) while skipping only `create.select`. The "trust boundary expansion" question doesn't apply here because we are not widening any internal callee for new callers.

Session: cron:clickhouse-ci-task-worker:20260525-201500

groeneai:
cc @evillique @KochetovNicolai could you review this? It fixes an assertion (STID 3282-3691) where parameter substitution was skipped for the entire CREATE statement when the SELECT body of a parameterized view contained a query parameter. `ReplaceQueryParameterVisitor` now substitutes parameters in the DDL parts of the create query (database, table, columns, storage, targets) while preserving the SELECT body's placeholders.

tuanpach (approved):

PR #105820, opened 2026-05-26
Fix UBSan signed-int overflow in flameGraph dealloc for size = INT64_MIN

groeneai:
`flameGraph(trace, size, ptr)` aborted the server in serverfuzz-enabled UBSan builds when the fuzzer fed a deallocation row with `size = INT64_MIN` (`-9223372036854775808`).

`AggregateFunctionFlameGraph.cpp:418` computed `UInt64 abs_size = -size;`, which is undefined behaviour for `size == INT64_MIN` because the magnitude `2^63` is not representable in `Int64` (UBSan: "negation of -9223372036854775808 cannot be represented in type 'Int64'").

Fix: cast `size` to `UInt64` before negating, so the operation runs in unsigned arithmetic where `-static_cast<UInt64>(INT64_MIN) == 2^63`. This is the canonical fix suggested by UBSan's own diagnostic ("cast to an unsigned type to negate this value to itself") and mirrors the existing pattern on the allocation path (`allocation->size = UInt64(size);` at line 407).

Regression test `04276_stid_3765_flamegraph_int64_min_dealloc.sql` exercises the `INT64_MIN` dealloc path. Verified locally on the ASan+UBSan build that the original reproducer aborted the server before the fix and runs cleanly after; the existing `04113_flame_graph_string_traces` and `04114_aggregate_function_flamegraph` tests still pass.

Report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=544f6da6dc6cc737e7f8fc8e3bcf7e7d5657f099&name_0=MasterCI&name_1=Stress%20test%20%28experimental%2C%20serverfuzz%2C%20arm_asan_ubsan%29
STID: 3765-3f7f

Fix a signed-integer overflow in the `flameGraph` aggregate function when the `size` argument equalled `INT64_MIN` (`-9223372036854775808`). On UBSan-enabled builds the violation aborted the server; in release builds the negation wrapped, producing a nonsensical absolute size. The dealloc path now negates the size in unsigned arithmetic.


groeneai:
### Pre-PR validation gate

```sql
SELECT arrayJoin(flameGraph(trace, sz, ptr)) FROM (
    SELECT CAST([18446744073709551600, 18446744073709551601] AS Array(UInt64)) AS trace,
           CAST(-9223372036854775808 AS Int64) AS sz,
           CAST(42 AS UInt64) AS ptr
);
```
On a local ASan+UBSan build, this aborts the server on every run with:
`SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /…/AggregateFunctionFlameGraph.cpp:418:31`

**b) Root cause explained?**
At line 418 the dealloc branch executes `UInt64 abs_size = -size;`. When `size == INT64_MIN` (`-2^63`), `-size` is signed-integer negation whose mathematical result `2^63` is not representable in `Int64` (whose max is `2^63 - 1`). In UBSan-enabled builds this triggers `halt_on_error` and aborts the server; in release builds the negation wraps back to `INT64_MIN`, so `abs_size` becomes `0x8000000000000000` reinterpreted as unsigned which happens to be the correct value, but only by undefined-behaviour accident.

The fix changes the negation domain to unsigned: `UInt64 abs_size = -static_cast<UInt64>(size);`. Unsigned arithmetic wraps with well-defined two's complement semantics, so `-static_cast<UInt64>(INT64_MIN) == 2^63`, which is the intended absolute value. This is exactly what UBSan's diagnostic suggests ("cast to an unsigned type to negate this value to itself") and matches the sibling positive-size path at line 407 (`allocation->size = UInt64(size);`).

**d) Test intent preserved / new tests added?**
- Existing `04113_flame_graph_string_traces` and `04114_aggregate_function_flamegraph` still pass against the patched binary.
- New regression test `04276_stid_3765_flamegraph_int64_min_dealloc.sql` exercises the `INT64_MIN` dealloc path. The fix does not weaken any assertion; it only widens defined behaviour to cover the full `Int64` domain.

- Before fix: same reproducer produces the UBSan abort and kills the server (`ATTEMPT_TO_READ_AFTER_EOF` on the follow-up query).
- After fix: same reproducer returns an empty result (unmatched dealloc, expected semantic), server stays alive, no UBSan output, follow-up queries succeed.

- I grepped the file for negation/abs patterns: line 418 is the only site. The positive-size branches at lines 388 and 407 already use the canonical `UInt64(size)` cast.
- No symmetric variants of this aggregate exist with the same pattern.
- The bad value is a legitimate user-controlled `Int64` input (the function signature accepts `Int64 size` from a SQL column), not an upstream defect. The fix lives at the exact computation that produces the undefined result.

Session: `cron:clickhouse-ci-task-worker:20260526-041500`

groeneai:
cc @KochetovNicolai @nickitat, could you review this? Small UBSan fix in the `flameGraph` aggregate function dealloc path: cast to `UInt64` before negating so `size == INT64_MIN` no longer triggers signed-integer overflow. Reproducer in the regression test.

Algunenano:
https://github.com/ClickHouse/ClickHouse/pull/105829 is better. Covers more use cases and it's not so verbose for no reason

PR #105828, opened 2026-05-26
Stabilize `Keeper` reconfig test retries

antonio2368:
Syncs the non-private `Keeper` test stabilization from https://github.com/ClickHouse/clickhouse-private/pull/54962 for https://github.com/ClickHouse/ClickHouse/pull/101757.

This makes `test_keeper_reconfig_replace_leader_in_one_command` robust under repeated runs by resetting `Keeper` state, closing stale ZooKeeper connections, and confirming idempotent retried `create` operations after `NodeExistsError`.

No matching open issue was found for `test_keeper_reconfig_replace_leader_in_one_command`.

Not required.

### Documentation entry:
- [x] Documentation is not required because this only changes an integration test.

### Version info
- Merged into: `26.6.1.162`

### DIFF
--- a/tests/integration/test_keeper_reconfig_replace_leader_in_one_command/test.py
+++ b/tests/integration/test_keeper_reconfig_replace_leader_in_one_command/test.py
@@ -1,10 +1,11 @@
 #!/usr/bin/env python3
 
 import time
-import typing as tp
+from multiprocessing.dummy import Pool
 from os.path import dirname, join, realpath
 
 import pytest
+from kazoo.exceptions import NodeExistsError
 
 import helpers.keeper_utils as ku
 from helpers.cluster import ClickHouseCluster, ClickHouseInstance
@@ -12,9 +13,9 @@ from helpers.cluster import ClickHouseCluster, ClickHouseInstance
 cluster = ClickHouseCluster(__file__)
 CONFIG_DIR = join(dirname(realpath(__file__)), "configs")
 
-node1 = cluster.add_instance("node1", main_configs=["configs/keeper1.xml"])
-node2 = cluster.add_instance("node2", main_configs=["configs/keeper2.xml"])
-node3 = cluster.add_instance("node3", main_configs=["configs/keeper3.xml"])
+node1 = cluster.add_instance("node1", main_configs=["configs/keeper1.xml"], stay_alive=True)
+node2 = cluster.add_instance("node2", main_configs=["configs/keeper2.xml"], stay_alive=True)
+node3 = cluster.add_instance("node3", main_configs=["configs/keeper3.xml"], stay_alive=True)
 node4 = cluster.add_instance("node4", stay_alive=True)
 zk1, zk2, zk3, zk4 = None, None, None, None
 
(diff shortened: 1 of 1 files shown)


SmitaRKulkarni (approved):
LGTM

groeneai:
Approach matches the investigation in the closed #105707. CIDB confirms 12 of 13 recent failures in the past 14 days land on this exact line `test.py:131` (`zk4.create(f"/test_four_{100 + i}", b"somedata")`), all with the same kazoo `NodeExistsError` traceback through `KazooClientWithImplicitRetries.create`. The targeted `create_or_confirm` here resolves that race.

The value-equality check (`assert zk.get(path)[0] == data`) on `NodeExistsError` correctly handles the lost-response retry case: the server committed the create, the response was dropped during a `ConnectionLoss`, kazoo retries and observes the now-existing node with the same value. Genuine collisions (different value) still surface as an assertion failure.

`reset_keepers()` + `stay_alive=True` on node1/2/3 is a useful improvement on top of #105707: with the coordination state wipe between runs, the test is now safe under `--count N` repeated runs. Targeting only the second `zk4.create` loop (and not the first `zk1.create` loop) is correct, since the race only manifests after the leader change.

LGTM.

groeneai:
@antonio2368 could you apply the `v26.5-must-backport` label to this PR? It is missing from 26.5 and the flake it fixed is still firing there.

The `NodeExistsError` at `test.py:131` in `test_keeper_reconfig_replace_leader_in_one_command` is now confined to 26.5. Over the last 120 days, on `pull_request_number = 0`, 26.5 has 11 FAIL against 696 OK while master, 26.7, 26.6, 26.4, 26.3 and 25.8 each have 0 FAIL. All 11 of the 26.5 rows carry `NodeExistsError` and all 11 carry `test.py:131`, the unguarded `zk4.create(f"/test_four_{100 + i}", b"somedata")`. Latest hit 2026-08-04 23:44:26, so about 1.6 percent of 26.5 runs.

Five more of the same signature landed on robot backport PRs targeting 26.5 (the backport/26.5 branches for 106802, 108477, 110233, 111039 and 111806), so it also taxes unrelated 26.5 backports.

This PR is the fix and it works. Counting FAIL rows for this test on master lineage since this merged at 2026-05-26 16:28:17, there are 8, and none of them is this signature: they are at `test.py:171`, `:130` and `:39`. The last `NodeExistsError` anywhere on master lineage is 2026-05-26 15:47:06, 41 minutes before the merge, against 87801 OK runs of this test on master lineage in the same window. 26.5 lacks the change: `git merge-base --is-ancestor a88db8eda388684e2ca670c12ee05f89046a3133 26.5` is false, and `create_or_confirm` occurs 0 times in the 26.5 copy of the test versus 2 times on master, 26.7 and 26.6.

`v26.5-must-backport` rather than `pr-must-backport`: the version label expands to the named release plus every newer active branch, so it targets 26.7, 26.6 and 26.5, and the first two already contain the merge commit, which leaves 26.5 as the only real target. The general label would also target 26.3 and 25.8, where this signature has never fired (0 FAIL against 987 and 423 OK) and where 25.8 does not run the `db disk` flavor that produced 10 of the 11 hits.

The cherry-pick is clean, so the robot should not need anyone to resolve anything: `git merge-tree --write-tree --merge-base=a88db8eda388684e2ca670c12ee05f89046a3133^1 26.5 a88db8eda388684e2ca670c12ee05f89046a3133` succeeds and the resulting tree differs from the current 26.5 head in that one test file only, +60/-10.

PR #105839, opened 2026-05-26
Fix STID 1941-1bfa: reject query parameter substitution as function name

groeneai:
The AST fuzzer keeps rediscovering the `Inconsistent AST formatting: the original AST: (STID: 1941-1bfa)` round-trip failure across unrelated PRs (see #104605). One subvariant: a query parameter substitution placed in a function-name position, e.g.

```sql
SELECT {FUNC:Identifier}(-1)
```

with `--param_FUNC=abs` aborts the server in debug builds.

Root cause: `ParserCompoundIdentifier` accepts `{p:Identifier}.tbl.fn` as a 3-part identifier with a parameter hole in the first slot. `getFunctionLayer` then calls `getIdentifierName(identifier)`, which returns the unset `ASTIdentifier::full_name` (empty when `name_parts` contains parameter holes). The result is an `ASTFunction` with `name=""`. `ASTFunction::formatImplWithoutAlias` then skips the name and emits just `(args)`, so re-parsing yields a plain parenthesized expression. The round-trip tree-hash check in `executeQueryImpl` raises the `STID 1941-1bfa` exception.

The fix rejects such an identifier as a function name at parse time in `getFunctionLayer` with a clear `SYNTAX_ERROR`. Parameter substitutions in column, table, and database positions are unaffected.

Pre-fix on the AST fuzzer reproducer:

```
Logical error: 'Inconsistent AST formatting: the original AST:
... (STID: 1941-1bfa)
```

Post-fix:

```
Code: 62. DB::Exception: Function name cannot be empty or a query parameter substitution. (SYNTAX_ERROR)
```

The new regression test `04277_stid_1941_1bfa_param_subst_function_name.sh` covers:
- the minimal single-part shape `{FUNC:Identifier}(-1)`
- the compound shape `{DB:Identifier}.test_table.COLUMNS(if(1, 1, id))` that the fuzzer found on #100412
- the `DESCRIBE TABLE (...)` wrapper used in the fuzzer output

It also checks legitimate parameter usages (column / table / database names) still work and `COLUMNS` matcher unaffected.

Related: requested by @alexey-milovidov on #100412 (2026-05-26T07:31:35Z, `STID 1941-1bfa` directive). Issue #104605 tracks the umbrella.

Pre-PR validation gate (the five questions) posted in a follow-up comment below.

Fix a server-aborting `LOGICAL_ERROR` `Inconsistent AST formatting: the original AST: (STID: 1941-1bfa)` raised by the AST round-trip check when a query parameter substitution was used in a function-name position, such as `{F:Identifier}(args)` or `{db:Identifier}.tbl.fn(args)`. Such usages now fail at parse time with a clear `SYNTAX_ERROR`.

Documentation entry for user-facing changes
- [x] Documentation is unchanged


groeneai:
**Pre-PR validation gate** (per TASK.md Phase 4 Step 7):

a) **Deterministic repro?** Yes. The minimal repro is `SELECT {FUNC:Identifier}(-1)` with `--param_FUNC=abs`. Pre-fix: server aborts with `Logical error: 'Inconsistent AST formatting: the original AST: ... (STID: 1941-1bfa)'` (signal 6) every time. Post-fix: returns `Code: 62. ... Function name cannot be empty or a query parameter substitution. (SYNTAX_ERROR)`.

b) **Root cause explained?** Yes. `ParserCompoundIdentifier` with `allow_query_parameter=true` accepts `{p:Identifier}.tbl.fn` as a 3-part identifier with one parameter hole. Constructed via `ASTIdentifier(std::vector<String> && name_parts_, bool special, ASTs && name_params)` it stores the parameter into `children` but does NOT call `resetFullName()` (gated by `name_params.empty()`), so `full_name` is `""`. `ASTIdentifier::name()` returns this empty `full_name`. `getFunctionLayer` then builds a `FunctionLayer(function_name="")`, which emits an `ASTFunction(name="")`. `ASTFunction::formatImplWithoutAlias` at the `if (!name.empty()) ostr << backQuoteIfNeed(name);` site (the comment there: *"Empty names are used rarely, to format queries with an extra pair of parentheses for external databases"*) writes only `(args)`. The re-parser interprets `(args)` as a parenthesized expression via `RoundBracketsLayer`, producing a different AST. The tree-hash mismatch fires the `STID 1941-1bfa` exception in `executeQueryImpl`.

c) **Fix matches root cause?** Yes. The fix rejects the empty `function_name` directly in `getFunctionLayer` — at the exact site that decides whether to enter `FunctionLayer` mode. No special-casing in the formatter, no extra fields in `ASTFunction`, no parser rewrites elsewhere. 13 lines including comment.

d) **Test intent preserved?** No existing tests are modified. `01550_query_identifier_parameters` (parameterized column / table / database names) and `02014_query_parameters` (general `{p:Type}` substitution) still pass against the fix; legitimate parameter positions are not affected. A new regression test `04277_stid_1941_1bfa_param_subst_function_name.sh` is added that asserts both the rejection (`SYNTAX_ERROR`) and that the legitimate uses still work.

- **WITHOUT** the fix: built locally; running `tests/clickhouse-test 04277` returned `Having 1 errors! 0 tests passed.` (server crashed in the middle of the test, exit 134).
- **WITH** the fix: `tests/clickhouse-test --test-runs 20 -- 04277` returns `20 tests passed. 0 tests skipped.` with full settings randomization on.

f) **Fix is general?** Yes. The check is placed once in `getFunctionLayer` and is reached from all three `function_name_parser`-feeding sites: `ParserFunction::parseImpl`, the table-function `ViewLayer` path, and the inner `OPERAND` state-machine site. No code path that creates an `ASTFunction` from a `ParserFunctionName`-parsed identifier bypasses it. The "trace upstream" question — *why does an identifier with empty `name()` reach a function-call position?* — is answered: `ASTIdentifier` legitimately leaves `full_name` empty when the identifier contains a parameter hole; the bug is that the **caller** then used that empty string as a function name without checking. The fix is at the call boundary, not on a defensive null check.

groeneai:
cc @evillique @george-larionov - could you review this? It is a parser-side fix for the chronic STID 1941-1bfa AST round-trip failure: rejects query parameter substitution as a function name (`{F:Identifier}(args)`) at parse time in `getFunctionLayer`, since such an identifier produces an `ASTFunction` with empty `name` that the formatter writes as just `(args)` and the round-trip check fails.

evillique:
@groeneai rebase on top of the fresh master

groeneai:
@evillique rebased on current master in 5de9b39b41a08bacbef87afe9e272fc094ac1834 (no conflicts). CI rerunning.

groeneai:
@evillique closing instead of rebasing. Master already fixes this exact case.

PR #106635 (merged 2026-06-07, abc40927d496) rejects a query-parameter substitution in a function-name position at the source, in `FunctionNameLayer::parseImpl` via `ASTIdentifier::isParam()`, and adds `chassert(!function_name.empty())` at the same `getFunctionLayer` line this PR edits. That chassert makes this PR's `throw` unreachable, which is exactly the throw-vs-chassert conflict on the rebase.

Their test `04319_query_parameter_as_function_name.sql` covers all three reproducers here, including the verbatim `DESCRIBE TABLE (SELECT {db:Identifier}.tbl.COLUMNS(...))` fuzzer case. No new master "Inconsistent AST formatting" failures for this subvariant since that merge.

Superseded by #106635, thanks @ vitlibar.

Issue #105830, opened 2026-05-26
Install packages job runs out of disk and fails tar extract


Algunenano:
The `Install packages (amd_release)` job intermittently fails when the runner's root filesystem fills up, surfacing as a `tar` short-write deep in the last sub-test.

Example failure on master: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=139d37962a73d4c6fb917fdd80211e6731d18095&name_0=MasterCI&name_1=Install%20packages%20%28amd_release%29&name_1=Install%20packages%20%28amd_release%29

```
+ tar xf /packages/clickhouse-keeper-26.6.1.139-amd64.tgz
tar: clickhouse-keeper-26.6.1.139/usr/bin/clickhouse-keeper: Wrote only 7680 of 10240 bytes
tar: Exiting with failure status due to previous errors
```

`df -h` from the same `job.log`: pre-job `12G` free on `/dev/root` (58G total, `t3.large`), post-job `8.4G` free. The job runs 12 install sub-tests sequentially on a single self-hosted runner (`STYLE_CHECK_AMD`) and the cumulative docker overlay / installed-package state on the host tips it over `ENOSPC` during the last extract.

### Possible approaches

- Grow the `STYLE_CHECK_AMD` AMI/launch-template root EBS volume from 58 GB to ~100 GB.
- Reclaim disk between sub-tests inside `ci/jobs/install_check.py` (`docker container prune -f` + `docker volume prune -f`) and log `df -h` per iteration.
- Add a pre-flight disk check at the start of `install_check.py` that fails fast with a clear message instead of letting `tar` produce a short-write later.
- Add the existing `docker_clean_up_hook.py` `post_hooks` to `install_check_master_jobs` in `ci/defs/job_configs.py` (currently only set on the PR variant).
- Split the job into per-format shards (deb / rpm / tgz-deb / tgz-rpm) so each runs on a fresh runner.

cc @leshikus @maxknv

Algunenano:
Should be fine after https://github.com/ClickHouse/ClickHouse/pull/105841

PR #105841, opened 2026-05-26
Add docker cleanup post-hook to install_check master jobs

leshikus:
The `install_check_master_jobs` config was missing the `docker_clean_up_hook.py` `post_hooks` entry that is already set on the PR variant `install_check_jobs`. This restores parity so the `STYLE_CHECK_AMD` runner is cleaned up between successive master runs of the `Install packages (amd_release)` job.

The post-hook runs at the end of the job, so it improves the runner state for the *next* job — it does not fix an ongoing `ENOSPC` failure on the last iteration of an in-flight run. If the underlying disk pressure recurs, further mitigations (pre-hook cleanup, larger EBS volume, or per-format shards) will be needed.

Related to #105830

...

### DIFF
--- a/ci/defs/job_configs.py
+++ b/ci/defs/job_configs.py
@@ -489,6 +489,7 @@ class JobConfigs:
             ],
         ),
         timeout=900,
+        post_hooks=["python3 ./ci/jobs/scripts/job_hooks/docker_clean_up_hook.py"],
     ).parametrize(
         Job.ParamSet(
             parameter="amd_release",
--- a/ci/jobs/scripts/job_hooks/docker_clean_up_hook.py
+++ b/ci/jobs/scripts/job_hooks/docker_clean_up_hook.py
@@ -2,27 +2,7 @@ from ci.praktika.utils import Shell
 
 
 def check():
-    print("Remove all images with names starting with clickhouse/clickhouse- (forced)")
-    Shell.check(
-        "docker images --format '{{.Repository}}:{{.Tag}}' | grep '^clickhouse/clickhouse-' | xargs -r docker rmi -f",
-        verbose=True,
-    )
-    print("Clean up non-latest images per each Repository")
-    Shell.check(
-        "docker images --format '{{.Repository}} {{.ID}} {{.CreatedAt}}' "
-        " | sort -k1,1 -k3,3M -k4,4n -k5,5n "
-        " | tac "
-        " | awk '!seen[$1]++ {next} {print $2}' "
-        " | xargs -r docker rmi",
-        verbose=True,
-    )
-    print("Clean up build cache")
-    Shell.check("docker builder prune -a -f", verbose=True)
-    print("Clean up stopped containers")
-    Shell.check("docker container prune -f", verbose=True)
-    Shell.check("docker system prune", verbose=True)
-    print("Clean up all tagged-but-unused images")
-    Shell.check("docker image prune -a -f", verbose=True)
+    Shell.check("docker system prune -a -f", verbose=True)
     return True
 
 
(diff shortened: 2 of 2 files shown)


Algunenano:
The hook helps but it does not actually clean the images this job uses.

`install_check.py` uses `clickhouse/install-rpm-test` and `clickhouse/install-deb-test`. Looking at `docker_clean_up_hook.py`:

1. The forced `docker rmi` only matches `^clickhouse/clickhouse-`, so it does not touch `clickhouse/install-*`.
2. The "non-latest per Repository" sweep keeps the current tag, so back-to-back master runs with the same digest leave the install-test images in place.
3. `docker container prune` is a no-op here since the job uses `docker run --rm`.
4. `docker system prune` without `-a` only removes dangling images.

So the install-test images (around 1 to 2 GB each) survive between runs. The PR is still a net positive because it sweeps `clickhouse/clickhouse-*` and dangling content left by other jobs on the same runner, but it does not reclaim the bulk that the install-packages job itself contributes.

To clean those too, the hook could either widen the grep to something like `^clickhouse/(clickhouse|install|stateless|stress|integration|binary|fasttest)-`, or just run `docker image prune -a -f`, which removes all tagged-but-unused images.

maxknv (approved):

maxknv:
> The hook helps but it does not actually clean the images this job uses.
> 
> `install_check.py` uses `clickhouse/install-rpm-test` and `clickhouse/install-deb-test`. Looking at `docker_clean_up_hook.py`:
> 
> 1. The forced `docker rmi` only matches `^clickhouse/clickhouse-`, so it does not touch `clickhouse/install-*`.
> 2. The "non-latest per Repository" sweep keeps the current tag, so back-to-back master runs with the same digest leave the install-test images in place.
> 3. `docker container prune` is a no-op here since the job uses `docker run --rm`.
> 4. `docker system prune` without `-a` only removes dangling images.
> 
> So the install-test images (around 1 to 2 GB each) survive between runs. The PR is still a net positive because it sweeps `clickhouse/clickhouse-*` and dangling content left by other jobs on the same runner, but it does not reclaim the bulk that the install-packages job itself contributes.
> 
> To clean those too, the hook could either widen the grep to something like `^clickhouse/(clickhouse|install|stateless|stress|integration|binary|fasttest)-`, or just run `docker image prune -a -f`, which removes all tagged-but-unused images.

I think it's fine to keep the images used for testing; they do not add up on each run, and should not take up a lot of space over time.

leshikus:
Thanks for the thorough review. Pushed a21133471b8 — collapsed the hook to a single `docker system prune -a -f`. Mapping your points to the new state:

| Point | Status |
|---|---|
| Forced `docker rmi` only matches `^clickhouse/clickhouse-`, misses `clickhouse/install-*` | `system prune -a -f` reclaims any unused image regardless of name |
| "Non-latest per Repository" sweep keeps current tag, so install-test images survive | `system prune -a` removes all unused, including `:latest` of an unused repo |
| `docker container prune` is a no-op (job uses `--rm`) | subsumed into `system prune` |
| `docker system prune` without `-a` only removes dangling images | now uses `-a` |
| Install-test images (1–2 GB each) survive between runs | reclaimed by `prune -a` |

The one thing this PR still doesn't fix — and never claimed to — is the *in-flight* `ENOSPC` on the failing iteration itself; the post-hook only helps the next job on the runner. That's called out in the description as future work (pre-hook cleanup, larger EBS, or per-format shards).

Trade-off of `prune -a`: base images like `ubuntu:24.04` get re-pulled on the next job that needs them, but that's seconds against multi-GB of disk reclaimed.

leshikus:
@Algunenano just a note, @maxknv suggested to start without changing the hook, add more cleaning if needed; I'll add `clickhouse/install-*` to cleanup

Algunenano (approved):

PR #105851, opened 2026-05-26
Keeper: reduce memory during received snapshot apply

### LINKED ISSUE #83260: Unload old storage in-memory when applying snapshot in Keeper
When Keeper applies snapshot it will deserialize the new storage and then replace the old storage with the new one.
Ref https://github.com/ClickHouse/ClickHouse/blob/master/src/Coordination/KeeperStateMachine.cpp#L659

This can be problematic as all nodes are kept in-memory.
Before deserializing, we can unload old storage and then load new snapshot.
Potential problem can occur if  new snapshot is invalid, so Keeper should terminate in that case because old storage is lost.

This behaviour should be toggled with a coordination setting as termination on invalid snapshot could be undesired under a cost of higher memory consumption.
Also, rocksdb storage stores nodes on disk, so the issue is less relevant.

antonio2368:
Reduce peak memory usage while applying received snapshots in ClickHouse Keeper with `KeeperMemoryStorage`.

The new path validates and pins the received snapshot, detaches the uncommitted tail, drops old in-memory storage, deserializes replacement storage, and replays the tail. This keeps `KeeperRocksStorage` on the existing compatibility path and closes https://github.com/ClickHouse/ClickHouse/issues/83260.

- Improvement

Reduce peak memory usage when applying received snapshots in ClickHouse Keeper with `KeeperMemoryStorage`.

### Documentation entry:

### Version info
- Merged into: `26.6.1.223`
- Backported to: `26.5.2.22`


src/Coordination/KeeperStorage.h
```diff
@@ -593,10 +597,42 @@ class KeeperStorage : public KeeperStorageBase
...
-    // Apply uncommitted state to another storage using only transactions
-    // with zxid > last_zxid
+    struct UncommittedStateForSnapshot
+    {
+        UncommittedStateForSnapshot();
+        ~UncommittedStateForSnapshot();
+        UncommittedStateForSnapshot(UncommittedStateForSnapshot &&) noexcept;
+        UncommittedStateForSnapshot & operator=(UncommittedStateForSnapshot &&) noexcept;
+        UncommittedStateForSnapshot(const UncommittedStateForSnapshot &) = delete;
+        UncommittedStateForSnapshot & operator=(const UncommittedStateForSnapshot &) = delete;
```
kssenii, line 607: `boost::noncopyable`?

src/Coordination/KeeperStateMachine.cpp
```diff
@@ -718,37 +721,158 @@ bool KeeperStateMachine<Storage>::apply_snapshot(nuraft::snapshot & s)
...
 
-    { /// deserialize and apply snapshot to storage
-        std::lock_guard lock(snapshots_lock);
-
-        if (s.get_last_log_idx() > latest_snapshot_meta->get_last_log_idx())
+    const auto validate_snapshot_to_apply = [&](const SnapshotMetadataPtr & current_snapshot_meta)
+    {
+        if (!current_snapshot_meta)
+        {
+            ProfileEvents::increment(ProfileEvents::KeeperSnapshotApplysFailed);
```
kssenii: I see this is an old profile event, but its name seems gramatically incorrect, was it meant as `Applys -> Applies`?

Also would it be better to update this metric from `SCOPE_EXIT` instead (under `is_snapshot_applied` flag which is set to `true` at the end of this method) to account for uncaught exceptions?

src/Coordination/KeeperStorage.h
```diff
@@ -593,10 +597,42 @@ class KeeperStorage : public KeeperStorageBase
...
+            int64_t log_idx = 0;
+        };
+
+        std::vector<Transaction> transactions;
+        std::list<Delta> deltas;
+
+        bool empty() const { return transactions.empty(); }
+    };
+
+    UncommittedStateForSnapshot copyUncommittedStateAfter(int64_t last_log_idx) const;
```
kssenii, line 623:
```suggestion
    /// Collect uncommitted transactions/deltas with log_idx > last_log_idx
    UncommittedStateForSnapshot copyUncommittedStateAfter(int64_t last_log_idx) const;
```
(If you think this is excessive - just ignore)

src/Coordination/KeeperStorage.h
```diff
@@ -593,10 +597,42 @@ class KeeperStorage : public KeeperStorageBase
...
+    UncommittedStateForSnapshot detachUncommittedStateAfter(int64_t last_log_idx);
```
kssenii, line 627:
```suggestion
    /// Like `copyUncommittedStateAfter`, but removes the returned transactions/deltas
    /// from this storage instead of copying them.
    UncommittedStateForSnapshot detachUncommittedStateAfter(int64_t last_log_idx);
```
(If you think this is excessive - just ignore)

src/Coordination/KeeperStateMachine.cpp
```diff
@@ -718,37 +721,158 @@ bool KeeperStateMachine<Storage>::apply_snapshot(nuraft::snapshot & s)
...
         }
-        if (s.get_last_log_idx() < latest_snapshot_meta->get_last_log_idx())
+        if (s.get_last_log_idx() < current_snapshot_meta->get_last_log_idx())
         {
             LOG_INFO(
                 log,
                 "A snapshot with a larger last log index ({}) was created, skipping applying this snapshot",
-                latest_snapshot_meta->get_last_log_idx());
-            return true;
+                current_snapshot_meta->get_last_log_idx());
```
kssenii: `s.get_last_log_idx()` could also be included in the log message?

src/Coordination/KeeperStateMachine.cpp
```diff
@@ -718,37 +721,158 @@ bool KeeperStateMachine<Storage>::apply_snapshot(nuraft::snapshot & s)
...
+            return false;
+        }
+
+        return true;
+    };
+
+    if constexpr (std::is_same_v<Storage, KeeperMemoryStorage>)
```
kssenii, line 763: May be add a small comment explaining the logic a bit? E.g. something like:
```
/// Three-phase apply to reduce peak memory:
///   1. Under snapshots_lock: validate metadata, pin the snapshot file.
///   2. Lock-free: read the file into memory and validate its version
///   3. Re-acquire snapshots_lock + take storage_lock exclusively:
///      detach the uncommitted tail, free the old storage, deserialize
///      the new one, replay the tail, swap in.
/// On any failure after step 3's `storage.reset()` we terminate — there
/// is no recoverable in-memory state to fall back to.
```

src/Coordination/KeeperStorage.cpp
```diff
@@ -1282,21 +1298,124 @@ void KeeperStorage<Container>::applyUncommittedState(KeeperStorage & other, int6
...
+        std::lock_guard lock(uncommitted_state.deltas_mutex);
+        /// After the first splice, this loop must not do throwing work. It
+        /// relies on single-element `std::list::splice` being `noexcept` and
+        /// integer set lookup not allocating.
+        for (auto it = uncommitted_state.deltas.begin(); it != uncommitted_state.deltas.end();)
+        {
+            auto current = it++;
+            if (zxids_to_apply.contains(current->zxid))
+                result.deltas.splice(result.deltas.end(), uncommitted_state.deltas, current);
+        }
```
kssenii: Would it make sense to put this part into separate `noexcept` method just for more explicit guarantees?

src/Coordination/KeeperStateMachine.cpp
```diff
@@ -718,37 +721,158 @@ bool KeeperStateMachine<Storage>::apply_snapshot(nuraft::snapshot & s)
...
+                }
+                catch (...)
+                {
+                    ProfileEvents::increment(ProfileEvents::KeeperSnapshotApplysFailed);
+                    LOG_FATAL(
+                        log,
+                        "Failed to apply snapshot {} after dropping old `KeeperMemoryStorage` "
+                        "(latest snapshot metadata index: {}): {}. Terminating to avoid inconsistent Keeper state",
+                        s.get_last_log_idx(),
+                        latest_snapshot_meta ? latest_snapshot_meta->get_last_log_idx() : 0,
```
kssenii:
```suggestion
                        latest_snapshot_meta ? toString(latest_snapshot_meta->get_last_log_idx()) : "(None)",
```

kssenii: Also inside the `try` we could have re-assigned `latest_snapshot_meta`, so it can mean different things depending on when the exception happened, so could make sense pre-save its value before the try if we want to see the previous version here in the log message.

src/Coordination/KeeperSnapshotManager.cpp
```diff
@@ -1052,6 +1063,32 @@ SnapshotDeserializationResult<Storage> KeeperSnapshotManager<Storage>::deseriali
     return result;
 }
 
+template<typename Storage>
+SnapshotMetadataPtr KeeperSnapshotManager<Storage>::deserializeSnapshotMetadataFromBuffer(nuraft::ptr<nuraft::buffer> buffer) const
+{
+    /// `nuraft::buffer::pos(0)` resets the cursor. This method must leave the
+    /// buffer at offset `0` on success and on throw.
```
kssenii, line 1070: Why on throw as well? As I see exception from `deserializeSnapshotMetadataFromBuffer` is not caught

src/Coordination/KeeperStorage.h
antonio2368: I would need to manually write move operators so it's not much better tbh

src/Coordination/KeeperStateMachine.cpp
antonio2368: Yes that typo is really old and it could break things if I just replace it. I can do it if you think it's worth it.

src/Coordination/KeeperSnapshotManager.cpp
antonio2368: Just defensive programming but your comment is valid.

src/Coordination/KeeperStorage.cpp
antonio2368: Yes, it would force termination on any exception which is what we want.

kssenii (approved):

src/Coordination/KeeperStateMachine.cpp
```diff
@@ -718,37 +724,168 @@ bool KeeperStateMachine<Storage>::apply_snapshot(nuraft::snapshot & s)
...
+                std::optional<uint64_t> latest_snapshot_meta_index_before_reset;
+                if (latest_snapshot_meta)
+                    latest_snapshot_meta_index_before_reset = latest_snapshot_meta->get_last_log_idx();
+
+                auto uncommitted_tail = storage->detachUncommittedStateAfter(s.get_last_log_idx());
+                storage.reset();
+
+                try
+                {
+                    auto snapshot_deserialization_result = snapshot_manager.deserializeSnapshotFromBuffer(snapshot_buf);
```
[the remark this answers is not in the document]
antonio2368: That's the behavior we want now.

PR #105856, opened 2026-05-26
Move git identity check to workflow pre-hook

Algunenano:
Replace the range with $local_sha --not --remotes: commits reachable from any remote-tracking ref are excluded, so anything brought in by git merge <remote>/<branch> is no longer scanned. Commits authored locally, either on the topic mainline or on a merged-in local side branch, remain in scope.

### Version info
- Merged into: `26.6.1.213`


.githooks/pre-push
```diff
@@ -8,40 +8,47 @@ blocked='ubuntu|root|ec2-user|centos|fedora|debian|admin|bitnami|azureuser|gce-u
...
         fi
-        if [[ $cn =~ $blocked_re ]]; then
-            echo "ERROR: Commit $sha has a default/generic identity: $cn <$ce>" >&2
-            failed=1
+
+        # For new branches, check commits not reachable from any remote branch
+        if [ "$remote_sha" = "0000000000000000000000000000000000000000" ]; then
+            range=("$local_sha" --not --remotes)
+        else
+            range=("$remote_sha..$local_sha")
```
azat: Can we instead use merge-base? That way we should have only commits belongs to topic branch excluding commits brings by i.e. merge commit I guess

.githooks/pre-push
```diff
@@ -8,40 +8,47 @@ blocked='ubuntu|root|ec2-user|centos|fedora|debian|admin|bitnami|azureuser|gce-u
...
-    # separated since `\n` cannot appear in `%an` / `%ae` / `%cn` / `%ce`.
-    while IFS=$'\n' read -r -d '' sha an ae cn ce; do
-        if [[ $an =~ $blocked_re ]]; then
-            echo "ERROR: Commit $sha has a default/generic identity: $an <$ae>" >&2
-            failed=1
+# Only run the per-commit scan when the local identity itself looks default/
+# generic. Otherwise we trust the push: walking the range would false-positive
+# on upstream merge ancestors that carry bot identities like
+# `robot-ch-test-poll`, which aren't the pusher's to fix.
+my_name=$(git config user.name)
```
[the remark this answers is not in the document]
Algunenano: The problem we are trying to solve is this:
* You work for a week in a feature, with proper author.
* You merge upstream/master, which contains commits with a bad author.
* You try to push. It fails because of commits that aren't yours.

If we validate pushed commits, and then we see they don't match your user.name then we are complaining to the wrong person. If we don't complain, then what's the point of validating. I would rather choose false negatives than false positives.

.githooks/pre-push
Algunenano: I think we can check against `--first-parent` I guess

.githooks/pre-push
```diff
@@ -15,12 +15,11 @@ while read -r local_ref local_sha remote_ref remote_sha; do
...
-    if [ "$remote_sha" = "0000000000000000000000000000000000000000" ]; then
-        range=("$local_sha" --not --remotes)
-    else
-        range=("$remote_sha..$local_sha")
-    fi
+    # Exclude anything already reachable from a remote-tracking branch. Commits
+    # brought in by merging a remote-tracking ref (e.g. `git merge private/master`)
+    # are skipped because they live under `refs/remotes/`; commits authored
+    # locally are not on any remote yet and remain in scope.
+    range=("$local_sha" --not --remotes)
```
azat, line 22:
```suggestion
    if [[ "$remote_sha" == "0000000000000000000000000000000000000000" ]]; then
        range=("$local_sha" --not --remotes="$2")
    else
        range=("--first-parent" "$remote_sha..$local_sha")
    fi
```

Maybe this will be better?

Algunenano: I guess you mean `--remotes=$1` right?

The only difference would be that this narrows the remote to the one you are pushing to, which reintroduces the problem if you are using multiple remotes (which is a common practice when working with a fork, or when synchronizing private). If you merge blessed/master while pushing to your local fork you'll get the same error as now if the fork isn't synchronized (who keeps it sync'd?).

I think we can keep this approach and iterate further if some issues appear

Algunenano: Also, reintroducing the branch means reintroducing the issue that AI complained (rightly so) about in https://github.com/ClickHouse/ClickHouse/pull/105856#discussion_r3305143805.

In any case, if this proves too much noise or complexity we can remove this prehook and find another way of avoiding the problem (I can tell AI bot to check for it and complain)

Algunenano: In fact, let's just do it. Let me drop this hook and instead add it to check_style.py

ci/jobs/check_style.py
```diff
@@ -262,6 +262,56 @@ def check_pylint():
...
+
+    info = Info()
+    if not info.base_branch:
+        return ""
+
+    # `actions/checkout` only mirrors the PR head, so `refs/remotes/origin/<base>`
+    # is not populated. Fetch the base branch explicitly and use FETCH_HEAD as
+    # the lower bound of the range.
+    res, _, err = Shell.get_res_stdout_stderr(
+        f"git fetch --no-tags --quiet origin {info.base_branch}"
```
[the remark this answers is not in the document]
Algunenano: I'll get the info via `gh` instead

ci/jobs/check_style.py
```diff
@@ -262,6 +262,77 @@ def check_pylint():
...
+        or os.environ.get("GH_TOKEN")
+        or os.environ.get("GITHUB_TOKEN")
+    )
+    headers = {"Accept": "application/vnd.github.v3+json"}
+    if token:
+        headers["Authorization"] = f"token {token}"
+
+    url = (
+        f"https://api.github.com/repos/{info.repo_name}/pulls/"
+        f"{info.pr_number}/commits?per_page=100"
```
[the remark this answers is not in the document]
Algunenano: At this point I think I'll need to move the check to clang tidy as I did with the submodules...

.githooks/pre-push
azat: I like pre-push more, but, from the other side it won't prevent from committing this (i.e. --no-verify).
Both ways is OK to me

ci/jobs/check_style.py
```diff
@@ -262,6 +262,77 @@ def check_pylint():
...
+    Reject commits in the PR that carry a default/generic author or committer
+    identity (e.g. `ubuntu`, `runner`, `robot-...`). Uses the GitHub PR commits
+    API so the check does not depend on local git history depth.
+    """
+    import requests
+
+    blocked = (
+        "ubuntu|root|ec2-user|centos|fedora|debian|admin|bitnami|"
+        "azureuser|gce-user|runner|jenkins|nobody|user|test|default|guest"
+    )
```
azat: Nice to have a list here

azat (approved):
Except for security concern looks OK, though I guess it is an not a problem either since the PR will still require `can be tested`

.githooks/pre-push
Algunenano: Ok, the other way is turning out way more complex to do without an excessive toll in the CI (unshallowing git is a no-no, using gh complex and limited, etc). I'll just keep it as it was in `7217774`.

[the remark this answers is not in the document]
Algunenano: Fixed

PR #105858, opened 2026-05-26
glibc-compat: honor SETPGROUP / RESETIDS / SETSIGDEF in posix_spawn stub (fixes "clickhouse chdig client")

azat:

Fix "clickhouse chdig client" (uses POSIX_SPAWN_SETPGROUP) by honoring SETPGROUP / RESETIDS / SETSIGDEF in posix_spawn stub

### Version info
- Merged into: `26.6.1.178`

### DIFF
--- a/base/glibc-compatibility/musl/posix_spawn.c
+++ b/base/glibc-compatibility/musl/posix_spawn.c
@@ -71,6 +71,8 @@ static int child(void *args_vp)
 		for (i=1; i<_NSIG; i++) {
 			if (sigismember(&attr->__sd, i) == 1) {
 				sa.sa_handler = SIG_DFL;
+				/* Safe only because sa is all-zero (SIG_DFL is 0): glibc's struct sigaction and the kernel rt_sigaction ABI have different field layouts. */
+				_Static_assert(SIG_DFL == 0, "Not compatible with kernel struct sigaction");
 				ret = __syscall(SYS_rt_sigaction, i, &sa, 0, _NSIG/8);
 				if (ret) goto fail;
 			}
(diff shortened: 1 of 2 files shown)


base/glibc-compatibility/musl/posix_spawn.c
```diff
@@ -44,12 +54,36 @@ static int child(void *args_vp)
...
+
+	/* Reset signal handlers requested via posix_spawnattr_setsigdefault.
+	 * Iterate via sigismember rather than poking sigset_t internals --
+	 * the local spawn.h pulls in glibc's sigset_t (which exposes __val,
+	 * not musl's __bits), so direct field access would not compile. */
+	if (attr->__flags & POSIX_SPAWN_SETSIGDEF) {
+		for (i=1; i<_NSIG; i++) {
+			if (sigismember(&attr->__sd, i) == 1) {
+				sa.sa_handler = SIG_DFL;
+				ret = __syscall(SYS_rt_sigaction, i, &sa, 0, _NSIG/8);
```
Algunenano, line 76:
```suggestion
				sa.sa_handler = SIG_DFL;
				/* Safe only because sa is all-zero (SIG_DFL is 0): glibc's struct sigaction and the kernel rt_sigaction ABI have different field layouts. */
				ret = __syscall(SYS_rt_sigaction, i, &sa, 0, _NSIG/8);
```

Algunenano (commented):
LGTM. Also note that `spawn.h` keeps `__flags` as int, rather than short int (in glibc). I guess this is fine and the layouts are compatible, so probably not worth changing

azat:
>Also note that spawn.h keeps __flags as int, rather than short int (in glibc)

Thanks, I will fix it, even though it is not strictly related here, and it is compatible

base/glibc-compatibility/musl/posix_spawn.c
azat: Indeed, thanks. Also added static_assert.

Algunenano (approved):

PR #105862, opened 2026-05-26
Fix Iceberg Avro NULL partition values written as 0 (#105852)

### LINKED ISSUE #105852: Iceberg Avro table gets wrong write from ClickHouse
### Describe the bug

It seems ClickHouse is writing a partition value incorrectly. Claude analysis:
```
getAvroType for Nullable just unwraps to the nested type — it returns "int" instead of ["null", "int"]. So the partition field in the manifest Avro schema is declared as a plain int, not a nullable union. Combined with the case Field::Types::Null: break; at line 419 that leaves the field at its default value of 0, NULL partition values get written as 0 in the manifest.

Two bugs in IcebergWrites.cpp:

Line 419-420: case Field::Types::Null: break; — does nothing when the partition value is NULL, leaving the Avro field at its default (0 for int) Utils.cpp:618-622: getAvroType strips Nullable and returns the inner type ("int" instead of ["null", "int"]), so the manifest schema doesn't even allow null partition values
That's why it only reproduces with partitioned tables — the data file itself encodes NULLs correctly, but the manifest partition metadata loses the NULL. Spark reads partition column values from the manifest, not the data file, so it sees 0.
```

### How to reproduce

Run [spark.py](https://github.com/user-attachments/files/28264997/spark.py) create
Then in ClickHouse:
```sql
CREATE TABLE t0 ENGINE = IcebergLocal(local, path = '/var/lib/clickhouse/user_files/lakehouses/t0/test/t0', format = 'Avro');
INSERT INTO TABLE t0 (c0) SELECT NULL;
```
Run `spark.py read` and Spark outputs 0 instead of NULL.

_No response_

groeneai:
Fix issue #105852: Iceberg writes silently turn NULL partition values into 0 (or the default of the inner type) on round-trip through Spark.

Two bugs in `src/Storages/ObjectStorage/DataLakes/Iceberg/`:

1. `Utils.cpp` `getAvroType` stripped `Nullable(T)` and returned only the inner Avro type, so the partition field in the manifest schema was declared as a non-nullable `T`.
2. `IcebergWrites.cpp` `generateManifestFile` partition-write loop had `case Field::Types::Null: break;`, a no-op that left the Avro field at its default value (`0` for `int`).

After the fix:

* `getAvroType` returns the Avro union `["null", T]` for a `Nullable(T)` partition column (matches the standard Iceberg manifest encoding).
* The partition write loop wraps each value of a nullable partition in `avro::GenericUnion`, selecting branch 0 for NULL and branch 1 for a concrete value. Non-nullable partitions keep their previous behavior.

New regression test `04279_105852_iceberg_avro_null_partition.sh` covers both directions: a `Nullable(Int32)`-partitioned local Iceberg table receives a NULL and a 42; the test asserts that `SELECT` round-trips NULL and that the per-file Avro manifests contain `(NULL)`, not `(0)`.

Validation:

* Before the fix, the reproducer wrote schema `{"name":"c0","type":"int"}` and the NULL row's manifest had partition `(0)`.
* After the fix, schema is `{"name":"c0","type":["null","int"]}` and the NULL row's manifest has partition `(NULL)`. SELECT also returns NULL.
* Existing partition-write tests `04201_iceberg_manifest_per_file_stats`, `04061_iceberg_update_partition_key_empty_block`, `04079_iceberg_write_retry_explicit_metadata` pass against the fixed binary.

Iceberg writes now preserve NULL values in `Nullable(T)` partition columns. Previously, a NULL written by ClickHouse showed up as the default value of the inner type (`0` for `int`) when read back by Spark or other Iceberg readers. Closes #105852.

- [x] Documentation is unchanged (the on-disk Iceberg manifest layout follows the spec; user-facing SQL behavior is unchanged).


groeneai:
**Pre-PR validation gate** (per session checklist):

* **a) Deterministic repro?** Yes. `CREATE TABLE t (c0 Nullable(Int32)) ENGINE = IcebergLocal('...') PARTITION BY c0; INSERT INTO t VALUES (NULL);` reliably writes `(0)` in the manifest before this change, `(NULL)` after.
* **b) Root cause explained?** `getAvroType(Nullable(T))` stripped the `Nullable` wrapper and returned `"T"`, so the manifest field schema accepted only inner-type values. The partition write loop then handled `Field::Types::Null` with a plain `break;`, leaving the field at its Avro default. Both bugs combined produced `0` where `NULL` was intended.
* **c) Fix matches root cause?** Yes. Schema now returns the Avro union `["null", T]` for `Nullable(T)`; data-write path now emits an `avro::GenericUnion` with branch 0 for NULL and branch 1 for the value.
* **d) Test intent preserved? / new tests added?** Existing tests `04201_iceberg_manifest_per_file_stats`, `04061_iceberg_update_partition_key_empty_block`, `04079_iceberg_write_retry_explicit_metadata` still pass; new test `04279_105852_iceberg_avro_null_partition.sh` covers a Nullable partition with both NULL and non-NULL values.
* **e) Demonstrated in both directions?** Yes. With the old binary the reproducer wrote schema `{"name":"c0","type":"int"}` and the NULL row's manifest had partition `(0)`. With the new binary, schema is `{"name":"c0","type":["null","int"]}` and the NULL row's manifest has partition `(NULL)`. SELECT also returns NULL.
* **f) Fix is general?** Yes. The `Nullable` branch in `getAvroType` covers all inner types it already supports, and the union wrap in the write loop is keyed off `partition_types[i]->isNullable()`. `getAvroType` has only one caller (`extendSchemaForPartitions` for partition fields), and the partition write loop is shared by INSERT, Compaction and Mutations via `generateManifestFile`, so all three write paths get the fix. No defensive null-guard was added at the symptom site; the fix is at the source (schema generator + write loop) where the bad state was produced.

Session: cron:clickhouse-ci-task-worker:20260526-141500

groeneai:
cc @alesapin @scanhex12 -- could you review this? It fixes a manifest-write bug where Iceberg `Nullable(T)` partition columns lose NULL values (issue #105852): the manifest schema declared the field as plain `T` and the write loop's NULL branch was a no-op, so external readers saw `0` instead of NULL.

tests/queries/0_stateless/04279_105852_iceberg_avro_null_partition.sh
```diff
@@ -0,0 +1,46 @@
...
+${CLICKHOUSE_CLIENT} --allow_insert_into_iceberg=1 --query "INSERT INTO ${TABLE} VALUES (NULL)"
+${CLICKHOUSE_CLIENT} --allow_insert_into_iceberg=1 --query "INSERT INTO ${TABLE} VALUES (42)"
+
+# Round-trip via the Iceberg engine itself: NULL must come back as NULL, not 0.
+${CLICKHOUSE_CLIENT} --query "SELECT c0 FROM ${TABLE} ORDER BY isNull(c0) DESC, c0"
+
+# Inspect the per-file manifests directly. The partition field's value in the
+# Avro tuple is what Spark and other Iceberg readers see; if the bug returns,
+# the NULL value will appear as 0.
+for manifest in $(find "${TABLE_PATH}/metadata" -maxdepth 1 -name '*.avro' -not -name 'snap-*.avro' -type f); do
```
PedroTadim: @groeneai should this be tested with parquet and orc formats as well?

groeneai: Good catch — done in 96dbd04c.

The fix itself lives in `generateManifestFile`, which always writes the manifest as Avro (per the Iceberg spec), independent of the data file format passed to the engine. To make that explicit in the test, I parametrized the loop over `Parquet`, `ORC`, and `Avro` data file formats and assert the same `(NULL)` / `(42)` partition tuples in the manifest for all three.

Verified locally: 10/10 passes with full randomization.

PedroTadim (approved):

groeneai:
CI finished on HEAD `b0bfff3c` (after the tidy-build fix and master merge). One real failure, not caused by this PR:

- **Upgrade check (amd_release)** — "Error message in clickhouse-server.log": chronic flaky on the Upgrade check, seen across 25+ unrelated PRs in the last 30 days (#101757, #100272, #104299, #106073, #106271, etc.). Tracked separately; unrelated to the Iceberg manifest change.

The remaining red marks (CH Inc sync, Mergeable Check, PR aggregator) are downstream consequences of that single chronic failure, not new problems. 143 checks pass.

Session: cron:our-pr-ci-monitor:20260604-003000

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
```diff
@@ -388,43 +388,69 @@ void generateManifestFile(
...
+                {
+                    union_field.selectBranch(1);
+                    union_field.datum() = make_value_datum();
+                }
+                partition_record.field(partition_columns[i]) = avro::GenericDatum(union_schema, union_field);
+            }
+            else
+            {
+                if (is_null_value)
                     throw Exception(
```
PedroTadim, line 449: @groeneai, why was the original exception removed? Does Avro support all the types?

groeneai:
## CI ledger — covered SHA: `b0bfff3c85f8ece08dae2d7f56898d6c16879615`

|---|---|---|
| `Upgrade check (amd_release)` | `Error message in clickhouse-server.log (see upgrade_error_messages.txt)` | chronic upgrade-check noise — fires across many recent PRs (groeneai/fix-106336-prewhere-in-pk-pruning, vector-similarity-pk-partial-filter, this PR, more-fuzz, fix-104463-alter-settings, set-and-join-tables-as-indices, etc.). NOT PR-caused; the upgrade-check logfile-error matcher trips on benign migration warnings. |
| `Build (arm_tidy)` | `Build ClickHouse` | clang-tidy strict-mode build failure — to investigate next tick if persists. May be PR-caused (Iceberg writer code). |

Awaiting human review on the Iceberg writer fix. Will investigate arm_tidy build separately if it persists past next CI run.

groeneai:
Correction to ledger above: `Build (arm_tidy) / Build ClickHouse` is also chronic — it fails across 30+ unrelated PRs in the last 14 days (e.g. `reset-session` 16x, `nullable-array` 11x, many docs/feature PRs). Not PR-caused; clang-tidy strict-mode is fragile in trunk. No actionable PR-side build issue.

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
groeneai: @PedroTadim, none of the prior exceptions were removed. The original code had:

- `default:` -> `throw BAD_ARGUMENTS, "Unsupported type to write into avro file {}"` for any unhandled `Field::Types`
- `case Field::Types::Null:` -> silent `break;` (the actual `#105852` bug: NULL was a no-op so the field stayed at its Avro default `0`/empty)

The change refactors the switch into a `make_value_datum` lambda. The `default:` -> `throw BAD_ARGUMENTS` is preserved verbatim inside that lambda (`IcebergWrites.cpp:408-413`). The only behavioral change for the existing supported types is replacing the silent `Field::Types::Null` no-op with explicit handling: NULLs go through the Avro `["null", T]` union for nullable partitions, and a `LOGICAL_ERROR` for the non-nullable case (line 449, the line you commented on). That `LOGICAL_ERROR` is NEW, but it is an assertion against a logically impossible state, not a replacement for a previously thrown exception.

On Avro support: Avro itself supports all the types covered by the switch (`int`, `long`, `string`, `float`, `double`, `bytes`, decimals as `bytes` or `fixed`). The mapping in this file is deliberately the narrow subset that `getAvroType` advertises in the manifest schema (`int`, `long`, `float`, `double`, `string`, plus the new `["null", T]` union for `Nullable`). Anything outside that set is rejected at schema generation time by `getAvroType`'s own `default:` -> `BAD_ARGUMENTS` (`Utils.cpp:631`), so the writer cannot reach the partition loop with an unsupported type unless the schema and value paths drift apart. The `default:` in `make_value_datum` exists exactly to catch that drift and throw with a precise message instead of writing garbage.

If you would prefer the non-nullable NULL case to be `BAD_ARGUMENTS` (user-facing) instead of `LOGICAL_ERROR` (programmer-facing assertion), I can switch it. Per spec a non-nullable partition column should never see a NULL `Field`, so I picked `LOGICAL_ERROR`.

PR #98685, opened 2026-03-04
Skip read-in-order-through-join optimization for parallel replicas

alexey-milovidov:
## Summary
- Skip `read_in_order_through_join` optimization when `ReadFromMergeTree` has parallel replicas enabled
- Applies to all three variants: sorting, aggregation-in-order, and distinct-in-order
- Prevents coordination mode mismatch ("Replica decided to read in Default mode, not in WithOrder") caused by plan construction differences between initiator and remote replicas

With parallel replicas, the initiator's local plan is built with `ignoreASTOptimizations()` while the remote replica is not, which can cause different plan structures. When `read_in_order_through_join` applies on one side but not the other, they end up with different coordination modes (WithOrder vs Default), triggering a LOGICAL_ERROR in `ParallelReplicasReadingCoordinator`.

This fix is consistent with the existing check that already skips read-in-order for parallel replicas in the Union case (lines 1409-1416 of `optimizeReadInOrder.cpp`).

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=98548&sha=8a781939137bab7b45291ac4d013e5f444ccf41a&name_0=PR&name_1=Stress%20test%20%28amd_tsan%29
https://github.com/ClickHouse/ClickHouse/pull/98548

## Test plan
- [ ] Verify the fix compiles
- [ ] CI passes stress tests without the coordination mode mismatch error

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix LOGICAL_ERROR "Replica decided to read in Default mode, not in WithOrder" when using `read_in_order_through_join` with parallel replicas.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### Version info
- Merged into: `26.3.1.430`

PR #95077, opened 2026-01-25
Fix coordination mode mismatch in parallel replicas with read-in-order optimization

matanper:
When using parallel_replicas_local_plan, different replicas could identify different columns as "fixed" (constant due to WHERE filtering) because each replica builds its own query plan independently. This caused replicas to compute different read directions, leading to coordination mode mismatch errors like "Replica N decided to read in WithOrder mode, not in ReverseOrder".

The fix uses AST-based fixed column detection (via getFixedSortingColumns) instead of DAG-based detection when in parallel replicas local plan mode. AST-based detection analyzes the original WHERE/PREWHERE clauses and produces deterministic results across all replicas.

The bug introduced in #94103

### Changelog category (leave one):
- Critical Bug Fix (crash, data loss, RBAC) or LOGICAL_ERROR

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

Fixed regression causing parallel replicas coordination failures with parallel_replicas_local_plan in read-in-order optimization.

PR #105865, opened 2026-05-26
Add regression test for set skip index with IS TRUE / IS FALSE / IS UNKNOWN

groeneai:
Follow-up to #99997 (which added the SQL truth-value predicates `IS TRUE`, `IS FALSE`, `IS UNKNOWN`, and their `IS NOT` variants). @PedroTadim asked on the inline review whether the set skip index is used for these predicates on a `Nullable(Bool)` column.

The answer is yes, the set index already prunes granules in all six cases because the new predicates lower to functions the set index recognises at the atom level (`isNotDistinctFrom(b, true|false)`, `isNull(b)`, `isDistinctFrom(b, true|false)`, `isNotNull(b)`). This PR pins that behaviour with a regression test.

The test creates a `MergeTree` table with `index_granularity = 8` and inserts 24 rows arranged so the three granules contain only true, only false, and only NULL values respectively. For each predicate it asserts:

- `count()` returns the right number of matching rows.
- `ProfileEvents['SelectedMarks']` shows the set index dropped the two unrelated granules for `IS TRUE` / `IS FALSE` / `IS UNKNOWN` (1/3 granules read) and the matching granule for the `IS NOT` forms (2/3 granules read).

...


groeneai:
Pre-PR validation gate (test-only PR):

- New test added (a-f items mostly N/A since there is no bug to fix): `tests/queries/0_stateless/04290_set_index_is_true_false_unknown.sql`.
- Regression catch: the reference asserts `SelectedMarks = 1/3` for every positive form and `2/3` for every negative form. A change that stops the set index from recognising `isNotDistinctFrom(b, true|false)`, `isNull(b)`, `isDistinctFrom(b, true|false)`, or `isNotNull(b)` would bump those counts to `3/3` and fail.
- 100/100 passes under default runner randomization, 10/10 under `enable_analyzer = 0`, 10/10 under `enable_parallel_replicas = 2` with `cluster_for_parallel_replicas = parallel_replicas`. Per-query `SETTINGS enable_parallel_replicas = 0` on each measured `SELECT` keeps `system.query_log` deterministic.
- No code changes, no production behavior change. The set index already prunes correctly via `MergeTreeIndexConditionSet::atomFromDAG` and `mayBeTrueOnGranule`; this just documents that with a regression test as @PedroTadim asked on ClickHouse/ClickHouse#99997.

session: cron:clickhouse-ci-task-worker:20260526-144500

groeneai:
cc @CurtizJ for review (comp-skip-index): tiny regression test pinning the set-skip-index pruning that the existing predicate lowering already gives us for the new IS TRUE / IS FALSE / IS UNKNOWN forms. Follow-up to #99997.

PedroTadim:
@groeneai can you check the test failures?

groeneai:
Triaged both failures. Both are chronic trunk issues unrelated to this PR.

This PR is test-only: 2 files in `tests/queries/0_stateless/` (`04290_set_index_is_true_false_unknown.sql` and its `.reference`), no C++ changes. Our test ran and passed (116 passed in the stateless lane).

### 1. Stress test (arm_release) -- `Logical error: Coordination mode mismatch for stream A: got B, expected C`

Origin: `DB::ParallelReplicasReadingCoordinator::handleInitialAllRangesAnnouncement` raising "got Default, expected WithOrder" (or vice versa). Chronic ParallelReplicasReadingCoordinator race tracked under STID 6117-726a + sibling STIDs (7334-5b92, 7334-48d5, 6117-68e9):

- 20 distinct PRs hit it in last 30 days (31 hits), 2 master hits today (2026-05-27 09:55 + 10:29).
- Authors and scopes are orthogonal: RowBinary defaults, Iceberg REST, DPhyp join reorder, LIMIT BY primary key prefix, http-table-as-file, this set-index test PR, etc.
- Prior umbrella attempts: #98467 and #98685 merged, #95077/#95052 closed. Underlying race is in coordinator init vs parallel-replicas pipeline construction.

This PR cannot affect parallel-replicas coordination.

### 2. Stateless tests (amd_asan_ubsan, db disk, distributed plan, sequential) -- `Server died` / `Lost connection`

`clickhouse-server.err.log` shows the cascade clearly: `system.<X>_log_sender.DistributedInsertQueue` repeatedly hitting `TOO_MANY_PARTS` and `MEMORY_LIMIT_EXCEEDED` (30.03 GiB attempt vs 29.80 GiB max), then `core.MergeMutate.1574-3816.zst` produced by a MergeMutate worker. No sanitizer report, no LOGICAL_ERROR in server.log. Tests reported as failing (`03716_mutations_parts_postpone_reasons_rmt`, `03714_queries_escaping_3`) are downstream of the server crash, not the cause.

Chronic on this exact lane:

- 30+ distinct unrelated PRs hit `Server died` on `Stateless tests (amd_asan_ubsan, db disk, distributed plan, sequential)` in last 14 days, 4 master hits on 2026-05-22.
- Same shape across runs: distributed-insert backlog under random_settings + db disk + ASAN -> memory wall -> MergeMutate crash -> remaining tests get "Lost connection".

Not reproducible from anything this PR adds.

### Recommendation

Both failures are infra/trunk, not PR-caused. The bot AI approval from 2026-05-26 stands; the test file passes consistently on every other lane. Happy to rerun if you would like a fresh set of CI runs.

PedroTadim:
@groeneai can you add a primary key test to check that it is also used with these predicates?

groeneai:
@PedroTadim added `tests/queries/0_stateless/04304_primary_key_is_true_false_unknown.sql` in 8de2d985eedc.

Same 24-row, three-granule layout (`false`/`true`/NULL × 8) but `b Nullable(Bool)` is the leading primary-key column (`ORDER BY (b, id)`, `allow_nullable_key = 1`). Trivial-count and implicit-projection optimisations are disabled per query so `SelectedMarks` reflects actual granule reads.

Observed `KeyCondition` behaviour (locally, 50/50 runs with random settings + random MergeTree settings):

| predicate | granules read / total |
|---|---|
| `IS UNKNOWN` | 2 / 3 |
| `IS NOT UNKNOWN` | 2 / 3 |
| `IS TRUE` / `IS FALSE` / `IS NOT TRUE` / `IS NOT FALSE` | 3 / 3 |

`KeyCondition` only special-cases `isNull` / `isNotNull` (see `src/Storages/MergeTree/KeyCondition.cpp` around line 642). The other four predicates lower to `isNotDistinctFrom` / `isDistinctFrom`, which it treats as the trivial `true` condition — so the test pins current behaviour and will flag any change. Happy to file a follow-up if you think `KeyCondition` should also recognise the distinct-from forms.

PR #105866, opened 2026-05-26
Reject wide integers and Decimal256 in icebergHash

Algunenano:
Closes #101875.

`icebergHash` accepted `Int128/UInt128/Int256/UInt256` (silently truncated to `Int64` via `getInt`) and `Decimal256` (silently truncated to `UInt128`), producing hash collisions. None of these types are defined by the [Iceberg spec](https://iceberg.apache.org/spec/#appendix-b-32-bit-hash-requirements), which only specifies hashing for `int` (32-bit), `long` (64-bit), and `decimal(P,S)` with P <= 38.

They are now rejected with `ILLEGAL_TYPE_OF_ARGUMENT` and a message pointing at the spec and suggesting a cast to `Int64`/`Decimal128`.

`icebergHash` and `icebergBucket` now reject `Int128`, `UInt128`, `Int256`, `UInt256`, and `Decimal256` arguments with an explicit error. Previously they silently truncated wider values and produced colliding hashes. The Iceberg spec only defines hashing for 32/64-bit integers and decimals with precision up to 38; cast to `Int64` or `Decimal128` to keep the previous behaviour for values that fit.

### Version info
- Merged into: `26.6.1.183`


antonio2368 (approved):
`getInt` seems like a really easy way to do the same thing in different places. It's unwanted truncation in best case.

PR #105888, opened 2026-05-26
Fix Iceberg position deletes returning wrong result with multi-file deletes

### LINKED ISSUE #91525: Position deletes incorrect result
### Company or project name

clickhouse

### Describe what's wrong

```
@pytest.mark.parametrize("use_roaring_bitmaps", [0, 1])
@pytest.mark.parametrize("format_version", ["2"])
@pytest.mark.parametrize("storage_type", ["s3", "azure", "local"])
def test_one_connectivity_group(started_cluster_iceberg_with_spark, format_version, storage_type, use_roaring_bitmaps):
    instance = started_cluster_iceberg_with_spark.instances["node1"]
    spark = started_cluster_iceberg_with_spark.spark_session
    TABLE_NAME = (
        "test_data_file_with_delete_file_compacted_"
        + format_version
        + "_"
        + storage_type
        + "_"
        + get_uuid_str()
    )

    spark.sql(
        f"""
        CREATE TABLE {TABLE_NAME} (a bigint, b string) USING iceberg TBLPROPERTIES ('format-version' = '2', 'write.update.mode'=
        'merge-on-read', 'write.delete.mode'='merge-on-read', 'write.merge.mode'='merge-on-read')
        """
    )
    spark.sql(f"INSERT INTO {TABLE_NAME} select id, char(id + ascii('a')) from range(0, 10)")
    spark.sql(f"INSERT INTO {TABLE_NAME} select id, char(id + ascii('a')) from range(10, 20)")

    files = default_upload_directory(
        started_cluster_iceberg_with_spark,
        storage_type,
        f"/iceberg_data/default/{TABLE_NAME}/",
        f"/iceberg_data/default/{TABLE_NAME}/",
    )

    settings = {
        "use_roaring_bitmap_iceberg_positional_deletes": use_roaring_bitmaps,
        "input_format_parquet_use_native_reader_v3": 1,
    }

    create_iceberg_table(storage_type, instance, TABLE_NAME, started_cluster_iceberg_with_spark, additional_settings=list(map(lambda kv: f'{kv[0]}={kv[1]}', settings.items())))
    assert int(instance.query(f"SELECT count() FROM {TABLE_NAME}")) == 20
    assert instance.query(f"SELECT a FROM {TABLE_NAME} ORDER BY a") == '\n'.join([str(x) for x in list(range(0, 20))]) + '\n'

    spark.sql(f"DELETE FROM {TABLE_NAME} WHERE a % 2 = 0")
    spark.sql(f"DELETE FROM {TABLE_NAME} WHERE a % 3 = 0")
    files = default_upload_directory(
        started_cluster_iceberg_with_spark,
        storage_type,
        f"/iceberg_data/default/{TABLE_NAME}/",
        f"/iceberg_data/default/{TABLE_NAME}/",
    )

    assert int(instance.query(f"SELECT count() FROM {TABLE_NAME}")) == 7
    assert instance.query(f"SELECT a FROM {TABLE_NAME} ORDER BY ALL") == '1\n5\n7\n11\n13\n17\n19\n'
```

Yes

### How to reproduce

.

### Expected behavior

_No response_

_No response_

### Additional context

_No response_

groeneai:
## Linked issues
- Closes #91525

For a v2 merge-on-read Iceberg table where a single position-delete file references rows in multiple data files, applying two or more such delete files to the same data file produced the wrong result. The streaming reader returned a different set of rows than the roaring-bitmap reader, and neither matched what Spark sees.

### Reproducer (full SQL in the regression test)

```sql
SET allow_experimental_insert_into_iceberg = 1;
SET allow_insert_into_iceberg = 1;
CREATE TABLE t (a Int64, b String) ENGINE = IcebergLocal('/path/', 'Parquet');
INSERT INTO t SELECT number, char(number + ascii('a')) FROM numbers(0, 10);
INSERT INTO t SELECT number, char(number + ascii('a')) FROM numbers(10, 10);
ALTER TABLE t DELETE WHERE a % 2 = 0;
ALTER TABLE t DELETE WHERE a % 3 = 0;
SELECT groupArray(a) FROM (SELECT a FROM t ORDER BY a);
-- Expected: [1,5,7,11,13,17,19]
-- Before fix (use_roaring_bitmap_iceberg_positional_deletes = 0): [1,3,7,11,13,17]
```

### Root cause

`IcebergStreamingPositionDeleteTransform` requires positions to arrive in ascending order per delete source. It relies on the `equals(file_path, current_data_file)` WHERE clause that `initializeDeleteSources` builds to filter the delete file down to rows for the current data file. That WHERE clause is wired into the Parquet reader as `FormatFilterInfo::filter_actions_dag`, which only drives `KeyCondition`-based row-group / page pruning. It does not filter individual rows. With one row group covering positions for several data files, rows for the wrong data files survive pruning and appear inside the chunk in `(file_path, pos)` order. From the perspective of one data file, positions then come out out of order (e.g. `5, 3, 9`), and the streaming merge silently produces the wrong deletion set.

`IcebergBitmapPositionDeleteTransform` already worked around this by re-checking the path in C++ before adding positions to the roaring bitmap, which is why the bug only surfaced with `use_roaring_bitmap_iceberg_positional_deletes = 0` (the default).

### Fix

Pulled the path matching into a shared helper `IcebergPositionDeleteTransform::filterChunkToCurrentDataFile`. Both readers now call the helper to drop rows that do not match the current data file before consuming the chunk:

- The streaming reader filters each chunk in place and keeps reading until it finds one with at least one matching row, restoring the ascending-positions invariant.
- The bitmap reader uses the same helper instead of inlining the four path variants; behaviour is unchanged.

### Tests

Added stateless test `04291_91525_iceberg_position_deletes_multi_file.sh` that creates an `IcebergLocal` table with two data files, runs two `ALTER ... DELETE` statements that each touch both data files, and verifies the result under all four combinations of `use_roaring_bitmap_iceberg_positional_deletes` and `input_format_parquet_use_native_reader_v3`. Without this fix the streaming variants return `[1,3,7,11,13,17]`; with it they return `[1,5,7,11,13,17,19]`. The Spark-based reproducer from the issue exercises the same code path.

Fixed Iceberg v2 merge-on-read position deletes returning wrong rows when a single delete file references multiple data files and several such delete files apply to the same data file. The streaming reader (`use_roaring_bitmap_iceberg_positional_deletes = 0`) now filters delete-file rows by `file_path` in C++ instead of relying on Parquet row-group pruning, restoring the ascending-positions invariant.

- [x] Documentation is unchanged (bug fix only).

### Version info
- Merged into: `26.6.1.248`

### DIFF
--- a/src/Storages/ObjectStorage/DataLakes/Iceberg/PositionDeleteTransform.cpp
+++ b/src/Storages/ObjectStorage/DataLakes/Iceberg/PositionDeleteTransform.cpp
@@ -7,6 +7,7 @@
 
 #include <Storages/ObjectStorage/DataLakes/Iceberg/PositionDeleteTransform.h>
 
+#include <Columns/IColumn.h>
 #include <Core/Settings.h>
 #include <Formats/FormatFactory.h>
 #include <Formats/ReadSchemaUtils.h>
@@ -140,6 +141,45 @@ size_t IcebergPositionDeleteTransform::getColumnIndex(const std::shared_ptr<IInp
     throw Exception(ErrorCodes::LOGICAL_ERROR, "Could not find column {} in chunk", column_name);
 }
 
+size_t IcebergPositionDeleteTransform::filterChunkToCurrentDataFile(Chunk & chunk, size_t filename_column_index) const
+{
+    const size_t num_rows = chunk.getNumRows();
+    if (num_rows == 0)
+        return 0;
+
+    const auto & filename_column = *chunk.getColumns()[filename_column_index];
+    const auto iceberg_uri = iceberg_object_info->info.data_object_file_path_key.serialize();
+    const auto storage_path = iceberg_object_info->getPath();
+
+    IColumn::Filter filter(num_rows, 0);
+    size_t num_matched = 0;
+    for (size_t i = 0; i < num_rows; ++i)
+    {
+        const auto file_to_delete = filename_column.getDataAt(i);
+        if (file_to_delete == iceberg_uri || file_to_delete == "/" + iceberg_uri
+            || file_to_delete == storage_path || file_to_delete == "/" + storage_path)
+        {
+            filter[i] = 1;
+            ++num_matched;
+        }
+    }
+
+    if (num_matched == num_rows)
+        return num_rows;
+
+    if (num_matched == 0)
+    {
+        chunk.clear();
+        return 0;
+    }
+
+    auto columns = chunk.detachColumns();
+    for (auto & column : columns)
+        column = column->filter(filter, static_cast<ssize_t>(num_matched));
+    chunk.setColumns(std::move(columns), num_matched);
+    return num_matched;
+}
+
 void IcebergBitmapPositionDeleteTransform::transform(Chunk & chunk)
 {
     DeletionVectorTransform::transform(chunk, bitmap);
@@ -149,24 +189,17 @@ void IcebergBitmapPositionDeleteTransform::initialize()
 {
     for (auto & delete_source : delete_sources)
     {
+        const auto position_index = getColumnIndex(delete_source, IcebergPositionDeleteTransform::positions_column_name);
+        const auto filename_index = getColumnIndex(delete_source, IcebergPositionDeleteTransform::data_file_path_column_name);
+
         while (auto delete_chunk = delete_source->read())
         {
-            auto position_index = getColumnIndex(delete_source, IcebergPositionDeleteTransform::positions_column_name);
-            auto filename_index = getColumnIndex(delete_source, IcebergPositionDeleteTransform::data_file_path_column_name);
+            if (filterChunkToCurrentDataFile(delete_chunk, filename_index) == 0)
+                continue;
 
-            auto position_column = delete_chunk.getColumns()[position_index];
-            auto filename_column = delete_chunk.getColumns()[filename_index];
-
-            auto iceberg_uri = iceberg_object_info->info.data_object_file_path_key.serialize();
-            auto storage_path = iceberg_object_info->getPath();
+            const auto & position_column = *delete_chunk.getColumns()[position_index];
             for (size_t i = 0; i < delete_chunk.getNumRows(); ++i)
-            {
-                auto position_to_delete = position_column->get64(i);
-                auto file_to_delete = filename_column->getDataAt(i);
-                if (file_to_delete == iceberg_uri || file_to_delete == "/" + iceberg_uri
-                    || file_to_delete == storage_path || file_to_delete == "/" + storage_path)
-                    bitmap.add(position_to_delete);
-            }
+                bitmap.add(position_column.get64(i));
         }
     }
 }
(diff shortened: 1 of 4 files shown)


groeneai:
### Pre-PR validation gate (per worker policy)

**a) Deterministic repro?** Yes. The stateless test `04291_91525_iceberg_position_deletes_multi_file.sh` reproduces 100% via `IcebergLocal`. The clickhouse-local one-liner:
```
SET allow_experimental_insert_into_iceberg=1, allow_insert_into_iceberg=1;
CREATE TABLE t (a Int64, b String) ENGINE = IcebergLocal('/tmp/ice/', 'Parquet');
INSERT INTO t SELECT number, char(number+ascii('a')) FROM numbers(0,10);
INSERT INTO t SELECT number, char(number+ascii('a')) FROM numbers(10,10);
ALTER TABLE t DELETE WHERE a%2=0;
ALTER TABLE t DELETE WHERE a%3=0;
SELECT groupArray(a) FROM (SELECT a FROM t ORDER BY a
  SETTINGS use_roaring_bitmap_iceberg_positional_deletes=0);
-- Before fix: [1,3,7,11,13,17]   After fix: [1,5,7,11,13,17,19]
```

**b) Root cause explained?** Delete file #2 stores rows sorted by `(file_path, pos)` as `(DF2, 5), (DF1, 3), (DF1, 9)`. The `equals(file_path, current_data_file)` WHERE clause attached to the Parquet reader feeds `KeyCondition`, which only prunes whole row groups. Inside a surviving row group the rows still come through in their stored order, so for DF1 the positions for source 1 reach `IcebergStreamingPositionDeleteTransform` as `[5, 3, 9]`. The streaming merge maintains a `std::set<(pos, source_idx)>` keyed by the smallest pos per source and assumes `delete_source->read()` returns ascending positions per source. With `[5, 3, 9]`, position 3 is consumed *after* `row_idx=5`, so it never deletes row 3 and instead consumes position 5 against the wrong data row. The result for both DF1 and DF2 becomes `[1, 3, 7]` (DF1 values) / `[11, 13, 17]` (DF2 values), losing `5` and `19` and keeping `3`. The bitmap reader does not have this problem because it re-checks the path explicitly in C++ before adding to the bitmap.

**c) Fix matches root cause?** Yes. The fix restores the ascending-positions invariant by filtering each delete chunk by `file_path` in C++ - the same predicate the bitmap reader already used - so the streaming merge once again receives `[3, 9]` for DF1 and `[5]` for DF2. The Parquet-level WHERE clause stays in place for row-group pruning; the C++ filter only handles rows that survive pruning.

**d) Test intent preserved?** No tests modified. New regression test added (`04291`) covering both readers (`roaring=0/1`) and both Parquet code paths (`input_format_parquet_use_native_reader_v3=0/1`).

**e) Demonstrated in both directions?** Yes. Before the source change the reproducer produces `count=7, sum=52, groupArray=[1,3,7,11,13,17]` (only 6 distinct values - `count()` returns 7 from manifest stats, the scan finds 6). After the change all four combinations of the two settings return `count=7, sum=73, groupArray=[1,5,7,11,13,17,19]`.

**f) Fix is general?** Yes. The same root cause affects every Iceberg position-delete file that references multiple data files (typical of v2 merge-on-read). The fix is in the shared base transform, mirrors what the bitmap reader has been doing, and covers both readers via the same helper. No symmetric bug remains: any future reader added next to the streaming / bitmap pair can call the same helper.

Session: `cron:clickhouse-ci-task-worker:20260526-153000`

groeneai:
cc @PedroTadim @scanhex12 @alesapin - per #91525 directive. PR ready for review. Two-file change in `src/Storages/ObjectStorage/DataLakes/Iceberg/PositionDeleteTransform.{cpp,h}` plus one stateless regression test. Root cause and pre-PR validation details above.

src/Storages/ObjectStorage/DataLakes/Iceberg/PositionDeleteTransform.cpp
george-larionov, line 68: On lines 159-160 we check the delete filepath against a few different values to handle inconsistent write formats, however here our WHERE filter will just filter against `iceberg_data_path`. Should this filter also be updated to match the 4 different versions of the filename? @scanhex12 and @alesapin i'm not familiar enough with the Iceberg code to know if this is a real issue or handled eslewhere, please advise.

george-larionov (approved):
This change looks good to me. My other comment is regarding code that was not touched in this PR.

PR #105889, opened 2026-05-26
Allow DP join reordering with parallel replicas

nickitat:
- Performance Improvement

DP JOIN reordering is now allowed with parallel replicas.

---

It looks like I really need it to implement support for JOINs with AutoPR. Otherwise, the local and distributed plans won't match in too many cases.

### DIFF
--- a/tests/queries/0_stateless/04277_parallel_replicas_join_dp_reorder.sql
+++ b/tests/queries/0_stateless/04277_parallel_replicas_join_dp_reorder.sql
@@ -1,0 +1,31 @@
+-- Regression test: the DP join-order optimizer must keep running when parallel replicas is enabled.
+-- Previously `QueryPlanOptimizationSettings` forced `query_plan_optimize_join_order_limit=0`
+-- whenever `allow_experimental_parallel_reading_from_replicas && max_parallel_replicas > 1`, which
+-- silently skipped join reordering for any PR query.
+
+DROP TABLE IF EXISTS pr_dp_small;
+DROP TABLE IF EXISTS pr_dp_large;
+
+CREATE TABLE pr_dp_small (x UInt32) ENGINE = MergeTree ORDER BY x AS SELECT number FROM numbers(100);
+CREATE TABLE pr_dp_large (x UInt32) ENGINE = MergeTree ORDER BY x AS SELECT number FROM numbers(100000);
+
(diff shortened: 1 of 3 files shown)


antaljanosbenjamin (approved):

tests/queries/0_stateless/04277_parallel_replicas_join_dp_reorder.sql
```diff
@@ -0,0 +1,59 @@
...
+DROP TABLE IF EXISTS pr_dp_a;
+DROP TABLE IF EXISTS pr_dp_b;
+DROP TABLE IF EXISTS pr_dp_c;
+
+CREATE TABLE pr_dp_a (x UInt32) ENGINE = MergeTree ORDER BY x AS SELECT number % 100   FROM numbers(100);
+CREATE TABLE pr_dp_b (x UInt32) ENGINE = MergeTree ORDER BY x AS SELECT number % 1000  FROM numbers(1000);
+CREATE TABLE pr_dp_c (x UInt32) ENGINE = MergeTree ORDER BY x AS SELECT number % 100   FROM numbers(100000);
+
+SET
+  enable_parallel_replicas = 1,
```
antaljanosbenjamin, line 23: Actually AI had a good comment:
>• Findings
>
>  - Medium: tests/queries/0_stateless/04277_parallel_replicas_join_dp_reorder.sql:22 does not pin automatic_parallel_replicas_mode = 0, and uses best-effort enable_parallel_replicas = 1. tests/clickhouse-test randomizes
>    automatic_parallel_replicas_mode to 2 25% of the time, and the analyzer disables enable_parallel_replicas when automatic mode is non-zero. In that run, the old buggy code would no longer force query_plan_optimize_join_order_limit = 0,
>    so this regression test can pass without exercising the intended parallel-replicas path. Add automatic_parallel_replicas_mode = 0; using enable_parallel_replicas = 2 or asserting the PR plan shape would make the coverage stronger.

nickitat: Indeed, thanks.

fm4v:
@groenai add test as performance test tests/queries/0_stateless/04277_parallel_replicas_join_dp_reorder.sql

PR #105890, opened 2026-05-26
Sparsity-aware part/granule pruning and trivial-count optimization

Algunenano:
- Experimental Feature

Experimental: serve `SELECT count(*) FROM t WHERE col <op> default(col)` from per-column sparsity statistics in `serialization.json` without any data scan when the predicate exactly partitions the column into defaults and non-defaults.

Recognised predicate shapes:
- `col = const` / `col != const` for `Int*` / `UInt*`, `String` / `FixedString`, `Date` / `DateTime` / `DateTime64`, `Decimal*`, `UUID`, `IPv4` / `IPv6`. Excluded for `Float*` (treats `+0.0` and `-0.0` separately), `Enum*` (named default is not bitwise zero), `Nullable`, `LowCardinality`, and composite types (`Tuple`, `Array`, `Map`) where SQL equality and the persisted counter diverge.
- `IS NULL` / `IS NOT NULL` on `Nullable` columns.
- `empty(col)` / `notEmpty(col)` on `String` columns.
- `col = true` / `col != true` on `Bool` columns.
- `col > 0`, `col >= 1`, `col < 1`, `col <= 0` on unsigned integer columns.
- Bare `col` / `NOT col` on `Int*`, `UInt*`, `Bool` columns (truthy test).

The constant side is compared against the column default via the least common supertype (so e.g. `dt_col = toDateTime64('…', 3)` is classified correctly); lossy conversions (`Decimal` truncation, out-of-range `Bool`, etc.) opt out rather than misclassify.

Two new settings, both experimental:
- `optimize_trivial_count_with_sparsity_filter` (default off) — extends `optimize_trivial_count_query` to the predicates above, served from the per-column `num_defaults` / `num_rows` counters with no scan. Honors `select_sequential_consistency` (the rewrite consults the same ZK-filtered active-parts list as `ReplicatedMergeTree::totalRows`).
- `compute_exact_num_defaults_for_sparse_columns` MergeTree setting (default off) — gates the O(rows) exact counting at part-write time. Without it the producer keeps the pre-existing sampled estimate and the part is silently opted out of the new optimization.

Also included as part of the sparsity work:
- `FilterWithCachedCount` preserves the sparse filter shape end-to-end so `MergeTreeRangeReader::optimize` shrinks reads on sparse conjuncts as aggressively as PREWHERE does on dense ones.
- `tryOptimizeTopK` runs in the second optimization pass, after PREWHERE has claimed sparse conjuncts, so `__topKFilter` no longer steals PREWHERE from a classifiable sparse predicate.
- `ColumnSparse::filter` walks the offsets array with 64-bit popcount, dropping O(rows) to O(non_defaults + rows/64).
- `IColumnHelper::getNumberOfDefaultRows` gets container-column overrides (`ColumnArray`, `ColumnMap`, `ColumnObject`, `ColumnTuple`) instead of the O(rows × subcolumns) virtual-call default.

### Part-statistics change

MergeTree already records `num_defaults` and `num_rows` per column in `serialization.json`, but the old `num_defaults` was a sampled estimate (sufficient for the sparse/dense kind decision, not for serving counts). This PR adds an `exact_num_defaults: true` flag to `SerializationInfo` that's set when the producer counted every default, and the trivial-count rewrite only fires when the flag is present.

Backward compatibility:
- **New server, legacy part** — the flag is absent in the JSON; the new reader defaults it to `false`; the trivial-count code guards on the flag so the part is opted out of the optimization. The pre-existing sampled counters are still used for the kind decision, exactly as before.
- **Old server, new part** — `serialization.json` for legacy parts is byte-identical because the new writer only emits the key when `true`; for parts written by the new server the JSON gains one extra key (`exact_num_defaults: true`). Old `SerializationInfo::fromJSON` parsing uses `Poco::JSON::Object`, which silently ignores unknown keys, so the older server reads the part normally.
- No format version bump and no migration: the change is purely additive in JSON and read-tolerant on both sides.

Tests under `tests/queries/0_stateless/04215_*`, `04278_*`, `04279_*`, `04281_*`, `04283_*`, `04285_*`, `04325_*`, `04326_*`, `04328_*` cover correctness across Nullable columns, persisted exact counters round-tripped through restart, decimal precision-loss guards, supertype-based classification, and sequential consistency on Replicated tables.

### Version info
- Merged into: `26.7.1.682` (included in `26.7` and later)


src/DataTypes/Serializations/SerializationInfo.cpp
```diff
@@ -65,29 +66,44 @@ void writeJSONKeyValue(std::string_view key, bool value, WriteBuffer & out)
...
-    size_t rows = column.size();
-    double ratio = column.getRatioOfDefaultRows(ColumnSparse::DEFAULT_ROWS_SEARCH_SAMPLE_RATIO);
-
-    num_rows += rows;
-    num_defaults += static_cast<size_t>(ratio * static_cast<double>(rows));
+    /// Count defaults exactly: downstream consumers (trivial count rewrite,
+    /// sparse-based part/granule pruning) cannot accept the sampling estimate.
+    bool was_empty = (num_rows == 0);
+    num_rows += column.size();
+    num_defaults += column.getNumberOfDefaultRows();
```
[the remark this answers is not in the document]
Algunenano: This is acceptable for evolution. getNumberOfDefaultRows has been optimized (in this PR) for the types that weren't already optimal

src/DataTypes/Serializations/SerializationSparse.cpp
```diff
@@ -209,6 +215,53 @@ size_t readOrGetCachedSparseOffsets(
     size_t & read_rows,
     size_t & skipped_values_rows)
 {
+    /// `getSubcolumnNameForStream(path + [SparseOffsets], encode_sparse_stream=true)` equals
```
Avogar: This code looks strange, I assume it's because of bad conflict resolution after my PR that changed this strange "optimized" (which I believe doesn't optimize anything and looks very bad) interaction with cache to regular `getElementFromSubstreamsCache/addElementToSubstreamsCache`

src/Processors/QueryPlan/ReadFromMergeTree.cpp
```diff
@@ -2708,6 +2722,28 @@ ReadFromMergeTree::AnalysisResultPtr ReadFromMergeTree::selectRangesToRead(
         {
             result.parts_with_ranges = MergeTreeDataSelectExecutor::filterPartsByPrimaryKeyAndSkipIndexes(filter_context, res_parts, result.index_stats);
 
+            /// Lazily create the share so the planning-mode analyzer (`filterMarkRangesBySparsityInfo`)
+            /// can persist its decompressed offsets here, and the scan-side `MergeTreeReader`s can
+            /// then serve their sparse-column reads from memory.
+            if (settings[Setting::use_sparsity_info_for_pruning] == SparsityPruningMode::Planning && !result.sparse_offsets_share)
+                result.sparse_offsets_share = std::make_shared<SparseOffsetsShare>();
```
Avogar: What do you think avout adding another setting to control the creating and usage of `SparseOffsetsShare`? So if there is a bug there we can fallback to rereading the offsets from disk. Or you think it doesn't make sense and we can just disable the pruning altogether in such case?

src/Storages/MergeTree/MergeTreeReaderWide.cpp
```diff
@@ -570,6 +591,28 @@ void MergeTreeReaderWide::prefetchForColumn(
         if (ISerialization::isEphemeralSubcolumn(substream_path, substream_path.size()))
             return;
 
+        /// The values stream is skipped during deserialization when the caller only wants
+        /// `SparseOffsets`; mirror that here so the prefetcher doesn't issue reads for the
+        /// `SparseElements` substream we will never consume.
+        if (only_read_sparse_offsets)
```
Avogar: IMO this should be done via setting in `EnumerateStreamsSettings`, we can just skip enumerating `SparseElements` if it's set, just like we do in deserialization with `skip_sparse_values_substream`. But If you think it's too much, then we can keep this.

src/Storages/MergeTree/SparseOffsetsShare.cpp
```diff
@@ -0,0 +1,265 @@
...
+    {
+        const auto it = per_part_surviving_ranges.find(part_entry.first);
+        if (it == per_part_surviving_ranges.end())
+            return true;
+        return pruneColumns(part_entry.second, it->second);
+    });
+}
+
+std::unique_ptr<SubstreamsCacheSparseOffsetsElement>
+SparseOffsetsShare::sliceFromBucket(
```
Avogar: One letter variable names makes it difficult to read the code in some places. Like `const auto * a = std::lower_bound(b, e, lo);` or `produce_total += c - a;` or `const auto * e = b + rd.size();`. Let's try to use something meaningful, even if it will be long names

src/Storages/MergeTree/SparsityFilter.h
```diff
@@ -0,0 +1,60 @@
...
+};
+
+struct RecognisedSparsityPredicate
+{
+    String column_name;
+    SparsityPredicateClass predicate_class;
+};
+
+/// Recognise a predicate that partitions one column of `table_expression_node` into
+/// default vs non-default rows. Returns nullopt for any other shape. Recognised forms:
```
Avogar, line 38: What about condition `WHERE numeric_or_bool_column` without any function?

Avogar (commented):
Amazing job! Looks good to me, just a few minor comments

src/DataTypes/Serializations/SerializationSparse.cpp
Algunenano: Yes, will simplify

src/Processors/QueryPlan/ReadFromMergeTree.cpp
Algunenano: I think we can disable pruning in that case. Even though it's declared as experimental it's running in CI via randomization, so it's well covered. And if we find an issue we can switch it off while we fix it

src/Storages/MergeTree/MergeTreeReaderWide.cpp
Algunenano: Makes sense and it fits nicely with the current enumerateStreams order. Will give it a try

src/Storages/MergeTree/SparsityFilter.h
Algunenano: Good idea. Both `WHERE col` and `WHERE NOT col` can be supported on bool and integer types

src/Processors/QueryPlan/ReadFromMergeTree.cpp
```diff
@@ -2851,6 +2865,28 @@ ReadFromMergeTree::AnalysisResultPtr ReadFromMergeTree::selectRangesToRead(
...
+
+            result.parts_with_ranges = MergeTreeDataSelectExecutor::filterMarkRangesBySparsityInfo(
```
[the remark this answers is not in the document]
Algunenano: > Could we run filterMarkRangesBySparsityInfo on result_parts_ranges after distributedIndexAnalysisOnReplicas, or explicitly disable/document the combination if that integration is not intended yet?

Yes, I think we should prune based on sparsity after distributed index analysis too. I'll have a look

CurtizJ:
Do we really need granule pruning by sparse offsets (`use_sparsity_info_for_pruning`) at all? I think `PREWHERE` on a sparse column already achieves almost the same, and pruning doesn't reduce the amount of read data.

For predicates of the `col != default` class, the pruned granules are all-default granules, which contribute nothing to the values substream. So with or without pruning we read the same bytes: all offsets plus all non-default values. The only class where pruning saves I/O is `col = default`, but on a sparse column that predicate matches the vast majority of rows, so such a query is unselective anyway. I'd expect a marginal improvement at best, and even a slowdown for queries with `LIMIT` because the offsets are eagerly read from all ranges in data part.

For comparison, here is what already happens with `PREWHERE` on a range that has only default values in sparse column:
1. Read the offsets (one integer per granule) and no values.
2. Create a sparse column whose values contain only one default row.
3. Evaluate a predicate on non-default values plus that one default row, and if the result for the default row is 0 (exactly the `col != default` class), the result stays sparse.
4. Granules where the filter is all-zero are skipped for all remaining columns (the existing mechanism in `MergeTreeRangeReader`).

So the I/O is identical, and the skipping already works; the inefficiency is only that `MergeTreeRangeReader` doesn't support `SparseFilterDescription` yet: `FilterWithCachedCount` expands the filter to a dense column. Fixing that is a small, self-contained change localized in `MergeTreeRangeReader` (keep the filter sparse in `FilterWithCachedCount`, count zero tails by binary search over the offsets), and it would benefit every query with a sparse `PREWHERE` column, not only the recognized predicate shapes.

So I'm afraid we're introducing a large and complex piece of code for a marginal improvement. I'd rather integrate `SparseFilterDescription` support into `MergeTreeRangeReader`.

`optimize_trivial_count_with_sparsity_filter` is a reasonable optimization since the query is answered from metadata only.

Algunenano:
Done

Algunenano:
Small issue with topK optimization killing prewhere. The previous method worked since it didn't depend on moving the condition to prewhere but the new one doesn't. Need to handle it properly with topK filters

alexey-milovidov:
Updating the branch to fix the `04001_join_reorder_through_expression` test, which was broken today in master.

Avogar (approved):

PR #105903, opened 2026-05-26
Require can be tested for BackportPR

### LINKED ISSUE #105901: BackportPR workflow does not use the trusted-contributor gate from the main PR workflow
`BackportPR` and the main `PR` workflow currently use different trust gates for external pull requests.

The main `PR` workflow includes `can_be_tested` in its `pre_hooks`, while `BackportPR` does not.

Sources:
- BackportPR: https://github.com/ClickHouse/ClickHouse/blob/master/ci/workflows/backport_branches.py
- PR: https://github.com/ClickHouse/ClickHouse/blob/master/ci/workflows/pull_request.py

Since both workflows handle `pull_request` events, I would expect them to use the same maintainer-approval / trusted-contributor gate unless the difference is intentional.

Expected behavior:
- `BackportPR` also uses `can_be_tested`, or
- the different behavior is documented and intentional

fallintoplace:
## Summary
- add the existing `can_be_tested` pre-hook to `BackportPR`
- align the release-branch PR workflow with the trusted-contributor gate already used by the main `PR` workflow

## Why
The main `PR` workflow already requires the trusted-contributor / maintainer-approval gate before running jobs. `BackportPR` handles `pull_request` events for release branches but was missing that same pre-hook.

This PR keeps the scope intentionally minimal and only adds the existing gate. It does not change runner labels or any other workflow policy.

## Validation
- `git diff --check`
- regenerated workflows locally with `python3 -m ci.praktika yaml`
- checked that the source diff stays minimal

## Related
- closes #105901

### DIFF
--- a/ci/workflows/backport_branches.py
+++ b/ci/workflows/backport_branches.py
@@ -3,6 +3,7 @@ from praktika import Workflow
 from ci.defs.defs import DOCKERS, SECRETS, ArtifactConfigs
 from ci.defs.job_configs import JobConfigs
 from ci.jobs.scripts.workflow_hooks.filter_job import should_skip_job
+from ci.jobs.scripts.workflow_hooks.trusted import can_be_tested
 
 workflow = Workflow.Config(
     name="BackportPR",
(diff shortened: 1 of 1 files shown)


maxknv:
Hi @fallintoplace, thanks for checking in. The `can_be_tested` label function is hard to explain without getting into internal details. It's not exactly the trust gate, and it doesn't really act as an extra guard on top of manual workflow approval, which is required for external contributions. So I believe this change is not required.

PR #105937, opened 2026-05-27
Fix std::bad_alloc on first-use JIT init in non-jemalloc macOS builds

### LINKED ISSUE #105868: JIT initialization fails on macOS arm64 for under-default aligned operator new
## Reproducer

On current `master` (`GIT_HASH=256c1a5a0f1bc3d95de93b2f5a15af7533b93ceb`), any expression that triggers first-use JIT initialization can fail in a fresh `clickhouse local` process on macOS arm64.

Minimal reproducer:

```bash
./clickhouse local --query "
SELECT sum(number + 1)
FROM numbers(384)
SETTINGS
    compile_expressions = 1,
    min_count_to_compile_expression = 0,
    max_threads = 1
"
```

Result before the local fix:

```text
Code: 1001. DB::Exception: std::bad_alloc: std::bad_alloc. (STD_EXCEPTION)
```

The original reproducer also fails on the 4th repetition with the default `min_count_to_compile_expression = 3`:

```bash
Q="WITH arrayMap(j -> toFloat32(j * 0.001), range(384)) AS ref SELECT arraySum(ref) SETTINGS max_threads = 1"
for i in 1 2 3 4; do
    printf '%s;\n' "$Q"
done | ./clickhouse local --multiquery
```

The 4th query is the first query that passes the default JIT compilation threshold.

## Root Cause Found Locally

This is broader than `arrayMap` or `Float32`.

The failure happens during first-use `CHJIT` initialization:

```text
llvm::allocate_buffer(unsigned long, unsigned long)
llvm::DenseMapBase<llvm::DenseMap<llvm::MCRegister, int, ...>>::grow(unsigned int)
createAArch64MCRegisterInfo(llvm::Triple const&)
llvm::CodeGenTargetMachineImpl::initAsmInfo()
llvm::AArch64TargetMachine::AArch64TargetMachine(...)
DB::CHJIT::getTargetMachine()
DB::CHJIT::CHJIT()
DB::ActionsDAG::compileFunctions(...)
```

`lldb` showed the failing allocation is tiny, not real OOM:

```text
llvm::DenseMap<llvm::MCRegister, int>::grow
requested size:      0x200
requested alignment: 0x4
```

`contrib/llvm-project/llvm/lib/Support/MemAlloc.cpp` calls aligned `operator new` whenever `__cpp_aligned_new` is available:

```cpp
::operator new(Size, std::align_val_t(Alignment), std::nothrow)
```

In the ClickHouse binary this reaches `Memory::newNoExcept(size, align)`, which currently forwards the requested alignment directly to `aligned_alloc`:

```cpp
__real_aligned_alloc(static_cast<size_t>(align), alignUp(size, static_cast<size_t>(align)))
```

On macOS arm64, `aligned_alloc(4, 512)` returns `nullptr`. A standalone libc++ `operator new(512, std::align_val_t(4), std::nothrow)` returns non-null, so this is not real memory pressure and not an LLVM query-planning issue. The ClickHouse allocation interceptor should treat under-default alignments as regular allocations.

## Local Fix Candidate

Normalize alignment inside `src/Common/memory.h`:

- if `align <= __STDCPP_DEFAULT_NEW_ALIGNMENT__`, use the regular allocation/tracking/deallocation path;
- use aligned allocation only for over-aligned requests.

That avoids changing the bundled LLVM fork and fixes all locally reproduced JIT cases.

Local validation after the ClickHouse-side fix:

```text
ninja -C build_jit_master_core clickhouse
=> ninja_exit=0

SELECT sum(number + 1)
FROM numbers(384)
SETTINGS compile_expressions = 1, min_count_to_compile_expression = 0, max_threads = 1;
=> 73920

SELECT sum(toFloat32(number * 0.001))
FROM numbers(384)
SETTINGS compile_expressions = 1, min_count_to_compile_expression = 0, max_threads = 1;
=> 73.53599999996368

WITH arrayMap(j -> toFloat32(j * 0.001), range(384)) AS ref
SELECT arraySum(ref)
SETTINGS compile_expressions = 1, min_count_to_compile_expression = 0, max_threads = 1;
=> 73.53599999996368
```

I also added a local stateless regression test that forces first-use expression JIT in a fresh `clickhouse local` process and checks the `ExpressionJIT: Compile expression` trace so it does not silently pass through the interpreter path.

## Environment

- Built locally from current `master`.
- `GIT_HASH`: `256c1a5a0f1bc3d95de93b2f5a15af7533b93ceb`
- Bundled LLVM in build: `21.1.5`
- `USE_EMBEDDED_COMPILER`: `ON`
- Version string: `26.6.1.1`
- Platform: macOS arm64

## Related Issues Checked

Related JIT issues, but not this allocator failure mode:

- #105662: planning-time SEGV through `LLVMFunction::isDeterministic`.
- #105797: SEGV in `CHJIT::runOptimizationPassesOnModule`.

### Version info
- Resolved by: #105937
- Merged into: `26.6.1.202`

Algunenano:
On macOS non-jemalloc builds, the first JIT compilation in `clickhouse local` fails with `std::bad_alloc: std::bad_alloc. (STD_EXCEPTION)`. LLVM's `DenseMap<MCRegister, int>::grow` requests `::operator new(512, std::align_val_t(4), std::nothrow)`, and ClickHouse's aligned `newNoExcept` forwarded that straight to system `aligned_alloc(4, 512)`, which returns `nullptr` (`EINVAL`) because POSIX requires alignment >= `sizeof(void *)`. macOS enforces this strictly; glibc happens to be lenient, which is why Linux non-jemalloc builds don't trip on it.

This patch mirrors what libc++ already does in `operator_new_aligned_impl` (`contrib/llvm-project/libcxx/src/new.cpp`): if the requested alignment is below `sizeof(void *)`, bump it up before calling `aligned_alloc`. The jemalloc path is unaffected — jemalloc's `je_aligned_alloc` sets `sopts.min_alignment = 1` and accepts any power-of-two alignment by design, which is why the official macOS binary (built `RelWithDebInfo` with jemalloc) does not hit this.

Closes #105868.

...

### Version info
- Merged into: `26.6.1.202`

### DIFF
--- a/src/Common/memory.h
+++ b/src/Common/memory.h
@@ -45,7 +45,14 @@ inline ALWAYS_INLINE void * newNoExcept(std::size_t size) noexcept
 
 inline ALWAYS_INLINE void * newNoExcept(std::size_t size, std::align_val_t align) noexcept
 {
-    return __real_aligned_alloc(static_cast<size_t>(align), alignUp(size, static_cast<size_t>(align)));
+    size_t alignment = static_cast<size_t>(align);
+#if !USE_JEMALLOC
+    /// POSIX `aligned_alloc` requires alignment >= `sizeof(void *)` (strictly enforced on macOS).
+    /// Mirror libc++'s `operator_new_aligned_impl` in `contrib/llvm-project/libcxx/src/new.cpp`.
+    if (alignment < sizeof(void *))
+        alignment = sizeof(void *);
+#endif
+    return __real_aligned_alloc(alignment, alignUp(size, alignment));
 }
 
 inline ALWAYS_INLINE void deleteImpl(void * ptr) noexcept


antonio2368 (approved):

antonio2368 (approved):

antonio2368 (approved):

antonio2368:
Sorry, it seems when GitHub shows 500 on approve it doesn't mean it didn't go through.
Also the PR so good I had to approve 3 times.

PR #105940, opened 2026-05-27
Fix projection Min-Max index pruning

Michicosun:

During index analysis, the min-max index should be used, built from the same metadata snapshot as the key condition. Before this patch, the projection parent part's min-max index was always used. This patch changes this behavior to use the min-max index loaded from the projection iself.

### Version info
- Merged into: `26.6.1.237`


tests/queries/0_stateless/04296_part_level_min_max_in_partitioned_projection.sql
```diff
@@ -0,0 +1,18 @@
...
+settings enable_block_number_column = 1, enable_block_offset_column = 1, allow_commit_order_projection = 1, part_minmax_index_columns = 'with_block_number_offset', add_minmax_index_for_block_number_column = 1, add_minmax_index_for_block_offset_column = 1, index_granularity = 8192;
+
+insert into t select 1, 1;
+insert into t select 1, 2;
+insert into t select 1, 3;
+insert into t select 1, 4;
+insert into t select 1, 5;
+optimize table t final;
+
+select 'Found' from t where _partition_id = '1' and _block_number = 3 order by (_block_number, _block_offset);
```
Michicosun, line 18: No rows were returned previously.

yariks5s (approved):

PR #105943, opened 2026-05-27
Make cascading materialized views see the row being inserted

ahmadov:

- Improvement

A materialized view whose inner SELECT reads the source storage via `INNER JOIN` back to the source can miss the row currently being inserted: `MergeTreeSink::consume` and `ReplicatedMergeTreeSink::consume` defer the part commit via the `delayed_chunk` pattern, but the MV cascade acquires its `StorageSnapshot` before that deferred commit runs.

The new setting `wait_for_part_commit_in_dependent_materialized_views` (default `false`) makes the sink finish the delayed chunk inside `consume` whenever the destination has at least one dependent view, committing the part before yielding to the cascade. When disabled, the existing parallel-write overlap is preserved.

### Version info
- Merged into: `26.6.1.518`


src/Storages/MergeTree/MergeTreeSink.cpp
```diff
@@ -211,9 +216,18 @@ void MergeTreeSink::consume(Chunk & chunk)
...
+        delayed_chunk = std::make_unique<MergeTreeDelayedChunk>();
+        delayed_chunk->partitions = std::move(partitions);
+        finishDelayedChunk();
+    }
+    else
+    {
+        finishDelayedChunk();
+        delayed_chunk = std::make_unique<MergeTreeDelayedChunk>();
+        delayed_chunk->partitions = std::move(partitions);
+    }
```
antaljanosbenjamin: What happens if `materialized_views_ignore_errors = 0, wait_for_part_commit_in_dependent_materialized_views = 1`. Wouldn't the data be committed to the table in case an exception is thrown in the depedant view? I think we should think about what is the correct way to handle this.

src/Storages/MergeTree/MergeTreeSink.cpp
```diff
@@ -68,6 +70,9 @@ MergeTreeSink::MergeTreeSink(
     , context(context_)
     , storage_snapshot(storage.getStorageSnapshotWithoutData(metadata_snapshot, context_))
     , deduplicate((*storage.getSettings())[MergeTreeSetting::non_replicated_deduplication_window] > 0 && storage.getDeduplicationLog() != nullptr)
+    , synchronously_commit_part_for_dependent_views(
```
antaljanosbenjamin: The same changes needs to be done in private for SMT, right?

ahmadov: yes, this public PR is fork of https://github.com/ClickHouse/clickhouse-private/pull/58692

src/Storages/MergeTree/MergeTreeSink.cpp
ahmadov: What about just documenting the behavior properly and let the user to decide to enable the setting since it's disabled by default?

src/Storages/MergeTree/MergeTreeSink.cpp
```diff
@@ -211,9 +216,18 @@ void MergeTreeSink::consume(Chunk & chunk)
     }
     deduplication_info->setPartWriterHashes(all_partwriter_hashes, chunk.getNumRows());
 
-    finishDelayedChunk();
-    delayed_chunk = std::make_unique<MergeTreeDelayedChunk>();
-    delayed_chunk->partitions = std::move(partitions);
+    if (synchronously_commit_part_for_dependent_views)
+    {
+        delayed_chunk = std::make_unique<MergeTreeDelayedChunk>();
```
CheSema: It looks as wrong sequence.
Here you looses what was delayed in `delayed_chunk` before.
The fist call `finishDelayedChunk();` has to be there. `synchronously_commit_part_for_dependent_views` adds the second one.

src/Storages/MergeTree/MergeTreeSink.cpp
CheSema: `materialized_views_ignore_errors=0` -- ignores the errors in MV's processors (select and sinks)
The data might either be commited in the dest table in that case or not. We do not provide guarantee here. We encourage users do reties, they will commit missed data.

`wait_for_part_commit_in_dependent_materialized_views=1` adds prefix guarantees indeed.

I do not see the issue here.  Or I do not understand the case.

src/Storages/MergeTree/MergeTreeSink.cpp
```diff
@@ -68,6 +70,9 @@ MergeTreeSink::MergeTreeSink(
     , context(context_)
     , storage_snapshot(storage.getStorageSnapshotWithoutData(metadata_snapshot, context_))
     , deduplicate((*storage.getSettings())[MergeTreeSetting::non_replicated_deduplication_window] > 0 && storage.getDeduplicationLog() != nullptr)
+    , synchronously_commit_part_for_dependent_views(
+        !DatabaseCatalog::instance().getDependentViews(storage_.getStorageID()).empty()
```
CheSema: We have one point where we resolve the dependencies, it is better to calculate this predicate there and here pass only a flag. We definitely already calculated it there.

src/Storages/MergeTree/MergeTreeSink.cpp
CheSema: No still `wait_for_part_commit_in_dependent_materialized_views=1` won't provide prefix commit guarantees.
If your insert several blocks they are handled by different instances of sync processors which work in parallel. Each parallel branch would have its own prefix. All together they would not have it.

prefix commit guarantees is only reachable here when insert works in one thread.

src/Storages/MergeTree/MergeTreeSink.cpp
CheSema: It make sense to make private part of that changes in the sync PR. I usually do it in that way.

src/Storages/MergeTree/MergeTreeSink.cpp
ahmadov: > No still `wait_for_part_commit_in_dependent_materialized_views=1` won't provide prefix commit guarantees. If you inserts several blocks they are handled by different instances of sync processors in parallel. Each branch would have its own prefix. prefix commit guarantees is only reachable here when insert works in one thread.

what solution would you recommend? Should we force to a single thread insert when the setting is enabled or just document the limitation?

CheSema: We do not have a goal to make prefix commit guarantee here. We just have to make sure that the settings description does not imply such promise.

antaljanosbenjamin: > The data might either be commited in the dest table in that case or not.

Based on our [docs](https://clickhouse.com/docs/operations/settings/settings#materialized_views_ignore_errors) (which might be misunderstood by me) if we have an insert chain `insert -> table -> mv -> mv_table` and `materialized_views_ignore_errors = 0`, then if the insertion into `mv_table` fails, the block won't be inserted into `table` neither. With this change I think this behavior is changed.

CheSema: That ineresting that you have such impression from the doc.
Could you please point to the exact part there?

I see there
```
Allows to ignore errors for MATERIALIZED VIEW, and deliver original block to the table regardless of MVs
```
And it says nothing about what would be inserted in target table in case of an error in MV part.

antaljanosbenjamin: "Allows" suggests that without that it is "prohibited", so that makes the impression for me that if `ignore_errors = 0` the original block won't be delivered.

If it would "ensure" that the block is delivered, then I would say it doesn't say anything about what happens if the errors are not ignored.

CheSema: I see, we better to alter that description then.

CheSema: More symboles in the description than it was, but it describes it with complete context.

https://github.com/ClickHouse/ClickHouse/pull/106044

CheSema (commented):
OK from me, but I do not see the private part in sync PR

ahmadov:
> OK from me, but I do not see the private part in sync PR

it should be done now.

CheSema (approved):

PR #105961, opened 2026-05-27
Kernel Helper fixes

PedroTadim:

1. If the Rust kernel builder configuration fails, consume it, then it won't leak. Fixes the CI leak: https://d1k2gkhrlfqv31.cloudfront.net/clickhouse-test-reports-private/json.html?PR=59140&sha=latest&name_0=PR&name_1=Stateless+tests+%28amd_asan_ubsan%2C+distributed+cache%2C+meta+in+keeper%2C+s3+storage%2C+parallel%2C+1%2F2%29 

2. For DeltaLake local tables, append a slash only if the path doesn't end with one already.

cc @Algunenano or @kssenii can you review?

### DIFF
--- a/src/Storages/ObjectStorage/DataLakes/DeltaLake/KernelHelper.cpp
+++ b/src/Storages/ObjectStorage/DataLakes/DeltaLake/KernelHelper.cpp
@@ -68,7 +68,14 @@ ffi::EngineBuilder * IKernelHelper::createBuilder() const
     catch (...)
     {
         /// The Rust FFI has no free_engine_builder; builder_build consumes it.
-        ffi::builder_build(builder);
+        /// The returned ExternResult owns either an engine handle (Ok) or an
+        /// error allocation (Err) — both must be released to avoid leaks.
+        try
+        {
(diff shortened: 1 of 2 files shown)


Algunenano:
Nice. I was also looking at a similar report right now too

Algunenano:
I'm reverting instead. Otherwise it's impossible to backport, and it's already causing flakiness

PedroTadim:
> I'm reverting instead. Otherwise it's impossible to backport, and it's already causing flakiness

Can we merge this first? Then revert the revert?

Algunenano:
> Can we merge this first? Then revert the revert?

No. This needs to be reviewed carefully, then if we merge this first we'll need to do double the work to backport. Reverting and reintroducing correctly saves us work, and in the meantime we are not blocking CI

PedroTadim:
> > Can we merge this first? Then revert the revert?
> 
> No. This needs to be reviewed carefully, then if we merge this first we'll need to do double the work to backport. Reverting and reintroducing correctly saves us work, and in the meantime we are not blocking CI

Sure, after tests and reviews pass :smile:

PR #105982, opened 2026-05-27
[#76518] Fix the input table function hint when the column is used twice in insert select

### LINKED ISSUE #76518: INSERT SELECT fails during schema inferece if the same identifier is used in projection list
### Company or project name

_No response_

### Describe what's wrong

`INSERT SELECT` fails if it projects the same identifier several times in the `SELECT` part.

**Note:** It doesn't work with `enable_analyzer` set to `0` and `1`. I added `analyzer` tag for visibility.

Yes

### How to reproduce

```sql
CREATE TABLE IF NOT EXISTS http (
    x Int64,
    y Int64
) ENGINE = MergeTree() ORDER BY ();
```

```sql
INSERT INTO http (
    x,
    y
) SELECT 
    size as `x`,
    size as `y`
FROM s3Cluster(...)
LIMIT 1;
```

### Expected behavior

`INSERT SELECT` succeeds.

Exception:
```
DB::Exception: Cannot add column size: column with this name already exists. (ILLEGAL_COLUMN) (version 25.1.5.31 (official build))
```

Stacktrace:
```
0. DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0x000000000dd109bb
1. DB::Exception::Exception(PreformattedMessage&&, int) @ 0x0000000008bae62c
2. DB::Exception::Exception<String&>(int, FormatStringHelperImpl<std::type_identity<String&>::type>, String&) @ 0x0000000008bc5acb
3. DB::ColumnsDescription::add(DB::ColumnDescription, String const&, bool, bool) @ 0x0000000013101ab9
4. DB::QueryAnalyzer::resolveTableFunction(std::shared_ptr<DB::IQueryTreeNode>&, DB::IdentifierResolveScope&, DB::QueryExpressionsAliasVisitor&, bool) @ 0x0000000011efb12d
5. DB::QueryAnalyzer::resolveQueryJoinTreeNode(std::shared_ptr<DB::IQueryTreeNode>&, DB::IdentifierResolveScope&, DB::QueryExpressionsAliasVisitor&) @ 0x0000000011f281ba
6. DB::QueryAnalyzer::resolveQuery(std::shared_ptr<DB::IQueryTreeNode> const&, DB::IdentifierResolveScope&) @ 0x0000000011eec246
7. DB::QueryAnalyzer::resolve(std::shared_ptr<DB::IQueryTreeNode>&, std::shared_ptr<DB::IQueryTreeNode> const&, std::shared_ptr<DB::Context const>) @ 0x0000000011eea6ac
8. DB::QueryAnalysisPass::run(std::shared_ptr<DB::IQueryTreeNode>&, std::shared_ptr<DB::Context const>) @ 0x0000000011ee9f80
9. DB::QueryTreePassManager::run(std::shared_ptr<DB::IQueryTreeNode>) @ 0x00000000125a6d76
10. DB::(anonymous namespace)::buildQueryTreeAndRunPasses(std::shared_ptr<DB::IAST> const&, DB::SelectQueryOptions const&, std::shared_ptr<DB::Context const> const&, std::shared_ptr<DB::IStorage> const&) (.llvm.1070560979882664992) @ 0x0000000012806d9d
11. DB::InterpreterSelectQueryAnalyzer::InterpreterSelectQueryAnalyzer(std::shared_ptr<DB::IAST> const&, std::shared_ptr<DB::Context const> const&, DB::SelectQueryOptions const&, std::vector<String, std::allocator<String>> const&) @ 0x0000000012804f7c
12. DB::InterpreterInsertQuery::buildInsertSelectPipeline(DB::ASTInsertQuery&, std::shared_ptr<DB::IStorage>) @ 0x00000000127ab5c8
13. DB::InterpreterInsertQuery::execute() @ 0x00000000127ae67f
14. DB::executeQueryImpl(char const*, char const*, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum, DB::ReadBuffer*) @ 0x0000000012b95cb4
15. DB::executeQuery(String const&, std::shared_ptr<DB::Context>, DB::QueryFlags, DB::QueryProcessingStage::Enum) @ 0x0000000012b90c9c
16. DB::TCPHandler::runImpl() @ 0x0000000013e86cac
17. DB::TCPHandler::run() @ 0x0000000013ea4839
18. Poco::Net::TCPServerConnection::start() @ 0x0000000017461687
19. Poco::Net::TCPServerDispatcher::run() @ 0x0000000017461ad9
20. Poco::PooledThread::run() @ 0x000000001742e05b
21. Poco::ThreadImpl::runnableEntry(void*) @ 0x000000001742c53d
22. ? @ 0x0000776ffd8cfac3
23. ? @ 0x0000776ffd960a04
```

### Additional context

Workaround:

```sql
INSERT INTO http (
    x,
    y
) SELECT 
    size as `x`,
    size + 0 as `y`
FROM s3Cluster(...)
LIMIT 1;
```

crakjie:
Closes: https://github.com/ClickHouse/ClickHouse/issues/76518

`INSERT ... SELECT` from a table function that infers its schema (`file`, `url`, `s3`, `s3Cluster`, `input`, ...) failed when the same source column was selected more than once:

```sql
INSERT INTO http (x, y) SELECT size AS x, size AS y FROM s3Cluster(...);
-- Cannot add column size: column with this name already exists. (ILLEGAL_COLUMN)
```

The structure of the table function is guessed from the insertion table by mapping every `SELECT` expression to the insertion table column at the same position. When the same identifier is selected twice, the same column was added to the hint twice, which is not allowed.

Now the column is added to the hint only once. A source column has a single type, so when the insertion table columns it is mapped to disagree on the type, the hint is ambiguous: it is dropped and the structure is inferred from the data instead, the same way it already happens when an asterisk is mixed with a column list. With `use_structure_from_insertion_table_in_table_functions = 1`, where the hint is mandatory, the ambiguity is reported as an error.

Both the analyzer (`QueryAnalyzer::resolveTableFunction`) and the old query interpretation path (`Context::executeTableFunction`) are fixed, because the query failed with `enable_analyzer` set to both `0` and `1`.

Fixed the error `Cannot add column ...: column with this name already exists` on `INSERT SELECT` from a table function such as `file`, `s3` or `input`, when the same source column is selected more than once.

---
Workflow [[PR](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105982&sha=latest&name_0=PR)]
Sync PR [[sync-upstream/pr/105982](https://github.com/search?q=head%3Async-upstream%2Fpr%2F105982+org%3AClickHouse+type%3Apr&type=pullrequests)]


alexey-milovidov:
Thank you for the fix! I pushed a follow-up commit on top of it.

Keeping the type of the first occurrence is not always right. When the insertion table columns that the repeated source column is mapped to have different types, the hint is ambiguous: for a table `(x Int64, y Float64)` and `INSERT INTO t (x, y) SELECT c, c FROM file(...)` over `{"c": 1.5}`, the first-occurrence type makes `c` be read as `Int64` and the query fails to parse the value, while plain schema inference handles it.

Now such a hint is dropped and the structure is inferred from the data instead, the same way it already happens when an asterisk is mixed with a column list. With `use_structure_from_insertion_table_in_table_functions = 1`, where the hint is mandatory, the ambiguity is reported as an error. This cannot regress anything, because today every duplicated identifier throws.

I also added a test for the `file` table function, and moved `Closes:` out of the HTML comment in the description so that the issue is closed automatically.

alexey-milovidov (approved):
Nice.

PR #105991, opened 2026-05-27
Batch ZooKeeper multi-requests in clearBlocksInPartition

cmcclure-twilio:
`clearBlocksInPartition` removed all deduplication block entries for a partition in a single ZooKeeper `multi` request. For tables with a large `replicated_deduplication_window`, that request could exceed ZooKeeper's `jute.maxbuffer` limit (1MB by default), causing `TRUNCATE TABLE` and `DROP PARTITION` to fail.

The removals are now split into batches of `zkutil::MULTI_BATCH_SIZE` (100) operations, with up to 16 batches in flight at once, consistent with other bulk-removal paths in the codebase such as `removeChildren` and `tryRemoveChildrenRecursive`. Non-user (hardware) ZooKeeper errors from a batch are propagated instead of being silently ignored, preserving the failure semantics of the original `tryMulti`-based implementation.

- Improvement

Only if you used Apache ZooKeeper (not recommended) instead of ClickHouse Keeper. Fix `TRUNCATE TABLE` and `DROP PARTITION` failing on tables with many deduplication blocks, where removing them in a single ZooKeeper request could exceed the default 1MB `jute.maxbuffer` limit.

### Version info
- Merged into: `26.7.1.485` (included in `26.7` and later)


src/Storages/StorageReplicatedMergeTree.cpp
```diff
@@ -8893,20 +8893,31 @@ void StorageReplicatedMergeTree::clearBlocksInPartition(
...
-    Coordination::Responses delete_responses;
-    auto code = zookeeper.tryMulti(delete_requests, delete_responses);
-    if (code != Coordination::Error::ZOK)
+
+    size_t total_removed = delete_requests.size();
+
+    /// Send removals in batches to avoid exceeding ZooKeeper's maximum message size.
+    /// Without batching, tables with large deduplication windows can produce multi-requests
+    /// exceeding 1MB (the default jute.maxbuffer), causing the operation to fail.
+    for (size_t batch_start = 0; batch_start < delete_requests.size(); batch_start += zkutil::MULTI_BATCH_SIZE)
```
Michicosun: This will basically execute all removal requests sequentially. Let's introduce some parallelism here - for example, 16 removal batches can be simultaneously in-flight.

Michicosun (changes requested):

src/Storages/StorageReplicatedMergeTree.cpp
cmcclure-twilio: Done. `clearBlocksInPartition` now keeps up to 16 batches in flight instead of sending them one at a time.

It issues each `MULTI_BATCH_SIZE`-sized batch via `asyncTryMultiNoThrow` and only blocks on the oldest in-flight future once the window of 16 is full, then drains the rest at the end. So a partition with many deduplication blocks is cleared with overlapping round-trips rather than a long strictly-sequential chain.

See commit `71d9ebe`.

tests/queries/0_stateless/4061_truncate_many_dedup_blocks.sql
```diff
@@ -0,0 +1,35 @@
...
+    SETTINGS replicated_deduplication_window = 10000;
+
+CREATE TABLE truncate_many_dedup2 (k UInt64)
+    ENGINE = ReplicatedMergeTree('/clickhouse/tables/{database}/test_04061/truncate_many_dedup', 'r2')
+    ORDER BY k
+    SETTINGS replicated_deduplication_window = 10000;
+
+-- Insert many individual blocks to create many deduplication entries in ZooKeeper.
+-- Each INSERT creates a separate dedup block entry.
+INSERT INTO truncate_many_dedup1 SELECT number FROM numbers(200) SETTINGS max_insert_block_size = 1, min_insert_block_size_rows = 1;
```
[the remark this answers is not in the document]
cmcclure-twilio: Addressed. The test now asserts batching deterministically instead of only checking row counts.

It was converted to a `.sh` test (`4061_truncate_many_dedup_blocks.sh`) that issues 120 separate `INSERT` statements. Each `INSERT` creates exactly one deduplication block node in ZooKeeper, so this exceeds `zkutil::MULTI_BATCH_SIZE` (100) and forces a correct implementation to split the removals into more than one `multi` request. (A single `INSERT ... SELECT FROM numbers(200)` is squashed server-side into one part and therefore one dedup node, which is why the previous row-count-only check didn't actually exercise batching.)

After the `TRUNCATE`, the test reads `system.zookeeper_log`, groups the `Remove` sub-requests for the table's `blocks`/`async_blocks`/`deduplication_hashes` nodes by `(session_id, xid)` (every sub-request of a `multi` shares the multi's `xid`), and asserts that no group exceeds 100 and that there is more than one group. The pre-fix code sends a single oversized `multi` and prints `0 0` (fails); the fixed code prints `1 1`. This makes the assertion independent of `jute.maxbuffer`.

See commits `71d9ebe` and `43dd935`.

src/Storages/StorageReplicatedMergeTree.cpp
```diff
@@ -8893,20 +8895,65 @@ void StorageReplicatedMergeTree::clearBlocksInPartition(
...
+
+            for (size_t i = 0; i < response.responses.size(); ++i)
+                if (response.responses[i]->error != Coordination::Error::ZOK)
+                    LOG_WARNING(log, "Error while deleting ZooKeeper path `{}`: {}, ignoring.",
+                                delete_requests[batch.batch_start + i]->getPath(), response.responses[i]->error);
+        }
+        in_flight.pop_front();
+    };
+
+    for (size_t batch_start = 0; batch_start < delete_requests.size(); batch_start += zkutil::MULTI_BATCH_SIZE)
```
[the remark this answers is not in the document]
cmcclure-twilio: Thanks — this is a fair observation. There are three other direct callers of `getClearBlocksInPartitionOps` that still send the removals in a single `tryMulti`: `dropPartImpl` (`StorageReplicatedMergeTree.cpp:9942`) and the two `PartMovesBetweenShardsOrchestrator` paths (`:435`, `:581`).

I'd prefer to keep this PR scoped to the `TRUNCATE TABLE` / `DROP PARTITION` path and handle those separately, because they are structurally different from `clearBlocksInPartition`:

- Each one commits the block removals **atomically together** with a log-version `Check` + a log-entry `Create` + a log `Set`, and retries the whole `multi` on `ZBADVERSION`. Splitting the removals into independent batches there would break that atomicity, so it needs its own design (e.g. pre-committing the removals in batches before the atomic log-entry commit) rather than a mechanical reuse of the batched helper.
- They also operate on a single-part range rather than a full-partition deduplication window, so the op count is bounded by the part's block range and is much smaller in practice than the `TRUNCATE`/`DROP PARTITION` case this PR targets.

I'll open a follow-up issue/PR to give those paths a bounded cleanup path. Let me know if you'd rather fold it in here.

cmcclure-twilio:
@Michicosun thanks for the review — both points are addressed:

1. **Deterministic batching assertion.** The test is now `4061_truncate_many_dedup_blocks.sh`, issuing 120 separate `INSERT`s (one dedup block node each, > `MULTI_BATCH_SIZE` = 100) and asserting via `system.zookeeper_log` that the `Remove` sub-requests were split across more than one `multi`, with no `multi` exceeding 100 ops. Pre-fix prints `0 0` (fails), fixed prints `1 1`, independent of `jute.maxbuffer`. (commits `71d9ebe`, `43dd935`)
2. **Parallelism.** `clearBlocksInPartition` now keeps up to 16 batches in flight via `asyncTryMultiNoThrow` instead of running them strictly sequentially. (commit `71d9ebe`)

I also restored the original failure contract after the async switch (re-throw on non-user/hardware Keeper errors) in `52dd789`.

Could you take another look when you have a chance?

### On the failing CI

- **`Stateless tests (arm_asan_ubsan, azure, sequential)` — unrelated flaky crash.** The segfault is in the object-storage metadata write path (`MetadataStorageFromDiskTransactionOperations` → `WriteBufferFromFile` dtor → ASan `free`) on a background INSERT/merge thread (`(no query)`), on the azure+ASan build. The new test `4061_truncate_many_dedup_blocks` does not appear anywhere in that shard's report or server log, and it isn't scheduled in the sequential shard at all. This looks like infra/flaky and should clear on a re-run.
- **AI Review — other dedup-cleanup callers.** Acknowledged in the inline thread; proposing to keep this PR scoped to `TRUNCATE`/`DROP PARTITION` and handle `dropPartImpl` / shard-move paths in a follow-up, since those commit removals atomically with their log entry and need a separate design.

src/Storages/StorageReplicatedMergeTree.cpp
```diff
@@ -8893,20 +8895,65 @@ void StorageReplicatedMergeTree::clearBlocksInPartition(
...
-            if (delete_responses[i]->error != Coordination::Error::ZOK)
-                LOG_WARNING(log, "Error while deleting ZooKeeper path `{}`: {}, ignoring.", delete_requests[i]->getPath(), delete_responses[i]->error);
+        auto & batch = in_flight.front();
+        auto response = batch.future.get();
+        if (response.error != Coordination::Error::ZOK)
+        {
+            /// Preserve the failure contract of the original tryMulti-based implementation:
+            /// non-user (hardware) errors such as ZCONNECTIONLOSS, ZSESSIONEXPIRED or ZOPERATIONTIMEOUT
+            /// must propagate so the caller does not proceed as if the blocks were removed. Only expected
+            /// per-op user errors (e.g. ZNONODE for an already-removed block) are logged and ignored.
```
Michicosun:
```suggestion
```

src/Storages/StorageReplicatedMergeTree.cpp
```diff
@@ -8893,20 +8895,65 @@ void StorageReplicatedMergeTree::clearBlocksInPartition(
...
+            if (!Coordination::isUserError(response.error))
+                throw zkutil::KeeperException(response.error);
+
+            for (size_t i = 0; i < response.responses.size(); ++i)
+                if (response.responses[i]->error != Coordination::Error::ZOK)
+                    LOG_WARNING(log, "Error while deleting ZooKeeper path `{}`: {}, ignoring.",
+                                delete_requests[batch.batch_start + i]->getPath(), response.responses[i]->error);
```
Michicosun: Only the first failed operation in a multi-request will have a valid error code; all others will have `ZRUNTIMEINCONSISTENCY`. 

Let's use `getFailedOpIndex(Coordination::Error code, const Coordination::Responses & responses)` and print only the first error here.

src/Storages/StorageReplicatedMergeTree.cpp
```diff
@@ -8893,20 +8895,65 @@ void StorageReplicatedMergeTree::clearBlocksInPartition(
...
+
+    const size_t total_removed = delete_requests.size();
+
+    /// Send removals in batches to avoid exceeding ZooKeeper's maximum message size.
+    /// Without batching, tables with large deduplication windows can produce multi-requests
+    /// exceeding 1MB (the default jute.maxbuffer), causing the operation to fail.
+    ///
+    /// Keep several batches in flight at once instead of waiting for each one to complete
+    /// before sending the next, so that partitions with many deduplication blocks are cleared
+    /// without a long sequence of round-trips to ZooKeeper.
```
Michicosun:
```suggestion
    /// Send removals in batches to avoid exceeding ZooKeeper's maximum message size.
```

src/Storages/StorageReplicatedMergeTree.cpp
cmcclure-twilio: Good catch — fixed in `7a69e5e`. Switched to `zkutil::getFailedOpIndex(response.error, response.responses)` and now log only the first failed op's path and error, instead of iterating every sub-response (the rest carry `ZRUNTIMEINCONSISTENCY`).

src/Storages/StorageReplicatedMergeTree.cpp
cmcclure-twilio: Done in `7a69e5e` — removed this comment. The behavior is now self-evident from the `isUserError`/`getFailedOpIndex` handling.

src/Storages/StorageReplicatedMergeTree.cpp
cmcclure-twilio: Done in `7a69e5e` — trimmed to the single line as suggested.

cmcclure-twilio:
@Michicosun thanks for the review — all three points addressed in `7a69e5e`:

1. **Log only the first failed op.** Now uses `zkutil::getFailedOpIndex(response.error, response.responses)` and logs just that op's path/error, instead of iterating all sub-responses (the rest carry `ZRUNTIMEINCONSISTENCY`).
2. Removed the failure-contract comment.
3. Trimmed the batching comment to a single line.

The failure contract is unchanged: non-user (hardware) errors still propagate via `KeeperException`; expected per-op user errors (e.g. `ZNONODE`) are logged and ignored. Could you take another look when you have a chance?

src/Storages/StorageReplicatedMergeTree.cpp
```diff
@@ -8893,20 +8895,56 @@ void StorageReplicatedMergeTree::clearBlocksInPartition(
 {
     Coordination::Requests delete_requests;
     getClearBlocksInPartitionOps(delete_requests, zookeeper, partition_id, min_block_num, max_block_num);
-    Coordination::Responses delete_responses;
-    auto code = zookeeper.tryMulti(delete_requests, delete_responses);
-    if (code != Coordination::Error::ZOK)
+
+    const size_t total_removed = delete_requests.size();
```
Michicosun: Let's track how many delete requests are completed successfully here. Instead of reporting the full count, it also makes sense to print how many requests failed.

tests/queries/0_stateless/4061_truncate_many_dedup_blocks.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# Each separate INSERT becomes its own deduplication block entry in ZooKeeper.
+# 120 inserts comfortably exceeds zkutil::MULTI_BATCH_SIZE (100), so the removals
+# during TRUNCATE must be split into more than one multi-request.
+NUM_INSERTS=120
+{
+    for i in $(seq 1 $NUM_INSERTS); do
+        echo "INSERT INTO truncate_many_dedup1 VALUES ($i);"
+    done
+} | $CLICKHOUSE_CLIENT --multiquery
```
Michicosun: let's use here

```
INSERT INTO truncate_many_dedup1 SELECT number FROM numbers(120) SETTINGS min_insert_block_size_rows=1, max_block_size=1;
```

Michicosun (commented):
Looks good!

Michicosun (approved):

alexey-milovidov:
The test does not reproduce the bug with ClickHouse Keeper.
It looks only relevant to ZooKeeper. I will change the category to "Improvement".

clickgapai:

### TL;DR
**Batch ZooKeeper multi-requests in clearBlocksInPartition**

**ClickGap verified:** reproduced it with the script below on `26.5`; tested 6 version(s) — 5 affected, 1 clean (see Affects).

### Reproduction
```
On a Keeper/ZooKeeper-backed ClickHouse server:
CREATE TABLE t (k UInt64) ENGINE = ReplicatedMergeTree('/clickhouse/tables/t','r1') ORDER BY k SETTINGS replicated_deduplication_window = 10000;
-- Run >100 SEPARATE inserts so >100 deduplication-block nodes exist in ZooKeeper:
INSERT INTO t VALUES (1); INSERT INTO t VALUES (2); /* ... repeat to >=150 ... */
TRUNCATE TABLE t SETTINGS replication_alter_partitions_sync = 2;
SYSTEM FLUSH LOGS zookeeper_log;
SELECT count() AS multis, max(c) AS max_removes_per_multi
FROM (SELECT count() AS c FROM system.zookeeper_log
      WHERE type='Request' AND op_num='Remove'
        AND match(path, '/clickhouse/tables/t/(blocks|async_blocks|deduplication_hashes)/')
      GROUP BY session_id, xid);
-- Pre-fix: 1 multi with >100 Remove ops -> can exceed jute.maxbuffer (1MB) and TRUNCATE/DROP PARTITION fails.
-- Fixed:   removals split into several multis of <=100 ops each.
```

### Bisect
- Bisected on master in several steps (confidence: MEDIUM); a single introducing PR could not be pinned (culprit confirmation was inconclusive).

### Affects

### Suggested next step

**Component:** `comp-replication`
**Severity:** `P2`

cc fix PR reviewer: @Michicosun

---
_ClickGap · Finding: `phase_d_pr105991`_




### Triage metadata
- Primary component: `comp-replication`
- Component owners (`comp-replication`): @alexey-milovidov @CheSema

PR #106000, opened 2026-05-27
Fix TSan data race in Parquet reader

azat:

Fix TSan data race in Parquet reader

Parquet::ReadManager::runBatchOfTasks() can mutate exception, while Prefetcher::runTask() may rethrow the same exception, and so runBatchOfTasks() may have a data-race

TSan report - https://pastila.nl/?003169db/e279818b622dcecc6ea34bdbbecc2504#9p8Zaqb6CrjGOd9QAF8JjA==GCM

### Version info
- Merged into: `26.6.1.261`


Algunenano (approved):
Looks good. I was thinking we might need to use this in some other places, because I remember we had to fix something like this in the past, but it looks like at least fibers don't need it right now. So it's all fine

alexey-milovidov (approved):

PR #106004, opened 2026-05-27
Fix ALTER TABLE partition operations silently failing for Bool partition keys

Avogar:

Fix ALTER TABLE partition operations silently failing for Bool partition keys. Closes https://github.com/ClickHouse/ClickHouse/issues/101722

### Version info
- Merged into: `26.6.1.570`


Diskein (approved):
Claude says that we could fix it in one place, but i'm not sure that's better because we are going to do unnecessary lookups.
   ```cpp
   explicit MergeTreePartition(Row value_) : value(std::move(value_))
   {
       /// Bool partition values are stored as UInt64 Fields (as INSERT/LOAD produce) so getID stays consistent.
       for (Field & field : value)
           if (field.getType() == Field::Types::Bool)
               field = field.safeGet<UInt64>();
   }
   ```

PR #106009, opened 2026-05-27
Fix JSON data misdetected as TSKV during format auto-detection

Avogar:

Fix JSON data misdetected as TSKV during format auto-detection. Closes https://github.com/ClickHouse/ClickHouse/issues/100797

### Version info
- Merged into: `26.7.1.67`


src/Processors/Formats/Impl/TSKVRowInputFormat.cpp
```diff
@@ -268,6 +268,8 @@ NamesAndTypesList TSKVSchemaReader::readRowAndGetNamesAndDataTypes(bool & eof)
     {
         bool has_value = readName(in, name_ref, name_buf);
         String name = String(name_ref);
+        if (name_ref.size() > 256)
```
george-larionov: This check should probably go before the previous line, otherwise we do a potentially wasteful copy. Also maybe worth adding a comment on why 256 was chosen.

src/Formats/ReadSchemaUtils.cpp
```diff
@@ -77,12 +77,12 @@ static const std::vector<String> & getFormatsOrderForDetection()
         "BSONEachRow",
         "JSONCompact",
         "Values",
-        "TSKV",
         "JSONObjectEachRow",
         "JSONColumns",
         "JSONCompactColumns",
         "JSONCompact",
```
george-larionov, line 82: Not changed here, but why is JSONCompact in the list twice?

src/Processors/Formats/Impl/TSKVRowInputFormat.cpp
george-larionov: Also, technically this limit is backwards incompatible in case someone is processing TSKV data with names longer than 256 bytes, are we sure we need it?

george-larionov (commented):
The change in ReadSchemaUtils.cpp fixes the issue but I have left some comments regarding other parts of the code.

alexey-milovidov:
The test didn't reproduce the bug.

alexey-milovidov:

george-larionov (approved):
LGTM.

PR #106016, opened 2026-05-27
Keep IcebergLocal writable after DETACH/ATTACH or server restart

### LINKED ISSUE #105899: SHOW ROW POLICIES does not support FORMAT clause
### Bug description

`SHOW ROW POLICIES` (and `SHOW MASKING POLICIES`) silently rejects the `FORMAT` clause and returns a syntax error, while other `SHOW` commands (e.g. `SHOW ROLES`, `SHOW DATABASES`) handle it correctly.

### How to reproduce

```sql
SHOW ROW POLICIES FORMAT TabSeparated;
```

**Error:**
```
Syntax error: failed at position 26 (TabSeparated):

SHOW ROW POLICIES Format TabSeparated

Expected one of: INTO OUTFILE, FORMAT, SETTINGS, ParallelWithClause, PARALLEL WITH, end of query
```

### Version

Reproducible on current `main`.

### Version info
- Resolved by: #107061
- Merged into: `26.6.1.837`

groeneai:
Fixes a regression where an `IcebergLocal` table became read-only after a `DETACH` + `ATTACH` cycle or a full server restart, with every `INSERT` failing as:

```
Code: 164. DB::Exception: Local object storage `Local` is readonly. (READONLY)
```

When reloading the table, `registerStorageObjectStorage` calls `createObjectStorage(context, /* is_readonly */ args.mode != LoadingStrictnessLevel::CREATE)` to suppress remote-side initialization side effects (for example, creating a container on Azure). The S3 and HDFS configurations already ignore this parameter; `StorageLocalConfiguration` was the outlier and propagated it into `LocalObjectStorageSettings::read_only`, which permanently routed every subsequent write through `throwIfReadonly`. The local filesystem has no analogous initialization side effect to suppress, so the safe behaviour is to ignore the flag (matching S3 and HDFS) and always construct a writable `LocalObjectStorage` for the `IcebergLocal` engine.

The disk-config based `<readonly>true</readonly>` path in `ObjectStorageFactory::registerLocalObjectStorage` is unaffected, so legitimate read-only `LocalObjectStorage` disks keep their read-only behaviour.

Reproducer from the issue:

```sql
SET allow_insert_into_iceberg = 1;
CREATE TABLE t0 (c0 Int) ENGINE = IcebergLocal('/path/to/lakehouse/t0');
INSERT INTO t0 VALUES (1);     -- ok
DETACH TABLE t0;
ATTACH TABLE t0;
INSERT INTO t0 VALUES (2);     -- before: READONLY exception; after: succeeds
```

Closes #85413.

Fix `IcebergLocal` table engine becoming read-only after a `DETACH` + `ATTACH` cycle or a server restart, which made every subsequent `INSERT` fail with `Local object storage Local is readonly. (READONLY)`.

### Version info
- Merged into: `26.6.1.254`


groeneai:
**Pre-PR validation gate**

a) **Deterministic repro?** Yes. `CREATE TABLE t0 (c0 Int) ENGINE = IcebergLocal(...); INSERT VALUES (1); DETACH TABLE t0; ATTACH TABLE t0; INSERT VALUES (2)` throws `Code: 164 Local object storage Local is readonly. (READONLY)` 100% of the time on master before the fix.
b) **Root cause explained?** Yes. `registerStorageObjectStorage` passes `is_readonly = (args.mode != LoadingStrictnessLevel::CREATE)` to `createObjectStorage` so that `ATTACH` (and server restart) skip remote-side initialization side effects such as Azure container creation. `StorageLocalConfiguration::createObjectStorage` was propagating that flag into the persistent `LocalObjectStorageSettings::read_only` field, after which every write went through `throwIfReadonly` and refused the operation. The intent was "skip init side effects", never "make this object storage permanently read-only".
c) **Fix matches root cause?** Yes. `StorageLocalConfiguration::createObjectStorage` now ignores the parameter and always constructs a writable `LocalObjectStorage`, matching `StorageS3Configuration::createObjectStorage` and `StorageHDFSConfiguration::createObjectStorage`. The disk-config driven `<readonly>true</readonly>` path in `ObjectStorageFactory::registerLocalObjectStorage` is unaffected, so legitimate read-only local disks keep their behaviour.
d) **Test intent preserved?** Yes, new test added. `04299_85413_iceberg_local_readonly_after_attach.sh` exercises `CREATE IcebergLocal -> INSERT -> DETACH -> ATTACH -> INSERT -> SELECT count() = 2`, which deterministically fails on the buggy code path and passes with the fix.
e) **Both directions demonstrated?** Yes. Against the pre-fix binary (the existing running server PID 25730 on port 9019, built before the fix) the new test fails with `READONLY` on the second INSERT. Against a freshly built server with the fix on port 9241 the test passes 20/20 runs in `tests/clickhouse-test --test-runs 20`.
f) **Fix is general?** Yes. The same bug pattern in S3 and HDFS is already absent there because their `createObjectStorage` ignores `is_readonly`. Azure uses it only for client construction. Local was the lone outlier. The fix is at the root cause site (`StorageLocalConfiguration::createObjectStorage`) rather than guarding the symptom in `throwIfReadonly`.

session: cron:clickhouse-ci-task-worker:20260527-213000

groeneai:
cc @scanhex12 @alesapin for review. Issue #85413 is assigned to @scanhex12; @alesapin owns the Iceberg area per the recent ownership decision. Fix is small (one branch in `StorageLocalConfiguration::createObjectStorage`) and brings local Iceberg behaviour in line with S3 and HDFS.

tiandiwonder (approved):

PR #105897, opened 2026-05-26
Validate UTF-8 of DeltaLake delta-kernel-rs FFI option values

groeneai:
The `delta-kernel-rs` Rust FFI decodes `KernelStringSlice` payloads as `&str`
via `std::str::from_utf8` and panics on invalid UTF-8. A panic that crosses the
`extern "C"` boundary aborts the server through `panic_in_cleanup`. Binary
credentials (for example raw `MD5(...)` bytes used as a SAS token) trip this
and abort a debug or sanitizer build.

`SELECT * FROM deltaLakeAzure('\0', MD5('')) SETTINGS allow_experimental_delta_kernel_rs = 1`
deterministically reproduces the abort. The stack trace shows
`set_builder_option @ ffi/src/lib.rs:488` reached from the `set_option` lambda
in `AzureKernelHelper::createBuilder`, with the offending payload being the
`azure_storage_sas_key` value populated from `MD5('')`.

The fix wraps `ffi::set_builder_option` with a C++ side check that runs the
value bytes through `DB::UTF8::isValidUTF8` first and throws `BAD_ARGUMENTS`
when the bytes are not valid UTF-8. The lambdas in `S3KernelHelper` and
`AzureKernelHelper` route through the wrapper so any user supplied credential,
region, endpoint or SAS token is validated before reaching the FFI boundary.
`LocalKernelHelper` does not call `set_builder_option` and is not affected.

Adds regression test `04294_deltalake_azure_invalid_utf8_credentials` covering
the reproducer.

Closes #104509
Closes #105895

### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Reverted

### Version info
- Merged into: `26.6.1.176`

Issue #105895, opened 2026-05-26
Received signal 6 (internal) (STID: 4793-48f3)


PedroTadim:
_Important: This issue was automatically generated and is used by CI for matching failures. DO NOT modify the body content. DO NOT remove labels._

Test name: Received signal 6 (internal) (STID: 4793-48f3)
CI report: [Stateless tests (amd_debug, parallel)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=105550&sha=5db5c66a4c135950ddd8ac21fe895a693ae75bec&name_0=PR&name_1=Stateless%20tests%20%28amd_debug%2C%20parallel%29)
[...]

Issue #104509, opened 2026-05-10
Server aborts via Rust panic_in_cleanup when delta-kernel-rs FFI panics across extern "C" boundary (STID 4793-48f3)


groeneai:
## Summary

ClickHouse server aborts (`Received signal 6`) with the Rust runtime message
`panicked while processing panic. aborting.` (`panic_in_cleanup`) when a Rust
panic unwinds across the `extern "C"` boundary of `delta-kernel-rs` FFI
functions. The panic is unrecoverable: the compiler-emitted cleanup code at
the FFI boundary calls `panic_cannot_unwind`, which calls `panic_in_cleanup`,
which terminates the process via `abort`.

CIDB STID **`4793-48f3`**: 8 hits / 7 distinct PRs / 0 master in the last
60 days. Hit rate is increasing — 4 hits in the last 11 days vs. 4 in the
prior 49. Three distinct unrelated PRs in the last 7 days, crossing the
documented escalation threshold:
- 2026-05-10 — PR #96978 — `AST fuzzer (amd_debug, targeted)`
- 2026-05-09 — PR #104465 — `Stateless tests (amd_debug, parallel)`
- 2026-05-07 — PR #104102 — `AST fuzzer (amd_debug, targeted)`

The same STID covers two distinct FFI call sites that share the same panic
machinery:

### Failure mode 1 — `set_builder_option` (`contrib/delta-kernel-rs/ffi/src/lib.rs:488`)

```rust
pub unsafe extern "C" fn set_builder_option(
    builder: &mut EngineBuilder,
    key: KernelStringSlice,
    value: KernelStringSlice,
) {
    let key = unsafe { String::try_from_slice(&key) };
    let value = unsafe { String::try_from_slice(&value) };
    // TODO: Return ExternalError if key or value is invalid? (builder has an error allocator)
    builder.set_option(key.unwrap(), value.unwrap());
}
```

`String::try_from_slice` returns `DeltaResult<String>` — `Err` on invalid
UTF-8. The `key.unwrap()` / `value.unwrap()` panics on `Err`, and that panic
unwinds across the `extern "C"` boundary → `panic_cannot_unwind` → abort.

C++ caller chain:
```
DeltaLake::AzureKernelHelper::createBuilder() const
  (src/Storages/ObjectStorage/DataLakes/DeltaLake/KernelHelper.cpp:245)
  → set_option lambda (KernelHelper.cpp:159)
  → ffi::set_builder_option
```

[...]

PR #106018, opened 2026-05-27
Validate UTF-8 of delta-kernel-rs FFI options and fix EngineBuilder leak (re-apply #105897)

groeneai:
Re-submits the change from #105897 ("Validate UTF-8 of DeltaLake delta-kernel-rs FFI option values"), which was merged in `a07ca558f023` and then reverted in `3ab7b26f7189` / #105963.

**Why the original was reverted.** ASan flagged a leak in the `s3 storage` stateless variant: the Rust FFI does not expose `free_engine_builder`, and the new `setBuilderOptionChecked` wrapper threw `BAD_ARGUMENTS` on the invalid-UTF-8 path before `ffi::builder_build` could consume the partially configured `ffi::EngineBuilder`. The original CI leak report (per #105961):
https://d1k2gkhrlfqv31.cloudfront.net/clickhouse-test-reports-private/json.html?PR=59140&sha=latest&name_0=PR&name_1=Stateless+tests+%28amd_asan_ubsan%2C+distributed+cache%2C+meta+in+keeper%2C+s3+storage%2C+parallel%2C+1%2F2%29

**What changed.** Combines the original UTF-8 validation with the leak fix prototyped by @PedroTadim in #105961 (closed when the revert landed):

1. `IKernelHelper::createBuilder` is now a non-virtual base method. It allocates the builder via `ffi::get_engine_builder`, invokes a new protected virtual `configureBuilder` hook, and, if `configureBuilder` throws, consumes the builder via `ffi::builder_build` and frees the resulting engine via `ffi::free_engine` before rethrowing.
2. `S3KernelHelper` and `AzureKernelHelper` now override `configureBuilder` and route every option through `setBuilderOptionChecked`, which rejects invalid UTF-8 with `BAD_ARGUMENTS` so it never reaches the FFI `&str` decode that panics with `panic_in_cleanup`.
3. `LocalKernelHelper` inherits the default empty `configureBuilder` and no longer appends a second `/` when the user-provided path already ends with one (also from #105961).

**Test.** `04294_deltalake_azure_invalid_utf8_credentials` uses the deterministic reproducer `SELECT * FROM deltaLakeAzure('\0', MD5(''))` from #105895. It is kept as a `.sh` test on purpose: the unfixed binary aborts the process, so a `.sql` test with `{ serverError BAD_ARGUMENTS }` cannot distinguish a crash from a clean error and Bugfix validation would not record a clean `FAIL` on master. With `clickhouse-local` + `grep`, the unfixed binary produces empty stdout and the fixed binary prints the diagnosis, so Bugfix validation works in both directions.

cc @CheSema (your concern on the original PR about a post-approval test commit landing without a full CI run is acknowledged: this re-submission runs the new test through full CI before merge), @Algunenano (revert author), @PedroTadim (leak fix author), @kssenii, @alesapin.

Fix a server-side abort triggered by passing non-UTF-8 byte sequences (for example raw `MD5(...)` bytes used as a SAS token) to the `deltaLakeAzure` table function or `DeltaLake` engine when `allow_experimental_delta_kernel_rs = 1`. The `delta-kernel-rs` Rust FFI panicked on invalid UTF-8, crossing the `extern "C"` boundary and aborting the server through `panic_in_cleanup`. Such inputs now return `BAD_ARGUMENTS` instead. Also fixes a memory leak of the `ffi::EngineBuilder` allocation that could occur when builder configuration failed.


groeneai:
**Pre-PR validation gate** (per worker policy, separate from the PR description):

| # | Question | Status |
|---|----------|--------|
| a | Deterministic repro? | YES. `SELECT * FROM deltaLakeAzure('\0', MD5('')) SETTINGS allow_experimental_delta_kernel_rs = 1` reproduces both the original FFI panic (without the validation) and the `EngineBuilder` leak (without the cleanup). The repro is from #105895 / @PedroTadim. |
| b | Root cause explained? | YES, two-part. (1) `delta-kernel-rs/ffi::set_builder_option` decodes the `(ptr, len)` slice via `std::str::from_utf8` and `.unwrap()`s the result, so any non-UTF-8 byte sequence panics across the `extern "C"` boundary and `panic_in_cleanup` aborts the server (#104509, #105895). (2) The original fix translated the panic to `BAD_ARGUMENTS` but did not free the partially configured `ffi::EngineBuilder`; the Rust FFI does not expose `free_engine_builder` and the builder is only released by `ffi::builder_build` (which consumes it and returns an engine handle that itself must be freed via `ffi::free_engine`), so throwing before `builder_build` ran leaked the builder allocation. This was the ASan finding that caused #105963 to revert #105897. |
| c | Fix matches root cause? | YES. (1) `setBuilderOptionChecked` runs `DB::UTF8::isValidUTF8` on the value bytes before the FFI call and throws `BAD_ARGUMENTS` on invalid input; nothing reaches the Rust `&str` decode. (2) `IKernelHelper::createBuilder` now wraps `configureBuilder` in a try/catch that, on failure, calls `ffi::builder_build` (consumes the builder, the Rust `Box::from_raw` releases it regardless of success/failure) and frees the resulting engine handle with `ffi::free_engine`. |
| d | Test intent preserved? | YES. Same query, same expected behavior (return `BAD_ARGUMENTS` to the user instead of aborting). The test is intentionally `.sh` not `.sql` so Bugfix validation can distinguish a process abort (empty stdout -> clean `FAIL` on master) from a clean error (grep matches `BAD_ARGUMENTS` -> `PASS` on the fixed binary). |
| e | Both directions demonstrated? | PARTIAL. The UTF-8 validation half was already CI-validated by the original #105897 (merged after Algunenano's `APPROVE`); the leak fix half is mechanically the same as the design in #105961, which the PR author wrote against the ASan leak report. Local two-direction verification requires a `delta-kernel-rs` build (Rust nightly-2026-03-22 + `ENABLE_LIBRARIES=ON`), which I do not have ready in this worker session; full CI here is the verification path. I will not request maintainer attention to merge until both the new test and the existing ASan stateless variants are green. |
| f | Fix is general, not a narrow patch? | YES. The cleanup lives on the base `IKernelHelper::createBuilder`, so it covers `S3KernelHelper`, `AzureKernelHelper`, and (vacuously) `LocalKernelHelper`; any future helper inherits it automatically. The UTF-8 validation is in the single helper used by both S3 and Azure (`LocalKernelHelper` does not call `set_builder_option` and is unaffected). I am NOT adding a defensive check at the original crash site only - the new structure ensures every `set_builder_option` call site is guarded and every throw path through `configureBuilder` releases the builder. |

PedroTadim:
@Algunenano you give a review first before enabling testing. This looks fine for me.

Algunenano (changes requested):
This PR addresses:
*  set_builder_option panicking on invalid UTF-8
* The EngineBuilder leak from the original revert

This PR does not address:
* ffi::snapshot failures reported in #104509 as `Failure mode 2`.
* Also, it seems that the cleanup path can re-trigger the same bug
* name is not validated. Right now it's internal only, but it's one step away from being public.

I mentioned when reviewing the other PR that it was ok but we should fix this properly. That PR had to be reverted. So let's fix this properly upstream, then apply checks and proper error handling here.

@groeneai I don't think you can work with rust submodules, so I'll deal with it

groeneai:
Acknowledged, closing. Happy to help with the C++ cleanup wrapper (createBuilder non-virtual base + try/catch + builder_build then free_engine) once the upstream delta-kernel-rs fix lands, if useful.

PR #101104, opened 2026-03-29
Fix exception in `GlobalPlannerContext::createColumnIdentifier` for UNION ALL queries

alexey-milovidov:
This PR adds a regression test for an exception that used to be raised by `GlobalPlannerContext::createColumnIdentifier` for some `UNION ALL` queries:

> Column identifier `m` is already registered

In `UNION ALL` queries, all branches share the same `GlobalPlannerContext`. When two branches referenced a table expression with no alias and the same column name, `createColumnIdentifier` would throw `Column identifier is already registered`. This was originally found by the AST Fuzzer.

The underlying issue is already fixed on master by:
- https://github.com/ClickHouse/ClickHouse/pull/101048 — `Fix exception "Column identifier is already registered" in planner`
- https://github.com/ClickHouse/ClickHouse/pull/101051 — `Fix Column identifier is already registered with additional_result_filter in UNION/EXCEPT`

So this PR keeps only the regression test (no source code changes) as a guard against future regressions.

Original report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=101023&sha=c1dde15b4607b68f7bbecd414d3d65fe2dc8e088&name_0=PR&name_1=Stress%20test%20%28amd_tsan%29 (https://github.com/ClickHouse/ClickHouse/pull/101023)

### Changelog category (leave one):
- CI Fix or Improvement (changelog entry is not required)

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.145`

PR #101051, opened 2026-03-29
Fix Column identifier is already registered with additional_result_filter in UNION/EXCEPT

alexey-milovidov:
When multiple UNION/EXCEPT branches share the same `GlobalPlannerContext` and each applies `additional_result_filter`, the `addAdditionalFilterStepIfNeeded` function creates a fake `TableNode` without an alias for each branch. Both branches try to register the same bare column identifier (e.g. `a`) in the shared context, causing the second to throw a LOGICAL_ERROR exception.

Fix by giving each fake table expression a unique alias so that column identifiers become unique across branches.

Fixes https://github.com/ClickHouse/ClickHouse/issues/99931

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=100947&sha=d0c5ef2315e4d001a6d34cf2e54a9ee2b176a538&name_0=PR&name_1=Stress%20test%20%28amd_debug%29
https://github.com/ClickHouse/ClickHouse/pull/100947

### Changelog category (leave one):
- Bug Fix (user-visible misbehavior in an official stable release)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
Fix LOGICAL_ERROR exception "Column identifier is already registered" when `additional_result_filter` setting is used with UNION or EXCEPT queries.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.81`

PR #106025, opened 2026-05-28
Fix LOGICAL_ERROR "Column identifier is already registered" in ALTER UPDATE with EXISTS

### LINKED ISSUE #106649: Analyzer LOGICAL_ERROR 'Column identifier id is already registered' on ALTER/lightweight UPDATE with IN-subquery joining derived tables (regression in 26.6)
## Describe the bug

A mutation (`ALTER TABLE ... UPDATE`, and likewise a lightweight `UPDATE`) whose `WHERE` contains `IN (subquery)`, where the subquery joins **two subquery-wrapped derived tables** that each project a column named `id`, fails analysis with:

```
Code: 49. DB::Exception: Column identifier id is already registered. (LOGICAL_ERROR)
```

`LOGICAL_ERROR` signals an internal invariant violation, so this is a bug regardless of the query. The **identical subquery used in a plain `SELECT` analyzes and runs fine** — only the mutation's `WHERE` predicate triggers it.

This is a **regression in 26.6**: works on `26.5.1.111`, fails on `26.6.1.458` and on ClickHouse Cloud `26.6.1.614`.

## How to reproduce

* ClickHouse version: **26.6.1.458** (also reproduced on Cloud `26.6.1.614`); **works on 26.5.1.111**.
* Reproduces in both `clickhouse local` and `clickhouse-server`. Tables can be empty — the error is raised at analysis time, so no data is needed.

```sql
CREATE TABLE gh    (id UInt64, x UInt8) ENGINE = MergeTree ORDER BY id;
CREATE TABLE prs   (id UInt64)          ENGINE = MergeTree ORDER BY id;
CREATE TABLE edges (from_id UInt64, to_id UInt64) ENGINE = MergeTree ORDER BY tuple();

ALTER TABLE gh UPDATE x = 1 WHERE id IN (
    SELECT i.id
    FROM (SELECT id FROM gh)  AS i
    JOIN edges AS r            ON r.to_id  = i.id
    JOIN (SELECT id FROM prs) AS pr        ON pr.id = r.from_id
) SETTINGS mutations_sync = 1;
```

Actual result on 26.6:

```
Received exception:
Code: 49. DB::Exception: Column identifier id is already registered. (LOGICAL_ERROR)
```

The same predicate runs fine as a plain query:

```sql
-- returns a result, no error, on 26.6:
SELECT count() FROM (
    SELECT i.id
    FROM (SELECT id FROM gh)  AS i
    JOIN edges AS r            ON r.to_id  = i.id
    JOIN (SELECT id FROM prs) AS pr        ON pr.id = r.from_id
);
```

### What triggers it (from minimization)

* The mutation form is required: `ALTER TABLE ... UPDATE ... WHERE id IN (...)` and lightweight `UPDATE ... WHERE id IN (...)` both fail; a plain `SELECT` of the same subquery does not.
* **Both** joined sources must be subquery-wrapped derived tables that project `id`. Replacing either `(SELECT id FROM t) AS a` with a plain `t AS a` makes the error disappear.
* **Two** joins are required — a single join does not trigger it.
* Independent of `UNION ALL` views or extra columns — the minimal case above uses plain tables projecting a single `id` column.

## Expected behavior

The mutation should execute (or, if the query were genuinely invalid, fail with a normal user-facing error) — not raise a `LOGICAL_ERROR`.

Caused by: https://github.com/ClickHouse/ClickHouse/pull/98884

### Version info
- Resolved by: #106025
- Merged into: `26.7.1.732` (included in `26.7` and later)
- Backported to: `26.6.2.58`

groeneai:
Adds a regression test for the chronic AST-fuzzer `LOGICAL_ERROR` "Column identifier ... is already registered" (STID `4697-4326`) on the mutations + `EXISTS` path, and for the rejection of correlated subqueries in mutations.

The two code fixes this test guards have since landed on `master` independently, so this PR is now **test-only** (per @novikd's review: "remove the code added in #106414, keep the test"; correlated subqueries should not be allowed in mutations):

1. Non-correlated `EXISTS` over a real source table no longer aborts the server. The `createUniqueAliasesIfNecessary` call on each detached subquery in `MutationsInterpreter` (added by #106414) gives the inner source nodes unique `__tableN` aliases, so `buildColumnIdentifier` no longer registers the same bare column name twice.
2. Correlated subqueries inside `ALTER UPDATE` / `ALTER DELETE` are rejected with a clean `NOT_IMPLEMENTED` by `correlated_subtrees.assertEmpty("in mutation filter"/"in mutation update")` in `MutationsInterpreter::prepareMutationStages`.

The earlier `ValidateNoCorrelatedSubqueryVisitor` gate that this PR carried added no behavior over `master` (every case in the test is already handled by the two pieces above — `Bugfix validation` confirms the test passes on the unfixed base), so it was removed and `MutationsInterpreter.cpp` is restored to `master`.

The test `04303_stid_4697_4326_mutation_exists_column_identifier` covers:
- non-correlated `EXISTS` over `system.one` and `NOT EXISTS` over `system.numbers` (must run, no crash);
- the AST-fuzzer `Dynamic` shape `UPDATE dynamic = NULL WHERE exists((SELECT DISTINCT *, currentDatabase() > database))`;
- three correlated shapes (`UPDATE`/`DELETE` with `EXISTS` referencing an outer column) that must error with `NOT_IMPLEMENTED`;
- the #106649 `id IN (... JOIN ...)` shape with two subquery-wrapped derived tables that both project `id`.

It is tagged `no-old-analyzer`: the bug class lives in the new analyzer's `EXISTS` rewrite; the old analyzer rejects these shapes with a different error before the planner runs.

Closes: https://github.com/ClickHouse/ClickHouse/issues/106649
Related: https://github.com/ClickHouse/ClickHouse/pull/106414

CI report: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=3120048c80b01b65ef75858d34e17b669505de6a&name_0=MasterCI&name_1=AST%20fuzzer%20%28arm_asan_ubsan%29

Add a regression test for `LOGICAL_ERROR` "Column identifier is already registered" on the `ALTER UPDATE`/`ALTER DELETE` + `EXISTS` path and for `NOT_IMPLEMENTED` rejection of correlated subqueries in mutations. The fixes themselves are already on `master`.

### Version info
- Merged into: `26.7.1.732` (included in `26.7` and later)
- Backported to: `26.6.2.58`


groeneai:
cc @CurtizJ @novikd — could you review this? Fix for chronic AST-fuzzer error "Column identifier is already registered" (STID `4697-4326`, issue #104877) on the mutations + EXISTS path. The analyzer rewrites `exists((subq))` to `1 IN (SELECT 1 FROM (subq) LIMIT 1)` in `only_analyze` mode, and the subquery tree reaches `buildSubqueryPlansForSetsAndAdd` with inner source nodes that lack `__tableN` aliases — causing both `prepareBuildQueryPlanForTableExpression` and `CollectSourceColumnsVisitor` to register the same bare column name. The fix re-runs `createUniqueAliasesIfNecessary` on the detached subquery tree before planning. The prior fix family (#100770, #101048, #101051, #101104) addressed the same error class on the SELECT side but did not cover this mutations path.

groeneai:
**Pre-PR validation gate** (session `cron:clickhouse-ci-task-worker:20260528-031500`):

a) **Deterministic repro?** Yes. Minimal command (debug build): `ALTER TABLE t UPDATE v = 0 WHERE exists((SELECT * FROM system.one)) SETTINGS mutations_sync = 2` aborts the server on every run with `LOGICAL_ERROR: Column identifier dummy is already registered`. The original AST-fuzzer query from CIDB (`ALTER TABLE test_updates (UPDATE dynamic = NULL WHERE exists((SELECT DISTINCT *, currentDatabase() > database)))`) reproduces with the same stack.

b) **Root cause explained?** Yes. In `only_analyze` mode (mutations dry_run during `validate()`), `QueryAnalyzer::resolve` rewrites `exists((subq))` -> `1 IN (SELECT 1 FROM (subq) LIMIT 1)`. `collectSets` stores that IN subquery in `PreparedSets`. `buildSubqueryPlansForSetsAndAdd` later detaches the subquery and creates a fresh `GlobalPlannerContext` per subquery. The detached tree's inner source nodes (the wrapping `QueryNode` and inner `TableNode`) lack the `__tableN` aliases that `createUniqueAliasesIfNecessary` would normally provide. `buildColumnIdentifier` thus produces a bare `dummy` for both the wrapping `QueryNode` (via `prepareBuildQueryPlanForTableExpression` -> `createColumnIdentifierOrGet`) and the inner `TableNode` (via `CollectSourceColumnsVisitor::enterImpl` -> `createColumnIdentifier`), and the second call hits the duplicate-insert throw.

c) **Fix matches root cause?** Yes. Re-running `createUniqueAliasesIfNecessary(query_tree, context_)` on the detached subquery before constructing the per-subquery `Planner` is the direct upstream of the missing-alias state: it assigns unique `__tableN` aliases to every `TableNode` / `TableFunctionNode` / nested `QueryNode` / `UnionNode` in the subquery, so the two registration sites produce distinct identifiers like `__table1.dummy` and `__table2.dummy`. Not a defensive widen of the assertion, not a `no-random-*` tag.

d) **Test intent preserved? / New test added?** Yes. New regression test `04300_stid_4697_4326_mutation_exists_column_identifier` covers the minimal trigger, the original AST-fuzzer query, and a `NOT EXISTS` variant. Existing tests for the related error class (`04064_planner_duplicate_column_identifier`, `04064_additional_result_filter_union_column_identifier`, `04066_union_all_column_identifier_collision`) still pass without modification.

e) **Demonstrated in both directions?** Yes.
- Without fix (stashed): `clickhouse-test 04300_stid_4697_4326` fails with `Having 1 errors!` (server crashes on the first mutation).
- With fix: 20/20 passes in `--test-runs 20`. Related fix-family tests: 15/15 passes over `--test-runs 5` for the three regression tests above.

f) **Fix is general, not a narrow patch?** Yes. The fix targets `buildSubqueryPlansForSetsAndAdd` -- the single place where mutations re-plan subqueries with their own per-subquery `GlobalPlannerContext`. Other related code paths handle this correctly:
- `Planner.cpp`'s `addBuildSubqueriesForSetsStepIfNeeded` (line ~1591) uses the regular SELECT path where `createUniqueAliasesIfNecessary` is already called upstream by `InterpreterSelectQueryAnalyzer`.
- `addBuildSubqueriesForMaterializedCTEsIfNeeded` plans CTE bodies that were already aliased before being attached to a `TableNode`.

Walked the callers of `createColumnIdentifier` / `createColumnIdentifierOrGet` (`CollectTableExpressionData.cpp:109`, `CollectTableExpressionData.cpp:158`, `PlannerJoinTree.cpp:494`) -- they all rely on the source node already having a unique alias. The fix re-establishes that invariant before the planner runs.

groeneai:
Pushed `3f0bc59d4d3d` adding `-- Tags: no-old-analyzer` to `04300_stid_4697_4326_mutation_exists_column_identifier.sql`.

Why this is correct, not a band-aid:
- The fix is in the new analyzer's `Planner` path (`createUniqueAliasesIfNecessary` re-run on each detached subquery tree from `PreparedSets`).
- The old analyzer never reaches `GlobalPlannerContext::registerIdentifier` for this query. Its `TreeRewriter::collectUsedColumns` rejects the AST-fuzzer query shape (`SELECT DISTINCT *, currentDatabase() > database`) with `UNKNOWN_IDENTIFIER` before any planner step runs, because the bare `database` column does not resolve in the old analyzer's `EXISTS` rewrite.
- So the bug class verified by this test cannot manifest on the old analyzer; running it there produces a different, unrelated error.

CI failure that motivated this:
https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=106025&sha=0d93b2f6bd9ece61ca2aef4f2c2a153f41860377&name_0=PR&name_1=Stateless%20tests%20%28amd_llvm_coverage%2C%20old%20analyzer%2C%20s3%20storage%2C%20DatabaseReplicated%2C%20WasmEdge%2C%20parallel%29

For reference, the other three failing checks on the current commit are not caused by this PR:

1. AST fuzzer (amd_debug) STID `0250-20a8` ("Column query tree node does not have valid source node after running C pass"): chronic trunk bug, multiple PRs in the last 7 days.
2. BuzzHouse (amd_msan) STID `2518-4199` (MSan use-of-uninitialized-value in `UserDefinedSQLFunctionVisitor`): chronic trunk MSan bug, tracked separately under issue #106040.
3. Bugfix validation: framework quirk where the inverter expects FAIL on master + PASS on PR, but the master-side `ALTER UPDATE` server-aborts (not a clean SQL error).

Pre-PR validation gate for this follow-up commit:

(a) Deterministic repro of the original failure: confirmed by the CI log linked above; query 9 (`ALTER TABLE test_updates_04300 (UPDATE dynamic = NULL WHERE exists((SELECT DISTINCT *, currentDatabase() > database))) SETTINGS mutations_sync = 2`) fails with `Code: 47. ... UNKNOWN_IDENTIFIER` on the old-analyzer variant.
(b) Root cause explained: see above. New-analyzer code path / old-analyzer code path divergence at `EXISTS` rewrite time.
(c) Fix matches root cause: `no-old-analyzer` is the standard mechanism in `tests/clickhouse-test` for this exact case ("test exercises new-analyzer-only behavior"). Sibling tests in the same fix family use `enable_analyzer = 1` query-level, but those tests are single-query; this test has multiple `ALTER ... mutations_sync = 2` queries and is more cleanly skipped wholesale.
(d) Test intent preserved: the test still exercises the same three patterns (minimal `EXISTS`, AST-fuzzer shape, `NOT EXISTS`) against the new analyzer where the bug actually lived.
(e) Both directions demonstrated: the pre-fix master HEAD (commit `f8f58430da41`) reproduces the original `Logical error: Column identifier dummy is already registered` exception when running this test, exactly the bug class the test guards. With the fix on this PR, the new-analyzer Stateless tests variant passes (162-job report column, only the old-analyzer variant fails). The change here removes the spurious old-analyzer failure, nothing else.

alexey-milovidov (approved):

alexey-milovidov:
@groeneai, why does the bugfix validation check fail to reproduce it?

groeneai:
@alexey-milovidov The test does reproduce the bug on `master` HEAD. The artifact `core.TCPHandler.1608-4962.zst.enc` on this run is the server core dump from the `LOGICAL_ERROR: 'Column identifier dummy is already registered'` that `abort_on_logical_error=true` (CI default in `tests/config/config.d/abort_on_logical_error.yaml`) converts to `SIGABRT` via `abortOnFailedAssertion` in `src/Common/Exception.cpp`. After the server dies, `tests/clickhouse-test`'s `check_server_liveness` raises `StopTesting`, the worker-to-parent `SIGTERM` feedback loop in `stop_tests` kicks the parent out, and the bash pipeline exits with `143` (`128 + SIGTERM`).

The reason that did not surface as "bug reproduced" is a 30-hour timing gap. PR #105643 widened `FTResultsProcessor`'s aborted-run synthesis from `runner_exit_code == STOP_TESTING_EXIT_CODE` (only `2`) to `runner_exit_code in {2, 137, 143, -15, -9}` and adds a synthetic `Server died` `FAIL` leaf that the bugfix-validation inverter would then flip to `OK`. That merged into `master` on `2026-05-29 18:14 UTC` (commit `db6d7a5e`). #106025's Bugfix validation ran on `2026-05-28 12:30 UTC`, 30 hours earlier, so at that SHA the runner exit `143` did not match the single-value check. The run took the `elif not s.success_finish` branch and produced `Tests = ERROR / "The test runner was terminated unexpectedly"`, and my #105789 inverter preserved that `ERROR` instead of overwriting it with the misleading "Failed to reproduce the bug".

I will merge `master` in to pick up #105643 and trigger a fresh run; with the wider `ABORTED_RUN_EXIT_CODES`, the bugfix-validation result should flip to `OK` / "bug reproduced".

novikd:
Should we allow correlated subqueries in mutations at all?

KochetovNicolai (changes requested):
Temporarily block.

groeneai:
@novikd Good question, and I want to be precise about what this PR does and does not do.

This PR does not add support for correlated subqueries in mutations. It fixes a server crash (`LOGICAL_ERROR: 'Column identifier ... is already registered'`, which becomes a SIGABRT under `abort_on_logical_error`) on `ALTER ... UPDATE/DELETE ... WHERE EXISTS((subquery))` shapes that the parser and analyzer already accept today. The regression test cases are predominantly uncorrelated, e.g.:

```sql
ALTER TABLE t UPDATE v = 0 WHERE exists((SELECT * FROM system.one)) SETTINGS mutations_sync = 2;
ALTER TABLE t UPDATE v = 1 WHERE NOT exists((SELECT * FROM system.numbers WHERE number > 1000000 LIMIT 1)) SETTINGS mutations_sync = 2;
```

The analyzer rewrites `exists((subq))` into `1 IN (SELECT 1 FROM (subq) LIMIT 1)` in only-analyze mode. The detached subquery source nodes have no `__tableN` alias, so two registration sites compute the same bare column identifier and the duplicate insert throws. The fix re-runs `createUniqueAliasesIfNecessary` on the detached subquery before building its per-subquery `Planner`, so the identifiers are unique again. No mutation semantics change.

If the team's preference is that such subqueries (correlated or not) should not be allowed in mutations at all, that is a separate decision. I am happy to instead implement a clean analyzer-level rejection that produces a proper user error rather than a crash. Either direction removes the abort. I just want to confirm which one you want before I touch it.

@KochetovNicolai Understood on the temporary block. Could you confirm what unblocks this: fix the crash as written here, or reject these queries with a clean error in the analyzer? I will follow whichever you and @novikd prefer.

KochetovNicolai:
@groeneai, please deprecate correlated subqueries in mutations.
non-correlated exists should work. Technically, we can allow correlated subqueries within scalar subqueries or IN (if it simplifies the code).

groeneai:
@KochetovNicolai @novikd Pivoted in 6a7a646.

Behavior:
- Non-correlated `EXISTS` in mutations: works (the existing `createUniqueAliasesIfNecessary` call in `buildSubqueryPlansForSetsAndAdd` is kept; without it the bare-`dummy` collision still aborts the server on the canonical reproducer).
- Correlated subqueries in mutations: rejected at analyzer time with `NOT_IMPLEMENTED: Correlated subqueries are not supported in mutations: ...`. The gate is an in-depth visitor over the resolved mutation expression that flags any `QueryNode` / `UnionNode` with correlated columns. It mirrors the existing IN-argument gate at `src/Analyzer/Resolve/resolveFunction.cpp:1139` but in `MutationsInterpreter`, so it covers `EXISTS` and any other correlated shape that survives analyzer rewrites.

Test `04303_stid_4697_4326_mutation_exists_column_identifier` extends with three new cases that must error with `NOT_IMPLEMENTED`:
- `ALTER ... UPDATE WHERE exists((SELECT 1 WHERE id > 0))`
- `ALTER ... UPDATE WHERE exists((SELECT * FROM numbers(10) WHERE number = id))`
- `ALTER ... DELETE WHERE exists((SELECT 1 WHERE v < 100))`

Existing non-correlated `EXISTS` and `NOT EXISTS` cases keep passing.

Re the carve-out you mentioned ("Technically, we can allow correlated subqueries within scalar subqueries or `IN` if it simplifies the code"): the current gate is strict - it rejects correlated wherever it appears. Correlated `IN` is already rejected globally (with a different message) by `resolveFunction.cpp:1139`. If you want me to relax this so correlated subqueries are allowed *only* when they appear as the right-hand argument of `IN` or as a scalar subquery, I can do that as a follow-up - it would need to teach the visitor to skip into those specific parents. Let me know which you prefer.

groeneai:
### Pre-PR validation gate

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. Without the crash-fix half (kept here): `ALTER TABLE t UPDATE v = 0 WHERE exists((SELECT * FROM system.one)) SETTINGS mutations_sync = 2` aborts the server (LOGICAL_ERROR on `dummy` column identifier collision in `GlobalPlannerContext::createColumnIdentifier`). For the new gate: `ALTER TABLE t UPDATE v = 99 WHERE exists((SELECT 1 WHERE id > 0))` deterministically returns `Code: 48 NOT_IMPLEMENTED: Correlated subqueries are not supported in mutations: ...`. |
| b | **Root cause explained?** | Yes (two halves). (1) Non-correlated `exists((subq))` is rewritten to `1 IN (SELECT 1 FROM (subq) LIMIT 1)`; the detached subquery reaches `buildSubqueryPlansForSetsAndAdd` with inner source nodes that have no `__tableN` aliases, so two registration sites compute the same bare column identifier for `dummy` and the duplicate-insert throws. The fix re-runs `createUniqueAliasesIfNecessary` per detached subquery before planning. (2) Correlated subqueries in mutations were the team-decided "don't support" surface (KochetovNicolai 2026-06-04T13:10:02Z); the new gate runs after `QueryAnalyzer::resolve` and rejects them at analyzer time before the planner can hit any LOGICAL_ERROR path. |
| c | **Fix matches root cause?** | Yes. Half (1) puts unique aliases on the source nodes that the planner then registers, so the duplicate identifier disappears at the source. Half (2) is a clean analyzer-time rejection at the precise scope (mutation expression list) where the team wants it; mirrors the existing IN-argument correlated-rejection gate at `resolveFunction.cpp:1139`. Neither half is a defensive guard near the throw site. |
| d | **Test intent preserved / new tests added?** | New regression test `04303_stid_4697_4326_mutation_exists_column_identifier` covers both halves: 3 non-correlated cases (must pass: minimal real-table EXISTS, AST-fuzzer-derived shape, NOT EXISTS) and 3 correlated cases (must error with `NOT_IMPLEMENTED`: UPDATE with `id > 0`, UPDATE with `number = id`, DELETE with `v < 100`). Existing `03925_exists_with_mutations_execute_on_initator` continues to pass (verified locally). |
| e | **Both directions demonstrated?** | Yes. Local ASan build, single-binary run on port 29200. Without the fix branch: server SIGABRTs on the minimal non-correlated repro. With the fix branch: non-correlated cases return rows as expected, correlated cases return `Code: 48 NOT_IMPLEMENTED` with no server abort. Output matches `.reference` exactly (zero-byte diff under `--multiquery`). |
| f | **Fix is general, not a narrow patch?** | Yes. Half (1) is a single per-subquery `createUniqueAliasesIfNecessary` invocation that handles every detached subquery in `PreparedSets`, so it covers any `exists` / IN-style rewrite that lands in the mutation planner path with a missing-alias source. Half (2) walks the entire mutation expression (filters AND update assignments) via `InDepthQueryTreeVisitor`, so any correlated `QueryNode` / `UnionNode` is caught regardless of which operator (EXISTS, scalar subquery, etc.) hosts it. Sibling correlated-subquery gates in IN argument resolution (`resolveFunction.cpp:1139`) and in JOIN resolution (`QueryAnalyzer.cpp:4852-4857`) are already in place; this adds the parallel mutation-scope gate. |

Session id: cron:clickhouse-ci-task-worker:20260604-135400

groeneai:
Pushed `58e4360e731b` adding the Issue #106649 shape to the regression test.

### Pre-PR validation gate

| # | Question | Answer |
|---|----------|--------|
| a | **Deterministic repro?** | Yes. On a master-HEAD debug build (`MutationsInterpreter.cpp` reverted to master), running `ALTER TABLE gh_04303 UPDATE x = 1 WHERE id IN (SELECT i.id FROM (SELECT id FROM gh_04303) AS i JOIN edges_04303 AS r ON r.to_id = i.id JOIN (SELECT id FROM prs_04303) AS pr ON pr.id = r.from_id) SETTINGS mutations_sync = 2;` deterministically aborts the server with `SIGABRT` (exit 134) on the same `Logical error: Column identifier id is already registered (STID: 4697-4326)` chronic family. |
| b | **Root cause explained?** | Yes (already covered by the existing fix in this PR; this commit only adds test coverage). The mutation path's `MutationsInterpreter::buildSubqueryPlansForSetsAndAdd` plans each detached subquery from `PreparedSets`. The `id IN ((SELECT id FROM x) AS a JOIN b JOIN (SELECT id FROM c) AS pr)` rewrite leaves the inner derived-table source nodes without `__tableN` aliases, so `prepareBuildQueryPlanForTableExpression` and `CollectSourceColumnsVisitor` register the same bare `id` column identifier twice and the second insert throws. Same mechanism as the EXISTS shape, different parent. |
| c | **Fix matches root cause?** | Yes. The per-detached-subquery `createUniqueAliasesIfNecessary` call already in this PR (the half kept after the team-direction pivot) covers any rewrite that lands in the mutation planner with unaliased source nodes — `exists((subq))` and the IN-subquery shape both flow through the same code path. No additional code change needed for #106649. This commit only extends the regression test. |
| d | **Test intent preserved / new tests added?** | New regression case added: a 3-table setup (`gh_04303`, `prs_04303`, `edges_04303`) executing the canonical Issue #106649 reproducer (ALTER UPDATE WHERE id IN with two subquery-wrapped derived tables joined twice, both projecting `id`), followed by `SELECT count() FROM gh_04303`. The existing EXISTS / NOT EXISTS / correlated cases are untouched. |
| e | **Both directions demonstrated?** | Yes. With master HEAD `MutationsInterpreter.cpp`: `clickhouse local --multiquery < 04303_*.sql` exits 134 (SIGABRT on the new query). With this PR's branch: same invocation exits 0 with the reference-matching output `0` (count of `gh_04303`, empty after the no-op ALTER UPDATE). Verified locally on a fresh debug build at the new commit. |
| f | **Fix is general, not a narrow patch?** | This commit is test-only — no source change. The underlying source fix (`createUniqueAliasesIfNecessary` on each detached `PreparedSets` subquery in `buildSubqueryPlansForSetsAndAdd`) was already general: it covers every detached subquery regardless of whether the mutation rewrite produced it from `EXISTS`, `IN`, or any other operator. The new test just verifies a second concrete shape is covered. |

Session id: cron:clickhouse-ci-task-worker:20260607-100600

groeneai:
### CI ledger — covered SHA `58e4360e731b`

CI fully finished on this SHA (Finish Workflow pass at 14:42:44Z). Triage of the only red:

| check | failure | classification | disposition |
|---|---|---|---|
| Integration tests (amd_asan_ubsan, db disk, old analyzer, 1/6) | `test_hedged_requests/test.py::test_stuck_replica` | chronic flaky, NOT PR-caused | Tracked in [PR #106571 task](https://github.com/ClickHouse/ClickHouse/pull/106571) (active investigation, P0). 30-day cross-PR: 9 distinct unrelated PRs (single hits each). PR diff is `MutationsInterpreter.cpp` + `04303` mutations test extension, completely unrelated to hedged-requests / replica timeout behavior. |

All other checks pass on this SHA. Mergeable Check fails downstream of the chronic flaky integration test only.

Session id: `cron:our-pr-ci-monitor:20260607-153000`

alexey-milovidov:
@groeneai, as the fix for flaky test has been merged, update the branch in this PR.

groeneai:
@alexey-milovidov merged `upstream/master` into the branch in commit `399eea2c02d`, picking up PR #106695 (test_stuck_replica SIGSTOP fix, merged 2026-06-08T11:25:25Z) which was the only red on the previous CI ledger. Awaiting fresh CI.

groeneai:
CI ledger for `399eea2c02d` (post-merge-master HEAD per @alexey-milovidov 2026-06-08T14:40:03Z directive). Workflow run 27149175588 completed 2026-06-08T19:40:51Z. 2 failures, both pre-existing / infra, NOT PR-caused:

| check | failure | classification | disposition |
|---|---|---|---|
| Stateless tests (amd_llvm_coverage, ParallelReplicas, s3 storage, parallel) | 04295_04278_materialized_cte_in_union_runtime_set | chronic flaky | 30-day cross-PR: 4 master hits + 19 distinct unrelated PRs. Active fix path via PR #106717. PR #106025 diff (`MutationsInterpreter.cpp` + 04303 SQL test) does not touch CTE / parallel replicas. |
| Docker server image (clickhouse-server:head-alpine-arm64) | failed to fetch `ubuntu:22.04` from Docker Hub: 504 Gateway Timeout | transient infra | Docker Hub outage during `FROM ubuntu:22.04 AS glibc-donor`. |
| Mergeable Check / PR | aggregator | downstream | Reflects the two above. |

PR is APPROVED (@alexey-milovidov 2026-05-31; pivot per @KochetovNicolai directive 2026-06-04 deprecates correlated subqueries in mutations). Awaiting reviewer ack of the master-merge.

alexey-milovidov:
@groeneai as the test 04295_04278_materialized_cte_in_union_runtime_set was fixed, update the branch.

groeneai:
@alexey-milovidov done. Merged latest master into the branch, picking up `b95136881bc` (PR #106543) which tags `04295_04278_materialized_cte_in_union_runtime_set` as `no-parallel-replicas`. That was the only red on the prior CI run (ParallelReplicas + s3 variant). New HEAD `98231edd414`. Our fix (`MutationsInterpreter.cpp` + the `04303` regression test) is unchanged by the merge; it is still the only diff vs master. CI is re-running.

groeneai:
CI ledger for `98231edd4143` (master-merge picking up the 04295 fix b95136881bc, per the update-branch request). CI fully finished (Finish Workflow SUCCESS at 16:40:39Z). 2 failures, neither PR-caused:

- Stateless tests (arm_asan_ubsan, azure, sequential): Server died + AddressSanitizer (STID 2046-4f7a). ASAN runtime-internal CHECK at sanitizer_allocator_secondary.h:200 on the azure sequential runner. Known azure+ASAN infra flake. 30-day cross-PR shows 1 hit here plus 1 unrelated master hit. Not in this PR diff (MutationsInterpreter.cpp + 04303 test).
- Stateless tests (amd_debug, distributed plan, s3 storage, parallel): 03752_attach_as_replicated_transaction_metadata, chronic flaky (7 unrelated PRs in 30d), tracked separately. Unrelated to this PR diff.

All other checks pass.

KochetovNicolai (approved):

groeneai:
CI finished on HEAD `d55c744ecc04` (master-merge commit). One failure, not PR-caused:

- **Integration tests (amd_tsan, 4/6)** / `test_zookeeper_fallback_session::test_fallback_session` — flaky. 30-day cross-PR: fails on 3 distinct unrelated PRs (#105499, #105056, this PR). Exercises Keeper/ZooKeeper session-fallback, timing-sensitive under TSAN. Tracked separately as a flaky integration test. This PR's diff (MutationsInterpreter.cpp + the 04303 SQL test) does not touch ZooKeeper fallback.

All other checks pass. The fix and the correlated-subqueries-in-mutations deprecation gate are unchanged by the master-merge. Ready for re-review / merge.

alexey-milovidov:
@groeneai, fix conflicts.

alexey-milovidov (approved):

groeneai:
@alexey-milovidov conflicts resolved. Merged master into the branch; new HEAD `d51e51a815776f961987e91e6426e4f04d4f431b`, PR is mergeable again.

The only conflict was in `MutationsInterpreter.cpp`: the `createUniqueAliasesIfNecessary` crash-fix landed independently on master via 01db4635ff7, so I kept master's comment for that line. This PR's net delta vs master is now just the correlated-subquery `NOT_IMPLEMENTED` gate plus the `04303` regression test.

alexey-milovidov:
@groeneai, now the CI says that the test does not reproduce the bug on the unfixed version.

tests/queries/0_stateless/04303_stid_4697_4326_mutation_exists_column_identifier.sql
```diff
@@ -0,0 +1,77 @@
...
+
+ALTER TABLE t_04303_corr UPDATE v = 99
+    WHERE exists((SELECT * FROM numbers(10) WHERE number = id))
+    SETTINGS mutations_sync = 2; -- { serverError NOT_IMPLEMENTED }
+
+ALTER TABLE t_04303_corr DELETE
+    WHERE exists((SELECT 1 WHERE v < 100))
+    SETTINGS mutations_sync = 2; -- { serverError NOT_IMPLEMENTED }
+
+-- Issue #106649 shape: ALTER UPDATE WHERE id IN (subquery joining two
```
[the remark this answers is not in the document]
groeneai: Done. Removed both `Closes #104877` occurrences from the PR body and the changelog entry; only `Closes #106649` (the correct tracker for this mutation column-identifier bug) remains. The test file already referenced #106649.

groeneai:
@alexey-milovidov, that result is now correct, because both fixes already landed on master.

Your commit 01db4635ff7 ("Fix logical error 'Column identifier ... is already registered' in mutations", 2026-06-04) added the `createUniqueAliasesIfNecessary` crash-fix (plus test 04305), and master's `MutationsInterpreter` now calls `correlated_subtrees.assertEmpty("in mutation filter"/"in mutation update")`, which throws `NOT_IMPLEMENTED` ("Correlated subqueries ... are not supported"). So the unfixed master binary no longer aborts on these queries, and Bugfix validation rightly reports "does not reproduce".

That makes both code pieces of this PR redundant with master. The only remaining net delta is the regression test `04303_stid_4697_4326_mutation_exists_column_identifier` (3 non-correlated EXISTS cases that must pass + 3 correlated cases that must error `NOT_IMPLEMENTED` + the #106649 IN-subquery shape).

Options:
1. Close this PR as superseded by 01db4635ff7, and I open a small test-only PR adding the 04303 cases not already covered by 04305 (the correlated `NOT_IMPLEMENTED` cases and the #106649 IN shape), changelog category "CI Fix or Improvement". This is the cleanest.
2. Keep this PR but reduce it to the test only, and switch the changelog category to "CI Fix or Improvement" since the code fix is already on master.

Which do you prefer? I'll keep only `Closes #106649` either way (the #104877 references are removed).

src/Interpreters/MutationsInterpreter.cpp
```diff
@@ -173,6 +175,31 @@ QueryTreeNodePtr prepareQueryAffectedQueryTree(const std::vector<MutationCommand
...
+            throw Exception(
+                ErrorCodes::NOT_IMPLEMENTED,
+                "Correlated subqueries are not supported in mutations: {}",
+                node->formatASTForErrorMessage());
+    }
+};
+
+void validateNoCorrelatedSubqueriesInMutationExpression(const QueryTreeNodePtr & expression)
+{
+    ValidateNoCorrelatedSubqueryVisitor visitor;
```
novikd: @groeneai, instead of creating another visitor, you can reuse the `traverseQueryTree` function.

novikd (commented):
Overall LGTM

novikd:
@groeneai check why [Bugfix validation (functional tests)](https://s3.amazonaws.com/clickhouse-test-reports/json.html?PR=106025&sha=d51e51a815776f961987e91e6426e4f04d4f431b&name_0=PR&name_1=Bugfix%20validation%20%28functional%20tests%29) failed. Also, merge the fresh master into the branch.

groeneai:
@novikd done, merged fresh master, new HEAD `356eec3e9b7`.

On Bugfix validation: it reports "Failed to reproduce the bug" because master already fixes this independently, so the regression test passes on the unfixed base and the inverter has nothing to flip. Two master pieces cover it:

- `createUniqueAliasesIfNecessary` (MutationsInterpreter.cpp:1605, alexey's 01db4635ff7) is the non-correlated `EXISTS` crash-fix.
- `correlated_subtrees.assertEmpty(...)` (MutationsInterpreter.cpp:1705 and 1767) already throws `NOT_IMPLEMENTED` for correlated subqueries, the exact error 04303 asserts.

So every case in 04303 (non-correlated `EXISTS` works, correlated -> `NOT_IMPLEMENTED`) already behaves correctly on the unfixed base, and the check cannot go green here regardless of re-runs.

The net delta of this PR vs master is now an earlier explicit gate (rejects correlated subqueries right after analyzer resolve, before the planner) plus the 04303 regression test. I can reduce it to a test-only PR with the `CI Fix or Improvement` category if you prefer; otherwise it is ready as-is.

novikd:
@groeneai, remove the code added in https://github.com/ClickHouse/ClickHouse/pull/106414 (keep the test). Correlated subqueries should not be allowed in the mutations.

novikd:
@alexey-milovidov I want to revert changes from https://github.com/ClickHouse/ClickHouse/pull/106414 in this PR, because IMHO mutations should not contain correlated subqueries. wdyt?

groeneai:
@novikd I built and tested both halves of this before touching the branch, and the result changes the picture:

**1. #106414 is a non-correlated crash fix, not correlated-subquery support.** Its only code change is one `createUniqueAliasesIfNecessary(query_tree, ...)` call in `buildSubqueryPlansForSetsAndAdd` (MutationsInterpreter.cpp:2030). I removed exactly that line on the current branch, rebuilt, and the server aborts:

```
DELETE FROM test_filter WHERE exists((SELECT DISTINCT * LIMIT 452));
-> Logical error: 'Column identifier dummy is already registered'  (SIGABRT)
   GlobalPlannerContext::createColumnIdentifier
   <- collectTableExpressionData <- Planner <- buildSubqueryPlansForSetsAndAdd
```

That query is **not** correlated (the inner `SELECT` just reads `system.one`); it is one of the cases in `04305_mutation_in_subquery_column_identifier`, the test you asked to keep. So reverting #106414 re-breaks a kept test rather than removing correlated support.

**2. Correlated subqueries in mutations are already rejected on master, independent of #106414.** `correlated_subtrees.assertEmpty(...)` at MutationsInterpreter.cpp:1736 (filter) and :1798 (update) throws `NOT_IMPLEMENTED: Correlated subqueries ... are not supported`. I disabled my own gate, rebuilt, and every correlated case in 04303 still gets a clean `Code: 48`, no crash. So "correlated subqueries should not be allowed in mutations" already holds.

**3. Consequence:** my analyzer-scope gate (the visitor at :184) is redundant with that `assertEmpty`. The only non-redundant part of this PR vs master is the 04303 regression test.

Given that, the revert as written is not what we want. The coherent options:

- **(a)** Drop my redundant gate, keep #106414's crash fix (already on master) and the 04303 test. This is effectively a test-only PR with the `CI Fix or Improvement` category.
- **(b)** If you specifically want correlated subqueries rejected earlier than today's `assertEmpty` (e.g. before the planner, with a mutation-specific message), I keep a gate but reuse `traverseQueryTree` per your other comment instead of a new visitor.

Reverting #106414's `createUniqueAliasesIfNecessary` is not on the table either way, since it guards the non-correlated crash. Which of (a)/(b) do you prefer? You also asked @ alexey-milovidov about touching #106414, so I will hold until you both agree.

src/Interpreters/MutationsInterpreter.cpp
alexey-milovidov: Resolved by removing the visitor entirely. The gate added no behavior over `master`: correlated subqueries in mutations are already rejected with `NOT_IMPLEMENTED` by `correlated_subtrees.assertEmpty("in mutation filter"/"in mutation update")` in `prepareMutationStages`, and the non-correlated `EXISTS` crash is already fixed by `createUniqueAliasesIfNecessary` (#106414). `MutationsInterpreter.cpp` is restored to `master`, so there is no longer a visitor (or a `traverseQueryTree` call) to maintain. The PR is now test-only.

alexey-milovidov:
Reduced this PR to **test-only** and merged fresh `master`. New HEAD `a22dc549f0a`.

**What changed**

- Removed the `ValidateNoCorrelatedSubqueryVisitor` gate; `src/Interpreters/MutationsInterpreter.cpp` is now byte-identical to `master`. This addresses @novikd's review ("remove the code … keep the test") and the `traverseQueryTree` thread (now moot — there is no visitor left). Net diff vs `master` is just the `04303` regression test.
- The gate was redundant. Both behaviors it guarded are already on `master`:
  - non-correlated `EXISTS` no longer aborts the server — `createUniqueAliasesIfNecessary` per detached subquery (#106414, `MutationsInterpreter.cpp:1605`/`:1999`);
  - correlated subqueries in mutations are rejected with `NOT_IMPLEMENTED` — `correlated_subtrees.assertEmpty("in mutation filter"/"in mutation update")` (`MutationsInterpreter.cpp:1705`/`:1767`).
- Changelog category switched from `Bug Fix` to `CI Fix or Improvement` (the fixes are already on `master`; this PR only adds coverage).

**Verification** (master-behavior binary, `clickhouse local`):

| case | result |
|---|---|
| `UPDATE … WHERE exists((SELECT * FROM system.one))` | `1 0 / 2 0`, no crash |
| `UPDATE dynamic = NULL WHERE exists((SELECT DISTINCT *, currentDatabase() > database))` | `1 1 / 2 1` |
| `UPDATE … WHERE NOT exists((SELECT … system.numbers … LIMIT 1))` | `1 10 / 2 20` (filter false) |
| `UPDATE … WHERE exists((SELECT 1 WHERE id > 0))` (correlated) | `Code: 48 NOT_IMPLEMENTED` |
| `UPDATE … WHERE exists((SELECT * FROM numbers(10) WHERE number = id))` (correlated) | `Code: 48 NOT_IMPLEMENTED` |
| `DELETE WHERE exists((SELECT 1 WHERE v < 100))` (correlated) | `Code: 48 NOT_IMPLEMENTED` |
| #106649 `id IN (… JOIN …)` shape | `count() = 0`, no crash |

Output matches the `.reference` exactly. `Bugfix validation` will continue to report "does not reproduce" — by design, since `master` already handles every case.

**One open decision for a human, @alexey-milovidov / @novikd**: I did **not** revert #106414 itself. As demonstrated earlier (and reproduced here), #106414 is the *non-correlated* `EXISTS` crash-fix, not correlated-subquery support — reverting it re-breaks the non-correlated case `DELETE … WHERE exists((SELECT DISTINCT * LIMIT 452))` (covered by `04305_mutation_in_subquery_column_identifier`). Correlated subqueries are already disallowed on `master` via `assertEmpty`. Reverting #106414 is a separate change from this PR's net diff; if you still want it, it should be its own PR with `04305` updated accordingly.

alexey-milovidov:
CI finished on HEAD `a22dc549f0a` (test-only). The PR is ready: net diff is just the `04303_stid_4697_4326_mutation_exists_column_identifier` test, `MutationsInterpreter.cpp` is byte-identical to `master`, AI Review is ✅ Approve, and `04303` passed everywhere it actually ran (`Fast test`, `arm_darwin`, every stateless lane, and the flaky-check shards on asan/msan/tsan/debug). `Bugfix validation` is correctly **SKIPPED** now that the category is `CI Fix or Improvement`.

The 3 remaining reds are all unrelated flakes — this PR adds only a stateless SQL test (no C++/Keeper changes):

- **Stress test (amd_tsan / arm_tsan)** — `Hung check failed, possible deadlock found` → known master-wide flake #107941 (the report already links it). The hung-query thread dumps reference `mt_with_pk` / `04411_t`, never `04303`.
- **Fast test (arm_darwin)** — `03231_keeper_cpr_mvr_commands` → a Keeper-client/Darwin flake unrelated to mutations; it also failed on `master` (PR `0`) on 2026-06-26 with the same "having stderror" signature. No tracking issue yet.

@groeneai, could you confirm/track the `03231_keeper_cpr_mvr_commands` arm_darwin flake (also seen on `master` 2026-06-26) and provide a fix in a separate PR if warranted? It's not caused by this test-only change. Otherwise the PR is good to merge once the flaky reds are re-run/overridden.

groeneai:
@alexey-milovidov Confirmed and tracked. `03231_keeper_cpr_mvr_commands` on `Fast test (arm_darwin)` is the same benign keeper-client teardown flake as `03135_keeper_client_find_commands`, and a fix already exists and is approved.

Root cause (stdout was correct on that run; only stderr tripped the harness): at `clickhouse-keeper-client` process exit the one-shot `-q` client destroys its `zkutil::ZooKeeper`, and `Coordination::ZooKeeper::finalize` calls `Poco::Net::SocketImpl::shutdown()` on an already-disconnected socket. On macOS that raises `Net Exception: Socket is not connected` (errno 57 / ENOTCONN), which gets logged to stderr. `Fast test` treats any stderr as failure ("having stderror"). Unrelated to this test-only change (no C++/Keeper code here).

Fix: #107438 ("Do not log benign ENOTCONN on socket shutdown during ZooKeeper finalize") downgrades exactly this log to TRACE at both `socket.shutdown()` sites in `ZooKeeper::finalize`. It is engine-side, so it covers the whole keeper-client teardown family (03135, 03231, ...) on arm_darwin. Approved by @alesapin on 2026-06-19, OPEN awaiting merge.

CIDB (30d, `Fast test (arm_darwin)`): family is arm_darwin-only across unrelated PRs, 0 on master. The `03231` hit on `master` 2026-06-26 was a different benign cause (`LLVM Profile Error: Invalid profile data to merge` injected into stdout on the coverage lane), not this stderr flake.

No separate PR needed here. #106025 is good to merge once the flaky reds are re-run/overridden.

groeneai:
### CI finish ledger — a22dc549f0a

CI fully finished (Finish Workflow: success). This PR is test-only (ALTER UPDATE column-identifier regression), APPROVED by @ alexey-milovidov + @ KochetovNicolai. Every failure has an owner; none is PR-caused.

|---|---|---|
| Fast test (arm_darwin) / 03231_keeper_cpr_mvr_commands | flaky: benign ENOTCONN stderr at keeper-client teardown (keeper-client family, 17 hits / 17 PRs / 30d, arm_darwin only) | #107438 (ours, open, approved by @ alesapin) |
| Stress test (amd_tsan) + (arm_tsan) / "Hung check failed, possible deadlock found" | deadlock: #107941 deep-nested formatQuery literal slowness (per @ alexey-milovidov FYI 2026-06-30) | #108892 (external, open) — perf fix in flight; hung-check family also #108212 (ours, merged) |
| Mergeable Check / PR | aggregator | rollup of the rows above |

Session id: cron:our-pr-ci-monitor:20260630-163000

groeneai:
### CI finish ledger — a22dc549f0a1

Every failure below has an owner: a fixing PR (ours or external), or a full-effort fix task whose fixing-PR link will be posted here when it opens. Only `CH Inc sync` is exempt.

|---|---|---|
| Fast test (arm_darwin) / 03231_keeper_cpr_mvr_commands | flaky (macOS-only benign `ENOTCONN` stderr noise at keeper-client ZooKeeper finalize; command output correct; 0 master, hits unrelated PRs) | #107438 (ours, open) |
| Stress test (amd_tsan) / Hung check failed, possible deadlock found | chronic stress-shutdown deadlock family (not PR-caused; this PR is a test-only regression test) | #108212 (ours, merged) / #105905 (ours, open) |
| Stress test (arm_tsan) / Hung check failed, possible deadlock found | same hung-check family | #108212 (ours, merged) / #105905 (ours, open) |

This PR's diff is test-only (`tests/queries/0_stateless/04303_stid_4697_4326_mutation_exists_column_identifier.{sql,reference}`); PR-caused gate = ZERO. The `03231` keeper-client flake and the stress Hung-check are both pre-existing trunk issues each owned by an open fixing PR above.

Session id: cron:our-pr-ci-monitor:20260701-043000

groeneai:
### CI finish ledger — 75ede47bc230eafa6d75aa71c05abbf8dabff34e

CI fully finished on head `75ede47bc230` (Finish Workflow: success, Mergeable Check: success). This PR is test-only (ALTER UPDATE column-identifier regression). Every failure below has an owner.

|---|---|---|
| Stress test (arm_tsan) — ThreadSanitizer data race (STID 1003-358c) | data race in `UniqExactSet::merge` -> `HashSetTable::merge` -> `HashTable::resize` (parallel uniqExact two-level merge); chronic trunk, NOT PR-caused (this diff is test-only; race spans 7 unrelated PRs + 1 master/30d) | #109389 (ours, merged 2026-07-07) — the failing build (06:35Z) predated the fix merge (10:49Z); merged master @ `0c72215052071` to pick it up. Zero STID 1003-358c recurrences anywhere after the fix merged. |

Session id: cron:our-pr-ci-monitor:20260708-173000

groeneai:
### CI finish ledger — 0c72215052071330d67357b72d15da25b3e64880

Clean CI. The master-merge that picked up #109389 (uniqExact two-level parallel-merge TSan fix) resolved the prior head's sole failure. No failures on this head; nothing PR-caused.

|---|---|---|
| (all checks) | — | green — 62 pass / 110 skipped, Finish Workflow + Mergeable pass |

Prior head 75ede47b carried Stress test (arm_tsan) STID 1003-358c (UniqExactSet::merge TSan data race, chronic trunk, not PR-caused), owned by #109389 (ours, merged 2026-07-07). Merging master to pick it up cleared it — 0 recurrences on this head.

Session id: cron:our-pr-ci-monitor:20260709-000000

PR #100603, opened 2026-03-24
Emit virtual row from read step

vdimir:
### Changelog category (leave one):
- Performance Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

* When `read_in_order_use_virtual_row` is enabled together with the new `read_in_order_use_virtual_row_per_block` setting, virtual row boundary information is now emitted after each block read from MergeTree, allowing the merge to reprioritize sources mid-stream for parts whose data is fully filtered out by WHERE/PREWHERE/JOIN. Close https://github.com/ClickHouse/ClickHouse/issues/99945

### Documentation entry for user-facing changes

- [x] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.4.1.899`

PR #106026, opened 2026-05-28
Fix mis-ordering with read_in_order_use_virtual_row_per_block on wide partial reads

groeneai:
Fixes an `ORDER BY ... LIMIT` correctness bug when `read_in_order_use_virtual_row_per_block = 1` is combined with wide parts that read partial granules. Output values can appear out of order around the seam between a stream's first and second emitted chunk, e.g. `99988, 99996, 99996, ...` instead of monotonically descending.

### Root cause

`MergeTreeSelectProcessor::buildVirtualRowFromIndex` computes the per-block boundary key from `index[next_mark]`, i.e. the start of the next granule. For wide parts (`MergeTreeReaderWide::canReadIncompleteGranules` returns true), a single block can contain only part of a granule, so a chunk whose `read_mark_ranges` reports `[N, N+1)` may carry just the last few rows of granule N while the rest of that granule lives in the previous forward chunk. For `MergeTreeInReverseOrderSelectAlgorithm` (which buffers all chunks and pops them LIFO, then `ReverseTransform` flips each chunk's rows), this means the index-derived key sits far below the actual values still pending from this source. `MergingSortedTransform` then deprioritizes the source and emits smaller values from other parts before returning to it, producing the mis-ordered output.

### Fix

Prefer extracting the per-block boundary PK from the chunk's actual data (`buildVirtualRowFromChunk`):

- forward reads: use the last row's PK -- a tight lower bound for the next chunk's data,
- reverse reads: use the first forward row's PK -- a tight upper bound on what the next emit can be after `ReverseTransform`.

When the chunk has no surviving rows (PREWHERE filtered everything) or the PK columns are not part of the query result, fall back to the original index-based path -- but only if the chunk's `num_read_rows` exactly equals the granule sizes covered by `read_mark_ranges`, so partial-granule reads no longer emit a misleading boundary.

Source columns from the chunk are materialized via `convertToFullIfNeeded` before `insertFrom` so the read works regardless of sparse / const / low-cardinality / replicated wrappers.

### Reproducer (fails on master, passes with this PR)

```sql
DROP TABLE IF EXISTS t_vrow;
CREATE TABLE t_vrow (CounterID UInt32, CounterID2 UInt32)
ENGINE = MergeTree ORDER BY CounterID
SETTINGS index_granularity = 28454, min_bytes_for_wide_part = 0;
SYSTEM STOP MERGES t_vrow;
-- 10 INSERTs of 100k rows each with `max_block_size = 14633`
INSERT INTO t_vrow SELECT sipHash64(number, 0) % 100000, rand32() % 100000 FROM numbers(100000) SETTINGS max_block_size = 14633;
-- ... 9 more INSERTs
SET optimize_read_in_order = 1, read_in_order_use_virtual_row = 1, read_in_order_use_virtual_row_per_block = 1, max_block_size = 14633;
SELECT CounterID FROM t_vrow ORDER BY CounterID DESC LIMIT 100;
```

Before the fix, the output transitions from `99988` straight back to `99996, 99996, 99994, ...` around row 50. After the fix, the output is monotonically non-increasing.

The feature was introduced in #100603. The existing test `04057_virtual_row_many_parts` does not pin `min_bytes_for_wide_part = 0` and only catches the bug when the random-settings runner hits the wide + small-`max_block_size` + small-`index_granularity` combination. A new stateless test `04301_virtual_row_per_block_wide_partial_granule` pins these settings and asserts both `ORDER BY ... DESC LIMIT 100` and `ASC LIMIT 100` are monotonic.

### Validation

- Reproducer above fails deterministically on master, passes on this branch.
- `04057_virtual_row_many_parts`: 50/50 successful runs with `--test-runs 50` and full random-settings randomization on a local debug build.
- All other `virtual_row` / `read_in_order` stateless tests still pass.

Fix wrong `ORDER BY ... LIMIT` output when reading wide `MergeTree` parts with `read_in_order_use_virtual_row_per_block = 1` and an `index_granularity` larger than `max_block_size`. The per-block virtual-row boundary is now derived from the chunk's actual data instead of the granule edge, so `MergingSortedTransform` no longer deprioritizes a source whose next chunk still contains larger values.

### DIFF
--- a/src/Storages/MergeTree/MergeTreeSelectProcessor.cpp
+++ b/src/Storages/MergeTree/MergeTreeSelectProcessor.cpp
@@ -311,14 +311,41 @@ void MergeTreeSelectProcessor::setVirtualRowConversions(
     read_in_reverse_order = read_in_reverse_order_;
 }
 
+ChunkAndProgress MergeTreeSelectProcessor::makeVirtualRowChunk(
+    const MergeTreeReadTask & current_task, Block pk_block) const
+{
+    const auto & data_part = current_task.getInfo().data_part;
+
+    Columns empty_columns;
+    empty_columns.reserve(result_header.columns());
+    for (size_t i = 0; i < result_header.columns(); ++i)
+        empty_columns.push_back(result_header.getByPosition(i).type->createColumn()->cloneEmpty());
+
+    Chunk chunk(std::move(empty_columns), 0);
+    auto part_level = data_part->info.level;
+    chunk.getChunkInfos().add(std::make_shared<MergeTreeReadInfo>(part_level, pk_block, virtual_row_conversions));
+
+    return {std::move(chunk), 0, 0, false, {}};
+}
+
 ChunkAndProgress MergeTreeSelectProcessor::buildVirtualRowFromIndex(
-    const MergeTreeReadTask & current_task, const MarkRanges & read_mark_ranges) const
+    const MergeTreeReadTask & current_task, const MarkRanges & read_mark_ranges, size_t num_read_rows) const
 {
     if (!virtual_row_conversions || read_mark_ranges.empty())
         return {};
 
     const auto & data_part = current_task.getInfo().data_part;
     const auto & index = data_part->getIndex();
+
+    /// The index-based key only correctly bounds the next chunk's data when this chunk's
+    /// physical reads exactly cover the granules listed in `read_mark_ranges`. When the
+    /// chunk started or ended mid-granule (possible only for wide parts, which allow
+    /// partial-granule reads), some rows of those granules live in adjacent chunks, so
+    /// `index[next_mark]` would be a much looser bound than the actual next data and
+    /// would mis-signal `MergingSortedTransform` to deprioritize this source.
+    size_t expected_rows = data_part->index_granularity->getRowsCountInRanges(read_mark_ranges);
+    if (num_read_rows != expected_rows)
+        return {};
 
     /// Forward order: the source will next produce data starting at back().end.
     /// Reverse order: MergeTreeInReverseOrderSelectAlgorithm returns chunks in reverse.
@@ -345,18 +372,50 @@ ChunkAndProgress MergeTreeSelectProcessor::buildVirtualRowFromIndex(
         column->insert((*(*index)[j])[next_mark]);
         pk_columns.push_back({std::move(column), header_col.type, header_col.name});
     }
-    Block pk_block(std::move(pk_columns));
+    return makeVirtualRowChunk(current_task, Block(std::move(pk_columns)));
+}
 
-    Columns empty_columns;
-    empty_columns.reserve(result_header.columns());
-    for (size_t i = 0; i < result_header.columns(); ++i)
-        empty_columns.push_back(result_header.getByPosition(i).type->createColumn()->cloneEmpty());
+ChunkAndProgress MergeTreeSelectProcessor::buildVirtualRowFromChunk(
+    const MergeTreeReadTask & current_task, const Chunk & data_chunk) const
+{
+    if (!virtual_row_conversions || data_chunk.getNumRows() == 0)
+        return {};
 
-    Chunk chunk(std::move(empty_columns), 0);
-    auto part_level = data_part->info.level;
-    chunk.getChunkInfos().add(std::make_shared<MergeTreeReadInfo>(part_level, pk_block, virtual_row_conversions));
+    size_t num_pk_columns = pk_block_header.columns();
 
-    return {std::move(chunk), 0, 0, false, {}};
+    /// PK columns must be present in `result_header` (i.e. the query selects them or they
+    /// were pulled in for sorting). Without them we cannot extract a tight boundary key.
+    std::vector<size_t> pk_positions;
+    pk_positions.reserve(num_pk_columns);
+    for (size_t j = 0; j < num_pk_columns; ++j)
+    {
+        const auto & pk_col_name = pk_block_header.getByPosition(j).name;
+        if (!result_header.has(pk_col_name))
+            return {};
+        pk_positions.push_back(result_header.getPositionByName(pk_col_name));
+    }
+
+    /// Forward: data is in ascending PK order. Last row's PK is the MAX value the merge
+    ///   just saw from this source, so it is a tight lower bound on the next data's PK.
+    /// Reverse: chunks are read forward (ascending PK) and emitted last-row-first by
+    ///   `ReverseTransform` downstream. After the chunk's reversed data is consumed, the
+    ///   next emitted value is bounded above by the chunk's FIRST forward row's PK.
+    size_t source_row = read_in_reverse_order ? 0 : (data_chunk.getNumRows() - 1);
+
+    const auto & data_columns = data_chunk.getColumns();
+    ColumnsWithTypeAndName pk_columns;
+    pk_columns.reserve(num_pk_columns);
+    for (size_t j = 0; j < num_pk_columns; ++j)
+    {
+        const auto & header_col = pk_block_header.getByPosition(j);
+        auto column = header_col.column->cloneEmpty();
+        /// Source may be wrapped (sparse, const, low-cardinality, replicated). `insertFrom`
+        /// requires a representation compatible with the destination, so fully materialize.
+        ColumnPtr source_column = data_columns[pk_positions[j]]->convertToFullIfNeeded();
+        column->insertFrom(*source_column, source_row);
+        pk_columns.push_back({std::move(column), header_col.type, header_col.name});
+    }
+    return makeVirtualRowChunk(current_task, Block(std::move(pk_columns)));
 }
 
 ChunkAndProgress MergeTreeSelectProcessor::read()
@@ -429,13 +488,20 @@ ChunkAndProgress MergeTreeSelectProcessor::read()
 
         auto result = readCurrentTask(*task, *algorithm);
 
-        /// Emit a virtual row update after each block, carrying the next mark's PK boundary.
+        /// Emit a virtual row update after each block, carrying the next chunk's PK boundary.
         /// This allows MergingSortedTransform to reprioritize sources when:
         /// - PREWHERE filters all rows (merge gets updated position without actual data)
         /// - A downstream filter (WHERE, JOIN) removes all rows (virtual row passes through filters)
+        ///
+        /// Prefer extracting the boundary key from the chunk's actual data — it is always a
+        /// tight, correct bound on the next data's PK regardless of granule alignment.
+        /// Fall back to the index-based boundary when the chunk has no surviving data (e.g.
+        /// PREWHERE filtered everything) or when the PK columns are not part of the result.
         if (virtual_row_conversions && !result.is_finished)
         {
-            auto vrow = buildVirtualRowFromIndex(*task, result.read_mark_ranges);
+            auto vrow = buildVirtualRowFromChunk(*task, result.chunk);
+            if (!vrow.chunk)
+                vrow = buildVirtualRowFromIndex(*task, result.read_mark_ranges, result.num_read_rows);
             if (vrow.chunk)
                 pending_virtual_row.emplace(std::move(vrow));
         }
--- a/src/Storages/MergeTree/MergeTreeSelectProcessor.cpp
+++ b/src/Storages/MergeTree/MergeTreeSelectProcessor.cpp
@@ -383,8 +383,11 @@ ChunkAndProgress MergeTreeSelectProcessor::buildVirtualRowFromChunk(
 
     size_t num_pk_columns = pk_block_header.columns();
 
-    /// PK columns must be present in `result_header` (i.e. the query selects them or they
-    /// were pulled in for sorting). Without them we cannot extract a tight boundary key.
+    /// `pk_block_header` carries the AST identifiers of the PK expression
+    /// (`primary_key.column_names[j]`). For an identity PK column this is the source column
+    /// name and `result_header.has` succeeds; for an expression PK like
+    /// `toStartOfMonth(date)` the chunk has only `date`, so the lookup fails and we fall
+    /// back to `buildVirtualRowFromIndex` (the index file stores the *evaluated* PK).
     std::vector<size_t> pk_positions;
     pk_positions.reserve(num_pk_columns);
     for (size_t j = 0; j < num_pk_columns; ++j)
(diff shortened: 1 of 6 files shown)


groeneai:
**Pre-PR validation gate** (TASK.md / `fix-ci-failure` skill):

a) **Deterministic repro?** YES. With `index_granularity = 28454`, `max_block_size = 14633`, `min_bytes_for_wide_part = 0`, 10 INSERTs of 100k rows each, `read_in_order_use_virtual_row_per_block = 1`, the query `SELECT CounterID FROM t ORDER BY CounterID DESC LIMIT 100` mis-orders at the same position on every run (output jumps back from `99988` to `99996` around row 50). Confirmed reproducible 80/80 by `clickhouse-test --diagnose-random-settings` in CI and 10/10 with different `sipHash64` seeds locally.

b) **Root cause explained?** YES. `MergeTreeSelectProcessor::buildVirtualRowFromIndex` uses `index[next_mark]` (the start of the next granule) as the per-block boundary key. With wide-part readers (`MergeTreeReaderWide::canReadIncompleteGranules` returns true) and `max_block_size < index_granularity`, a chunk that reports `read_mark_ranges = [N, N+1)` may carry only the last few rows of granule N -- the rest live in the previous forward chunk. The `MergeTreeInReverseOrderSelectAlgorithm` reads all chunks forward and pops them LIFO, then `ReverseTransform` flips each chunk's rows downstream. So the very first emitted chunk's index-derived key is `MIN(granule N)`, far below the actual values still pending from this source in the next-popped chunk. `MergingSortedTransform` then deprioritizes the source and emits smaller values from other parts before returning to it.

c) **Fix matches root cause?** YES. The fix derives the boundary PK from the chunk's actual data (`buildVirtualRowFromChunk`): forward reads use the last row's PK (tight lower bound on next), reverse reads use the first forward-row's PK (tight upper bound after `ReverseTransform`). Index-based path is kept as a fallback for 0-row chunks (PREWHERE-filtered) and queries that do not select the PK columns, but guarded by `num_read_rows == getRowsCountInRanges(read_mark_ranges)` so it never fires for partial-granule reads. This targets the exact mechanism rather than papering over a symptom.

d) **Test intent preserved / new tests added?** YES. The original `04057_virtual_row_many_parts` is unchanged and still passes. New stateless test `04301_virtual_row_per_block_wide_partial_granule` pins the trigger settings (`index_granularity = 28454`, `min_bytes_for_wide_part = 0`, `max_block_size = 14633`) and asserts both `ORDER BY ... DESC LIMIT 100` and `ASC LIMIT 100` produce a monotonic sequence -- this directly tests the regression instead of relying on the random-settings runner to hit the combination.

e) **Demonstrated in both directions?** YES. Without the fix the local reproducer fails on every run; with the fix `04057_virtual_row_many_parts` passes 50/50 with `--test-runs 50` + full random-settings randomization on a local debug build, and the deterministic local reproducer returns monotonic output for both ASC and DESC over 10 different `sipHash64` seeds. Other `virtual_row` / `read_in_order` stateless tests still pass.

f) **Fix is general, not a narrow patch?**
- **Same bug pattern elsewhere?** The fix replaces the boundary computation for the whole per-block virtual-row code path (forward and reverse, partial-granule and full-granule), so it is not specific to one trigger combination.
- **Fixing root cause vs guarding symptom?** Root cause: index marks alone cannot describe the boundary between two chunks when the underlying reader is allowed to read partial granules. The fix replaces marks with the chunk's actual PK data, which always describes the correct boundary. The fallback path additionally validates that the underlying reader did not in fact split a granule before emitting an index-based bound.

Session id: `cron:clickhouse-ci-task-worker:20260528-043000`

groeneai:
cc @KochetovNicolai @vdimir -- could you review this? It fixes an ORDER BY ... LIMIT mis-ordering when `read_in_order_use_virtual_row_per_block = 1` is combined with wide parts that read partial granules. The per-block boundary key is now derived from the chunk's actual data instead of `index[next_mark]`, which previously understated the boundary whenever a chunk spanned only part of the granules in `read_mark_ranges` (typical for the wide reader). vdimir, you authored the per-block feature in #100603 -- happy to revisit if you have a different design in mind.

groeneai:
Thanks for catching this. The `convertToFullIfNeeded` + `insertFrom` path was wrong: `pk_block_header` builds the destination column from the PK type, so for a `LowCardinality(...)` PK the destination is `ColumnLowCardinality`, while my materialization stripped the source down to the nested column. The result is the type mismatch you describe (`ILLEGAL_COLUMN` in release; debug builds trip the `chassert` in `IColumn::assertTypeEquality`).

Confirmed the exception locally with `LowCardinality(String)` and `LowCardinality(Int32)` PKs before the fix, then switched to a `Field` round-trip:

```cpp
Field field;
data_columns[pk_positions[j]]->get(source_row, field);
column->insert(field);
```

`IColumn::get(size_t, Field&)` and `IColumn::insert(Field)` are pure-virtual and universally implemented, so this works for any source representation (sparse / const / `LowCardinality` / replicated) and any destination wrapper without special-casing `ColumnLowCardinality`. The path runs at most once per chunk, so the per-value `Field` cost is irrelevant.

Pushed fbc34589f5 with:

1. The `Field`-based insertion in `buildVirtualRowFromChunk`
2. Regression test `04302_virtual_row_per_block_lc_pk` covering `LowCardinality(String)` and `LowCardinality(Int32)` PKs in both directions with the same trigger settings (`min_bytes_for_wide_part = 0`, `index_granularity = 28454`, `max_block_size = 14633`, `read_in_order_use_virtual_row_per_block = 1`)

Local results with the new binary (debug):
- `04302_virtual_row_per_block_lc_pk`: 20/20 passes, output monotonic in both directions for both LC types
- `04301_virtual_row_per_block_wide_partial_granule` (existing): 20/20 passes, unchanged
- `tmp/repro_lc_pk.sql` (the standalone reproducer that crashed pre-fix): now returns the LIMIT correctly

Pre-PR validation gate for this delta:
- (a) Deterministic repro of the LC bug before fix: 1/1 exception with `chassert` in `IColumn::assertTypeEquality`
- (b) Root cause: destination is `ColumnLowCardinality` (from PK type), source materialized to nested full column, `insertFrom` rejects the type pair
- (c) Fix matches root cause: `Field` round-trip is destination-and-source-agnostic
- (d) Existing test intent preserved (04301 still passes 20/20), new LC PK case covered by 04302
- (e) Both directions demonstrated (exception before, 20/20 pass after) on the same reproducer
- (f) General: `Field` works for any wrapper combination, not just LC

Comment text in `MergeTreeSelectProcessor.cpp` also drops two em-dashes per maintainer style guidance.

vdimir:
@groeneai where this repro came from? Was is caught by CI? Do we have any test which tests similar randomized scenarios with different block size/index granularity and virtual row?

groeneai:
> where this repro came from? Was is caught by CI?

Yes. Master CIDB hit on 2026-05-27 at commit `20435b5d9a50d8586e7668ebae7406f52474dcc3`, job `Stateless tests (amd_debug, parallel)` - `04057_virtual_row_many_parts` failed with a result-diff. The randomized-settings runner self-diagnosed it as deterministic under that randomization (80/80 failed) and isolated the culprits to:

```
--max_block_size 14633
--index_granularity 28454
--min_bytes_for_wide_part 0
```

(plus `read_in_order_use_virtual_row=1` from the random pool). The new test `04301_virtual_row_per_block_wide_partial_granule.sql` pins exactly those three values (`max_block_size=14633`, `index_granularity=28454`, `min_bytes_for_wide_part=0`) plus `read_in_order_use_virtual_row_per_block=1`, since `04057` does not pin them and only fails when the random runner hits all three together - which is rare. Earlier sightings: PR #102192 amd_msan WasmEdge (2026-05-17) had the same `--min_bytes_for_wide_part 0` shape; PR #100603 amd_msan flaky check (2026-04-08) was the original feature PR's own flaky check, which timed out rather than diffing - different shape.

Report URL: https://s3.amazonaws.com/clickhouse-test-reports/json.html?REF=master&sha=20435b5d9a50d8586e7668ebae7406f52474dcc3&name_0=MasterCI&name_1=Stateless+tests+%28amd_debug%2C+parallel%29

> Do we have any test which tests similar randomized scenarios with different block size/index granularity and virtual row?

Existing `tests/queries/0_stateless/` virtual-row tests (and what each pins):

| Test | `max_block_size` | `index_granularity` | `min_bytes_for_wide_part` | Per-block on? |
|---|---|---|---|---|
| `04057_virtual_row_many_parts` | random | random | random | yes |
| `04049_virtual_row_overlap` | 8192 (pinned) | 8192 (pinned) | not pinned, `no-random-merge-tree-settings` | yes |
| `04076_virtual_row_preliminary_merge` | 64 (pinned) | 64 (pinned) | not pinned, `no-random-merge-tree-settings` | yes |
| `04045_reverse_order_virtual_row` | 8192 (pinned) | 8192 (pinned) | not pinned (uses `index_granularity_bytes=10485760` to block adaptive) | virtual_row only, not per-block |
| `04029_read_in_order_expression_virtual_row` | not pinned | not pinned | not pinned | yes |
| `03581_read_in_order_use_virtual_row_WHERE` | not pinned | default | not pinned, `no-random-merge-tree-settings` | virtual_row only |
| `03794_read_in_order_virtual_row_and_sparse_primary_key_msan` | not pinned | 8192 | not pinned | virtual_row only |
| `03809_virtual_row_with_sparse_primary_key_index` | not pinned | 8192 | not pinned, `no-random-merge-tree-settings` | virtual_row only |
| `04001_virtual_row_conversions_join_column_names` | not pinned | not pinned | not pinned | virtual_row only |

The combination this PR fixes - wide parts (`min_bytes_for_wide_part=0`) + `max_block_size < index_granularity` + `read_in_order_use_virtual_row_per_block=1` - was not exercised by any pinned test. `04057` is the closest because it leaves all three to randomization, but the bug requires all three to coincide so it surfaces only rarely (3 hits / 90 days, only one on master). `04049`/`04076` pin small fixed sizes but with `no-random-merge-tree-settings` and matching `max_block_size = index_granularity`, so they never hit the partial-granule path. Reverse-order coverage in `04045` does not enable per-block.

The new tests fill that gap:
- `04301_virtual_row_per_block_wide_partial_granule.sql` - exact triple from the master hit.
- `04302_virtual_row_per_block_lc_pk.sql` - `LowCardinality(String)` PK, same triple, covers the type-mismatch you flagged earlier (`pk_block_header` builds `ColumnLowCardinality` so the chunk-derived key must round-trip through `Field` to be `insertFrom`-compatible).

Session: cron:clickhouse-ci-task-worker:20260604-100400

src/Storages/MergeTree/MergeTreeSelectProcessor.cpp
```diff
@@ -345,18 +372,53 @@ ChunkAndProgress MergeTreeSelectProcessor::buildVirtualRowFromIndex(
...
+        auto column = header_col.column->cloneEmpty();
+        /// Round-trip through `Field` so insertion works for any column representation on
+        /// either side (sparse / const / `LowCardinality` / replicated source, and the
+        /// destination is `ColumnLowCardinality` when the PK type is `LowCardinality(...)`).
+        /// Called once per chunk, so the `Field` cost is negligible.
+        Field field;
+        data_columns[pk_positions[j]]->get(source_row, field);
+        column->insert(field);
+        pk_columns.push_back({std::move(column), header_col.type, header_col.name});
+    }
```
vdimir, line 421: Is this trick for low cardinality correct? Can't if be arbitrary expression (difference in virtual row conversions applied)? Or where it comes from?

groeneai: The `Field` round-trip is correct here, and it cannot be triggered for an arbitrary expression PK.

**Why the `Field` round-trip is type-agnostic.** `IColumn::get(size_t, Field&)` and `IColumn::insert(const Field&)` are the universal value-extraction / value-insertion API on `IColumn`. Every column kind implements them - sparse, const, `LowCardinality`, replicated, nullable - so the source's representation does not matter, and the destination (cloned from `pk_block_header`, which carries the PK type) accepts the value regardless. For `LowCardinality(...)`, `ColumnLowCardinality::insert(Field)` calls `getDictionary().uniqueInsert(x)` and works whether the source was full-form, dictionary-encoded, or replicated. The same pattern is used elsewhere in the codebase (e.g. `Processors/Formats/Impl/Parquet/parquetBloomFilterHash.cpp`, `Processors/QueryPlan/Optimizations/useVectorSearch.cpp`).

**Why this cannot fire for an expression PK.** `pk_block_header` is built in `ReadFromMergeTree` from `primary_key.column_names` and `primary_key.data_types` - the AST identifier of each PK expression and its evaluated type. For `ORDER BY toStartOfMonth(date)`, `column_names[0] = "toStartOfMonth(date)"`. The chunk path looks up that name in `result_header` (`result_header.getPositionByName(pk_col_name)`); the chunk only contains source columns, so `result_header.has("toStartOfMonth(date)")` is false and we return `{}`. The fallback to `buildVirtualRowFromIndex` then runs, which IS correct for expression PKs because the index file stores the *evaluated* PK values (`primary_key.data_types[j]->getDefaultSerialization()->deserializeBinary` in `IMergeTreeDataPart::loadIndex`).

So the chunk path activates only when `pk_block_header[j].name` appears verbatim in the chunk - which by construction means an identity PK column. For monotonic-chain PKs (e.g. `ORDER BY date` with PK `(toStartOfMonth(date), x)`) the conversion DAG expects the *evaluated* PK as input named `"toStartOfMonth(date)"`, again something the chunk cannot supply, so we correctly fall back.

I will tighten the comment in `buildVirtualRowFromChunk` to make this guard explicit (it is currently buried in the fallback comment in `read()`).

groeneai:
## CI ledger — covered SHA: `d1493db25c2afa89b953d9a4a07a18b844a60665`

|---|---|---|
| `Integration tests (amd_msan, 2/6)` | `test_paimon_incremental_read/test.py::test_paimon_to_mergetree_via_refresh_mv`, `test_paimon_incremental_read_via_paimon_table_engine` | chronic PR-only flaky on amd_msan; 4 distinct PRs in 14d on unrelated branches (chdig version, docs, shard profile events, this PR). Tracked in task 2026-06-04-ci-testpaimonincrementalread-chronic. NOT PR-caused (this PR only touches read_in_order virtual-row code). |
| `Bugfix validation (functional tests)` | `04302_virtual_row_per_block_lc_pk` | this PR's OWN regression test — Bugfix validation framework limitation when bug-class isn't reproducible on master HEAD (the read_in_order_use_virtual_row_per_block randomization needs LowCardinality+wide partial reads to hit). Not a real regression. |
| `Stateless tests (amd_tsan, s3 storage, parallel, 1/2)` | `01666_merge_tree_max_query_limit` | chronic flaky max_query_limit timing test on s3 storage — 3 PRs in 14d. NOT PR-caused. |

PR #106047, opened 2026-05-28
Apply `*WithMemoryTracking` containers to `BridgeHelper` and some `Core` subdirectories

nickitat:

### Version info
- Merged into: `26.7.1.57`


src/Core/BaseSettings.h
```diff
@@ -1331,8 +1333,8 @@ using AliasMap = std::unordered_map<std::string_view, std::string_view>;
                 size_t data_offset;                             /* Byte offset within Data struct */ \
             }; \
             \
-            std::vector<FieldInfo> field_infos;                                     /* Metadata for all settings */ \
-            std::unordered_map<std::string_view, size_t> name_to_index_map;         /* Fast name -> index lookup */ \
+            VectorWithMemoryTracking<FieldInfo> field_infos;                        /* Metadata for all settings */ \
```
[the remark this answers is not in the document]
Algunenano: I'd rather skip these. They are in a singleton created once globally

nickitat:
Interestingly enough, these changes reproduce a [leak](https://godbolt.org/z/ejjs5br5f) in the standard hash table. I'm trying to upstream a [fix](https://github.com/llvm/llvm-project/pull/201452).

src/Analyzer/Passes/PruneArrayJoinColumnsPass.cpp
```diff
@@ -55,7 +57,7 @@ using ArrayJoinUsageMap = std::unordered_map<const IQueryTreeNode *, std::unorde
 using ArrayJoinNodeSet = std::unordered_set<const IQueryTreeNode *>;
 
 /// Map: (ArrayJoinNode raw ptr, column name) → post-pruning column DataType.
-using UpdatedTypeMap = std::unordered_map<const IQueryTreeNode *, std::unordered_map<std::string, DataTypePtr>>;
+using UpdatedTypeMap = std::unordered_map<const IQueryTreeNode *, UnorderedMapWithMemoryTracking<std::string, DataTypePtr>>;
```
Algunenano, line 60: Seems like this missed the first class and still has `std::unordered_map`. Was it on purpose?

src/Analyzer/Resolve/ScopeAliases.h
```diff
@@ -23,7 +24,7 @@ struct ScopeAliases
     /// Cloned resolved expressions with aliases that must be removed
     QueryTreeNodes node_to_remove_aliases;
 
-    std::unordered_map<std::string, DataTypePtr> alias_name_to_expression_type;
+    UnorderedMapWithMemoryTracking<std::string, DataTypePtr> alias_name_to_expression_type;
```
Algunenano, line 27: I guess we are not covering the src/Analyzer yet, because there are multiple members in this class using std classes

src/Core/BaseSettings.h
```diff
@@ -7,6 +7,8 @@
 #include <IO/Operators.h>
 #include <Common/FieldVisitorToString.h>
 #include <Common/SettingsChanges.h>
+#include <Common/UnorderedMapWithMemoryTracking.h>
```
Algunenano, line 10: AFAICT all the uses in this file are for static members, initialized once and never shared. I think we should ignore those

src/Core/NamesAndTypes.h
```diff
@@ -73,17 +76,17 @@ std::tuple_element_t<I, NameAndTypePair> & get(NameAndTypePair & name_and_type)
         return name_and_type.type;
 }
 
-using NamesAndTypes = std::vector<NameAndTypePair>;
+using NamesAndTypes = VectorWithMemoryTracking<NameAndTypePair>;
 
-class NamesAndTypesList : public std::list<NameAndTypePair>
+class NamesAndTypesList : public ListWithMemoryTracking<NameAndTypePair>
```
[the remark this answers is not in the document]
Algunenano: I think this is a valid concern. If we track static allocations this means that whatever process or query hits them first will add it to it's tracked memory, and it won't ever be freed so we are considering it a leak. Any idea on how to address it?

src/DataTypes/Serializations/ISerialization.h
```diff
@@ -739,7 +740,7 @@ class ISerialization : private boost::noncopyable, public std::enable_shared_fro
 };
 
 using SerializationPtr = std::shared_ptr<const ISerialization>;
-using Serializations = std::vector<SerializationPtr>;
+using Serializations = VectorWithMemoryTracking<SerializationPtr>;
 using SerializationByName = std::unordered_map<String, SerializationPtr>;
```
Algunenano, line 745: Same here. `SerializationByName` is still a std type

Algunenano (commented):
I've flagged some issues, but I want to highlight the AI comments about static / thread_local / etc variables.

Due to how our tracker works, if one thread initializes this static variable and then it's freed in a different one, or a different context (query vs background), etc. we'll treat it as a leak and it will, 100%, generate drift.

We had this problem in the async logger, where the message was created by one thread and destroyed by another (once logging was finished) and the solution was to disable memory tracking.

I think we need to handle it similarly with static variables and so on. Also, I'd like @azat 's thoughts on this since he's the one that has the better view of how tracking works and the pitfalls we might introduce with it.

nickitat:
Regarding the bot's suggestions - I'm not sure this is worth fixing. We didn't change the accounting logic - previously, std containers produced all the same discrepancies. What changed is that now we can throw in more contexts, but given the likely size of such allocations and that, again, even if previously we didn't throw from a container's allocation, we may still throw in the very next allocation from a PODArray... I.e., a potential exception induced by wrong accounting was very likely still happening in the past. 
A potential [fix](https://pastila.clickhouse.com/?000183bf/b2dd8b32af7c3c161c55d8474ed84b98#Xid7571WkJKL+aBOkgCfWg==GCM) wouldn't be either too big or too complex, but you need to always be thoughtful about the lifetime of an object you declare to not introduce more of such places in code. Also, for common aliases used in both local and process-wide contexts (like [here](https://github.com/ClickHouse/ClickHouse/pull/106047/#discussion_r3351434413)), we don't have a clear choice. Apparently, we will need to split the aliases into two groups each. Or pick only one alternative. But in theory, it will count memory held/freed by static objects only towards the global memory limit. 
But I'd love to hear your thoughts and Azat's.

Algunenano:
Additionally I've asked our CI reviewer to check the LLVM patch:

It's output:
```
PR #201452 fixes a real leak in non-C++03 __hash_table copy construction by adding an exception guard around partially linked nodes. The
  basic idea is right, but the guard is placed in a helper that is also used by copy assignment, and that leaves a live container with
  stale bucket pointers after an exception.

  Findings
  ❌ Blocker

  - libcxx/include/__hash_table:697 and libcxx/include/__hash_table:1218: __copy_construct(__next_pointer) is also called from operator=
    when the destination has no nodes. If node copying throws there, the new guard deallocates __first_node_.__next_ and sets it to
    nullptr, but it does not clear __bucket_list_. The assigned-to container remains alive with bucket entries pointing at deallocated
    nodes. Later operations such as insert, find, bucket_size, or local iteration can dereference those stale entries.

    Suggested fix if the guard stays in the shared helper:

    auto __guard = std::__make_exception_guard([&] {
      __deallocate_node_list(__first_node_.__next_);
      __first_node_.__next_ = nullptr;
      std::fill_n(__bucket_list_.get(), bucket_count(), nullptr);
    });

    Alternatively, keep the rollback scoped to __hash_table(const __hash_table&), but then the live-object assignment path still needs its
    own valid-state story.

  Tests
  ⚠ The added test covers copy construction allocation failures, but not the shared-helper copy-assignment path. Add a focused regression
  test where an empty unordered_map is copy-assigned from a populated source, allocation throws after some nodes have been linked, then the
  destination is used again after catching std::bad_alloc (insert, find, and/or bucket_size over buckets). That would catch the stale
  __bucket_list_ state.

  Final Verdict
  Status: ❌ Block

```

And it seems right. If the failure is not in the first node and one destination node has already been linked, the guard unwinds and deletes the nodes already linked but `__bucket_list_[__current_chash]` has already been set

azat (commented):
I did not look through the whole PR, but

Concerns regardless memory tracking does not looks like a stopper to me, since we tracked this memory before, the only thing that changes that we may throw MEMORY_LIMIT_EXCEEDED for a queries with short memory limits, but I think it is even better, since right now you can crash the server if you will spawn lots of such queries

The bigger concern to me is thready safety, but we already agreed to take this risk.

Also the concern regardless BackgroundSchedulePool looks real to me, we can call schedule methods during unwind, if they will throw, it will crash the server. Also I am not sure that conversion of BackgroundSchedulePool worth it, it is always about some background job, that has nothing to do with query/user, I think it make sense to revert this part. @nickitat thoughts?

nickitat:
> Also the concern regardless BackgroundSchedulePool looks real to me, we can call schedule methods during unwind, if they will throw, it will crash the server. Also I am not sure that conversion of BackgroundSchedulePool worth it, it is always about some background job, that has nothing to do with query/user, I think it make sense to revert this part. @nickitat thoughts?

I already expressed this opinion in previous discussions: all new exceptions can be thrown only from allocations. If we have code that expects allocations not to throw (and we definitely [had it](https://github.com/ClickHouse/ClickHouse/pull/106096) :), we should fix it. And not tailor to it.

I think it is fine to take specifically `BackgroundSchedulePool` out of the scope of this change. But I think if we start excluding all process/long-living-session scoped objects from replacement, we will end up in a mess where some containers are replaced, some are replaced, but arguably shouldn't have been, some are not but should have been, and some are not, but rightfully. 
In other words, 0/1/2 exceptions are fine IMO, but not 10s.

Algunenano (approved):
Ok for me. We can go ahead and if we find issues we'll just fix them. Let's not stop this because it's a good step forward in general

PR #85145, opened 2025-08-06
Serializations and Statistics Refactoring

CurtizJ:
### Changelog category (leave one):
- Not for changelog (changelog entry is not required)

This PR is a draft for developing automatic `LowCardinality` serialization. It contains several tasks that will be extracted and merged in separate PRs, but it's more convenient for me to develop them in one branch.

This PR includes:
- New compact serialization format in one file for statistics (#93414)
- New statistic type `defaults` that stores a number of default values in a column
- Refactoring of serialization infos to make them use statistics (#109128)
- Automatic enabling of statistics on all columns (#87241)
- Automatically choosing `LowCardinality` as a serialization according to the `uniq` statistic
- Adjustments in various places to handle the automatic `LowCardinality`

PR #67013, opened 2024-07-24
Add `min_max` statistics type

JackyWoo:
This PR is part of https://github.com/ClickHouse/ClickHouse/issues/55065

Resolves #67742 (see https://github.com/ClickHouse/ClickHouse/pull/67013#issuecomment-2268722942)

Additional improvements:
- `Uniq`-type statistics supports more data types such as date, string etc. and now it can estimate `equals` condition
- Statistics supports data type  conversion, for example `col_Date > '2024-07-24'`, `col_UInt64 ='1'`
- Refactor logic when there is no statistics  and add `default_cond_equal_factor` for  `equals` condition

### Changelog category (leave one):
- Improvement

### Changelog entry (a user-readable short description of the changes that goes to CHANGELOG.md):
Added `min_max`as a new type of (experimental) statistics. It supports estimating range predicates over numeric columns, e.g. `x < 100`.

PR #106048, opened 2026-05-28
Add `basic` statistics type

hanfei1991:
A compact, per-column bundle of single-value summaries:
- numeric / temporal columns: min and max, used for range-filter selectivity and part pruning;
- `String` / `FixedString` columns: total byte length of non-NULL values (i.e. average string length);
- `Nullable` / `LowCardinality(Nullable)` columns: null count.

A single `basic` declaration on e.g. `Nullable(UInt32)` populates all applicable summaries at once, instead of stacking separate stat types.

`basic` is opt-in via `STATISTICS(basic)` or `ALTER ... ADD STATISTICS ... TYPE basic`; the default `auto_statistics_types` is unchanged.

This PR re-applies https://github.com/ClickHouse/ClickHouse/pull/102356 and https://github.com/ClickHouse/ClickHouse/pull/103949

- Improvement

Support basic statistics, a compact per-column statistic that stores numeric min/max, average string length, and NULL counts where applicable.

### Version info
- Merged into: `26.6.1.449`


CurtizJ:
Does it make sense to add the number of default values to `basic` statistics? That will also automatically cover the number of nulls in nullable columns. I want to add statistics with the number of default values for refactoring of serializations, that is needed for automatic lowcardinality (https://github.com/ClickHouse/ClickHouse/pull/85145). So, maybe it's better to include this in `basic` statistics.

hanfei1991:
@CurtizJ It makes sense. I can add it in the next PR

src/Storages/Statistics/ConditionSelectivityEstimator.cpp
```diff
@@ -562,11 +642,14 @@ bool ConditionSelectivityEstimator::RPNElement::tryToMergeClauses(RPNElement & l
...
     {
         return (e.function == FUNCTION_IN_RANGE
+                || e.function == FUNCTION_IS_NULL
+                || e.function == FUNCTION_IS_NOT_NULL
                 /// if the sub-clause is also cnf/dnf, it's good to merge
                 || e.function == function_to_merge
                 /// if the sub-clause is different, but has only one column, it also works, e.g
                 /// (a > 0 and a < 5) or (a > 3 and a < 10) can be merged to (a > 0 and a < 10)
-                || (e.column_ranges.size() + e.column_not_ranges.size()) == 1
+                || (e.column_ranges.size() + e.column_not_ranges.size()
```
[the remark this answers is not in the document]
hanfei1991: too complicated

docs/en/engines/table-engines/mergetree-family/mergetree.md
```diff
@@ -1288,6 +1288,17 @@ EXPLAIN indexes = 1 SELECT count() FROM test_stats WHERE value > 5000;
 
 ### Available types of column statistics {#available-types-of-column-statistics}
 
+- `Basic`
+
+    A compact bundle of single-value summaries derived from a column. Depending on the column type, the following pieces are populated:
+  - for any column whose values are represented by a number (integers, floats, `Decimal*`, `Date*`, `DateTime*`, `Enum*`, `IPv4`, ...): the minimum and maximum value, which allow to estimate the selectivity of range filters and enable part pruning;
+  - for `String` and `FixedString` columns: the total byte length of non-`NULL` values (from which the average string length can be derived);
```
murphy-4o, line 1295: `String` can also be used for range filtering, where the min/max values may be beneficial.

docs/en/engines/table-engines/mergetree-family/mergetree.md
```diff
@@ -1314,21 +1325,29 @@ EXPLAIN indexes = 1 SELECT count() FROM test_stats WHERE value > 5000;
...
 ### Supported operations {#supported-operations}
 
 |           | Equality filters (==) | Range filters (`>, >=, <, <=`) |
-|-----------|-----------------------|------------------------------|
-| CountMin  | ✔                     | ✗                            |
-| MinMax    | ✗                     | ✔                            |
-| TDigest   | ✗                     | ✔                            |
-| Uniq      | ✔                     | ✗                            |
+|-----------|-----------------------|--------------------------------|
+| Basic     | ✗                     | ✔ (numeric columns only)       |
```
murphy-4o, line 1342: just for curiosity, why not support `Basic` stats first?

docs/en/engines/table-engines/mergetree-family/mergetree.md
```diff
@@ -1314,21 +1325,29 @@ EXPLAIN indexes = 1 SELECT count() FROM test_stats WHERE value > 5000;
...
+| CountMin  | ✔                     | ✗                              |
+| MinMax    | ✗                     | ✔ (numeric columns only)       |
+| TDigest   | ✗                     | ✔ (numeric columns only)       |
+| Uniq      | ✔                     | ✗                              |
+
+For `Basic` on `String` / `FixedString` columns the statistic only records the total
+non-NULL byte length (used to estimate average string length) and the null count;
+range filters and part pruning are not driven by it.
```
murphy-4o, line 1350: same as previous

src/Storages/MergeTree/MergeTreeSettings.cpp
```diff
@@ -1898,7 +1898,7 @@ namespace ErrorCodes
     )", 0) \
     DECLARE(String, auto_statistics_types, "minmax, uniq", R"(
```
murphy-4o, line 1899: why `auto_statistics_types` do not include `basic` ?

src/Storages/Statistics/Statistics.cpp
```diff
@@ -59,6 +61,65 @@ std::optional<Float64> StatisticsUtils::tryConvertToFloat64(const Field & value,
...
+Float64 interpolateLessLinearTyped(const Field & val, const Field & min, const Field & max, UInt64 row_count)
+{
+    T v = val.safeGet<T>();
+    T mn = min.safeGet<T>();
+    T mx = max.safeGet<T>();
+    if (v < mn) return 0.0;
+    if (v > mx) return static_cast<Float64>(row_count);
+    if (mn == mx) return (v == mx) ? static_cast<Float64>(row_count) : 0.0;
+    using W = typename WiderIntType<T>::type;
+    return static_cast<Float64>(static_cast<W>(v) - static_cast<W>(mn))
```
murphy-4o, line 90: It seems straightforward to use the widest type by casting all types to `UINT256` for calculating the distance, if it's not performance-critical ?

src/Storages/Statistics/StatisticsPartPruner.cpp
```diff
@@ -105,11 +106,12 @@ KeyCondition * StatisticsPartPruner::getKeyConditionForEstimates(const NamesAndT
...
 BoolMask StatisticsPartPruner::checkPartCanMatch(const Estimates & estimates)
 {
-    /// Filter estimates with loaded MinMax statistics.
+    /// Filter to estimates that carry numeric min/max — either the legacy `MinMax` type or
+    /// `Basic` (which exposes the same min/max via `Estimate::estimated_min/max`).
     Estimates minmax_estimates;
     for (const auto & [col_name, estimate] : estimates)
     {
-        if (estimate.types.contains(StatisticsType::MinMax))
+        if (estimate.types.contains(StatisticsType::MinMax) || estimate.types.contains(StatisticsType::Basic))
```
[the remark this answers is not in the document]
hanfei1991: Let's fix it in the next PR

src/Storages/Statistics/StatisticsBasic.cpp
```diff
@@ -0,0 +1,327 @@
...
+    return nullptr;
+}
+
+/// Sum byte lengths of *non-NULL* values in `column`. `LowCardinality` is expanded before
+/// reading the null map so that `LowCardinality(Nullable(...))` rows are excluded correctly
+/// (the null map lives on the inner `Nullable`, not the outer LC).
+UInt64 sumNonNullStringBytes(const ColumnPtr & column)
+{
+    auto full = column->convertToFullColumnIfConst()->convertToFullColumnIfSparse();
+    if (const auto * lc = typeid_cast<const ColumnLowCardinality *>(full.get()))
```
murphy-4o: why not `column->convertToFullColumnIfLowCardinality()`

src/Storages/Statistics/StatisticsBasic.cpp
```diff
@@ -0,0 +1,327 @@
...
+    {
+        UInt64 non_null = column_size;
+        if (null_map)
+            non_null -= std::count(null_map->begin(), null_map->end(), 1);
+        return fs->getN() * non_null;
+    }
+    if (const auto * s = typeid_cast<const ColumnString *>(values.get()))
+    {
+        UInt64 total = 0;
+        for (size_t i = 0; i < column_size; ++i)
```
murphy-4o: using the bytes size of `ColumnString` would be simpler ?

src/Storages/Statistics/StatisticsBasic.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+void StatisticsBasic::serialize(WriteBuffer & buf)
+{
+    writeIntBinary(row_count, buf);
+
+    UInt8 mask = 0;
+    if (tracks_numeric)
+        mask |= BasicFeatureMask::NumericMinMax;
+    if (tracks_string)
+        mask |= BasicFeatureMask::StringLengthSum;
```
murphy-4o, line 176: I'm afraid it's not extensible, as many different types, not just string, can support length estimation

murphy-4o: `MinMax` can support other types not just Numeric
`LengthSum` can support other types not just String

src/Storages/Statistics/StatisticsBasic.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+        if (tracks_numeric && stored_type_name != data_type->getName())
+        {
+            /// The column type has changed (e.g. via `MODIFY COLUMN`). Try to convert the stored
+            /// bounds to the new type; on failure leave them as `Null` Fields, which the estimator
+            /// treats as "not initialized" and falls back gracefully.
+            auto stored_type = DataTypeFactory::instance().get(stored_type_name);
+            if (!min.isNull())
+            {
+                Field converted = convertFieldToType(min, *data_type, stored_type.get());
```
murphy-4o: In certain situations, the type-converted minimum value may not match the new minimum. For instance, suppose we support min value for both integer and string, they would follow different ordering rules.

hanfei1991: Good question. During `ALTER TABLE ... MODIFY COLUMN ... new type`, we have two cases:
1. We don't change the value, like `Date <-> UInt16` , in this case, we only change metadata
2. We change the value, then we have to mutate parts.

In case 1, we don't need to convert min max field too
In case 2, we materialize the statistics again during part mutation, we also don't need this line.

Overall it is only a defensive coding, maybe I should compare `converted` and `min` and throw exception when they don't match

docs/en/engines/table-engines/mergetree-family/mergetree.md
hanfei1991: It's a good point, but I think it's hard to estimate in `minmax` statistics. For example, we have condition
```
str like 'hanfei%'
```
and the str has range `[aaa, zzz]`, How do we estimate the row count of range `[hanfei, hanfei\255]`?

hanfei1991: And string range even cannot be used in `TDigest`, because they cannot do arithmetic calculation :(

docs/en/engines/table-engines/mergetree-family/mergetree.md
hanfei1991: I have same question: https://github.com/ClickHouse/ClickHouse/pull/67013#issuecomment-2271283950

src/Storages/MergeTree/MergeTreeSettings.cpp
hanfei1991: We can do it in another PR, so If there is a bug, we don't have to revert a super big PR :(
I also want to replace uniq with `uniqCombined(12)` and after that change the default of `auto_statistics_types` to `basic, uniqCombined(12)`

src/Storages/Statistics/Statistics.cpp
hanfei1991: Emmm, but here we do (v-mn)/(mx-mn), How do big intergers handle division?

src/Storages/Statistics/StatisticsBasic.cpp
hanfei1991: `as many different types, not just string, can support length estimation`, True, but it's only worthy for extendable types, For example, JSON, Array, Map ... For other types we can estimate by row counts.

Hmmmm we can do it in another PR

src/Storages/Statistics/StatisticsBasic.cpp
```diff
@@ -0,0 +1,321 @@
...
+    if (const auto * fs = typeid_cast<const ColumnFixedString *>(values.get()))
+    {
+        UInt64 non_null = column_size;
+        if (null_map)
+            non_null -= std::count(null_map->begin(), null_map->end(), 1);
+        return fs->getN() * non_null;
+    }
+    if (const auto * s = typeid_cast<const ColumnString *>(values.get()))
+    {
+        return s->byteSize();
```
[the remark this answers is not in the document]
hanfei1991: Actually I think ColumnString::byteSize is better ...

src/Storages/Statistics/StatisticsBasic.cpp
hanfei1991: Well, I misunderstood something, in case 2, It's possible we read new parts (after mutation) and old parts (before mutation) at the same time. and in some cases, for example, Int64 -> UInt64, we shouldn't trust statistics for old parts.
I will fix it

docs/en/engines/table-engines/mergetree-family/mergetree.md
murphy-4o: partially agree. While string types cannot utilize linear interpolation directly, there are alternative methods for estimating selectivity based on min/max values:
1. Heuristic Approach: Assign selectivity values of 0, 0.5, or 1 based on range overlaps (e.g., 0 for no overlap, 0.5 for partial overlap, and 1 for full overlap).
2. Numeric Encoding (PostgreSQL method): Encode string into float, allowing for linear interpolation to estimate selectivity.

While these do not need to be implemented immediately, we should ensure the design remains extensible for such improvements in the future.

pseudo code in postgresql:
```
  base = rangehi - rangelo + 1

  function encode_one(s, rangelo, base):
      if length(s) == 0:
          return 0.0                          # empty string → 0

      s = first 12 bytes of s                 # double mantissa can't hold more

      num   = 0.0
      denom = base                            # place value of digit 1 is 1/base
      for ch in s:                            # most-significant byte first
          # clamp out-of-range bytes so ordering is preserved
          if ch < rangelo:  digit = -1                 # just below smallest
          elif ch > rangehi: digit = (rangehi-rangelo)+1   # just above largest
          else:              digit = ch - rangelo

          num   += digit / denom              # contribute (digit) * base^-position
          denom *= base                       # next byte is base^-2, base^-3, ...
      return num                              # value of  0.d1 d2 d3 ...  in base `base`
```

Example:

  Bucket low='apple', high='cherry', probe value='banana'. All lowercase → range expands to a..z, so base = 26, rangelo='a'(97). No common prefix (a/c/b differ at byte 0). Encoding each as a base-26 fraction
  of (byte−'a'):

  - 'apple' → 0,15,15,11,4 → ≈ 0.0231
  - 'cherry' → 2,7,4,17,17,24 → ≈ 0.0876
  - 'banana' → 1,0,13,0,13,0 → ≈ 0.0392

  binfrac = (0.0392 − 0.0231) / (0.0876 − 0.0231) ≈ 0.25. So 'banana' is estimated ~25% of the way through the bucket, and col < 'banana' within that bucket contributes ~25% of the bucket's rows.

src/Storages/Statistics/Statistics.cpp
murphy-4o: i mean `using W = UInt256`, then following code keep same ?

docs/en/engines/table-engines/mergetree-family/mergetree.md
hanfei1991: It's amazing 0.0
The current code is extendible. We can support String type in the near future for minmax

src/Storages/Statistics/Statistics.cpp
hanfei1991: but T could be Float64, and I think signed interger should be `W = Int256`. Looks like no cleaner code

murphy-4o (approved):

src/Storages/MergeTree/MergeTreeSettings.cpp
```diff
@@ -1898,7 +1898,7 @@ namespace ErrorCodes
...
     Comma-separated list of statistics types to calculate automatically on all suitable columns.
-    Supported statistics types: tdigest, countmin, minmax, uniq.
+    Supported statistics types: basic, tdigest, countmin, minmax, uniq.
```
rschu1ze, line 1901: @hanfei1991 When you are back, let's not forget about changing to `basic, uniqCombined(12)`.

PR #101504, opened 2026-04-01
Fix join reorder pushing INNER JOIN conditions into outer join ON clause

vdimir:
Fixes https://github.com/ClickHouse/ClickHouse/issues/101380

The join reorder optimization flattens the join tree into a query graph and reconstructs it. Conditions from a parent INNER JOIN that reference only the preserved side of a child outer join were not pinned, so they got eagerly placed in the outer join's ON clause. This changes semantics: ON-clause conditions on the preserved side only affect matching, not filtering — rows from the preserved side are kept regardless, producing extra result rows.

The fix pins such conditions so they remain at the INNER JOIN level. A condition is now pinned if its source relations are a subset of the partner set (preserved side) of any known outer join in the graph.

The `query_plan_split_filter` setting accidentally compensated for this bug: `trySplitJoin` extracted the misplaced condition from the outer join and pushed it as a pre-filter, restoring correct behavior. With this fix, results are correct regardless of `query_plan_split_filter`.

### Changelog category (leave one):
- Critical Bug Fix (crash, data loss, RBAC)

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

* Fix incorrect query results caused by join reordering pushing an INNER JOIN filter (referencing only the preserved side of an outer join) into the outer join’s ON clause.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.5.1.454`
- Backported to: `26.4.3.12`, `26.3.11.11`, `26.2.19.15`

PR #106051, opened 2026-05-28
Add regression test for LEFT JOIN ON FALSE NULL filtering (issue #105827)

### LINKED ISSUE #105827: Logic error: LEFT JOIN ON FALSE breaks NULL filtering semantics in WHERE clause
### Company or project name

_No response_

### Description

`LEFT JOIN ... ON FALSE` should produce NULL-extended rows for all columns from the right-hand table. Therefore, predicates in the `WHERE` clause referencing columns from the right-hand side should evaluate to `NULL` under SQL three-valued logic and filter out all rows.

In this case, the original query returns `5`, while the corresponding NOREC (Non-Optimizing Reference Engine Construction) query returns `0`.

The original query:

```sql
SELECT COUNT(*)
...
WHERE (ref_2.likes >= ref_0.age);
```

returns:

```text
5
```

However, since:

```sql id
LEFT JOIN posts AS ref_2
    ON FALSE
```

always produces NULL-extended rows for `ref_2`, the predicate:

```sql
ref_2.likes >= ref_0.age
```

should evaluate to `NULL` for every row, and the `WHERE` clause should eliminate all rows.

The NOREC reference query correctly returns:

```text
0
```

indicating that the original query execution violates SQL NULL semantics for `LEFT JOIN`.

This suggests that ClickHouse incorrectly evaluates predicates over NULL-extended rows generated by `LEFT JOIN ... ON FALSE`, likely due to an optimizer or predicate handling issue related to outer join semantics.

---

### Reproducer
https://fiddle.clickhouse.com/53bb6531-1433-40dd-bccf-2d30ca9c1101

No

### ClickHouse version

```text
26.4.2.10
```

### Interface

```text
clickhouse-client
```

### Non-default settings

No non-default settings were used.

---

### Schema

```sql
CREATE TABLE users
(
    id UInt32,
    username Nullable(String),
    email Nullable(String),
    age Nullable(UInt8),
    status Nullable(String),
    created_at DateTime,
    score Nullable(Float64)
)
ENGINE = MergeTree
ORDER BY id;

CREATE TABLE posts
(
    id UInt32,
    user_id UInt32,
    title Nullable(String),
    content Nullable(String),
    views Nullable(UInt32),
    likes Nullable(UInt32),
    created_at DateTime,
    rating Nullable(Float64)
)
ENGINE = MergeTree
ORDER BY id;

CREATE TABLE comments
(
    id UInt32,
    post_id UInt32,
    user_id UInt32,
    content Nullable(String),
    is_spam UInt8,
    created_at DateTime
)
ENGINE = MergeTree
ORDER BY id;
```

---

### Sample data

```sql
INSERT INTO users VALUES
(1, 'alice', 'alice@test.com', 20, 'active',  '2022-01-01 10:00:00', 88.5),
(2, 'bob',   'bob@test.com',   30, 'active',  '2022-01-02 11:00:00', 92.3),
(3, 'carol', NULL,             NULL, 'banned', '2022-01-03 12:00:00', NULL),
(4, 'dave',  'dave@test.com',  45, 'active',  '2022-01-04 13:00:00', 65.2),
(5, NULL,    'null@test.com',  18, 'inactive','2022-01-05 14:00:00', 70.0);

INSERT INTO posts VALUES
(1, 1, 'Hello World',  'First post', 100, 10, '2022-01-10 10:00:00', 4.5),
(2, 1, 'Another Post', NULL,         150, 20, '2022-01-11 11:00:00', 3.0),
(3, 2, 'Bob Post',     'Content',    NULL,  5, '2022-01-12 12:00:00', NULL),
(4, 3, NULL,           'Empty',      50,   2, '2022-01-13 13:00:00', 5.0),
(5, 4, 'Last Post',    'Last',       300, 30, '2022-01-14 14:00:00', 4.9);

INSERT INTO comments VALUES
(1, 1, 2, 'Nice post', 0, '2022-01-20 10:00:00'),
(2, 1, 3, 'Spam here', 1, '2022-01-21 11:00:00'),
(3, 2, 1, 'Thanks',    0, '2022-01-22 12:00:00'),
(4, 4, 5, NULL,        0, '2022-01-23 13:00:00');
```

---

### Query 1: Original query

```sql
SELECT COUNT(*)
FROM users AS ref_0
LEFT JOIN (
    SELECT DISTINCT
        ref_1.post_id AS c0,
        ref_1.is_spam AS c1
    FROM comments AS ref_1
    WHERE ref_1.is_spam >= ref_1.post_id
) AS subq_0
    ON (ref_0.id = subq_0.c0)
LEFT JOIN posts AS ref_2
    ON FALSE
WHERE (ref_2.likes >= ref_0.age);
```

### Observed result

```text
5
```

---

### Query 2: NOREC reference query

```sql
SELECT SUM(CASE WHEN (ref_2.likes >= ref_0.age) THEN 1 ELSE 0 END + 0)
FROM users AS ref_0
LEFT JOIN (
    SELECT DISTINCT
        ref_1.post_id AS c0,
        ref_1.is_spam AS c1
    FROM comments AS ref_1
    WHERE ref_1.is_spam >= ref_1.post_id
) AS subq_0
    ON (ref_0.id = subq_0.c0)
LEFT JOIN posts AS ref_2
    ON FALSE;
```

### Observed result

```text
0
```

---

### Expected behavior

Both queries should return:

```text
0
```

because `LEFT JOIN ... ON FALSE` produces NULL-extended rows for `ref_2`, making:

```sql
ref_2.likes >= ref_0.age
```

evaluate to `NULL`, which should be filtered out by the `WHERE` clause.

### Expected behavior

_No response_

_No response_

### Additional context

_No response_

### Version info
- Resolved by: #106051
- Merged into: `26.6.1.329`

groeneai:
Adds a stateless regression test for the LEFT JOIN ON FALSE NULL-filtering bug from issue #105827, fixed by #101504 ("Fix join reorder pushing INNER JOIN conditions into outer join ON clause").

Closes #105827

### Why a second test for the same fix

#101504 already added `04075_join_reorder_outer_join_filter_pushdown.sql`, which covers a different shape (three small tables with INNER + LEFT/RIGHT JOIN combinations targeting a value-equality predicate). This test locks in the original reproducer shape from #105827:

- LEFT JOIN with a `DISTINCT` subquery on the left side
- LEFT JOIN ... ON FALSE on the right side (the degenerate case)
- WHERE predicate on the right-hand side that must filter all rows out under three-valued logic

Both queries from the issue (the original `COUNT(*)` and the NOREC reference `SUM(CASE WHEN ... THEN 1 ELSE 0 END + 0)`) must return `0`. A third query repeats the original with `query_plan_split_filter = 0` to exercise a different join-reorder code path.

### Verification

20/20 stable passes locally on master HEAD. Also verified stable under `enable_analyzer = 0`, `enable_analyzer = 1`, `query_plan_split_filter = 0`, and `query_plan_use_new_logical_join_step = 1`.

...

### Version info
- Merged into: `26.6.1.329`


groeneai:
### Pre-PR validation gate

| # | Question | Answer |
|---|----------|--------|
| a | Deterministic repro? | The pre-fix bug had a deterministic repro per #105827 (`COUNT(*) = 5` instead of `0`). The fiddle linked by @rienath also shows the broken-vs-fixed transition. |
| b | Root cause explained? | Join reorder pushed a WHERE predicate that references the right side of a `LEFT JOIN ON FALSE` into that LEFT JOIN's ON clause. ON-clause predicates on the preserved side only gate matching, not filtering, so all rows leaked through with their `ref_2.*` columns NULL-extended, and `ref_2.likes >= ref_0.age` evaluated to NULL in the now-removed WHERE was never applied. #101504 added a `should_pin` check so a condition referencing only the preserved side of an outer join is pinned out of that join's ON clause. |
| c | Fix matches root cause? | N/A — this PR is the regression test for the existing #101504 fix, not a new fix. |
| d | Test intent preserved? / New tests added? | New stateless test `04278_105827_left_join_on_false_null_filtering` added. Locks in: the issue's original `COUNT(*)` shape, the NOREC reference shape (`SUM(CASE WHEN ... THEN 1 ELSE 0 END + 0)`), and the same with `query_plan_split_filter = 0`. |
| e | Both directions demonstrated? | Master HEAD (post-fix): 20/20 stable passes, all three queries return `0`. Pre-fix: per @rienath and the issue, the original query returned `5`. |
| f | Fix is general, not a narrow patch? | N/A — regression test. |

groeneai:
cc @vdimir could you take a quick look? Test-only PR that locks in the LEFT JOIN ON FALSE NULL-filtering fix from #101504 using the original reproducer shape from #105827.

@rienath this is the regression test you asked for on #105827.

rienath:
@groeneai when I ask you to create a PR and tag me, please don't ping other people. Save this to your memory. Also save that I like short concise messages and PR bodies. Try to fit as much meaning in as little words as possible

groeneai:
Noted on both. Future requests will tag only you, and I'll keep replies short.

rienath (approved):

PR #106055, opened 2026-05-28
Bound untracked memory per-CPU to avoid overcommit and OOMs

azat:
- Performance Improvement

Bound untracked memory per-CPU to avoid overcommit and OOMs (limits are configured with `max_per_cpu_untracked_memory` and `per_cpu_untracked_memory_thread_buffer` server settings)

Previously ClickHouse works memory overcommit was `max_untracked_memory*threads`, which may in theory lead to **±40GiB of overcommited memory** in a worst case scenario and **10K threads** and default settings

By bounding maximum untracked memory to `ncpu*max_per_cpu_untracked_memory` we can avoid this massive memory overcommit, and also with default `max_server_memory_usage_to_ram_ratio=0.9`, all this memory is already reserved, so for now per-cpu memory is not reserved explicitly.

There is also very small (32Ki by default) thread local limit, so possible overcommit is `max(nthreads*per_cpu_untracked_memory_thread_buffer, ncpu*max_per_cpu_untracked_memory)`, which should be covered by default `max_server_memory_usage_to_ram_ratio` already.

_Note, librseq gives ±1% ([in the best case](https://pastila.nl/?012f91fd/e6d4eb2ee438c17950e9836bd505512d#mC4O/p41MeNXF2gcCVmbQA==GCM)), and complexity does not worth it_

### Version info
- Merged into: `26.7.1.157` (included in `26.7` and later)


src/Common/CurrentMemoryTracker.cpp
```diff
@@ -78,17 +80,25 @@ AllocationTrace CurrentMemoryTracker::allocImpl(Int64 size, bool enforce_memory_
 
         Int64 previous_untracked_memory = current_thread->untracked_memory;
         current_thread->untracked_memory += size;
-        if (current_thread->untracked_memory > current_thread->untracked_memory_limit)
+
+        bool flush_per_cpu = DB::PerCPUMemory::alloc(size, current_thread->per_cpu_untracked_memory);
```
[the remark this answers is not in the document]
azat: Existing performance tests is enough, new benchmark will not match the production setup since it allocates in smaller batches then the per-thread limit of 32KiB

src/Common/MemoryTrackerSwitcher.cpp
```diff
@@ -35,6 +39,8 @@ MemoryTrackerSwitcher::~MemoryTrackerSwitcher()
     thread_tracker->setParent(prev_memory_tracker_parent);
     current_thread->untracked_memory = prev_untracked_memory;
     current_thread->untracked_memory_blocker_level = prev_untracked_memory_blocker_level;
+    current_thread->per_cpu_untracked_memory.pending_alloc = prev_per_cpu_pending_alloc;
+    current_thread->per_cpu_untracked_memory.pending_free = prev_per_cpu_pending_free;
```
Ergus: Considering the Bot comment I see this and I would say that seems sensible to save `alloc_cpu`, `free_cpu`, `nallocs`, `nfrees` for a complete CPU/Thread state, else, at least document that these are intentionally left unrestored and explain why the precision loss is acceptable

tests/integration/test_early_memory_limit_exception/test.py
```diff
@@ -19,38 +41,41 @@ def started_cluster():
...
+    retry(lambda: "(total) memory limit exceeded" not in instance.exec_in_container(["bash", "-c", f"curl {endpoint}"]))
+    retry(lambda: instance.query("select 1", user="default"))
 
-    assert "(total) memory limit exceeded" in instance.query_and_get_error("select 1", user="Bob")
-    instance.query("system free memory")
-    instance.query("DROP USER IF EXISTS Alex")
-    instance.query("DROP USER IF EXISTS Bob")
+    retry(lambda: "(total) memory limit exceeded" in instance.query_and_get_error("select 1", user="Bob"))
+    retry(lambda: instance.query("system free memory"))
+    retry(lambda: instance.query("DROP USER IF EXISTS Alex"))
```
Ergus: Why we need the retry here? 
IIRC retry swallows all exceptions for all attempts and then raises a generic AssertionError. So if a regression causes `DROP USER` to OOM reliably, the test would spin 100 times and look like a flaky test rather than a clear regression signal

tests/integration/test_early_memory_limit_exception/test.py
```diff
@@ -19,38 +41,41 @@ def started_cluster():
...
+    retry(lambda: instance.query("DROP USER IF EXISTS Bob"))
 
     instance.replace_in_config(
-        "/etc/clickhouse-server/config.d/users_to_ignore_early_memory_limit_check.xml",
+        "/etc/clickhouse-server/config.d/server.yaml",
         "default , Alex ",
         "default",
     )
-    instance.query("system reload config", user="default")
+    retry(lambda: instance.query("system reload config", user="default"))
```
Ergus, line 82: In the tests the retry should only cover the three or four calls that race against profile application, not the entire test including cleanup. I am missing something?

src/Common/CurrentMemoryTracker.cpp
```diff
@@ -78,17 +80,25 @@ AllocationTrace CurrentMemoryTracker::allocImpl(Int64 size, bool enforce_memory_
...
+
+        if (current_thread->untracked_memory > current_thread->untracked_memory_limit || flush_per_cpu)
         {
             Int64 current_untracked_memory = current_thread->untracked_memory;
             current_thread->untracked_memory = 0;
 
             try
             {
-                return memory_tracker->allocImpl(current_untracked_memory, enforce_memory_limit);
+                /// We cannot AllocationTrace that got here, since it's smaple_probability was calculated on the size that may not match original allocation size
```
Ergus: // We cannot ??? AllocationTrace that got here...

Also: `sed 's/smaple/sample/'`

Ergus (commented):
This is labelled "Performance Improvement" but the primary goal is OOM prevention... a correctness/safety fix.

So I think this is an Improvement,

src/Common/CurrentMemoryTracker.cpp
azat: "return", fixed

src/Common/MemoryTrackerSwitcher.cpp
azat: Fixed

tests/integration/test_early_memory_limit_exception/test.py
azat: The only queries that need retries are those between "SYSTEM ALLOCATE" and "SYSTEM FREE", you are right that this one does not need retries. Removed

azat:
>This is labelled "Performance Improvement" but the primary goal is OOM prevention... a correctness/safety fix.

It is marked as so to run performance tests, and in some sense "OOM prevention" can be "performance improvement"

azat:
@Ergus I do not like current implementation, will keep it as draft for now, and ping you/undraft once it will be ready

azat:
@Ergus now it should be better, it uses reservation on per-cpu counters instead

src/Common/examples/memory_tracker_benchmark.cpp
```diff
@@ -0,0 +1,279 @@
...
+        std::string_view a{argv[i]};
+        auto next = [&]() -> const char *
+        {
+            if (i + 1 >= argc)
+                return nullptr;
+            return argv[++i];
+        };
+        if (a == "--threads")
+        {
+            if (const auto * v = next()) opts.threads = static_cast<int>(std::strtol(v, nullptr, 10));
```
Ergus, line 77: This if-else format is permitted?

src/Common/PerCPUMemory.h
```diff
@@ -0,0 +1,227 @@
...
+public:
+    static constexpr Int64 DEFAULT_BUDGET = 8 * 1024 * 1024;
+    static constexpr Int64 DEFAULT_THREAD_BUFFER = 32 * 1024;
+    /// max_per_cpu_untracked_memory == 0 maps here: no side ever leaves the range, disabling the
+    /// per-CPU bound (only the per-thread cap applies).
+    static constexpr Int64 UNLIMITED_BUDGET = std::numeric_limits<Int64>::max() / 2;
+
+    static int numberOfCPUs()
+    {
+        Int64 n = ::sysconf(_SC_NPROCESSORS_CONF);
```
Ergus, line 60: I have a question here. Isn't it a bit more accurate to use the cgroup limit instead of the global system one?

Ergus (approved):
IIUC with this change the new default value of 8Mb forces earlier flushes and many-thread workloads that previously rode the loose `threads * max_untracked_memory` overcommit can now hit `MEMORY_LIMIT_EXCEEDED` earlier than before an upgrade.

This is a behavior change for the user; the only escape is explicitly setting `max_per_cpu_untracked_memory=0`.

1. Could this impact performance in a significant way?
2. Do we need to add the backward incompatible label?

src/Common/PerCPUMemory.h
azat: cgroups limits ignored by intention since you can change them on fly (in theory there is even CPU hotplug, but I haven't seen it in use at all in my experience)

src/Common/examples/memory_tracker_benchmark.cpp
azat: I think it is OK for a very small cases like this one, i.e. some example code, some options parsing, short strings

azat:
> IIUC with this change the new default value of 8Mb forces earlier flushes and many-thread workloads that previously rode the loose threads * max_untracked_memory overcommit can now hit MEMORY_LIMIT_EXCEEDED earlier than before an upgrade.

Yes, and this is the intention, per-thread max_untracked_memory allows too huge overcommit
But note that usually threads executed for some period of time on one CPU, so it should be OK performance wise

> This is a behavior change for the user; the only escape is explicitly setting max_per_cpu_untracked_memory=0.

Disabling by tunning server setting is OK, I doubt that anyone should tune it, only if it will have problems, and if it is, we we will either increase/disable or revert it completely.

> Could this impact performance in a significant way?

Our performance comparison has queries that are sensitive to memory allocations, so in other words it should show if there is significant slowdown (and actually one of previous implementation has bugs and it shows this), while now it looks more or less OK (need to take a look at the results one more time)

> Do we need to add the backward incompatible label?

I don't think so

src/Common/CurrentMemoryTracker.cpp
```diff
@@ -77,19 +79,30 @@ AllocationTrace CurrentMemoryTracker::allocImpl(Int64 size, bool enforce_memory_
...
+
             Int64 current_untracked_memory = current_thread->untracked_memory;
             current_thread->untracked_memory = 0;
 
             try
             {
-                return memory_tracker->allocImpl(current_untracked_memory, enforce_memory_limit);
+                /// We cannot return the AllocationTrace from here, since its sample_probability was calculated on the (batched) flushed size, which may not match the original allocation size.
+                if (current_untracked_memory > 0)
+                    std::ignore = memory_tracker->allocImpl(current_untracked_memory, enforce_memory_limit, /*query_tracker=*/ nullptr, /*_sample_probability=*/ 0.0);
```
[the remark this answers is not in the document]
azat: This is OK for now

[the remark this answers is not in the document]
azat: Fixed

src/Common/CurrentMemoryTracker.cpp
```diff
@@ -77,19 +79,30 @@ AllocationTrace CurrentMemoryTracker::allocImpl(Int64 size, bool enforce_memory_
...
+                else
+                    std::ignore = memory_tracker->free(-current_untracked_memory, /*_sample_probability=*/ 0.0);
             }
             catch (...)
             {
                 current_thread->untracked_memory += previous_untracked_memory;
```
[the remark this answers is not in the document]
azat: This is existing bug, please do not report such, only new that this PR introduces. I will address this later,

src/Common/ThreadStatus.cpp
```diff
@@ -116,6 +117,9 @@ ThreadStatus::ThreadStatus()
     last_rusage = std::make_unique<RUsageCounters>();
 
     memory_tracker.setDescription("Thread");
+    /// memory_tracker is already parented to total_memory_tracker, so a thread that never attaches
+    /// to a group still honors total_memory_tracker_sample_probability.
+    resolveMemorySampleConfig();
```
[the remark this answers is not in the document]
azat: Minor, but fixed

PR #106069, opened 2026-05-28
Expand Nested prefix in ARRAY JOIN over ALIAS column

KochetovNicolai:
In the new analyzer, `ARRAY JOIN <name>` over a Nested prefix used to fail with `TYPE_MISMATCH` whenever a same-named `ALIAS` column resolved to a non-`Array`/`Map` expression (for example a `Tuple`). The identifier resolver looks up the column map first, finds the alias, and never reaches the nested-prefix expansion path; meanwhile the legacy analyzer treats the name as a marker for the underlying per-field arrays.

Fix the regression by retrying the resolution as a Nested prefix in `QueryAnalyzer::resolveArrayJoin` whenever a bare identifier resolves to a non-`Array`/`Map` type. The nested-expansion logic that already lives in `IdentifierResolver::tryResolveIdentifierFromStorage` is hoisted into a new static helper `IdentifierResolver::tryResolveIdentifierAsNestedPrefix` so both call sites share the same code path.

Better compatibility with the old analyzer. If the table has columns like `x.a Array, x.b Array, x String`, prefer arrays for `ARRAY JOIN x`.

### Version info
- Merged into: `26.6.1.410`


src/Analyzer/Resolve/QueryAnalyzer.cpp
```diff
@@ -4603,6 +4639,53 @@ void QueryAnalyzer::resolveArrayJoin(QueryTreeNodePtr & array_join_node, Identif
...
+        /// immediate child), so the retry also fires when the ARRAY JOIN wraps a join
+        /// tree, e.g. `FROM s JOIN t ON ... ARRAY JOIN s.loc`.
+        if (auto * resolved_column_node = array_join_expression->as<ColumnNode>())
+        {
+            auto current_result_type = resolved_column_node->getResultType();
+            if (current_result_type && !isArray(current_result_type) && !isMap(current_result_type))
+            {
+                auto column_source = resolved_column_node->getColumnSourceOrNull();
+                const auto & array_join_input = array_join_node_typed.getTableExpression();
+                if (column_source && isInArrayJoinSourceTree(column_source.get(), array_join_input.get()))
```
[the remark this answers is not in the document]
KochetovNicolai: I think the example `WITH loc AS l SELECT l FROM t ARRAY JOIN l` is a valid case. If this query works, it's as if nested column `loc` was aliased as `l`.

novikd: It seems that `isInArrayJoinSourceTree` is not needed. We should just check that the column source is TableNode.

src/Analyzer/Resolve/QueryAnalyzer.cpp
```diff
@@ -373,6 +373,42 @@ static bool isFromJoinTree(const IQueryTreeNode * node_source, const IQueryTreeN
     return false;
 }
 
+/// Variant of `isFromJoinTree` that also descends into `ArrayJoinNode` children. Used to
+/// check whether a column's source table participates in an ARRAY JOIN's input subtree,
+/// which may itself be (or contain) another ARRAY JOIN when ARRAY JOINs are chained.
+static bool isInArrayJoinSourceTree(const IQueryTreeNode * node_source, const IQueryTreeNode * tree_node)
```
novikd: Can be removed

novikd (approved):
LGTM

clickgapai:

### TL;DR
**Expand Nested prefix in ARRAY JOIN over ALIAS column**

**ClickGap verified:** reproduced it with the script below on `master`; tested 5 version(s) — 5 affected, 0 clean (see Affects).

### Reproduction
```
CREATE TABLE t_array_join_alias_nested (id String, `loc.x` Array(String), `loc.y` Array(String), loc Tuple(x String, y String) ALIAS tuple(`loc.x`[1], `loc.y`[1])) ENGINE = MergeTree ORDER BY id; INSERT INTO t_array_join_alias_nested (id, `loc.x`, `loc.y`) VALUES ('a', ['x1','x2'], ['y1','y2']); SELECT loc.x FROM t_array_join_alias_nested ARRAY JOIN loc SETTINGS enable_analyzer = 1;
```
[▶ Run on ClickHouse Fiddle](https://fiddle.clickhouse.com/a7b14a98-9fd9-478c-bc04-86da43cbd845)

### Bisect
- Bisect not performed — no regression boundary was established from the affects matrix above (the bug did not reproduce on master, or no clean baseline release was found to bound the search).

### Affects
- **Reproduces on**: `master, 26.5, 26.4, 26.3, 25.8`

### Suggested next step
Manual triage of the `comp-query-analyzer` path is needed. The reproducer above should be a fast first check.

**Severity:** `P2`

cc fix PR author: @KochetovNicolai
cc fix PR reviewer: @novikd

---
_ClickGap · Finding: `phase_d_pr106069`_

PR #106070, opened 2026-05-28
Improve canceling queries in LimitByTransform and LimitBySortedStreamTransform

rvasin:
- Improvement

Reduced cancellation latency for the positive forms of LIMIT BY (LimitByTransform, LimitBySortedStreamTransform): KILL QUERY and Ctrl+C now interrupt processing within a single chunk instead of waiting for the entire step to complete.

### Version info
- Merged into: `26.8.1.63` (included in `26.8` and later)


rvasin:
@vdimir could you please assign `can be tested` label?

tests/queries/0_stateless/04266_kill_limit_by_transform.sh
```diff
@@ -0,0 +1,53 @@
...
+        number % 100000000 AS category,
+        number AS value
+    FROM numbers(100000000)
+    LIMIT 1 BY category
+    FORMAT Null
+    SETTINGS max_block_size=100000000, max_threads=1, max_rows_to_read=0
+" || exit 1
+
+# Test transformInOrder path
+test_limit_by_cancellation "inorder" "
```
tiandiwonder: The "inorder" case does not actually exercise transformInOrder — both test queries run transformCommon.

  LimitByStep only selects the in-order path when an upstream sort property has the LIMIT BY columns as a prefix
  (LimitByStep::applyOrder: in_order = sort_description.hasPrefix(columns)). The "inorder" query has no ORDER BY and its keys are
  derived string expressions (cat1/cat2), which are not a prefix of any input ordering, so in_order stays false. EXPLAIN PIPELINE
  confirms it produces LimitByTransform, not LimitByTransform (InOrder) — same as the "common" query. So the cancellation checks
  added to transformInOrder are never executed by any test.

  To genuinely cover transformInOrder, use a MergeTree table whose sort key is the LIMIT BY key and read it in order — this yields
  LimitByTransform (InOrder) with only a non-blocking RemoveVirtualRowTransform above it (verified via EXPLAIN PIPELINE):

  CREATE TABLE t (k UInt64) ENGINE = MergeTree ORDER BY k
    AS SELECT number FROM numbers(...);

  SELECT k FROM t ORDER BY k LIMIT 1 BY k FORMAT Null
    SETTINGS optimize_read_in_order = 1, max_block_size = <large>;

  Avoid SELECT number FROM numbers(N) ORDER BY number LIMIT 1 BY number: it does reach LimitByTransform (InOrder), but it inserts a
  blocking MergeSortingTransform ahead of it. A KILL would then be cancelled inside the sort (which consumes all input before
  emitting), so it would test sort cancellation rather than transformInOrder. The read-in-order plan above has no blocking sort, so
  the kill lands in the LimitBy stage.

tiandiwonder:
please fix the two comments by clickhouse-gh first.

tiandiwonder:
It's a real problem having array join with limit, but the current fix only applies to limit. what I am thinking is a more complete fix for array join, including limit, filter, etc.

rvasin:
> please fix the two comments by clickhouse-gh first.

@tiandiwonder please advice about the tests.
For both paths transformCommon and transformInOrder I have carefully selected the queries in such way what LimitByTransform is executed long time and other processors like ExpressionTranform are short. So when KILL QUERY works it always make existing in LimitByTransform transformCommon/transformInOrder, But I agree with clickhouse-gh that it;s not evident. 
I have been trying to improve functional tests framework by adding function similar to  `wait_for_log_line`  which we already have in integration tests. But I think it's overkill for function tests. 
In fact I have improved cancellation in `PostgreSQLSource` and other sources the approach is the following src/Processors/Sources/PostgreSQLSource.cpp
```cpp
template<typename T>
Chunk PostgreSQLSource<T>::generate()
{
    LOG_TEST(getLogger("PostgreSQLSource"), "Generate a chunk from stream");

    /// Check if source was cancelled or completed
    if (is_completed.load() || isCancelled())
        return {};
```

Then in my integration test tests/integration/test_postgresql_kill_query/test.py

```python
    node1.wait_for_log_line("Generate a chunk from stream")
    time.sleep(1)

    node1.query(f"KILL QUERY WHERE query_id='{query_id}' SYNC")
```

I am planning to remove the functional test for LimitByTransform and add new **integration** test. 
In src/Processors/Transforms/LimitByTransform.cpp
```cpp
void LimitByTransform::transformInOrder(Chunk & chunk)
{
    LOG_TEST(getLogger("LimitByTransform"), "exec transformInOrder");
    
```
Then in the integration test:
```python
    node1.wait_for_log_line("exec transformInOrder")
    time.sleep(1)

    node1.query(f"KILL QUERY WHERE query_id='{query_id}' SYNC")
```

I think this the only way to make the tests 100% correct: when LimitByTransform start we send `KILL QUERY`.  
Are you agree with this approach? I may quickly implement it.

If needed it's possible also track the exit message like:
```cpp
    for (UInt64 row = 0; row < num_rows; ++row)
    {
        if (isCancelled())
        {
             LOG_TEST(getLogger("LimitByTransform"), "transformInOrder exit loop");
            stopReading();
            return;
        }
```
then in test:
```python
node1.wait_for_log_line("transformInOrder exit loop")
```
But may be it's overkill too. Let me to know your opinion.

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -74,11 +83,23 @@ void LimitByTransform::transformInOrder(Chunk & chunk)
     UInt64 num_rows = chunk.getNumRows();
     auto columns = chunk.detachColumns();
 
+    if (isCancelled())
```
Ergus: This check should probably be moved before the `detachColumns` because detachColumns() moves the entire column data out of chunk and into the local columns vector.

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -45,11 +42,23 @@ void LimitByTransform::transformCommon(Chunk & chunk)
...
```
Ergus: Look at the last comment

tests/queries/0_stateless/04266_kill_limit_by_transform.sh
```diff
@@ -0,0 +1,53 @@
...
+        concat(toString(number), repeat('x', 50)) AS cat1,
+        concat(toString(number), repeat('y', 50)) AS cat2,
+        number AS value
+    FROM numbers(100000000)
+    LIMIT 1 BY
+        cat1,
+        cat2
+    FORMAT Null
+    SETTINGS max_block_size = 100000000, max_threads = 1, max_rows_to_read=0
+" || exit 1
```
Ergus: IIUC in order requires an `ORDER BY`

You need an explicit ORDER BY whose columns are a prefix of the LIMIT BY columns like (untested):

```
SELECT number AS key
  FROM numbers(100000000)
  ORDER BY key          -- SortingStep → SortScope::Global, desc = {key}
  LIMIT 1 BY key        -- applyOrder called: hasPrefix({"key"}) = true-> in_order = true
  FORMAT Null
  SETTINGS max_block_size = 100000000, max_threads = 1
```

 Even with ORDER BY key, there's a new problem: the SortingStep itself is a heavy operation on 100M rows and it also respects isCancelled(). The KILL QUERY issued immediately after wait_for_query_to_start is just as likely probably more likely to cancel inside the merge-sort merge phase than inside transformInOrder.

Ergus (dismissed):
The changelog describes the mechanism rather than the user-visible impact. 

Maybe better something like: 

```
Reduced cancellation latency for queries using LIMIT BY: KILL QUERY and Ctrl+C now interrupt processing within a single chunk instead of waiting for the entire LIMIT BY step to complete.
```

Both test cases end up calling `transformCommon`. The `isCancelled()` check added to `transformInOrder` is dead code from the test's perspective

tests/queries/0_stateless/04266_kill_limit_by_transform.sh
rvasin: I have replaced function test with integration test where I use  ORDER BY  now. and I issue KILL QUERY in the integration test right after entering into `LimitByTransform::transformInOrder`

src/Processors/Transforms/LimitByTransform.cpp
rvasin: @Ergus I think placing ` if (isCancelled())` after `detachColumns()` is correct. Because in case of cancellation we return empty chunk. This leads to `has_output = false` in `ISimpleTransform::work()`. And the taken chunk  is freed when the local `auto columns` goes out of scope.

tests/queries/0_stateless/04266_kill_limit_by_transform.sh
rvasin: In the integration test I have replaced it with:
```sql
SELECT
    number AS key1,
    number + 1 AS key2,
    number AS value
FROM numbers(100000000)
ORDER BY key1 ASC, key2 ASC
LIMIT 1 BY key1, key2
FORMAT Null
SETTINGS max_block_size=100000000, max_threads=1, max_rows_to_read=0
```
the LimitByTransform::transformInOrder runs for 4 seconds for this query on my 48 machine. So it's hard to increase LimitByTransform::transformInOrder run time for example to 30 seconds without increasing RAM consumption. In integration tests there is the RAM limit for 10 GB. I will see how it will run on GitHub CI (on my local runs it is running fine)..

rvasin:
> The changelog describes the mechanism rather than the user-visible impact.
> 
> Maybe better something like:
> 
> ```
> Reduced cancellation latency for queries using LIMIT BY: KILL QUERY and Ctrl+C now interrupt processing within a single chunk instead of waiting for the entire LIMIT BY step to complete.
> ```
> 
Thank you for the suggestion. I have updated the changelog.

> Both test cases end up calling `transformCommon`. The `isCancelled()` check added to `transformInOrder` is dead code from the test's perspective

Yes. Thank you for noticing it. I have updated the query as explained in other threads.

rvasin:
@tiandiwonder do you have any questions/suggestions to this PR? It seems like all tests and checks were successfully passed.

tiandiwonder:
@rvasin please merge the current pr with latest master, thanks.

rvasin:
@tiandiwonder I have resolved the merge conflicts. Could you restart the pipeline? I see Fast test and others were cancelled.

src/Processors/Transforms/LimitByTransform.cpp
```diff
@@ -228,6 +229,12 @@ void LimitByTransform::consumeImpl(Method & hash_method, const ColumnRawPtrs & g
     size_t current_run_group_idx = 0;
     for (UInt64 row_idx = 0; row_idx < row_count; ++row_idx)
     {
+        if (isCancelled())
```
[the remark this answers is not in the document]
rvasin: I have changed the changelog to reflect what this pull request is only for the positive forms of LIMIT BY.

tests/integration/test_limit_by_transform_kill_query/test.py
```diff
@@ -0,0 +1,88 @@
...
+        except Exception as e:
+            thread_error[0] = e
+
+    query_thread = threading.Thread(target=execute_query)
+    query_thread.start()
+
+    node1.wait_for_log_line(log_line_pattern)
+    time.sleep(1)
+
+    node1.query(f"KILL QUERY WHERE query_id='{query_id}' SYNC")
```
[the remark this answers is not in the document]
rvasin:
```
Please assert the cancellation branch itself, e.g. emit a LOG_TEST immediately before stopReading in each row-loop cancellation path and wait for that line
```
this is probably the good advice but it could probably make the test flaky because probably we may miss the LOG_TEST text before stopReading.

tests/integration/test_limit_by_transform_kill_query/test.py
```diff
@@ -0,0 +1,88 @@
...
+            )
+            assert "DB::Exception: Query was cancelled" in error
+        except Exception as e:
+            thread_error[0] = e
+
+    query_thread = threading.Thread(target=execute_query)
+    query_thread.start()
+
+    node1.wait_for_log_line(log_line_pattern)
+    time.sleep(1)
```
tiandiwonder: `test_sorted_kill_query` fails in CI here: the sorted query finishes during this 1s window before `KILL` lands, so the assert sees an empty error (`'Query was cancelled' in ''`). The blocking sort completes before the `LimitBySortedStreamTransform` log line fires, then the transform drains its rows faster than `sleep` + `KILL` — the hashmap variant stays alive long enough but the sorted one doesn't. It needs a deterministic hold (e.g. a failpoint inside the transform) so the sorted path is still running when `KILL` is issued.

src/Processors/Transforms/LimitByTransform.cpp
tiandiwonder, line 247: optional: on this cancel-return, `transform` still materializes the partially-filled `output_slices` below before stopping.

the sorted loop returns from `transform` directly so it skips that; a post-`consumeImpl` `isCancelled()` check before `materializeSlicesIntoChunk` would make the two paths symmetric and skip the wasted work.

tests/integration/test_limit_by_transform_kill_query/test.py
rvasin: Yes, thank you for noticing it. I am working on it. Right now test_sorted_kill_query is flaky (thanks to the recent CI run, because previous CI runs were fine).

src/Processors/Transforms/LimitByTransform.cpp
rvasin: Again, thank you for noticing it. I will add new `isCancelled()` check in `void LimitByTransform::transform(Chunk & chunk)`:

```cpp
    if (isCancelled())
        return;

    /// No row from this chunk survived `LIMIT BY`.
    if (output_slices.empty())
        return;

    const UInt64 output_row_count = materializeSlicesIntoChunk(…);
```
It will make both path symmetric.

tests/integration/test_limit_by_transform_kill_query/test.py
rvasin: Done. Re-worked it via failpoint mechanism.

rvasin:
@tiandiwonder could you please review PR again? i have been struggling to find the correct SQL/row count for `test_sorted_kill_query`, because the run time/RAM consumption is totally different for amd_tsan and arm_release builds. Then I have found the correct combination and the test was successfully passed on CI. But I was warring that with upgrading ARM CI runners the test may become flaky. `clickhouse-gh` have noticed it too and suggested to use FailPoint mechanism, So I refactored the code to use FailPoint approach and it seems now the test is very stable and reliable. Please review and make your further suggestions if needed.

tiandiwonder:
@rvasin looks good now. Would you plese merge with latest master and resolve the conflicts?

rvasin:
@tiandiwonder I have resolved merge conflicts. One failed test is not related.

rvasin:
@tiandiwonder do you have more suggestions for this PR?

Ergus:
@tiandiwonder I requested some changes a while ago, but the PR is assigned to you, so if you approve it it can go on without a new review/aprovel from me

rvasin:
@Ergus could you please confirm that all changes you requested where implemented? If not let me to know what is not yet implemented. I will try to implement it.

tiandiwonder (approved):

tests/integration/test_limit_by_transform_kill_query/test.py
```diff
@@ -0,0 +1,95 @@
...
+                query_id=query_id,
+            )
+            assert "DB::Exception: Query was cancelled" in error
+        except Exception as e:
+            thread_error[0] = e
+
+    query_thread = threading.Thread(target=execute_query)
+    query_thread.start()
+
+    node1.query(f"SYSTEM WAIT FAILPOINT {fault_name} PAUSE")
```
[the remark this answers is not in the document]
tiandiwonder: @rvasin would you please fix it?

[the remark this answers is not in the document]
rvasin: @tiandiwonder yes, it will fix it today.

rvasin:
@tiandiwonder please take a look at the pool request again.

PedroTadim:
@rvasin are there any other transforms that could also improve query cancellation?

rvasin:
> @rvasin are there any other transforms that could also improve query cancellation?

@PedroTadim In my company I am assigned to make analysis (and improvements) of ClickHouse codebase for query cancellation because it's critical for our clients. I have done several pull requests for most of Sources which were critical (we had to restart ClickHouse to cancel query), some pull requests for transforms (not too critical used in most queries like ExpressionTranform) and output formats. 

Here is my analysis related to Transforms only:

1. High Impact (long-running, common queries, ClickHouse is blocked for minutes)

- AggregatingTransform
- SortingTransform / MergeSortingTransform
- MergeJoinTransform
- JoiningTransform / (due to HashJoin/GraceHashJoin)

2. Medium Impact (ClickHouse is blocked for seconds)

- DistinctTransform - my [pull request](https://github.com/ClickHouse/ClickHouse/pull/108588) in review
- DistinctSortedStreamTransform - my [pull request](https://github.com/ClickHouse/ClickHouse/pull/111061) in review
- FilterTransform - my [pull request](https://github.com/ClickHouse/ClickHouse/pull/111061) status: closed because it was supersed by #112705 which is in progress.
- AggregatingInOrderTransform
- CubeTransform / RollupTransform
- MergingAggregatedTransform
- IntersectOrExceptTransform
- CheckSortedTransform (I am working on it now)
- CreatingSetsTransform
- ArrayJoinTransform

3. Lower Impact (shorter loops or less common)

- TotalsHavingTransform
- TTLTransform / TTLCalcTransform
- TTLDeleteFilterTransform
- TTLColumnAlgorithm / TTLDeleteAlgorithm / TTLUpdateInfoAlgorithm
- PartialSortingTransform
- CheckConstraintsTransform
- LazyMaterializingTransform
- FilterSortedStreamByRange
- ReverseTransform
- TotalsHavingTransform

PR #106078, opened 2026-05-28
Add silk fibers aware socket implementations

mstetsyuk:
- Improvement

Add silk fiber aware secure and plain socket implementations. 

### Version info
- Merged into: `26.6.1.814`


nikitamikhaylov:
@mstetsyuk what's the difference with https://github.com/ClickHouse/silk/blob/main/src/perf/fiber-http.h ?

mstetsyuk:
2 changes on the silk side are required:
1. Add `silk::FiberScheduler::connect` (https://github.com/ClickHouse/silk/pull/52), otherwise [this](https://github.com/ClickHouse/ClickHouse/pull/106078/changes/3634b361dba55baab5395df4750a99362369da75) is required to make sure that the socket is unblocking - required by [this logic](https://github.com/ClickHouse/ClickHouse/blob/master/base/poco/NetSSL_OpenSSL/src/SecureSocketImpl.cpp#L522).
2. `cancelIo` should not do `future->result = nullptr` on cancel, as there is a race between our thread calling cancel and the kernel consuming bytes from TCP connection into our userspace buffer. Failure mode: kernel reads bytes from a TCP connection, writes them to our userspace buffer, and we set `future->result = nullptr`. The bytes are forever lost, so when we try to read the data again (with a higher timeout, for instance), we won't see them.

mstetsyuk:
> what's the difference with https://github.com/ClickHouse/silk/blob/main/src/perf/fiber-http.h ?

This PR is basically https://github.com/ClickHouse/silk/blob/main/src/perf/fiber-http.h + TLS + cosmetic changes/simplifications/minor improvements (honouring the receive/send timeout) + tests

(I was under the impression that https://github.com/ClickHouse/silk/blob/main/src/perf/fiber-http.h is just for internal perf tests in the silk library - while this PR is meant for normal production usage in our codebase)

mstetsyuk:
This change is required to implement this feature: https://github.com/ClickHouse/ClickHouse/pull/106078

src/IO/SilkFiberStreamSocketImpl.cpp
```diff
@@ -0,0 +1,208 @@
+#include <IO/SilkFiberStreamSocketImpl.h>
```
vadimskipin, line 1: I would suggest to drop the silk part of a name - just FiberStreamSocket.
I hope there will be just one fiber library in use.

src/IO/SilkFiberStreamSocketImpl.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+    const Poco::Timespan elapsed = Poco::Timestamp() - started;
+    timeout = (elapsed < timeout) ? (timeout - elapsed) : Poco::Timespan(0);
+
+    if (r)
+        error(r, "poll");
+    return (triggered & static_cast<uint64_t>(events)) != 0;
+}
+
+int FiberStreamSocketImpl::sendBytes(const void * buffer, int length, int flags)
```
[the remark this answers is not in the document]
vadimskipin: I believe this one is important. But note that the throttler function just sleeps now - need to changes this. Must return a sleep duration so fiber code can do FinerScheduler::sleep

mstetsyuk:
> I believe this one is important. But note that the throttler function just sleeps now - need to changes this. Must return a sleep duration so fiber code can do FinerScheduler::sleep

Apparently, we can make it even simpler. It's the responsibility of the caller to supply the throttler, so I think we can just add `Silk::Throttler` which overrides sleep: https://github.com/ClickHouse/ClickHouse/pull/106078/commits/079062091a2e366b2f17e63203dbb3c9db8c9464

And if the user of this class wants to set a throttler, it's on them to use the right one.

src/IO/SilkFiberStreamSocketImpl.cpp
mstetsyuk: There is another one, apparently (`boost::context`).

By the way, what do you think about the `Silk` namespace? I think it should be useful to gate silk-aware sockets, connection pool, throttlers etc.

mstetsyuk:
> I believe this one is important. But note that the throttler function just sleeps now - need to changes this. Must return a sleep duration so fiber code can do FinerScheduler::sleep

@vadimskipin, addressed by adding a one-liner throttler implementation that does `silk::FiberScheduler::sleep` and not `std::this_thread::sleep`.

nikitamikhaylov:
Unit tests failure should be fixed by: https://github.com/ClickHouse/ClickHouse/pull/106856

vadimskipin (approved):

clickgapai:
Hi @mstetsyuk @nikitamikhaylov @vadimskipin — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: PR body is essentially template-only; it does not explain the Throttler `sleep` virtualization, the Poco `BIO_METHOD` injection, or the `supportsExternalPolling` contract enforced in `setAsyncCallback`.
- Suggested addition:

  > Describe: (1) the new build flag USE_SILK and its glibc/x86-64-v2/liburing requirements; (2) that `Throttler::sleep` is now virtual so `Silk::Throttler` can yield the fiber instead of blocking the thread; (3) that `setAsyncCallback` now rejects callbacks on sockets without external-polling support (Silk fiber sockets), since they use the blocking-future model; (4) the custom `BIO_METHOD` injection used to route OpenSSL I/O through the fiber scheduler.

alexey-milovidov:
Merged with broken CI :crying_cat_face:

PR #106092, opened 2026-05-29
Support AT TIME ZONE and AT LOCAL postfix operators

### LINKED ISSUE #105950: Support `AT TIME ZONE` postfix operator on datetime values
ClickHouse rejects the standard `<datetime> AT TIME ZONE <zone>` postfix syntax that PostgreSQL and most other engines accept:

```sql
SELECT TIMESTAMP '2024-01-01 12:00:00' AT TIME ZONE 'UTC';
-- Code: 62. DB::Exception: Syntax error
```

The functionality is already present via `toTimeZone(dt, 'UTC')`, so this is a parser-level ask: accept the postfix `AT TIME ZONE <expr>` and `AT LOCAL` and desugar to the existing function.

Postgres reference: https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-ZONECONVERT

```sql
-- Postgres / desired ClickHouse behaviour
SELECT TIMESTAMP '2001-02-16 20:38:40' AT TIME ZONE 'America/Denver';
SELECT timestamptz '2001-02-16 20:38:40-05' AT TIME ZONE 'America/Denver';
```

Workaround today: `toTimeZone(ts, 'America/Denver')`.

### Version info
- Resolved by: #106092
- Merged into: `26.7.1.208` (included in `26.7` and later)

lvzhipin03:
The standard SQL `AT TIME ZONE` operator (and its `AT LOCAL` shorthand) is widely used in PostgreSQL and other databases to convert datetime values between time zones. ClickHouse already provides the equivalent `toTimeZone` function, but previously rejected the postfix syntax at parse time with a syntax error.

This change adds parser-level support for both forms as pure syntactic sugar — no new AST nodes, no runtime overhead:

- `expr AT TIME ZONE zone` → `toTimeZone(expr, zone)`
- `expr AT LOCAL` → `toTimeZone(expr, timeZone())`

AT TIME ZONE is added to the binary-operator table at priority 13 (above *///% at 12, and above +/- at 11), matching PostgreSQL's precedence: `a * ts AT TIME ZONE 'tz'` binds as `a * (ts AT TIME ZONE 'tz')`, and `ts + interval AT TIME ZONE 'tz'` binds as `ts + (interval AT TIME ZONE 'tz')`. To convert after arithmetic, explicit parentheses are required: `(ts + interval) AT TIME ZONE 'tz'`.

`AT LOCAL` is handled as a special postfix case in `tryParseOperator` before the operator-table loop, because it takes no right-hand operand; it desugars to `toTimeZone(expr, timeZone())` where `timeZone()` returns the current session timezone.

Two new keywords are added to `CommonParsers.h`: `AT` and `AT_TIME_ZONE`. The existing `LOCAL` and `TIME_ZONE` keywords are reused.

Closes #105950

- New Feature

### Changelog entry:
Added support for the standard SQL `AT TIME ZONE` and `AT LOCAL` postfix operators as syntactic sugar for `toTimeZone`. The expression `expr AT TIME ZONE zone` is now equivalent to `toTimeZone(expr, zone)`, and `expr AT LOCAL` is equivalent to `toTimeZone(expr, timeZone())`.

### Version info
- Merged into: `26.7.1.208` (included in `26.7` and later)


tiandiwonder:
>AT TIME ZONE is added to the binary-operator table at priority 10 (between || and +/-), matching PostgreSQL's precedence: a + b AT TIME ZONE 'tz' correctly folds to toTimeZone(a + b, 'tz').
It seems not wrong: 
  
INTERVAL 1 HOUR + ts AT TIME ZONE 'America/Denver'

```
  - ClickHouse:(INTERVAL 1 HOUR + ts) AT TIME ZONE 'Denver'
  - PostgreSQL:INTERVAL 1 HOUR + (ts AT TIME ZONE 'Denver') 
```

tests/queries/0_stateless/04299_at_time_zone.sql
tiandiwonder, line 1: One note on the tests: as an operator, AT TIME ZONE is only exercised against +. The other
  operator-level cases aren't covered — I checked them on this PR's build and they all behave correctly, so these are coverage gaps rather than bugs, but they're the parts most likely to regress later. Worth adding:

```
  -- precedence / associativity (formatQuery pins the grouping)
  SELECT formatQuery($$SELECT dt AT TIME ZONE 'UTC' = dt2$$);          -- toTimeZone(dt, 'UTC') = dt2   (binds tighter than
  '=')
  SELECT formatQuery($$SELECT dt AT TIME ZONE 'A' AT TIME ZONE 'B'$$); -- toTimeZone(toTimeZone(dt, 'A'), 'B')  (chained,
  left-assoc)
  SELECT formatQuery($$SELECT t AT TIME ZONE tz FROM x$$);             -- toTimeZone(t, tz)  (column operands)

  -- error paths
  SELECT 'x' AT TIME ZONE 'UTC';   -- { serverError ILLEGAL_TYPE_OF_ARGUMENT }
  SELECT 1 AT FOO;                 -- { clientError SYNTAX_ERROR }

  -- backward-compat of the new AT keyword: `at` / `local` as alias and column name
  SELECT 1 AS at, number AS local FROM numbers(1);
```

lvzhipin03: This case has been added to the test file.

lvzhipin03:
> > AT TIME ZONE is added to the binary-operator table at priority 10 (between || and +/-), matching PostgreSQL's precedence: a + b AT TIME ZONE 'tz' correctly folds to toTimeZone(a + b, 'tz').
> > It seems not wrong:
> 
> INTERVAL 1 HOUR + ts AT TIME ZONE 'America/Denver'
> 
> ```
>   - ClickHouse:(INTERVAL 1 HOUR + ts) AT TIME ZONE 'Denver'
>   - PostgreSQL:INTERVAL 1 HOUR + (ts AT TIME ZONE 'Denver') 
> ```

Thanks for catching this — you're right that the old priority 10 didn't match PostgreSQL. The example you cited (INTERVAL 1 HOUR + ts AT TIME ZONE 'Denver') was exactly the case where they diverged.

I've changed AT TIME ZONE (and AT LOCAL) to priority 12 (same as *///%, above +/-), so now INTERVAL 1 HOUR + ts AT TIME ZONE 'Denver' parses as INTERVAL 1 HOUR + (ts AT TIME ZONE 'Denver'), matching PostgreSQL. The PR description has been updated accordingly.

As a consequence, ts + interval AT TIME ZONE 'tz' (interval on the right) now produces ILLEGAL_TYPE_OF_ARGUMENT — explicit parentheses are required: (ts + interval) AT TIME ZONE 'tz', same as PostgreSQL.

src/Parsers/ExpressionListParsers.cpp
```diff
@@ -3215,6 +3215,7 @@ const std::vector<std::pair<std::string_view, Operator>> ParserExpressionImpl::o
     {toStringView(Keyword::GLOBAL_IN),     Operator("globalIn",        9,  2)},
     {toStringView(Keyword::GLOBAL_NOT_IN), Operator("globalNotIn",     9,  2)},
     {"||",            Operator("concat",          10, 2, OperatorType::Mergeable)},
+    {toStringView(Keyword::AT_TIME_ZONE),        Operator("toTimeZone",      13, 2)},
```
tiandiwonder, line 3218: This registers `toTimeZone` into the query-plan operator pretty-printer, but the canonical function name is `toTimezone` (this spelling is just the alias), so the lookup never matches — a no-op today. It's fragile though: if the spelling is ever 'corrected' to `toTimezone`, every `toTimezone(...)` call would render as `AT TIME ZONE` in `EXPLAIN PLAN pretty=1`. Cleaner to skip it explicitly, like `match` is.

src/Functions/serverConstants.cpp
```diff
@@ -72,12 +77,20 @@ namespace
...
     class FunctionTimezone final : public FunctionServerConstantBase<FunctionTimezone, String, DataTypeString>
     {
     public:
         static constexpr auto name = "timezone";
         static FunctionPtr create(ContextPtr context) { return std::make_shared<FunctionTimezone>(context); }
-        explicit FunctionTimezone(ContextPtr context) : FunctionServerConstantBase(DateLUT::instance().getTimeZone(), context->isDistributed()) {}
+        explicit FunctionTimezone(ContextPtr context)
+            : FunctionServerConstantBase(
+                DateLUT::instance().getTimeZone(),
+                context->isDistributed() && context->getSettingsRef()[Setting::session_timezone].value.empty())
```
tiandiwonder, line 92: Folding here is correct only while `session_timezone` stays query-wide constant across shards — i.e. the setting actually propagates and no shard overrides it (profile/readonly). Fine in the normal path and 04300 covers it; just flagging that folding to the initiator's value rests on that assumption.

tests/queries/0_stateless/04299_at_time_zone.sql
```diff
@@ -0,0 +1,52 @@
...
+-- Must produce the same result as the function form
+SELECT (TIMESTAMP '2001-02-16 20:38:40' AT TIME ZONE 'America/Denver') = toTimeZone(TIMESTAMP '2001-02-16 20:38:40', 'America/Denver');
+
+-- AT LOCAL converts to the session timezone (UTC in this test)
+SELECT TIMESTAMP '2001-02-16 20:38:40' AT LOCAL;
+
+-- AT LOCAL must equal toTimeZone(expr, timeZone())
+SELECT (TIMESTAMP '2001-02-16 20:38:40' AT LOCAL) = toTimeZone(TIMESTAMP '2001-02-16 20:38:40', timeZone());
+
+-- Timezone expression can be any expression, not just a literal
```
tiandiwonder: This comment says "any expression", but the docs were updated to say a constant string by default (a column needs `allow_nonconst_timezone_arguments=1`). Worth syncing so the test doesn't keep the old wording alive.

tests/queries/0_stateless/04299_at_time_zone.sql
```diff
@@ -0,0 +1,52 @@
...
+SELECT TIMESTAMP '2001-02-16 20:38:40' + INTERVAL 1 HOUR AT LOCAL; -- { serverError ILLEGAL_TYPE_OF_ARGUMENT }
+
+-- INTERVAL on the left — AT TIME ZONE binds to ts, matching PostgreSQL
+-- INTERVAL 1 HOUR + ts AT TIME ZONE 'tz'  →  INTERVAL 1 HOUR + (ts AT TIME ZONE 'tz')
+SELECT INTERVAL 1 HOUR + TIMESTAMP '2001-02-16 20:38:40' AT TIME ZONE 'America/Denver';
+
+-- precedence / associativity (formatQuery pins the grouping)
+SELECT formatQuery($$SELECT dt AT TIME ZONE 'UTC' = dt2$$);          -- toTimeZone(dt, 'UTC') = dt2 (binds tighter than '=')
+SELECT formatQuery($$SELECT dt AT TIME ZONE 'A' AT TIME ZONE 'B'$$); -- toTimeZone(toTimeZone(dt, 'A'), 'B') (chained, left-assoc)
+SELECT formatQuery($$SELECT t AT TIME ZONE tz FROM x$$);             -- toTimeZone(t, tz) (column operands)
```
tiandiwonder, line 44: This only pins the parse via `formatQuery` — the non-const timezone column path (`tz` as a real column, which needs `allow_nonconst_timezone_arguments`) is never actually executed.

tests/queries/0_stateless/04299_at_time_zone.sql
```diff
@@ -0,0 +1,52 @@
...
+SELECT formatQuery($$SELECT a * ts AT TIME ZONE 'UTC'$$);            -- a * toTimeZone(ts, 'UTC') (AT binds tighter than *)
+
+-- error paths
+SELECT 'x' AT TIME ZONE 'UTC';   -- { serverError ILLEGAL_TYPE_OF_ARGUMENT }
+SELECT 1 AT FOO;                  -- { clientError SYNTAX_ERROR }
+
+-- backward-compat of the new AT keyword: `at` / `local` as alias and column name
+SELECT 1 AS at, number AS local FROM numbers(1);
```
tiandiwonder: `at`/`local` are exercised as aliases but not as real column names — the keyword regression that matters most is a table that actually has a column named `at` or `local`.

src/Parsers/ExpressionListParsers.cpp
lvzhipin03: Fixed in commit 4de83668cc5b077a945d80710a636dc9235c6070

tests/queries/0_stateless/04299_at_time_zone.sql
lvzhipin03: Fixed in commit 01b26b8e816561090f272897b2d0e73289460bce

tests/queries/0_stateless/04299_at_time_zone.sql
lvzhipin03: Added an execution test in c1149d3f48c8d7858b87db34232ac21f66c017c1

tests/queries/0_stateless/04299_at_time_zone.sql
lvzhipin03: Changed the test in 6213eb4990449d721b12da2e4d78ee1282d65e49

tests/queries/0_stateless/04300_at_local_distributed.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+# Tags: distributed
+# Regression test for AT LOCAL / timeZone() on Distributed tables.
+#
+# timeZone() is query-wide constant only when session_timezone is explicitly set
+# (the setting is propagated to remote shards). When session_timezone is empty,
```
tiandiwonder: optional: this header comment is now stale.

it still describes the empty-`session_timezone` per-shard path, but after the negative cases were removed nothing exercises that path anymore.

tests/queries/0_stateless/04300_at_local_distributed.sh
```diff
@@ -0,0 +1,33 @@
...
+# the effective timezone falls back to the server's local timezone which may
+# differ per shard, so timeZone() keeps per-shard evaluation (old behavior).
+
+CUR_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=../shell_config.sh
+. "$CUR_DIR"/../shell_config.sh
+
+$CLICKHOUSE_CLIENT -q "
```
tiandiwonder: optional: this could be a `.sql` test.

it's a single `$CLICKHOUSE_CLIENT -q` call with no shell logic, and `.sql` binds each `-- { serverError }` hint to its own statement, avoiding the multiquery attribution ambiguity that made the dropped negatives hard to pin down.

tests/queries/0_stateless/04300_at_local_distributed.sh
```diff
@@ -0,0 +1,33 @@
...
+    INSERT INTO t_at_local_dist_local VALUES ('2001-02-16 20:38:40');
+
+    CREATE TABLE t_at_local_dist AS t_at_local_dist_local
+        ENGINE = Distributed(test_shard_localhost, currentDatabase(), t_at_local_dist_local, rand());
+
+    -- AT LOCAL works when session_timezone is explicitly set (propagated to all shards)
+    SELECT dt AT LOCAL FROM t_at_local_dist ORDER BY dt SETTINGS session_timezone = 'UTC';
+
+    -- AT TIME ZONE with a constant string always works
+    SELECT dt AT TIME ZONE 'America/Denver' FROM t_at_local_dist ORDER BY dt;
```
tiandiwonder: optional: keep the negative cases (pinned with `allow_nonconst_timezone_arguments = 0`) instead of deleting them.

they were the PR's only distributed per-shard non-constant-timezone coverage, and only succeeded because the functional-test profile defaults that setting to `1`; `00515_enhanced_time_zones.sql` shows the pinned pattern.

tests/queries/0_stateless/04300_at_local_distributed.sh
lvzhipin03: fix in commit [626b07c](https://github.com/ClickHouse/ClickHouse/pull/106092/commits/626b07c5d0aca04b9161e92b9b067692f22cbc3e)

tests/queries/0_stateless/04300_at_local_distributed.sh
lvzhipin03: fix in commit [626b07c](https://github.com/ClickHouse/ClickHouse/pull/106092/commits/626b07c5d0aca04b9161e92b9b067692f22cbc3e)

tests/queries/0_stateless/04300_at_local_distributed.sh
lvzhipin03: fix in commit [626b07c](https://github.com/ClickHouse/ClickHouse/pull/106092/commits/626b07c5d0aca04b9161e92b9b067692f22cbc3e)

lvzhipin03:
@tiandiwonder Gentle ping — CI is green and the branch is conflict-free. Are there any outstanding issues on your end, or is this ready for merge?

alexey-milovidov (approved):
LGTM

PR #106094, opened 2026-05-29
Forbid creating minmax skip index on JSON (Object) columns to prevent NO_COMMON_TYPE exception when inserting mixed-type arrays

linhaojie:
When a JSON column has a minmax skip index, inserting data containing arrays with mixed element types (e.g., ["string", 1, false]) caused a crash in SerializationDynamic::serializeBinary(Field) because FieldToDataType tried to find a common supertype for incompatible element types via getLeastSupertype<Throw>.

The minmax index on a whole JSON column has very limited practical value (it only works for full object comparison like WHERE json > '{"a":"2"}'::JSON, not for subcolumn filtering). And the fundamental issue of losing type information through Field extraction makes it fragile regardless.

So forbidding JSON in minmax index is the better approach here

Forbid creating minmax skip index on JSON (Object) columns to prevent NO_COMMON_TYPE exception when inserting mixed-type arrays. Use typed subcolumns instead (e.g., `INDEX idx json.field TYPE minmax`). 

### Version info
- Merged into: `26.7.1.622` (included in `26.7` and later)


linhaojie:
@Diskein Could you help review this PR? It modifies FieldToDataType.cpp and ColumnDynamic.cpp to handle mixed-type arrays in JSON columns with skip indexes. I'm investigating the CI failures in arm_asan and arm_msan builds.

linhaojie:
@Avogar  Could you please take a look at this fix? It touches SerializationDynamic and FieldToDataType — areas you're most familiar with. Thanks!

src/Columns/ColumnDynamic.cpp
```diff
@@ -250,7 +254,29 @@ void ColumnDynamic::insert(const Field & x)
     }
 
     /// If we cannot insert field into current variant column, extend it with new variant for this field from its type.
-    auto field_data_type = applyVisitor(FieldToDataType(), x);
+    auto field_data_type = applyVisitor(FieldToDataType<LeastSupertypeOnError::Null>(), x);
+    if (!field_data_type)
```
Avogar: I don't like this approach. If we want to have `Dynamic` for types without a least supertype, we need to add new `LeastSupertypeOnError::Dynamic` similar to `LeastSupertypeOnError::Variant` and use it here and in `SerializationDynamic.`


And in general, usage of the JSON in minmax index sounds stragne to me, is this a real use case?

The overall problem with minmax index is the usage of Fields, we lose information about actual stored data types inside the object paths during Field extraction. Maybe it even can lead to problems even without the exception. So ideally we would need to get rid of Fields in the minmax indexes at all and use IColumn always.

Avogar (changes requested):

src/Columns/ColumnDynamic.cpp
Avogar: Maybe it makes sense to forbid the JSON column in minmax index...

linhaojie: > Maybe it makes sense to forbid the JSON column in minmax index...

Thanks for the review! You're right — I agree that forbidding JSON in minmax index is the better approach here. 

As you mentioned, the minmax index on a whole JSON column has very limited practical value (it only works for full object comparison like `WHERE json > '{"a":"2"}'::JSON`, not for subcolumn filtering). And the fundamental issue of losing type information through Field extraction makes it fragile regardless.

I'll simplify this PR to just add `isObject(column.type)` check in `minmaxIndexValidator`, consistent with how Dynamic/Variant are already forbidden. This will give a clear error message guiding users to use typed subcolumns instead (e.g., `INDEX idx json.field TYPE minmax`).

src/Storages/MergeTree/MergeTreeIndexMinMax.cpp
```diff
@@ -482,7 +484,8 @@ void minmaxIndexValidator(const IndexDescription & index, bool attach)
                 column.type->getName(), column.name);
         }
 
-        if (isDynamic(column.type) || isVariant(column.type))
+        auto nested_type = removeNullable(column.type);
```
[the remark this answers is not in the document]
linhaojie: 1. **Recursive type rejection**: Replaced the manual `isDynamic || isVariant || isObject` check with `column.type->hasDynamicStructure() || isVariant(nested_type)`. `hasDynamicStructure()` recursively checks through Nullable, Array, Tuple, Map, and LowCardinality wrappers — so `Nullable(JSON)`, `Array(JSON)`, `Tuple(Int, JSON)`, etc. are all properly rejected now.

2. **Existing test updated**: `04076_json_in_min_max_index_bug.sql` now verifies CREATE TABLE rejection instead of skip-index filtering, consistent with the new contract.

3. **Test coverage expanded**: Added `Array(JSON)` nested-type rejection case alongside `Nullable(JSON)` and plain `JSON`.

Regarding the `attach` path (finding #2, dismissed): this is consistent with existing behavior for Dynamic/Variant — the validator skips on attach, and users with legacy metadata will see the error on INSERT and need to `ALTER TABLE ... DROP INDEX`.

tests/queries/0_stateless/04076_json_in_min_max_index_bug.sql
```diff
@@ -1,21 +1,10 @@
 -- Tags: no-parallel-replicas
```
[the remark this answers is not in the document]
linhaojie: d

src/Storages/MergeTree/MergeTreeIndexMinMax.cpp
```diff
@@ -482,7 +484,8 @@ void minmaxIndexValidator(const IndexDescription & index, bool attach)
...
+        if (column.type->hasDynamicStructure() || isVariant(nested_type))
```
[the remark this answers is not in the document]
linhaojie: Added a recursive `contains_variant` check using `forEachChild`, which now rejects container-wrapped Variants like `Array(Variant(UInt8, String))` and `Tuple(Variant(...), Int32)`. This closes the gap where these types passed `isComparable` and `hasDynamicStructure` but later hit `SerializationVariant::serializeBinary(Field)` which throws NOT_IMPLEMENTED.

src/Columns/ColumnDynamic.cpp
linhaojie: @Avogar Hi! I've simplified the PR as you suggested — now it just forbids JSON/Dynamic/Variant columns in minmax index at validation time using `hasDynamicStructure()` + recursive Variant check. The FieldToDataType changes have been removed entirely.

Could you please re-review when you have a moment? Thanks!

src/Storages/MergeTree/MergeTreeIndexMinMax.cpp
```diff
@@ -482,7 +482,19 @@ void minmaxIndexValidator(const IndexDescription & index, bool attach)
...
+        std::function<bool(const IDataType &)> contains_variant = [&](const IDataType & t) -> bool
+        {
+            if (isVariant(t))
+                return true;
+            bool found = false;
+            t.forEachChild([&](const IDataType & child) { found = found || contains_variant(child); });
+            return found;
+        };
+
+        if (column.type->hasDynamicStructure() || contains_variant(*column.type))
```
Avogar: A few moments:
- I merged my [PR](https://github.com/ClickHouse/ClickHouse/pull/105468) with check for nested Variant/Dynamic types here, we need to resolve conflicts
- I am afraid we have to add a compatibility setting to allow creating minmax index on JSON but disable it by default. Because otherwise existing tables with such index will fail to startup on upgrade.

linhaojie: > A few moments:
> 
> * I merged my [PR](https://github.com/ClickHouse/ClickHouse/pull/105468) with check for nested Variant/Dynamic types here, we need to resolve conflicts
> * I am afraid we have to add a compatibility setting to allow creating minmax index on JSON but disable it by default. Because otherwise existing tables with such index will fail to startup on upgrade.

@Avogar  Thanks for the feedback! I'll rebase on your merged #105468 and add isObject to check_not_dynamic_or_variant.Regarding the compatibility setting — I see two approaches and would like your preference:

Approach A: Add ContextPtr parameter to the Validator function signature and pass it through MergeTreeIndexFactory::validate(). This would require changing all validator signatures (~10 files, 7 validators), but keeps the check inside minmaxIndexValidator.

Approach B: Follow the existing allow_suspicious_indices pattern — add a MergeTree setting (e.g., allow_minmax_index_for_json) and check it in MergeTreeData::checkMetadata() before calling validate(), similar to how allow_suspicious_indices works at line 968-972 and 1057-1064. This avoids changing the validator interface.

In both cases:
Default = false (JSON minmax index is forbidden for new tables)
attach path still skips validation (existing tables can load)
Users with existing JSON minmax indexes can set the flag to suppress the error if needed

Which approach do you prefer? I'm leaning toward B since it's less invasive for a bugfix PR.

Avogar: The MergeTree setting is better I think. For existing tables we can do similar to `allow_suspicious_indices` - don't check it on attach.

src/Core/Settings.cpp
```diff
@@ -1076,6 +1076,11 @@ In CREATE TABLE statement allows creating columns of type FixedString(n) with n
 )", 0) \
     DECLARE(Bool, allow_suspicious_indices, false, R"(
 Reject primary/secondary indexes and sorting keys with identical expressions
+)", 0) \
+    DECLARE(Bool, allow_minmax_index_for_json, false, R"(
```
Avogar, line 1134: There is no need in session setting after we added MergeTree setting

linhaojie: > There is no need in session setting after we added MergeTree setting

Removed the session/query-level setting entirely (`src/Core/Settings.cpp`). 
   `allow_minmax_index_for_json` is now a pure MergeTree table-level setting only, 
   consistent with your suggestion.

linhaojie:
**Key design decisions:**

- Object check: isObject(type) + forEachChild recursive scan in checkProperties, rejecting JSON / Nullable(JSON) / Array(JSON) etc.

- Setting evaluation uses engine-specific defaults via virtual dispatch:
1.  ALTER path: reads from *getDefaultSettings() (resolves to getReplicatedMergeTreeSettings() for ReplicatedMergeTree), then overlays new_metadata.settings_changes. This correctly handles RESET SETTING (setting removed from metadata → falls back to default false → rejects existing JSON index).
2. Construction path: reads from (*getSettings())[] which already contains server config + SETTINGS clause (merged by registerStorageMergeTree). Avoids calling virtual getDefaultSettings() during MergeTreeData construction where vtable is not yet dispatched to the subclass.

- Compatibility: attach always skips validation; registered in getMergeTreeSettingsChangesHistory under 26.6 (old=true, new=false).

**Test coverage (04278_json_minmax_index_mixed_type_array.sql):**

- Default rejection: JSON, Nullable(JSON), Array(JSON)
- Table-level SETTINGS allow_minmax_index_for_json = 1 allows creation
- ALTER TABLE RESET SETTING rejected while JSON minmax index exists
- DETACH/ATTACH + ALTER TABLE MODIFY SETTING upgrade/recovery path
- Typed subcolumn index (json.field) unaffected

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -930,13 +931,26 @@ static void checkKeyExpression(const ExpressionActions & expr, const Block & sam
     }
 }
 
+static bool getAllowMinmaxIndexForJSON(const StorageInMemoryMetadata & metadata, bool default_value)
```
Avogar: Now I understand why we have `allow_suspicious_indices` both as session setting and MergeTree setting. With just a `MergeTree` setting it makes it difficult to run alters. Now I think it's better to follow the pattern with `allow_suspicious_indices` and return 2 settings `allow_minmax_index_for_json` to simplify the code of alter (so it will be almost the same as for  `allow_suspicious_indices`, or `allow_minmax_index_for_json` requires some additional logic?).

So seems like my suggestion to use single `MergeTree` setting was not good, sorry for that.

linhaojie: > Now I understand why we have `allow_suspicious_indices` both as session setting and MergeTree setting. With just a `MergeTree` setting it makes it difficult to run alters. Now I think it's better to follow the pattern with `allow_suspicious_indices` and return 2 settings `allow_minmax_index_for_json` to simplify the code of alter (so it will be almost the same as for `allow_suspicious_indices`, or `allow_minmax_index_for_json` requires some additional logic?).
> 
> So seems like my suggestion to use single `MergeTree` setting was not good, sorry for that.

  Thanks, this makes sense. I changed the implementation to follow the same pattern as `allow_suspicious_indices`: `allow_minmax_index_for_json` now exists both as a query/session setting and as a `MergeTree` table setting.

  With this, `CREATE` and replica metadata replay use the `MergeTree` table/server setting, while `ALTER` validation can be suppressed by the query-level setting. This removes the special `settings_changes` overlay and the ReplicatedMergeTree
  mixed-ALTER special case.

  The intended ALTER pattern is now:

  ALTER TABLE t
      ADD INDEX ...,
      MODIFY SETTING allow_minmax_index_for_json = 1
  SETTINGS allow_minmax_index_for_json = 1;

  For ReplicatedMergeTree, the query-level setting only suppresses validation on the replica executing the ALTER. The table/server-level setting still has to be effective on replicas before they replay the replicated metadata containing the JSON
  minmax index.

  I also updated the stateless test to cover the failure of `ADD INDEX + MODIFY SETTING` without query `SETTINGS`, the successful query-level override path, `RESET SETTING`, and the ReplicatedMergeTree single-replica behavior.

linhaojie: @Avogar   Hi , I hit one more edge case with the dual-setting approach and would like your opinion.

  Current implementation follows `allow_suspicious_indices`:

  - `allow_minmax_index_for_json` exists both as a query/session setting and as a MergeTree table setting.
  - `checkProperties()` starts from the table setting, but when `local_context` exists it currently replaces it with the query setting.

  This causes a problem for tables that already opted in persistently:

  CREATE TABLE t
  (
      j JSON,
      INDEX idx j TYPE minmax GRANULARITY 1
  )
  ENGINE = MergeTree
  ORDER BY tuple()
  SETTINGS allow_minmax_index_for_json = 1;

  ALTER TABLE t ADD COLUMN x UInt8;

  The `ALTER ADD COLUMN` does not affect the index, but `checkAlterIsPossible()` builds the full post-ALTER `new_metadata` and calls `checkProperties(new_metadata, old_metadata, ..., local_context)`. Inside `checkProperties()` we iterate over
  all `new_metadata.secondary_indices`, not only indices affected by this ALTER. Since the query-level setting defaults to false, it overrides the table-level true and the existing JSON minmax index is rejected.

  So I see two possible directions:

  1. Keep following the `allow_suspicious_indices` shape, but change the precedence for this setting to:
     `effective_allow = table_level_allow || query_level_allow`.
     Then add an extra guard so query-only `ALTER ... ADD INDEX ... SETTINGS allow_minmax_index_for_json = 1` cannot persist a JSON minmax index unless the post-ALTER table-level setting is also enabled.

  2. Instead, make the JSON minmax validation delta-aware and only reject newly added or affected JSON minmax indices during ALTER. Existing metadata would not be revalidated by unrelated ALTERs.

  My current preference is option 1 because it keeps the existing full `checkProperties()` validation model and only fixes the settings semantics. But option 2 may be closer to “don’t reject existing metadata on unrelated ALTER”.

  Which behavior would you prefer?

Avogar: I would prefer option 1 as well. And let's add a test for it

src/Storages/MergeTree/MergeTreeData.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <mutex>
+#include <optional>
```
Avogar: It's not needed anymore, let's remove

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -1049,6 +1055,24 @@ void MergeTreeData::checkProperties(
                         checkSuspiciousIndices(index_expression_ptr);
                 }
 
+
```
Avogar: Let's avoid double blank lines for better style

linhaojie: done

src/Storages/MergeTree/MergeTreeData.h
linhaojie: done

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -4875,8 +4969,20 @@ void MergeTreeData::checkAlterIsPossible(const AlterCommands & commands, Context
         }
     }
 
+    auto settings_after_alter = getDefaultSettings();
+    if (new_metadata.settings_changes)
+        settings_after_alter->applyChanges(new_metadata.settings_changes->as<const ASTSetQuery &>().changes);
+    const bool allow_minmax_index_for_json_after_alter = (*settings_after_alter)[MergeTreeSetting::allow_minmax_index_for_json];
```
Avogar: I don't understand why we need it, it only overcomplicates things and makes code quality worse

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -1049,6 +1094,55 @@ void MergeTreeData::checkProperties(
...
+                                {
+                                    throw Exception(ErrorCodes::BAD_ARGUMENTS,
+                                        "{} data type of column {} is not allowed in minmax index when only query-level setting "
+                                        "'allow_minmax_index_for_json = 1' is enabled. Use MergeTree table setting "
+                                        "'allow_minmax_index_for_json = 1' to make this compatibility escape hatch persistent",
+                                        idx_column.type->getName(), idx_column.name);
+                                }
+
+                                const bool same_json_minmax_index_already_existed = same_json_minmax_index_existed(index);
+                                if (supportsReplication() && !same_json_minmax_index_already_existed)
```
Avogar: I don't understand why we need this, it only overcomplicates things and makes code quality worse.

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -1049,6 +1094,55 @@ void MergeTreeData::checkProperties(
...
 
+                if (!attach && index.type == "minmax")
+                {
+                    for (const auto & idx_column : index.sample_block)
+                    {
+                        auto check_json = [&](const IDataType & type)
+                        {
+                            if (isObject(type))
+                            {
+                                if (!allow_minmax_index_for_json)
```
Avogar: Why do we have 2 separate checks for `allow_minmax_index_for_json` and `table_allows_minmax_index_for_json` now? It looks strange. It's totally ok to create a minmax index on JSON if only session level setting is enabled and no MergeTree setting is enabled.

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -954,6 +957,16 @@ void MergeTreeData::checkProperties(
...
 
+    /// For ALTER, allow_minmax_index_for_json_table_setting is the post-ALTER durable table setting.
+    /// Query setting may suppress the current validation, but cannot create JSON minmax index metadata
+    /// without an already effective table-level allowance.
+    const bool current_table_allows_minmax_index_for_json = (*getSettings())[MergeTreeSetting::allow_minmax_index_for_json];
+    const bool table_allows_minmax_index_for_json = allow_minmax_index_for_json_table_setting.value_or(
+        current_table_allows_minmax_index_for_json);
+    bool allow_minmax_index_for_json = current_table_allows_minmax_index_for_json;
+    if (local_context)
+        allow_minmax_index_for_json |= local_context->getSettingsRef()[Setting::allow_minmax_index_for_json];
```
Avogar: I don't understand why we need to take into account table_allows_minmax_index_for_json. We should allow to create this index either when session setting is enabled or when table setting is enabled.

Avogar (changes requested):
After my last review things became overcomplicated and stragne.

src/Storages/MergeTree/MergeTreeData.cpp
linhaojie: Thanks, now I understand the intended direction.

  I agree that the current version overcomplicates this by treating allow_minmax_index_for_json as a durable metadata invariant. That was my attempt to avoid replicated replay problems, but it does not match the allow_suspicious_indices-style
  behavior.

  I’ll simplify the patch:
  - allow creating a JSON minmax index when either the MergeTree table setting or the session/query setting is enabled;
  - remove the post-ALTER durable setting calculation;
  - remove table_allows_minmax_index_for_json and the duplicated checks;
  - remove the ReplicatedMergeTree-specific new-index guard / same_json_minmax_index_existed logic;
  - adjust the tests so query-level allowance is accepted, including the replicated case.

  So the check will become a simple creation guard controlled by table setting OR session/query setting, instead of trying to enforce a replicated metadata invariant.

src/Storages/MergeTree/MergeTreeData.cpp
```diff
@@ -954,6 +957,10 @@ void MergeTreeData::checkProperties(
     if (local_context)
         allow_suspicious_indices = local_context->getSettingsRef()[Setting::allow_suspicious_indices];
 
+    bool allow_minmax_index_for_json = (*getSettings())[MergeTreeSetting::allow_minmax_index_for_json];
+    if (local_context)
+        allow_minmax_index_for_json |= local_context->getSettingsRef()[Setting::allow_minmax_index_for_json];
```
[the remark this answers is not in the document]
linhaojie: I don't think anyone would do that.

src/Storages/MergeTree/MergeTreeData.cpp
linhaojie: @Avogar You're right — the two separate checks were unnecessarily complex.

  I've simplified it to a single variable with `|=` semantics:

  Either session/query setting or table setting being true allows the index — no separate table_allows_minmax_index_for_json variable anymore.

  I also fixed the settings-only ALTER paths (both StorageMergeTree and StorageReplicatedMergeTree) where RESET SETTING / MODIFY SETTING ... = 0 could bypass the check and leave the table in an inconsistent state.

  Could you please re-review when you have a moment? Thanks!

src/Storages/MergeTree/MergeTreeData.cpp
linhaojie: Same as above

Avogar (approved):

Avogar:
Let's ignore the comments from the AI bot, we don't have to follow it strictly

linhaojie:
> Let's ignore the comments from the AI bot, we don't have to follow it strictly

@Avogar Got it, thanks for the clarification! I'll focus on your feedback going forward.  There was a merge conflict with master, I've resolved it

clickgapai:

### TL;DR
**Forbid creating minmax skip index on JSON (Object) columns to prevent NO_COMMON_TYPE exception when inserting mixed-type arrays**

**ClickGap verified:** reproduced it with the script below on `26.5`; tested 6 version(s) — 5 affected, 1 clean (see Affects); bisected to a candidate range — a single introducing PR could not be pinned (see Bisect).

### Reproduction
```sql
SET enable_json_type = 1;
CREATE TABLE t_json_minmax_bug
(
    id Int32,
    col1 JSON,
    INDEX col_idx col1 TYPE minmax GRANULARITY 1
) ENGINE = MergeTree() ORDER BY id;
INSERT INTO t_json_minmax_bug VALUES (1, '{"data":["CP-018673", 1, false]}');
-- Pre-fix: crashes with Code: 386 NO_COMMON_TYPE "There is no supertype for types String, Int8, Bool" during minmax granule serialization.
-- Fixed (PR #106094): CREATE is rejected with BAD_ARGUMENTS ("JSON data type ... is not allowed in minmax index").
```
[▶ Run on ClickHouse Fiddle](https://fiddle.clickhouse.com/da8d3108-2c49-40b6-b8ba-cca9770860bc)

### Bisect
- The bisect boundary was rejected — PR #101146 only adds a test for Map/Nested types with ALTER ADD COLUMN and lazy materialization; it does not touch minmax skip index creation logic or JSON type handling, so it cannot cause a bug related to minmax indexes on JSON columns. — a coincidental boundary, not the cause; no introducing PR pinned.

### Affects

### Suggested next step
The bug reproduces as far back as the oldest tested release (`26.1`) and the only converged boundary is topically unrelated — this is a long-standing / pre-existing bug in `comp-skip-index`, not a recent regression; no introducing PR exists within the tested range.

**Component:** `comp-skip-index`
**Severity:** `P2`


---
_ClickGap · Finding: `phase_d_pr106094`_




### Triage metadata
- Primary component: `comp-skip-index`
- Component owners (`comp-skip-index`): @shankar-iyer @rschu1ze

PR #106098, opened 2026-05-29
Set CTCACHE_S3_READ_ONLY for PR builds to fix AccessDenied errors

maxknv:
PR builds run on untrusted runners that lack `s3:PutObject` permission on the shared cache bucket. `clang-tidy-cache.py` was trying to upload entries on every cache miss and logging errors like:

```
ERROR:clang-tidy-cache.py:Error calling S3:put_object An error occurred (AccessDenied) when calling the PutObject operation: User: arn:aws:sts::542516086801:assumed-role/untrusted_runner/... is not authorized to perform: s3:PutObject
```

The fix mirrors the existing `SCCACHE_S3_READ_ONLY` pattern already in place for sccache: set `CTCACHE_S3_READ_ONLY=true` for PR builds (`pr_number > 0`) so untrusted runners can still read from the shared S3 cache but never attempt writes.

### Version info
- Merged into: `26.6.1.240`


pamarcos (approved):

leshikus (approved):
Looks good

PR #100586, opened 2026-03-24
Position based unused column removal

antaljanosbenjamin:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

Identify columns by position (instead of by name) when removing unused columns in the query plan. This enables unused-column removal when duplicated column names are present.

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

### Version info
- Merged into: `26.6.1.195`

PR #106104, opened 2026-05-29
Fix LOGICAL_ERROR for bare-column row policy filters

### LINKED ISSUE #106099: Row policy with a bare-column USING expression: LOGICAL_ERROR "Duplicate column name ... in row policy actions output" (26.x regression)
### Company or project name

_No response_

### Describe what's wrong

A **permissive row policy whose `USING` expression is a bare physical column reference** makes every query that reads that same column fail with a server-side `LOGICAL_ERROR`:

```
Code: 49. DB::Exception: Duplicate column name c0 in row policy actions output. (LOGICAL_ERROR)
```

This is a **regression in 26.x**: the same statements work correctly on 25.10 and 25.12. It fires with **default settings** and is **independent of `enable_analyzer`** (both `0` and `1` fail).

The error originates in the query-plan optimization pass `tryRemoveUnusedColumns` → `ReadFromMergeTree::removeUnusedColumns`: when the row-policy filter actions DAG references a bare column that is also a read column, the unused-column-removal optimization produces a duplicate column node in the row-policy actions output.

Found by a SQLancer-style row-policy oracle (`CREATE ROW POLICY ... USING <generated predicate>`), where the generated predicate was occasionally a bare column used as a truthy filter.

Yes — reproduces on `26.6.1.229` (head). Does **not** reproduce on `25.12.11.4` or `25.10.7.6` (returns the filtered rows as expected), so this is a 26.x regression.

### How to reproduce

`26.6.1.229`:

```sql
CREATE TABLE t (c0 Int32) ENGINE = MergeTree ORDER BY tuple();
CREATE ROW POLICY pol ON t USING c0 TO ALL;

SELECT c0 FROM t;
-- 26.6: Code: 49 ... Duplicate column name c0 in row policy actions output. (LOGICAL_ERROR)
-- 25.10 / 25.12: returns 1, 2, 3  (USING c0 keeps rows where c0 != 0)
```

Boundary conditions (all on `26.6.1.229`):

```sql
SELECT c1 FROM t;          -- OK   (query does not read the policy column)
SELECT count() FROM t;     -- Code 49
SELECT * FROM t;           -- Code 49 (the column is read)

-- any wrapping of the USING expression avoids it:
CREATE ROW POLICY pol ON t USING c0 + 0          TO ALL;  -- SELECT c0 -> OK
CREATE ROW POLICY pol ON t USING c0 != 0         TO ALL;  -- SELECT c0 -> OK
CREATE ROW POLICY pol ON t USING materialize(c0) TO ALL;  -- SELECT c0 -> OK
```

So the trigger is specifically a **bare column identifier** as the policy filter, combined with that column being read by the query.

### Expected behavior

`SELECT c0 FROM t` returns the rows kept by the policy (`c0 != 0`), i.e. `1, 2, 3` — the behavior on 25.10 / 25.12. A bare-column `USING` expression must not raise an internal `LOGICAL_ERROR`.

```
Code: 49. DB::Exception: Duplicate column name c0 in row policy actions output. (LOGICAL_ERROR)

3. DB::Exception::Exception<...>(int, FormatStringHelperImpl<...>, ...)
4. DB::ReadFromMergeTree::removeUnusedColumns(std::vector<unsigned long, ...> const&, bool)::$_2::operator()(DB::ActionsDAG const&, ...)
5. DB::ReadFromMergeTree::removeUnusedColumns(std::vector<unsigned long, ...> const&, bool)
6. DB::QueryPlanOptimizations::tryRemoveUnusedColumns(DB::QueryPlan::Node*, std::list<DB::QueryPlan::Node, ...>&, ...)
7. DB::QueryPlanOptimizations::optimizeTreeFirstPass(DB::QueryPlanOptimizationSettings const&, DB::QueryPlan::Node&, ...)
8. DB::QueryPlan::optimize(DB::QueryPlanOptimizationSettings const&)
9. DB::QueryPlan::buildQueryPipeline(...)
```

### Additional context

_No response_

### Version info
- Resolved by: #106438
- Merged into: `26.7.1.663` (included in `26.7` and later)
- Backported to: `26.6.2.51`

groeneai:
Closes #106099.

A row policy whose `USING` expression is a bare physical column reference (e.g. `CREATE ROW POLICY pol ON t USING c0 TO ALL`) made every query that reads `c0` fail on master with `Code: 49. DB::Exception: Duplicate column name c0 in row policy actions output. (LOGICAL_ERROR)`. The bug is a 26.x regression and is independent of `enable_analyzer`.

**Cause #1 — over-strict duplicate-output check.** `buildFilterInfo` (`Planner/Utils.cpp`) and `generateFilterActions` (`Interpreters/InterpreterSelectQuery.cpp`) append every required table input as an output of the filter actions DAG so that filtered columns remain available downstream. For a bare-column `USING` the filter result IS the input node, so the same `ActionsDAG::Node` ends up in the outputs list twice. `FilterTransform` already copes with this: `Block::erase(filter_column_name)` removes a single entry, leaving the other as the passthrough.

PR #100586 (`Position based unused column removal`) added a `NameSet` duplicate check at the top of `ReadFromMergeTree::removeUnusedColumns` that throws on this legitimate pattern. The fix relaxes it to use the node pointer in addition to the result name: same-pointer duplicates (the filter+passthrough trick) are allowed, while two distinct nodes with the same result name (a real bug for the position-based optimization) still abort. The position-based logic later in `removeUnusedColumns` operates on `output_header` (table columns, always uniquely named), not on the filter DAG outputs, so a same-pointer duplicate does not affect it.

**Cause #2 — projection optimization throws on non-`UInt8` filter columns.** Once the validator stops masking the first bug, `SELECT count() FROM t` (and any other aggregation) still fails with `Code: 59. ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER`. `projectionsCommon.cpp::findInOutputs` rejects any filter whose type is not `UInt8` / `Nullable(UInt8)`, even though `FilterTransform` itself accepts every type that satisfies `canBeUsedInBooleanContext`. `Int32`, `Int64`, etc. are valid runtime filters. The fix returns `nullptr` instead of throwing, so the projection optimizer gracefully gives up and the normal read path runs.

Regression test `04301_106099_row_policy_bare_column_duplicate.sql` covers: the reproducer from the issue, `SELECT *`, `SELECT count()`, `SELECT c0, sum(c1) GROUP BY c0`, `SELECT <other column>`, `enable_analyzer = 0` and `= 1`, and the wrapped-`USING` workarounds (`c0 + 0`, `c0 != 0`, `materialize(c0)`).

Fix `LOGICAL_ERROR` "Duplicate column name in row policy actions output" when a row policy `USING` expression is a bare column reference (e.g. `USING c0`) and the same column is read by the query. Also fix `ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER` thrown by projection optimization for any row policy whose filter type is not `UInt8` / `Nullable(UInt8)`.


groeneai:
**Pre-PR validation gate** (session `cron:clickhouse-ci-task-worker:20260529-113000`):

**(a) Deterministic repro?** Yes. `CREATE TABLE t (c0 Int32) ENGINE = MergeTree ORDER BY tuple(); CREATE ROW POLICY pol ON t USING c0 TO ALL; SELECT c0 FROM t` throws `Code: 49. LOGICAL_ERROR` on master 100% of the time on debug builds (aborts the server). With the patches in this PR, all reproducer queries from issue #106099 return the expected rows.

**(b) Root cause explained?** Yes, see the PR description. Two distinct 26.x regressions stack: PR #100586 (`Position based unused column removal`, 2026-05-27) added a name-based duplicate-output check that rejects the historical filter+passthrough pattern, and the projection helper at `projectionsCommon.cpp:129` (introduced 2026-03-30) throws `ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER` for any non-`UInt8` filter column even though `FilterTransform` accepts every `canBeUsedInBooleanContext` type.

**(c) Fix matches root cause?** Yes. (1) The validator is relaxed to use node pointer + result name, so legitimate same-pointer duplicates are tolerated while different-node same-name duplicates (the real failure mode for position-based optimization) still abort. (2) The projection helper returns `nullptr` instead of throwing, so the optimizer gracefully gives up and the normal read path handles the query.

**(d) Test intent preserved?** Yes. Existing row policy tests still pass: `01308_row_policy_and_trivial_count_query`, `01295_create_row_policy`, `01296_create_row_policy_in_current_database`, `01710_projection_row_policy`. The new test `04301_106099_row_policy_bare_column_duplicate` covers the reproducer, `SELECT *`, `SELECT count()`, `GROUP BY`, `SELECT <other column>`, `enable_analyzer = 0` and `= 1`, and the wrapped-`USING` workarounds.

**(e) Bidirectional demonstration?** Yes. Stashed the two `src/Processors/QueryPlan/...` patches, rebuilt from origin/master state, ran the regression test against the unfixed binary — the server crashed with `Logical error: 'Duplicate column name c0 in row policy actions output'` (exit code 134 / `SIGABRT`). Restored the patches, rebuilt, re-ran — test passes 20/20 with randomization.

groeneai:
cc @antaljanosbenjamin @novikd could you take a look? The first regression is in `ReadFromMergeTree::removeUnusedColumns` (PR #100586) and the second is in `projectionsCommon.cpp::findInOutputs` — both in `comp-query-optimizer`.

antaljanosbenjamin:
@groeneai Why not just ensure the same column is not added twice?

groeneai:
@antaljanosbenjamin Tested it. Producer-side dedup works for `buildFilterInfo` (Planner/Utils.cpp:564-566): skip the redundant push when `filter_input_node == filter_actions_outputs[0]` and flip `remove_filter_column = false`. With strict validator restored, all `enable_analyzer = 1` queries from the reproducer pass.

The legacy `generateFilterActions` path is harder. `ExpressionAnalyzer.cpp:2096` unconditionally sets `do_remove_column = true`, so stripping the duplicate there breaks `enable_analyzer = 0`: `FilterTransform` erases the only `c0` and the query fails with `NOT_FOUND_COLUMN_IN_BLOCK`. The override would need to become conditional on "filter is a computed column, not a bare input".

Three options:

1. Current PR (validator relax only): smallest change; both paths keep the duplicate.
2. Producer dedup in `buildFilterInfo` plus validator relax for the legacy duplicate.
3. Producer dedup in both paths plus conditional override in `ExpressionAnalyzer.cpp:2096`.

I lean toward (3) since it matches your question, but it touches `do_remove_column` propagation. Which do you prefer and I'll push the follow-up commit?

antaljanosbenjamin:
@groeneai Do the third.

groeneai:
@antaljanosbenjamin Done in d338e52e3288.

Producer-side dedup:
- `Planner/Utils.cpp::buildFilterInfo` (analyzer): when `filter_input_node == filter_actions_outputs[0]`, skip the redundant passthrough push and set `remove_filter_column = false`.
- `Interpreters/InterpreterSelectQuery.cpp::generateFilterActions` (legacy): same logic, plus `do_remove_column = false` so `FilterTransform` keeps the column.
- `Interpreters/ExpressionAnalyzer.cpp`: dropped the unconditional `row_policy_info->do_remove_column = true;` override so the producer's decision survives.
- `Processors/QueryPlan/ReadFromMergeTree.cpp`: restored the strict name-based duplicate validator from PR #100586.

Bot review on the test: addressed in the same commit. Data now includes `(0, 0)` and explicit `count()` / `countIf(c0 = 0)` queries, so the test fails not only on `LOGICAL_ERROR` reappearance but also if the bare-column policy silently stops being applied (a `count() == 3` mismatch catches it).

- Reproducer queries from #106099 pass under both `enable_analyzer = 0` and `enable_analyzer = 1`.
- `count()`, `SELECT *`, `SELECT c1`, and the three wrapped-`USING` variants (`c0 + 0`, `c0 != 0`, `materialize(c0)`) all return `3` rows (the `c0 = 0` row is filtered out).
- Existing row-policy tests still pass: `01308_row_policy_and_trivial_count_query`, `01295_create_row_policy`, `01296_create_row_policy_in_current_database`, `01710_projection_row_policy`, `01851_fix_row_policy_empty_result`, `02131_row_policies_combination`.

The Cause 2 fix in `projectionsCommon.cpp::findInOutputs` (graceful `nullptr` for non-`UInt8` filters) is unchanged.

Pre-PR validation gate (Option 3):
- Deterministic repro: yes -- single SQL file, no randomization sensitivity.
- Root cause: bare-column filter expressions produce a same-pointer duplicate in `FilterDAGInfo::actions.getOutputs()` because the producers (`buildFilterInfo` / `generateFilterActions`) unconditionally append every input as a passthrough output without checking whether it is already the filter result; PR #100586's name-based validator then rejects it.
- Fix matches root cause: producers now skip the redundant push and set `do_remove_column` consistently with whether the filter column was a bare input.
- Test intent preserved/strengthened: the `c0 = 0` row addition keeps the original `LOGICAL_ERROR` coverage and adds policy-applied coverage.
- Both directions demonstrated: locally confirmed via direct queries and via the test file diff against the new reference.
- General fix (not symptom): both planner paths fixed at the source; symmetric subsystems (`additional_filter_info`, `parallel_replicas_custom_filter_info`) are unaffected because their explicit `do_remove_column = true` sets are kept for the non-bare cases they target (those are passed-in `ASTPtr` row-level filters, never the bare-input case).

groeneai:
Pushed `75557e5` to address the CI failures on the prior Option 3 fixup.

**Fast test on `02680_illegal_type_of_filter_projection`** + **Style check `ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER is defined but not used`**: same root cause. The Option 3 commit turned the type check in `findInOutputs` from `throw` into a blanket `return nullptr`. Row policy `USING <int_column>` now works, but a user-issued `PREWHERE` with a non-`UInt8` filter type also stopped erroring -- which is exactly what `02680` was added to verify.

Surgical fix: `findInOutputs` gets a `lenient_type_check` parameter (default `false`). Only the `row_level_filter` call site passes `true`. `PREWHERE` and `FilterStep` callers keep the original throw, so user-issued filters of the wrong type still surface `ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER`.

### Pre-PR validation gate

a) **Deterministic repro?** Yes. `02680` query reproduces the throw on demand on the new build.
b) **Root cause explained?** The blanket `return nullptr` covered three call sites; row policies need leniency but PREWHERE/FilterStep do not.
c) **Fix matches root cause?** Differentiates the three call sites via `lenient_type_check`, only relaxing the one for row policies.
d) **Test intent preserved?** Both `02680` (throw expectation) and `04301` (row policy passthrough) keep their original semantics. `04301` was not weakened.
   - `02680`: throws `Code: 59 ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER` (matches reference's `serverError` annotation).
   - `04301`: matches reference under both `enable_analyzer = 0` and `1`.
   - Sweep: `01308_row_policy_and_trivial_count_query`, `01295_create_row_policy`, `01296_create_row_policy_in_current_database` all pass on the new build.
   - Row policy `USING` with `Int8` / `Float32` / `UInt16` columns all filter rows correctly.
f) **Fix is general?** Yes. Only the row policy code path takes the lenient branch; sibling code paths (`additional_filter_info`, `parallel_replicas_custom_filter_info`) do not use bare-input filters and are unaffected.

**Bugfix validation** check was inconclusive on both `d338e52` and the prior `ff66be3` (server core dump while running the regression test against `master HEAD` -- the LOGICAL_ERROR on the bare-column path crashes the debug binary instead of throwing cleanly). That state predates this commit and is the same on both prior commits; it does not block the actual functional CI.

groeneai:
Addressed in `a498f399b613` — three fixes layered on top of Option 3:

1. **`generateFilterActions` dedup actually deduplicates now.** The previous loop iterated over `inputs` (which is already deduped by the analyzer) but never reset the outputs list, so when the prerequisite columns happened to include the filter column itself the analyzer emitted the same output node twice and the strict validator restored in `d338e5` fired `LOGICAL_ERROR "Duplicate column name c0 in row policy actions output"` on every `SELECT c0` against a `USING c0` policy. Resetting `filter_outputs = {filter_result_node}` before re-exporting the remaining inputs fixes this.

2. **Dropped the unconditional `additional_filter_info->do_remove_column = true` override** at `InterpreterSelectQuery.cpp:1001` (your option (b)). `generateFilterActions` now sets the correct value: `false` for a bare-input filter, `true` for a computed filter. The matching override on `parallel_replicas_custom_filter_info` was always a no-op (that filter is generated as a computed `modulo(...) = 0` expression) — dropped for consistency.

3. **Made the `FilterStep` call site in `projectionsCommon::buildImpl` lenient** about non-`UInt8` filter columns. `SELECT count() FROM t SETTINGS additional_table_filters = {'t':'<bare-int-column>'}` regressed under the previous fixup because the strict `findInOutputs` call for `FilterStep` threw `ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER` even though `FilterTransform` itself accepts any `canBeUsedInBooleanContext` type. The PREWHERE call site above stays strict, so `02680_illegal_type_of_filter_projection` keeps surfacing the type error for explicit user `PREWHERE <non-bool>`.

Regression test additions in `04301`:
- `SELECT c0 FROM t SETTINGS additional_table_filters = {'t':'c0'}` under both `enable_analyzer = 0` and `1` (your reproducer)
- Matching `count()` cases that previously hit `ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER`

### Pre-PR validation gate

- **(a) Deterministic repro?** Yes. Without `a498f399b613` the row policy + `SELECT c0` reproducer aborts on `ReadFromMergeTree::removeUnusedColumns`' duplicate-name validator (`LOGICAL_ERROR`, Signal 6 in debug). `SELECT count()` with bare `additional_table_filters` deterministically returns Code 59 `ILLEGAL_TYPE_OF_COLUMN_FOR_FILTER`. Both repros are in `04301`.
- **(b) Root cause explained?** Yes, three causal chains: (1) the previous dedup loop iterated over inputs but never reset outputs, missing the analyzer's prereq-induced duplication; (2) the unconditional override at line 1001 overruled the producer's correct decision for bare-input filters; (3) the strict findInOutputs at `projectionsCommon.cpp:212` rejected non-`UInt8` filter columns even though `FilterTransform` accepts them, breaking projection optimization for bare-integer additional filters.
- **(c) Fix matches root cause?** Yes. Reset outputs at the producer; drop the overriding writes; pass `lenient_type_check = true` at the single call site that matters. No widened assertions or `no-random-*` tags.
- **(d) Test intent preserved AND strengthened?** Yes. `02680` still throws as before (PREWHERE call site stays strict). `04301` now covers `additional_table_filters` bare-column + `count()` shapes that previously failed silently in CI.
- **(e) Both directions demonstrated locally?** Yes. Built `build/programs/clickhouse` debug binary; `04301` is 20/20 passes against the new reference. `02680` still returns `Code 59. DB::Exception: Illegal type Float64 of column ... for filter`. Existing tests `01295`, `01308`, `01710_projection_row_policy`, `01710_projection_additional_filters`, `01851`, `02131`, `03033`, `03591`, `03641` all pass unchanged.
- **(f) Fix is general, not a narrow patch?** Yes. The dedup fix at `generateFilterActions` covers the legacy analyzer for both `row_policy` and `additional_table_filters` callers. The analyzer-path equivalent in `Planner/Utils.cpp::buildFilterInfo` already had the right behavior (resets outputs to `expression_nodes` before adding passthroughs). The lenient `findInOutputs` change at `projectionsCommon.cpp:212` covers every projection-optimization call into a `FilterStep`, not just bare-Int32; PREWHERE and the row-level-filter call sites are untouched.

tests/queries/0_stateless/04301_106099_row_policy_bare_column_duplicate.sql
```diff
@@ -0,0 +1,78 @@
...
+CREATE ROW POLICY pol_106099 ON t_106099 USING materialize(c0) TO ALL;
+SELECT c0 FROM t_106099 ORDER BY c0;
+SELECT count() FROM t_106099;
+
+DROP ROW POLICY pol_106099 ON t_106099;
+
+-- Same producer-side dedup also runs for `additional_table_filters`. Without
+-- the fix at `InterpreterSelectQuery.cpp` the override would force
+-- `FilterTransform` to erase the only `c0` output, breaking `SELECT c0`.
+SELECT c0 FROM t_106099 ORDER BY c0 SETTINGS additional_table_filters = {'t_106099':'c0'}, enable_analyzer = 0;
```
antaljanosbenjamin, line 69: We need a test with row policy + additional_table_filters and probably prewhere and normal filters.

antaljanosbenjamin (commented):
In general reduce the number and lengthiness of comments. Make them concise and refrain from explaining other parts of the code: the other parts might change overtime which would make the comment misleading.

groeneai:
@antaljanosbenjamin Addressed in 613234708a82. Each comment now states what the local code does and why, without explaining other code paths (`FilterTransform`, `ReadFromMergeTree::removeUnusedColumns`, `generateFilterActions`, `canBeUsedInBooleanContext`, the constant-rewrite logic).

Same commit also addresses the bot's earlier UInt8 projection concern: when `findInOutputs` would have rewritten a passthrough filter column to constant `1`, lenient-mode callers now bail on projection so `SELECT c0` returns the real values. Test `04301_106099_row_policy_bare_column_duplicate` adds a `UInt8` projection block under both analyzer paths.

antaljanosbenjamin (commented):
It is getting too complicated. Let's revert all the changes and make ReadFromMergeTree not to remove unused columns in case of duplicated column names.

antaljanosbenjamin:
@groeneai can you resolve conflicts?

antaljanosbenjamin:
@groeneai I resolved the conflicts for you

antaljanosbenjamin:
Probably #106438 is better

groeneai:
@antaljanosbenjamin Agreed, closing this in favor of #106438. Your position-based pruning refactor is much cleaner than the Option 3 fixups, and your 04305 plus 02680 tests cover the same cases. Will monitor #106438.

groeneai:
Superseded by #106438 (cleaner approach by @antaljanosbenjamin).

PR #106105, opened 2026-05-29
Fix `system.dictionaries` returning 0 rows with partial `SHOW DICTIONARIES` revoke

Avogar:

Fix `system.dictionaries` returning 0 rows with partial `SHOW DICTIONARIES` revoke. 

### Version info
- Merged into: `26.6.1.243`
- Backported to: `26.5.2.20`, `26.4.4.25`, `26.3.13.21`, `25.8.25.18`


src/Storages/System/StorageSystemCompletions.cpp
```diff
@@ -312,8 +312,7 @@ void fillDataWithPolicies(MutableColumns & res_columns, const ContextPtr & conte
 void fillDataWithDictionaries(MutableColumns & res_columns, const ContextPtr & context)
 {
     const auto & access = context->getAccess();
-    if (!access->isGranted(AccessType::SHOW_DICTIONARIES))
-        return;
+    const bool need_to_check_access_for_dictionaries = !access->isGranted(AccessType::SHOW_DICTIONARIES);
```
Ergus, line 315: Isn't this condition good enough. I mean, is access is not Granted checking a second condition may open potential breaches.

Avogar: `access->isGranted(AccessType::SHOW_DICTIONARIES)` will return false if there is a revoke on `SHOW_DICTIONARIES` on any of the database/table/dictionary. So we won't see all other dictionaries that we have access to. Instead, in this case we should check access to each dictionary independently in a loop. If `access->isGranted(AccessType::SHOW_DICTIONARIES)` returns true, it means we have access to all dictionaries.

Ergus (approved):

src/Storages/System/StorageSystemCompletions.cpp
grantholly-clickhouse: This was new to me too, apparently we handle this differently for `AccessType::SHOW_TABLES`.  Here's an example: https://github.com/ClickHouse/ClickHouse/blob/master/src/Storages/System/StorageSystemTables.cpp#L518

clickgapai:

### TL;DR
**Fix `system.dictionaries` returning 0 rows with partial `SHOW DICTIONARIES` revoke**

### Reproduction
```
clickhouse-client -mn -q "CREATE DICTIONARY d (id UInt64, value String) PRIMARY KEY id SOURCE(CLICKHOUSE(TABLE 'numbers' DB 'system')) LIFETIME(MIN 0 MAX 300) LAYOUT(FLAT()); CREATE ROLE r; GRANT SHOW DICTIONARIES, SELECT ON *.* TO r; REVOKE SHOW DICTIONARIES, SELECT ON system.non_existing_table FROM r; CREATE USER u IDENTIFIED WITH no_password; GRANT r TO u;"; clickhouse-client --user u -q "SELECT name FROM system.dictionaries WHERE name='d'"  -- 0 rows on buggy build (partial SHOW DICTIONARIES revoke hides ALL dicts), 1 row when fixed
```

### Bisect

### Affects

### Suggested next step
Manual triage of the `Storages` path is needed. The reproducer above should be a fast first check.

**Component:** `Storages`
**Severity:** `P2`

cc fix PR author: @Avogar
cc fix PR reviewer: @Ergus

---
_ClickGap · Finding: `phase_d_pr106105`_

PR #106107, opened 2026-05-29
Add native Mapbox Vector Tile (MVT) encoding functions

saarthak2002:
## Motivation

Web map clients (MapLibre GL, Mapbox GL) render [Mapbox Vector Tiles](https://github.com/mapbox/vector-tile-spec) (MVT) — a protobuf format — natively. Serving data from ClickHouse as slippy-map tiles currently requires an external service that reads rows out of ClickHouse and re-encodes them into MVT bytes, adding a network hop and a separate component to operate even though the data and aggregation already live in ClickHouse.

These functions let a complete vector tile be produced entirely in SQL and served straight off the HTTP interface with `FORMAT RawBLOB`, so a `/tile/{z}/{x}/{y}` endpoint becomes a single parameterized query behind a `predefined_query_handler`.

## What's added

- `MVTEncodeGeom(geometry, zoom, tile_x, tile_y[, extent[, buffer[, clip]]])` — scalar; projects a geometry given in longitude/latitude into the tile-local pixel space (Web Mercator, top-left origin), clips it to the tile expanded by `buffer` pixels (default `1`), and snaps to whole pixels so `GROUP BY MVTEncodeGeom(...)` performs pixel clustering. Returns a `Geometry`, or `NULL` when the input is fully clipped out. Point, line and polygon geometry is supported (`Point`, `LineString`, `MultiLineString`, `Ring`, `Polygon`, `MultiPolygon`, `Geometry`).
- `MVTEncode('layer'[, extent[, feature_id_name[, stringify_unsupported]]])(geometry[, properties])` — aggregate; collects one feature per row into the bytes of a single-layer tile. `properties` is a named tuple whose element names become attribute keys and whose types map to MVT value types. `feature_id_name` optionally promotes an unsigned-integer element to the MVT Feature `id`; `stringify_unsupported` encodes otherwise-unsupported types as their text value instead of raising.
- `MVTBoundingBox(zoom, tile_x, tile_y[, margin])` and `MVTBoundingBoxMercator(...)` — return a tile's bounding box, so the `WHERE` prefilter can run against indexed `longitude`/`latitude` (or Mercator) columns.

Also available under their PostGIS names as aliases: `ST_AsMVTGeom` for `MVTEncodeGeom` and `ST_AsMVT` for `MVTEncode`. All in the `Geo` category. Documentation: `docs/en/sql-reference/functions/geo/mvt.md`. Clustering is expressed in SQL (`GROUP BY` + `count()`), not as a function parameter.

## Example

```sql
WITH
    1 AS buffer,
    4096 AS extent,
    MVTBoundingBox({z:UInt8}, {x:UInt32}, {y:UInt32}, buffer / extent) AS bb
SELECT MVTEncode('points')(geom, tuple(cluster_count)::Tuple(cluster_count UInt64))
FROM
(
    SELECT MVTEncodeGeom((lon, lat)::Point, {z:UInt8}, {x:UInt32}, {y:UInt32}) AS geom, count() AS cluster_count
    FROM points
    WHERE lon BETWEEN bb.1 AND bb.3 AND lat BETWEEN bb.2 AND bb.4
    GROUP BY geom
)
SETTINGS allow_suspicious_types_in_group_by = 1
```

## Notes for reviewers

- The MVT protobuf bytes are written directly (`src/AggregateFunctions/MVTProtobuf.h`) rather than via the generic protobuf library, to avoid per-feature message allocations.
- The Web Mercator world spans exactly `2^32` so that a tile of size `2^(32 - zoom)` tessellates the world without drift at high zoom.
- Clipping lives in the scalar `MVTEncodeGeom`, which returns a `Geometry` (a `Variant` that carries `NULL` natively for fully-clipped input). This avoids a `Nullable(Tuple(...))` return, which is a normally-disallowed type and faults under expression JIT.
- The aggregate state begins with a format-version byte, and the parser used by `merge`/`deserialize` validates lengths with size-based bounds checks (`length > end - pos`) so a corrupted state cannot cause a pointer overflow or out-of-bounds read.
- The properties tuple must have explicit element names; since aliases inside `tuple(...)` aren't propagated to element names, they're named with a cast, e.g. `tuple(count())::Tuple(cluster_count UInt64)`.
- Two limitations are documented: polygon clipping fixes orientation and closure but not self-intersections (a bow-tie ring is dropped or passed through, not repaired), and geometry is clipped on the floating-point projection before being rounded to the pixel grid (a coordinate that would round onto the boundary can be clipped).

Tested with stateless tests `04299_mvt_encode` and `04300_mvt_clip`, whose emitted bytes were decoded back and checked against the official `vector_tile.proto` (via `protoc`) and the canonical `mapbox-vector-tile` library, plus end-to-end by serving tiles over HTTP and rendering them in QGIS.

- New Feature

Added functions for serving Mapbox Vector Tiles directly from SQL: `MVTEncodeGeom` projects a geometry into the pixel space of a slippy-map tile and clips it, `MVTEncode` aggregates the projected geometries of a group into the binary bytes of a single-layer tile, and `MVTBoundingBox` / `MVTBoundingBoxMercator` return the bounding box of a tile for restricting rows to it. Point, line and polygon geometry is supported. Also available under the PostGIS aliases `ST_AsMVTGeom` and `ST_AsMVT`.

### Version info
- Merged into: `26.6.1.1112`


saarthak2002:
Hello,

Thanks for the review and assignment! The remaining red checks look like unrelated with this change:

- Stateless tests (arm_asan_ubsan, targeted) 
   [WARNING: Timeout exceeded [2818], sending SIGTERM to process group [5725]]
- Integration tests (amd_tsan, 2/6)
  [test_kafka_read_consumers_in_parallel]

Could you re-run these two jobs when you get a chance please?

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,286 @@
...
+doc_type: 'reference'
+---
+
+## Overview {#overview}
+
+[Mapbox Vector Tiles](https://github.com/mapbox/vector-tile-spec) (MVT) are the protobuf-encoded tiles that web map
+clients such as MapLibre and Mapbox GL render natively. ClickHouse can build such tiles entirely in SQL with a pair of
+cooperating functions:
+
+- `mvtEncodeGeom` — a scalar function that projects a geographic point into the tile-local pixel space of a slippy-map
```
Algunenano: These names are incorrect:
* MVT is an acronym and it should be in all caps.
* There is already standard names in PostGIS, then adopted by others like DuckDB, to use ST_AsMVT / ST_AsMVTGeom. We can have other names if we need to, simpler ones, but keep the alias and have a similar parameter order and names.

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,286 @@
...
+  tile.
+- `mvtEncode` — an aggregate function that collects the projected points of a group into the binary bytes of a
+  single-layer tile, optionally clipping features to the tile.
+
+Two helper functions, `mvtTileBBox` and `mvtTileBBoxMercator`, return the bounding box of a tile so that rows can be
+restricted to it in the `WHERE` clause using an index.
+
+The resulting bytes are a complete tile that can be returned directly over the HTTP interface with `FORMAT RawBLOB`.
```
Algunenano, line 27: Why do we return RawBLOB and not protobuf?

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,286 @@
...
+
+Only point geometry is supported.
```
Algunenano: We should support the Geometry type, not just points

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,286 @@
...
+(
+    SELECT mvtEncodeGeom(lon, lat, 10, 550, 335) AS geom, count() AS cluster_count
+    FROM points
+    GROUP BY geom
+);
+```
+
+Omit the inner `GROUP BY` (and `count()`) to emit one feature per input row instead of clustered features.
+
+## mvtTileBBox {#mvttilebbox}
```
Algunenano: ST_TileEnvelope

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Restricting rows to a tile {#restricting-rows-to-a-tile}
+
+A tile must only contain the points that belong to it. This is best expressed as two cooperating steps: a cheap,
+index-using bounding-box predicate in the `WHERE` clause (performance), and the `buffer` clip parameter of `mvtEncode`
+(correctness). The clip drops features outside the tile during aggregation, so even a loose bounding-box
+predicate cannot leak out-of-tile points into the result.
+
+```sql
+WITH mvtTileBBox({z:UInt8}, {x:UInt32}, {y:UInt32}, 0.0156) AS bb   -- margin = 64 / 4096
```
Algunenano: Let's have good defaults to work with webmercator, as ST_TileEnvelope does

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+A `GET /tile/10/550/335` then returns the encoded tile. Omit the inner `GROUP BY` and `count()` for unclustered tiles.
+
+## Limitations {#limitations}
+
+- Point geometry only; line and polygon geometry are not supported. Clipping (the `buffer` parameter of `mvtEncode`)
+  is likewise point-only.
+- The Web Mercator projection clamps latitude to `±85.05112878°` and does not handle antimeridian-crossing inputs.
+- No explicit feature `id` field.
```
Algunenano: Why? It should be optional and supported

src/AggregateFunctions/AggregateFunctionMVTEncode.cpp
```diff
@@ -0,0 +1,515 @@
...
+{
+
+/// Which Mapbox Vector Tile `Value` variant a property column is encoded as.
+enum class MvtValueKind : UInt8
+{
+    String, /// string_value (field 1)
+    Float, /// float_value (field 2)
+    Double, /// double_value (field 3)
+    Int, /// int_value (field 4)
+    UInt, /// uint_value (field 5)
```
Algunenano, line 89: I don't see sint or bool

src/AggregateFunctions/AggregateFunctionMVTEncode.cpp
```diff
@@ -0,0 +1,515 @@
...
+};
+
+struct MvtProperty
+{
+    String key;
+    MvtValueKind kind;
+};
+
+/// Variable-length aggregate state held in the arena: a flat buffer of self-contained feature records.
```
Algunenano, line 102: This is not ok. If a million rows have the same value this will add the value a million times, not sharing it. Yes, this works but it will create massive tile sizes for no good reason. MVT is designed to share properties, so this is a must.

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
```diff
@@ -0,0 +1,515 @@
...
+        const DataTypePtr base = removeNullable(recursiveRemoveLowCardinality(element_type));
+        WhichDataType which(base);
+
+        if (which.isStringOrFixedString())
+            return MvtValueKind::String;
+        if (which.isFloat32())
+            return MvtValueKind::Float;
+        if (which.isFloat64())
+            return MvtValueKind::Double;
+        if (which.isNativeInt())
```
Algunenano: Date32 (which is a Int32) can be matched to ::Int here

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
```diff
@@ -0,0 +1,515 @@
...
+        throw Exception(
+            ErrorCodes::ILLEGAL_TYPE_OF_ARGUMENT,
+            "Property of type {} is not supported by aggregate function {}. Supported types are String, FixedString, "
+            "Float32, Float64, (U)Int8/16/32/64, Date and DateTime (optionally Nullable and/or LowCardinality); cast "
+            "other types with toString() or a numeric cast",
+            element_type->getName(),
+            function_name);
+    }
+
+    static void appendFixedInt64(String & out, Int64 value)
```
Algunenano: Please prefer anonymous namespaces to static functions

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
```diff
@@ -0,0 +1,515 @@
...
+    {
+        const UInt64 bits = static_cast<UInt64>(value);
+        for (size_t i = 0; i < sizeof(bits); ++i)
```
Algunenano: Let's use unalignedStoreLittleEndian instead of redefining helpers. Something along the lines:
```
void appendFixedInt64(String & out, Int64 value)
{
    char buf[sizeof(Int64)];
    unalignedStoreLittleEndian<Int64>(buf, value);
    out.append(buf, sizeof(buf));
}
```

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
```diff
@@ -0,0 +1,515 @@
...
+            out.push_back(static_cast<char>((bits >> (8 * i)) & 0xFF));
+    }
+
+    static Int64 readFixedInt64(const char *& pos, const char * end)
+    {
+        if (sizeof(UInt64) > static_cast<size_t>(end - pos))
+            throw Exception(ErrorCodes::INCORRECT_DATA, "Corrupted mvtEncode aggregate state: truncated coordinate");
+        UInt64 bits = 0;
```
Algunenano: Same here: unalignedLoadLittleEndian

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+    static UInt64 readVarint(const char *& pos, const char * end)
+    {
+        UInt64 result = 0;
+        int shift = 0;
+        while (true)
+        {
+            if (pos >= end || shift >= 64)
+                throw Exception(ErrorCodes::INCORRECT_DATA, "Corrupted mvtEncode aggregate state: truncated varint");
+            const unsigned char byte = static_cast<unsigned char>(*pos++);
```
Algunenano: Check IO/VarInt.h, which already implements this

src/AggregateFunctions/MVTProtobuf.h
```diff
@@ -0,0 +1,88 @@
...
+    Varint = 0,
+    Fixed64 = 1,
+    LengthDelimited = 2,
+    Fixed32 = 5,
+};
+
+/// Append an unsigned integer as a protobuf base-128 varint.
+inline void writeVarint(std::string & out, UInt64 value)
+{
+    while (value >= 0x80)
```
Algunenano: IO/Varint.h for this, and for zigzag. Let's not reimplement everything.

Algunenano (changes requested):
Partial review. Only up to the aggregation

I don't quite understand why you decided to use a manual write instead of using protobuf, which with your architecture is only used at the end (in `insertResultInto`). What benefits those is bring? Because the need to review and verify for correctness is a clear drawback, specially when the behaviour is not set and proven.

Please pay attention to the AI review comments and either address them or dismiss them.

Also, please consider carefully what functions you are adding (many already existed in the project) and follow a similar architecture and setup as other functions, reusing existing code, data structures, etc.

src/AggregateFunctions/AggregateFunctionMVTEncode.cpp
saarthak2002: The values in the emitted tile are shared- [insertResultInto](https://github.com/ClickHouse/ClickHouse/blob/b18f6a19add7764c05ce456df0967cf61ba00b83/src/AggregateFunctions/AggregateFunctionMvtEncode.cpp#L356-L361) dedups into the layer's value pool, so a million identical values produce one pool entry and every feature references it by index via tags:

https://github.com/ClickHouse/ClickHouse/blob/b18f6a19add7764c05ce456df0967cf61ba00b83/src/AggregateFunctions/AggregateFunctionMvtEncode.cpp#L356-L361

This is covered by a regression test — a tile with one distinct value is strictly smaller than the same two features with distinct values, which can only hold if the value is interned once:
https://github.com/ClickHouse/ClickHouse/blob/b18f6a19add7764c05ce456df0967cf61ba00b83/tests/queries/0_stateless/04299_mvt_encode.sql#L38-L51

If the intermediate state duplication is an issue however, I can also maintain a dictionary with remapping  on merge for the state. But the massive tile sizes should not appear because of the value pooling during final protobuf assembly in `insertResultInto`. I'll also fix the misleading comment here.

Algunenano: Yes sorry. I thought this was building the tile already (it's possible), but then realized it didn't and never went back to delete the comment. My bad

saarthak2002: No worries!

docs/en/sql-reference/functions/geo/mvt.md
saarthak2002: Reworked to support full geometry: mvtEncodeGeom/mvtEncode now operate on the Geometry type (points, lines, polygons and their multi/ring variants) instead of just point tuples, reusing geometryConverters.h + Boost.Geometry, etc. This was a major rework- so will address other comments now gradually.

Screenshot 2026-06-02 at 12 27 28 PM

ClickTilerGeom1

ClickTilerGeom2

alexey-milovidov:

docs/en/sql-reference/functions/geo/mvt.md
saarthak2002: The tile is protobuf- mvtEncode returns a String holding the raw MVT protobuf bytes. This is just an example of a custom http handler that uses RawBLOB to write the raw bytes over an http channel that a client library like MapBox would expect.

saarthak2002:
Thanks for the review! Here's what's addressed so far:

- IO/VarInt.h helpers: the writers now use `writeVarUInt`, `readVarUInt` and `encodeZigZag` from IO/VarInt.h, and `unalignedStoreLittleEndian`
- Aliases: added `ST_TileEnvelope`, `ST_AsMVT`, and `ST_AsMVTGeom` for `mvtTileBBoxMercator`, `mvtEncode`, and `mvtEncodeGeom`. I kept the primary mvt* names in lowerCamelCase to match ClickHouse's convention (`geoToH3`, `pointInPolygon`) with the ST_* forms as aliases
- Full Geometry support: the functions now accept the generic `Geometry` type (and the concrete geo types), with clipping handled in `mvtEncodeGeom`.
- Value types: added the `sint` and `bool` value encodings, plus `Date32` handling.
- clickhouse-gh bot: all inline comments addressed.

On manual encoding vs using the protobuf library: only the Tile/Layer/Feature framing could reuse an existing implementation thats what `MVTProtobuf.h` does right now. The geometry field itself is a hand written uint32 command stream regardless (MoveTo / LineTo / ClosePath with zig-zag deltas, built in `emitLineTo` / `emitRing` / `emitPolygon` etc). So adopting the library would replace only the ~80 lines of outer framing, not the geometry encoding. In my testing the manual output is byte-for-byte identical to the library on every tile size, and keeping it manual avoids checking in a `.proto` and wiring up codegen.

That said, I can switch the framing over to the protobuf library if needed

Algunenano:
Thanks. I'll have a look next week. Now I remember how painful it was to make sure polygons were valid on MVT after  the flip + grid. You seem to have gone the approach of correct + intersect + correct + grid and no correct at the end, which IIRC it led to some issues and it might be possible to just do all the process and correct on integer coordinates at the end, but I think we can live with this process and evolve in the future if needed. Just mentioning in case you have any thoughts, but no need to address it.

saarthak2002:
> Thanks. I'll have a look next week. Now I remember how painful it was to make sure polygons were valid on MVT after the flip + grid. You seem to have gone the approach of correct + intersect + correct + grid and no correct at the end, which IIRC it led to some issues and it might be possible to just do all the process and correct on integer coordinates at the end, but I think we can live with this process and evolve in the future if needed. Just mentioning in case you have any thoughts, but no need to address it.

Thanks for the feedback. The final correct does happen, just in the aggregate in `mvtEncode`'s `emitRing`, not in the scalar:

https://github.com/ClickHouse/ClickHouse/blob/2940746873041b23d276e8311b7510d04c0a0c39/src/AggregateFunctions/AggregateFunctionMvtEncode.cpp#L301-L313

It prunes consecutive duplicates, drops zero-area rings, and corrects the winding. So for a call like `mvtEncode('layer')(mvtEncodeGeom(geom, z, x, y))` the integer-coordinate correction is applied at the end.

The reason to split it up was that `mvtEncode`'s argument does not have to come from `mvtEncodeGeom`. A user can feed `mvtEncode` any tile-space geometry. The validity rules apply to the command stream generated by `mvtEncode` as per the tile spec, so it has to re-establish them itself. Thanks!

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+**Arguments**
+
+- `geometry` — Geometry in longitude/latitude degrees. Longitude is clamped to `[-180, 180]` and latitude to the Web Mercator range `[-85.05112878, 85.05112878]`. [`Point`](../../data-types/geo.md) / [`LineString`](../../data-types/geo.md) / [`MultiLineString`](../../data-types/geo.md) / [`Ring`](../../data-types/geo.md) / [`Polygon`](../../data-types/geo.md) / [`MultiPolygon`](../../data-types/geo.md) / [`Geometry`](../../data-types/geo.md).
+- `zoom` — Slippy-map zoom level, in the range `[0, 32]`. [`UInt8`](../../data-types/int-uint.md).
+- `tile_x` — Tile column index, in the range `[0, 2^zoom - 1]`. [`UInt32`](../../data-types/int-uint.md).
+- `tile_y` — Tile row index, in the range `[0, 2^zoom - 1]`. [`UInt32`](../../data-types/int-uint.md).
+- `extent` — Optional tile extent in pixels per side, in the range `[1, 2147483647]`. Defaults to `4096`, the Mapbox Vector Tile default. [`UInt32`](../../data-types/int-uint.md).
+- `buffer` — Optional clip buffer in pixels, in the range `[0, 2147483647]`. Defaults to `256`. [`UInt32`](../../data-types/int-uint.md).
```
Algunenano: I remember 256 being too much, and basically kept as is for compatibility. I don't think it's a good default as renderers don't need this much extra buffer to do stitching which, talking from memory, is the main reason to use it. I'd consider 0 or 1 to be better defaults, even if it differs from PostGIS. Any thoughts?

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Restricting rows to a tile {#restricting-rows-to-a-tile}
+
+A tile must only contain the geometry that belongs to it. This is best expressed as two cooperating steps: a cheap,
+index-using bounding-box predicate in the `WHERE` clause (performance), and the clip of `mvtEncodeGeom` (correctness).
+The clip drops geometry outside the tile, so even a loose bounding-box predicate cannot leak out-of-tile geometry into
+the result.
+
+```sql
+WITH mvtTileBBox({z:UInt8}, {x:UInt32}, {y:UInt32}, 0.0625) AS bb   -- margin = 256 / 4096
```
Algunenano: Let's use bounding_box instead of bb, and use explicit values for the margin.

Something like: `WITH 1 as buffer, 4096 as extent, mvtTileBBox({z:UInt8}, {x:UInt32}, {y:UInt32}, buffer / extent) as bounding_box`...

src/Functions/mvtEncodeGeom.cpp
```diff
@@ -0,0 +1,473 @@
...
+    constexpr UInt8 LineString = 0;
+    constexpr UInt8 MultiLineString = 1;
+    constexpr UInt8 MultiPolygon = 2;
+    constexpr UInt8 Point = 3;
+    constexpr UInt8 Polygon = 4;
+    constexpr UInt8 Ring = 5;
+}
+
+/// 0xFFFFFFFF: the Web Mercator projection spans the full UInt32 range (matching the materialized mercator_x/y columns).
+constexpr double mercator_max = 4294967295.0;
```
Algunenano: This value here is used incorrectly, it should be 2^32. Using `2^32 - 1` leads to a drift.

An example showcases it much better: The point (lon = 0, lat = 0) lies, by definition, exactly on the top-left corner of tile (2^(z-1), 2^(z-1)) at every zoom z ≥ 1. So mvtEncodeGeom must return pixel (0,0) for it at every zoom, independent of extent:

```
SELECT
    z,
    mvtEncodeGeom(CAST('(0.0, 0.0)', 'Point'), z, t, t, 4096, 0, false) AS pixel_no_clip,
    mvtEncodeGeom(CAST('(0.0, 0.0)', 'Point'), z, t, t) AS pixel_clipped
FROM
(
    SELECT
        CAST(arrayJoin([1, 10, 16, 20, 24, 28, 30, 32]), 'UInt8') AS z,
        CAST(bitShiftLeft(CAST('1', 'UInt64'), z - 1), 'UInt32') AS t
)

Query id: bd8ac62a-2fb2-4901-9233-17b6f7f57f60

   ┌──z─┬─pixel_no_clip─┬─pixel_clipped─┐
1. │  1 │ (0,0)         │ (0,0)         │
2. │ 10 │ (0,0)         │ (0,0)         │
3. │ 16 │ (0,0)         │ (0,0)         │
4. │ 20 │ (-1,-1)       │ (-1,-1)       │ <<<<<< Wrong already
5. │ 24 │ (-8,-8)       │ (-8,-8)       │ <<<<<< Wrong 
6. │ 28 │ (-128,-128)   │ (-128,-128)   │
7. │ 30 │ (-512,-512)   │ ᴺᵁᴸᴸ          │ <<<<< It has gone even beyond the extent, so more wrong
8. │ 32 │ (-2048,-2048) │ ᴺᵁᴸᴸ          │
   └────┴───────────────┴───────────────┘

8 rows in set. Elapsed: 0.002 sec. 
```


The world is scaled to a range that is 1 unit too short, so every projected coordinate is short by f mercator units (f = fraction of the world, here 0.5). In pixels that is f * extent / 2^(32-zoom): invisible below zoom ~17, but up to a full tile of displacement at zoom 32 near the antimeridian.

src/Functions/mvtTileBBox.cpp
```diff
@@ -0,0 +1,261 @@
...
+namespace ErrorCodes
+{
+    extern const int ARGUMENT_OUT_OF_BOUND;
+}
+
+namespace
+{
+
+/// 0xFFFFFFFF: the Web Mercator projection used by `mvtEncodeGeom` spans the full UInt32 range.
+constexpr double mercator_max = 4294967295.0;
```
Algunenano: Same bug with this value here too.

Query that shows it:

```
SELECT
    z,
    mvtTileBBoxMercator(z, t, t) AS last_tile_bbox,
    (last_tile_bbox.3) - (last_tile_bbox.1) AS width,
    bitShiftLeft(CAST('1', 'UInt64'), 32 - z) AS expected_width
FROM
(
    SELECT
        CAST(arrayJoin([1, 16, 24, 30, 31, 32]), 'UInt8') AS z,
        CAST((bitShiftLeft(CAST('1', 'UInt64'), z) - 1), 'UInt32') AS t
)

Query id: 06fac381-e2e2-4e24-b57d-28de83456483

   ┌──z─┬─last_tile_bbox─────────┬──────width─┬─expected_width─┐
1. │  1 │ {                     ↴│ 2147483647 │     2147483648 │ -- 2.15 billion
   │    │↳  "min_x": 2147483648,↴│            │                │ -- 2.15 billion
   │    │↳  "min_y": 2147483648,↴│            │                │ -- 2.15 billion
   │    │↳  "max_x": 4294967295,↴│            │                │ -- 2.15 billion
   │    │↳  "max_y": 4294967295 ↴│            │                │ -- 2.15 billion
   │    │↳}                      │            │                │ -- 2.15 billion
2. │ 16 │ {                     ↴│      65535 │          65536 │
   │    │↳  "min_x": 4294901760,↴│            │                │
   │    │↳  "min_y": 4294901760,↴│            │                │
   │    │↳  "max_x": 4294967295,↴│            │                │
   │    │↳  "max_y": 4294967295 ↴│            │                │
   │    │↳}                      │            │                │
3. │ 24 │ {                     ↴│        255 │            256 │
   │    │↳  "min_x": 4294967040,↴│            │                │
   │    │↳  "min_y": 4294967040,↴│            │                │
   │    │↳  "max_x": 4294967295,↴│            │                │
   │    │↳  "max_y": 4294967295 ↴│            │                │
   │    │↳}                      │            │                │
4. │ 30 │ {                     ↴│          3 │              4 │
   │    │↳  "min_x": 4294967292,↴│            │                │
   │    │↳  "min_y": 4294967292,↴│            │                │
   │    │↳  "max_x": 4294967295,↴│            │                │
   │    │↳  "max_y": 4294967295 ↴│            │                │
   │    │↳}                      │            │                │
5. │ 31 │ {                     ↴│          1 │              2 │
   │    │↳  "min_x": 4294967294,↴│            │                │
   │    │↳  "min_y": 4294967294,↴│            │                │
   │    │↳  "max_x": 4294967295,↴│            │                │
   │    │↳  "max_y": 4294967295 ↴│            │                │
   │    │↳}                      │            │                │
6. │ 32 │ {                     ↴│          0 │              1 │
   │    │↳  "min_x": 4294967295,↴│            │                │
   │    │↳  "min_y": 4294967295,↴│            │                │
   │    │↳  "max_x": 4294967295,↴│            │                │
   │    │↳  "max_y": 4294967295 ↴│            │                │
   │    │↳}                      │            │                │
   └────┴────────────────────────┴────────────┴────────────────┘

6 rows in set. Elapsed: 0.002 sec. 
```


That single constant is used in 3 places that produce the wrong numbers above: the clamp bound for the box edges (executeImpl, the std::clamp(..., 0.0, mercator_max) calls around lines 153-156), and the divisors in mercatorXToLongitude / mercatorYToLatitude (lines 36-46). With 2^32 the clamp no longer truncates the last tile (width becomes exactly expected_width, max_x = 4294967296), and min_lon of the lon-0 tile edge becomes exactly

src/AggregateFunctions/AggregateFunctionMVTEncode.cpp
```diff
@@ -0,0 +1,714 @@
...
+    }
+
+    /// Render one property value as the body of an MVT `Value` message into `out`.
+    void renderValue(const MvtProperty & property, const IColumn & column, size_t row, String & out) const
+    {
+        switch (property.kind)
+        {
+            case MvtValueKind::String:
+            {
+                const std::string_view value = column.getDataAt(row);
```
Algunenano, line 180: For Fixedstring this is including the final NULL character, which is incorrect.

Verification query:
```
SELECT mvtEncode('t')((0., 0.)::Point::Geometry, tuple(toFixedString('ab', 4))::Tuple(c1 FixedString(4)))
     = mvtEncode('t')((0., 0.)::Point::Geometry, tuple('ab')::Tuple(c1 String));
```

Should be the same, but isn't.

src/Functions/MVTEncodeGeom.cpp
```diff
@@ -0,0 +1,473 @@
...
+            if (result.empty())
+            {
+                builder.addNull();
+                return;
+            }
+            roundToGrid(result);
+            builder.addMultiLineString(result);
+        };
+
+        auto process_polygons = [&](MultiPolygon<BPoint> polygons, const Projection & projection, const BBox & box, bool clip)
```
Algunenano, line 306: As I feared this is very limited and will produce polygons that do not respect the MVT spec. It  mishandles invalid (self-intersecting) polygons, or even polygons that were valid but become invalid after projection or clipping.

The polygon path runs boost::geometry::intersection after boost::geometry::correct. bg::correct only fixes ring orientation and closure; it does not repair self-intersections, and bg::intersection's precondition is valid input. Self-intersecting rings therefore produce undefined results: the feature can be silently lost, mangled, or inverted into a full-tile polygon. Self-intersecting rings are common in real data (the examples below are from the PostGIS regression suite for ST_AsMVTGeom, which hits exactly these cases).

* Bowtie fully inside the tile: the feature is silently lost with clipping, and passed through invalid without it

```
SELECT mvtEncodeGeom([[(10., 10.), (50., 50.), (10., 50.), (50., 10.), (10., 10.)]]::Polygon, 2, 2, 1) AS clip_on,
       mvtEncodeGeom([[(10., 10.), (50., 50.), (10., 50.), (50., 10.), (10., 10.)]]::Polygon, 2, 2, 1, 4096, 256, 0) AS clip_off;
```

* clip_on:  NULL
* clip_off: [[[(455,3639),(2276,1461),(455,1461),(2276,3639),(455,3639)]]]

clip_on should be the two triangles of the bowtie (PostGIS returns both); instead the whole feature disappears. clip_off emits the self-intersection unrepaired, which mvtEncode then encodes into a tile that violates the MVT 2.1 polygon rules.

* Bowtie crossing the tile: half the geometry is replaced by a wrong quadrilateral

```
SELECT mvtEncodeGeom(CAST('[[(-10., -5.), (100., 60.), (-10., 60.), (100., -5.), (-10., -5.)]]', 'Polygon'), 2, 2, 1, 4096, 0)

Query id: 0e9aa6b5-db02-41d1-8adb-1ff9f47b04c9

   ┌─mvtEncodeGeom(CAST('[[(-10.⋯olygon'), 2, 2, 1, 4096, 0)─┐
1. │ [[[(4096,662),(0,662),(0,995),(4096,3991),(4096,662)]]] │
   └─────────────────────────────────────────────────────────┘
```

Expected: the two clipped lobes of the bowtie (PostGIS regress [PG43](https://github.com/postgis/postgis/blob/24e75f75b94ce51ce88fad7a4b9fadae12ae3ef3/regress/core/mvt.sql#L272)) -> https://github.com/postgis/postgis/blob/24e75f75b94ce51ce88fad7a4b9fadae12ae3ef3/regress/core/mvt_expected#L52

* Worst case: a 4-vertex self-crossing sliver inverts into a polygon covering the entire tile

This shape is PostGIS regress [PG45](https://github.com/postgis/postgis/blob/24e75f75b94ce51ce88fad7a4b9fadae12ae3ef3/regress/core/mvt.sql#L291) ("invalid geometry after simplification with invalid clipping", a real-world shape), translated to lon/lat:

```
SELECT mvtEncodeGeom(CAST('[[(15.403, 0.667), (15.335, 0.280), (15.398, -0.186), (15.331, 0.374), (15.403, 0.667)]]', 'Polygon'), 4, 8, 8, 4096, 16)

Query id: 53665850-9931-44fe-8e3d-5868d9d95435

   ┌─mvtEncodeGeom(CAST('[[(15.403, 0.667), (15.335⋯03, 0.667)]]', 'Polygon'), 4, 8, 8, 4096, 16)─┐
1. │ [[[(2797,-16),(2803,34),(2796,-16),(-16,-16),(-16,4112),(4112,4112),(4112,-16),(2797,-16)]]] │
   └──────────────────────────────────────────────────────────────────────────────────────────────┘
```

The output ring contains all four corners of the buffered clip window (-16,-16)..(4112,4112): a sub-pixel sliver in the source data renders as the whole tile filled. PostGIS returns NULL here.

Nested interior rings (a valid-looking but OGC-invalid POLYGON whose holes contain rings) are similarly mis-encoded (PostGIS regress [PG17](https://github.com/postgis/postgis/blob/master/regress/core/mvt.sql#L109): wagyu resolves them even-odd, here they all become holes of the exterior).

Recommendation: do what PostGIS does in mvt_clip_and_validate and clip/validate polygons with mapbox/wagyu (https://github.com/mapbox/wagyu) instead of bg::intersection. Wagyu was written for exactly this job: it clips to a box and outputs polygons that are valid under the MVT winding rules, handling self-intersections and nested rings by construction. It would be a new (small) submodule and it's header only, which we can include just for this file.

It operates on integer coordinates, so the natural integration point is after rounding to the pixel grid, which would also align the clip-vs-snap ordering with PostGIS. Lines and points can stay on bg::intersection.

src/Functions/MVTEncodeGeom.cpp
```diff
@@ -0,0 +1,473 @@
...
+            projection = Projection{static_cast<double>(tile_x) * tile_size, static_cast<double>(tile_y) * tile_size, static_cast<double>(extent) / tile_size};
+            const double lo = -static_cast<double>(buffer);
+            const double hi = static_cast<double>(extent) + static_cast<double>(buffer);
+            box = BBox(BPoint(lo, lo), BPoint(hi, hi));
+        };
+
+        /// Process one geometry of a concrete Boost type into the result builder.
+        auto process_point = [&](BPoint p, const Projection & projection, const BBox & box, bool clip)
+        {
+            projection(p);
```
Algunenano, line 276: PostGIS snaps to the integer pixel grid first and clips second; this PR clips first (covered_by/bg::intersection on the raw projection) and rounds after. Geometry within half a pixel of the clip boundary behaves differently.

A point 0.2 px outside the tile rounds exactly onto the boundary, yet is dropped:

```
SELECT
    mvtEncodeGeom((90.005, 30.), 2, 2, 1, 4096, 0) AS clip_on,
    mvtEncodeGeom((90.005, 30.), 2, 2, 1, 4096, 0, 0) AS clip_off

Query id: 34690690-85d3-415d-b8e6-056b004a3b99

   ┌─clip_on─┬─clip_off────┐
1. │ ᴺᵁᴸᴸ    │ (4096,2664) │
   └─────────┴─────────────┘
```

Expected (PostGIS): (4096, 2664) with clipping too.

Same for lines: A line wiggling sub-pixel around a tile edge gets fragmented into disjoint pieces with a gap, instead of one line along the edge:


```

SELECT
    mvtEncodeGeom(CAST('[(0.005, 10.), (-0.005, 20.), (-0.006, 30.), (0.004, 40.), (0.005, 50.)]', 'LineString'), 2, 2, 1, 4096, 0) AS clip_on,
    mvtEncodeGeom(CAST('[(0.005, 10.), (-0.005, 20.), (-0.006, 30.), (0.004, 40.), (0.005, 50.)]', 'LineString'), 2, 2, 1, 4096, 0, 0) AS clip_off

Query id: d322e512-1902-4e0c-9893-a31ddaf4f2ac

   ┌─clip_on────────────────────────────────────────────┬─clip_off─────────────────────────────────────────┐
1. │ [[(0,3639),(0,3403)],[(0,2329),(0,2107),(0,1461)]] │ [[(0,3639),(0,3167),(0,2664),(0,2107),(0,1461)]] │
   └────────────────────────────────────────────────────┴──────────────────────────────────────────────────┘
```

Every vertex rounds to x = 0, so the expected clipped result is the single line. Rendered, the fragments show up as dashed tile borders.

I'd suggest to round to the pixel grid before the covered_by/bg::intersection step instead of after.

I'll include a specific comment about polygons since that's a more complex case.

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,303 @@
...
+
+For point data this works just as well against plain `longitude`/`latitude` columns by building the point inline with
+`mvtEncodeGeom((lon, lat)::Point, …)`. To cluster coincident features, or to add an index-using bounding-box prefilter
+for large tables, extend the inner query as shown in [Clustering](#clustering) and
+[Restricting rows to a tile](#restricting-rows-to-a-tile).
+
+## Limitations {#limitations}
+
+- The Web Mercator projection clamps latitude to `±85.05112878°` and does not handle antimeridian-crossing inputs.
+- No explicit feature `id` field.
```
Algunenano: This is still important, some MVT renderers use ids for feature-state, stitching, etc.

src/Functions/MVTEncodeGeom.cpp
```diff
@@ -0,0 +1,473 @@
...
+            builder.addPoint(p);
+        };
+
+        auto process_lines = [&](MultiLineString<BPoint> lines, const Projection & projection, const BBox & box, bool clip)
+        {
+            bg::for_each_point(lines, projection);
+            MultiLineString<BPoint> result;
+            if (clip)
+                bg::intersection(lines, box, result);
+            else
```
Algunenano, line 293: This does not reject empty LineStrings because `dispatch_line/dispatch_ring/dispatch_polygon` wrap the input in Multi containers. The subsequent guard tests result.empty(), which is false for a container holding empty elements.

```
SELECT mvtEncodeGeom([]::LineString, 2, 2, 1, 4096, 0, 0);   -- today [[]],   expect NULL
SELECT mvtEncodeGeom([]::Polygon, 2, 2, 1, 4096, 0, 0);      -- today [[[]]], expect NULL
SELECT mvtEncodeGeom([[]]::MultiPolygon, 2, 2, 1, 4096, 0, 0); -- today [[[]]], expect NULL
```

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,303 @@
...
+
+**Parameters**
+
+- `layer_name` — Name of the vector tile layer. [`String`](../../data-types/string.md).
+- `extent` — Tile extent in pixels per side, in the range `[1, 2147483647]`. Defaults to `4096`. [`UInt32`](../../data-types/int-uint.md).
+
+**Arguments**
+
+- `geometry` — Tile-space geometry, for example from `mvtEncodeGeom`. [`Geometry`](../../data-types/geo.md).
+- `properties` — Optional named tuple of feature attributes. Element names become attribute keys. [`Tuple`](../../data-types/tuple.md).
```
Algunenano, line 112: In the future maybe we could support JSON, but for now I think it's ok

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
```diff
@@ -0,0 +1,714 @@
...
+        if (which.isFloat64())
+            return MvtValueKind::Double;
+        if (isBool(base))
+            return MvtValueKind::Bool;
+        if (which.isNativeInt() || which.isDate32())
+            return MvtValueKind::Sint;
+        if (which.isNativeUInt() || which.isDate() || which.isDateTime())
+            return MvtValueKind::UInt;
+
+        throw Exception(
```
Algunenano: One design decision we could take, which is what Postgis does, is to stringify any types that aren't directly supported, so any type is naturally supported. Maybe we could make this an option, that way big integers would be supported for example.

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
```diff
@@ -0,0 +1,714 @@
...
+    VectorWithMemoryTracking<MvtProperty> properties;
+
+    static MvtValueKind kindForType(const DataTypePtr & element_type, const String & function_name)
+    {
+        const DataTypePtr base = removeNullable(recursiveRemoveLowCardinality(element_type));
+        WhichDataType which(base);
+
+        if (which.isStringOrFixedString())
+            return MvtValueKind::String;
+        if (which.isFloat32())
```
Algunenano: We can support BFloat16 by using Float32, since it's a superset

src/Functions/mvtEncodeGeom.cpp
```diff
@@ -0,0 +1,473 @@
...
+    }
+};
+
+/// Snap every coordinate of a (tile-space) geometry to the integer pixel grid, matching the MVT geometry model and
+/// letting `GROUP BY` collapse sub-pixel duplicates into a single cluster.
+/// Rounds to the nearest integer and normalises negative zero to positive zero so coordinates clipped to an axis
+/// do not surface as "-0".
+inline Float64 roundCoordinate(Float64 value)
+{
+    const Float64 rounded = std::round(value);
```
Algunenano: Minor, just mentioning for completeness: std::round/llround (ties away from zero) while PostGIS does rint (ties to even). Here + toTileCoordinate

docs/en/sql-reference/functions/geo/mvt.md
```diff
@@ -0,0 +1,303 @@
...
+
+```text
+┌─bbox────────────────────────────────────────────┐
+│ (-180,-85.05112877980659,180,85.05112877980659)  │
+└──────────────────────────────────────────────────┘
+```
+
+## mvtTileBBoxMercator {#mvttilebboxmercator}
+
+The Web Mercator counterpart of `mvtTileBBox`, also available under its PostGIS name `ST_TileEnvelope`. Returns the
```
Algunenano: After checking the usage in Postgis and here I see I was wrong suggesting this alias, since in Postgis this is used to prefilter (based on the BBOX) and to give the extent to ST_AsMVTGeom. OTOH here ST_AsMVTGeom is covered automatically and we use `mvtTileBBox` to do the filter. mvtTileBBoxMercator would be to filter on the tile/MVT coordinates, which is unlikely to happen. Better to drop it to avoid confusion and provide good examples as to how to do filtering in ClickHouse (done under `Restricting rows to a tile`).

src/AggregateFunctions/AggregateFunctionMVTEncode.cpp
```diff
@@ -0,0 +1,714 @@
...
+};
+
+/// Variable-length aggregate state held in the arena: a flat buffer of self-contained feature records.
+/// Each record is `[UInt8 geometry_type][varint geometry_length][geometry command bytes]` followed, for every
+/// property column in tuple order, by a presence byte and (when present) a varint-length-prefixed pre-rendered MVT
+/// `Value` message. Records are self-contained and order-independent, so `merge` is a plain buffer concatenation.
+/// Value interning into the layer's shared value pool, and the final protobuf assembly, are deferred to
+/// `insertResultInto`, which is where MVT property sharing happens.
+struct AggregateFunctionMvtEncodeData
+{
```
Algunenano, line 109: AFAICT this is safe and covers all possible issues with malformed data which is great.

One issue that I see is that the intermediate steps will grow per row noticeable since we don't deduplicate until the end. This is ok, but with aggregation state-s it also means we won't be able to easily change it in the future.

I would suggest to add a version here, initially hardcoded to 0 and using the first byte for it, so we can easily evolve it in the future if we want to optimize it further

Algunenano (changes requested):
Sorry it took me a while to review and understand differences vs PostGIS (which I consider the reference). Most of the issues flagged are not critical, and if they are too complex we could choose to document the different behaviour and address them in the future as they won't change the API. The only exception is the version in the aggregation function state.

src/AggregateFunctions/AggregateFunctionMVTEncode.cpp
saarthak2002: Done - serialize now writes a state-format version byte and deserialize validates it for future evolution without  breaking

saarthak2002:
Hello @Algunenano,

Thank you for your detailed feedback. Sorry for the delay, it took me some time to review everything. I have fixed the one critical one you pointed out related to the versioning. As you mentioned, some of them are quite elaborate improvements like using wagyu instead of bg::intersection for polygons and snapping to the grid before clipping - would you like me to open a new PR with these changes documented after this one is merged ? Documenting these as known limitations that can be worked on without changing the API. Thank you.

I will continue to address some of the more minor things you pointed out above here.

docs/en/sql-reference/functions/geo/mvt.md
saarthak2002: Makes sense- changed the default buffer to 1 in mvtEncodeGeom and docs

docs/en/sql-reference/functions/geo/mvt.md
saarthak2002: Done!

docs/en/sql-reference/functions/geo/mvt.md
saarthak2002: dropped the ST_TileEnvelope alias

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
saarthak2002: BFloat16 is now accepted

src/Functions/mvtEncodeGeom.cpp
saarthak2002: changed- `roundCoordinate` and `toTileCoordinate` use `rint`/`llrint`

Algunenano:
@saarthak2002 Thank you. I'll have a look soon, ideally this week. Let's document the existing issues around polygons so we can implement and improve them in the future if the need is important (for perf / correctness).

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
saarthak2002: added stringify_unsupported flag (optional 4th positional parameter). When set, property types not directly supported (big integers, etc) are encoded as their text string value

src/AggregateFunctions/AggregateFunctionMVTEncode.cpp
saarthak2002: Fixed and added a test

src/Functions/MVTEncodeGeom.cpp
saarthak2002: Fixed — empty LineString/Polygon/MultiPolygon now return NULL

docs/en/sql-reference/functions/geo/mvt.md
saarthak2002: Added an optional feature_id_name parameter (PostGIS equivalent) - existing calls are unchanged (optional opt-in)

src/Functions/mvtEncodeGeom.cpp
saarthak2002: Fixed — mercator_max now uses `2^32` instead of `2^32 - 1`

src/Functions/mvtTileBBox.cpp
saarthak2002: Fixed - mercator_max now uses `2^32` instead of `2^32 - 1`

docs/en/sql-reference/functions/geo/mvt.md
Algunenano: This is still an issue. The convention is to capitalize acronyms. Like JSONextract

src/AggregateFunctions/AggregateFunctionMvtEncode.cpp
```diff
@@ -423,9 +453,65 @@ class AggregateFunctionMvtEncode final
...
+                }
+                if (!id_index)
+                    throw Exception(
+                        ErrorCodes::BAD_ARGUMENTS,
+                        "The feature_id_name '{}' of aggregate function {} is not an element of the properties tuple",
+                        feature_id_name_, getName());
+
+                const DataTypePtr id_base = removeNullable(recursiveRemoveLowCardinality(elements[*id_index]));
+                const WhichDataType id_which(id_base);
+                if (!id_which.isNativeInt() && !id_which.isNativeUInt())
```
Algunenano: Let's only accept unsigned integers and reject signed ones since they can't be represented.  It's easier to validate and avoids the situation where some features have ids and some don't because you used a column of the wrong type

tests/queries/0_stateless/04299_mvt_encode.sql
```diff
@@ -55,6 +55,25 @@ SELECT '-- mvtEncode: BFloat16 properties are encoded as float values (same as F
 SELECT mvtEncode('t')((0.0, 0.0)::Point::Geometry, tuple(toBFloat16(1.5))::Tuple(c BFloat16))
      = mvtEncode('t')((0.0, 0.0)::Point::Geometry, tuple(toFloat32(1.5))::Tuple(c Float32));
 
+SELECT '-- mvtEncode: a FixedString property trims its trailing NUL padding (matches the String value)';
+SELECT mvtEncode('t')((0.0, 0.0)::Point::Geometry, tuple(toFixedString('ab', 4))::Tuple(c1 FixedString(4)))
+     = mvtEncode('t')((0.0, 0.0)::Point::Geometry, tuple('ab')::Tuple(c1 String));
+
+SELECT '-- mvtEncode: feature_id_name emits the named integer element as the Feature id and drops it from the tags';
+SELECT mvtEncode('t', 4096, 'fid')((0.0, 0.0)::Point::Geometry, tuple(toUInt64(0), 'x')::Tuple(fid UInt64, name String))
```
Algunenano: We need to validate the outputs are correct and match what's expected, not that they are different from not having id declared. Please generate the output, validate it with a ref, and compare it here

tests/queries/0_stateless/04299_mvt_encode.sql
```diff
@@ -55,6 +55,25 @@ SELECT '-- mvtEncode: BFloat16 properties are encoded as float values (same as F
...
+    != mvtEncode('t')((0.0, 0.0)::Point::Geometry, tuple('x')::Tuple(name String));
+SELECT mvtEncode('t', 4096, 'fid')((0.0, 0.0)::Point::Geometry, tuple(CAST(NULL, 'Nullable(UInt64)'), 'x')::Tuple(fid Nullable(UInt64), name String))
+     = mvtEncode('t')((0.0, 0.0)::Point::Geometry, tuple('x')::Tuple(name String));
+
+SELECT '-- mvtEncode: feature_id_name must name an existing integer element';
+SELECT mvtEncode('t', 4096, 'missing')((0.0, 0.0)::Point::Geometry, tuple(toUInt64(1))::Tuple(fid UInt64)); -- { serverError BAD_ARGUMENTS }
+SELECT mvtEncode('t', 4096, 'fid')((0.0, 0.0)::Point::Geometry, tuple('s')::Tuple(fid String)); -- { serverError ILLEGAL_TYPE_OF_ARGUMENT }
+
+SELECT '-- mvtEncode: stringify_unsupported encodes otherwise-unsupported types (e.g. Int128) as their text value';
+SELECT mvtEncode('t', 4096, '', 1)((0.0, 0.0)::Point::Geometry, tuple(toInt128('170141183460469231731687303715884105727'))::Tuple(big Int128))
```
Algunenano: Same here

Algunenano (changes requested):
Review:

Function naming:
* ClickHouse keeps uppercase for acronyms in function names (JSONEncode, URLHash, MACNumToString). Let's call functions MVTEncode and so on.
* Also, unless the name is clear or known we also avoid truncating words. Geom is understandable, Bbox isn't unless you already know. mvtTileBBox -> MVTBoundingBox (Tile is the T), mvtTileBBoxMercator -> MVTBoundingBoxMercator

Docs:
* Please document the limitations of the current approach to clipping and validating the geometries in the docs.

Other:
* Let's accept just unsigned integers of the id.
* Tests should not simply compare feature enabled != feature disabled. They should verify the output is correct according to the spec. Please generate them, verify them and put them in the reference file.

saarthak2002:
Changes as per review:
- Functions have been renamed to follow conventions (MVTEncode, MVTEncodeGeom, MVTBoundingBox, MVTBoundingBoxMercator)
- Feature id rejects signed integers
- Tests have been updated to validate correct output for stringify_unsupported and feature_id_name for MVTEncode
- Docs now describe the limitations about clipping and polygons

Algunenano (approved):
@saarthak2002 Thanks a lot. Looks great. I'll finish reviewing the (unrelated) CI failures and merge it

fm4v:
@groeneai enrich tests: check boundary values and equivalence classes. Check results with standard.

groeneai:
@fm4v on it. Since this PR is merged and lives on a fork, I'll deliver the enrichment as a separate follow-up PR against master (extending `04299_mvt_encode` / `04300_mvt_clip`, not weakening any existing assertion).

Plan, grounded in the actual code paths (`MVTEncodeGeom.cpp`, `AggregateFunctionMVTEncode.cpp`, `MVTProtobuf.h`):

**1. Check results against the standard.** The strongest "verify against spec" we can do without a new dependency is to assert the emitted bytes against the worked examples in vector-tile-spec 2.1 (§4.1 layer/feature structure, §4.3 the `CommandInteger`/`ParameterInteger` geometry encoding, §4.4 the Value variants). Concretely:
- Decode the produced tile back to a human-readable form in SQL and put the decoded command stream in the `.reference` (geometry type, MoveTo/LineTo/ClosePath ids + counts, zig-zag-decoded deltas, tags, value pool), so the reference shows *what the spec says the bytes mean*, not just an opaque hex blob.
- Pin the spec's own example feature (the §4.3.5.1 single MoveTo point and the §4.3.5.4 polygon-with-2-rings command sequence `[9 3 6 ... 15]`) and assert our encoder produces exactly that command stream.

**2. Boundary values.** zoom 0 and 32; tile indices `0` and `2^zoom - 1` (already partly covered); extent `1` and `2147483647`; buffer `0`; coordinates exactly on the tile edge, one pixel inside, one pixel outside, and spanning the buffer; the `Int32`-delta and `Int32`-coordinate limits in `MVTEncode`; ties-to-even rounding at `x.5`; latitude exactly `±85.05112878`; longitude wrap at `±180`.

**3. Equivalence classes.** geometry kind (Point / Line / MultiLine / Ring / Polygon-with-hole / MultiPolygon); clip outcome (fully inside / fully outside -> NULL / partially clipped / on boundary); degenerate (collinear ring, sub-pixel line, zero-area near `Int32::max`); each `MVTValueKind` -> its Value field number (string/float/double/sint/uint/bool, FixedString trimming, stringify fallback, the `Nullable`/`LowCardinality` wrappers); feature-id present / NULL / rejected types; value interning (shared vs distinct); `-State`/`-Merge` round-trip.

I'll build and run locally, then open the follow-up PR and link it here. Anything you want weighted differently (e.g. a specific reference decoder you'd prefer over an in-SQL decode)?

groeneai:
Follow-up PR with the enriched tests: https://github.com/ClickHouse/ClickHouse/pull/108929 — decodes MVTEncode output through the official vector_tile.proto and asserts the §4.3.5 worked examples byte-for-byte, plus equivalence classes and boundary values.

clickgapai:
Hi @saarthak2002 @Algunenano — while reviewing this PR I found the following:

- **Coverage gap:** `MVTEncode` aggregate state `serialize`/`deserialize` round-trip is untested — https://github.com/ClickHouse/ClickHouse/pull/109601

PR #106108, opened 2026-05-29
Update NuRaft Keeper settings

antonio2368:
Update `NuRaft` to commit `62ab4edc621a65432badf60d7f85fa8695debeff`.

This brings the upstream leader-side admission limit for uncommitted log entries and the append-entries backward-probe throttle, then exposes both through Keeper `coordination_settings` so they can be configured in ClickHouse.

- Improvement

Added Keeper `coordination_settings` for `NuRaft` uncommitted log entry admission limiting and append-entries backward-probe throttling.

- [x] Documentation is not required because this change exposes internal Keeper tuning settings.

### Version info
- Merged into: `26.6.1.241`
- Backported to: `26.5.2.23`


al13n321:
This breaks the new dispatcher a little, fix in https://github.com/ClickHouse/NuRaft/pull/115

PR #106102, opened 2026-05-29
catalogs with schema evolutions + concurrent checks

scanhex12:
### Changelog category (leave one):
- Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
catalogs with schema evolutions + concurrent checks

### Version info
- Merged into: `26.6.1.431`

PR #106111, opened 2026-05-29
Fix mutation issues

scanhex12:

Fix mutation issues:
1. 1 snapshot instead of 2
2. Fix with catalogs

### Version info
- Merged into: `26.6.1.654`


src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -361,68 +366,69 @@ static bool writeMetadataFiles(
...
-    auto storage_metadata_name = path_resolver.resolve(metadata_info.path);
     Int64 parent_snapshot = -1;
     if (metadata->has(Iceberg::f_current_snapshot_id))
         parent_snapshot = metadata->getValue<Int64>(Iceberg::f_current_snapshot_id);
 
-    Int64 total_rows = 0;
-    Int64 total_bytes = 0;
-    Int64 total_files = 0;
-    for (const auto & [_, delete_filename] : delete_filenames.delete_file)
+    auto sum_files = [](const DataFileWriteResultWithStats & set, Int64 & rows, Int64 & bytes, Int64 & files)
```
asya-ch: nit: Initializing ints by reference is not super standard. Could we perhaps return a tuple or a struct instead? For example, tuple could look like:
```
auto sum_files = [](const DataFileWriteResultWithStats & set) 
    -> std::tuple<Int64, Int64, Int64> {
    Int64 rows = 0, bytes = 0, files = 0;
    // ... accumulate ...
    return {rows, bytes, files};
};

// Usage:
auto [delete_rows, delete_bytes, delete_files_count] = sum_files(delete_files);
```

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -490,62 +494,54 @@ static bool writeMetadataFiles(
...
+            generateManifestList(
+                path_resolver,
+                metadata,
+                object_storage,
+                context,
+                manifest_entries,
+                new_snapshot,
+                manifest_entry_sizes,
+                *buffer_manifest_list,
+                /* content_type */ Iceberg::FileContentType::POSITION_DELETE,
```
asya-ch, line 506: Could make it more robust via assert that `per_entry_content_types` is not empty. I was confused why we are passing `POSITION_DELETE` here (until I looked at the next function).

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -348,7 +352,8 @@ static std::optional<WriteDataFilesResult> writeDataFiles(
 }
 
 static bool writeMetadataFiles(
-    DataFileWriteResultWithStats & delete_filenames,
+    DataFileWriteResultWithStats & delete_files,
+    std::optional<DataFileWriteResultWithStats> & data_files,
```
asya-ch: Let's make `delete_files` and `data_files` const ref.

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -435,52 +441,50 @@ static bool writeMetadataFiles(
         }
     };
 
-    try
+    auto write_manifest_entries = [&](DataFileWriteResultWithStats & set, Iceberg::FileContentType content_type, SharedHeader sample_block)
     {
-        for (const auto & [partition_key, delete_filename] : delete_filenames.delete_file)
+        for (const auto & [partition_key, data_file] : set.delete_file)
```
asya-ch: I find it confusing we use `delete_file` regardless of whether it's `delete_file` or `data_file`. 

Let's change the field name to something more generic (e.g. `files`, `files_by_partition`, etc.).

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -361,68 +366,69 @@ static bool writeMetadataFiles(
...
+
+    Int64 delete_rows = 0;
+    Int64 delete_bytes = 0;
+    Int64 delete_files_count = 0;
+    sum_files(delete_files, delete_rows, delete_bytes, delete_files_count);
+
+    Int64 data_rows = 0;
+    Int64 data_bytes = 0;
+    Int64 data_files_count = 0;
+    if (data_files)
```
asya-ch: nit: I think it would be cleaner to make `data_files` an empty map in the non-update case so we can avoid all these `if (data_files)` checks.

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -361,68 +366,69 @@ static bool writeMetadataFiles(
     Poco::JSON::Object::Ptr partititon_spec,
     Int32 partition_spec_id,
     std::optional<ChunkPartitioner> & chunk_partitioner,
-    Iceberg::FileContentType content_type,
-    SharedHeader sample_block,
-    bool write_metadata_json_file)
+    SharedHeader delete_sample_block,
```
asya-ch: Don't need `delete_sample_block` as a param -- can generate instead in the function by calling helper.

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -577,19 +573,60 @@ void mutate(
...
+                persistent_table_components.metadata_compression_method,
+                /* force_fetch_latest_metadata */ true,
+                /* ignore_explicit_metadata_file_path */ true);
+            last_version = last_version_info.version;
+            metadata_path = last_version_info.path;
+            compression_method = last_version_info.compression_method;
+        }
+        else
+        {
+            DataLake::TableMetadata table_metadata;
```
asya-ch: Nice catch.

I don't think this is the right place for the logic though. There are other places in the file where we call `getLatestOrExplicitMetadataFileAndVersion()` and don't consult the catalog for latest updates. 

Maybe this should be implemented in the same file as `getLatestOrExplicitMetadataFileAndVersion()`, and consistently called for all three call sites in this file.

src/Storages/ObjectStorage/DataLakes/Iceberg/MetadataGenerator.cpp
```diff
@@ -156,6 +156,11 @@ MetadataGenerator::NextMetadataResult MetadataGenerator::generateNextMetadata(
     else
     {
         summary->set(Iceberg::f_operation, Iceberg::f_overwrite);
+        if (added_files > 0 || added_records > 0)
```
asya-ch: Do we need the if condition here? Would be simpler without.

Also, can we move the lines that repeated across the `if/else` out. For example, changed_partition_count, added_files_size, added_files, added_records (if it's ok to record zero).

tests/integration/test_database_iceberg/test.py
```diff
@@ -1026,3 +1026,66 @@ def test_invalid_auth_header_format(started_cluster):
             """
         )
     assert "Invalid auth header format" in str(err.value)
+
+
+def test_writes_mutate_update(started_cluster):
```
asya-ch, line 1158: Thanks for grabbing the test!

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
```diff
@@ -494,38 +501,32 @@ void generateManifestList(
         };
         entry.field(Iceberg::f_added_snapshot_id) = new_snapshot->getValue<Int64>(Iceberg::f_metadata_snapshot_id);
         auto summary = new_snapshot->getObject(Iceberg::f_summary);
+
+        Int64 entry_added_rows = 0;
+        if (entry_content == Iceberg::FileContentType::DATA)
+        {
+            if (summary->has(Iceberg::f_added_records))
```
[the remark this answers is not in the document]
asya-ch: This is just a refactor, right? Maybe we can open an issue and fix in a separate PR? Once system.iceberg_files is merged, can use it to test.

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
```diff
@@ -494,38 +501,32 @@ void generateManifestList(
...
+                entry_added_rows = summary->getValue<Int64>(Iceberg::f_added_records);
+        }
+        else if (summary->has(Iceberg::f_added_position_deletes))
+            entry_added_rows = summary->getValue<Int64>(Iceberg::f_added_position_deletes);
```
asya-ch: Why use added_position_deletes for entry_added_rows?

scanhex12: https://iceberg.apache.org/spec/#manifest-lists

Number of entries in the manifest that have status ADDED

ADDED is not just about data files, it also includes position delete files

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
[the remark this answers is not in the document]
scanhex12: It's a fix for the `added_rows_count` field. Please let it be here without separated PRs

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -724,17 +690,17 @@ void alter(
     while (i < MAX_TRANSACTION_RETRIES)
     {
         auto log = getLogger("IcebergMutations");
-        auto [last_version, metadata_path, compression_method] = getLatestOrExplicitMetadataFileAndVersion(
+        auto [last_version, metadata_path, compression_method] = getLatestMetadataFileAndVersionWithCatalog(
```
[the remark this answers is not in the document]
scanhex12: I fixed it in this PR https://github.com/ClickHouse/ClickHouse/pull/106102

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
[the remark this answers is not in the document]
asya-ch: I just mean, in a separate PR, let's address the issue the bot is talking about -- that summary can encompass multiple manifest entries, so the values here are not correct. Not asking you to remove anything from this one.

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
```diff
@@ -494,38 +501,32 @@ void generateManifestList(
...
-                Iceberg::f_added_rows_count);
-        }
-        else
-        {
-            set_versioned_field(summary->getValue<Int64>(Iceberg::f_added_position_deletes), Iceberg::f_added_rows_count);
-        }
-        set_versioned_field(
-            0,
-            Iceberg::f_existing_rows_count);
+        set_versioned_field(entry_added_rows, Iceberg::f_added_rows_count);
```
asya-ch: I'm confused on a couple points:
(1) Why was `added_rows_count` implementation previously incorrect? It looks functionally equivalent to me, not sure if I'm missing something.
(2) Why are we no longer setting `added_position_deletes` and `deleted_rows_count`?

src/Storages/ObjectStorage/DataLakes/Iceberg/Mutations.cpp
```diff
@@ -361,67 +367,63 @@ static bool writeMetadataFiles(
...
-        new_snapshot = result.snapshot;
-        storage_manifest_list_name = path_resolver.resolve(result.manifest_list_path);
-    }
     auto manifest_entries_in_storage = std::make_shared<Strings>();
     std::vector<Iceberg::IcebergPathFromMetadata> manifest_entries;
     std::vector<Int64> manifest_entry_sizes;
+    std::vector<Iceberg::FileContentType> per_entry_content_types;
 
-    auto cleanup = [object_storage, &delete_filenames, &path_resolver, manifest_entries_in_storage, storage_manifest_list_name, storage_metadata_name]()
+    auto cleanup = [object_storage, &delete_files, &data_files, &path_resolver, manifest_entries_in_storage, storage_manifest_list_name]()
```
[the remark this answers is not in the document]
asya-ch: @scanhex12, let's address it please.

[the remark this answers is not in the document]
asya-ch: @scanhex12, let's fix this please. I know it's not introduced by your PR, but it's a good catch from the bot.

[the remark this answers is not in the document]
scanhex12: Yes, I fixed this

src/Storages/ObjectStorage/DataLakes/Iceberg/IcebergWrites.cpp
```diff
@@ -520,23 +527,27 @@ void generateManifestList(
         };
         entry.field(Iceberg::f_added_snapshot_id) = new_snapshot->getValue<Int64>(Iceberg::f_metadata_snapshot_id);
         auto summary = new_snapshot->getObject(Iceberg::f_summary);
+
         if (version == 1)
         {
             set_versioned_field(1, Iceberg::f_added_files_count);
             set_versioned_field(std::stoi(summary->getValue<String>(Iceberg::f_total_data_files)), Iceberg::f_existing_files_count);
             set_versioned_field(0, Iceberg::f_deleted_files_count);
             if (summary->has(Iceberg::f_added_position_deletes))
```
asya-ch, line 537: Can fix it in another PR if preferred, but there is a bunch of repetition across this `if/else` despite `set_versioned_field` handling both cases.

src/Storages/ObjectStorage/DataLakes/Iceberg/MetadataGenerator.cpp
```diff
@@ -145,22 +145,16 @@ MetadataGenerator::NextMetadataResult MetadataGenerator::generateNextMetadata(
 
     auto parent_snapshot = getParentSnapshot(parent_snapshot_id);
     Poco::JSON::Object::Ptr summary = new Poco::JSON::Object;
-    if (num_deleted_rows == 0)
+    summary->set(Iceberg::f_operation, num_deleted_rows == 0 ? Iceberg::f_append : Iceberg::f_overwrite);
+    summary->set(Iceberg::f_added_data_files, std::to_string(added_files));
+    summary->set(Iceberg::f_added_records, std::to_string(added_records));
```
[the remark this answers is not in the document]
asya-ch: @scanhex12 -- This looks correct to me, could we for now just move added_records back to how it was (one-line change)? LMK if I'm missing something.

In the future, we should probably invest in adding more tests for the statistics writer.

src/Storages/ObjectStorage/DataLakes/Iceberg/ExpireSnapshotsExecute.cpp
```diff
@@ -670,22 +671,27 @@ ExpireSnapshotsResult expireSnapshots(
     if (!common_path.starts_with('/'))
         common_path = "/" + common_path;
 
+    if (catalog && catalog->isTransactional())
```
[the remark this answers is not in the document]
scanhex12: such a weird case tbh

PR #106113, opened 2026-05-29
Self-deduplication for ActionsDAG

yariks5s:
- Improvement

Provides the logic to remove duplicate calculations during query execution in different scenarios.

### Version info
- Merged into: `26.6.1.982`


src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,205 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+        parent[&node] = &node;
+        size[&node] = 1;
+    }
+}
+
+const ActionsDAG::Node * ActionsDAG::EquivalenceClasses::find(const Node * n) const
+{
+    auto it = parent.find(n);
+    if (it == parent.end() || it->second == n)
+        return it == parent.end() ? n : it->second;
```
george-larionov: I believe this could just be `return n;` right?

tests/queries/0_stateless/04300_dedup_subtrees_on_filter_merge.sql
```diff
@@ -0,0 +1,68 @@
...
+    INNER JOIN (SELECT number FROM numbers(10000)) AS r
+    ON l.number = r.number AND l.number > 0 AND r.number > 0
+    SETTINGS enable_join_runtime_filters = 0
+);
+
+SELECT count() FROM (SELECT 1 AS x WHERE x = 1 AND 1 = x);
+
+SELECT count() FROM numbers(100) WHERE number > 0 AND number > 0;
+
+SELECT count() FROM numbers(100) WHERE rand() % 256 < 256 AND rand() % 256 < 256;
```
george-larionov: What are these 3 tests testing? It seems to me like they will work no matter the deduplication behavior

tests/queries/0_stateless/04300_dedup_subtrees_on_filter_merge.sql
```diff
@@ -0,0 +1,68 @@
...
+
+-- Two arrayMap with different lambdas but same signature must not collapse
+-- Outer filter uses `x + 2`; correct answer is `number = 3` (3 + 2 = 5)
+SELECT number FROM (
+    SELECT number, arrayMap(x -> x + 1, [number])[1] AS a FROM numbers(5)
+) WHERE arrayMap(x -> x + 2, [number])[1] = 5;
+
+-- If the outer rand() were merged with inner x, the predicate would collapse to
+-- x = x and return 100, correct behavior is two independent rands -> ~0 matches
+SELECT count() FROM (SELECT rand() AS x FROM numbers(100)) WHERE rand() = x;
```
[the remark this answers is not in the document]
george-larionov: I guess you could instead do something like `SELECT count() < 20` to make it even less likely.

george-larionov (approved):
It looks good to me overall (I saw you fixed the final (hopefully) issue that the bot found). However I think it is worth it to add a performance test to show some perf gains for this improvement.

src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,216 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+        return false;
+    if (a->size() != b->size())
+        return false;
+    if (hasDummyInside(a) || hasDummyInside(b))
+        return false;
+    Field fa;
+    Field fb;
+    a->get(0, fa);
+    b->get(0, fb);
+    return fa == fb;
```
KochetovNicolai: Please don't use field comparison. Use IColumn::compareAt. Check the types are the same before it. Also check the columns are the same (probably, compare typeid). Some columns can allow compareAt for different types (e.g., LowCardinality) - we can check these corner cases as well.

src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,216 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+}
+
+struct NodeKey
```
KochetovNicolai: It is really similar to `ActionsDAG::Node`.
Maybe at least store `ActionsDAG::Node*` inside, and maybe something else, if we want comparable stuff?

src/Interpreters/ActionsDAG.h
```diff
@@ -240,6 +242,26 @@ class ActionsDAG
...
+    {
+    public:
+        explicit EquivalenceClasses(const ActionsDAG & dag);
+
+        const Node * find(const Node * n) const;
+        void unite(const Node * a, const Node * b);
+
+    private:
+        mutable std::unordered_map<const Node *, const Node *> parent;
+        std::unordered_map<const Node *, size_t> size;
```
KochetovNicolai: I would probably propose to make an inner class `EquivalenceClass{size, representative}`

src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,216 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+        /// even if their current values match
+        if (node.type == ActionType::COLUMN && !node.is_deterministic_constant)
+            continue;
+        if (node.type == ActionType::FUNCTION && node.function_base)
+        {
+            /// `rand()` / `randConstant` may produce different values per call - merging changes semantics
+            if (!node.function_base->isDeterministic())
+                continue;
+        }
+        /// Lambda-typed values (like results of `FunctionCapture`) carry inner DAG that we don't see here
```
KochetovNicolai: btw we can compare inner dags as well (todo)

src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,216 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+        {
+            case ActionType::INPUT:
+            case ActionType::PLACEHOLDER:
+                /// skipped above
+                break;
+            case ActionType::COLUMN:
+                /// keep the name - column names are observable for name-sensitive parents
+                /// (`formatRowNoNewline` uses them as JSON keys), and downstream code
+                /// looks up outputs by name
+                key.name = node.result_name;
```
KochetovNicolai: I think we should not compare names.
Outputs can be handled separately: take the representative from the equivalent class and add an alias on top to match names (if needed).

src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,216 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+                key.column = node.column;
```
KochetovNicolai: Also, I think we should treat all the constant-folded nodes as equivalent. E.g. function `plus(1, 1)` will be equialent constant `2`

src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,216 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+                break;
+            case ActionType::ALIAS:
+                /// keep alias name in the key - downstream may look up by name
+                key.name = node.result_name;
+                if (!node.children.empty())
+                    key.children.push_back(ec.find(node.children.front()));
+                break;
+            case ActionType::FUNCTION:
+                if (node.function_base)
+                    key.name = node.function_base->getName();
```
KochetovNicolai: I was thinking about avoiding comparison by name if possible. Also, maybe at the same time solve the issue with case-sensitive functions. Probably we can put into IFunction[Base] a ptr to the FunctionFactory::FunctionCreator. I expect it to be a single creator object shared by all the function aliases as well.

src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,216 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+            case ActionType::COLUMN:
+                /// keep the name - column names are observable for name-sensitive parents
+                /// (`formatRowNoNewline` uses them as JSON keys), and downstream code
+                /// looks up outputs by name
+                key.name = node.result_name;
+                key.column = node.column;
+                break;
+            case ActionType::ALIAS:
+                /// keep alias name in the key - downstream may look up by name
+                key.name = node.result_name;
```
KochetovNicolai: Same. Alias names should be ignored and class of alias is always a class of child.
Outputs should be handled separately.

src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,216 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+                {
+                    for (const auto * c : node.children)
+                        key.children.push_back(ec.find(c));
+                }
+                break;
+            case ActionType::ARRAY_JOIN:
+                break;
+        }
+
+        auto it = seen.find(key);
```
KochetovNicolai: I would prefer just a DFS over a hashing approach.
The sketch:
* inputs are always different classes of equivalents
* define a class of equivalence for every child first
* only non-trivial cases are constants (actually, nodes with filled ColumnConst) and functions
* for constants, we can indeed keep a hash map (hash is a SipHash updated from IDataType and IColumn; equality uses IDataType->equals and IColumn->compareAt)
* for functions, probably a hash table : CreatorPtr -> list of candidates, where every candidate is a list of classes of the arguments. In this case, we can do a linear search. Another option - add a hash of the candidate to the key as well.

src/Interpreters/ActionsDAG.cpp
```diff
@@ -868,6 +868,216 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+            /// first-seen wins as root, so EXPLAIN names stay stable
+            ec.unite(it->second, &node);
+        else
+            seen.emplace(std::move(key), &node);
+    }
+
+    return ec;
+}
+
+void ActionsDAG::applyEquivalenceClasses(const EquivalenceClasses & ec)
```
KochetovNicolai: Some nodes should be removed in the end.
I think it's correct to remove every node that is not class-representative or output.

src/Interpreters/ActionsDAG.h
```diff
@@ -240,6 +242,26 @@ class ActionsDAG
 
     void removeAliasesForFilter(const std::string & filter_name);
 
+    /// Union-find over DAG nodes. Callers populate it from whatever notion of
+    /// equivalence they need (structural, equi-join, ...)
+    class EquivalenceClasses
```
KochetovNicolai: Not sure we need this in the interface.
Probably keep just `deduplicateSubtrees` (or, even, combine with removeUnusedActions)

tests/queries/0_stateless/03130_convert_outer_join_to_inner_join.reference
```diff
@@ -238,13 +238,11 @@ Positions: 4 0 1 2
...
             Prewhere filter
-            Prewhere filter column: and(notEquals(__table1.id, 0_UInt8), notEquals(__table1.id, 0_UInt8)) (removed)
+            Prewhere filter column: notEquals(__table1.id, 0_UInt8) (removed)
             Actions: INPUT : 0 -> id UInt64 : 0
                      COLUMN Const(UInt8) -> 0_UInt8 UInt8 : 1
-                     FUNCTION notEquals(id : 0, 0_UInt8 : 1) -> notEquals(__table1.id, 0_UInt8) UInt8 : 2
-                     FUNCTION notEquals(id : 0, 0_UInt8 :: 1) -> notEquals(__table1.id, 0_UInt8) UInt8 : 3
-                     FUNCTION and(notEquals(__table1.id, 0_UInt8) :: 3, notEquals(__table1.id, 0_UInt8) :: 2) -> and(notEquals(__table1.id, 0_UInt8), notEquals(__table1.id, 0_UInt8)) UInt8 : 1
-            Positions: 0 1
+                     FUNCTION notEquals(id : 0, 0_UInt8 :: 1) -> notEquals(__table1.id, 0_UInt8) UInt8 : 2
```
KochetovNicolai, line 244: Nice!

src/Functions/IFunction.h
```diff
@@ -478,9 +499,10 @@ class IFunctionOverloadResolver : public std::enable_shared_from_this<IFunctionO
 private:
 
     DataTypePtr getReturnTypeWithoutLowCardinality(const ColumnsWithTypeAndName & arguments) const;
-};
 
-using FunctionOverloadResolverPtr = std::shared_ptr<IFunctionOverloadResolver>;
+    /// mutable beacuse it's set after construction by FunctionFactory, resolvers themselves are otherwise immutable
```
KochetovNicolai, line 503: As an option, we can change buildImpl to return `std::shared_ptr<IFunctionBase>` (or, even better, std::unique_ptr<IFunctionBase>`), and move it into `FunctionBasePtr = std::shared_ptr<const IFunctionBase>` during `build`.

I found about 35 places by searching. So let's do it in post-refactoring if needed.

src/Interpreters/ActionsDAG.cpp
```diff
@@ -881,6 +881,264 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
 
+namespace
+{
+
+/// dummy columns (ColumnSet for IN, ColumnFunction for lambdas) don't have a Field-representable value
+bool hasDummyInside(const ColumnPtr & col)
+{
+    if (!col)
+        return false;
+    if (const auto * cc = typeid_cast<const ColumnConst *>(col.get()))
```
KochetovNicolai: Apparently, this is redundant after the recent refactoring. Now `column` in the Node is always ColumnConst. So you can work with ColumnConstPtr directly and simplify some casts/checks.

src/Interpreters/ActionsDAG.cpp
```diff
@@ -881,6 +881,264 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+{
+    const ActionsDAG::Node * sample;
+};
+
+struct ConstantKeyHash
+{
+    size_t operator()(const ConstantKey & k) const
+    {
+        SipHash h;
+        h.update(k.sample->result_type->getName());
```
KochetovNicolai: Why not use IDataType::updateHash ?

src/Interpreters/ActionsDAG.cpp
```diff
@@ -881,6 +881,264 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+        {
+            if (!node.children.empty())
+                ec.unite(&node, node.children.front());
+            continue;
+        }
+
+        /// `rand()` / `randConstant` and similar give different values per call
+        if (!node.isDeterministic())
+            continue;
+        /// Lambda-typed values (like results of FunctionCapture) carry inner DAG we don't see
```
KochetovNicolai, line 1034: TODO: compare inner DAGs

src/Interpreters/ActionsDAG.cpp
```diff
@@ -881,6 +881,264 @@ void ActionsDAG::removeAliasesForFilter(const std::string & filter_name)
...
+        }
+
+        auto & candidates = functions_by_creator[factory_handle];
+
+        bool merged = false;
+        for (const auto & cand : candidates)
+        {
+            if (!cand.representative->result_type->equals(*node.result_type))
+                continue;
+            if (cand.arg_classes != arg_classes)
```
KochetovNicolai: I think there can be a case when the stored classes for function candidates are stale.

It should not be a problem if we traverse the nodes in topologically-sorted order.
Note that `nodes` are likely topologically sorted by the way we build it, but it is not an invariant.

KochetovNicolai (approved):

PR #90341, opened 2025-11-19
Improve `ClusterDiscovery` module .

RinChanNOWWW:
### Changelog category (leave one):

- Bug Fix

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):

- Fixed cluster discovery updates in multicluster setups with `multicluster_root_path` to prevent delays and missing ZooKeeper updates

### Documentation entry for user-facing changes

- [ ] Documentation is written (mandatory for new features)

PR #106114, opened 2026-05-29
Fix cluster discovery re-registration after Keeper session expiry

### LINKED ISSUE #106112: Cluster discovery nodes not re-registered after Keeper session expiry
### Describe what's wrong

After a Keeper session expiry (e.g., during Keeper rolling restart), ClickHouse data nodes do not re-register their ephemeral discovery znodes under `/clickhouse/discovery/<cluster>/shards`. Observer nodes see an incomplete cluster and silently return partial query results when `skip_unavailable_shards=1`.

The self-healing code exists in `upsertCluster()` (line 447 of `ClusterDiscovery.cpp`) but is unreachable because `Flags::wait()` blocks indefinitely after session expiry — watches are invalidated and `any_need_update` is never set.

Yes

### How to reproduce

Config on data nodes:
```xml
<clickhouse>
  <allow_experimental_cluster_discovery>1</allow_experimental_cluster_discovery>
  <remote_servers>
    <my_cluster>
      <discovery>
        <path>/clickhouse/discovery/my-cluster-disco</path>
        <shard from_env="SHARD"/>
        <secret from_env="SHARED_SECRET"/>
      </discovery>
    </my_cluster>
  </remote_servers>
</clickhouse>
```

Config on observer (query) nodes:
```xml
<clickhouse>
  <allow_experimental_cluster_discovery>1</allow_experimental_cluster_discovery>
  <remote_servers>
    <my_cluster>
      <discovery>
        <path>/clickhouse/discovery/my-cluster-disco</path>
        <observer/>
        <secret from_env="SHARED_SECRET"/>
      </discovery>
    </my_cluster>
  </remote_servers>
</clickhouse>
```

Steps:
1. Start cluster, verify all data nodes registered: `SELECT count() FROM system.zookeeper WHERE path = '/clickhouse/discovery/my-cluster-disco/shards'`
2. Rolling restart all Keeper nodes (fast enough that data node session timeouts are exceeded)
3. Wait 5+ minutes
4. Check registrations again — count will be 0 (or only nodes restarted after Keeper came back)
5. Restart an observer node — `SELECT * FROM system.clusters WHERE cluster = 'my_cluster'` shows only the re-registered nodes
6. Distributed queries return partial results silently

### Expected behavior

After Keeper session expiry and reconnection, data nodes should re-register their ephemeral discovery znodes automatically. The `force_update_interval` (2 min) and the self-check in `upsertCluster()` should catch the missing registration.

No errors in logs. The bug is silent — the main thread is blocked on `cv.wait()` and never wakes up to detect the missing registration.

### Additional context

Root cause in `src/Interpreters/ClusterDiscovery.cpp`:

`Flags::wait()` blocks indefinitely:
```cpp
cv.wait(lk, [this]() -> bool { return any_need_update || stop_flag; });
```

After session expiry:
1. Ephemeral znodes deleted by Keeper (correct)
2. ZK watch callbacks invalidated on old session (correct)
3. No watches fire on new session → `any_need_update` never set → `cv.wait()` blocks forever
4. `force_update_interval` check and `registerInZk` re-registration code exist but are inside the loop body that never executes

sharadgaur:
### Changelog category
- Bug Fix

### Changelog entry
Fix cluster discovery nodes not re-registering after Keeper session expiry, which caused distributed queries to silently return partial results.

### Description

After a Keeper session expiry (e.g. during Keeper rolling restart), data nodes do not re-register their ephemeral discovery znodes. The main loop in `ClusterDiscovery::runMainThread` blocks indefinitely on `Flags::wait()` because ZK watches are invalidated on the old session and `cv.wait()` is never notified.

The `force_update_interval` check and re-registration logic in `upsertCluster()` already exist but are unreachable because `wait()` never returns.

#### Regression

Introduced in b932cce8805 ("Fix code review", v25.12) which removed the timeout from `Flags::wait()`:

```
Before b932cce (v25.5 - v25.11):
  clusters_to_update->wait(5s, finished);   // cv.wait_for(5s) - loop runs every 5s

After b932cce (v25.12+):
  clusters_to_update->wait(finished);        // cv.wait() - blocks forever
```

The previous commit bfd7003f326 ("New server setting cluster_discovery_wait_timeout") had added a configurable timeout with `cv.wait_for()` and a fallback to `cv.wait()` when timeout=0. The follow-up code review commit `b932cce` removed the timeout parameter entirely, leaving only the blocking `cv.wait()`.

#### Timeline from production incident

- **Day 0**: Keeper pods rolling restarted (config change). All data node sessions expired. Ephemeral discovery znodes deleted. No data nodes re-registered.
- **Day 10**: One data node restarted due to infrastructure churn. Re-registered successfully. Now 1 out of 96 nodes in ZK discovery path.
- **Day 13**: Observer (query) node restarted. Got fresh state from ZK: only 1 shard visible. Distributed queries silently returned partial results. The other observer (not restarted) still had 42/48 shards cached in memory.

#### Fix

Restore the timeout on `Flags::wait()` using `cv.wait_for()`, passing `force_update_interval` (2 min) from `runMainThread`. This ensures the loop unblocks periodically regardless of watch state, allowing `upsertCluster()` to run the existing self-check:

```cpp
// Line 447 - already existed, just unreachable before this fix
if (!contains(node_uuids, current_node_name))
{
    LOG_ERROR(log, "Can't find current node, will register again");
    registerInZk(zk, cluster_info);
}
```

#### Changes

1. **`Flags::wait()`**: added optional timeout parameter (default=zero for backwards compatibility), uses `cv.wait_for()` when timeout > 0
2. **`runMainThread()`**: passes `force_update_interval` to `wait()`
3. **New integration test**: `test_cluster_discovery_keeper_restart` — stops all ZK nodes, waits for session expiry, restarts ZK, verifies all data nodes re-register

Closes #106112


sharadgaur:
### Local reproduction

Reproduced the bug locally using docker compose with ClickHouse 26.3 (3 keeper nodes, 4 data nodes):

```
=== Step 1: Verify discovery works ===
Nodes in cluster: 4
ZK registrations: 4

=== Step 2: Stop ALL keeper nodes ===
Waiting 30s for sessions to expire (session_timeout=10s)...

=== Step 3: Restart keeper nodes ===
Waiting 20s for quorum...

=== Step 4: Check ZK registrations ===
ZK registrations after keeper restart: 0

=== Step 5: Wait up to 3 min for re-registration ===
  10s: cluster_nodes=4  zk_registrations=0
  20s: cluster_nodes=4  zk_registrations=0
  ...
  180s: cluster_nodes=4  zk_registrations=0

=== Final Result ===
Final cluster nodes: 4 (expected 4)
Final ZK registrations: 0 (expected 4)
❌ FAILED: Bug reproduced
```

Key observations:
- After keeper restart, all 4 ephemeral discovery znodes are gone (expected)
- Data nodes reconnect to keeper but never re-register (the bug)
- `cluster_nodes=4` because the observer caches the old cluster state in memory - this masks the problem until the observer restarts
- Waited 3 minutes with no re-registration, confirming `cv.wait()` blocks forever

tanner-bruce:
I wonder if when the keeper conn is terminated and then reconnected it should also double check that the registration is still there. Otherwise there could be up to two minutes before the node is re-registered

tests/integration/test_cluster_discovery/test.py
```diff
@@ -177,6 +177,73 @@ def test_cluster_discovery_no_crash_on_empty_static_cluster(start_cluster):
     )
 
 
+def test_cluster_discovery_keeper_restart(start_cluster):
```
tiandiwonder, line 181: The regression test doesn't reliably validate re-registration.

  test_cluster_discovery_keeper_restart asserts the observer's system.clusters view, but that view is served from the in-memory
  cluster_impls cache (ClusterDiscovery::getClusters), not from Keeper. The cache was already populated with all 5 nodes by
  the pre-test check_on_cluster([observer], 5), and nothing clears it on session loss — it's only mutated on the success paths
  of upsertCluster/removeCluster.

tiandiwonder: To make it a reliable regression test, assert the actual children under /clickhouse/discovery/test_auto_cluster — e.g. via
  system.zookeeper or cluster.query_zookeeper — or restart the observer before checking system.clusters so it reads fresh state
  from Keeper.

src/Interpreters/ClusterDiscovery.cpp
```diff
@@ -132,10 +132,17 @@ class ClusterDiscovery::Flags
             flags.erase(key);
     }
 
-    std::unordered_map<T, bool> wait(bool & finished)
+    /// Wait for any flag to be set, or until timeout expires.
+    /// The timeout ensures the caller's periodic maintenance (e.g. force_update_interval
+    /// checks and ephemeral znode re-registration) runs even when ZK watches are lost
+    /// after a Keeper session expiry.
+    std::unordered_map<T, bool> wait(bool & finished, std::chrono::milliseconds timeout = std::chrono::milliseconds::zero())
```
tiandiwonder: Latent footgun (orthogonal to the 2-min-vs-5s choice): the new `wait` defaults to "block forever".

The timeout is optional with `default = zero`, and a zero timeout takes the plain `cv.wait(...)` branch below — i.e. block forever, which is precisely the regression this PR fixes. The only current caller passes `force_update_interval`, so the default path is dead code today, but any future caller that uses the default silently reintroduces the same infinite-block bug.

Note the irony: `bfd7003f326`'s signature `wait(std::chrono::milliseconds timeout, bool & finished)` made the timeout **mandatory**, so this class of bug was impossible to write; adding the default here gives that safety back up. Suggest keeping `timeout` mandatory — or at minimum, never defaulting it to a value that means "block forever".

src/Interpreters/ClusterDiscovery.cpp
```diff
@@ -698,7 +705,7 @@ bool ClusterDiscovery::runMainThread(std::function<void()> up_to_date_callback)
     while (!finished)
     {
         bool all_up_to_date = true;
-        auto clusters = clusters_to_update->wait(finished);
+        auto clusters = clusters_to_update->wait(finished, force_update_interval);
```
tiandiwonder, line 741: The 2-minute interval is a deliberate trade-off, not a regression — and the right call here (non-blocking note).

The bug's only real invariant is "the loop must not sleep unboundedly"; **any** finite timeout satisfies it, so `5s` vs `2 min` is a tuning choice, not a correctness one. Reusing `force_update_interval` is the cleaner option:

- It makes the **existing** "force-read every 2 min" logic (`watch.elapsedSeconds() > force_update_interval`) actually reachable, instead of adding a second independent 5s cycle that mostly spins without doing a force update.
- Lower background cost: every wakeup runs the full main loop (`findDynamicClusters` + dynamic-cluster reconciliation); a 5s cadence is far more aggressive for many-node / many-cluster deployments, for a self-heal that only matters during a rare degraded-Keeper event.
- The previously-removed `cluster_discovery_wait_timeout` was never actually wired — `bfd7003f326` stored it in `wait_timeout_secs` but `runMainThread` still called `wait(5s, finished)` hardcoded — so not resurrecting that knob is correct cleanup, not an omission.

For this failure class recovery latency is not the binding constraint anyway: the linked incident took ~13 days to surface because the observer's in-memory cache masked the depleted topology; 2 min vs ~4 min worst case is negligible against that. **No change requested on the interval.**

(Minor, optional) Since the same `force_update_interval` is used both as the `wait` timeout and as the per-cluster force-update threshold, worst-case re-registration is ~`2 ×` it (≈4 min). Decoupling via a separate constant wait timeout is a nice-to-have, not required.

tiandiwonder:
Related PR #90341

tests/integration/test_cluster_discovery/test.py
```diff
@@ -177,6 +177,73 @@ def test_cluster_discovery_no_crash_on_empty_static_cluster(start_cluster):
...
+    cluster.wait_zookeeper_nodes_to_start(zk_nodes)
+
+    # With the fix, re-registration should happen within force_update_interval (2 min).
+    # Use retry loop instead of blind sleep to avoid unnecessarily slow test.
+    for attempt in range(24):  # 24 * 10s = 240s max
+        time.sleep(10)
+        try:
+            result = int(
+                observer.query(
+                    "SELECT count() FROM system.clusters WHERE cluster = 'test_auto_cluster'"
```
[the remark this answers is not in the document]
sharadgaur: The test now waits for the actual Keeper children under `/clickhouse/discovery/test_auto_cluster/shards` via `system.zookeeper`, so it no longer relies on the observer's cached `system.clusters`. 

After the Keeper-side registration count recovers, the test restarts the observer and verifies that a fresh observer sees all nodes in `system.clusters`.

tests/integration/test_cluster_discovery/test.py
sharadgaur: Done , I used both checks: first `system.zookeeper` for the source-of-truth registration count, then an observer restart followed by `system.clusters` to verify fresh discovery state.

src/Interpreters/ClusterDiscovery.cpp
sharadgaur: Good point, fixed. The timeout is mandatory now, so there is no default path that can silently fall back to blocking forever.

sharadgaur:
> I wonder if when the keeper conn is terminated and then reconnected it should also double check that the registration is still there. Otherwise there could be up to two minutes before the node is re-registered

Agreed, immediate revalidation on a new Keeper session would reduce recovery latency further.

I kept this PR scoped to restoring the existing bounded self-healing path, so recovery is no longer "never" and happens on the `force_update_interval`. If you prefer, I can either add reconnect-triggered revalidation here or follow up separately.

src/Interpreters/ClusterDiscovery.cpp
```diff
@@ -698,7 +702,7 @@ bool ClusterDiscovery::runMainThread(std::function<void()> up_to_date_callback)
...
```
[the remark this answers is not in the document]
sharadgaur: fixed,  newly discovered dynamic clusters are now added to `clusters_to_update` with `need_update=false` in the dynamic insert path. That keeps them in the periodic force-update set, so if their Keeper watch is later invalidated by session expiry, `upsertCluster()` can still run and refresh/remove stale membership.

I also extended `test_dynamic_clusters` to cover this case: `test_auto_cluster3` is discovered after observer startup, then Keeper is restarted to expire sessions, then the member is stopped. The test verifies the observer eventually drops the stale dynamic cluster membership instead of keeping it forever.

src/Interpreters/ClusterDiscovery.cpp
sharadgaur: Agreed, thanks. I kept `force_update_interval` as the wait timeout and did not bring back the old server setting.

tests/integration/test_cluster_discovery/test_dynamic_clusters.py
```diff
@@ -93,3 +120,24 @@ def test_cluster_discovery_startup_and_stop(start_cluster):
...
+    # test_auto_cluster3 was discovered dynamically after observer startup. It must be kept
+    # in the observer's periodic update set, otherwise a Keeper session expiry can invalidate
+    # its children watch and leave the observer with stale membership forever.
+    zk_nodes = ["zoo1", "zoo2", "zoo3"]
+    cluster.stop_zookeeper_nodes(zk_nodes)
+    time.sleep(30)
+    cluster.start_zookeeper_nodes(zk_nodes)
+    cluster.wait_zookeeper_nodes_to_start(zk_nodes)
+
+    wait_for_registration_count(
```
tiandiwonder: Race: after the Keeper restart this only waits for `test_auto_cluster3` to re-register, but the restart expired the ephemeral znodes of **all three** dynamic clusters. `test_auto_cluster1` (node0) and `test_auto_cluster2` (node2) re-register on their own nodes' independent ~2-4 min self-heal timers and are not waited on here. Combined with the `len`-based early return in `get_clusters_hosts` (see other comment), `get_clusters_hosts(observer, 2)` below can return on a wrong transient 2-set, e.g. `{cluster1/node0, cluster3/node3}` (cluster3 not yet dropped after the node3 kill, cluster2 not yet back), so `assert clusters == expect6` fails. This matches the current CI failure.

Fix: after the restart, wait for **all** relevant registrations to recover before killing node3 — `/clickhouse/discovery2/test_auto_cluster1/shards`, `/clickhouse/discovery2/test_auto_cluster2/shards`, and `/clickhouse/discovery2/test_auto_cluster3/shards`.

(The product side is correct — `set(cluster_name, false)` does make all three self-heal — they just converge at independent times; the test asserts before convergence.)

tests/integration/test_cluster_discovery/test_dynamic_clusters.py
```diff
@@ -34,18 +34,45 @@ def start_cluster():
...
 
 
-def get_clusters_hosts(node, expected):
-    count = 30
+def get_clusters_hosts(node, expected, retries=30):
     while True:
         resp = node.query("SELECT cluster, host_name FROM system.clusters ORDER BY cluster, host_name FORMAT JSONCompact")
         hosts = json.loads(resp)["data"]
-        if count <= 0 or len(hosts) == expected:
+        if retries <= 0 or len(hosts) == expected:
```
tiandiwonder, line 41: This early-returns as soon as `len(hosts) == expected`, not when the membership actually matches. During the multi-cluster recovery transient above, the row count can hit 2 with the wrong set (e.g. `{cluster1, cluster3}`), and `get_clusters_hosts(observer, 2, retries=300)` then returns that wrong 2-set, failing `assert clusters == expect6`.

Fix: wait until `hosts == expected` (full set match), not just `len(hosts) == expected`. With an exact-match poll plus the 300s budget it converges even through the cluster3-removal transient. This is the core correctness fix.

tests/integration/test_cluster_discovery/test_dynamic_clusters.py
sharadgaur: Thanks, fixed in 4d47bc501ad.

After the Keeper restart, the test now waits for all relevant registrations to recover before killing node3. Small note: based on the test configs, test_auto_cluster1 and test_auto_cluster2 register under `/clickhouse/discovery/...`, while test_auto_cluster3 registers under `/clickhouse/discovery2/...`, so I used those concrete paths.

tests/integration/test_cluster_discovery/test_dynamic_clusters.py
sharadgaur: Fixed in 4d47bc501ad.

I added an exact-match polling helper for `system.clusters`, so this no longer returns just because the row count matches. It now waits for the full expected membership set before asserting.

tiandiwonder:
@sharadgaur please don't force push, thanks.

sharadgaur:
> @sharadgaur please don't force push, thanks.

got it, I was keeping the PR as a single clean commit while addressing review comments.

tiandiwonder (approved):

sharadgaur:
@tiandiwonder thank you for the review.

It looks like merge is currently blocked on CI. The remaining failures I see are around `Upgrade check (amd_release)` / `CH Inc sync`, and they do not look related to this Keeper session re-registration change to me, but I might be missing something.

Is there anything I should fix on my side to make CI green, or should we rerun the failed checks? I can also rebase if that is preferred.

src/Interpreters/ClusterDiscovery.cpp
```diff
@@ -779,6 +783,9 @@ bool ClusterDiscovery::runMainThread(std::function<void()> up_to_date_callback)
                 continue;
             }
             auto & cluster_info = cluster_info_it->second;
+            /// Keep newly discovered dynamic clusters in the periodic update set so they can self-heal
+            /// if their ZooKeeper watches are later lost after a Keeper session expiry.
+            clusters_to_update->set(cluster_name, false);
```
tiandiwonder, line 834: the self-heal still has one unreachable corner: a one-node dynamic cluster that goes empty before the session expiry. `removeCluster` drops the `clusters_to_update` entry and watch callback but leaves the stale `clusters_info` record, so after expiry the root rescan dedups the still-existing znode against that residue — it lands in neither `clusters_to_insert` nor the update set, and when the node returns nothing notices (`node -> empty -> session expiry -> node returns` never recovers). erasing the `clusters_info` entry in `removeCluster`'s dynamic path would let the next scan rediscover it (and also fixes remove-then-recreate-same-name); worth a regression test walking that exact sequence.

tiandiwonder (changes requested):

src/Interpreters/ClusterDiscovery.cpp
sharadgaur: `removeCluster` dynamic path now erases the `clusters_info` record, so the next root scan rediscovers the cluster as a fresh dynamic insert. That covers `node -> empty -> session expiry -> node returns` and also remove-then-recreate-same-name. Added `test_dynamic_cluster_recovers_after_empty_then_session_expiry` walking that exact sequence, it times out without the fix since recovery there depends purely on the root rescan (the member re-registering doesn't change the root's children, so no watch fires).

The plain erase needed a few things on top:

- `upsertCluster` returns `Updated/NeedRetry/Empty` instead of bool and the caller does the removal after the call returns, so no caller is left holding a dangling `ClusterInfo&`. `removeCluster` takes the name by value for the same reason, callers pass `cluster_info.name` which lives inside the map node being erased.
- `initialUpdate` iterates a snapshot of names since empty dynamic clusters can now erase their record during startup.
- a child watch registered before the removal still fires on session expiry and re-adds a flags entry for the erased cluster. The update loop defers to the insert pass when the same name was just rediscovered, and drops orphaned flags entries instead of logging "Unknown cluster" on every wake.
- empty static clusters keep the previous force-update cadence (watch restarted before removal).

PR #106115, opened 2026-05-29
Add path validation in delta lake

scanhex12:

Continuation of https://github.com/ClickHouse/ClickHouse/pull/96754 but for delta lake

### Version info
- Merged into: `26.6.1.380` (included in `26.6` and later)
- Backported to: `26.5.2.39`, `26.4.4.38`, `26.3.17.76`, `25.8.29.33`


src/Storages/ObjectStorage/DataLakes/Common/Common.cpp
```diff
@@ -39,4 +47,21 @@ std::vector<String> listFiles(
...
+{
+    auto norm_base = std::filesystem::path(table_path).lexically_normal();
+    auto combined = (norm_base / relative_path).lexically_normal();
+
+    auto rel = combined.lexically_relative(norm_base);
+
+    if (rel.empty() || rel.begin()->string() == "..")
+        throw Exception(
+            ErrorCodes::PATH_ACCESS_DENIED,
+            "Data lake path `{}` should be inside the table directory `{}`",
```
kssenii, line 60: May be use `pathStartsWith` from `filesystemHelpers.h` which is usually used for similar checks in our code?

kssenii (approved):

clickgapai:
Hi @scanhex12 @kssenii — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: The description is essentially the changelog template with no explanation of the vulnerability (crafted Delta log `add.path` traversal reading arbitrary files), how `resolvePathInsideTable` enforces containment, or why both Delta readers are covered.
- Suggested addition:

  > Explain that a malicious `_delta_log` could set `add.path` to a relative (`..`) or absolute path outside the table directory, causing `deltaLake`/`deltaLakeLocal` to read arbitrary files. The fix introduces `resolvePathInsideTable`, which rejects any resolved path not contained in the table directory (via `pathStartsWith`), and applies it to both the delta-kernel-rs reader and the legacy `DeltaLakeMetadata` reader, plus the Iceberg explicit-metadata path.

PR #106116, opened 2026-05-29
FileCache metrics improvement

kssenii:
- Improvement

FileCache metrics improvement

### Version info
- Merged into: `26.6.1.517`
- Backported to: `26.5.3.3`


antonio2368 (approved):

clickgapai:
Hi @kssenii @antonio2368 — a couple of suggestions to make this PR easier for reviewers and the changelog reader.

- Why: PR body is empty apart from the changelog template; it does not explain the metric rename (which renames a `system.metric_log` column) nor the new `FilesystemCachePriorityQueueElements`/`FilesystemCacheInvalidatedElements` gauges and the Main-vs-Query queue gating.
- Suggested addition:

  > Describe: (1) the rename of ProfileEvent `FilesystemCacheEvictedFileSegmentsDuringPriorityIncrease` to `FilesystemCacheDowngradedFileSegments` and that it now counts protected->probationary downgrades rather than evictions; (2) the two new CurrentMetrics gauges and that they only track the global (Main) priority queue, not per-query queues; (3) that eviction byte accounting now uses reserved size.

